PR #1077, opened 2023-11-03
FIX: Skip adaption prompt tests with new transformers versions

BenjaminBossan:
Adaption prompt is failing with transformers v4.35.0. This PR skips the adaption prompt tests so that CI is green again. The PR also adds an error when users try to use adaption prompt with that version, instructing them to use an older transformers version instead.

This should be removed as soon as the issue is fixed in PEFT/transformers.


younesbelkada (approved):
Makes sense, thank you very much @BenjaminBossan !

PR #1078, opened 2023-11-03
Enable Dynamic Resizing of Model Embedding Matrix during merge_and_unload() in PEFT Models

mrm8488:
This pull request introduces a new feature that addresses a common scenario in the fine-tuning of Large Language Models (LLMs): the addition of new special tokens to the tokenizer. Traditionally, when new tokens are introduced, the model's embedding matrix needs to be expanded to accommodate the updated vocabulary size, which could be cumbersome and error-prone.

To streamline this process, we've added the functionality to dynamically resize the model's embedding matrix within the merge_and_unload() method of peft models. Users can now simply pass the updated vocabulary size as a parameter to the method. The merge_and_unload() function will then adjust the embedding matrix dimensions accordingly.

cc: @younesbelkada ;)

### DIFF
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -376,7 +376,9 @@ class LoraModel(BaseTuner):
             )
         return peft_config
 
-    def _unload_and_optionally_merge(self, merge=True, progressbar: bool = False, safe_merge: bool = False):
+    def _unload_and_optionally_merge(
+        self, merge=True, progressbar: bool = False, safe_merge: bool = False, new_vocab_length: int = 0
+    ):
         if merge:
             if getattr(self.model, "quantization_method", None) == "gptq":
                 raise ValueError("Cannot merge LORA layers when the model is gptq quantized")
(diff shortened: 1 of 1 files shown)


pacman100:
Hello @mrm8488, thank you for the PR. I am missing the context here. The expanded embedding matrix will have random weights for the added tokens. How will that help, trying to understand the usage of the feature being added?

pacman100 (commented):
Hello @mrm8488, thank you for the PR. I am missing the context here. The expanded embedding matrix will have random weights for the added tokens. How will that help, trying to understand the usage of the feature being added?

mrm8488:
Hi, @pacman100. I agree with you. Did not take it into account. The problem here is that when I resize the embedding matrix because I have added special tokens for the training (fine-tuning), once it has finished, the peft model (adapter) does not store such info and ofc when you do `merge_and_upload()` this new "condition" is not taken into account. How can we deal with such case? Mmm, maybe after resizing the embedding matrix I have to upload the model to the hub and fine-tune the resized model, that way, on merge and upload the base model will be the right one :)

PR #1085, opened 2023-11-06
[`Docker`] Update Dockerfile to force-use transformers main

younesbelkada:
# What does this PR do?

Optimum pins transformers to <4.35.0, leading our docker images to have transformers==4.34.1, which is not intended since we want to test PEFT against transformers and accelerate main

This PR fixes it, there is no need to update the cpu docker file as optimum is not installed there

cc @pacman100 @BenjaminBossan

### DIFF
--- a/docker/peft-gpu/Dockerfile
+++ b/docker/peft-gpu/Dockerfile
@@ -29,15 +29,6 @@ ENV PATH /opt/conda/envs/peft/bin:$PATH
 # Activate our bash shell
 RUN chsh -s /bin/bash
 SHELL ["/bin/bash", "-c"]
-# Activate the conda env and install transformers + accelerate from source
-RUN source activate peft && \
-    python3 -m pip install --no-cache-dir \
-    librosa \
-    "soundfile>=0.12.1" \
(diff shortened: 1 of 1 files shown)


younesbelkada:
@BenjaminBossan as far as I know we've been always testing on transformers main: https://github.com/huggingface/peft/blob/main/docker/peft-gpu/Dockerfile#L38 except until the latest optimum release

BenjaminBossan:
Oh, so this is more of a problem of the order in which the packages are installed. Is there a reason why `bitsandbytes optimum auto-gptq` are installed in a separate block further down below? Maybe they should be installed in the first block, and then we add a second block to install every package that needs to be installed from source? Otherwise, we are prone to run into similar issues in the future if any dependency adds another pin to a version.

younesbelkada:
Makes sense yes, let me update that as well

BenjaminBossan (approved):
Thanks for investigating this issue and providing a solution, looks good.

younesbelkada:
Thanks @BenjaminBossan !

PR #995, opened 2023-10-05
TST: Add regression tests

BenjaminBossan:
This is a first step towards adding regression tests to the project. These tests allow us to load old adapters with new PEFT versions and ensure that the output generated by the model does not change.

The PR includes a framework for how to add regression artifacts and how to run regression tests based on those artifacts. Right now, only bnb + LoRA is covered, but it should be straightforward to add more tests.

## Description

In general, for regression tests, we need two steps:

1. Creating the regression artifacts, in this case the adapter checkpoint and the expected output of the model.
2. Running the regression tests, i.e. loading the adapter and checking that the output of the model is the same as the expected output.

My approach is to re-use as much code as possible between those two steps. Therefore, the same test script can be used for both, with only an environment variable to distinguish between the two. Step 1 is invoked by calling:

`REGRESSION_CREATION_MODE=True pytest tests/regression/test_regression.py`

and to run the second step, we call:

`pytest tests/regression/test_regression.py`

### Creating regression artifacts

The first step will create an adapter checkpoint and an output for the given PEFT version and test setting in a new directory. E.g. it will create a directory `tests/regression/lora_opt-125m_bnb_4bit/0.5.0/` that contains adapter_model.bin and output.pt.

Before this step runs, there is a check that the git repo is clean (no dirty worktree) and that the commit is tagged (i.e. corresponds to a release version of PEFT). Otherwise, we may accidentally create regression artifacts that do not correspond to any PEFT release.

The easiest way to get such a clean state (say, for PEFT v0.5.0) is by checking out a tagged commit, e.g:

`git checkout v0.5.0`

before running the first step.

The first step will also skip the creation of regression artifacts if they already exist.

It is possible to circumvent all the aforementioned checks by setting the environment variable `REGRESSION_FORCE_MODE` to True like so:

`REGRESSION_FORCE_MODE=True REGRESSION_CREATION_MODE=True pytest tests/regression/test_regression.py`

You should only do this if you know exactly what you're doing.

###  Running regression tests

The second step is much simpler. It will load the adapters and the output created in the first step, and compare the output to the output from a new PEFT model using the loaded adapter. The outputs should be the same.

If more than one version is discovered for a given test setting, all of them are tested.

## Notes

As is, the adapters are stored in the git repo itself. Since they're relatively small, the total size of the repo is still reasonable. However, it could be better to store those adapters on HF Hub instead. This would, however, make things a bit more complicated (not sure how to parse directories etc. on Hub).

[...]

PR #1088, opened 2023-11-07
FEAT: Make safe serialization the default one

younesbelkada:
# What does this PR do?

As per title, to be in line with the current efforts in the OSS ecosystem let's also make in PEFT safe serialization the default behaviour

Adapted the tests and added some regression tests! 

cc @BenjaminBossan @pacman100

### DIFF
--- a/tests/testing_common.py
+++ b/tests/testing_common.py
@@ -269,7 +269,7 @@ class PeftCommonTester:
 
         self.assertTrue(dummy_output.requires_grad)
 
-    def _test_save_pretrained_regression(self, model_id, config_cls, config_kwargs):
+    def _test_save_pretrained(self, model_id, config_cls, config_kwargs, safe_serialization=True):
         # ensure that the weights are randomly initialized
         if issubclass(config_cls, LoraConfig):
             config_kwargs = config_kwargs.copy()
@@ -287,7 +287,10 @@ class PeftCommonTester:
         model = model.to(self.torch_device)
 
         with tempfile.TemporaryDirectory() as tmp_dirname:
-            model.save_pretrained(tmp_dirname, safe_serialization=False)
+            if safe_serialization:
+                model.save_pretrained(tmp_dirname)
+            else:
+                model.save_pretrained(tmp_dirname, safe_serialization=False)
 
             model_from_pretrained = self.transformers_class.from_pretrained(model_id)
             model_from_pretrained = PeftModel.from_pretrained(model_from_pretrained, tmp_dirname)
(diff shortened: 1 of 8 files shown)


tests/testing_common.py
```diff
@@ -311,14 +368,14 @@ def _test_save_pretrained(self, model_id, config_cls, config_kwargs):
...
-            self.assertTrue(os.path.exists(os.path.join(tmp_dirname, "adapter_model.bin")))
+            # check if `adapter_model.safetensors` is present
+            self.assertTrue(os.path.exists(os.path.join(tmp_dirname, "adapter_model.safetensors")))
 
             # check if `adapter_config.json` is present
             self.assertTrue(os.path.exists(os.path.join(tmp_dirname, "adapter_config.json")))
 
-            # check if `pytorch_model.bin` is not present
-            self.assertFalse(os.path.exists(os.path.join(tmp_dirname, "pytorch_model.bin")))
+            # check if `model.safetensors` is not present
```
younesbelkada, line 377: This change in necessary as now in transformers we don't save anymore `pytorch-model.bin` but the safetensors model

BenjaminBossan (commented):
No full review yet, but some general questions:

1. Do we want to do a deprecation/future warning cycle before making safetensors the default? How was it done in transformers?
2. Do we want to run `_test_save_pretrained` twice, once with safetensors, once with pytorch?

younesbelkada:
Thanks @BenjaminBossan !
1. not sure how they did it in transformers, let me double check that

> 2. Do we want to run _test_save_pretrained twice, once with safetensors, once with pytorch?

Yes this is what I did in the PR, I kept `test_save_pretrained` as is and replaced the expected file names with safetensors file names and added a new test to check if one calls `save_pretrained` with `safe_serialization=False` it leads to the same behavior as we had before this PR

BenjaminBossan:
> I kept `test_save_pretrained` as is and replaced the expected file names with safetensors file names and added a new test to check if one calls `save_pretrained` with `safe_serialization=False` it leads to the same behavior as we had before this PR

Oh I see. I wonder if `_test_save_pretrained` could not be changed slightly so that it takes an argument `use_safetensors` or something like that which can be parameterized. Then the same test can be re-used without duplicating it whole.

I think the new test is not quite a regression test, as regression testing would require the model to be persisted with one PEFT version and loaded with another PEFT version. I think we should really try to advance #995, which adds regression tests which would also help cover this case.

tests/testing_common.py
```diff
@@ -287,7 +287,10 @@ def _test_save_pretrained(self, model_id, config_cls, config_kwargs):
         model = model.to(self.torch_device)
 
         with tempfile.TemporaryDirectory() as tmp_dirname:
-            model.save_pretrained(tmp_dirname)
+            if safe_serialization:
```
younesbelkada, line 290: I prefer to not pass `safe_serialization=safe_serialization` to test the native behaviour, let me know if this makes sense

younesbelkada: yes sorry I meant the default behaviour

younesbelkada:
We are post the patch release now

BenjaminBossan (approved):
Thanks for making the switch and adjusting the tests. I think we can merge now, although I would imagine some users will be surprised that their code suddenly produces safetensors. We should put a big note about this on the release notes once we publish the next version.

PR #1092, opened 2023-11-08
Parallel linear Lora

zhangsheng377:
We implemented the Lora algorithm for megatron's distributed layer ColumnParallelLinear and RowParallelLinear.

Due to the particularity of megatron creating a distributed layer, the required megatron information needs to be injected before executing Lora:
```python
from megatron.arguments import core_transformer_config_from_args
from megatron import get_args

config = core_transformer_config_from_args(get_args())

lora_config = LoraConfig(
    r=16,
    lora_alpha=32,
    target_modules=['query_key_value', 'dense', 'dense_h_to_4h', 'dense_4h_to_h'],
    lora_dropout=0.0,
    bias="none",
    megatron_config=config,
    megatron_core="megatron.core",
)
```

It has been verified on the megatron and megatron-deepspeed frameworks.

### DIFF
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -117,6 +117,32 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    megatron_config: Optional[dict] = field(
+        default=None,
+        metadata={
+            "help": (
+                "The TransformerConfig from Megatron, it is used to create LoRA's parallel linear layer."
+                "You can get it like this, `core_transformer_config_from_args(get_args())`, "
+                "this two functions are from Megatron."
+                "You need to specify this parameter when you want to loraize the ColumnParallelLinear and "
+                "RowParallelLinear layers of megatron."
+                "It should be noted that we may not be able to use the `save_pretrained` and `from_pretrained` "
+                "functions, because TransformerConfig may not necessarily be serialized."
+                "But when using megatron, we can use `get_peft_model_state_dict` function and "
+                "megatron's framework, they can also save and load models and configurations."
+            )
+        },
+    )
+    megatron_core: Optional[str] = field(
+        default="megatron.core",
+        metadata={
+            "help": (
+                "The core module from Megatron, it is used to judge and create LoRA's parallel linear layer. "
+                "It only needs to be passed in when you need to use your own modified megatron core module. "
+                "Otherwise, it will use the default value `megatron.core`. "
+            )
+        },
+    )
 
     def __post_init__(self):
         self.peft_type = PeftType.LORA
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import importlib
 import math
 import operator
 import re
@@ -259,6 +260,10 @@ class LoraModel(BaseTuner):
         else:
             target_base_layer = target
 
+        megatron_core = None
+        if lora_config.megatron_config:
+            megatron_core = importlib.import_module(lora_config.megatron_core)
+
         if loaded_in_8bit and isinstance(target_base_layer, bnb.nn.Linear8bitLt):
             eightbit_kwargs = kwargs.copy()
             eightbit_kwargs.update(
@@ -297,6 +302,28 @@ class LoraModel(BaseTuner):
                 )
                 kwargs["fan_in_fan_out"] = lora_config.fan_in_fan_out = False
             new_module = Linear(target, adapter_name, **kwargs)
+        elif megatron_core and isinstance(
+            target_base_layer,
+            (megatron_core.tensor_parallel.ColumnParallelLinear, megatron_core.tensor_parallel.RowParallelLinear),
+        ):
+            from .tp_layer import LoraParallelLinear
+
+            megatron_kwargs = kwargs.copy()
+            megatron_config = lora_config.megatron_config
+            if isinstance(megatron_config, dict):
+                transformer_config_class = megatron_core.transformer.transformer_config.TransformerConfig
+                megatron_config = transformer_config_class(**lora_config.megatron_config)
+            megatron_kwargs["megatron_config"] = megatron_config
+            if megatron_kwargs["fan_in_fan_out"]:
+                warnings.warn(
+                    "fan_in_fan_out is set to True but the target module is `ColumnParallelLinear` "
+                    "or `RowParallelLinear`. "
+                    "Setting fan_in_fan_out to False."
+                )
+                megatron_kwargs["fan_in_fan_out"] = lora_config.fan_in_fan_out = False
+            new_module = LoraParallelLinear(
+                base_layer=target, adapter_name=adapter_name, backend=megatron_core.tensor_parallel, **megatron_kwargs
+            )
         elif isinstance(target_base_layer, Conv1D):
             if not kwargs["fan_in_fan_out"]:
                 warnings.warn(
--- a/src/peft/tuners/lora/tp_layer.py
+++ b/src/peft/tuners/lora/tp_layer.py
@@ -1,0 +1,158 @@
+from typing import Any
+
+import torch
+import torch.nn as nn
+import torch.nn.init as init
+
+from .layer import LoraLayer
+
+
+class LoraParallelLinear(nn.Module, LoraLayer):
+    """
+    When the target layer parallel_linear is RowParallelLinear, in order to keep the input and output shapes
+    consistent, we need to split the lora matrix A into rows, and the lora_B at this time should be a complete linear
+    layer; In the same way, when the target layer is ColumnParallelLinear, we perform column segmentation on lora_B,
+    while lora_A is still a complete linear layer.
+    """
+
+    def __init__(
+        self,
+        base_layer,
+        adapter_name: str,
+        backend,
+        r: int = 0,
+        lora_alpha: int = 1,
+        lora_dropout: float = 0.0,
+        fan_in_fan_out: bool = False,
+        init_lora_weights: bool = True,
+        **kwargs,
+    ):
+        super().__init__()
+        LoraLayer.__init__(self, base_layer=base_layer)
+
+        self.backend = backend
+        self.is_paralle_a = isinstance(base_layer, backend.RowParallelLinear)
+        self.fan_in_fan_out = fan_in_fan_out
+        self._active_adapter = adapter_name
+
+        megatron_config = kwargs["megatron_config"]
+        parallel_linear_kwargs = {"megatron_config": megatron_config}
+        init_method = init.xavier_normal_
+        if hasattr(megatron_config, "init_method"):
+            init_method = megatron_config.init_method
+        input_is_parallel = True
+        gather_output = False
+        if isinstance(base_layer, self.backend.RowParallelLinear):
+            input_is_parallel = base_layer.input_is_parallel
+        else:
+            gather_output = base_layer.gather_output
+        self.update_layer(
+            adapter_name,
+            r,
+            lora_alpha,
+            lora_dropout,
+            init_lora_weights,
+            init_method,
+            input_is_parallel,
+            gather_output,
+            **parallel_linear_kwargs,
+        )
+
+        self.is_target_conv_1d_layer = False
+
+    def update_layer(
+        self,
+        adapter_name,
+        r,
+        lora_alpha,
+        lora_dropout,
+        init_lora_weights,
+        init_method=init.xavier_normal_,
+        input_is_parallel=True,
+        gather_output=False,
+        **parallel_linear_kwargs,
+    ):
+        if r <= 0:
+            raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
+        self.r[adapter_name] = r
+        self.lora_alpha[adapter_name] = lora_alpha
+        if lora_dropout > 0.0:
+            lora_dropout_layer = nn.Dropout(p=lora_dropout)
+        else:
+            lora_dropout_layer = nn.Identity()
+
+        self.lora_dropout[adapter_name] = lora_dropout_layer
+
+        megatron_config = parallel_linear_kwargs["megatron_config"]
+        # lora needs to be forced to upgrade to 32-bit precision, otherwise it will overflow
+        megatron_config.params_dtype = torch.float32
+        if self.is_paralle_a:
+            lora_a = self.backend.RowParallelLinear(
+                input_size=self.in_features,
+                output_size=r,
+                bias=False,
+                input_is_parallel=input_is_parallel,
+                skip_bias_add=True,
+                init_method=init_method,
+                config=megatron_config,
+            )
+            lora_b = nn.Linear(in_features=r, out_features=self.out_features, bias=False, dtype=torch.float32)
+        else:
+            lora_a = nn.Linear(in_features=self.in_features, out_features=r, bias=False, dtype=torch.float32)
+            lora_b = self.backend.ColumnParallelLinear(
+                input_size=r,
+                output_size=self.out_features,
+                bias=False,
+                gather_output=gather_output,
+                init_method=init_method,
+                config=megatron_config,
+            )
+        self.lora_A[adapter_name] = lora_a
+        self.lora_B[adapter_name] = lora_b
+        self.scaling[adapter_name] = lora_alpha / r
+        if init_lora_weights:
+            self.reset_lora_parameters(adapter_name)
+
+        weight = getattr(self.get_base_layer(), "weight", None)
+        if weight is not None:
+            # the layer is already completely initialized, this is an update
+            if weight.dtype.is_floating_point or weight.dtype.is_complex:
+                self.to(weight.device, dtype=weight.dtype)
+            else:
+                self.to(weight.device)
+        self.set_adapter(self.active_adapters)
+
+    def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any):
+        previous_dtype = x.dtype
+        # If weight is used for matrix multiplication here, the final aggregation operation of the original
+        # parallel_linear layer will be missing, so we need to directly call its forward function to obtain the
+        # output of the original parallel_linear layer.
+        if self.disable_adapters:
+            if self.merged:
+                self.unmerge()
+            result, bias = self.base_layer(x, *args, **kwargs)
+        elif self.merged:
+            result, bias = self.base_layer(x, *args, **kwargs)
+        else:
+            result, bias = self.base_layer(x, *args, **kwargs)
+            for active_adapter in self.active_adapters:
+                if active_adapter not in self.lora_A.keys():
+                    continue
+                lora_A = self.lora_A[active_adapter]
+                lora_B = self.lora_B[active_adapter]
+                dropout = self.lora_dropout[active_adapter]
+                scaling = self.scaling[active_adapter]
+                x = x.to(lora_A.weight.dtype)
+
+                lora_result = lora_A(dropout(x))
+                if isinstance(lora_result, tuple):
+                    lora_result = lora_result[0]
+                lora_result = lora_B(lora_result)
+                if isinstance(lora_result, tuple):
+                    lora_result = lora_result[0]
+                lora_result = lora_result * scaling
+
+                result = result + lora_result
+
+        result = result.to(previous_dtype)
+        return result, bias
--- a/tests/test_lora_megatron.py
+++ b/tests/test_lora_megatron.py
@@ -1,0 +1,167 @@
+#!/usr/bin/env python3
+
+# coding=utf-8
+# Copyright 2023-present the HuggingFace Inc. team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import copy
+import importlib
+import os
+import unittest
+
+import torch
+import torch.nn.init as init
+
+from peft import LoraConfig, PeftModel, get_peft_model, get_peft_model_state_dict
+
+
+def is_megatron_available() -> bool:
+    return importlib.util.find_spec("megatron") is not None
+
+
+if is_megatron_available():
+    from megatron.core import parallel_state, tensor_parallel
+    from megatron.core.tensor_parallel.random import model_parallel_cuda_manual_seed
+    from megatron.core.transformer.module import MegatronModule
+    from megatron.core.transformer.transformer_config import TransformerConfig
+
+    world_size = 1
+    rank = 0
+
+    def initialize_distributed():
+        print(f"Initializing torch.distributed with rank: {rank}, world_size: {world_size}")
+        torch.cuda.set_device(0)
+        init_method = "tcp://"
+        master_ip = os.getenv("MASTER_ADDR", "localhost")
+        master_port = os.getenv("MASTER_PORT", "6001")
+        init_method += master_ip + ":" + master_port
+        torch.distributed.init_process_group(backend="nccl", world_size=world_size, rank=rank, init_method=init_method)
+
+    def destroy_model_parallel():
+        parallel_state.destroy_model_parallel()
+        torch.distributed.barrier()
+
+    def initialize_model_parallel(
+        tensor_model_parallel_size=1,
+        pipeline_model_parallel_size=1,
+        virtual_pipeline_model_parallel_size=None,
+        pipeline_model_parallel_split_rank=None,
+    ):
+        parallel_state.destroy_model_parallel()
+        if not torch.distributed.is_initialized():
+            initialize_distributed()
+        parallel_state.initialize_model_parallel(
+            tensor_model_parallel_size,
+            pipeline_model_parallel_size,
+            virtual_pipeline_model_parallel_size,
+            pipeline_model_parallel_split_rank,
+        )
+
+    class DummyModule(MegatronModule):
+        def __init__(self, config: TransformerConfig):
+            super().__init__(config)
+            self.linear = tensor_parallel.ColumnParallelLinear(
+                input_size=10,
+                output_size=10,
+                config=config,
+                init_method=init.xavier_normal_,
+                bias=False,
+                gather_output=False,
+            )
+            self.lm_head = tensor_parallel.RowParallelLinear(
+                input_size=10,
+                output_size=10,
+                config=config,
+                init_method=init.xavier_normal_,
+                bias=False,
+                input_is_parallel=True,
+            )
+
+        def forward(self, input):
+            x = self.linear(input)[0]
+            x = self.lm_head(x)[0]
+            return x
+
+    class TestMegatronLora(unittest.TestCase):
+        def setUp(self):
+            initialize_model_parallel(1, 1)
+            model_parallel_cuda_manual_seed(123)
+            transformer_config = {
+                "num_layers": 2,
+                "hidden_size": 12,
+                "num_attention_heads": 4,
+                "use_cpu_initialization": True,
+            }
+            config = TransformerConfig(**transformer_config)
+            self.megatron_module = DummyModule(config=config).cuda()
+            self.dummy_module = copy.deepcopy(self.megatron_module).cuda()
+
+            lora_config = LoraConfig(
+                lora_alpha=16,
+                lora_dropout=0.1,
+                r=64,
+                bias="none",
+                target_modules=["linear", "lm_head"],
+                megatron_config=config,
+                megatron_core="megatron.core",
+            )
+            self.megatron_module = get_peft_model(self.megatron_module, lora_config)
+
+        def tearDown(self):
+            destroy_model_parallel()
+
+        def test_megatron_lora_module(self):
+            megatron_module = self.megatron_module
+            self.assertTrue(isinstance(megatron_module, PeftModel))
+
+            for name, module in megatron_module.named_modules():
+                if name.endswith("linear"):
+                    self.assertTrue(hasattr(module, "lora_A"))
+                    self.assertTrue(hasattr(module, "lora_B"))
+                if name.endswith("linear.lora_A.default"):
+                    self.assertTrue(isinstance(module, torch.nn.Linear))
+                if name.endswith("linear.lora_B.default"):
+                    self.assertTrue(isinstance(module, tensor_parallel.ColumnParallelLinear))
+
+                if name.endswith("lm_head.lora_A.default"):
+                    self.assertTrue(isinstance(module, tensor_parallel.RowParallelLinear))
+                if name.endswith("lm_head.lora_B.default"):
+                    self.assertTrue(isinstance(module, torch.nn.Linear))
+
+        def test_forward(self):
+            x = torch.ones((2, 4, 10)).cuda()
+            megatron_module_result = self.megatron_module(x)
+            dummt_module_result = self.dummy_module(x)
+
+            # Because lora_B is initialized with 0, the forward results of two models should be equal before backward.
+            self.assertTrue(megatron_module_result.equal(dummt_module_result))
+
+        def test_backward(self):
+            optimizer = torch.optim.AdamW(self.megatron_module.parameters())
+            loss_fn = torch.nn.CrossEntropyLoss()
+
+            x = torch.randn(2, 4, 10, requires_grad=True).cuda()
+            label = torch.randint(10, (2 * 4,)).cuda()
+
+            output = self.megatron_module(x)
+            output = output.reshape(2 * 4, 10)
+            loss = loss_fn(output, label)
+
+            loss.backward()
+            optimizer.step()
+
+        def test_get_peft_model_state_dict(self):
+            peft_state_dict = get_peft_model_state_dict(self.megatron_module)
+
+            for key in peft_state_dict.keys():
+                self.assertTrue("lora" in key)
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -61,6 +61,9 @@ class LoraLayer(BaseTunerLayer):
         elif hasattr(base_layer, "infeatures") and hasattr(base_layer, "outfeatures"):
             # QuantLinear
             in_features, out_features = base_layer.infeatures, base_layer.outfeatures
+        elif hasattr(base_layer, "input_size") and hasattr(base_layer, "output_size"):
+            # Megatron ColumnParallelLinear,RowParallelLinear
+            in_features, out_features = base_layer.input_size, base_layer.output_size
         else:
             raise ValueError(f"Unsupported layer type {type(base_layer)}")
 


zhangsheng377:
@BenjaminBossan

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,155 @@
...
+            if self.is_paralle_a:
+                lora_a = tensor_parallel.RowParallelLinear(input_size=self.in_features, output_size=r, bias=False,
+                                                           input_is_parallel=kwargs.get('input_is_parallel', True),
+                                                           skip_bias_add=True, init_method=init_method,
+                                                           config=config)
+                lora_b = nn.Linear(in_features=r, out_features=self.out_features, bias=False, dtype=torch.float32)
+            else:
+                lora_a = nn.Linear(in_features=self.in_features, out_features=r, bias=False, dtype=torch.float32)
+                lora_b = tensor_parallel.ColumnParallelLinear(input_size=r, output_size=self.out_features, bias=False,
+                                                              gather_output=kwargs.get('gather_output', False),
```
fearblackcat, line 44: why can't  use lora_a and lora_b both tensor_parallel

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,155 @@
...
+            result, bias = self.parallel_linear_class.forward(self, x)
+            for active_adapter in self.active_adapters:
+                if active_adapter not in self.lora_A.keys():
+                    continue
+                lora_A = self.lora_A[active_adapter]
+                lora_B = self.lora_B[active_adapter]
+                dropout = self.lora_dropout[active_adapter]
+                scaling = self.scaling[active_adapter]
+                x = x.to(lora_A.weight.dtype)
+
```
fearblackcat, line 143: not see amp_02 with apex for layer norm
linear_in can use ColumnParallelLinear 
linear_out can use  RowParallelLinear if gather_out

src/peft/tuners/lora/tp_layer.py
zhangsheng377: > why can't use lora_a and lora_b both tensor_parallel

If the target module is RowParallelLinear, the input maybe is parallel, so the lora_a must be RowParallelLinear. At this time, the lora_a's output is complete matrix, although lora_b can be linear or tensor_parallel, but if use tensor_parallel, we will add one more scatter op. Because of lora matrix is small, the efficiency may not necessarily be improved.

And if the target module is ColumnParallelLinear, unless we judge the output is or not gather_out, the most convenient way is lora_b is ColumnParallelLinear, and then the lora_a can be linear or tensor_parallel. But as the first reason, lora matrix is small, if we use linear, the more copy in gpu has little influence, but if we use tensor_parallel, we will add one more scatter op, the communication overhead is big, so the efficiency may not necessarily be improved.

src/peft/tuners/lora/tp_layer.py
zhangsheng377: > not see amp_02 with apex for layer norm linear_in can use ColumnParallelLinear linear_out can use RowParallelLinear if gather_out

As the first reason, lora matrix is small, so if we use linear, the more copy in gpu has little influence, but if we use tensor_parallel, we will add one more scatter op, the communication overhead is big, so the efficiency may not necessarily be improved.

BenjaminBossan:
Great, thanks for the PR.

Before going through a full review, I have some points/questions:

Could you please provide a bit of more context what users can expect when using this functionality?

> Due to the particularity of megatron creating a distributed layer, the required megatron information needs to be injected before executing Lora:

It would be great if we could find a way to avoid that, I can also check later if I have any ideas.

Finally, it would be great to have unit tests for the new feature, or at the very least an example to see it in action.

zhangsheng377:
> Great, thanks for the PR.
> 
> Before going through a full review, I have some points/questions:
> 
> Could you please provide a bit of more context what users can expect when using this functionality?

For llama or other large models, the deepspeed framework is not easy to use if the model is too large. We now use our own modified megatron-deepspeed framework to train the cluster. Therefore, the model structure uses the ParallelLinear of megatron, but at this time we want to use Lora for fine-tuning, so we want to extend lora to support the ParallelLinear.




> 
> > Due to the particularity of megatron creating a distributed layer, the required megatron information needs to be injected before executing Lora:
> 
> It would be great if we could find a way to avoid that, I can also check later if I have any ideas.
> 
> Finally, it would be great to have unit tests for the new feature, or at the very least an example to see it in action.

Ok, I use this script to finetune a llama7B with alpaca on our own magetron-deepspeed framework :
```shell
export CUDA_VISIBLE_DEVICES_=0,1,2,3
export ASCEND_RT_VISIBLE_DEVICES=${CUDA_VISIBLE_DEVICES_}
ASCEND_RT_VISIBLE_DEVICES_ARRAY=(${CUDA_VISIBLE_DEVICES_//,/ })
echo "${ASCEND_RT_VISIBLE_DEVICES_ARRAY[@]}"

# the number of parameters is not aligned
export LD_LIBRARY_PATH=/usr/local/lib:/home/anaconda3/lib:$LD_LIBRARY_PATH
export HCCL_CONNECT_TIMEOUT=1200
export COMBINED_ENABLE=1
source /home/xxx/Ascend/set_env.sh

GPUS_PER_NODE=${#ASCEND_RT_VISIBLE_DEVICES_ARRAY[@]} 
echo $GPUS_PER_NODE
# Change for multinode config
MASTER_ADDR=localhost
MASTER_PORT=6001
NNODES=1
NODE_RANK=0
WORLD_SIZE=$(($GPUS_PER_NODE*$NNODES))

TP=1
PP=1

DATA_PATH=xxx
LOAD_CHECKPOINT_PATH=xxx
SAVE_CHECKPOINT_PATH=xxx
TOKENIZER_PATH=xxx

DS_CONFIG=deepspeed_config_13B_1.json
ZERO_STAGE=2

MICRO_BATCH=4
GRADIENT_ACCUMULATION_STEP=8
GLOBAL_BATCH=$(($MICRO_BATCH * $GRADIENT_ACCUMULATION_STEP * $WORLD_SIZE))
EPOCH=2
TRAIN_ITERS=$((52000 / $GLOBAL_BATCH * $EPOCH))
echo $TRAIN_ITERS
SAVE_INTERVAL=$(($TRAIN_ITERS / 2))
echo $SAVE_INTERVAL

export HCCL_OP_BASE_FFTS_MODE_ENABLE=TRUE

cat <<EOT > $DS_CONFIG
{
    "fp16": {
        "enabled": true,
        "loss_scale": 0,
        "loss_scale_window": 1000,
        "initial_scale_power": 8,
        "hysteresis": 2,
        "min_loss_scale": 1
    },

    "optimizer": {
        "type": "Adam"
    },

    "zero_optimization": {
        "stage": $ZERO_STAGE,
        "allgather_partitions": true,
        "allgather_bucket_size": 1e8,
        "overlap_comm": true,
        "reduce_scatter": true,
        "reduce_bucket_size": 1e8,
        "contiguous_gradients": true
    },

    "gradient_accumulation_steps": ${GRADIENT_ACCUMULATION_STEP},
    "train_batch_size": $GLOBAL_BATCH,
    "train_micro_batch_size_per_gpu":$MICRO_BATCH,
    "zero_allow_untested_optimizer": true
}
EOT

ds_args=""
ds_args=" --deepspeed ${ds_args}"
ds_args=" --no-pipeline-parallel ${ds_args}"
ds_args=" --deepspeed_config=$DS_CONFIG ${ds_args}"
ds_args=" --zero-stage=$ZERO_STAGE ${ds_args}"
ds_args=" --deepspeed-activation-checkpointing ${ds_args}"

#deepspeed --master_port ${MASTER_PORT} --include localhost:${CUDA_VISIBLE_DEVICES_} pretrain_llama.py \
deepspeed --master_port ${MASTER_PORT} pretrain_llama.py \
       --DDP-impl local \
       --no-contiguous-buffers-in-local-ddp \
       --tensor-model-parallel-size ${TP} \
       --pipeline-model-parallel-size ${PP} \
       --num-layers 32 \
       --hidden-size 4096 \
       --ffn-hidden-size 11008 \
       --num-attention-heads 32 \
       --micro-batch-size $MICRO_BATCH \
       --global-batch-size $GLOBAL_BATCH \
       --seq-length 1024 \
       --max-position-embeddings 2048 \
       --train-iters ${TRAIN_ITERS} \
       --lr-decay-iters ${TRAIN_ITERS} \
       --save $SAVE_CHECKPOINT_PATH \
       --load $LOAD_CHECKPOINT_PATH \
       --data-path $DATA_PATH \
       --tokenizer-name-or-path $TOKENIZER_PATH \
       --tokenizer-not-use-fast \
       --data-impl mmap \
       --split 949,50,1 \
       --distributed-backend nccl \
       --lr 2e-5 \
       --lr-decay-style cosine \
       --min-lr 0 \
       --weight-decay 0. \
       --clip-grad 1.0 \
       --lr-warmup-iters 100 \
       --checkpoint-activations \
       --log-interval 1 \
       --save-interval ${SAVE_INTERVAL} \
       --eval-interval 1000 \
       --eval-iters 10 \
       --use-cpu-initialization \
       --lora-target-modules query_key_value dense gate_proj up_proj down_proj \
       --lora-r 16 \
       --lora-alpha 32 \
       --is-instruction-dataset \
       --seed 42 \
       $ds_args \
       --optimizer fused_adam \
       --fp16 | tee logs/train_7B_deepspeed.log
```
We compare it with the model using torch.linear for lora fine-tuning, and the loss error is less than the absolute value 0.001.



This is the inference result of our model:
image

zhangsheng377:
@BenjaminBossan Add an unit test to test Lora for the Parallel linear.
You can run it in peft dir:
```shell
pytest tests/test_lora_megatron.py
```
And it is all passed in my env.

zhangsheng377:
@BenjaminBossan @pacman100

zhangsheng377:
@BenjaminBossan @pacman100 Sorry to bother you, but can you help me review it?

BenjaminBossan (changes requested):
Thanks for making some changes to integrate this better with LoRA and adding tests. I still think we need to find a better way to configure how this feature can be used, I made a suggestion in the comments. What do you think?

Furthermore, could you please run `make style`?

src/peft/tuners/lora/config.py
```diff
@@ -130,3 +130,13 @@ def __post_init__(self):
         # if target_modules is a regex expression, then layers_pattern should be None
         if isinstance(self.target_modules, str) and self.layers_pattern is not None:
             raise ValueError("`layers_pattern` cannot be used when `target_modules` is a str.")
+
+
+class MegatronConfig:
```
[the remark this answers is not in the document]
zhangsheng377: > > Due to the particularity of megatron creating a distributed layer, the required megatron information needs to be injected before executing Lora
> 
> I still don't understand this fully. Ideally, we should find a solution that allows these config options to just live on `LoraConfig`.
> 
> What I can tell is that in `tp_layer.py`, the `LoraParallelLinear` inherits from `MegatronConfig.tensor_parallel`, which is probably where the issue is. It seems to me that there are multiple backends and you need the backend to be determined before the `LoraParallelLinear` class is created, is that right?
> 
> One solution that comes to mind is that we create one class for each possible backend. The code could look somewhat like this:
> 
> ```python
> class _LoraParallelLinearBase:
>     def __init__(...)  # same code as LoraParallelLinear, but use self.backend for backend-specific stuff
> 
> # assume we have the backends foo and bar:
> 
> class LoraParallelLinearFoo(_LoraParallelLinearBase, foo.ColumnParallelLinear, foo.RowParallelLinear, LoraParalleLayer):
>     def __init__(self, *args, **kwargs)
>         self.backend = foo
>         super().__init__(*args, **kwargs)
> 
> class LoraParallelLinearBar(_LoraParallelLinearBase, bar.ColumnParallelLinear, bar.RowParallelLinear, LoraParalleLayer):
>     def __init__(self, *args, **kwargs)
>         self.backend = bar
>         super().__init__(*args, **kwargs)
> ```
> 
> Probably a few more changes need to be made, but I hop this gets the idea across.
> 
> If this doesn't work for some reason, I'm sure we can find other ways, maybe by creating the class at runtime. As is, however, I think it's difficult for the user to make use of the feature.


Okay, I'll try to think of a way without explicit injection.

Let me first explain why we need to design it this way:
Because currently when people use the two warehouses Megatron-LM or Megatron-DeepSpeed, most of them will not install them into the environment. Instead, they will make some magic changes locally and then start training directly in the source code directory. (like us)

So I plan to change to using the native megatron class by default, so that explicit injection is not needed. But I plan to keep the set injection interface for people like us who have special needs.

BenjaminBossan:
Hi @zhangsheng377 I wanted to inform you that we merged #1106, which is a substantial refactor of PEFT and created some merge conflicts for your PR. The most notable change from that PR is that we refactored the adapter layers to take a `base_layer` argument now, which is a reference to the original module. This means that you have to slightly rewrite `LoraParalleLayer` and `LoraParallelLinear`. Take a look at `LoraLayer` and `Linear` LoRA layer, I hope that the change should be straightforward. If you have questions, please let me know and I'll help you with the change.

src/peft/tuners/lora/config.py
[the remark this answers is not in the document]
zhangsheng377: > > Due to the particularity of megatron creating a distributed layer, the required megatron information needs to be injected before executing Lora
> 
> I still don't understand this fully. Ideally, we should find a solution that allows these config options to just live on `LoraConfig`.
> 
> What I can tell is that in `tp_layer.py`, the `LoraParallelLinear` inherits from `MegatronConfig.tensor_parallel`, which is probably where the issue is. It seems to me that there are multiple backends and you need the backend to be determined before the `LoraParallelLinear` class is created, is that right?
> 
> One solution that comes to mind is that we create one class for each possible backend. The code could look somewhat like this:
> 
> ```python
> class _LoraParallelLinearBase:
>     def __init__(...)  # same code as LoraParallelLinear, but use self.backend for backend-specific stuff
> 
> # assume we have the backends foo and bar:
> 
> class LoraParallelLinearFoo(_LoraParallelLinearBase, foo.ColumnParallelLinear, foo.RowParallelLinear, LoraParalleLayer):
>     def __init__(self, *args, **kwargs)
>         self.backend = foo
>         super().__init__(*args, **kwargs)
> 
> class LoraParallelLinearBar(_LoraParallelLinearBase, bar.ColumnParallelLinear, bar.RowParallelLinear, LoraParalleLayer):
>     def __init__(self, *args, **kwargs)
>         self.backend = bar
>         super().__init__(*args, **kwargs)
> ```
> 
> Probably a few more changes need to be made, but I hop this gets the idea across.
> 
> If this doesn't work for some reason, I'm sure we can find other ways, maybe by creating the class at runtime. As is, however, I think it's difficult for the user to make use of the feature.

After thinking about it, I plan to put the config used when creating the tp distributed layer into LoraConfig. In this case, I might as well put the reference to tensor_parallel into LoraConfig~:laughing:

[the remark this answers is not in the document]
zhangsheng377: > > Due to the particularity of megatron creating a distributed layer, the required megatron information needs to be injected before executing Lora
> 
> I still don't understand this fully. Ideally, we should find a solution that allows these config options to just live on `LoraConfig`.
> 
> What I can tell is that in `tp_layer.py`, the `LoraParallelLinear` inherits from `MegatronConfig.tensor_parallel`, which is probably where the issue is. It seems to me that there are multiple backends and you need the backend to be determined before the `LoraParallelLinear` class is created, is that right?
> 
> One solution that comes to mind is that we create one class for each possible backend. The code could look somewhat like this:
> 
> ```python
> class _LoraParallelLinearBase:
>     def __init__(...)  # same code as LoraParallelLinear, but use self.backend for backend-specific stuff
> 
> # assume we have the backends foo and bar:
> 
> class LoraParallelLinearFoo(_LoraParallelLinearBase, foo.ColumnParallelLinear, foo.RowParallelLinear, LoraParalleLayer):
>     def __init__(self, *args, **kwargs)
>         self.backend = foo
>         super().__init__(*args, **kwargs)
> 
> class LoraParallelLinearBar(_LoraParallelLinearBase, bar.ColumnParallelLinear, bar.RowParallelLinear, LoraParalleLayer):
>     def __init__(self, *args, **kwargs)
>         self.backend = bar
>         super().__init__(*args, **kwargs)
> ```
> 
> Probably a few more changes need to be made, but I hop this gets the idea across.
> 
> If this doesn't work for some reason, I'm sure we can find other ways, maybe by creating the class at runtime. As is, however, I think it's difficult for the user to make use of the feature.

@BenjaminBossan I refactored the code according to the idea of ​​not explicitly injecting, please help me review it again. 
Now, we can use it like this:
```python
from megatron.arguments import core_transformer_config_from_args
from megatron import get_args
from megatron.core import tensor_parallel

config = core_transformer_config_from_args(get_args())

lora_config = LoraConfig(
    r=16,
    lora_alpha=32,
    target_modules=['query_key_value', 'dense', 'dense_h_to_4h', 'dense_4h_to_h'],
    lora_dropout=0.0,
    bias="none",
    megatron_config=config,
    tensor_parallel=tensor_parallel,
)
```


By the way, maybe I will resolve the conflicts and unit test code tomorrow.

tests/test_lora_megatron.py
```diff
@@ -0,0 +1,120 @@
...
+
+if is_megatron_available():
+    from megatron.core.transformer.module import Float16Module, MegatronModule
+    from megatron.core.transformer.transformer_config import TransformerConfig
+    from megatron.core.tensor_parallel.random import model_parallel_cuda_manual_seed
+    from megatron.core import tensor_parallel, parallel_state
+    from peft.tuners import lora
+
+
+    class Utils:
```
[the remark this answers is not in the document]
zhangsheng377: ok~

BenjaminBossan:
@zhangsheng377 Thanks a lot for your continued work on this PR. The usability looks much nicer now! There is still a merge conflict left. Don't hesitate to ask if you need help with resolving it.

tests/test_lora_megatron.py
```diff
@@ -0,0 +1,120 @@
...
+                target_modules=["linear"],
+            )
+            self.megatron_module = inject_adapter_in_model(lora_config, self.megatron_module)
+
+        def tearDown(self):
+            Utils.destroy_model_parallel()
+
+        def test_megatron_lora_module(self):
+            megatron_module = self.megatron_module
+            assert megatron_module
```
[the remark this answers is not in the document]
zhangsheng377: > What is this checking?

Well, it turned out that I just wanted to verify that the object was indeed created. Now change it to verify the type of this object.

tests/test_lora_megatron.py
```diff
@@ -0,0 +1,120 @@
...
+            self.lm_head = tensor_parallel.RowParallelLinear(input_size=10, output_size=10, config=config,
+                                                             init_method=init.xavier_normal_, bias=False)
+
+        def forward(self, input):
+            x = self.linear(input)[0]
+            x = self.lm_head(x)[0]
+            return x
+
+
+    class TestPeft(unittest.TestCase):
```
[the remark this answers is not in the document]
zhangsheng377: > AFAICT, this test just checks that the LoRA adapter exists on the model, is that right? Is there anything specific to `LoraParallelLinear` in the test?

Well, what I originally wanted to do was to verify that the tensor_parallel layer can indeed be lora-ized. But you are right, now there is verification of the lora layer type created by different tensor_parallel layers.

zhangsheng377:
> @zhangsheng377 Thanks a lot for your continued work on this PR. The usability looks much nicer now! There is still a merge conflict left in `lora/model.py`. Don't hesitate to ask if you need help with resolving it.

Ha, thank you for your concern. In fact, the conflict resolution was completed yesterday, but the changes in the main line were quite large, so I trained the llama model again to compare the loss and make sure there were no problems before submitting the code.

zhangsheng377:
@BenjaminBossan Sorry I forgot make style. Now that it has been reformed, please trigger the workflows again.

PS: Happy Thanksgiving.

src/peft/tuners/lora/config.py
```diff
@@ -117,6 +119,25 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    if is_megatron_available():
```
BenjaminBossan, line 122: Hmm, I don't think we should guard these options like this. There should be no harm of always adding them, even if megatron is not installed. Then, in the `__post_init__`, we can raise an error if `megatron_config is not None` or `tensor_parallel is not None` in case megatron is not installed.

If the config is conditional like this, we can run into trouble if someone saves the config when megatron is installed and later tries to load it in an env without megatron, because all of a sudden there are unknown fields.

src/peft/tuners/lora/config.py
```diff
@@ -130,3 +151,7 @@ def __post_init__(self):
         # if target_modules is a regex expression, then layers_pattern should be None
         if isinstance(self.target_modules, str) and self.layers_pattern is not None:
             raise ValueError("`layers_pattern` cannot be used when `target_modules` is a str.")
+
+        if is_megatron_available():
+            if self.tensor_parallel is not None or self.megatron_config is not None:
+                assert self.tensor_parallel is not None and self.megatron_config is not None
```
BenjaminBossan, line 157: I don't understand this check, it checks the same thing as the `if` condition? Also, let's not use `assert` but raise proper errors instead.

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,160 @@
...
+        if weight is not None:
+            # the layer is already completely initialized, this is an update
+            if weight.dtype.is_floating_point or weight.dtype.is_complex:
+                self.to(weight.device, dtype=weight.dtype)
+            else:
+                self.to(weight.device)
+        self.set_adapter(self.active_adapters)
+
+
+def get_lora_paralle_linear(
```
BenjaminBossan, line 72:
```suggestion
def get_lora_parallel_linear(
```

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,160 @@
...
+    base_layer, adapter_name: str, backend, r: int = 0, lora_alpha: int = 1, lora_dropout: float = 0.0, **kwargs
+):
+    class LoraParallelLinear(nn.Module, LoraParalleLayer):
```
BenjaminBossan, line 75: Why is this class defined inside the function, wouldn't it work if we define it on root level? We could pass the `backend` argument as an `__init__` parameter.

If this is changed, I think we can also remove `get_lora_paralle_linear` completely. Furthermore, I think we could remove the `LoraParalleLayer` class completely, which doesn't do much at the moment, only introducing the `update_layer` method. If we later want to add support for more layer types besides `Linear`, we can still factor this out into a mixin class for instance.

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,160 @@
...
+import torch
+import torch.nn as nn
+import torch.nn.init as init
+
+from .layer import LoraLayer
+
+
+class LoraParalleLayer(LoraLayer):
+    def __init__(self, base_layer, backend, is_paralle_a: bool = False):
+        LoraLayer.__init__(self, base_layer=base_layer)
```
BenjaminBossan, line 12: This should also work, right?

```suggestion
        super().__init__(base_layer=base_layer)
```

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,160 @@
...
+        if r <= 0:
+            raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
+        self.r[adapter_name] = r
+        self.lora_alpha[adapter_name] = lora_alpha
+        if lora_dropout > 0.0:
+            lora_dropout_layer = nn.Dropout(p=lora_dropout)
+        else:
+            lora_dropout_layer = nn.Identity()
+
+        self.lora_dropout.update(nn.ModuleDict({adapter_name: lora_dropout_layer}))
```
BenjaminBossan, line 26:
```suggestion
        self.lora_dropout[adapter_name] = lora_dropout_layer
```

zhangsheng377:
image

@BenjaminBossan Sorry, I neglected to install ruff this morning, so I mistakenly thought that style was no longer a problem.
But it should be fine now. But unlike the workflow, I did not find the problem of config.py locally. If it continue to report errors, please help me find out where the problem is.

src/peft/tuners/lora/config.py
```diff
@@ -117,6 +119,25 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    if is_megatron_available():
+        megatron_config: Optional[ModuleType] = field(
```
BenjaminBossan, line 123: Is `megatron_config` really a `ModuleType`, isn't it a `TransformerConfig` from megatron? AFAICT, this is a `dataclass`. When we save the `adapter_config.json`, the `dataclass` will be converted to a dict/json. So when we load it back in, it's still a dict. We need to parse the `TransformerConfig` in that case.

src/peft/tuners/lora/config.py
```diff
@@ -117,6 +119,25 @@ class LoraConfig(PeftConfig):
...
+            default=None,
+            metadata={
+                "help": (
+                    "The TransformerConfig of Megatron, it used to create lora's parallel linear. "
+                    "You can get it like this, `core_transformer_config_from_args(get_args())`"
```
BenjaminBossan, line 128: Let's add that the two functions are from megatron.

src/peft/tuners/lora/config.py
```diff
@@ -117,6 +119,25 @@ class LoraConfig(PeftConfig):
...
+                )
+            },
+        )
+        tensor_parallel: Optional[ModuleType] = field(
+            default=None,
+            metadata={
+                "help": (
+                    "The tensor_parallel module of Megatron, it used to judge and create lora's parallel linear. "
```
BenjaminBossan, line 136:
```suggestion
                    "The tensor_parallel module of Megatron, it is used to judge and create LoRA's parallel linear layer. "
```

BenjaminBossan (changes requested):
Thank you so much for updating the PR, it looks very good now. I'm not knowledgeable on megatron, so my review focuses more in the integration with PEFT.

As you can see, I left a couple of comments, please take a look. Mostly, I think we need to make some changes to the config to make saving and loading possible. Also, I think we can simplify the class structure. Finally, I have some questions concerning the `forward` method.

> I did not find the problem of config.py locally. If it continue to report errors, please help me find out where the problem is

The issue is that the `from types` import should come before `from typing`.

> PS: Happy Thanksgiving.

Thanks, but where I am, there is no Thanksgiving :) In case you celebrate it, happy Thanksgiving to you too.

src/peft/tuners/lora/config.py
zhangsheng377: "The TransformerConfig from Megatron" like this?

src/peft/tuners/lora/config.py
```diff
@@ -117,6 +119,25 @@ class LoraConfig(PeftConfig):
...
+        megatron_config: Optional[ModuleType] = field(
+            default=None,
+            metadata={
+                "help": (
+                    "The TransformerConfig of Megatron, it used to create lora's parallel linear. "
+                    "You can get it like this, `core_transformer_config_from_args(get_args())`"
+                )
+            },
+        )
+        tensor_parallel: Optional[ModuleType] = field(
```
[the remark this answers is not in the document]
zhangsheng377: @BenjaminBossan Do you mean that make ModuleType to str?

src/peft/tuners/lora/config.py
zhangsheng377: > I don't understand this check, it checks the same thing as the `if` condition? Also, let's not use `assert` but raise proper errors instead.

This check means is that either both of them are None, or neither of item is None.

src/peft/tuners/lora/config.py
zhangsheng377: > "The TransformerConfig from Megatron" like this?

"The TransformerConfig from Megatron" like this?

zhangsheng377: > Let's add that the two functions are from megatron.

"The TransformerConfig from Megatron" like this?

src/peft/tuners/lora/config.py
zhangsheng377: > Hmm, I don't think we should guard these options like this. There should be no harm of always adding them, even if megatron is not installed. Then, in the `__post_init__`, we can raise an error if `megatron_config is not None` or `tensor_parallel is not None` in case megatron is not installed.
> 
> If the config is conditional like this, we can run into trouble if someone saves the config when megatron is installed and later tries to load it in an env without megatron, because all of a sudden there are unknown fields.

What you said makes sense, but if the if judgment is removed and we change the ModuleType to TransformerConfig later, we must import megatron, but megatron may not be installed in the environment.

src/peft/tuners/lora/config.py
zhangsheng377:
```python
        if is_megatron_available():
            if (self.tensor_parallel is not None or self.megatron_config is not None) and (
                self.tensor_parallel is None or self.megatron_config is None
            ):
                raise ValueError("`tensor_parallel` and `megatron_config` must both have value or neither have value.")
```
Now, it has been modified like this.

src/peft/tuners/lora/config.py
BenjaminBossan: Sorry, I meant to refer to `core_transformer_config_from_args` and `get_args`

src/peft/tuners/lora/config.py
```diff
@@ -153,5 +153,7 @@ def __post_init__(self):
             raise ValueError("`layers_pattern` cannot be used when `target_modules` is a str.")
 
         if is_megatron_available():
-            if self.tensor_parallel is not None or self.megatron_config is not None:
-                assert self.tensor_parallel is not None and self.megatron_config is not None
+            if (self.tensor_parallel is not None or self.megatron_config is not None) and (
+                self.tensor_parallel is None or self.megatron_config is None
+            ):
```
BenjaminBossan, line 158: I see the intent now. Couldn't this be simplified to:

`if (self.tensor_parallel is None) is not (self.megatron_config is None):`

src/peft/tuners/lora/config.py
BenjaminBossan: Yes, we cannot change it to `TransformerConfig`, that's true. But at the end of the day, that's just a `dataclass`, rigth? Maybe we can just put a dict here, which will be passed like this: `TransformerConfig(**peft_config.megatron_transformer_config_kwargs)`. Alternatively, we could have a proxy `dataclass` in PEFT with the same fields as `TransformerConfig`.

src/peft/tuners/lora/config.py
zhangsheng377: > Sorry, I meant to refer to `core_transformer_config_from_args` and `get_args`

Oh, ok. Now is like this:
```python
            metadata={
                "help": (
                    "The TransformerConfig from Megatron, it is used to create LoRA's parallel linear. "
                    "You can get it like this, `core_transformer_config_from_args(get_args())`, this two functions are from megatron. "
                )
            },
```

src/peft/tuners/lora/config.py
zhangsheng377: > Yes, we cannot change it to `TransformerConfig`, that's true. But at the end of the day, that's just a `dataclass`, rigth? Maybe we can just put a dict here, which will be passed like this: `TransformerConfig(**peft_config.megatron_transformer_config_kwargs)`. Alternatively, we could have a proxy `dataclass` in PEFT with the same fields as `TransformerConfig`.

Yes, now I have added a sentence in tp_layer.py. If it is a dict, I will use it to initialize TransformerConfig.

But what type should be used for tensor_parallel? I don't quite understand what you mean by "put a string here".

tests/test_lora_megatron.py
```diff
@@ -0,0 +1,128 @@
...
+                    self.assertTrue(isinstance(module, torch.nn.Linear))
+
+            x = torch.ones((2, 10)).cuda()
+            self.assertTrue(megatron_module(x).shape == (2, 10))
+
+        def test_get_peft_model_state_dict(self):
+            peft_state_dict = get_peft_model_state_dict(self.megatron_module)
+
+            for key in peft_state_dict.keys():
+                self.assertTrue("lora" in key)
```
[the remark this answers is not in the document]
zhangsheng377: > We should add a test that we can save and load the model with parallel linear layers using `save_pretrained` and `from_pretrained`.

Yes, save_pretrained and from_pretrained should indeed be added to the unit test, although because each process will independently store its own part of the model after megatron splits the model, we actually do not use these two functions now. But you inspired me and I will try to use these two functions in our code later. Thanks.

src/peft/tuners/lora/config.py
BenjaminBossan: So what I meant is that we could allow users to set `tensor_parallel = "megatron.core.tensor_parallel"`, and then we can import that using `module = __import__(config.tensor_parallel)` (or using importlib). This way, it's a simple string which we can serialize, and at the same time it defines exactly what module we intend on using. WDYT?

src/peft/tuners/lora/config.py
zhangsheng377: yes

src/peft/tuners/lora/config.py
```diff
@@ -117,6 +119,25 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    if is_megatron_available():
+        megatron_config: Optional[ModuleType] = field(
+            default=None,
+            metadata={
+                "help": (
+                    "The TransformerConfig of Megatron, it used to create lora's parallel linear. "
```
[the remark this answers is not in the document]
zhangsheng377: > Could you please also add a sentence about when and why users should use this option?

ok~

src/peft/tuners/lora/config.py
zhangsheng377: > So what I meant is that we could allow users to set `tensor_parallel = "megatron.core.tensor_parallel"`, and then we can import that using `module = __import__(config.tensor_parallel)` (or using importlib). This way, it's a simple string which we can serialize, and at the same time it defines exactly what module we intend on using. WDYT?

Wow, this is indeed a good idea.

Because when I created TransformerConfig, I found that we even customized this and added some attributes. So I now replaced tensor_parallel in LoraConfig with megatron_core so that I can get the tensor_parallel module and transformer_config module at the same time.

@BenjaminBossan Can you help me take a look at this modification?

src/peft/tuners/lora/tp_layer.py
zhangsheng377: > Why is this class defined inside the function, wouldn't it work if we define it on root level? We could pass the `backend` argument as an `__init__` parameter.
> 
> If this is changed, I think we can also remove `get_lora_paralle_linear` completely. Furthermore, I think we could remove the `LoraParalleLayer` class completely, which doesn't do much at the moment, only introducing the `update_layer` method. If we later want to add support for more layer types besides `Linear`, we can still factor this out into a mixin class for instance.

Yes, I don't need `get_lora_paralle_linear` and `LoraParalleLayer` this base class~

src/peft/tuners/lora/tp_layer.py
zhangsheng377: I will remove `get_lora_paralle_linear` func.

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,160 @@
...
+                        lora_result = lora_result[0]
+                    lora_result = lora_B(lora_result)
+                    if isinstance(lora_result, tuple):
+                        lora_result = lora_result[0]
+                    lora_result = lora_result * scaling
+
+                    result = result + lora_result
+
+            result = result.to(previous_dtype)
+            return result, bias
```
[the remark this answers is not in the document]
zhangsheng377: > I'm a bit surprised that we can return `bias` here. Wouldn't that break the caller, who assumes that `Linear` just returns a single tensor?

Haha, megatron's ColumnParallelLinear and RowParallelLinear do return result and bias.
General models, such as llama, do not use bias, it is just None. But I remember it seems to be a bloom model, so this bias is used.

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,160 @@
...
+
+        def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any):
+            previous_dtype = x.dtype
+            # If weight is used for matrix multiplication here, the final aggregation operation of the original
+            # parallel_linear layer will be missing, so we need to directly call its forward function to obtain the
+            # output of the original parallel_linear layer.
+            if self.disable_adapters:
+                if self.merged:
+                    self.unmerge()
+                result, bias = self.base_layer(x, *args, **kwargs)
```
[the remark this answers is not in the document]
zhangsheng377: > Where does `bias` come from?

Megatron's ColumnParallelLinear and RowParallelLinear do return result and bias~

src/peft/tuners/lora/tp_layer.py
zhangsheng377: > This should also work, right?

Yes, but now `LoraParalleLayer` will be removed.

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,160 @@
...
+            lora_dropout_layer = nn.Identity()
+
+        self.lora_dropout.update(nn.ModuleDict({adapter_name: lora_dropout_layer}))
+
+        config = parallel_linear_kwargs["config"]
+        # lora needs to be forced to upgrade to 32-bit precision, otherwise it will overflow
+        config.params_dtype = torch.float32
+        init_method = parallel_linear_kwargs.get("init_method", init.xavier_normal_)
+        input_is_parallel = parallel_linear_kwargs.get("input_is_parallel", True)
+        gather_output = parallel_linear_kwargs.get("gather_output", False)
```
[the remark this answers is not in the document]
zhangsheng377: > update_layer

`init_method` is OK, but `input_is_parallel` and `gather_output` correspond to `RowParallelLinear` and `ColumnParallelLinear` respectively, and they are not available at the same time. So does this situation also need to become a parameter of the `update_layer` function? Default is None?

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,160 @@
...
+        if r > 0:
```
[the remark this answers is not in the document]
zhangsheng377: > This is not needed because we already check `r <= 0` at the beginning, right? Yes, we should also remove the check in the default LoRA layer :)

Ha, ok~~

src/peft/tuners/lora/tp_layer.py
zhangsheng377: > 

ok~

src/peft/tuners/lora/config.py
zhangsheng377: > 我不明白这个检查，它检查与`if`条件相同的东西？另外，我们不要使用`assert`，而是提出适当的错误。



> I don't understand this check, it checks the same thing as the `if` condition? Also, let's not use `assert` but raise proper errors instead.

Megatron.core is now used by default, based on megatron_config, so this check is no longer needed.

tests/test_lora_megatron.py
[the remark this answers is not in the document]
zhangsheng377: > We should add a test that we can save and load the model with parallel linear layers using `save_pretrained` and `from_pretrained`.

I found that some values ​​in TransformerConfig are function types and cannot be serialized.
But fortunately, when we use megatron, we will also use the megatron framework to save models and configurations, so save_pretrained and from_pretrained are not needed yet.

src/peft/tuners/lora/model.py
```diff
@@ -258,6 +258,10 @@ def _create_new_module(lora_config, adapter_name, target, **kwargs):
         else:
             target_base_layer = target
 
+        megatron_core = None
+        if lora_config.megatron_config:
+            megatron_core = __import__(lora_config.megatron_core, fromlist=["tensor_parallel", "transformer"])
```
BenjaminBossan, line 263: So I checked the [Python docs](https://docs.python.org/3/library/functions.html#import__) and they mention that:

> This is an advanced function that is not needed in everyday Python programming, unlike [importlib.import_module()](https://docs.python.org/3/library/importlib.html#importlib.import_module).

It's not quite clear to me what makes this "advanced", but it would probably be better to switch to `importlib.import_module`.

src/peft/tuners/lora/config.py
```diff
@@ -117,6 +117,26 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    megatron_config: Optional[dict] = field(
+        default_factory=dict,
+        metadata={
+            "help": (
+                "The TransformerConfig from Megatron, it is used to create LoRA's parallel linear layer. "
+                "You can get it like this, `core_transformer_config_from_args(get_args())`, this two functions are from megatron. "
+                "You need to specify this parameter when you want to loraize the ColumnParallelLinear and RowParallelLinear layers of megatron. "
```
BenjaminBossan, line 126: Line is too long, trailing whitespace can be removed.

src/peft/tuners/lora/config.py
```diff
@@ -117,6 +117,26 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    megatron_config: Optional[dict] = field(
+        default_factory=dict,
```
BenjaminBossan, line 121: Since this is supposed to be an optional argument, how about making the default `None`?

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,145 @@
...
+        super().__init__()
+        LoraLayer.__init__(self, base_layer=base_layer)
+
+        self.backend = backend
+        self.is_paralle_a = isinstance(base_layer, backend.RowParallelLinear)
+        self.fan_in_fan_out = fan_in_fan_out
+        self._active_adapter = adapter_name
+
+        megatron_config = kwargs["megatron_config"]
+        parallel_linear_kwargs = {"megatron_config": megatron_config}
```
BenjaminBossan, line 39: If `megatron_config` is `None`, would this still work? Should we put a default `TransformerConfig` here instead?

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,145 @@
...
+        if hasattr(megatron_config, "init_method"):
+            parallel_linear_kwargs["init_method"] = megatron_config.init_method
+        if isinstance(base_layer, self.backend.RowParallelLinear):
+            parallel_linear_kwargs["input_is_parallel"] = base_layer.input_is_parallel
+        else:
+            parallel_linear_kwargs["gather_output"] = base_layer.gather_output
```
BenjaminBossan, line 45: For these 3 arguments and for `megatron_config`, is there really a point in putting them inside of `parallel_linear_kwargs`? They are unpacked anyway when passed to `self.update_layer`, so it just seems like an unnecessary indirection. I think it would be easier to just pass all of them explicitly and not have any `**kwargs`.

tests/test_lora_megatron.py
```diff
@@ -0,0 +1,136 @@
...
+            self.lm_head = tensor_parallel.RowParallelLinear(
+                input_size=10, output_size=10, config=config, init_method=init.xavier_normal_, bias=False
+            )
+
+        def forward(self, input):
+            x = self.linear(input)[0]
+            x = self.lm_head(x)[0]
+            return x
+
+    class TestPeft(unittest.TestCase):
```
BenjaminBossan, line 84:
```suggestion
    class TestMegatron(unittest.TestCase):
```

or `TestMegatronLora` would also work. `TestPeft` doesn't make too much sense in PEFT ;)

tests/test_lora_megatron.py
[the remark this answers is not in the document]
BenjaminBossan: I see, thanks. Could you maybe add a comment about this to the documentation inside the config?

tests/test_lora_megatron.py
```diff
@@ -0,0 +1,136 @@
...
+                if name.endswith("linear.lora_B.default"):
+                    self.assertTrue(isinstance(module, tensor_parallel.ColumnParallelLinear))
+
+                if name.endswith("lm_head.lora_A.default"):
+                    self.assertTrue(isinstance(module, tensor_parallel.RowParallelLinear))
+                if name.endswith("lm_head.lora_B.default"):
+                    self.assertTrue(isinstance(module, torch.nn.Linear))
+
+            x = torch.ones((2, 10)).cuda()
+            self.assertTrue(megatron_module(x).shape == (2, 10))
```
BenjaminBossan, line 130: Could you please also check that the results are finite? Would it be possible to also check the backwards pass working?

BenjaminBossan (changes requested):
Thanks so much, I think the implementation is much cleaner now and easier to understand. From my point of view, there are only a few details left, otherwise we should be good to merge.

tests/test_lora_megatron.py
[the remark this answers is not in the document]
zhangsheng377: > I see, thanks. Could you maybe add a comment about this to the documentation inside the config?

ok, now is like this:
```python
metadata={
            "help": (
                "The TransformerConfig from Megatron, it is used to create LoRA's parallel linear layer. "
                "You can get it like this, `core_transformer_config_from_args(get_args())`, this two functions are from megatron. "
                "You need to specify this parameter when you want to loraize the ColumnParallelLinear and RowParallelLinear layers of megatron. "
                "It should be noted that TransformerConfig may not necessarily be serialized, which means that we may not be able to use the save_pretrained and from_pretrained functions to save and load LoRA's parameters and configurations. "
                "But luckily, when we use megatron, we can use megatron's framework to save and load models and configurations. "
            )
        },
```

src/peft/tuners/lora/model.py
zhangsheng377: > So I checked the [Python docs](https://docs.python.org/3/library/functions.html#import__) and they mention that:
> 
> > This is an advanced function that is not needed in everyday Python programming, unlike [importlib.import_module()](https://docs.python.org/3/library/importlib.html#importlib.import_module).
> 
> It's not quite clear to me what makes this "advanced", but it would probably be better to switch to `importlib.import_module`.

ok~

src/peft/tuners/lora/config.py
zhangsheng377: > Line is too long, trailing whitespace can be removed.

ok~

src/peft/tuners/lora/tp_layer.py
zhangsheng377: > If `megatron_config` is `None`, would this still work? Should we put a default `TransformerConfig` here instead?

ok~

src/peft/tuners/lora/config.py
zhangsheng377: > Since this is supposed to be an optional argument, how about making the default `None`?

ok.

src/peft/tuners/lora/tp_layer.py
zhangsheng377: > If `megatron_config` is `None`, would this still work? Should we put a default `TransformerConfig` here instead?

Because in model.py, we will judge whether megatron_config has value, and if so, we will import megatron_core. Only when megatron_core is successfully imported will it go to tp_layer.py.
Therefore, the megatron_config here can work. And because we don’t know the configured environment information, we can’t give a default TransformerConfig.

src/peft/tuners/lora/tp_layer.py
zhangsheng377: > For these 3 arguments and for `megatron_config`, is there really a point in putting them inside of `parallel_linear_kwargs`? They are unpacked anyway when passed to `self.update_layer`, so it just seems like an unnecessary indirection. I think it would be easier to just pass all of them explicitly and not have any `**kwargs`.

Yes, it is indeed clearer now.

tests/test_lora_megatron.py
zhangsheng377: > `TestMegatronLora`

Ha~ you are right.

tests/test_lora_megatron.py
zhangsheng377: > Could you please also check that the results are finite? Would it be possible to also check the backwards pass working?

OK, now I directly compare it with the forward results of the original model, and verify the backward of the lora model.

src/peft/tuners/lora/tp_layer.py
BenjaminBossan: I see, thanks

src/peft/tuners/lora/tp_layer.py
```diff
@@ -0,0 +1,158 @@
...
+        super().__init__()
+        LoraLayer.__init__(self, base_layer=base_layer)
+
+        self.backend = backend
+        self.is_paralle_a = isinstance(base_layer, backend.RowParallelLinear)
+        self.fan_in_fan_out = fan_in_fan_out
+        self._active_adapter = adapter_name
+
+        megatron_config = kwargs["megatron_config"]
+        parallel_linear_kwargs = {"megatron_config": megatron_config}
```
BenjaminBossan, line 39: Can we get rid of `parallel_linear_kwargs` completely, and just pass `megatron_config` directly to `update_layer`?

BenjaminBossan (approved):
Thanks a lot for this PR. It looks pretty good now, I have a nit but it's no big deal.

Before merging, I just want to discuss the newly added tests. As is, they would not be run by the Github CI at any point, so in theory bugs could be introduced without us noticing. I have no personal experience, but the requirements for running megatron seem to be quite high, so I'm not sure if we could make it run on our GPU runners. I wonder if there is a way we can still make it work.

src/peft/tuners/lora/tp_layer.py
zhangsheng377: > Can we get rid of `parallel_linear_kwargs` completely, and just pass `megatron_config` directly to `update_layer`?

I have tried this, but because `megatron_config` cannot provide a default value, and if it is a function parameter without a default value, my pycharm will report that the `update_layer` function does not match the base method.

zhangsheng377:
> Thanks a lot for this PR. It looks pretty good now, I have a nit but it's no big deal.
> 
> Before merging, I just want to discuss the newly added tests. As is, they would not be run by the Github CI at any point, so in theory bugs could be introduced without us noticing. I have no personal experience, but the requirements for running megatron seem to be quite high, so I'm not sure if we could make it run on our GPU runners. I wonder if there is a way we can still make it work.

Well, I only used one gpu card to run the newly added unit test locally, and the resource requirements are not large.
I installed Megatron from the source code, but it does have a lot of dependencies and the installation is a bit troublesome, but it should be okay. It should take less than an afternoon for me to configure the environment.

Or you should be able to install Megatron-DeepSpeed, after all, it is actually the backend of our current magic modification.

https://github.com/huggingface/peft/pull/1092#issuecomment-1805044931

BenjaminBossan:
Unfortunately, I did not manage to successfully build APEX and it seems I'm not the only one, judging from all the open issues. Therefore, I couldn't test it locally. If you have a recipe to get this all to run, which we could use for a CI job, that would be great.

src/peft/tuners/lora/tp_layer.py
BenjaminBossan: Oh I see. Too bad, as this isn't really an error, but I guess we can keep it as is.

BenjaminBossan:
Heads up, there is a small merge conflict, should be easy to fix.

zhangsheng377:
> Heads up, there is a small merge conflict, should be easy to fix.

Yes, it's done.

zhangsheng377:
> Unfortunately, I did not manage to successfully build APEX and it seems I'm not the only one, judging from all the open issues. Therefore, I couldn't test it locally. If you have a recipe to get this all to run, which we could use for a CI job, that would be great.

My apex is version 0.1, and it seems that it should be installed directly via pip.
Can you post the error report and take a look?

zhangsheng377:
> > My apex is version 0.1, and it seems that it should be installed directly via pip.
> 
> Could you tell me how to do that? The `apex` package on PyPI seems to be unrelated.
> 
> What I did is follow the instructions [here](https://github.com/NVIDIA/apex#from-source). Several users reported issues with this. Some suggested solutions included checking out specific tags or commits, but none of those I tried worked for me.

I think I did not install it from the source code. You can try `pip install apex` directly.

BenjaminBossan:
> I think I did not install it from the source code. You can try `pip install apex` directly.

I think this will install the wrong package. When checking on [PyPI](https://pypi.org/project/apex/), the description says:

> Authentication, Form Library, I18N/L10N, Flash Message Template (not associated with Pyramid, a Pylons project)

zhangsheng377:
> Authentication, Form Library, I18N/L10N, Flash Message Template (not associated with Pyramid, a Pylons project)

(xx) [root@localhost peft]# pip install apex
Looking in indexes: http://mirrors.aliyun.com/pypi/simple/
Requirement already satisfied: apex in /root/miniconda3/envs/xx/lib/python3.9/site-packages/apex-0.1-py3.9.egg (0.1)
Requirement already satisfied: packaging>20.6 in /root/miniconda3/envs/xx/lib/python3.9/site-packages (from apex) (23.2)

Maybe you can change the index-url? Or specify version?

BenjaminBossan:
> Looking in indexes: http://mirrors.aliyun.com/pypi/simple/

Interesting. The host is unfortunately insecure (http) and I don't know who this is. Therefore, we cannot use this index for the CI, as it could start hosting malicious code at any point in the future.

Anyway, even if we cannot find a way right now to make the tests work, we can still proceed. Hopefully, we can find a better way in the future, maybe Nvidia manages to provide a package that can be reliably installed soon.

BenjaminBossan:
@zhangsheng377 Thanks so much for this wonderful PR.

thincal:
> on our own magetron-deepspeed framework

@zhangsheng377 so that can we use the latest https://github.com/microsoft/Megatron-DeepSpeed with this peft now ?

zhangsheng377:
> > on our own magetron-deepspeed framework
> 
> @zhangsheng377 so that can we use the latest https://github.com/microsoft/Megatron-DeepSpeed with this peft now ?

Yes, you can.

thincal:
Thanks for your information. And another question, with Megatron-DeepSpeed used, what's the format for input model and result model in this LoRA fine-tuning with PEFT ? @zhangsheng377

zhangsheng377:
> Thanks for your information. And another question, with Megatron-DeepSpeed used, what's the format for input model and result model in this LoRA fine-tuning with PEFT ? @zhangsheng377

@thincal In Megatron-DeepSpeed, the base model is megatron's model, so you can see my ut in this pr.

kota-iizuka:
@zhangsheng377 Thank you for your great work. I'd like to try your code, so could you show me the `pretrain_llama.py` that you mentioned in the comment https://github.com/huggingface/peft/pull/1092#issuecomment-1801646874 ? I think it's a script that allows you to specify `--lora-target-modules` from the command line, but I couldn't find anything after a quick search.

zhangsheng377:
> @zhangsheng377 Thank you for your great work. I'd like to try your code, so could you show me the `pretrain_llama.py` that you mentioned in the comment [#1092 (comment)](https://github.com/huggingface/peft/pull/1092#issuecomment-1801646874) ? I think it's a script that allows us to specify `--lora-target-modules` from the command line, but I couldn't find anything after a quick search.

Haha, You will need to adapt the Megatron code to PEFT. For example, modify the `model_provider` function in the Megatron root directory `pretrain_gpy.py` file.

By the way, '--lora-target-modules' is a parameter I added to Megatron myself, and you can use your own adaptation process.

kota-iizuka:
@zhangsheng377 Thank you. I added the conversion to the LoRA model at the end of `model_provider()` in the example https://github.com/microsoft/Megatron-DeepSpeed/tree/main/examples_deepspeed/finetune_hf_llama, but I got an error when loading the model. This sample includes a process to convert the huggingface model to the megatron model, and I think that the LoRA model conversion is also required there.

zhangsheng377:
> Thank you. I added the conversion to the LoRA model at the end of `model_provider()` in the example https://github.com/microsoft/Megatron-DeepSpeed/tree/main/examples_deepspeed/finetune_hf_llama, but I got an error when loading the model. This sample includes a process to convert the huggingface model to the megatron model, and I think that the LoRA model conversion is also required there.

Things to note when loading the model: 1. The original hf model may not have lora parameters. 2. If you are converting the lora plug-in, you need to write the conversion code yourself.

kota-iizuka:
> If you are converting the lora plug-in, you need to write the conversion code yourself.

I understand that there is no standard way to load weight in LoRA training. I will try converting it now. Thank you.

PR #1099, opened 2023-11-09
Loftq

yxli2123:
Hi,

We would like to add our method, called LoftQ, to peft repo. Here's the paper link of our method https://arxiv.org/abs/2310.08659.


BenjaminBossan:
Thanks so much for this PR, LoftQ looks very interesting. I didn't do a full review yet, but have some more general points to start with:

1. Do you have any tests or examples that show this in action?
2. Could you please run `make style` on the the code?
3. From my superficial understanding, LoftQ is mainly an initialization scheme for LoRA when using quantized models, which works with different quantization methods. Would it be possible to integrate this completely into our LoRA implementation instead of creating a new adapter? E.g. right now, users can configure `init_lora_weights=True/False`. It would be super convenient if they could pass `init_lora_weights="loftq"` and then they get a normal LoRA model but with LoftQ-initialized weights. Maybe I'm missing something and that would not be possible, please let me know. If this is conceptually possible but you need support in implementing it in PEFT, let us know and we can help.

yxli2123:
Hi,

Thanks for your advice. We will integrate LoftQ into init_lora_weights.

Best,
Yixiao

PR #1094, opened 2023-11-08
fix the failing CI tests

pacman100:
### What does this PR do?
1. Fixes the failing CI tests: https://github.com/huggingface/peft/actions/runs/6792850417/job/18466760365 and https://github.com/huggingface/peft/actions/runs/6792850417/job/18466760175

PR #1100, opened 2023-11-09
FIX: Failing nightly CI tests due to IA3 config

BenjaminBossan:
Same idea as in PR as #1094, but for yet more ill-configured IA³ configs. The tests are now failing because we do stricter checks on incorrect IA³ configs.

These are the failing tests:

https://github.com/huggingface/peft/actions/runs/6806502406/job/18507828825
https://github.com/huggingface/peft/actions/runs/6806502406/job/18507829021

With this fix, they now pass locally.


younesbelkada (approved):
Thanks for fixing!

Issue #1079, opened 2023-11-04
Token classification LoRA example broken


srama2512:
### System Info

Python version: 3.9.18
PEFT version: 0.5.0
Transformers: 4.35.0
Safetensors: 0.4.0

System: Ubuntu 22.04, CUDA 11.8

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

I tried running the token-classification-lora example from https://huggingface.co/docs/peft/task_guides/token-classification-lora. I run into this error:

```
  File "/home/ubuntu/lora_token_classification.py", line 157, in <module>
    trainer.train()
  File "/home/ubuntu/mambaforge/envs/llm/lib/python3.9/site-packages/transformers/trainer.py", line 1555, in train
    return inner_training_loop(
  File "/home/ubuntu/mambaforge/envs/llm/lib/python3.9/site-packages/transformers/trainer.py", line 1937, in _inner_training_loop
    self._maybe_log_save_evaluate(tr_loss, model, trial, epoch, ignore_keys_for_eval)
  File "/home/ubuntu/mambaforge/envs/llm/lib/python3.9/site-packages/transformers/trainer.py", line 2282, in _maybe_log_save_evaluate
    self._save_checkpoint(model, trial, metrics=metrics)
  File "/home/ubuntu/mambaforge/envs/llm/lib/python3.9/site-packages/transformers/trainer.py", line 2350, in _save_checkpoint
    self.save_model(output_dir, _internal_call=True)
  File "/home/ubuntu/mambaforge/envs/llm/lib/python3.9/site-packages/transformers/trainer.py", line 2843, in save_model
    self._save(output_dir)
  File "/home/ubuntu/mambaforge/envs/llm/lib/python3.9/site-packages/transformers/trainer.py", line 2901, in _save
    self.model.save_pretrained(
  File "/home/ubuntu/mambaforge/envs/llm/lib/python3.9/site-packages/peft/peft_model.py", line 174, in save_pretrained
    safe_save_file(
  File "/home/ubuntu/mambaforge/envs/llm/lib/python3.9/site-packages/safetensors/torch.py", line 281, in save_file
    serialize_file(_flatten(tensors), filename, metadata=metadata)
  File "/home/ubuntu/mambaforge/envs/llm/lib/python3.9/site-packages/safetensors/torch.py", line 467, in _flatten
[...]

PR #1101, opened 2023-11-09
[`core`] Fix safetensors serialization for shared tensors

younesbelkada:
# What does this PR do?

Fixes https://github.com/huggingface/peft/issues/1079

Safetensors do not support tensor aliasing, before saving the model state dict, we need to loop over the state dict and check for aliased tensors and `clone()` them to avoid any error

cc @BenjaminBossan @pacman100

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -203,6 +203,7 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
             os.makedirs(output_dir, exist_ok=True)
 
             if safe_serialization:
+                # Section copied from: https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L2111-L2134
                 # Safetensors does not allow tensor aliasing.
                 # We're going to remove aliases before saving
                 ptrs = collections.defaultdict(list)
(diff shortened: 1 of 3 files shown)


src/peft/peft_model.py
```diff
@@ -199,6 +203,25 @@ def save_pretrained(
             os.makedirs(output_dir, exist_ok=True)
 
             if safe_serialization:
+                # Safetensors does not allow tensor aliasing.
```
BenjaminBossan, line 206: This block is basically copied from https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L2111-L2134, right? Let's add a comment.

src/peft/peft_model.py
```diff
@@ -199,6 +203,25 @@ def save_pretrained(
...
+                # We're going to remove aliases before saving
+                ptrs = collections.defaultdict(list)
+                for name, tensor in output_state_dict.items():
+                    # Sometimes in the state_dict we have non-tensor objects.
+                    # e.g. in bitsandbytes we have some `str` objects in the state_dict
+                    if isinstance(tensor, torch.Tensor):
+                        ptrs[id_tensor_storage(tensor)].append(name)
+                    else:
+                        # In the non-tensor case, fall back to the pointer of the object itself
+                        ptrs[id(tensor)].append(name)
```
younesbelkada, line 216: Not sure if this is needed, I added that in the transformers code to handle bnb state dict case and should be safe to keep it but happy to remove it as well

BenjaminBossan: Probably it's better to follow the same steps as transformers.

BenjaminBossan (approved):
Thanks for investigating this. Just a minor comment about adding a comment, apart from this, LGTM.

src/peft/peft_model.py
younesbelkada: Ok perfect!

PR #1102, opened 2023-11-09
Change to 0.6.1.dev0

younesbelkada:
As per title, as we are going to do a minor release (edit: patch release)

cc @pacman100 @BenjaminBossan


BenjaminBossan (approved):
Looks good, thanks. In semver parlence, it's a patch release.

younesbelkada:
Indeed yes!

PR #980, opened 2023-09-29
Combine multiple (IA)^3 Adapters and delete (IA)^3 adapters

alexrs:
### Problem
$(IA)^3$ models supports multiple adapters, however, there are a few missing features (compared to LoRA):
- Delete adapters: There is no method to delete an adapter from an `IA3Model`.
- Add Weighted Adapter: There is no method to add a weighted combination of multiple adapters.

### Solution
- Add a `delete_adapter` method based on the one in `LoraModel`.
- Add a `add_weighted_adapter` method based on the one in `LoraModel`. This method, however, is a simplified version of the LoRA one. As $(IA)^3$ injects trainable vectors, I have deleted both the `cat` and `svd` options.

### Other minor modifications
- I have deleted the `self.scaling` in `IA3Layer` as it was not used. I believe this was a leftover from the LoRA implementation this layer seems to be based on.

### Discussion
I have tested this locally using a simple script (see below) but I have not run any automated tests yet. What is the best way to test these changes?

<details>
  <summary>Test script</summary>

```py
import os

import datasets
import torch
from peft import PeftModel, PeftConfig
from transformers import (
    AutoModelForCausalLM,
    AutoTokenizer,
    BitsAndBytesConfig,
)

def run_model(
    output_dir: str,
    cache_dir: str,
    dataset_dir: str,
    peft_strategy: str,
):
    # Load dataset from the hub
    datasets.config.DOWNLOADED_DATASETS_PATH = dataset_dir

    tokenizer = AutoTokenizer.from_pretrained(
        "meta-llama/Llama-2-7b-hf", cache_dir=cache_dir
    )

    quantization_config = BitsAndBytesConfig(load_in_8bit=True)

    config = PeftConfig.from_pretrained(os.path.join(output_dir, peft_strategy, "general"))

    model = AutoModelForCausalLM.from_pretrained(
        config.base_model_name_or_path,
        torch_dtype=torch.float,
        quantization_config=quantization_config,
        # device_map="auto",
        cache_dir=cache_dir,
    )

    model = PeftModel.from_pretrained(model, os.path.join(output_dir, peft_strategy, "general"))
    model.load_adapter(os.path.join(output_dir, peft_strategy, "code"), "code")
    model.load_adapter(os.path.join(output_dir, peft_strategy, "creative"), "creative")

    model.add_weighted_adapter(["code", "creative"], [0.5, 0.5], "code_creative")
    model.set_adapter("code_creative")

    prompt = f"""
    ### Input:
[...]

PR #1105, opened 2023-11-09
Refactor adapter deletion

BenjaminBossan:
## Description

The job of deleting an adapter is now transferred to the adapter layer,
instead of the adapter model. This makes it easier for users or other
libraries who don't use the adapter model to delete adapters.

## Implementation

The code should now be more generic, relying less on hard-coded
attributes.

As a precaution, I also changed the type of `adapter_layer_names` from
list to tuple, as it should not be mutated.

When deleting the active adapter, the logic for choosing the new active
adapter has been changed slightly to ensure consistency across layers.
In practice, this should rarely make a difference. ~~An error is now
raised if the last remaining adapter is deleted.~~

## Test coverage has been increased:

- Deleting adapters is now also tested for custom models.
- It is also tested for LoHa, LoKr, not only LoRA.
- I added a test for deleting the non-active adapter.

## Not implemented

I did not add adapter deletion to IA³, since it is included in #980. LMK
if it should be added here instead.

## Other changes

Fixed a bug in updating IA³ layers. For `Conv2d`, the method is called
`update_layer`, `update_layer_conv2d` does not exist.

### DIFF
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -661,29 +661,16 @@ class LoraModel(BaseTuner):
         del self.peft_config[adapter_name]
 
         key_list = [key for key, _ in self.model.named_modules() if "lora" not in key]
+        new_adapter = None
         for key in key_list:
             _, target, _ = _get_submodules(self.model, key)
             if isinstance(target, LoraLayer):
-                for attr in [
-                    "r",
-                    "lora_alpha",
-                    "scaling",
-                    "lora_A",
-                    "lora_B",
-                    "lora_embedding_A",
-                    "lora_embedding_B",
-                    "lora_dropout",
-                ]:
-                    if adapter_name in getattr(target, attr):
-                        getattr(target, attr).pop(adapter_name)
-                if adapter_name in target.active_adapters:
-                    resetting_active_adapter = (
-                        list(self.peft_config.keys())[0] if len(self.peft_config) > 0 else "default"
-                    )
-                    warnings.warn(
-                        f"Adapter {adapter_name} was active which is now deleted. Setting active adapter to {resetting_active_adapter}. "
-                    )
-                    target.set_adapter(resetting_active_adapter)
+                target.delete_adapter(adapter_name)
+                if new_adapter is None:
+                    new_adapter = target.active_adapters[:]
+
+        if new_adapter:
+            self.active_adapter = new_adapter
 
     def merge_and_unload(self, progressbar: bool = False, safe_merge: bool = False):
         r"""
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -669,8 +669,7 @@ class LoraModel(BaseTuner):
                 if new_adapter is None:
                     new_adapter = target.active_adapters[:]
 
-        if new_adapter:
-            self.active_adapter = new_adapter
+        self.active_adapter = new_adapter or []
 
     def merge_and_unload(self, progressbar: bool = False, safe_merge: bool = False):
         r"""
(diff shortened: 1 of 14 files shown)


src/peft/tuners/lora/model.py
```diff
@@ -661,29 +661,16 @@ def delete_adapter(self, adapter_name: str):
...
-                    warnings.warn(
-                        f"Adapter {adapter_name} was active which is now deleted. Setting active adapter to {resetting_active_adapter}. "
-                    )
-                    target.set_adapter(resetting_active_adapter)
+                target.delete_adapter(adapter_name)
+                if new_adapter is None:
+                    new_adapter = target.active_adapters[:]
+
+        if new_adapter:
+            self.active_adapter = new_adapter
```
pacman100, line 673: is this required as `set_adapter` is being called in `delete_adapter` method?

src/peft/tuners/tuners_utils.py
```diff
@@ -351,6 +354,54 @@ def set_adapter(self, adapter_names: str | list[str]):
...
+            active_adapters = self.active_adapters[:]
+            active_adapters.remove(adapter_name)
+            if active_adapters:
+                self.set_adapter(active_adapters)
+            else:
+                # no active adapters left, set a new default adapter
+                # here we get the list of all adapters existing adapter names and choose the first one
+                remaining_adapters = self._all_available_adapter_names()
+                if not remaining_adapters:
+                    raise ValueError("You tried to delete the only adapter in the model, this is not possible.")
```
pacman100, line 396: let's support deleting all adapters, in that case it will simply run the base model forward, right?

tests/testing_common.py
```diff
@@ -815,42 +816,69 @@ def _test_training_prompt_learning_tasks(self, model_id, config_cls, config_kwar
...
+            base_model_name_or_path=model_id,
+            **config_kwargs,
+        )
+        if config.peft_type not in supported_peft_types:
             return
 
         model = self.transformers_class.from_pretrained(model_id)
+        if isinstance(config.target_modules, str):
+            # TODO this should be doable
+            self.skipTest("Multiple adapters cannot currently be added when target_modules is a string.")
```
pacman100, line 832: why is that?

pacman100 (commented):
Super work @BenjaminBossan for refactoring deletion, fixes bugs and nice tests! 🚀

pacman100:
> I did not add adapter deletion to IA³, since it is included in https://github.com/huggingface/peft/pull/980. LMK
if it should be added here instead.

Let's take this up later by suggesting in that PR the new way of deleting adapters of this PR

src/peft/tuners/lora/model.py
BenjaminBossan: Yes, unfortunately. Although we call `set_adapter` via `delete_adapter` on the target layers, the `LoraModel` itself also needs its active adapters updated because they are separate. If I comment out this line, tests start failing because the model's active adapter is still `delete_me`.

src/peft/tuners/tuners_utils.py
BenjaminBossan: Okay, I allowed this, also extended the tests to cover this.

Note that we cannot call `peft_model(**input)` when there is no active adapter at all. However, we can call `peft_model.base_model(**input)`, which is what I test.

tests/testing_common.py
BenjaminBossan: No idea, I haven't checked. But it's unrelated to this PR.

pacman100 (approved):
Thank you @BenjaminBossan

tests/testing_common.py
BenjaminBossan: Created a fix for that: https://github.com/huggingface/peft/pull/1111

PR #1069, opened 2023-11-01
Mixed adapter types

BenjaminBossan:
This is a POC to show how we could achieve mixing adapter types such as LoRA and LoKr.

## Description

The very general idea is that we can already mix multiple adapters of the same type, e.g. add two LoRA adapters, but right now we fail when trying to mix different types. This restriction has been lifted by adding a new class `PeftMixedModel` which deals with different adapter types.

The usage looks something like this:

```python
base_model = ...
config0 = LoraConfig(...)
# set mixed=True
peft_model = get_peft_model(base_model, config0, mixed=True)
config1 = LoHaConfig(...)
peft_model.add_adapter(config1, "other")
peft_model.set_adapter(["default", "other"])
peft_model(x)
```

At this point, both adapters are active at the same time.

Existing code should not be affected by this change, since users need to opt into this behavior by setting `mixed=True`.

Also interesting is that this method can be used for a single adapter type but with very different configs. Right now, we have limited support for that (e.g. for LoRA, different `r` values by using `rank_pattern`), but with this, we don't need to special case the differing arguments anymore.

## Implementation

Apart from adding the new `PeftMixedModel` class to replace `PeftModel`, I added a new class `LycorisModel` which replaces `LoraModel`, `LoHaModel` etc. This class checks the config type and then uses the corresponding `LoraModel`, `LoHaModel` etc. to create the adapter.

Another crucial change I had to make was to adopt the "base layer pattern". This is the pattern that was, for instance, used to speed up initialization in LoRA bnb layers in PR #994.

The main change is that the adapter layer wraps the original layer and calls forward on that layer, instead of doing stuff like this:

```python
F.linear(input, transpose(self.weight, self.fan_in_fan_out), bias=self.bias)
```

which completely circumvents the call to the target layer's `forward` method. With the base layer pattern, we now call the target layer's `forward` method. Therefore, if the target layer is another adapter layer, we call its forward method correctly.

This change has the nice side benefit that we no longer need to use `_init_empty_weights` -- in fact, we don't initialize any of the target layer's weights anymore, since we have a reference to it.

Note that same as for the bnb layers, this should not be backwards incompatible, since the adapter weights and their `state_dicts` are not affected by this change.

I pondered the possibility to implement this via hooks, which may be more elegant, but that would require larger changes in the adapter layers like LoRA `Linear.forward` to be rewritten to only return the diff. The chosen approach seemed less disruptive to me.

## Somewhat unrelated changes

[...]

Issue #1058, opened 2023-10-27
Numerical instabilities with LoHa


BenjaminBossan:
### System Info

Latest PEFT main.

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

```python
import torch
from torch import nn
from peft import LoraConfig, LoHaConfig, get_peft_model

class MLP(nn.Module):
    def __init__(self, bias=True):
        super().__init__()
        self.lin0 = nn.Linear(10, 20, bias=bias)
        self.relu = nn.ReLU()
        self.drop = nn.Dropout(0.5)
        self.lin1 = nn.Linear(20, 2, bias=bias)
        self.sm = nn.LogSoftmax(dim=-1)

    def forward(self, X):
        X = X.float()
        X = self.lin0(X)
        X = self.relu(X)
        X = self.drop(X)
        X = self.lin1(X)
        X = self.sm(X)
        return X

def test_it(config_cls, device):
    input = torch.arange(90).reshape(9, 10).to(device)
    torch.manual_seed(0)
    if issubclass(config_cls, LoraConfig):
        config = config_cls(target_modules=["lin1"], init_lora_weights=False)
    else:
        config = config_cls(target_modules=["lin1"], init_weights=False)
    peft_model = get_peft_model(MLP(), config).eval().to(device)
    output = peft_model(input)
    is_finite = torch.isfinite(output).all().item()
    return is_finite

if __name__ == "__main__":
    n = 100
    print("LoHa")

    device = "cpu"
    print("Testing on CPU")
    successes = [test_it(LoHaConfig, device) for _ in range(n)]
    print(f"Success rate: {sum(successes) / n}")

    device = "cuda"
    print("\nTesting on GPU")
    successes = [test_it(LoHaConfig, device) for _ in range(n)]
    print(f"Success rate: {sum(successes) / n}")

    print("Lora")

    device = "cpu"
    print("Testing on CPU")
    successes = [test_it(LoraConfig, device) for _ in range(n)]
    print(f"Success rate: {sum(successes) / n}")

    device = "cuda"
    print("\nTesting on GPU")
    successes = [test_it(LoraConfig, device) for _ in range(n)]
    print(f"Success rate: {sum(successes) / n}")
```

### Expected behavior

[...]

PR #994, opened 2023-10-05
ENH: Refactor LoRA bnb layers for faster initialization

BenjaminBossan:
Partly addresses #896

## Description

After speeding up normal LoRA layer initialization, this PR improves initialization speed of bnb LoRA layers.

The method to achieve this is different from the one used before, namely this time the base layer is stored as a reference on the LoRA layer. This allows us to avoid calling `__init__` on the bnb layer, which is what is slow.

## Notes

We cannot use the same method as for the normal LoRA layers, (i.e. calling the super class's `__init__` with meta device) because the bnb layers have extra logic that still creates unnecessary weights (this is something that could potentially be fixed by bnb).

However, the way used here could also be a solution to the normal layers, so if we want to have consistency, the normal layers could be refactored to use the same approach.

Another small advantage of this approach is that we can now correctly handle any `*args, **kwargs` that are passed to `forward`. If we had this consistently in all layers, it would allow us to handle extra arguments like `lora_scale` in the future.

Interestingly, even though we now save the base layer as a reference, which results in a different `state_dict`, the existing models can still be loaded successfully. This is because the adapter `state_dict` is not affected by the change, so users can still load their existing adapters.

The only problem would occur if users dump the whole model, i.e. base model and adapter, using `torch.save` and then trying to load with `torch.load`. For those users, we could theoretically provide a script to convert the `state_dict` (i.e. renaming some keys).

To ensure that the old adapters can still be loaded successfully, I'm working at the same time on adding regression tests. I'll create a separate PR for those to avoid blowing up this one.

## Tests

I ran a test on `bloomz-1b1` for how long it takes to create the `PeftModel`, the results are:

8bit: 1108.34 ms > 26.82 ms
4bit: 1101.96 ms > 23.69 ms

Script was:

```python
import time
from contextlib import contextmanager

import torch
from transformers import AutoModelForCausalLM, BitsAndBytesConfig
from peft import LoraConfig, get_peft_model

@contextmanager
def timed(msg):
    start = time.perf_counter()
    yield
    end = time.perf_counter()
    print(f"{msg} took {1000 * (end - start):.2f} ms")

torch.manual_seed(1000)
model_id = "bigscience/bloomz-1b1"
test_8bit = True

if test_8bit:
    model = AutoModelForCausalLM.from_pretrained(
        model_id,
        load_in_8bit=True,
    )
    config = LoraConfig(
        r=8,
        init_lora_weights=False,
    )
else:
    bnb_config = BitsAndBytesConfig(
[...]

PR #1106, opened 2023-11-09
Refactor base layer pattern

BenjaminBossan:
Refactor all tuners (where it applies) to use the "base layer pattern". This means that the adapter layer will always hold a reference to the original layer that it modifies. This pattern is already partly used (e.g. LoRA bnb, gptq layers), now it is consistently used everywhere it makes sense.

This PR is a companion PR to #1069, where I first added these changes. They are now extracted to a separate PR to make code review easier and to advance more quickly.

## Implementation

The main change is that the adapter layer wraps the original layer and calls forward on that layer, instead of doing stuff like this:

```python
F.linear(input, transpose(self.weight, self.fan_in_fan_out), bias=self.bias)
```

which completely circumvents the call to the target layer's `forward` method. With the base layer pattern, we now call the target layer's `forward` method. Therefore, if the target layer is another adapter layer (which will be crucial for mixed adapters), we call its forward method correctly. Also, this should allow passing extra arguments, like `lora_scale` to `forward`.

This change has the nice side benefit that we no longer need to use `_init_empty_weights` -- in fact, we don't initialize any of the target layer's weights anymore, since we have a reference to it. There is thus no risk of having slow but superfluous initialization of layers.

Moreover, I could greatly simplify `merge_and_unload` by just using the `base_layer` instead of having to create a completely new layer. I haven't measured it, but this should speed up the operation considerably.

Note that same as for the bnb layers, this should be backwards incompatible, since the adapter weights and their `state_dicts` are not affected by this change.

## Somewhat unrelated changes

1. During debugging, I got very annoyed with the fact that the `repr`s of adapter layers and normal PyTorch layers are hard to distinguish, e.g. the type is just `"Linear"`. Now, for adapter layers, it is prefixed by the adapter type, e.g. `"lora.Linear"`.
2. For LoHa and LoKr, I had to change the init of weights when using `init_weights=False`. This is because of what is discussed in #1058.

## TODOs

Before merging this, I would like to do some regression tests to ensure that everything still works as expected. This is especially important to ensure that old adapter checkpoints can still be loaded. In theory, this should work for the same reason as discussed in #994 but better be safe than sorry.


BenjaminBossan:
**Update**: Based on my regression testing PR (#995), I created a bunch of regression tests to check this PR against v0.6.1. This means that I created adapters with v0.6.1 and stored them, as well as the model output. Then I switched to this PR, loaded the adapters successfully and also produced the exact same output.

This confirms that the PR does not break existing code or checkpoints, which is what we expected based on how the changes are implemented, and based on the same refactor on bnb layers we merged earlier. This gives me high confidence that this PR is not BC breaking.

Here is the list of settings that I tested:

- adalora_mlp
- adalora_opt-350m
- ia3_conv2d
- ia3_mlp
- ia3_no_ff_mlp
- ia3_opt-350m
- lora_conv2d
- lora_emb_conv1d
- lora_mlp
- lora_mlp_modules_to_save
- lora_opt-350m
- lora_opt-350m_bnb_4bit
- lora_opt-350m_bnb_8bit

Note that I couldn't test a couple of settings:

1. LoHa and LoKr: When using `init_weights=False`, they are initialized with `torch.empty`, thus results are not reproducible, see #1058. In this very PR, that behavior is changed so that in the future, we get reproducible results with those tuners.
2. AdaLoRA with 4bit and 8bit. For 8bit, I got `TypeError: unsupported operand type(s) for +=: 'dict' and 'Tensor'`, for 4bit I got the issue described in #1113.

It would have been nice to include those, but I see very little danger that we get BC breakage in those.

pacman100 (commented):
Super work @BenjaminBossan on this huge refactoring to make things simple, readable, faster and robust! 🔥🚀✨

Left few comments

src/peft/tuners/lora/bnb.py
```diff
@@ -169,27 +167,29 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
 
             return result
 
+        def __repr__(self) -> str:
```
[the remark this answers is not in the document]
pacman100: it doesn't, checked locally.

src/peft/tuners/adalora/bnb.py
```diff
@@ -85,37 +74,28 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
 
 if is_bnb_4bit_available():
 
-    class SVDLinear4bit(bnb.nn.Linear4bit, AdaLoraLayer):
+    class SVDLinear4bit(torch.nn.Module, AdaLoraLayer):
```
younesbelkada, line 77: This might be breaking no? Some setups might consider some checks such as `isinstance(module, bnb.nn.Linear4bit)`, suddenly these layers will not inherit anymore from bnb layers

src/peft/tuners/adalora/bnb.py
```diff
@@ -85,37 +74,28 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
         # Low-rank matrix for SVD-based adaptation
         def __init__(
             self,
-            adapter_name,
-            in_features,
-            out_features,
+            base_layer: torch.nn.Module,
```
younesbelkada, line 81: For backward compatibility we should maybe consider keeping the same arg order, ignore the unused kwargs and put the new arguments at the end. This might be too much of an edge case but we never know. What do you think?

src/peft/tuners/__init__.py
```diff
@@ -27,10 +27,3 @@
 from .prefix_tuning import PrefixEncoder, PrefixTuningConfig
 from .prompt_tuning import PromptEmbedding, PromptTuningConfig, PromptTuningInit
 from .multitask_prompt_tuning import MultitaskPromptEmbedding, MultitaskPromptTuningConfig, MultitaskPromptTuningInit
-
-# Mapping of tuners that support direct plugging
```
younesbelkada, line 31: Why this has been removed?

src/peft/tuners/adalora/layer.py
```diff
@@ -73,32 +82,31 @@ def reset_lora_parameters(self, adapter_name):
             nn.init.normal_(self.lora_B[adapter_name], mean=0.0, std=0.02)
 
 
-class SVDLinear(nn.Linear, AdaLoraLayer):
+class SVDLinear(nn.Module, AdaLoraLayer):
     # SVD-based adaptation by a dense layer
     def __init__(
         self,
+        base_layer: nn.Module,
```
younesbelkada, line 89: Same comment above about BC, I would slowly deprecate them instead of changing the signature of these public classes

src/peft/tuners/adalora/layer.py
```diff
@@ -73,32 +82,31 @@ def reset_lora_parameters(self, adapter_name):
...
-        if fan_in_fan_out:
-            self.weight.data = self.weight.data.T
-
-        nn.Linear.reset_parameters(self)
+        self._active_adapter = adapter_name
         self.update_layer(adapter_name, r, lora_alpha, lora_dropout, init_lora_weights)
-        self.set_adapter(adapter_name)
+
+    @property
+    def weight(self) -> torch.Tensor:
```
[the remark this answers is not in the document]
younesbelkada: Same question

src/peft/tuners/adalora/layer.py
```diff
@@ -149,19 +158,16 @@ def get_delta_weight(self, adapter) -> torch.Tensor:
             / (self.ranknum[adapter] + 1e-5)
         )
 
-    def _linear(self, input: torch.Tensor) -> torch.Tensor:
-        return F.linear(input, transpose(self.weight, self.fan_in_fan_out), bias=self.bias)
-
-    def forward(self, x: torch.Tensor) -> torch.Tensor:
+    def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
```
younesbelkada, line 161: Nice!

src/peft/tuners/adalora/layer.py
```diff
@@ -176,8 +182,12 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
 
         return result
 
+    def __repr__(self) -> str:
+        rep = super().__repr__()
+        return "adalora." + rep
```
younesbelkada, line 187: Can you maybe add few comments here elaborating: 

> During debugging, I got very annoyed with the fact that the reprs of adapter layers and normal PyTorch layers are hard to distinguish, e.g. the type is just "Linear". Now, for adapter layers, it is prefixed by the adapter type, e.g. "lora.Linear".

src/peft/tuners/adalora/model.py
```diff
@@ -212,7 +211,7 @@ def _create_new_module(lora_config, adapter_name, target, **kwargs):
                     f"Target module {target} is not supported. "
                     f"Currently, only `torch.nn.Linear` and `Conv1D` are supported."
                 )
-            new_module = SVDLinear(adapter_name, in_features, out_features, bias=bias, **kwargs)
+            new_module = SVDLinear(target, adapter_name, bias=bias, **kwargs)
```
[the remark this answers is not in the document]
younesbelkada: Indeed, `bias` seems to be not present in the init signature of these classes (and also not used above)

src/peft/tuners/loha/layer.py
```diff
@@ -171,96 +193,109 @@ def get_delta_weight(self, adapter_name: str) -> torch.Tensor:
...
     """LoHa implemented in Linear layer"""
 
     def __init__(
         self,
-        in_features: int,
-        out_features: int,
-        bias: bool = True,
-        device: Optional[Union[str, torch.device]] = None,
-        dtype: Optional[torch.dtype] = None,
+        base_layer: nn.Module,
```
younesbelkada, line 228: Same comment about init signature

younesbelkada (commented):
Thanks a lot for your great work @BenjaminBossan ! I left few open questions in addition to what @pacman100 said ! 
I think we should run integration tests before merging, as well as slow tests to make sure everything works as expected

BenjaminBossan (commented):
@younesbelkada Thanks for the review. You brought up some good comments about backwards compatibility in case users use certain classes directly (not through `get_peft_model` etc.). I think it's basically impossible to make the changes suggested here without breaking this very low level API. I believe there will be very few, if any, users who do this (maybe transformers integration??). Also note that we already adopted this pattern partly (e.g. bnb LoRA layers) and so far there have been no complaints about it. Therefore, my suggestion would be to accept this possible breaking change.

However, what we could do is to add a migration guide somewhere (maybe docs?). There, I could write down what users who actually use these classes directly should be doing to make their code work. Do you think this will be good enough?

src/peft/tuners/adalora/bnb.py
younesbelkada: ok sounds great yeah, I think this is too much of an edge case, let's just be careful on the release notes and maybe add this PR on the next minor release (0.7.0)

src/peft/tuners/adalora/bnb.py
younesbelkada: agreed!

src/peft/tuners/adalora/layer.py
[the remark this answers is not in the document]
pacman100: Thank you for the info, helps!

pacman100 (approved):
Thank you @BenjaminBossan for the great work, LGTM! 🚀

younesbelkada (approved):
Impressive and inspiring work @BenjaminBossan !

PR #1107, opened 2023-11-09
Added num_dataloader_workers arg to fix Windows issue

lukaskuhn-lku:
Running the train_dreambooth.py script on Windows leads to the common issue that the pytorch dataloader crashes or is extremly slow because num_workers > 0 is not well supported ([Pytorch Issue#12831](https://github.com/pytorch/pytorch/issues/12831)).

I have therefore added the parser arg `--num_dataloader_workers` which by default is set to 1 (same default as before but now possible to configure from the outside). I also added a small sentence to the documentation, warning others about this common pitfall.

### DIFF
--- a/docs/source/task_guides/dreambooth_lora.mdx
+++ b/docs/source/task_guides/dreambooth_lora.mdx
@@ -83,6 +83,7 @@ accelerate launch train_dreambooth.py \
   --output_dir=$OUTPUT_DIR \
   --train_text_encoder \
   --with_prior_preservation --prior_loss_weight=1.0 \
+  --num_dataloader_workers=1 \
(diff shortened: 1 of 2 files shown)


BenjaminBossan (approved):
Thanks for adding the argument, I think it useful to have.

Just a question, if you set `persistent_workers=True` (while leaving `num_workers=1`), do you still run into the slowdown? If not, we could consider enabling this for Windows users.

lukaskuhn-lku:
Thanks! 

Yes, I still run into the issue if I set `num_workers=1` and `persistent_workers=True`, so this doesn't fix the Windows issues.

BenjaminBossan (approved):
Thanks for checking. The change LGTM.

PR #1109, opened 2023-11-10
avoid unnecessary import

winglian:
there is no need to import auto_gptq if we're not going to use it. We're seeing issues downstream with the latest auto-gptq if it happens to be installed but not needed.


BenjaminBossan (approved):
Thanks, this is a reasonable change.

PR #1112, opened 2023-11-10
Correctly deal with  `ModulesToSaveWrapper` when using Low-level API

younesbelkada:
# What does this PR do?

Currently when using the low-level API of PEFT, the `modules_to_save` gets silently ignored. The PR addressed that.

cc @BenjaminBossan @pacman100 

Will also test the integration tests and add corresponding tests on transformers once we release a PEFT version with it

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -25,7 +25,7 @@ from torch import nn
 from peft.utils import COMMON_LAYERS_PATTERN
 
 from ..config import PeftConfig
-from ..utils import _get_submodules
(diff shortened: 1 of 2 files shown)


younesbelkada:
Integration tests reports (passed) are here: https://github.com/huggingface/peft/actions/runs/6826639872/job/18566990274

BenjaminBossan (approved):
Thanks for addressing this oversight, looks good.

pacman100 (approved):
Thank you @younesbelkada for fixing the low-level API to account for additional trainable modules support! 🤗

PR #1115, opened 2023-11-10
TST: Add regression tests 2

BenjaminBossan:
This PR supersedes #995. The description below is copied and modified from that PR. For some technical reasons, it was easier for me to create a new PR than to update the previous one, sorry for that.

## Description

In general, for regression tests, we need two steps:

1. Creating the regression artifacts, in this case the adapter checkpoint and the expected output of the model.
2. Running the regression tests, i.e. loading the adapter and checking that the output of the model is the same as the expected output.

My approach is to re-use as much code as possible between those two steps. Therefore, the same test script can be used for both, with only an environment variable to distinguish between the two. Step 1 is invoked by calling:

`REGRESSION_CREATION_MODE=True pytest tests/regression/test_regression.py`

and to run the second step, we call:

`pytest tests/regression/test_regression.py`

### Creating regression artifacts

The first step will create an adapter checkpoint and an output for the given PEFT version and test setting in a new directory. E.g. it will create a directory `tests/regression/lora_opt-125m_bnb_4bit/0.5.0/` that contains adapter_model.bin and output.pt.

Before this step runs, there is a check that the git repo is clean (no dirty worktree) and that the commit is tagged (i.e. corresponds to a release version of PEFT). Otherwise, we may accidentally create regression artifacts that do not correspond to any PEFT release.

The easiest way to get such a clean state (say, for PEFT v0.5.0) is by checking out a tagged commit, e.g:

`git checkout v0.5.0`

before running the first step.

The first step will also skip the creation of regression artifacts if they already exist.

It is possible to circumvent all the aforementioned checks by setting the environment variable `REGRESSION_FORCE_MODE` to True like so:

`REGRESSION_FORCE_MODE=True REGRESSION_CREATION_MODE=True pytest tests/regression/test_regression.py`

You should only do this if you know exactly what you're doing.

### Running regression tests

The second step is much simpler. It will load the adapters and the output created in the first step, and compare the output to the output from a new PEFT model using the loaded adapter. The outputs should be the same.

If more than one version is discovered for a given test setting, all of them are tested.

## Notes

As is, the adapters are stored in the git repo itself. Since they're relatively small, the total size of the repo is still reasonable. However, it could be better to store those adapters on HF Hub instead. This would, however, make things a bit more complicated (not sure how to parse directories etc. on Hub).

The regression tests in this included in this PR were used to check that #1106 still allows to load checkpoints created with PEFT v0.6.1.


pacman100 (commented):
Thank you @BenjaminBossan for adding the extensive regression tests along with cmd configurations to automatically enable/skip regression tests 🚀!

One major issue for this approach is that it will bloat the package heavily and we should actively avoid that. I would prefer that we create the models and outputs and save them as revisions based on version on HF Hub and load it from there for the regression tests.

BenjaminBossan:
> One major issue for this approach is that it will bloat the package heavily and we should actively avoid that. I would prefer that we create the models and outputs and save them as revisions based on version on HF Hub and load it from there for the regression tests.

Yes, this increases the size when cloning the repo, though the pip package is unaffected. I'll investigate how the artifacts can be uploaded to the hub instead.

BenjaminBossan:
@pacman100 I think your concern should be addressed. I changed the code so that all regression artifacts are stored on HF Hub. You can check them here: https://huggingface.co/peft-internal-testing/regression-tests/tree/main

Note that I had to do a bit of trickery to get the regression artifacts for PEFT v0.6.2 because regression tests need to be run on a clean tagged commit, but of course for v0.6.2 the regression tests didn't exist yet. The output should be correct though. I also changed the code to use safetensors, since that's our new standard.

In the future, if this PR is merged, no trickery should be necessary to create the regression artifacts.

I ran a regression test on current main with the artifacts from v0.6.2 and they all passed.

For now, the regression tests are not run automatically on CI. Once we agree if and how we want to do that, I can add that in a separate PR.

(Note that github shows `Files changed 57` but those files have been cleaned up, so after a squash and merge, they should not be part of the git history)

BenjaminBossan:
@younesbelkada Thanks for the review. I tried importing the two copied functions but always get an

> ImportError: attempted relative import with no known parent package

If you know a solution to this, please LMK.

Regarding the AdaLora tests: I did a check and they're still failing. Not sure why. I'd propose to investigate those later.

The unnecessary comment was removed.

younesbelkada (approved):
Thanks ! Can we also add a job on the nightly runners to run the regression tests on GPU?

PR #1122, opened 2023-11-12
Add LoftQ to LoRA tuner

yxli2123:
Add LoftQ for LoRA initialization. People can pass `init_lora_weights='loftq'` to apply LoftQ. The original `init_lora_weights=True` will be changed to `init_lora_weights='zero'`, which uses the initialization in LoRA paper.


yxli2123:
> Add LoftQ for LoRA initialization. People can pass `init_lora_weights='loftq'` to apply LoftQ. The original `init_lora_weights=True` will be changed to `init_lora_weights='zero'`, which uses the initialization in LoRA paper.

Here's the LoftQ paper link: https://arxiv.org/abs/2310.08659

src/peft/tuners/lora/config.py
```diff
@@ -76,12 +76,13 @@ class LoraConfig(PeftConfig):
             "the final layer `classifier/score` are randomly initialized and as such need to be trainable and saved."
         },
     )
-    init_lora_weights: bool = field(
-        default=True,
+    init_lora_weights: Optional[str] = field(
```
pacman100, line 79: this will lead to tests failing as they are passing boolean flag `False` in many of them.

src/peft/tuners/lora/layer.py
```diff
@@ -200,21 +234,21 @@ def __init__(
         lora_dropout: float = 0.0,
         fan_in_fan_out: bool = False,  # Set this to True if the layer to replace stores weight like (fan_in, fan_out)
         is_target_conv_1d_layer: bool = False,
+        weight_copy: Optional[torch.Tensor] = None,  # pass weight for LoftQ
         **kwargs,
     ) -> None:
-        init_lora_weights = kwargs.pop("init_lora_weights", True)
+        init_lora_weights = kwargs.pop("init_lora_weights", "zero")
```
pacman100, line 240: this change isn't reflected for the other layers.

src/peft/tuners/lora/model.py
```diff
@@ -318,8 +317,42 @@ def _create_new_module(lora_config, adapter_name, target, **kwargs):
...
+            if lora_config.loftq_fake:
+                new_module = Linear(
+                    adapter_name,
+                    in_features,
+                    out_features,
+                    weight_copy=target.weight if lora_config.init_lora_weights == "loftq" else None,
+                    bias=bias,
+                    **kwargs,
+                )
+            elif loaded_in_8bit or lora_config.loftq_bits == 8:
```
pacman100, line 331: lines 261 and 272 would run when loaded in 8 bit and 4 bit, respectively. This and the below logic would never run if the model is already loaded in 8/4 bits.

src/peft/utils/other.py
```diff
@@ -464,6 +465,194 @@ def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
     return tensor.device, unique_id, storage_size(tensor)
 
 
+class NFQuantizer:
```
pacman100, line 468: I think this should be placed in `utils/loftq_utils.py`

pacman100 (changes requested):
Thank you @yxli2123 for the awesome work on integration LoftQ which is an innovative way of initializing the LoRA weights to be quantization aware ✨! Left few comments that needs to be addressed.

src/peft/tuners/lora/config.py
yxli2123: I have changed logics to compare `if init_lora_weights ==True or init_lora_weights == 'zero'` and `elif init_lora_weights  == 'loftq'`

yxli2123:
I have changed the code. Please review them again. Also, I'm not able to run the test files in my local machine. Are the results from the test files a mandatory requirement?

src/peft/tuners/lora/bnb.py
```diff
@@ -35,14 +36,17 @@ def __init__(
...
             super().__init__()
-            LoraLayer.__init__(self, in_features=base_layer.in_features, out_features=base_layer.out_features)
+            LoraLayer.__init__(
+                self, in_features=base_layer.in_features, out_features=base_layer.out_features, **kwargs
+            )
             self.base_layer = base_layer
 
-            init_lora_weights = kwargs.pop("init_lora_weights", True)
-            self.update_layer(adapter_name, r, lora_alpha, lora_dropout, init_lora_weights)
+            init_lora_weights = kwargs.pop("init_lora_weights", 'zero')
```
BenjaminBossan, line 48: Let's not add the `"zero"` option. Can we not just have `init_lora_weights={True, False, "loftq"}`?

src/peft/tuners/lora/config.py
```diff
@@ -76,12 +76,13 @@ class LoraConfig(PeftConfig):
             "the final layer `classifier/score` are randomly initialized and as such need to be trainable and saved."
         },
     )
-    init_lora_weights: bool = field(
+    init_lora_weights: Optional[Union[str, bool]] = field(
```
BenjaminBossan, line 79: Why do we need `Optional` here? `Union[str, bool]` seems sufficient.

src/peft/tuners/lora/config.py
```diff
@@ -76,12 +76,13 @@ class LoraConfig(PeftConfig):
...
         default=True,
         metadata={
             "help": (
-                "Whether to initialize the weights of the Lora layers with their default initialization. Don't change "
-                "this setting, except if you know exactly what you're doing."
+                "Whether to initialize the weights of the Lora layers with their default initialization. "
+                "Pass `True` or `'zero'` to initialize the weights with A ~ Gaussian and B=0."
+                "Pass `'loftq'` to initialize the weights with the LoftQ initialization."
```
BenjaminBossan, line 85: Can we maybe add a sentence about when users should use loftq initialization. Most users will not be aware. Moreover, let's add a sentence that passing `False` should generally be avoided and is for debugging purposes.

src/peft/tuners/lora/config.py
```diff
@@ -117,6 +118,25 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    bits_pattern: Optional[dict] = field(
```
BenjaminBossan, line 121: Hmm, there are a bunch of new parameters. Maybe it would be better to have a separate `LoftQConfig` class only for the LoftQ-parameters. We could then pass this config to `LoraConfig`. So in practice, it would look something like this for users:

```python
loftq_config = LoftQConfig(bits_pattern=..., loftq_bits=...)
lora_config = LoraConfig(r=16, ..., loftq_config=loftq_config)
```

By default, `loftq_config` on `LoraConfig` would be `None`, but if users pass it, it means that they want to use LoftQ-initialization. How does that sound?

@pacman100 WDYT about that?

src/peft/tuners/lora/layer.py
```diff
@@ -60,7 +65,8 @@ def _init_empty_weights(self, cls, *args, **kwargs) -> None:
         cls.__init__(self, *args, device="meta", **kwargs)
         self.to_empty(device=final_device)
 
-    def update_layer(self, adapter_name, r, lora_alpha, lora_dropout, init_lora_weights):
+    def update_layer(self, adapter_name, r, lora_alpha, lora_dropout,
+                     init_lora_weights: Optional[Union[str, bool]], weight_copy: Optional[torch.Tensor] = None):
```
BenjaminBossan, line 69: I think we can avoid the need to have `weight_copy` as a new argument to `update_layer` and all of the other changes related to that. This requires #1106 to be merged first. To summarize the change from that PR, it will result in the LoRA layers having a new attribute, `self.base_layer`, which is the reference to the original layer that we're adopting. You can already see that on the bnb layers. Therefore, inside of the `loftq_init_parameters` method, we will be able to fetch the weight by calling `self.base_layer.weight` (or, to be precise, `self.get_base_layer().weight`). I hope that the PR will be merged soon, it should make this PR easier overall.

src/peft/tuners/lora/layer.py
```diff
@@ -151,6 +167,23 @@ def reset_lora_parameters(self, adapter_name):
...
+    def loftq_init_parameters(self, adapter_name, weight, num_bits, num_iter):
+        qweight, lora_A, lora_B = loftq_init(weight, num_bits, self.r[adapter_name], num_iter)
+        if adapter_name in self.lora_A.keys():
+            self.lora_A[adapter_name].weight.data = lora_A
+            self.lora_B[adapter_name].weight.data = lora_B
+        if adapter_name in self.lora_embedding_A.keys():
+            self.lora_embedding_A[adapter_name].data = lora_A
+            self.lora_embedding_B[adapter_name].data = lora_B
+        # TODO: better deal with this instead of creating weight here
+        if getattr(self, "base_layer", None) is None:
```
BenjaminBossan, line 179: Can you please explain what is happening here? Are we converting the weight of the original layer to a quantized weight? Why is that required? I would rather only apply LoftQ when the weights are already quantized, as it is a requirement for LoftQ, right? We could raise an error if a user wants to apply LoftQ on non-quantized weights.

src/peft/tuners/lora/layer.py
```diff
@@ -151,6 +167,23 @@ def reset_lora_parameters(self, adapter_name):
...
+            self.weight = nn.Parameter(qweight.to(weight.dtype), requires_grad=False)
+        elif is_bnb_available():
```
BenjaminBossan, line 181: Similar question: Should the weights not already be quantized? Is this to move them to CPU? If yes, why is that needed?

src/peft/tuners/lora/layer.py
```diff
@@ -553,11 +589,11 @@ def get_delta_weight(self, adapter) -> torch.Tensor:
         else:
             # conv2d 3x3
             output_tensor = (
-                F.conv2d(
-                    weight_A.permute(1, 0, 2, 3),
-                    weight_B,
-                ).permute(1, 0, 2, 3)
-                * self.scaling[adapter]
+                    F.conv2d(
```
BenjaminBossan, line 592: Let's undo these unrelated changes.

src/peft/tuners/lora/model.py
```diff
@@ -257,7 +257,6 @@ def _create_new_module(lora_config, adapter_name, target, **kwargs):
         loaded_in_8bit = kwargs.pop("loaded_in_8bit", False)
         loaded_in_4bit = kwargs.pop("loaded_in_4bit", False)
         bias = kwargs.pop("bias", False)
-
```
BenjaminBossan, line 260: Please undo these unrelated changes.

src/peft/tuners/lora/config.py
```diff
@@ -117,6 +118,25 @@ class LoraConfig(PeftConfig):
...
+        default_factory=dict,
+        metadata={
+            "help": (
+                "The mapping from layer names or regexp expression to ranks which are different from the default bits specified by `bits`. "
+                "For example, `{model.decoder.layers.0.encoder_attn.k_proj: 2`}"
+            )
+        },
+    )
+    loftq_bits: Optional[int] = field(default=4, metadata={"help": "Quantization bits for LoftQ"})
```
BenjaminBossan, line 130: The help for this option, as well as for `loftq_iter` is a bit short. Could you please elaborate, e.g. what are valid options (IIUC only 4 and 8), or at least provide a reference for users to better understand the options?

src/peft/tuners/lora/config.py
```diff
@@ -117,6 +118,25 @@ class LoraConfig(PeftConfig):
...
+    loftq_iter: Optional[int] = field(default=1, metadata={"help": "Alternating iterations for LoftQ"})
+    loftq_fake: Optional[bool] = field(
+        default=True,
+        metadata={
+            "help": "True: use fp16/fp32; used for first time to save weights."
+            "False: use bitsandbytes 4bit linear models. weights can't be saved."
+            "Recommend to set to True, save the weights and load the saved weights in 4 bits."
```
BenjaminBossan, line 137: Could you please explain why this is needed? What weights are we talking about? Usually, we only save the adapter weights, not the weights of the original model.

src/peft/utils/loftq_utils.py
```diff
@@ -0,0 +1,190 @@
+import torch
+from scipy.stats import norm
+
+
+class NFQuantizer:
+    def __init__(self, num_bits=2, device="cuda", method="normal", block_size=64, *args, **kwargs):
```
BenjaminBossan, line 6: Is there any way for the user to control what `method` is being used here? If not, should it be added as an option or should the `uniform` method just be removed?

src/peft/utils/loftq_utils.py
```diff
@@ -0,0 +1,190 @@
+import torch
```
BenjaminBossan, line 1: AFAICT, this code is mostly the same as from your LoftQ repo. Could you please add a reference?

src/peft/tuners/lora/model.py
BenjaminBossan: Could you please elaborate on that? What are the constraints? I think it would be much more intuitive if we require the base model to be quantized when users want to apply LoftQ initialization.

BenjaminBossan (changes requested):
Thank you so much for adding the LoftQ initialization to PEFT. This looks very promising and useful. I assume that you are the first author of the paper?

I had a couple of comments regarding the implementation. There are a few areas where I think we can improve it so that it's as simple as possible for users to apply. I also think that part of the implementation can be simplified if we merge #1106 first, I mention that in one of the comments too.

Could you please also run `make style` on your changes?

After we have streamlined the design, we also need to work on adding examples/tests/documentation. This doesn't need to be 100% complete, but as is, users would have a hard time discovering the feature and knowing when and how to use it. But let's work on this once we have figured out the implementation.

PR #1129, opened 2023-11-14
Add Poly

TaoSunVoyage:
Add codes for the following papers:
- https://arxiv.org/abs/2202.13914
- https://arxiv.org/abs/2211.03831

with official implementation at https://github.com/microsoft/mttl


TaoSunVoyage:
Hi @pacman100, could you please have a look at this PR?

I'd like to implement two methods, [Polytropon](https://arxiv.org/abs/2202.13914) and [Multi-Head Routing ](https://arxiv.org/abs/2211.03831). And their official implementation is at https://github.com/microsoft/mttl.

BenjaminBossan:
Thanks a lot for providing this tuning method. I didn't have a closer look yet, but wanted to ask if it was possible to add some tests and documentation? Or would you have a working example of using the method to fine-tune a model?

Some questions I had at first glance:

1. In the linked repo, it seems that the method is applied to both LoRA and IA³. Is your implementation only for LoRA?
2. How is the router being used?

TaoSunVoyage:
Hi @BenjaminBossan, thanks for your comments. Sure, I do have some working examples and I will add a clean tutorial notebook for the method. 

Regrading your questions:

1. Yes, i only have it for LoRA now but happy to apply it for IA3 later, or maybe some other methods as well. 
2. The router can be viewed as a learnable weighting vector. it's used to combine different adapters like LoRAs.

BenjaminBossan:
> Yes, i only have it for LoRA now but happy to apply it for IA3 later, or maybe some other methods as well.

That would be fantastic, thanks!

> The router can be viewed as a learnable weighting vector. it's used to combine different adapters like LoRAs.

Thanks for explaining, I'll probably have to read the details of the papers first.

src/peft/tuners/poly/config.py
```diff
@@ -0,0 +1,57 @@
+from dataclasses import dataclass, field
+from typing import List, Optional, Union
+
+from peft.config import PeftConfig
+from peft.utils import PeftType
+
+
+@dataclass
+class PolyConfig(PeftConfig):
```
BenjaminBossan, line 9: Let's add a docstring. This would also be a good place to reference the paper(s).

src/peft/tuners/poly/layer.py
```diff
@@ -0,0 +1,153 @@
...
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from peft.tuners.tuners_utils import BaseTunerLayer
+from .config import PolyConfig
+from .router import get_router
+
+
+class PolyLayer(BaseTunerLayer):
```
BenjaminBossan, line 12: Let's add a docstring.

src/peft/tuners/poly/layer.py
```diff
@@ -0,0 +1,153 @@
...
+                A = torch.einsum("bqs,qsdr->bqdr", (mixing_weights, poly_lora_A))
+                B = torch.einsum("bqs,qsrd->bqrd", (mixing_weights, poly_lora_B))
+
+                A = A.reshape(bs, self.in_features, r)
+                B = B.transpose(1, 2).reshape(bs, r, self.out_features)
+
+                result += x.bmm(A).bmm(B) / r
+
+        result = result.to(previous_dtype)
+        return result
```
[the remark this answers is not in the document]
TaoSunVoyage: done.

src/peft/tuners/poly/layer.py
```diff
@@ -0,0 +1,153 @@
...
+                poly_config.n_splits,
+                poly_config.n_skills,
+                poly_config.r,
+                self.out_features // poly_config.n_splits,
+            )
+        )
+        self.poly_router[adapter_name] = get_router(poly_config)
+
+        if poly_config.init_poly_weights:
+            self.reset_poly_parameters(adapter_name)
```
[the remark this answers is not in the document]
TaoSunVoyage: done. as suggested, when `init_weights=False`, `lora_B` is initialized randomly.

src/peft/peft_model.py
```diff
@@ -813,6 +819,18 @@ def forward(
...
+                    input_ids=input_ids,
+                    attention_mask=attention_mask,
+                    inputs_embeds=inputs_embeds,
+                    labels=labels,
+                    output_attentions=output_attentions,
+                    output_hidden_states=output_hidden_states,
+                    return_dict=return_dict,
+                    task_ids=task_ids,
+                    **kwargs,
+                )
```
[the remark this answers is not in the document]
TaoSunVoyage: done. i did it this way:
```python
if peft_config.peft_type == PeftType.POLY:
    kwargs["task_ids"] = task_ids
return self.base_model(
    ...,
    **kwargs,
)
```

src/peft/tuners/poly/config.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+    modules_to_save: Optional[List[str]] = field(
+        default=None,
+        metadata={
+            "help": "List of modules apart from Poly layers to be set as trainable and saved in the final checkpoint. "
+            "For example, in Sequence Classification or Token Classification tasks, "
+            "the final layer `classifier/score` are randomly initialized and as such need to be trainable and saved."
+        },
+    )
+    init_poly_weights: bool = field(
```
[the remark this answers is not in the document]
TaoSunVoyage: done

src/peft/tuners/poly/config.py
```diff
@@ -0,0 +1,57 @@
...
+            ),
+        },
+    )
+
+    poly_type: str = field(
+        default="poly",
+        metadata={"help": "Type of Poly module to be used."},
+    )
+    n_tasks: int = field(
+        default=None,
```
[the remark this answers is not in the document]
TaoSunVoyage: yep, i set the default value to 1.

src/peft/tuners/poly/layer.py
```diff
@@ -0,0 +1,153 @@
...
+                    torch.nn.init.kaiming_uniform_(param, a=math.sqrt(5))
+                    self.poly_lora_A[adapter_name].data[split, skill, :, :] = param.T
+
+            torch.nn.init.zeros_(self.poly_lora_B[adapter_name])
+
+            # initialized router
+            self.poly_router[adapter_name].reset()
+
+
+class Linear(nn.Linear, PolyLayer):
```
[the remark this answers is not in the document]
TaoSunVoyage: done, i have updated it following the new `LoRA` implementation.

src/peft/tuners/poly/layer.py
```diff
@@ -0,0 +1,153 @@
...
+    # Lora implemented in a dense layer
+    def __init__(
+        self,
+        adapter_name: str,
+        in_features: int,
+        out_features: int,
+        poly_config: PolyConfig,
+        task_id_ptr: dict,
+        **kwargs: object,
```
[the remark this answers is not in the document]
TaoSunVoyage: oh yes, have fixed it

src/peft/tuners/poly/layer.py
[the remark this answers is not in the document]
TaoSunVoyage: done, i have updated it following the new `LoRA` implementation.

src/peft/tuners/poly/layer.py
```diff
@@ -0,0 +1,153 @@
...
+                    continue
+
+                r = self.r[active_adapter]
+                poly_router = self.poly_router[active_adapter]
+                poly_lora_A = self.poly_lora_A[active_adapter]
+                poly_lora_B = self.poly_lora_B[active_adapter]
+
+                task_ids = self.task_ids
+                repeat = x.size(0) // task_ids.size(0)
+                # this repeat follows the patten in `model.predict()` line 152
```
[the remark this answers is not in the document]
TaoSunVoyage: i dropped these lines as `repeat` is supposed to be 1 all the time.

src/peft/tuners/poly/config.py
TaoSunVoyage: done!

src/peft/tuners/poly/model.py
```diff
@@ -0,0 +1,129 @@
...
+        new_module.weight = child.weight
+        if hasattr(child, "bias"):
+            new_module.bias = child.bias
+        if getattr(child, "state", None) is not None:
+            new_module.state = child.state
+            new_module.to(child.weight.device)
+
+        # dispatch to correct device
+        for name, module in new_module.named_modules():
+            if "poly_" in name:
```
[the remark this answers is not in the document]
TaoSunVoyage: done

TaoSunVoyage:
Hi @BenjaminBossan, thanks for the detailed comments! I have updated the implementation following your suggestions. Would you have time for a look?

> * AFAICT, MHR and its variants from the 2nd paper is not implemented, is that correct? Do you have an idea if the existing implementation could be extended (in a future PR) to add those as well?

Actually no, as implemented in the MS repo, one can use the variant `MHR` from the 2nd paper by simply setting `n_splits` greater than 1.

> * Right now, there are no examples/docs/tests. We can help by creating the tests in a separate PR, but it would be great to have at least one of docs or examples. Otherwise, it will be almost impossible for users to discover this newly added adapter.

I have added a simple example notebook using Flan-T5 and SuperGLUE. Please check `examples/poly/peft_poly_seq2seq_with_generate.ipynb`

> * If you were copying some code directly from the MS repo, could you please mark the code as such (possibly adding permalinks too). That way, we can more easily reference the code.

Yep, i have also added permalinks in the code.

src/peft/tuners/poly/model.py
```diff
@@ -0,0 +1,151 @@
...
+            target_base_layer = target.get_base_layer()
+        else:
+            target_base_layer = target
+
+        if isinstance(target_base_layer, torch.nn.Linear):
+            return Linear(target, adapter_name, poly_config, **kwargs)
+        else:
+            raise ValueError(
+                f"Target module {target} is not supported. Currently, only the following modules are supported: "
+                "`torch.nn.Linear`, `torch.nn.Embedding`, `torch.nn.Conv2d`, `transformers.pytorch_utils.Conv1D`."
```
BenjaminBossan, line 95:
```suggestion
                "`torch.nn.Linear`."
```

src/peft/tuners/poly/model.py
```diff
@@ -0,0 +1,151 @@
...
+            if model_config["model_type"] not in TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING:
+                raise ValueError("Please specify `target_modules` in `peft_config`")
+            peft_config.target_modules = set(
+                TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING[model_config["model_type"]]
+            )
+        return peft_config
+
+    @classmethod
+    def _add_task_id_pre_hook(cls, module, args, kwargs):
+        task_ids = kwargs.get("task_ids", None)
```
BenjaminBossan, line 142: Well, this is not quite what I had in mind. My idea was to get rid of the `.task_id` attribute completely. Instead, the forward hook would be added to `poly.Linear` and would inject the `task_id` argument to the `kwargs` of `poly.Linear`.

I guess this method also works, but WDYT?

BenjaminBossan (changes requested):
Fantastic, thanks for making the adjustments. This already looks quite good.

Could you please run `make style` so that the CI passes?

Apart from that, I only have a few smaller comments left. At the end of the day, we should also add docs and tests, but it's fine if they are added in a subsequent PR.

> Actually no, as implemented in the MS repo, one can use the variant MHR from the 2nd paper by simply setting n_splits greater than 1.

Oh I see, thanks for clarifying. Is there also an equivalent to 𝙼𝙷𝚁 -z and 𝙼𝙷𝚁 -μ?

src/peft/tuners/poly/model.py
TaoSunVoyage: i was also wondering how to make this work properly and "Pythonically"
```python
# when use `PeftModel.forward`
PeftModel.forward(..., task_ids=...)
PeftModel.base_model.forward(..., task_ids=...)
# we in fact call `PolyModel.forward`
PolyModel.forward(..., task_ids=...)
# so i added the hook here and make `task_ids` as an attribute of `poly.Linear`
```
i am not sure how to inject `task_ids` into certain sub-modules like `poly.Linear` of one model like `PolyModel` as you suggested. sub-modules are somehow chained together? - one's input is another's output.

and most importantly, sometimes the `kwargs` is not given to `Linear` layers in`forward` (see Llama below), so it won't be passed into any pre-foward hook? i guess adding a hook here for `{q,k,v}_proj` would not help?  🤔
https://github.com/huggingface/transformers/blob/514de24abfd4416aeba6a6455ad5920f57f3567d/src/transformers/models/llama/modeling_llama.py#L366C13-L366C54
```python
query_states = self.q_proj(hidden_states)
key_states = self.k_proj(hidden_states)
value_states = self.v_proj(hidden_states)
```
also the doc of [register_forward_pre_hook](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_forward_pre_hook)
> If `with_kwargs` is true, the forward pre-hook will be passed **the kwargs given to the forward function**.

src/peft/tuners/poly/model.py
```diff
@@ -0,0 +1,151 @@
...
+from .config import PolyConfig
+from .layer import Linear, PolyLayer
+
+
+class PolyModel(BaseTuner):
+    prefix: str = "poly_"
+
+    def __init__(self, model, config, adapter_name) -> None:
+        super().__init__(model, config, adapter_name)
+        self.register_forward_pre_hook(PolyModel._add_task_id_pre_hook, with_kwargs=True)
```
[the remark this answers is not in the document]
TaoSunVoyage: you are right, having the handle alive here could be annoying. it would be better to clean it up after using, maybe with a (post-)forward hook or similar.

src/peft/tuners/poly/model.py
TaoSunVoyage: aha, i managed to make this work as you suggested 😄

using `task_ids` as `kwargs` instead of attribute would indeed be better.

TaoSunVoyage:
Hi @BenjaminBossan, thanks for your prompt reply 😄 

> Could you please run `make style` so that the CI passes?

Done!

> Apart from that, I only have a few smaller comments left. At the end of the day, we should also add docs and tests, but it's fine if they are added in a subsequent PR.

I have fixed the hooks as you suggested. And, we would love to have your assistance in creating proper tests 🙏 

> Oh I see, thanks for clarifying. Is there also an equivalent to 𝙼𝙷𝚁 -z and 𝙼𝙷𝚁 -μ?

Not yet, actually, I was wondering the right way to implement these two methods in `peft` (maybe in a subsequent PR). 

From what I understand, 𝙼𝙷𝚁 -z and 𝙼𝙷𝚁 -μ are meant for a second round of fine-tuning. Imagine you have a model tuned with 𝙼𝙷𝚁, which gives you a unique routing vector for each task and a set of shared LoRAs. When a new task shows up, there are two ways to adjust the model for it:

- 𝙼𝙷𝚁 -z: freeze all the LoRAs and tune a brand new routing vector from scratch.
- 𝙼𝙷𝚁 -μ: tune all the LoRAs using a 1/N routing vector.

src/peft/tuners/poly/model.py
TaoSunVoyage: hi @BenjaminBossan, I want to ensure I’ve understood your suggestion correctly. Are you proposing that we assign one `hook_fn` to each `poly.Linear` within our `PolyModel` like this?

```python
class PolyModel(BaseTuner):
    def __init__(self, model, config, adapter_name) -> None:
        super().__init__(model, config, adapter_name)
        self.hook_handles = dict()
        self._add_hooks()

    def _add_hooks(self):
        for name, module in self.model.named_modules():
            if isinstance(module, PolyLayer) and isinstance(module, nn.Module):
                self.hook_handles[name] = module.register_forward_pre_hook(
                    **hook_fn**, with_kwargs=True)
    ...

def hook_fn(module, args, kwargs):
    # inject task_ids into kwargs, but how?
    return args, kwargs
```

I’m trying to figure out how to implement `hook_fn` such that it can access and inject `task_ids`, which are provided to the model during the `PolyModel.forward` call.

One approach I thought of is defining `hook_fn` during the forward pass of `PolyModel`. However, this seems problematic in a multi-GPU training scenario, as it would result in multiple hooks being added to the same layer, which is what currently happens in the code [here](https://github.com/TaoSunVoyage/peft/blob/1a7dab4ec3e447d1de8d6af1d24b2fd17400e666/src/peft/tuners/poly/model.py#L139)

BenjaminBossan: > However, this seems problematic in a multi-GPU training scenario, as it would result in multiple hooks (one per gpu) being added to the same layer

Good point. Ideally, we could do an atomic "check if hook is present and only add it if not" operation, but I don't think that's possible. Regardless though, I think this should work correctly if we add the hook during `forward`, as that is already device-specific, i.e. after the inputs have been split. I ran a dummy script with DDP and accelerate, adding a forward hook to the child module during the `forward` of the parent module, and it works:

```python
class ChildModule(nn.Module):
    def __init__(self):
        super().__init__()
        self.lin0 = nn.Linear(10, 20)
        self.relu = nn.ReLU()
        self.lin1 = nn.Linear(20, 2)

    def forward(self, X, task_ids=None):
        idx = accelerator.local_process_index
        msg = f"{idx}: {task_ids[0]}"
        print(msg)
        X = X.float()
        X = self.lin0(X)
        X = self.relu(X)
        X = self.lin1(X)
        return X

class ParentModule(nn.Module):
    def __init__(self):
        super().__init__()
        self.mod = ChildModule()

    def forward(self, X, **kwargs):
        def pre_hook(module, args, kwargs):
            X, *_ = args
            task_ids = -100 * X
            kwargs["task_ids"] = task_ids
            return (X,), kwargs

        handle = self.mod.register_forward_pre_hook(pre_hook, with_kwargs=True)

        try:
            result = self.mod(X, **kwargs)
        finally:
            handle.remove()
        return result
```

In the child's `forward`, I print the `task_ids` and they are distinct for each process index, so I think there is no danger of different processes overriding one another's hooks.

BenjaminBossan (commented):
Thanks for the updates, we're getting close to finishing the PR. I added a comment about the hook situation, please take a look.

> And, we would love to have your assistance in creating proper tests

I haven't thought deeply about this, but it's not going to be as easy as some of the other methods, such as LoRA, given that we need the different tasks. The main purpose of the test should be: Supposing that the current code is correct, how can we ensure that future changes to PEFT do not accidentally break poly?

With that in mind, it might be a good idea to use the provided notebook as a starting point for a functional test that trains a simple model on a few dummy tasks and does some sanity checks at the end (say, loss decreased, weights updated, this type of thing). WDYT?

> From what I understand, 𝙼𝙷𝚁 -z and 𝙼𝙷𝚁 -μ are meant for a second round of fine-tuning.

I think it's totally fine to leave those out for now, maybe a comment could be added to that effect.

BenjaminBossan:
@TaoSunVoyage LMK once the PR is ready for another review round.

TaoSunVoyage:
Hi @BenjaminBossan, i think the PR is ready for review.

src/peft/tuners/poly/model.py
```diff
@@ -0,0 +1,173 @@
...
+
+        def pre_hook(_, args, kwargs):
+            kwargs["task_ids"] = task_ids
+            return args, kwargs
+
+        handles = []
+
+        for module in self.model.modules():
+            if isinstance(module, Linear):
+                handle = module.register_forward_pre_hook(pre_hook)
```
BenjaminBossan, line 152:
```suggestion
                handle = module.register_forward_pre_hook(pre_hook, with_kwargs=True)
```

src/peft/tuners/poly/router.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+EPS = 1e-12
+
+
+def get_router(poly_config: PolyConfig) -> nn.Module:
+    if poly_config.poly_type == "poly":
+        return PolyRouter(poly_config)
+    else:
+        raise NotImplementedError
```
BenjaminBossan, line 17: Can we raise a `ValueError` with a helpful error message here?

BenjaminBossan (changes requested):
Thanks a lot, this is very close to being finished now from my point of view.

I found a small error in the handling for the hooks. With my suggested change, it should be fixed. Also, could you please add the license notice to the top of each file, same as for the other files in PEFT?

Thanks also for preparing the notebook. I noticed some small areas for improvement, maybe you could adjust those:

- Upload the notebook including the outputs (not the cleared cells) - that way, a user can get an idea of this method on GitHub and doesn't need to run the notebook first.
- If you make that change, I think you can remove the lines containing the model repr (4th cell, 2nd to last cell), as it's quite big and not super important.
- You set `safe_serialization=False`, is there a need for that? I could safe and load with `safe_serialization=True`.
- There are some double comments, like `# # of tasks`.

I also took a stab at adding at least a simple functional test for poly. This is what I came up with:

```python
import os
import unittest

import torch
from transformers import AutoModelForSeq2SeqLM, AutoTokenizer
from peft import PolyConfig, get_peft_model, TaskType, PeftModel

class TestPoly(unittest.TestCase):
    def test_poly(self):
        torch.manual_seed(0)
        model_name_or_path = "google/flan-t5-small"

        atol, rtol = 1e-6, 1e-6
        r = 8  # rank of lora in poly
        n_tasks = 3  # # of tasks
        n_skills = 2  # # of skills (loras)
        n_splits = 4  # # of heads
        lr = 1e-2
        num_epochs = 10

        tokenizer = AutoTokenizer.from_pretrained(model_name_or_path)
        base_model = AutoModelForSeq2SeqLM.from_pretrained(model_name_or_path)

        peft_config = PolyConfig(
            task_type=TaskType.SEQ_2_SEQ_LM,
            poly_type="poly",
            r=r,
            n_tasks=n_tasks,
            n_skills=n_skills,
            n_splits=n_splits,
        )

        model = get_peft_model(base_model, peft_config)

        # generate some dummy data
        text = os.__doc__.splitlines()
        self.assertTrue(len(text) > 10)
        inputs = tokenizer(text, return_tensors="pt", padding=True)
        inputs["task_ids"] = torch.arange(len(text)) % n_tasks
        inputs["labels"] = tokenizer((["A", "B"] * 100)[:len(text)], return_tensors="pt")["input_ids"]

        # simple training loop
        model.train()
        optimizer = torch.optim.Adam(model.parameters(), lr=lr)
        losses = []
        for _ in range(num_epochs):
            outputs = model(**inputs)
            loss = outputs.loss
            loss.backward()
            optimizer.step()
            optimizer.zero_grad()
            losses.append(loss.item())

        # loss improved by at least 50%
        self.assertLess(losses[-1], 0.5 * losses[0])

        # check that saving and loading works
        torch.manual_seed(0)
        model.eval()
        logits_before = model(**inputs).logits
        tokens_before = model.generate(**inputs)

        with model.disable_adapter():
            logits_disabled = model(**inputs).logits
            tokens_disabled = model.generate(**inputs)

        self.assertFalse(torch.allclose(logits_before, logits_disabled, atol=atol, rtol=rtol))
        self.assertFalse(torch.allclose(tokens_before, tokens_disabled, atol=atol, rtol=rtol))

        # saving and loading
            model.save_pretrained(tmp_dir)
            base_model = AutoModelForSeq2SeqLM.from_pretrained(model_name_or_path)
            loaded = PeftModel.from_pretrained(base_model, tmp_dir)

        torch.manual_seed(0)
        output_after = loaded(**inputs).logits
        tokens_after = loaded.generate(**inputs)
        self.assertTrue(torch.allclose(logits_before, output_after, atol=atol, rtol=rtol))
        self.assertTrue(torch.allclose(tokens_before, tokens_after, atol=atol, rtol=rtol))
```

As you can see, it's super basic, but it covers the majority of added lines:

```
src/peft/tuners/poly/layer.py                            93     12    87%   35, 42, 76, 95-100, 129, 156-157
src/peft/tuners/poly/model.py                           109     24    78%   39, 48, 58, 61-63, 66-70, 86, 93, 106-112, 126-128, 133, 142
src/peft/tuners/poly/router.py                           40      5    88%   17, 23, 27, 48, 50
```

Probably you have some ideas on how to expand the tests to be more precise. But regardless, WDYT about adding something like this?

src/peft/tuners/poly/model.py
TaoSunVoyage: oh, thanks for this. I thought i've fixed this line

src/peft/tuners/poly/router.py
TaoSunVoyage: sure, that would be helpful

TaoSunVoyage:
Hi @BenjaminBossan, thanks a lot for your help!

> I found a small error in the handling for the hooks. With my suggested change, it should be fixed.

Indeed, I missed pushing the fix for the hooks; apologies for the oversight. It's sorted out now.

> Also, could you please add the license notice to the top of each file, same as for the other files in PEFT?

The license notice has been added. Thanks for the reminder.

> * Upload the notebook including the outputs (not the cleared cells) - that way, a user can get an idea of this method on GitHub and doesn't need to run the notebook first.
> * If you make that change, I think you can remove the lines containing the model repr (4th cell, 2nd to last cell), as it's quite big and not super important.

I've uploaded the notebook with outputs included, and removed the specified lines to streamline the content.

> * You set `safe_serialization=False`, is there a need for that? I could safe and load with `safe_serialization=True`.

Regarding `safe_serialization`, I had no particular reason for setting it to `False`. I wasn't entirely up to speed on the feature, but it does appear to be beneficial. I'm glad to switch it to True.

> * There are some double comments, like `# # of tasks`.

I've cleaned up the double comments.

> Probably you have some ideas on how to expand the tests to be more precise. But regardless, WDYT about adding something like this?

I agree, adding the basic test you've proposed is a good idea. It provides a clear and efficient way to verify the main functionality of our code. At the moment, I don't have additional suggestions for more precise tests, as it seems to cover the crucial aspects of our code well.

BenjaminBossan (approved):
There is nothing more to add from my side, thanks for the fantastic work.

@pacman100 @younesbelkada could one of you (or both) also give this a review please?

TaoSunVoyage:
@BenjaminBossan Really appreciate your review!

I'll await @pacman100 and @younesbelkada 's feedback post-holiday season.

Wishing everyone a joyful break!

BenjaminBossan:
Sorry for the delay. We plan on having another review ready by the end of the week. It just takes a bit of time for new methods to be reviewed, since there are many additions at once.

pacman100 (approved):
Wow! Thank you @TaoSunVoyage for the efforts put into supporting Poly! Briefly went over the 2 papers and learnt a lot! The implementation is super clean, super nice example combining superglue tasks and nice tests! Thank you @BenjaminBossan for driving the review of the PR! 🔥🚀✨

BenjaminBossan:
Thanks so much @TaoSunVoyage for this excellent PR and your patience.

TaoSunVoyage:
Thanks @BenjaminBossan @pacman100 ! It feels really great to contribute 🚀

TaoSunVoyage:
> Thanks for the contribution. One quick question, what I we just want to do 1 task with MHR? @TaoSunVoyage
> 
> I tried this and got errors: .../site-packages/peft/tuners/poly/router.py", line 66, in forward raise ValueError("task_ids should not be None.")
> 
> Seems it requires task_id and you cannot set n_task to be 1 and just use MHR?

Hi @hohoCode, apology for the late reply. i think you have to set `task_id=0` when `n_task=1` in current version. 

hope it helps.

A11en0:
Maybe a bug when I using poly on Mistral-7B. The error information is:

```python
 File "/home/yhge/miniconda3/envs/dpmoe/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1532, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/xxx/miniconda3/envs/dpmoe/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1541, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/xxx/CTA/DP-MoE/src/peft/src/peft/tuners/poly/layer.py", line 152, in forward
    mixing_weights = poly_router(task_ids=task_ids, input_ids=x)
  File "/home/xxx/miniconda3/envs/dpmoe/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1532, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/xxx/miniconda3/envs/dpmoe/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1541, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/yhge/CTA/DP-MoE/src/peft/src/peft/tuners/poly/router.py", line 66, in forward
    raise ValueError("task_ids should not be None.")
ValueError: task_ids should not be None.
```

this error is encounter in the backward stage.

TaoSunVoyage:
Hi @A11en0, `task_ids` should be passed to the model explicitly (same as `input_ids`, ...) otherwise this error would be raised.

PR #1131, opened 2023-11-14
TST: Improve requires grad testing:

BenjaminBossan:
## Description

Previously, the corresponding tests were testing only whether specific parameters had `requires_grad` `True` or `False`. Now, all parameters are being checked. This is more rigorous.

Also, tests for `Embedding`, `Conv1D`, `Conv2d` were added (LoRA and IA³), thus superseding PR #1115.

Finally, tests for LoHa and LoKr were added.

## Note

I considered moving the tests to a separate module, as they were getting quite big and this would help with readability. For now, I left them in the same module because it leads to a better diff view and is thus easier to review. LMK if I should move the tests to a separate file.

### DIFF
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -941,14 +941,25 @@ class RequiresGradTester(unittest.TestCase):
 
     """
 
+    def check_requires_grad(self, model, *params_expected: str):
+        # Check that only the given parameters have requires_grad=True, and all others have requires_grad=False.
+        # Calling without arguments besides the model means that all parameters should have requires_grad=False.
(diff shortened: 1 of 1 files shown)


pacman100 (approved):
Thank you @BenjaminBossan for the extensive tests to check that gradients are being set properly!

tests/test_custom_models.py
```diff
@@ -941,45 +941,64 @@ class RequiresGradTester(unittest.TestCase):
 
     """
 
+    def check_requires_grad(self, model, *params_expected: str):
```
pacman100, line 944: Neat!

PR #1132, opened 2023-11-14
FEAT: Merging only specified `adapter_names` when calling `merge`

younesbelkada:
# What does this PR do?

Partially addresses: https://github.com/huggingface/diffusers/issues/5320 

This PR exposes a new argument in `merge` termed as `adapter_names`, if a user passes an explicit `adapter_names` then only adapter names inside that array will be merged. This PR will make things smoother and easier to integrate on diffusers side

Added also a nice CI test

cc @pacman100 @BenjaminBossan

### DIFF
--- a/tests/test_decoder_models.py
+++ b/tests/test_decoder_models.py
@@ -167,6 +167,19 @@ class PeftDecoderModelTester(unittest.TestCase, PeftCommonTester):
             },
         )
     )
+    def test_merge_layers_multi(self, test_name, model_id, config_cls, config_kwargs):
+        self._test_merge_layers_multi(model_id, config_cls, config_kwargs)
+
+    @parameterized.expand(
+        PeftTestConfigManager.get_grid_parameters(
+            {
+                "model_ids": PEFT_DECODER_MODELS_TO_TEST,
+                "lora_kwargs": {"init_lora_weights": [False]},
+                "ia3_kwargs": {"init_ia3_weights": [False]},
+                "task_type": "CAUSAL_LM",
+            },
+        )
+    )
     def test_merge_layers_nan(self, test_name, model_id, config_cls, config_kwargs):
         self._test_merge_layers_nan(model_id, config_cls, config_kwargs)
 
--- a/tests/testing_common.py
+++ b/tests/testing_common.py
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import copy
 import json
 import os
 import pickle
@@ -560,6 +561,48 @@ class PeftCommonTester:
 
         logits_merged_from_pretrained = model_from_pretrained(**dummy_input)[0]
         self.assertTrue(torch.allclose(logits_merged, logits_merged_from_pretrained, atol=atol, rtol=rtol))
+
+    def _test_merge_layers_multi(self, model_id, config_cls, config_kwargs):
+        if config_cls not in (LoraConfig, IA3Config):
+            # Merge layers only supported for LoRA and IA³
+            return
+        if ("gpt2" in model_id.lower()) and (config_cls != LoraConfig):
+            self.skipTest("Merging GPT2 adapters not supported for IA³ (yet)")
+
+        model = self.transformers_class.from_pretrained(model_id)
+        config = config_cls(
+            base_model_name_or_path=model_id,
+            **config_kwargs,
+        )
+        model = get_peft_model(model, config)
+        model.add_adapter("adapter-2", config)
+
+        model = model.to(self.torch_device)
+
+        model_copy = copy.deepcopy(model)
+
+        if config.peft_type not in ("IA3", "LORA"):
+            with self.assertRaises(AttributeError):
+                model = model.merge_and_unload()
+
+        dummy_input = self.prepare_inputs_for_testing()
+        model.eval()
+
+        with torch.no_grad():
+            logits_all_adapters = model(**dummy_input)[0]
+
+        model_merged_1 = model.merge_and_unload(adapter_names=["adapter-2"])
+
+        with torch.no_grad():
+            logits_merged_1 = model_merged_1(**dummy_input)[0]
+
+        model_fully_merged = model_copy.merge_and_unload()
+
+        with torch.no_grad():
+            logits_fully_merged = model_fully_merged(**dummy_input)[0]
+
+        self.assertTrue(torch.allclose(logits_all_adapters, logits_fully_merged, atol=1e-4, rtol=1e-4))
+        self.assertFalse(torch.allclose(logits_all_adapters, logits_merged_1, atol=1e-4, rtol=1e-4))
 
     def _test_generate(self, model_id, config_cls, config_kwargs):
         model = self.transformers_class.from_pretrained(model_id)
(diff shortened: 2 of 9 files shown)


tests/testing_common.py
```diff
@@ -561,6 +562,48 @@ def _test_merge_layers(self, model_id, config_cls, config_kwargs):
...
+            **config_kwargs,
+        )
+        model = get_peft_model(model, config)
+        model.add_adapter("adapter-2", config)
+
+        model = model.to(self.torch_device)
+
+        model_copy = copy.deepcopy(model)
+
+        if config.peft_type not in ("IA3", "LORA"):
```
BenjaminBossan, line 584: Because of the first check in the test, this will never be called, right?

tests/testing_common.py
```diff
@@ -561,6 +562,48 @@ def _test_merge_layers(self, model_id, config_cls, config_kwargs):
         logits_merged_from_pretrained = model_from_pretrained(**dummy_input)[0]
         self.assertTrue(torch.allclose(logits_merged, logits_merged_from_pretrained, atol=atol, rtol=rtol))
 
+    def _test_merge_layers_multi(self, model_id, config_cls, config_kwargs):
```
BenjaminBossan, line 565: The test was added but is not called in `test_decoder_models.py` etc.

tests/testing_common.py
```diff
@@ -561,6 +562,48 @@ def _test_merge_layers(self, model_id, config_cls, config_kwargs):
...
+
+        if config.peft_type not in ("IA3", "LORA"):
+            with self.assertRaises(AttributeError):
+                model = model.merge_and_unload()
+
+        dummy_input = self.prepare_inputs_for_testing()
+        model.eval()
+
+        with torch.no_grad():
+            logits_all_adapters = model(**dummy_input)[0]
```
BenjaminBossan, line 592: Is this really from all adapters? `"adapter-2"` is not active, is it?

tests/testing_common.py
```diff
@@ -561,6 +562,48 @@ def _test_merge_layers(self, model_id, config_cls, config_kwargs):
...
+        with torch.no_grad():
+            logits_merged_1 = model_merged_1(**dummy_input)[0]
+
+        model_fully_merged = model_copy.merge_and_unload()
+
+        with torch.no_grad():
+            logits_fully_merged = model_fully_merged(**dummy_input)[0]
+
+        self.assertTrue(torch.allclose(logits_all_adapters, logits_fully_merged, atol=1e-4, rtol=1e-4))
+        self.assertFalse(torch.allclose(logits_all_adapters, logits_merged_1, atol=1e-4, rtol=1e-4))
```
BenjaminBossan, line 605: I'm not sure if the test is rigorous enough. Just because `model_merged_1` is different from the others is not sufficient to prove that adapter2 was merged. How about this process:

1. create PEFT model with adapter1 (default)
2. generate logits1 with adapter1
3. add adapter2
4. activate adapter2 (without merge)
5. generate logits2 with adapter2
6. check that logits1 are different from logits2
7. activate adapter1 again
8. check that logits are again like logits1
9. `merge_and_unload` adapter2
10. the unloaded model should produce logits like logits2

Furthermore, it would be good to have a similar test, only that at step 9, `merge_and_unload` is called on both adapter1 and adapter2 (so one active and one inactive adapter).

BenjaminBossan (changes requested):
Thanks for adding this feature, it should be very useful. Implementation-wise, it looks good, but I have a few remarks:

I noticed that only IA³ and (Ada)LoRA were changed, but the same logic should also work for LoHa and LoKr. Could you please update those as well?

Regarding the test, I think it's not being called right now, I also made a suggestion on how I think it could be improved.

tests/testing_common.py
younesbelkada: AH yes nice catch, fixed it!

tests/testing_common.py
younesbelkada: True, removed it!

tests/testing_common.py
younesbelkada: True, this is only on the active adapter and not all adapters

tests/testing_common.py
```diff
@@ -561,6 +561,54 @@ def _test_merge_layers(self, model_id, config_cls, config_kwargs):
         logits_merged_from_pretrained = model_from_pretrained(**dummy_input)[0]
         self.assertTrue(torch.allclose(logits_merged, logits_merged_from_pretrained, atol=atol, rtol=rtol))
 
+    def _test_merge_layers_multi(self, model_id, config_cls, config_kwargs):
+        supported_peft_types = [PeftType.LORA, PeftType.LOHA, PeftType.LOKR, PeftType.IA3]
+
+        if ("gpt2" in model_id.lower()) and (config_cls != LoraConfig):
```
pacman100, line 567:
```suggestion
        if ("gpt2" in model_id.lower()) and (config_cls == IA3Config):
```

pacman100 (approved):
Thank you @younesbelkada for extending the functionality of the `merge` util.

younesbelkada:
The tests should be fixed now thanks to @pacman100 's suggestions! the PR is ready for another round of review

tests/testing_common.py
```diff
@@ -561,6 +562,63 @@ def _test_merge_layers(self, model_id, config_cls, config_kwargs):
...
+
+        model = self.transformers_class.from_pretrained(model_id)
+        config = config_cls(
+            base_model_name_or_path=model_id,
+            **config_kwargs,
+        )
+
+        if config.peft_type not in supported_peft_types:
+            return
+
```
BenjaminBossan, line 579: Let's load the model here, instead of l.571, because we don't need to load it when we skip anyway.

tests/testing_common.py
```diff
@@ -561,6 +562,63 @@ def _test_merge_layers(self, model_id, config_cls, config_kwargs):
...
+        self.assertFalse(torch.allclose(logits_merged_all, logits_adapter_2, atol=1e-3, rtol=1e-3))
+        self.assertFalse(torch.allclose(logits_merged_all, logits_adapter_1, atol=1e-3, rtol=1e-3))
+
+        model_merged_adapter_2 = model_copy.merge_and_unload(adapter_names=["adapter-2"])
+
+        with torch.inference_mode():
+            logits_merged_adapter_2 = model_merged_adapter_2(**dummy_input)[0]
+
+        self.assertTrue(torch.allclose(logits_merged_adapter_2, logits_adapter_2, atol=1e-3, rtol=1e-3))
```
BenjaminBossan, line 620: How about also adding a test for `merge_and_unload(adapter_names=["default"])`? Should just be a few more lines for a little bit of extra safety.

tests/test_decoder_models.py
```diff
@@ -157,6 +157,19 @@ def test_from_pretrained_config_construction(self, test_name, model_id, config_c
...
+        PeftTestConfigManager.get_grid_parameters(
+            {
+                "model_ids": PEFT_DECODER_MODELS_TO_TEST,
+                "lora_kwargs": {"init_lora_weights": [False]},
+                "ia3_kwargs": {"init_ia3_weights": [False]},
+                "task_type": "CAUSAL_LM",
+            },
+        )
+    )
+    def test_merge_layers_multi(self, test_name, model_id, config_cls, config_kwargs):
```
BenjaminBossan, line 170: I don't think we need these tests for all model types besides decoder. But maybe let's also add them to the custom model tests, as we check a broader range of adapters there.

BenjaminBossan (changes requested):
Thanks @younesbelkada for adding this feature and @pacman100 for figuring out the missing puzzle piece. I have some comments about testing, besides that it LGTM.

BenjaminBossan (approved):
Thanks a lot for extending the test coverage, LGTM!

PR #1136, opened 2023-11-16
[`Tests`] Fix daily CI

younesbelkada:
# What does this PR do?

With the recent changes of making safetensors default saving format, we need to update the slow tests accordingly

cc @BenjaminBossan @pacman100


BenjaminBossan (approved):
Thanks, looks good. Just a suggestion: We could also use `SAFETENSOR_WEIGHTS_NAME`.

PR #1137, opened 2023-11-16
Raise error when `modules_to_save` is specified and multiple adapters are being unloaded

pacman100:
### What does this PR do?
1. Raise error when `modules_to_save` is specified and multiple adapters are being unloaded


src/peft/tuners/lora/model.py
```diff
@@ -377,6 +377,15 @@ def _unload_and_optionally_merge(
             if getattr(self.model, "quantization_method", None) == "gptq":
                 raise ValueError("Cannot merge LORA layers when the model is gptq quantized")
 
+        adapters_to_consider = adapter_names or self.active_adapters
+        is_modules_to_save_available = any(
+            self.peft_config[adapter].modules_to_save is not None for adapter in adapters_to_consider
+        )
+        if is_modules_to_save_available and len(adapters_to_consider) > 1:
```
BenjaminBossan, line 384: If `adapter_names` is a str, this will trigger an error even though it would be fine, right?

BenjaminBossan (changes requested):
Thanks for stepping up and addressing this issue so quickly.

The same change is needed for `ia3/model.py` and `lycoris_utils.py`. Maybe it can be factored out into a utility function?

For general context: This is an oversight from #1132, which was discussed internally. Since the PR is very recent, this change should not affect any users.

src/peft/tuners/lora/model.py
pacman100: `adapter_names` is a `List[str]`, right?

src/peft/tuners/tuners_utils.py
```diff
@@ -299,6 +299,16 @@ def unmerge_adapter(self):
             if isinstance(module, BaseTunerLayer):
                 module.unmerge()
 
+    def _unloading_checks(self, adapter_names: List[str]):
```
BenjaminBossan, line 302:
```suggestion
    def _unloading_checks(self, adapter_names: Optional[List[str]]):
```

src/peft/tuners/tuners_utils.py
```diff
@@ -299,6 +299,16 @@ def unmerge_adapter(self):
...
+        adapters_to_consider = adapter_names or self.active_adapters
+        is_modules_to_save_available = any(
+            self.peft_config[adapter].modules_to_save is not None for adapter in adapters_to_consider
+        )
+        if is_modules_to_save_available and len(adapters_to_consider) > 1:
+            raise ValueError(
+                "models having `modules_to_save` specified in LoraConfig support only single adapter for unloading."
```
BenjaminBossan, line 309: I tried to come up with an error message that is a bit clearer, WDYT?

```suggestion
                "Cannot unload multiple adapters that specify `modules_to_save`."
```

src/peft/tuners/tuners_utils.py
```diff
@@ -299,6 +299,16 @@ def unmerge_adapter(self):
             if isinstance(module, BaseTunerLayer):
                 module.unmerge()
 
+    def _unloading_checks(self, adapter_names: List[str]):
+        adapters_to_consider = adapter_names or self.active_adapters
+        is_modules_to_save_available = any(
+            self.peft_config[adapter].modules_to_save is not None for adapter in adapters_to_consider
```
BenjaminBossan, line 305:
```suggestion
            self.peft_config[adapter].modules_to_save for adapter in adapters_to_consider
```

How about this change? This way, if `modules_to_save = []`, it is not counted here.

BenjaminBossan (commented):
Thanks for making the adjustments. I only have some small comments, the rest looks good.

BenjaminBossan (approved):
Great, thanks! From my point of view, can be merged as soon as CI has passed.

Issue #1138, opened 2023-11-16
Linear4bit.merge() got an unexpected keyword argument 'adapter_names'


khiempm209:
### System Info

I got this error when i use the 'merge_and_unload()' to merge my own trained lora apdapter to the Mistral model. I thought you guys forgot to put the 'adapter_names' argument to the merge function in Linear4bit class.

File /opt/conda/lib/python3.10/site-packages/peft/tuners/lora/model.py:390, in LoraModel._unload_and_optionally_merge(self, merge, progressbar, safe_merge, adapter_names)
    388 if hasattr(target, "base_layer"):
    389     if merge:
--> 390         target.merge(safe_merge=safe_merge, adapter_names=adapter_names)
    391     self._replace_module(parent, target_name, target.get_base_layer(), target)
    392 elif isinstance(target, ModulesToSaveWrapper):
    393     # save any additional trainable modules part of `modules_to_save`

TypeError: Linear4bit.merge() got an unexpected keyword argument 'adapter_names'

### Who can help?

@pacman100 @younesbelkada @say

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [X] My own task or dataset (give details below)

### Reproduction

vi_model_final = PeftModel.from_pretrained(vi_model_merge, model_name)
vi_model_final = vi_model_final.merge_and_unload()
vi_model_final.eval()

----> 2 vi_model_final = vi_model_final.merge_and_unload()
      3 vi_model_final.eval()

File /opt/conda/lib/python3.10/site-packages/peft/tuners/lora/model.py:658, in LoraModel.merge_and_unload(self, progressbar, safe_merge, adapter_names)
    630 def merge_and_unload(
    631     self, progressbar: bool = False, safe_merge: bool = False, adapter_names: Optional[List[str]] = None
    632 ):
    633     r"""
    634     This method merges the LoRa layers into the base model. This is needed if someone wants to use the base model
    635     as a standalone model.
   (...)
    656     ```
    657     """
--> 658     return self._unload_and_optionally_merge(
    659         progressbar=progressbar, safe_merge=safe_merge, adapter_names=adapter_names
    660     )

File /opt/conda/lib/python3.10/site-packages/peft/tuners/lora/model.py:390, in LoraModel._unload_and_optionally_merge(self, merge, progressbar, safe_merge, adapter_names)
    388 if hasattr(target, "base_layer"):
    389     if merge:
[...]

PR #1139, opened 2023-11-16
[`core` / `LoRA`] Add `adapter_names` in bnb layers

younesbelkada:
# What does this PR do?

Fixes https://github.com/huggingface/peft/issues/1138

cc @BenjaminBossan @pacman100


BenjaminBossan (approved):
Thanks for fixing this quickly. Since only LoRA bnb layers support merging for now, this should be sufficient.

PR #1143, opened 2023-11-17
CI Add Python 3.11 to test matrix

BenjaminBossan:


younesbelkada (approved):
Makes sense, thanks @BenjaminBossan !

PR #1146, opened 2023-11-17
FIX: A few issues with AdaLora, extending GPU tests

### LINKED ISSUE #1113: AdaLora + bnb not working
### System Info

-

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

The issue is this line:

https://github.com/huggingface/peft/blob/49ddefa83443052ca9c0ee56d99476f16f375bdc/src/peft/tuners/adalora/bnb.py#L144

In AdaLoRA, `lora_A` and `lora_B` are not `ModuleDict`s but `ParameterDict`s, so `lora_A[adapter_name].weight.dtype` does not exist, it should just be `lora_A[adapter_name].dtype`.

Furthermore, using AdaLoRA with 8bit bnb gives NaNs for me for opt-125m.

### Expected behavior

AdaLoRA + bnb should work.

BenjaminBossan:
This PR fixes a handful of issues with AdaLora, should resolve #1113.

## Description

1. `lora_A.weight.device` was called but for AdaLora, `lora_A` is an `nn.Paramter`, not an `nn.Module`, so the weight attribute does not exist. `lora_A.device` is sufficient.
2. For 8bit, an inplace operation failed because it was on a view. Now the operation is no longer inplace.
3. The loss term of the model output is not necessarily a torch tensor. In the test, it was a dict and did not contain an actual loss. Therefore, I added a check to make sure the loss is a torch tensor. Is there a better way?

## Notes

Running `pytest tests/test_gpu_examples.py -k adalora` locally (with GPU) passes. Ideally, someone else can confirm, as normal unit tests won't catch this.

If this is merged before #1115, skipping AdaLora tests in that PR can be removed.

### DIFF
--- a/src/peft/tuners/adalora/bnb.py
+++ b/src/peft/tuners/adalora/bnb.py
@@ -70,7 +70,8 @@ if is_bnb_available():
                 if requires_conversion:
                     output = output.to(expected_dtype)
                 output = output * scaling / ranknum
-                result += output
+                # inplace operation on view is forbidden for MatMul8bitLtBackward, so avoid it
+                result = result + output
             return result
 
         def __repr__(self) -> str:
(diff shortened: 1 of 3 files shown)


younesbelkada (approved):
Amazing clean up, thanks @BenjaminBossan !

PR #1152, opened 2023-11-20
DOC Update a few places in the README

### LINKED ISSUE #1151: Mention that Mistral is supported as well
In https://huggingface.co/docs/peft/index and the README, we list supported models. I think Mistral is supported as well (and probably few others)

BenjaminBossan:
Resolves #1151

- fix bits_and_bytes => bitsandbytes
- add a few links
- add mistral to list of supported models

A question: What is the criterion for a model to be considered to be "supported" for a given adapter? E.g. "ChatGLM" is listed as having support for all methods, but we only have an entry for it in `TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING`, not in `TRANSFORMERS_MODELS_TO_IA3_TARGET_MODULES_MAPPING` etc. So should we really put a ✅ in all those columns? For mistral, I only added a ✅ for LoRA, as that's the only entry for mistral that we have, even though I'm pretty sure we can just use the same settings for other adapters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,7 +228,7 @@ An example is provided in [this notebook](https://github.com/huggingface/peft/bl
 
 ## Models support matrix
 
-Find models that are supported out of the box below. Note that PEFT works with almost all models -- it it is not listed, you just need to [do some manual configuration](https://huggingface.co/docs/peft/developer_guides/custom_models).
+Find models that are supported out of the box below. Note that PEFT works with almost all models -- if it is not listed, you just need to [do some manual configuration](https://huggingface.co/docs/peft/developer_guides/custom_models).
 
 ### Causal Language Modeling
 | Model        | LoRA | Prefix Tuning  | P-Tuning | Prompt Tuning  | IA3 |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -223,11 +223,13 @@ DeepSpeed version required `v0.8.0`. An example is provided in `~examples/condit
   ```
 
 ### Example of PEFT model inference using 🤗 Accelerate's Big Model Inferencing capabilities
-An example is provided in `~examples/causal_language_modeling/peft_lora_clm_accelerate_big_model_inference.ipynb`. 
+An example is provided in [this notebook](https://github.com/huggingface/peft/blob/main/examples/causal_language_modeling/peft_lora_clm_accelerate_big_model_inference.ipynb).
 
 
 ## Models support matrix
 
+Find models that are supported out of the box below. Note that PEFT works with almost all models -- it it is not listed, you just need to [do some manual configuration](https://huggingface.co/docs/peft/developer_guides/custom_models).
```
stevhliu, line 231:
```suggestion
Find models that are supported out of the box below. Note that PEFT works with almost all models -- if it is not listed, you just need to [do some manual configuration](https://huggingface.co/docs/peft/developer_guides/custom_models).
```

stevhliu (commented):
just a small typo i noticed :)

BenjaminBossan:
> A question: What is the criterion for a model to be considered to be "supported" for a given adapter?

I added a note about this in our internal docs discussion page, so that we don't forget about it. I'll go ahead and merge this PR then.

PR #1153, opened 2023-11-20
Delete IA3 adapter

alexrs:
### What

As discussed in https://github.com/huggingface/peft/pull/980 (specifically https://github.com/huggingface/peft/pull/980#issuecomment-1815902304), we are breaking that PR into smaller PRs that can be merged faster. 

In this first PR, we include the ability to delete $(IA)^3$ adapters. 

### How
A new method `delete_adapter` has been added to the `IA3Model` class. This method is based on https://github.com/huggingface/peft/blob/f1ecfa6ae6eba599ae89decbf47b339d8c9d39a3/src/peft/tuners/lora/model.py#L608

### Test Plan
- Added `PeftType.IA3` to `testing_common.py#_test_delete_adapter`
- Run the tests and checked that they pass

### DIFF
--- a/src/peft/tuners/ia3/layer.py
+++ b/src/peft/tuners/ia3/layer.py
@@ -27,12 +27,9 @@ from peft.utils import transpose
 class IA3Layer(BaseTunerLayer):
     # All names of layers that may contain adapter weights
     adapter_layer_names = ("ia3_l",)
-    # All names of other parameters that may contain adapter-related parameters
-    other_layer_names = ("scaling",)
 
     def __init__(self, base_layer: nn.Module, is_feedforward: bool, **kwargs) -> None:
         self.base_layer = base_layer
-        self.scaling = {}
         self.ia3_l = nn.ParameterDict({})
         # Mark the weight as unmerged
         self._disable_adapters = False
@@ -294,7 +291,7 @@ class Conv2d(nn.Module, IA3Layer):
                     base_layer.bias.data = torch.mul(base_layer.bias.data, scaling.data)
 
     def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
-        previous_dtype = x.dtype
+        dtype = previous_dtype = x.dtype
 
         if self.disable_adapters:
             if self.merged:
--- a/tests/testing_common.py
+++ b/tests/testing_common.py
@@ -885,7 +885,7 @@ class PeftCommonTester:
             self.assertIsNotNone(param.grad)
 
     def _test_delete_adapter(self, model_id, config_cls, config_kwargs):
-        supported_peft_types = [PeftType.LORA, PeftType.LOHA, PeftType.LOKR]
+        supported_peft_types = [PeftType.LORA, PeftType.LOHA, PeftType.LOKR, PeftType.IA3]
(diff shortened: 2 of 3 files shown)


src/peft/tuners/ia3/layer.py
```diff
@@ -27,12 +27,9 @@
 class IA3Layer(BaseTunerLayer):
     # All names of layers that may contain adapter weights
     adapter_layer_names = ("ia3_l",)
-    # All names of other parameters that may contain adapter-related parameters
-    other_layer_names = ("scaling",)
 
     def __init__(self, base_layer: nn.Module, is_feedforward: bool, **kwargs) -> None:
         self.base_layer = base_layer
-        self.scaling = {}
```
alexrs, line 35: As far as I understand, we do not use `self.scaling` and $(IA)^3$ does not have this parameter.

src/peft/tuners/ia3/layer.py
```diff
@@ -294,7 +291,7 @@ def unmerge(self) -> None:
                     base_layer.bias.data = torch.mul(base_layer.bias.data, scaling.data)
 
     def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
-        previous_dtype = x.dtype
+        dtype = previous_dtype = x.dtype
```
alexrs, line 294: `dtpye` was not defined but it is used in https://github.com/huggingface/peft/blob/f1ecfa6ae6eba599ae89decbf47b339d8c9d39a3/src/peft/tuners/ia3/layer.py#L314

tests/testing_common.py
```diff
@@ -905,7 +905,7 @@ def _test_delete_adapter(self, model_id, config_cls, config_kwargs):
         self.assertFalse(adapter_to_delete in model.peft_config)
         self.assertEqual(model.active_adapters, ["default"])
 
-        key_list = [key for key, _ in model.named_modules() if "lora" not in key]
+        key_list = [key for key, _ in model.named_modules() if model.prefix not in key]
```
BenjaminBossan, line 908: Good amendment, but this made me curious why the code worked at all, because `"lora"` was hard-coded but we also test LoHa and LoKr. It turns out the that prefix check is totally unnecessary and the code can be simplified to:

`key_list = [key for key, _ in model.named_modules()]`

Could you please fix that? Same for `_test_delete_inactive_adapter` and `_test_weighted_combination_of_adapters`. Thanks.

BenjaminBossan (changes requested):
Thank you so much Alejandro for factoring out the deletion-feature and creating this super clean PR. Not only do you show great understanding of the PEFT code base, but also improved some unrelated lines on top, perfect!

I have only a few minor comments, they should be easy to fix and then we're good to merge.

tests/testing_common.py
alexrs: I was also curious about why this worked at all. I'll fix it!

alexrs:
> Not only do you show great understanding of the PEFT code base, but also improved some unrelated lines on top, perfect!

Thanks! I've spent quite a few hours navigating the codebase for my MSc thesis, and it has helped me a lot. Thanks for the good work! 👏

BenjaminBossan (approved):
Fantastic work, thanks a lot!

PR #1159, opened 2023-11-21
add options to save or push model

callanwu:
Add options to either push the model to pub or save it locally, and deprecate 'use_auth_token' that is no longer in use.


BenjaminBossan:
Thanks for updating the examples and notebooks. I see that you put the explanations in a string using `"""`. IMO, the better fit would be to use comments in the Python scripts and Markdown cells in the notebooks. Could you please change that?

callanwu:
ic, I will change it in the notebooks.

callanwu:
@BenjaminBossan hi,
i changed, check it:]

BenjaminBossan (changes requested):
Thanks for updating the PR. I still see a few issues:

- You add `{dataset_name}` to the model name, is there a specific reason for that or is it just your preference?
- Would the `push_to_hub` call work the way you wrote it? There is no repo id in the name. Normally, it should be something like `myrepo/mymodel`, right?
- I also don't think that `push_to_hub` gets a `state_dict` argument.
- "You can push model to hub" > "You can push **the** model to **Hugging Face Hub**"
- "token to use as HTTP bearer authorization for remote files": The sentence is missing a word, not sure what exactly it wants to express
- `peft_prefix_tuning_clm.ipynb` has not been updated to use markdown cells
- The .py files still use a string and not a comment

callanwu:
- The reason for adding `{dataset_name}` is to make it more specific that can recognize more easily.
- When `token` is filled in, the user id is recognized. So `myrepo/mymodel` is not needed.
- I will modify the argument based on https://github.com/huggingface/transformers/blob/3bc50d81e6c70d63e59d635106bac6a561b47681/src/transformers/utils/hub.py#L759

I will improve the PR according to your suggestions:)

callanwu:
I made some changes. Are there any further improvements needed?

BenjaminBossan (approved):
Thanks a lot.

PR #1160, opened 2023-11-21
[Feature] Support OFT

okotaku:
OFT: [Controlling Text-to-Image Diffusion by Orthogonal Finetuning](https://arxiv.org/abs/2306.07280)

### DIFF
--- a/src/peft/tuners/oft/layer.py
+++ b/src/peft/tuners/oft/layer.py
@@ -1,0 +1,393 @@
+# coding=utf-8
+# Copyright 2023-present the HuggingFace Inc. team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import math
+import warnings
+from typing import Any, List, Optional, Set, Tuple
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from peft.tuners.lycoris_utils import LycorisLayer
+
+
+class OFTLayer(nn.Module, LycorisLayer):
+    # All names of layers that may contain adapter weights
+    adapter_layer_names = ("oft_r",)
+    # other_param_names is defined on parent class
+
+    def __init__(self, base_layer: nn.Module):
+        super().__init__()
+        LycorisLayer.__init__(self, base_layer)
+
+        # OFT info
+        self.oft_r = nn.ParameterDict({})
+        self.coft = {}
+        self.eps = {}
+        self.block_share = {}
+
+    @property
+    def _available_adapters(self) -> Set[str]:
+        return {*self.oft_r}
+
+    def create_adapter_parameters(self, adapter_name: str, r: int, shape: Tuple[int, ...], block_share: bool):
+        if block_share:
+            self.oft_r[adapter_name] = nn.Parameter(torch.empty(1, shape[1] // r, shape[1] // r))
+        else:
+            self.oft_r[adapter_name] = nn.Parameter(torch.empty(r, shape[1] // r, shape[1] // r))
+
+    def reset_adapter_parameters(self, adapter_name: str):
+        nn.init.zeros_(self.oft_r[adapter_name])
+
+    def reset_adapter_parameters_random(self, adapter_name: str):
+        nn.init.kaiming_uniform_(self.oft_r[adapter_name], a=math.sqrt(5))
+
+    def update_layer(
+        self,
+        adapter_name: str,
+        r: int,
+        alpha: float,
+        rank_dropout: float,
+        module_dropout: float,
+        init_weights: bool,
+        use_effective_conv2d: bool = False,
+        coft: bool = False,
+        eps: float = 6e-5,
+        block_share: bool = False,
+        **kwargs,
+    ) -> None:
+        """Internal function to create oft adapter
+
+        Args:
+            adapter_name (`str`): Name for the adapter to add.
+            r (`int`): Rank for the added adapter.
+            alpha (`float`): Alpha for the added adapter.
+            rank_dropout (`float`): The dropout probability for rank dimension during training.
+            module_dropout (`float`): The dropout probability for disabling adapter during training.
+            init_weights (`bool`): Whether to initialize weights.
+            use_effective_conv2d (`bool`, *optional*, defaults to `False`):
+                Use parameter effective decomposition for Conv2d with ksize > 1.
+            coft (`bool`): Whether to use the constrainted variant of OFT or not.
+            eps (`float`): The control strength of COFT. The freedom of rotation. Only has an effect if `coft` is set to True.
+            block_share (`bool`): Whether to share the OFT parameters between blocks or not.
+        """
+
+        self.r[adapter_name] = r
+        self.alpha[adapter_name] = alpha
+        self.scaling[adapter_name] = alpha / r
+        self.rank_dropout[adapter_name] = rank_dropout
+        self.module_dropout[adapter_name] = module_dropout
+        self.coft[adapter_name] = coft
+        self.eps[adapter_name] = eps
+        self.block_share[adapter_name] = block_share
+
+        # Determine shape of OFT weights
+        base_layer = self.get_base_layer()
+        if isinstance(base_layer, nn.Linear):
+            shape = tuple(base_layer.weight.shape)
+        elif isinstance(base_layer, nn.Conv2d):
+            use_effective_conv2d = use_effective_conv2d and base_layer.kernel_size != (1, 1)
+            if use_effective_conv2d:
+                shape = (base_layer.out_channels, base_layer.in_channels, *base_layer.kernel_size)
+            else:
+                shape = (
+                    base_layer.out_channels,
+                    base_layer.in_channels * base_layer.kernel_size[0] * base_layer.kernel_size[1],
+                )
+        else:
+            raise TypeError(f"OFT is not implemented for base layers of type {type(base_layer).__name__}")
+
+        # Create weights with provided shape
+        self.create_adapter_parameters(adapter_name, r, shape, block_share)
+
+        # Initialize weights
+        if init_weights:
+            self.reset_adapter_parameters(adapter_name)
+        else:
+            self.reset_adapter_parameters_random(adapter_name)
+
+        # Move new weights to device
+        weight = getattr(self.get_base_layer(), "weight", None)
+        if weight is not None:
+            # the layer is already completely initialized, this is an update
+            if weight.dtype.is_floating_point or weight.dtype.is_complex:
+                self.to(weight.device, dtype=weight.dtype)
+            else:
+                self.to(weight.device)
+        self.set_adapter(self.active_adapters)
+
+    def merge(self, safe_merge: bool = False, adapter_names: Optional[List[str]] = None) -> None:
+        if self.merged:
+            warnings.warn(
+                f"Already following adapters were merged {','.join(self.merged_adapters)}. "
+                f"You are now additionally merging {','.join(self.active_adapters)}."
+            )
+        if adapter_names is None:
+            adapter_names = self.active_adapters
+
+        for active_adapter in adapter_names:
+            if active_adapter in self._available_adapters:
+                base_layer = self.get_base_layer()
+
+                orig_weights = base_layer.weight.data
+                if isinstance(base_layer, nn.Linear):
+                    orig_weights = torch.transpose(orig_weights, 0, 1)
+                elif isinstance(base_layer, nn.Conv2d):
+                    orig_weights = orig_weights.view(
+                        [
+                            base_layer.kernel_size[0] * base_layer.kernel_size[1] * base_layer.in_channels,
+                            base_layer.out_channels,
+                        ]
+                    )
+                orig_weights = torch.mm(self.get_delta_weight(active_adapter), orig_weights)
+                if isinstance(base_layer, nn.Linear):
+                    orig_weights = torch.transpose(orig_weights, 0, 1)
+                elif isinstance(base_layer, nn.Conv2d):
+                    orig_weights = orig_weights.view(
+                        [
+                            base_layer.out_channels,
+                            base_layer.in_channels,
+                            base_layer.kernel_size[0],
+                            base_layer.kernel_size[1],
+                        ]
+                    )
+
+                if not torch.isfinite(orig_weights).all():
+                    raise ValueError(
+                        f"NaNs detected in the merged weights. The adapter {active_adapter} seems to be broken"
+                    )
+
+                base_layer.weight.data = orig_weights
+                self.merged_adapters.append(active_adapter)
+
+    def unmerge(self) -> None:
+        if not self.merged:
+            warnings.warn("Already unmerged. Nothing to do.")
+            return
+        while len(self.merged_adapters) > 0:
+            active_adapter = self.merged_adapters.pop()
+            if active_adapter in self._available_adapters:
+                base_layer = self.base_layer
+                orig_weights = base_layer.weight.data
+                if isinstance(base_layer, nn.Linear):
+                    orig_weights = torch.transpose(orig_weights, 0, 1)
+                elif isinstance(base_layer, nn.Conv2d):
+                    orig_weights = orig_weights.view(
+                        [
+                            base_layer.kernel_size[0] * base_layer.kernel_size[1] * base_layer.in_channels,
+                            base_layer.out_channels,
+                        ]
+                    )
+                delta_weight_inv = torch.inverse(self.get_delta_weight(active_adapter))
+                orig_weights = torch.mm(orig_weights, delta_weight_inv)
+                if isinstance(base_layer, nn.Linear):
+                    orig_weights = torch.transpose(orig_weights, 0, 1)
+                elif isinstance(base_layer, nn.Conv2d):
+                    orig_weights = orig_weights.view(
+                        [
+                            base_layer.out_channels,
+                            base_layer.in_channels,
+                            base_layer.kernel_size[0],
+                            base_layer.kernel_size[1],
+                        ]
+                    )
+                self.base_layer.weight.data = orig_weights
+
+    def get_delta_weight(self, adapter_name: str) -> torch.Tensor:
+        scaling = self.scaling[adapter_name]
+        rank = self.r[adapter_name]
+        coft = self.coft[adapter_name]
+        eps = self.eps[adapter_name]
+        opt_r = self.oft_r[adapter_name]
+
+        if coft:
+            with torch.no_grad():
+                opt_r.copy_(self._project_batch(opt_r, eps=eps))
+
+        # Perform rank dropout during training - drop rows of addition weights
+        rank_dropout = self.rank_dropout[adapter_name]
+        if self.training and rank_dropout:
+            drop = (torch.rand(opt_r.size(0)) > rank_dropout).to(opt_r.dtype)
+            drop = drop.view(-1, *[1] * len(opt_r.shape[1:])).to(opt_r.device)
+            # TODO: Investigate if there should be a scaler like in normal dropout during training
+            # Original implementation doesn't have it
+            # https://github.com/KohakuBlueleaf/LyCORIS/blob/eb460098187f752a5d66406d3affade6f0a07ece/lycoris/modules/OFT.py#L193
+            drop /= drop.mean()
+            opt_r *= drop
+
+        orth_rotate = self._cayley_batch(opt_r * scaling)
+        weight = self._block_diagonal(orth_rotate, rank)
+
+        return weight
+
+    def _cayley_batch(self, data: torch.Tensor) -> torch.Tensor:
+        b, r, c = data.shape
+        # Ensure the input matrix is skew-symmetric
+        skew = 0.5 * (data - data.transpose(1, 2))
+        I = torch.eye(r, device=data.device).unsqueeze(0).expand(b, r, c)
+
+        # Perform the Cayley parametrization
+        Q = torch.bmm(I - skew, torch.inverse(I + skew))
+
+        return Q
+
+    def _block_diagonal(self, oft_r: torch.Tensor, rank: int) -> torch.Tensor:
+        if oft_r.shape[0] == 1:
+            # block share
+            blocks = [oft_r[0, ...] for i in range(rank)]
+        else:
+            blocks = [oft_r[i, ...] for i in range(rank)]
+
+        # Use torch.block_diag to create the block diagonal matrix
+        A = torch.block_diag(*blocks)
+
+        return A
+
+    def _project_batch(self, oft_r, eps=1e-5):
+        # scaling factor for each of the smaller block matrix
+        eps = eps * 1 / torch.sqrt(torch.tensor(oft_r.shape[0]))
+        I = (
+            torch.zeros((oft_r.size(1), oft_r.size(1)), device=oft_r.device, dtype=oft_r.dtype)
+            .unsqueeze(0)
+            .expand_as(oft_r)
+        )
+        diff = oft_r - I
+        norm_diff = torch.norm(oft_r - I, dim=(1, 2), keepdim=True)
+        mask = (norm_diff <= eps).bool()
+        out = torch.where(mask, oft_r, I + eps * (diff / norm_diff))
+        return out
+
+    def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
+        previous_dtype = x.dtype
+
+        if self.disable_adapters:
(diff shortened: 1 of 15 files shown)


lukaskuhn-lku:
Really cool stuff! Thanks for taking over the issue, as I was busy implemented something else first

BenjaminBossan:
Thanks for the PR. Could you please run `make style`, so that the CI can run successfully?

okotaku:
@BenjaminBossan Done!

okotaku:
I uploaded trained weight.

https://huggingface.co/takuoko/oft_test/tree/main

```
from pathlib import Path

import torch
from diffusers import DiffusionPipeline, AutoencoderKL
from peft import PeftModel

checkpoint = 'takuoko/oft_test'
prompt = 'yoda pokemon'

vae = AutoencoderKL.from_pretrained(
    'madebyollin/sdxl-vae-fp16-fix',
)
pipe = DiffusionPipeline.from_pretrained(
    'stabilityai/stable-diffusion-xl-base-1.0', vae=vae,
    )
pipe.to('cuda')
pipe.unet = PeftModel.from_pretrained(pipe.unet, checkpoint, adapter_name="default")

image = pipe(
    prompt,
    num_inference_steps=50,
    height=1024,
    width=1024,
).images[0]
image.save('demo.png')
```

src/peft/tuners/oft/model.py
```diff
@@ -0,0 +1,115 @@
...
+        `torch.nn.Module`: The OFT model.
+
+    Example:
+        ```py
+        >>> from diffusers import StableDiffusionPipeline
+        >>> from peft import OFTModel, OFTConfig
+
+        >>> config_te = OFTConfig(
+        ...     r=8,
+        ...     lora_alpha=32,
```
lukaskuhn-lku, line 48: should be alpha not lora_alpha

src/peft/tuners/oft/model.py
```diff
@@ -0,0 +1,115 @@
...
+        ...     target_modules=["k_proj", "q_proj", "v_proj", "out_proj", "fc1", "fc2"],
+        ...     rank_dropout=0.0,
+        ...     module_dropout=0.0,
+        ...     init_weights=True,
+        ... )
+        >>> config_unet = OFTConfig(
+        ...     r=8,
+        ...     lora_alpha=32,
```
lukaskuhn-lku, line 56: same here: should be alpha, not lora_alpha

lukaskuhn-lku: also there is no rank_dropout in OFTConfig, so this needs to be deleted as well

src/peft/tuners/oft/layer.py
```diff
@@ -0,0 +1,398 @@
...
+                module_dropout = self.module_dropout[active_adapter]
+
+                if (not self.training) or (self.training and torch.rand(1) > module_dropout):
+                    weight = self._get_delta_activations(active_adapter, weight, *args, **kwargs)
+
+            if weight is None:
+                # skip all active_adapter
+                result = self.base_layer(x, *args, **kwargs)
+            else:
+                result = self._forward_from_weight(x, weight, *args, **kwargs)
```
BenjaminBossan, line 287: Hmm, can we refactor this to always use `self.base_layer(x, *args, **kwargs)` by adding the delta activations on top? This probably requires to move the updates inside the loop, instead of accumulating all the updated before applying them. This may also simplify some code, as I think we can avoid the `prev_weight` part.

The problem is that if we don't call `self.base_layer(...)`, we completely sidestep that layer's `forward` function. This is okay when we know it's just a normal `nn.Linear`, but we want to allow nested adapters, so the base layer could for instance be an `lora.Linear` layer, in which case this would give the wrong result.

src/peft/tuners/oft/model.py
```diff
@@ -0,0 +1,112 @@
...
+from torch import nn
+
+from peft.tuners.lycoris_utils import LycorisConfig, LycorisTuner
+
+from .layer import Conv2d, Linear, OFTLayer
+
+
+class OFTModel(LycorisTuner):
+    """
+    Creates Orthogonal Finetuning model from a pretrained model. The method is partially described in
```
BenjaminBossan, line 30: What part is not described if it's only "partially" described?

examples/oft_dreambooth/oft_dreambooth_inference.ipynb
```diff
@@ -0,0 +1,81 @@
...
+   "outputs": [],
+   "source": [
+    "prompt = \"A photo of a sks dog\"\n",
+    "image = pipe(\n",
+    "    prompt,\n",
+    "    num_inference_steps=50,\n",
+    "    height=512,\n",
+    "    width=512,\n",
+    ").images[0]\n",
+    "#image.save('demo.png')\n",
```
BenjaminBossan, line 50: Remove this?

examples/oft_dreambooth/oft_dreambooth_inference.ipynb
```diff
@@ -0,0 +1,81 @@
+{
```
BenjaminBossan, line 1: The notebook assumes that the user ran the training script first to create an adapter, right? Let's mention that. Otherwise, we could maybe copy https://huggingface.co/takuoko/oft_test to our PEFT HF repo and link that one.

BenjaminBossan (changes requested):
Thank you so much @okotaku for this PR. It's already in a super clean state, integrates really well with PEFT and supports many features like merging. Excellent work!

Also, thanks @lukaskuhn-lku for already providing feedback.

I added a couple of points, please check them out. None should be really big issues, so hopefully we can get this merged soon.

Regarding the faithfulness of the implementation: I read the paper and gained a high level understanding, but could not always match that to the actual code. So I'm putting some trust into the code here that it works as intended, based on your judgment, the training results, and on the code of the [other repo](https://github.com/Zeju1997/oft) which this is partly based on.

src/peft/tuners/oft/layer.py
```diff
@@ -0,0 +1,398 @@
...
+                self.merged_adapters.append(active_adapter)
+
+    def unmerge(self) -> None:
+        if not self.merged:
+            warnings.warn("Already unmerged. Nothing to do.")
+            return
+        while len(self.merged_adapters) > 0:
+            active_adapter = self.merged_adapters.pop()
+            if active_adapter in self._available_adapters:
+                base_layer = self.base_layer
```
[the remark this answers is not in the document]
lukaskuhn-lku: @okotaku Fixed in my PR

src/peft/tuners/oft/layer.py
```diff
@@ -0,0 +1,398 @@
...
+                elif isinstance(base_layer, nn.Conv2d):
+                    orig_weights = orig_weights.reshape(
+                        [
+                            base_layer.out_channels,
+                            base_layer.in_channels,
+                            base_layer.kernel_size[0],
+                            base_layer.kernel_size[1],
+                        ]
+                    )
+                self.base_layer.weight.data = orig_weights
```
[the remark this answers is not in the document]
lukaskuhn-lku: @okotaku Fixed in my PR

examples/oft_dreambooth/oft_dreambooth_inference.ipynb
lukaskuhn-lku: @okotaku Fixed this in my PR

examples/oft_dreambooth/oft_dreambooth_inference.ipynb
lukaskuhn-lku: I mentioned it now in [PR#2](https://github.com/okotaku/peft/pull/2). Don't know how to integrate the oft_test adapter into the PEFT HF repo so I would need some guidance to help with that.

src/peft/tuners/oft/layer.py
```diff
@@ -0,0 +1,398 @@
...
+        if coft:
+            with torch.no_grad():
+                opt_r.copy_(self._project_batch(opt_r, eps=eps))
+
+        orth_rotate = self._cayley_batch(opt_r * scaling)
+        weight = self._block_diagonal(orth_rotate, rank)
+
+        return weight
+
+    def _cayley_batch(self, data: torch.Tensor) -> torch.Tensor:
```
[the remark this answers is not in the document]
lukaskuhn-lku: I found some other occurrences and added all of them as permanent comments in [PR#2](https://github.com/okotaku/peft/pull/2) :)

src/peft/tuners/oft/model.py
okotaku: It's copied from https://github.com/huggingface/peft/blob/main/src/peft/tuners/loha/model.py#L30.
I delete it.

src/peft/tuners/oft/layer.py
```diff
@@ -0,0 +1,398 @@
...
+        b, r, c = data.shape
+        # Ensure the input matrix is skew-symmetric
+        skew = 0.5 * (data - data.transpose(1, 2))
+        I = torch.eye(r, device=data.device).unsqueeze(0).expand(b, r, c)
+
+        # Perform the Cayley parametrization
+        Q = torch.bmm(I - skew, torch.inverse(I + skew))
```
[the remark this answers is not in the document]
okotaku: https://github.com/Zeju1997/oft/blob/main/oft-t2i/oft_utils/attention_processor.py#L505

The official implementation uses inverse, which is why I use it.
However, I understand your opinion, and I will train models using `pinv`. If it works, we can add `pinv` option.

src/peft/tuners/oft/layer.py
```diff
@@ -0,0 +1,398 @@
...
+                        [
+                            base_layer.kernel_size[0] * base_layer.kernel_size[1] * base_layer.in_channels,
+                            base_layer.out_channels,
+                        ]
+                    )
+                delta_weight = self.get_delta_weight(active_adapter)
+                if orig_weights.shape[0] != delta_weight.shape[1]:
+                    # when in channels is not divisible by r
+                    delta_weight = delta_weight[: orig_weights.shape[0], : orig_weights.shape[0]]
+                orig_weights = torch.linalg.solve(delta_weight, orig_weights)
```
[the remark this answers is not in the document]
okotaku: It is difficult to guarantee that an exact solution will always be found. Do you have any alternative plans?

src/peft/tuners/oft/layer.py
okotaku: @BenjaminBossan 
I have refactored `layer.py`.
To achieve this method, I fixed OFT method similar to kohya_ss.

https://github.com/kohya-ss/sd-scripts/blob/95ae56bd22c285ccb2fe5fca96d92f39842bb99b/networks/oft.py#L77-L89

examples/oft_dreambooth/oft_dreambooth_inference.ipynb
BenjaminBossan: I think it's okay now as is, this way users who just open the notebook know what they should do first.

src/peft/tuners/oft/layer.py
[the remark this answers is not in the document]
BenjaminBossan: Nice. Probably the authors had a good reason to use exact inverse and not pinv, but it could be worth a try.

tests/test_stablediffusion.py
```diff
@@ -115,13 +130,17 @@ def prepare_inputs_for_testing(self):
...
             },
         )
     )
     def test_merge_layers(self, test_name, model_id, config_cls, config_kwargs):
         if config_cls == LoHaConfig:
             # TODO: This test is flaky with PyTorch 2.1 on Windows, we need to figure out what is going on
             self.skipTest("LoHaConfig test is flaky")
+        if config_cls == OFTConfig:
+            # TODO: This test is flaky with PyTorch 2.1 on Windows, we need to figure out what is going on
+            self.skipTest("OFTConfig test is flaky")
```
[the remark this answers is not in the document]
BenjaminBossan: I ran the test a couple of times and it passed on my Linux machine. I don't have Windows to test it. How big is the difference, are we talking about a very small deviation or something like completely random results? Does setting a seed help with flakiness?

If nothing helps and the deviation on Windows is very small, we could increase the tolerance a bit. Otherwise, let's skip the test on Windows.

src/peft/tuners/oft/layer.py
```diff
@@ -0,0 +1,378 @@
...
+            result = self.base_layer(x, *args, **kwargs)
+        else:
+            result = self.base_layer(x, *args, **kwargs)
+            if len(result.shape) == 4:
+                result = result.permute(0, 2, 3, 1)
+
+            base_layer = self.get_base_layer()
+            base_bias = base_layer.bias
+            if base_bias is not None:
+                result = result - base_bias.data
```
BenjaminBossan, line 287: Not super happy that this is here. I assume it is necessary for correctness? Could you please explain and maybe add a comment?

BenjaminBossan (commented):
@okotaku Thanks for the changes, the PR looking very mature and not much more should be needed to merge it.

I wasn't quite sure if this PR was ready for review or not, so I just did a light pass on it and left a few comments. LMK if/when this is ready.

src/peft/tuners/oft/layer.py
okotaku: @BenjaminBossan In my quick check, there are some differences.

```
import torch


def cayley_batch(data):
        b, r, c = data.shape
        # Ensure the input matrix is skew-symmetric
        skew = 0.5 * (data - data.transpose(1, 2))
        # I = torch.eye(r, device=data.device).unsqueeze(0).repeat(b, 1, 1)
        I = torch.eye(r, device=data.device).unsqueeze(0).expand(b, r, c)

        # Perform the Cayley parametrization
        Q = torch.bmm(I - skew, torch.inverse(I + skew))

        return Q

def block_diagonal(R):
        if len(R.shape) == 2:
            # Create a list of R repeated block_count times
            blocks = [R] * r
        else:
            # Create a list of R slices along the third dimension
            blocks = [R[i, ...] for i in range(r)]

        # Use torch.block_diag to create the block diagonal matrix
        A = torch.block_diag(*blocks)

        return A

in_features = 128
r = 4
eps = 6e-5
R_shape = [r, in_features // r, in_features // r]
R = torch.zeros(R_shape[1], R_shape[1])
R = torch.stack([R] * r)
eps = eps * R_shape[1] * R_shape[1]

# generate random R
orth_rotate = cayley_batch(torch.rand(*R.shape) * 0.001)
block_diagonal_matrix = block_diagonal(orth_rotate)

from torch import nn
attn = nn.Linear(32, 128, bias=True)
fix_filt = attn.weight.data
base_bias = attn.bias.data if attn.bias is not None else None
filt = torch.mm(torch.transpose(fix_filt, 0, 1), block_diagonal_matrix)
filt = torch.transpose(filt, 0, 1)

x = torch.rand(1, 32)

out = x
out = nn.functional.linear(input=out, weight=filt, bias=base_bias)

out2 = attn(x)
out2 = torch.matmul(out2, block_diagonal_matrix)

out3 = attn(x) - base_bias
out3 = torch.matmul(out3, block_diagonal_matrix) + base_bias
```

out == out3, out2 != out3

src/peft/tuners/oft/layer.py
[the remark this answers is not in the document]
okotaku: When I replaced `torch.inverse` with `torch.pinverse`, it raised the error `RuntimeError: expected scalar type Float but found Half`.  To avoid this error, I tried disabling autocast by setting `autocast enbale=False`. However, this resulted in longer training times compared to using `torch.inverse`. Therefore, I believe there is no need to support `pinv`.

```
if pinv and torch.is_autocast_enabled():
    device = 'cuda' if data.is_cuda else 'cpu'
    # to avoid `RuntimeError: expected scalar type Float but found Half`
    with torch.autocast(device_type=device, enabled=False):
        Q = torch.bmm(I - skew, torch.pinverse(I + skew))
else:
    Q = torch.bmm(I - skew, torch.inverse(I + skew))
```

okotaku:
@BenjaminBossan Now, PR is ready for me.

src/peft/tuners/oft/layer.py
```diff
@@ -0,0 +1,377 @@
...
+                            base_layer.kernel_size[1],
+                        ]
+                    )
+
+                if safe_merge and not torch.isfinite(orig_weights).all():
+                    raise ValueError(
+                        f"NaNs detected in the merged weights. The adapter {active_adapter} seems to be broken"
+                    )
+
+                base_layer.weight.data = orig_weights
```
BenjaminBossan, line 170: Nit: I find the variable name `orig_weights` a bit confusing, since at this point, they're no longer the "original weights" but have been transformed quite a lot. Maybe `new_weights` could be used once `orig_weights` have been changed.

src/peft/tuners/oft/layer.py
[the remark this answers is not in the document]
BenjaminBossan: Thanks for testing, I wasn't aware that `pinv` doesn't work with half precision. As you wrote, it doesn't make sense then.

Something I wonder though is why get an issue with half precision in the first place. The `I` tensor that is created is in float32, which means that `I + skew` is also upcast to float32, even if `data` is half precision. How is it possible to get the runtime error then when using `pinv`?

Also, in my testing, `torch.inverse` also requires full precision:

```python
torch.inverse(torch.rand(10, 10).half())
# => RuntimeError: linalg.inv: Low precision dtypes not supported. Got Half
```

Any idea what could explain this?

src/peft/tuners/oft/layer.py
```diff
@@ -0,0 +1,377 @@
...
+                            base_layer.out_channels,
+                            base_layer.in_channels * base_layer.kernel_size[0] * base_layer.kernel_size[1],
+                        ]
+                    )
+                    orig_weights = torch.transpose(orig_weights, 0, 1)
+                delta_weight = self.get_delta_weight(active_adapter)
+                if orig_weights.shape[1] != delta_weight.shape[1]:
+                    # when in channels is not divisible by r
+                    delta_weight = delta_weight[: orig_weights.shape[1], : orig_weights.shape[1]]
+                delta_inv = torch.linalg.inv(delta_weight)
```
BenjaminBossan, line 196: Nit: Sometimes, `torch.linalg.inv` is used, sometimes `torch.inverse`, but they are the same function. Let's be consistent.

src/peft/tuners/oft/layer.py
BenjaminBossan: > out2 = attn(x)
> out2 = torch.matmul(out2, block_diagonal_matrix)
> out3 = attn(x) - base_bias
> out3 = torch.matmul(out3, block_diagonal_matrix) + base_bias

Yes, a difference is expected when in one case, we first add the bias, then multiply, in the other we first multiply, then add the bias. I understand why we would first want to accumulate all the multiplications before applying the bias, it's just unfortunate that it means we have to manually remove and add the bias, which I think could be error prone. I have no solution really, except to add a comment why this is done.

BenjaminBossan (approved):
Thanks so much for the updates, I only have some nits left and a question about the precision when using inverse. None of these are blockers, up to you if you want to take them up.

In the meantime, I'll check if we can get an additional reviewer in.

BenjaminBossan:
Oh no, a test is failing. It was passing for me locally, so at first this seemed strange. Well, it turns out that it fails on CPU but works with CUDA, which is why it passes locally but not on CI. With CPU, the differences are actually quite big, with an average of 45, and one pixel even 252. Therefore, this is not a simple precision issue but smells like a bug to me.

src/peft/tuners/oft/layer.py
```diff
@@ -0,0 +1,377 @@
...
+            init_weights (`bool`): Whether to initialize weights.
+            coft (`bool`): Whether to use the constrainted variant of OFT or not.
+            eps (`float`):
+                The control strength of COFT. The freedom of rotation. Only has an effect if `coft` is set to True.
+            block_share (`bool`): Whether to share the OFT parameters between blocks or not.
+        """
+
+        self.r[adapter_name] = r
+        self.alpha[adapter_name] = alpha
+        self.scaling[adapter_name] = alpha / r
```
pacman100, line 85: This isn't used in the paper, right?

pacman100 (commented):
Thank you @okotaku for doing an exceptional job at adding OFT method! 🤗🚀✨
The PR is in great shape to be merged once all tests pass! 🚀
Left a comment.

okotaku:
@BenjaminBossan I feel it is the same issue as `LoHaConfig`. That's why they skipped `test_merge_layers`.
https://github.com/okotaku/peft/blob/main/tests/test_stablediffusion.py#L122-L124

When I tested `lora_kwargs`, `loha_kwargs` and `oft_kwargs` in CPU, all of them failed.
https://github.com/okotaku/peft/blob/oft/tests/test_stablediffusion.py#L131-L133

src/peft/tuners/oft/layer.py
okotaku: Exactly.

src/peft/tuners/oft/layer.py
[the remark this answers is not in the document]
okotaku: Hmm...
`torch.inverse` works well with `torch.autocast`. However, I don't have any ideas to explain it.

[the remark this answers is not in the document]
okotaku: And the error was raised in the backward process.

```
 File "/usr/local/lib/python3.10/dist-packages/mmengine/optim/optimizer/optimizer_wrapper.py", line 196, in update_params
    self.backward(loss)
  File "/usr/local/lib/python3.10/dist-packages/mmengine/optim/optimizer/amp_optimizer_wrapper.py", line 125, in backward
    self.loss_scaler.scale(loss).backward(**kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/_tensor.py", line 491, in backward
    torch.autograd.backward(
  File "/usr/local/lib/python3.10/dist-packages/torch/autograd/__init__.py", line 204, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: expected scalar type Float but found Half
```

src/peft/tuners/oft/layer.py
pacman100: then it would be better to remove these.

BenjaminBossan:
@okotaku Good point, I had forgotten about this issue.

Regarding the removal of the scaling parameter: Before, the default scaling was 1/8, now it is 1. Do you think this will have an impact on results?

BenjaminBossan:
Heads up, there is a small merge conflict, should be easy to fix.

okotaku:
@BenjaminBossan 

>Regarding the removal of the scaling parameter: Before, the default scaling was 1/8, now it is 1. Do you think this will have an impact on results?

I trained a model without scaling, and I feel that they have no issues.

image0_step_49

image0_step_48

pacman100 (approved):
Thank you @okotaku for doing a great job on adding OFT and addressing all the comments, LGTM! 🤗🚀✨

However, I do like the results when scaling was used 😅 but let's keep the implementation close to paper at present.

BenjaminBossan:
Thanks so much @okotaku and @lukaskuhn-lku for this amazing addition and high quality PR.

okotaku:
@pacman100 Haha, I like with scaling too ;)
@BenjaminBossan @lukaskuhn-lku 
Thank you for all your contributions!

BenjaminBossan:
Hey @okotaku @lukaskuhn-lku I encountered a small issue with OFT, maybe you have an idea. My problem is that for certain configs, OFT results in an identity transform, even though `init_weights=False` was set. Given the script below:

```python
import torch
from torch import nn
from peft import get_peft_model, OFTConfig

class SimpleNet(nn.Module):
        super().__init__()
        self.lin0 = nn.Linear(10, 20, bias=bias)
        self.relu = nn.ReLU()
        self.lin1 = nn.Linear(20, 5, bias=bias)

    def forward(self, X):
        X = X.float()
        X = self.lin0(X)
        X = self.relu(X)
        X = self.lin1(X)
        return X

input = torch.arange(30).reshape(3, 10)

torch.manual_seed(0)
base_model = SimpleNet()
output_base = base_model(input)

torch.manual_seed(1)
config = OFTConfig(target_modules=["lin1"], init_weights=False)            # FAILS
# config = OFTConfig(target_modules=["lin1"], init_weights=False, r=4)     # PASSES
# config = OFTConfig(target_modules=["lin0"], init_weights=False)          # PASSES
# config = OFTConfig(target_modules=["lin0", "lin1"], init_weights=False)  # PASSES
peft_model = get_peft_model(base_model, config)
output_oft = peft_model(input)

assert not torch.allclose(output_base, output_oft, atol=1e-5, rtol=1e-5)
```

as mentioned in the comments, the first config fails, whereas the others pass the test. Digging into the code, the problem is that `_cayley_batch` returns a `torch.ones((8, 1, 1))`, which can be traced back to the skew being `torch.zeros((8, 1, 1))`, which is not surprising, given that `data.shape` is 8, 1, 1.

https://github.com/huggingface/peft/blob/6a57472665b2b712a84e2bedd98945038283f7cc/src/peft/tuners/oft/layer.py#L229-L238

The trailing dimensions are always 1 if the dimension of the weight is lower or equal to rank `r`. This may be a bit of an edge case, but could for instance happen if someone uses OFT to fine-tune a classification layer.

Do you think `_cayley_batch` could be adjusted for this edge case or is this something we just have to accept?

okotaku:
@BenjaminBossan That's right.
`_cayley_batch` cannot be adjusted for this edge. We have two options:
1. Raise an error when r >= dim.
2. Accept and ignore it.

PR #1163, opened 2023-11-21
Mixed adapter models

BenjaminBossan:
This PR is a follow up to, and supersedes, #1069. That PR is no longer necessary, as the base layer refactor PR has been merged.

## Description

This PR allows to add adapters of different types, e.g. LoRA and LoHa:

```python
base_model = ...
config0 = LoraConfig(...)
peft_model = get_peft_model(base_model, config0, mixed=True) config1 = LoHaConfig(...)
peft_model.add_adapter(config1, "other")
peft_model.set_adapter(["default", "other"])
peft_model(x)
```

At this point, both adapters are active at the same time.

Existing code should not be affected by this change, since users need to opt into this behavior by setting `mixed=True`, and a completely different class is being used (`PeftMixedModel`).

Also interesting is that this method can be used for a single adapter type but with very different configs. Right now, we have limited support for that (e.g. for LoRA, different `r` values by using `rank_pattern`), but with this, we don't need to special case the differing arguments anymore.

## Not implemented

- I'm not yet sure if the same logic can be applied to IA³ or if it may fail because IA³ can apply its scaling to the input, not the output.
- It is currently not possible to represent a mixed adapter model as a single config. I think we can come up with a solution but I don't think it is necessary for a first version of this.
- Saving and loading is not yet implemented for mixed models for the reason above.

Those could potentially be added in a future PR.


pacman100 (approved):
Amazing work @BenjaminBossan! 🔥🚀🤗

Super clean, nice APIs, readable with comprehensive tests 🔥! Left a nit suggestion.

tests/test_mixed.py
```diff
@@ -0,0 +1,621 @@
...
+    def _get_model(self, model_cls, peft_config=None, adapter_name=None, seed=0):
+        torch.manual_seed(0)  # always use seed 0 for base model, seed for adapters may differ
+        base_model = model_cls().eval().to(self.torch_device)
+        if peft_config is None:
+            return base_model
+
+        torch.manual_seed(seed)
+        assert adapter_name is not None
+        peft_model = get_peft_model(base_model, peft_config, adapter_name=adapter_name, mixed=True)
+        return peft_model.eval().to(self.torch_device)
```
younesbelkada, line 67: Shouldn't `get_peft_model` always call `eval()` before returning the model ? for `MixedModel` since it is intended for inference perhaps we should call `.eval()` before returning the model (if not already the case)

younesbelkada (approved):
Impressive and inspiring work @BenjaminBossan ! I left one open question that is not a blocker, shall we also add few lines in the main README?

src/peft/mixed_model.py
```diff
@@ -0,0 +1,320 @@
...
+PEFT_TYPE_TO_MODEL_MAPPING = {
+    PeftType.LORA: LoraModel,
+    PeftType.LOHA: LoHaModel,
+    PeftType.LOKR: LoKrModel,
+    PeftType.ADALORA: AdaLoraModel,
+    PeftType.IA3: IA3Model,
+}
+
+
+def _prepare_model_for_gradient_checkpointing(model: nn.Module) -> None:
```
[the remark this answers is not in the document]
BenjaminBossan: The original method has no such argument:

https://github.com/huggingface/peft/blob/fb607d00adc7b8aa1694f58a786ce2cd5681710d/src/peft/peft_model.py#L387

Maybe you meant `prepare_model_for_kbit_training`?

tests/test_mixed.py
BenjaminBossan: I think it's better to be consistent here and not automatically call `eval()`, since we don't do that for the other cases. Also, I think in principle, training should be possible, although there probably isn't much point in doing so.

pacman100 (approved):
Thank you @BenjaminBossan for iterating, LGTM! 🚀

PR #1164, opened 2023-11-21
Dataset was loaded twice in 4-bit finetuning script

lukaskuhn-lku:
Was working on something unrelated and saw that the dataset was always loaded twice in the finetuning script so I removed the double occurence.


BenjaminBossan (approved):
Thanks for removing this duplication.

PR #1165, opened 2023-11-22
Remove explicit declaration of 'label' arg to add support for Causal Models without LM head

huzama:
When calling the forward pass after applying peft to Causal Models (without LM Head), an error occurs:
`"TypeError: forward() got an unexpected keyword argument 'labels'".`

The bug was fixed by removing the explicit declaration of arg 'label' and allowing **kwargs to handle the input label.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -978,6 +978,7 @@ class PeftModelForCausalLM(PeftModel):
         input_ids=None,
         attention_mask=None,
         inputs_embeds=None,
+        labels=None,
         output_attentions=None,
         output_hidden_states=None,
         return_dict=None,
@@ -992,6 +993,7 @@ class PeftModelForCausalLM(PeftModel):
                 return self.base_model(
                     input_ids=input_ids,
                     attention_mask=attention_mask,
+                    labels=labels,
                     output_attentions=output_attentions,
                     output_hidden_states=output_hidden_states,
                     return_dict=return_dict,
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Thanks for working on this fix. Do you have an example where the current code is failing and the fix solves it? This would help in writing a test.

huzama:
Sure, here is a snippet that can reproduce the error:
```
from peft import PrefixTuningConfig, TaskType, get_peft_model
from transformers import AutoModel

model_ = AutoModel.from_pretrained("gpt2")


config = PrefixTuningConfig(
    task_type=TaskType.CAUSAL_LM, inference_mode=False, num_virtual_tokens=16
)

model = get_peft_model(model_, config)

model(input_ids=model_.dummy_inputs["input_ids"])

```


> Output:
> TypeError: forward() got an unexpected keyword argument 'labels'
 


The error occurs when the base model does not accept the 'labels' parameter, which is the case for models without the lm_head layer.

_I also updated the code to make minimal changes while fixing the bug, so as not to break anything else._

BenjaminBossan:
Great, thanks for providing a snippet that demonstrates the issue. Do you think you could write a unit test that basically contains that snippet? It's always nice when fixing a bug to have a unit test that breaks without the fix but works with the fix. We can assist you with adding that test if you need some support.

pacman100:
Hello, CausalLM does have `labels` as param. Please use the task `TaskType.FEATURE_EXTRACTION` with `AutoModel` that doesn't use any cls/lm head

pacman100:
I don't think there is need for this PR given that task `TaskType.FEATURE_EXTRACTION` is meant to `AutoModel` class

huzama:
@pacman100 Thank you for your comment, you are absolutely right. I will close this PR.

PR #1063, opened 2023-10-29
Align offloaded parameters before merging

blbadger:
Fixes issue #868  to allow for merging and unloading using `auto` device alignment.   

For models of sufficient size relative to memory constraints, choosing `auto` device alignment causes `accelerate` to perform lazy offloaded parameter loading just before a forward call. Calling `merge_and_unload` on a model with offloaded parameters results in errors due to these offloaded parameters.

This PR addresses that issue by activating the appropriate `accelerate` device alignment hook methods for each module that is to be merged.  This loads offloaded parameters which are otherwise on the "meta' device (if the model does not fit on the execution device) to allow merging and then offloads the merged parameters to avoid out-of-memory errors.

The implementation is fairly straightforward with one exception: it seems that we must avoid calling`self._replace_module(parent, target_name, new_module, target)` as doing so leads to a memory leak as the `new_module` does not have the appropriate `AlignDevicesHook` hook.  Another approach would be to add this hook and proceed with replacement, but this may not be necessary for successful merging right now.

The main benefit compared to the existing approach of loading models using `cpu` is that this method should succeed even when cpu cannot load the model in its entirety, as each module is loaded and merged and unloaded sequentially.

PR #1167, opened 2023-11-22
Add test for merge_and_unload of offloaded models

blbadger:
Adds a test to determine if the merging and offloading of (offloaded) LoRA parameters succeeds given a model with offloaded modules.

This PR submits the test from PR #1063, which addresses problems in merging and unloading offloaded models.  

@BenjaminBossan


BenjaminBossan:
Thanks @blbadger for creating this new PR.

Unfortunately, I still encounter the [same error](https://github.com/huggingface/peft/pull/1063#issuecomment-1820719371) when running the test locally

> Errors in offloaded models upon calling PeftModel.from_pretrained() are typical of the old problem of a lack of nested modules causing a LoRA parameter loading issue. I think this is a version problem, as for example in my peft dev environment the test uses peft==0.6.0.dev0 (although I am not sure why this would be) whereas running the test externally and pulling from the head, ie peft==0.6.3.dev0 passes.

I'm working on PEFT `main`, so this can't really be the issue. I tried different values for `memory_limits` but to no avail.

Btw I uploaded some compatible LoRA weights to our testing repo: https://huggingface.co/peft-internal-testing/gpt2-lora-random. So `lora_checkpoint = "peft-internal-testing/gpt2-lora-random"` should work. Also used `safetensors` for good measure.

blbadger:
> Unfortunately, I still encounter the [same error](https://github.com/huggingface/peft/pull/1063#issuecomment-1820719371) when running the test locally
> I'm working on PEFT `main`, so this can't really be the issue. I tried different values for `memory_limits` but to no avail.

I think I see what you mean, switching to `main` and installing gives me the same error.  Installing the latest version of `accelerate` should fix this issue.

> Btw I uploaded some compatible LoRA weights to our testing repo: https://huggingface.co/peft-internal-testing/gpt2-lora-random. So `lora_checkpoint = "peft-internal-testing/gpt2-lora-random"` should work. Also used `safetensors` for good measure.

Thanks very much for uploading! Unfortunately the test fails with the safetensor adapter, the final assertion fails.  Checking the values suggests that this is not a rounding error either.

pacman100 (approved):
Thank you @blbadger for adding this test, LGTM! 🚀

pacman100 (changes requested):
but fails if I use `peft-internal-testing/gpt2-lora-random`

pacman100:
>  I have not tried working with adapters saved in safetensors before, but can look into this if using safetensors would be preferable.

That would be great given that `safetensors` is not the default format in HF libraries.

tests/test_gpu_examples.py
```diff
@@ -933,3 +936,49 @@ def test_causal_lm_training_mutli_gpu(self):
...
+        """
+        tokenizer = AutoTokenizer.from_pretrained(self.causal_lm_model_id)
+        config = AutoConfig.from_pretrained(self.causal_lm_model_id)
+        with init_empty_weights():
+            model = AutoModelForCausalLM.from_config(config)
+
+        memory_limits = {0: "0.5GIB", "cpu": "5GIB"}
+        # offloads around half of all transformer modules
+        device_map = infer_auto_device_map(model, max_memory=memory_limits)
+        assert 0 in device_map.values()
```
BenjaminBossan, line 968: Could you please always use `self.assertX` for consistency?

tests/test_gpu_examples.py
```diff
@@ -933,3 +936,49 @@ def test_causal_lm_training_mutli_gpu(self):
...
+        lora_checkpoint = "blbadger/gpt2-lora"
+
+        # max_memory must be specified again to avoid defaulting to 'auto' device map
+        model = PeftModel.from_pretrained(model, lora_checkpoint, max_memory=memory_limits)
+        input_tokens = tokenizer.encode("Four score and seven years ago", return_tensors="pt")
+        pre_merge_olayer = model(input_tokens)[0]
+        model = model.merge_and_unload()
+        post_merge_olayer = model(input_tokens)[0]
+
+        self.assertTrue(torch.all(pre_merge_olayer == post_merge_olayer))
```
BenjaminBossan, line 984:
```suggestion
        self.assertTrue(torch.allclose(pre_merge_olayer, post_merge_olayer))
```

tests/test_gpu_examples.py
```diff
@@ -933,3 +936,49 @@ def test_causal_lm_training_mutli_gpu(self):
...
+        # offloads around half of all transformer modules
+        device_map = infer_auto_device_map(model, max_memory=memory_limits)
+        assert 0 in device_map.values()
+        assert "cpu" in device_map.values()
+
+        model = AutoModelForCausalLM.from_pretrained(self.causal_lm_model_id, device_map=device_map)
+        assert len({p.device for p in model.parameters()}) == 2
+
+        # LoRA adapters on ['c_attn', 'c_proj', 'c_fc']
+        lora_checkpoint = "blbadger/gpt2-lora"
```
BenjaminBossan, line 975: Thinking more about this, do we even need a lora checkpoint on the Hub? I believe it would be easier to create a LoRA model on the fly:

```python
        torch.manual_seed(0)
        config = LoraConfig(
            task_type="CAUSAL_LM",
            init_lora_weights=False,
        )
        model = AutoModelForCausalLM.from_pretrained(self.causal_lm_model_id)

            model.save_pretrained(tmp_dir)
            # load the model with device_map
            model = AutoModelForCausalLM.from_pretrained(self.causal_lm_model_id, device_map=device_map).eval()
            self.assertTrue(len({p.device for p in model.parameters()}) == 2)
            model = PeftModel.from_pretrained(model, tmp_dir, max_memory=memory_limits)
```

WDYT?

BenjaminBossan (changes requested):
So indeed the test passes if I install the latest accelerate from main. This means we should skip the test if the accelerate version is too low. Could you please fix that? Technically, this also means that we should pin the min accelerate version in the dependencies, but I think this feature is enough of a corner case that we don't need to do that. WDYT?

I investigated further why the pickle checkpoint works but safetensors fails. The issue is that the pickle checkpoint was initialized without `init_lora_weights=False`. To wit:

```python
>>> model.base_model.model.transformer.h[0].attn.c_attn.lora_B["default"].weight
Parameter containing:
tensor([[0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.],
        ...,
        [0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.]], device='cuda:0')
```

Now to the question why this fails with `init_lora_weights=False`, which I used for the safetensors checkpoint:

In general, the values are very close, below is a sample of 3000 points:

image

However, there is a systematic error:

image

When I change the `memory_limits` to use all GPU or all CPU, the test passes, so the error originates from the mixing of CPU and GPU.

tests/test_gpu_examples.py
blbadger: This sounds good to me! As long as we load via `.from_pretrained` the test should work well (once we get the merging accuracy fixed that is)

tests/test_gpu_examples.py
blbadger: Changed to assertTrue

blbadger:
Thanks for the reviews @pacman100 @BenjaminBossan !  Good catch on `init_lora_weights=False` as well.  

For the version, I suppose it is your preference.  Perhaps this is enough of a corner issue to not require the latest `accelerate`, but then maybe we should add a warning for merging so the error you found is not silent.

The parameter merging accuracy issue seems to be tricky one. As far as I can tell so far, the error lies in the failure to onload adapter parameters during `get_delta_weight`, as the adapters lose their hooks somewhere between loading and getting the delta weight.  I will work on tracking this down to fix the issue:)

Also on close inspection, the final assertion also fails when LoRA adapters are set on certain modules (`wte` or `wpe` or both) even when the module exists only on the cpu (and has no align devices hook). Here it looks like a delta weight is not calculated at all, although I have not looked into this issue very long as it is apparently different than the one we are dealing with for offloaded parameters.

blbadger:
> I investigated further why the pickle checkpoint works but safetensors fails. The issue is that the pickle checkpoint was initialized without `init_lora_weights=False`. To wit:

After some effort I have worked out a solution to this issue: first each LoRA layer must call the device alignment hook to onload A and B params (on the leaf nodes as parents don't have the hook), merging can occur, then the merged weights must be used to update the base layer's weight map and the params are offloaded again.  The new test passes when this is performed regardless of the device map:)

@BenjaminBossan would you like me to put in another PR with this implementation? We could add on to #1063 but it would probably be simpler to just make it a new one.

BenjaminBossan:
Superseded by #1190

Issue #1155, opened 2023-11-21
A (possible) bug in lora merging method: add_weighted_adapter


jihuishan:
### System Info

peft 0.6.1

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [X] My own task or dataset (give details below)

### Reproduction

In both line 565 and 578 of add_weighted_adapter from, peft/tuners/lora/model.py, lora_A and lora_B are upgraded by
 target_lora_A.data += current_adapter_lora_A.data * weight * target.scaling[adapter]
target_lora_B.data += current_adapter_lora_B.data
It is quite confusing (and performance degrading) that the upgrade of target_lora_B is not relevant with the weight, although this seems like a delibrate design (as target_lora_A is upgraded correctly). 

### Expected behavior

When using add_weighted_adapter, me expect all params of each adapter is weightedly combined, instead of weightedly averaging across lora_A while evenly averaging for lora_B. 
During my experiments on merging several loras for a unified model, the performance drops signicantly if the problem above is not solved. For example, lets say there are five adapters with same configs are trained individually on five datasets, and when using a set of weights: [1,0,0,0,0] for add_weighted_adapter to merge these adatpers, me expect the result to be the same of using the first adapter without averaging, which is how weighted averaging works, right? But actually, there is a huge gap between the performance of these two implementations, which should have been almost the same. 
A simple fix is to change both line 565 and 578 by adding the weight: 
target_lora_B.data += current_adapter_lora_B.data ---> target_lora_B.data += current_adapter_lora_B.data * weight
And it works. Assigning a set of weights of [1,0,0,0,0] for five adapters works nearly the same as merely using the first adapter now.
Thank you for your amazing work for the community, and the patience reading this.

jihuishan:
@BenjaminBossan Thank you for your response and patience in reproducing the results, and I'm sorry for the confusing way of referring the codes (me've never issued a question before.)
There may be an explanation about the problem. When merging LoRA adapters, we hope that: $W = W_0 + αB_1A_1 + βB_2A_2 + ……$
Therefore, it is understandable that it was implemented with no weights on B, because due to the formula above, simply putting a weight on A is enough. 

However, this will be a little bit different due to the way we merge adapters. Lets say the weights are $[1,0,0]$ for three adapters, then:
[...]

PR #1169, opened 2023-11-22
fix `add_weighted_adapter` method

pacman100:
### what does this PR do?
1. Fixes https://github.com/huggingface/peft/issues/1155


younesbelkada (approved):
Makes sense, thanks!

BenjaminBossan (approved):
Thanks for the quick fix.

I guess, this could be considered not backwards compatible, because the same code would produce different outcomes now. I think this is fine in this case, as "linear" is not the default and as the previous behavior could be considered a "bug". But we should be sure to highlight this change in the next release notes.

pacman100:
> But we should be sure to highlight this change in the next release notes.
Yes, noted.

PR #1177, opened 2023-11-23
Fix dockerfile build

younesbelkada:
As discussed offline cc @BenjaminBossan @muellerzr


younesbelkada:
I can confirm this leads to successfully building the docker image: https://github.com/huggingface/peft/actions/runs/6970869476/job/18969759804

BenjaminBossan (approved):
Thanks a lot for fixing the issue.

PR #1186, opened 2023-11-25
Added lora support for phi

umarbutler:
This pull request adds support for lora training [`microsoft/phi-1_5`](https://huggingface.co/microsoft/phi-1_5) by creating a new target modules mapping for models of the `model_type` `phi`.

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -507,6 +507,7 @@ TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING = {
     "codegen": ["qkv_proj"],
     "mistral": ["q_proj", "v_proj"],
     "stablelm": ["q_proj", "v_proj"],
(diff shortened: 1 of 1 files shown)


BenjaminBossan (approved):
Thanks, the change looks good. One question before merging: Do we want to update the list of supported models in the README.md as well? That should make it more discoverable.

Also, out of curiosity, does PEFT also work with `FusedDense` from flash_attn?

pacman100:
It should work with `FusedDense ` given that it inherits from `nn.Linear` https://github.com/Dao-AILab/flash-attention/blob/ce3e7280f81869fda78578e91e79e8d11763a3e3/flash_attn/ops/fused_dense.py#L139. Am I missing something, @BenjaminBossan?

BenjaminBossan:
> It should work with `FusedDense ` given that it inherits from `nn.Linear`

I see, thanks for the pointer. I agree it looks like it should just work.

PR #1187, opened 2023-11-25
Updating multi LoRA inference examples notebook to be usable

lucapericlp:
Right now, the base model isn't available.

[Related issue](https://github.com/tloen/alpaca-lora/issues/598) recognizing same thing, using the base model suggested there.


lucapericlp:
Closing because the rest of the notebook, `tloen/alpaca-lora-7b` uses an older verison of `peft` whose serialised config isn't compatible with current version of `peft`.

PR #936, opened 2023-09-15
[`CI`] Pin diffusers 

younesbelkada:
Fixes the current failing CI by pinning diffusers to before 0.21

The 0.21 version introduced a new way to load and run lora layers, and also introduced the `scale` parameter for forward pass. As this behaviour seems to break inference with PEFT, let's pin diffusers version and wait for the more global diffusers + peft integration coming soon 

cc @pacman100 @BenjaminBossan

PR #1189, opened 2023-11-27
ENH: Different initialization methods for LoRA

BenjaminBossan:
This PR adds the possibility to use different initialization methods for LoRA, as is a requirement for a completely backwards compatible adoption of PEFT in diffusers.

## Description

The default is still the same as always, namely the one from the reference implementation by Microsoft. On top of that, it is now possible to pass `init_lora_weights='gaussian'` to initialize the LoRA weights in the same way as is default for diffusers, namely with a normal distribution which is scaled by 1/r.

The init method currently applies to LoRA linear and conv layers, but not embedding layers, which are always initialized from a normal distribution (and are probably irrelevant for diffusers).

In the future, similar extensions could be added for other adapter methods.

## Notes

For testing, a rather simple test is added which calculates the Kolmogorov-Smirnov distance between the weights of a LoRA layer and the expected distribution. If someone has a better idea for a test, please let me know.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,9 @@ extras = {}
 extras["quality"] = ["black ~= 22.0", "ruff>=0.0.241", "urllib3<=2.0.0"]
 extras["docs_specific"] = ["hf-doc-builder"]
 extras["dev"] = extras["quality"] + extras["docs_specific"]
-extras["test"] = extras["dev"] + ["pytest", "pytest-cov", "pytest-xdist", "parameterized", "datasets", "diffusers<0.21.0"]
+extras["test"] = extras["dev"] + [
+    "pytest", "pytest-cov", "pytest-xdist", "parameterized", "datasets", "diffusers<0.21.0", "scipy"
+]
 
 setup(
     name="peft",
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 from dataclasses import dataclass, field
-from typing import List, Optional, Union
+from typing import Literal, List, Optional, Union
 
 from peft.config import PeftConfig
 from peft.utils import PeftType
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -18,7 +18,9 @@
 extras["quality"] = ["black ~= 22.0", "ruff>=0.0.241", "urllib3<=2.0.0"]
 extras["docs_specific"] = ["hf-doc-builder"]
 extras["dev"] = extras["quality"] + extras["docs_specific"]
-extras["test"] = extras["dev"] + ["pytest", "pytest-cov", "pytest-xdist", "parameterized", "datasets", "diffusers<0.21.0"]
+extras["test"] = extras["dev"] + [
+    "pytest", "pytest-cov", "pytest-xdist", "parameterized", "datasets", "diffusers<0.21.0", "scipy"
```
sayakpaul, line 22: Why's the version restriction needed on the `diffusers` side?

sayakpaul (approved):
Very nice! Particularly, the tests are very thorough! 

How should this be reflected in https://github.com/huggingface/diffusers/pull/5419?

setup.py
BenjaminBossan: It was pinned in #936 and can probably be unpinned now.

BenjaminBossan:
> How should this be reflected in [huggingface/diffusers#5419](https://github.com/huggingface/diffusers/pull/5419)?

Hmm, not sure about that PR specifically. In general, for diffusers to use this feature, every time that a `LoraConfig` is being created, `init_lora_weights="gaussian"` should be passed as an additional argument. On top

1. diffusers needs to depend on a PEFT version that contains that feature, or
2. add the argument in a backwards compatible way, as older PEFT versions would error out when encountering this unknown argument

sayakpaul:
Hmm the changes sound good to me. @patrickvonplaten WDYT?

pacman100 (approved):
Thank you @BenjaminBossan for adding support for different weight initialization methods for LoRA. The tests are really good, never thought that I would see statistical significance analysis in tests. LGTM! 🤩

pacman100:
> Before merging, should we wait until we have finalized the change required on diffusers side, in case we find that this PR doesn't quite cut it?

Makes sense

sayakpaul:
I can confirm that this works: https://wandb.ai/sayakpaul/dreambooth-lora/runs/bub8wjc3?workspace=user-sayakpaul

BenjaminBossan:
Great, thanks for checking it.

PR #1190, opened 2023-11-27
Extend merge_and_unload to offloaded models

### LINKED ISSUE #868: merge_and_unload issue?
### System Info

I am using the latest dev version of transformers, accelerate and peft (installed via !pip install -q -U git+<URL>) installed via Google Colab. 

This worked a few days ago, but now when I attempt to merge an adapter back into the base model and then save to hub, the size is much smaller than that of the base model and it can't be loaded (generates error "Cannot copy out of meta tensor; no data!" when I attempt to do so). 

The function I am using to merge the PeftModel back into the base model is:
**(Code begins here)**

```python
def merge_adapter(lora_id, merged_id):
    config = PeftConfig.from_pretrained(lora_id)

    model_id = config.base_model_name_or_path
        model_id,
        torch_dtype=dtype,
        device_map="auto",
        offload_folder="offload"
        )

    adapter = PeftModel.from_pretrained(
        model,
        lora_id,
        torch_dtype=dtype,
        device_map="auto",
        offload_folder="offload"
    )

    model = adapter.merge_and_unload(progressbar=True)
    tokenizer = AutoTokenizer.from_pretrained(model_id) #, use_fast=False)

    model.save_pretrained(
        merged_id,
        push_to_hub=True,
        repo_id=merged_id,
        private=True,
        #max_shard_size="4GB"
    #,
    )

    tokenizer.save_pretrained(
        merged_id,
        push_to_hub=True,
        repo_id=merged_id,
        private=True,
    )
```

The base model is 'meta-llama/Llama-2-7b-hf', which has two bin files, 1 @ 9.98GB and 1 @ 3.5GB. Previously when I ran this code, the merged model would be the same size. Now it only produces a single file @ 1.07GB. 

This may be an error with the library, although I am not seeing any bug reports to indicate this. It may also be an error with the training code in the first place, my upload code, the HF library or anything else. 

If anyone has any solutions, please let me know. Otherwise, if this is a bug, I guess this is my first bug report. 

### Who can help?

_No response_

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

This is the Colab I have been using. The data and LoRAs are private, but the two I am playing with are LLama 2 7B QLoRA fine-tuned on 1) a chat dataset and 2) a big hunk of raw text split into paragraphs (a piece of fanfiction). 

https://colab.research.google.com/drive/15g2NU2wJ9fOvY3PJCCN5dVDYV8KSXbeS?usp=sharing

### Expected behavior

The merged model should be created and be the same size as the base model, and I should be able to load it using AutoModelForCausalLM.from_pretrained as I was able to a few days ago.

blbadger:
This PR allows for the merging of offloaded LoRA modules.

Addresses issue #868 (`merge_and_unload` issue) and updates PR #1063, and is necessary for the test in PR #1167 to pass (that test is included here).

As noted in PR #1167,  model merging is currently unsuccessful. This PR addresses that issue by onloading each LoRA layer, merging with the base layer, re-making the base layer's offloaded parameter state dictionary before offloading the parameters accordingly.

@BenjaminBossan @pacman100 

Fixes #868


BenjaminBossan:
Thanks @blbadger for your continued work on this. This issue proves to be more complex than initially thought. The handling of hooks is very error prone and it's too easy to miss a place where it needs to be done. I wonder if we could maybe make some change on the accelerate side to simplify this. @SunMarc wanted to take a look too.

blbadger:
> Thanks @blbadger for your continued work on this. This issue proves to be more complex than initially thought. The handling of hooks is very error prone and it's too easy to miss a place where it needs to be done. I wonder if we could maybe make some change on the accelerate side to simplify this. @SunMarc wanted to take a look too.

No problem! I have tested on larger models as well and they pass too, with a little less-restricted `atol` margin of error on the `torch.allclose`. I am working on refactoring some of the code to reduce duplication, and certainly be open to suggestions to moving some to accelerate as well.

src/peft/tuners/lora/layer.py
```diff
@@ -244,8 +269,15 @@ def merge(self, safe_merge: bool = False, adapter_names: Optional[List[str]] = N
             adapter_names = self.active_adapters
 
         for active_adapter in adapter_names:
+
             if active_adapter in self.lora_A.keys():
                 base_layer = self.get_base_layer()
+                if (
```
BenjaminBossan, line 275: We have some repetition here of the same code, and it's also a bit distracting of the proper merging code. I think we can solve this by writing a `@contextmanager` that is basically

```python
@contextmanger
def some_name(base_layer):
    if (
        hasattr(base_layer, "_hf_hook")
        and isinstance(base_layer._hf_hook, AlignDevicesHook)
        and base_layer._hf_hook.offload
    ):
        base_layer._hf_hook.pre_forward(base_layer)

    yield

    if (
        hasattr(base_layer, "_hf_hook")
        and isinstance(base_layer._hf_hook, AlignDevicesHook)
        and base_layer._hf_hook.offload
    ):
        base_layer._hf_hook.weights_map = {
            name: param.to("cpu") for name, param in named_module_tensors(base_layer)
        }
        # offload weights map to disk if original device is the disk
        if torch.device("meta") in base_layer._hf_hook.original_devices.values():
            offload_folder = "offload_dir"
            offload_state_dict(offload_folder, base_layer._hf_hook.weights_map)
        base_layer._hf_hook.post_forward(base_layer, torch.tensor([]))
```

Then, here we can just use: `with some_name(base_layer) ...`. It leads to one more level of indentation, but I think that's a worthwhile trade-off.

src/peft/tuners/lora/layer.py
```diff
@@ -260,6 +292,20 @@ def merge(self, safe_merge: bool = False, adapter_names: Optional[List[str]] = N
...
+                    hasattr(base_layer, "_hf_hook")
+                    and isinstance(base_layer._hf_hook, AlignDevicesHook)
+                    and base_layer._hf_hook.offload
+                ):
+                    base_layer._hf_hook.weights_map = {
+                        name: param.to("cpu") for name, param in named_module_tensors(base_layer)
+                    }
+                    # offload weights map to disk if original device is the disk
+                    if torch.device("meta") in base_layer._hf_hook.original_devices.values():
+                        offload_folder = "offload_dir"
```
BenjaminBossan, line 306: Hmm, this creates a folder in the current working directory, right? This should be a temp folder instead.

BenjaminBossan (changes requested):
Thanks a lot @blbadger for your continued work on this. There is a merge conflict, could you please resolve it, it should be quick. I have a couple of comments, please check them out.

src/peft/tuners/lora/layer.py
blbadger: It probably should be, I have changed this to use a tempfile.TemporaryDirectory() instead.

src/peft/tuners/lora/layer.py
blbadger: Good idea! Implemented for all LoRA layers

src/peft/tuners/lora/layer.py
```diff
@@ -189,6 +191,29 @@ def unscale_layer(self, scale=None) -> None:
                 self.scaling[active_adapter] /= scale
 
 
+def onload_delta_wrapper(func):
+    """
+    Performs LoRA parameter onloading and offloading required for get_delta_weight() to function properly with
+    offloaded modules.
+    """
+
+    def wrapper(self, adapter):
```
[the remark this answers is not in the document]
blbadger: Hmm I was thinking of LoRA originally but maybe the approach would be more general. You are referring to the `BaseTunerLayer.adapter_layer_names` correct?  

One simple way to make a general version would be to activate the hooks for the parent and target in `_unload_and_optionally_merge()`, perform the merge, update the weight map, and offload again.  The problem is that doing so typically does not activate the sub-module LoRA A and B hooks (and modifying the offloaded weight map is tricky too), hence the current approach.  I will look into this some more, however.

src/peft/tuners/lora/layer.py
```diff
@@ -232,6 +236,53 @@ def unscale_layer(self, scale=None) -> None:
...
+
+    if (
+        hasattr(base_layer, "_hf_hook")
+        and isinstance(base_layer._hf_hook, AlignDevicesHook)
+        and base_layer._hf_hook.offload
+    ):
+        base_layer._hf_hook.weights_map = {name: param.to("cpu") for name, param in named_module_tensors(base_layer)}
+        # offload weights map to disk if original device is the disk
+        if torch.device("meta") in base_layer._hf_hook.original_devices.values():
+            offload_folder = tempfile.TemporaryDirectory()
```
BenjaminBossan, line 281: Can we use the `TemporaryDirectory` as a context manager, otherwise I think there is no guarantee that it will be cleaned up. Or is it necessary for it _not_ to be cleaned up right away?

src/peft/tuners/lora/layer.py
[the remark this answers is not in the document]
BenjaminBossan: > You are referring to the BaseTunerLayer.adapter_layer_names correct?

Correct.

> I will look into this some more, however.

Thanks, let me know if you figure something out or need help.

BenjaminBossan (commented):
Nice, I think we're approach a general solution that is at the same time not too distracting for the existing code. I left some comments.

src/peft/tuners/lora/layer.py
blbadger: On second thought I think it is indeed best to have a string for the directory here, as this is how the disk-offloaded directory is usually initially specified.  We do need the directory to be persistent, as the updated disk-offloaded weights are saved here. The original method (making a new folder) was I suppose not very efficient, so the updated way is to find the original offload directory and update that.

blbadger:
Thanks for your review @BenjaminBossan ! I have changed the disk-offloaded updates to what I think is the optimal approach with respect to memory: we look up the original offload folder and re-write there. 

As for a general approach for merging all tuner parameters, I realize now that `_unload_and_optionally_merge()` is specific for LoRA and that we would need to modify `BaseTuner.merge_adapter()` instead.  I have combined our context manager and wrapper into one context manager and applied this to `merge_adapter`, and this should work for any tuner with a base layer.  The test passes for the refactor too:)

However, it appears that the `_unload_and_optionally_merge()`  methods for each tuner type typically don't use `BaseTuner.merge_adapter()` so merging.  This would likely not be too difficult, but might be beyond the scope of this PR.

To sum up, we now have a merging operation that should work for any tuner and any device, with the following TODOs that I reckon are best as separate PRs:

1. incorporate `BaseTuner.merge_adapter()`  into each tuner unload and merge (should not be hard).
2.  Add compatibility for `PeftModel.from_pretrained` for disk-offloaded models.

Let me know what you think!

src/peft/tuners/tuners_utils.py
```diff
@@ -32,6 +35,47 @@
...
+        if name in ["", "base_layer"]:
+            continue
+        if hasattr(module, "_hf_hook") and isinstance(module._hf_hook, AlignDevicesHook) and module._hf_hook.offload:
+            module._hf_hook.pre_forward(module)
+
+    if hasattr(layer, "base_layer") and (
+        hasattr(layer.base_layer, "_hf_hook")
+        and isinstance(layer.base_layer._hf_hook, AlignDevicesHook)
+        and layer.base_layer._hf_hook.offload
+    ):
```
BenjaminBossan, line 50: Since this is the exact same check is below, how about storing this as a variable and re-using it below? The value cannot change in-between, right?

src/peft/tuners/tuners_utils.py
```diff
@@ -32,6 +35,47 @@
 logger = logging.getLogger(__name__)
 
 
+@contextmanager
+def onload_layer(layer):
```
BenjaminBossan, line 39: Could we have a short docstring that explains what is happening here.

src/peft/tuners/tuners_utils.py
```diff
@@ -32,6 +35,47 @@
...
+    for name, module in layer.named_modules():
```
BenjaminBossan, line 40: Since this is the same loop as before, how about storing these modules in a variable and re-using it below? So something along the lines of:

```python
modules_with_align_device_hooks = [m for name, m in layer.named_modules() if (name not in ["", "base_layer"]) and hasattr(module, "_hf_hook") and isinstance(module._hf_hook, AlignDevicesHook) and module._hf_hook.offload]
```

The long conditional could be factored out into a local function or something like that for readability.

src/peft/tuners/tuners_utils.py
```diff
@@ -32,6 +35,47 @@
...
+            continue
+        if hasattr(module, "_hf_hook") and isinstance(module._hf_hook, AlignDevicesHook) and module._hf_hook.offload:
+            module._hf_hook.post_forward(module, torch.tensor([]))
+
+    if hasattr(layer, "base_layer") and (
+        hasattr(layer.base_layer, "_hf_hook")
+        and isinstance(layer.base_layer._hf_hook, AlignDevicesHook)
+        and layer.base_layer._hf_hook.offload
+    ):
+        layer.base_layer._hf_hook.weights_map = {
```
BenjaminBossan, line 69: Could you please add a comment about why the device map should be CPU?

src/peft/tuners/lora/layer.py
```diff
@@ -235,7 +235,6 @@ def unscale_layer(self, scale=None) -> None:
 # Below code is based on https://github.com/microsoft/LoRA/blob/main/loralib/layers.py
 # and modified to work with PyTorch FSDP
 
-
```
BenjaminBossan, line 238: Could you please undo the changes in this file, since they're all cosmetic?

tests/test_gpu_examples.py
```diff
@@ -39,10 +40,13 @@
     WhisperTokenizer,
 )
 
+# A full testing suite that tests all the necessary features on GPU. The tests should
```
BenjaminBossan, line 43: Why move the comment here?

tests/test_gpu_examples.py
```diff
@@ -944,6 +944,51 @@ def test_causal_lm_training_multi_gpu(self):
...
+            model = AutoModelForCausalLM.from_pretrained(self.causal_lm_model_id, device_map=device_map).eval()
+            self.assertTrue(len({p.device for p in model.parameters()}) == 2)
+            model = PeftModel.from_pretrained(model, tmp_dir, max_memory=memory_limits)
+
+        input_tokens = tokenizer.encode("Four score and seven years ago", return_tensors="pt")
+        pre_merge_olayer = model(input_tokens)[0]
+        model = model.merge_and_unload()
+        post_merge_olayer = model(input_tokens)[0]
+
+        self.assertTrue(torch.allclose(pre_merge_olayer, post_merge_olayer))
```
BenjaminBossan, line 989: Unfortunately, this test fails for me:

```python
>>> torch.testing.assert_close(pre_merge_olayer, post_merge_olayer)
*** AssertionError: Tensor-likes are not close!

Mismatched elements: 301501 / 301542 (100.0%)
Greatest absolute difference: 2.2849273681640625 at index (0, 0, 11552) (up to 1e-05 allowed)
Greatest relative difference: 0.0675124004483223 at index (0, 0, 1862) (up to 1.3e-06 allowed)
```

Clearly, the values are highly correlated, but the difference is not negligible. Of course, we could set a higher tolerance to make the test pass, but I wonder if something fishy is going on here. When I run this completely on GPU, the assert passes.

When inspecting the differences, this is what I saw (sample of 0.1% of all values):

image

Do you have any idea what could be happening here?

blbadger: This is my fault, I pushed my test changes to the wrong branch (this was checking `model.merge_and_unload` which currently does not use `model.merge_adapter` and should not work). This has been corrected now!

tests/test_gpu_examples.py
blbadger: Not sure why this was moved, it has been returned now

src/peft/tuners/tuners_utils.py
blbadger: They should not change while merging, so a variable has been added in place of the second check

src/peft/tuners/tuners_utils.py
blbadger: Added a docstring

src/peft/tuners/tuners_utils.py
blbadger: Added comment

blbadger:
@BenjaminBossan no problem, thanks for all your help making this as tight as possible! I agree it is looking much better now.  I have made the changes you suggested and fixed the test. Thanks for the heads up on the deep speed context manager too, I will take a look at that as well.

src/peft/tuners/tuners_utils.py
```diff
@@ -32,6 +35,54 @@
 logger = logging.getLogger(__name__)
 
 
+@contextmanager
+def onload_layer(layer):
+    r"""
+    Moves a module's sub-modules to the execution device before context, after remakes the model state dict if
```
BenjaminBossan, line 41: I think the docstring could be clarified a bit. This is only for modules with offloaded weights, right? So maybe worth mentioning that this context is a no-op for weights without offloading.

"after" => "afterwards" or "after that"?
"remakes" => "restores"?

BenjaminBossan (changes requested):
Thanks, this looks super clean now.

I did some local testing and the test passed now, nice! As to why some tests are failing on CI, I think that's because you "stole" the `@require_torch_gpu` from the `LoftQTests` right below :)

I did some further testing and I think the new context manager should also be added to `unmerge_adapter` and to `LoraModel._unload_and_optionally_merge`. After I did that, the test will also pass for `unmerge_adapter` and `merge_and_unload`. If you agree, could you please make these additions and extend the test accordingly?

Eventually, the context manager could also be added to other adapters, but this could be done in a subsequent PR.

src/peft/tuners/tuners_utils.py
blbadger: I re-worded the docstring, hopefully that is better now.  Indeed this context manager does nothing if all the module's sub-modules are onloaded, although we call it on all tuner layers in case they have offloaded components.

blbadger:
@BenjaminBossan awesome, good idea to add compatibility with unmerging and merging with offloading. I have added that with the appropriate expansion on the tests:)

src/peft/tuners/lora/model.py
```diff
@@ -437,22 +437,21 @@ def _unload_and_optionally_merge(
...
                 continue
-
-            if hasattr(target, "base_layer"):
-                if merge:
-                    target.merge(safe_merge=safe_merge, adapter_names=adapter_names)
-                self._replace_module(parent, target_name, target.get_base_layer(), target)
-            elif isinstance(target, ModulesToSaveWrapper):
-                # save any additional trainable modules part of `modules_to_save`
-                setattr(parent, target_name, target.modules_to_save[target.active_adapter])
+            with onload_layer(module):
```
BenjaminBossan, line 447: I wonder: Would it be sufficient to use `with onload(target)`? I think that should work and it allows us to simplify the code a little bit.

BenjaminBossan (commented):
Thanks a lot. Just a small comment from my side, and could you please re-run `make style`? Otherwise, this LGTM.

src/peft/tuners/lora/model.py
blbadger: Agreed, and the code has been simplified accordingly

blbadger:
We should be good to go, only the `target` is offloaded now and `make style` has been run too.

BenjaminBossan (approved):
Thanks a ton for your patience and diligence. This LGTM now.

Could one of @pacman100 or @younesbelkada please also take a look?

Optional: If @muellerzr or @SunMarc could check the accelerate-side, it would be great.

younesbelkada (approved):
Looks pretty clean on my side ! Thanks for working on this @blbadger !

younesbelkada:
I suggest to merge it and leaving it on main for few days just to double check all is good on other integrations side

src/peft/tuners/tuners_utils.py
```diff
@@ -32,6 +35,56 @@
 logger = logging.getLogger(__name__)
 
 
+@contextmanager
+def onload_layer(layer):
+    r"""
+    A utility for modifying a module containing one or more tuners and a base layer, any of which are offloaded. Moves
+    a module's sub-modules to the execution device before some action is performed, after that that the state
```
SunMarc, line 42:
```suggestion
    a module's sub-modules to the execution device before some action is performed, after that the state
```

SunMarc (commented):
Thanks @blbadger for working on this ! Overall, the PR looks very good. I left a comment about disk offload.

src/peft/tuners/tuners_utils.py
```diff
@@ -32,6 +35,56 @@
...
+
+    if base_layer_offload:
+        # re-make weights map (must be on cpu to send params to the disk via memmap if disk offload)
+        layer.base_layer._hf_hook.weights_map = {
+            name: param.to("cpu") for name, param in named_module_tensors(layer.base_layer)
+        }
+        # offload weights map to disk if original device is the disk
+        if torch.device("meta") in layer.base_layer._hf_hook.original_devices.values():
+            # rewrite directory with merged weights
+            offload_state_dict(offload_folder, layer.base_layer._hf_hook.weights_map)
```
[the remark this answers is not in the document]
blbadger: It is true that the test only checks CPU offloading even though the context manager function is designed for both CPU and disk offloading. This is mostly because `PeftModel.from_pretrained` is currently incompatible with disk offloading (as far as I know) due the disk offload directory not being accessed from the base model properly.  I am happy to work on fixing this problem in the near future so the disk offloading test may be added, but thought it might be out-of-scope for this PR.  In the meantime I have added a todo comment for disk offloading in the test:)

The context manager and test description docstrings have been updated to reflect which offloading destinations each currently supports.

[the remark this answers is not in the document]
SunMarc: Thanks for clarifying this ! While we are unable to test the disk offload, the code looks good. So, I'm not against keeping it.

SunMarc (approved):
Thanks for iterating @blbadger !

blbadger:
My pleasure, and thank you very much for your reviews @younesbelkada @SunMarc @BenjaminBossan !

PR #1195, opened 2023-11-29
DOC: How to configure new transformers models

BenjaminBossan:
## Description

I believe that new transformers architectures could be the most common case of users wanting to apply PEFT on a model that is not supported out of the box. Thus I added a section specifically to help users configure their configs for new transformers models.

As I wanted to point users to a single file that contains all the existing transformers model mappings, I created a new file `src/peft/utils/constants.py`, which contains all the mappings that previously lived in `src/peft/utils/other.py`. LMK if that makes sense.

## Notes

To be absolutely backwards compatible, I re-imported the moved constants into `other.py`. This way, if there is code that imports them directly from there, it should continue to work.

To avoid getting a linter error for unused imports, I added those constants to the `__all__` list in `other.py`.


stevhliu (commented):
Nice doc! Some suggestions:

- I think maybe also explicitly mentioning this section is for setting up a `PeftConfig` (or any other config) for new transformers architectures will make it easier for users to immediately grasp. 
- A brief code snippet would also be cool to demonstrate how to setup the config. While there are examples shown above, some users may just skip to this last section, and it'd be a better experience if they didn't have to scroll back up to see how to set up the config.

BenjaminBossan:
Thanks for the feedback @stevhliu. I applied your suggestions, tried to clarify the context better, and gave a code example. Could you please check again?

docs/source/developer_guides/custom_models.md
```diff
@@ -199,3 +199,26 @@ peft_model.print_trainable_parameters()
...
+config = LoraConfig(
+    target_modules=["q_proj", "v_proj"],
+    ...,  # other LoRA arguments
+)
+peft_model = get_peft_model(my_mistral_model, config)
+```
+
+If that doesn't help, check the existing modules in your model architecture with the `named_modules` method and try to identify the attention layers, especially the key, query, and value layers. Those will often have names such as `c_attn`, `query`, `q_proj`, etc. The key layer is not always adapted, and ideally, you should check whether including it results in better performance.
+
+Additionally, linear layers are common targets to be adapted. Their names will often contain the strings `fc` or `dense`.
```
younesbelkada, line 222:
```suggestion
Additionally, linear layers are common targets to be adapted (e.g. in [QLoRA paper](https://arxiv.org/abs/2305.14314), authors suggest to adapt them as well). Their names will often contain the strings `fc` or `dense`.
```

younesbelkada (approved):
Great clean up ! Left one suggestion

PR #1196, opened 2023-11-29
[docs] API docs

stevhliu:
Expands and reorganizes the API docs to include other supported adapter methods, the `BaseTuner` class, and `Lycoris` classes.

### DIFF
--- a/docs/source/package_reference/tuners.md
+++ b/docs/source/package_reference/tuners.md
@@ -16,7 +16,7 @@ rendered properly in your Markdown viewer.
 
 # Tuners
 
-A tuner (or adapter) is a module that can be plugged into a `torch.nn.Module`. [`BaseTuner`] base class for other tuners and provides shared methods and attributes for preparing an adapter configuration and replacing a target module with the adapter module. [`BaseTunerLayer`] is a tuner mixin offering methods and attributes for managing adapters such as merging and unmerging, and activating and disabling adapters.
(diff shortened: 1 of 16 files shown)


BenjaminBossan (changes requested):
Thanks a lot Steven for extending the API docs and even writing a short summary for all methods. Apart from a typo, this LGTM.

I think we should also go through the code and add docstring for a couple of important methods. I'll try to do that today.

BenjaminBossan:
One more thing I noticed, there is an "UNDEFINED" section in the docs:

image

docs/source/package_reference/tuners.md
```diff
@@ -1,43 +1,27 @@
...
+rendered properly in your Markdown viewer.
 
-[[autodoc]] tuners.prefix_tuning.PrefixEncoder
+-->
 
-## Prompt tuning
+# Tuners
 
-[[autodoc]] tuners.prompt_tuning.PromptTuningConfig
+A tuner (or adapter) is a module that can be plugged into a `torch.nn.Module`. [`BaseTuner`] base class for other tuners and provides shared methods and attributes for preparing an adapter configuration and replacing a target module with the adapter module. [`BaseTunerLayer`] is a tuner mixin offering methods and attributes for managing adapters such as merging and unmerging, and activating and disabling adapters.
```
BenjaminBossan, line 19:
```suggestion
A tuner (or adapter) is a module that can be plugged into a `torch.nn.Module`. [`BaseTuner`] base class for other tuners and provides shared methods and attributes for preparing an adapter configuration and replacing a target module with the adapter module. [`BaseTunerLayer`] is a base class for adapter layers. It offers methods and attributes for managing adapters such as activating and disabling adapters.
```

`BaseTunerLayer` is not really a mixin, since it has abstract methods. Also, it doesn't implement (un)merging, these need to be implemented by child classes if the corresponding method supports it.

BenjaminBossan (approved):
Thanks a lot, LGTM.

PR #1200, opened 2023-11-30
Don't set config attribute on custom models

BenjaminBossan:
## Description

Initially, we had the issue that it was sometimes assumed that models had a `config` attribute, as is given for transformers models. This made PEFT fail with custom models, so we made a change to set a dummy config on those.

However, this can lead to issues down the line. For example, when users use the `Trainer` class from transformers, they can stumble upon lines like this:

https://github.com/huggingface/transformers/blob/62ab32b2997846526cdffd629c72c47bc7b4f215/src/transformers/integrations/integration_utils.py#L636-L637

https://github.com/huggingface/transformers/blob/62ab32b2997846526cdffd629c72c47bc7b4f215/src/transformers/integrations/integration_utils.py#L729-L730

Here transformers assumes that if `config` attribute exists on the model, it must have a `to_json_string` method or a `to_dict` method (as it assumes the config to be a `PretrainedConfig` instance). Therefore, in order not to trip up transformers, it is best not to set any config at all.

## Note

I _think_ this should be fully backwards compatible, as the `config` attribute should not affect checkpoints. LMK if I'm missing something.

## Other solutions

1. Alternatively, transformers could be changed to check each time when the config attributes exists, if it is a `PretrainedConfig` instance, but that would be a much larger change (albeit a cleaner one).
2. Instead of creating a `dict` type `config` on custom models, we could create a `PretrainedConfig`. This would probably require setting a couple of fields correctly, as the config should not be empty.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -121,7 +121,6 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
             self.base_model = cls(model, {adapter_name: peft_config}, adapter_name)
             self.set_additional_trainable_modules(peft_config, adapter_name)
 
-        self.config = getattr(self.base_model, "config", {"model_type": "custom"})
         if getattr(model, "is_gradient_checkpointing", True):
             model = self._prepare_model_for_gradient_checkpointing(model)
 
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
Thanks for the review. I would suggest to merge this after the upcoming PEFT release, just in case it causes some issues I haven't thought of.

PR #1204, opened 2023-12-01
ENH: Enable OFT adapter for mixed adapter models

BenjaminBossan:
This PR makes it possible to use the newly added OFT adapter in mixed adapter type models, similar to LoRA, LoHa, etc.

## Notes

Adding the integration was pretty straightforward, which is a good sign.

The difficult part was actually about the tests. This stems from the fact that OFT is (if my understanding is correct) never commutative. What I mean is that even if the adapters are applied to the last layer of a model, it makes a difference whether we apply, say, first LoRA, then OFT vs first OFT, then LoRA.

This is different for the other adapters that were added so far for mixed models, as they basically do:

- `Xa = X + dXa`
- `Xab = Xa + dXb = X + dXa + dXb = X + dXb + dXa = Xb + dXa = Xba`

IIUC, this is not true for OFT, so when OFT is used, I had to ensure that no test was applied that (implicitly) assumes commutativity. Ping @okotaku and @lukaskuhn-lku is my understanding correct?

Furthermore, I had to increase the model size, see this comment:

https://github.com/huggingface/peft/pull/1160#issuecomment-1836107235

### DIFF
--- a/src/peft/mixed_model.py
+++ b/src/peft/mixed_model.py
@@ -35,6 +35,7 @@ from .tuners import (
     LoKrModel,
     LoraModel,
     MixedModel,
+    OFTModel,
 )
 from .utils import PeftType, _set_adapter, _set_trainable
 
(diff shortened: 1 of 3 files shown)


lukaskuhn-lku:
As far as my understanding of OFT goes your assumption about commutativity are correct

okotaku:
@BenjaminBossan It's correct.

Issue #1198, opened 2023-11-30
Potentially a bug


ngocbh:
https://github.com/huggingface/peft/blob/2b901ee57230559aaf39867c7698f6aca3617162/src/peft/tuners/tuners_utils.py#L260C32-L260C32

inject_adapters have its own `model` argument but, in line 260 and 263, it accesses model parameters using `self.model`. I guess it's not a safe choice.

BenjaminBossan:
Thanks for bringing this up. The `model` and `self.model` are indeed the same object (check `model is self.model`), so the code works correctly when used as intended. I agree, however, that this is confusing and it would be better if `model` was used throughout. If you want to create a PR to fix that, we'd be happy to merge it.

PR #1205, opened 2023-12-01
Fix model argument issue (#1198)

ngocbh:
A quick fix to issue #1198 that makes sure one model is used throughout `inject_adapter` function in `BaseTuner`.


ngocbh:
The reason I want to `_mark_only_adapters_as_trainable` can also work without passing the model argument is the `BaseTuner` is already a model. So if we want to use this function outside the class we might have to call `model._mark_only_adapters_as_trainable(model.model)`. I think this is the main use case of this function outside the class. But calling like this seems a little bit weird to me. That's why I let this function work without passing the model argument. What do you think? 

If you still prefer to make `model` a required argument, I will change it back.

BenjaminBossan:
> I think this is the main use case of this function outside the class.

I'm not quite sure I understand this. Do you mean that PEFT users may want to call `model._mark_only_adapters_as_trainable`? If that is so, I think we should add an explicit method/function for that, as `_mark_only_adapters_as_trainable` is supposed to be "private" and thus not called directly in user code.

ngocbh:
Sorry for the delay, I fixed the issue according to your suggestions.

BenjaminBossan (approved):
Looks good, thanks.

PR #1207, opened 2023-12-04
remove HF tokens

yxli2123:
Hi, I just found I leaked my personal HF tokens. I just removed and will deactivate it on my side.


yxli2123:
No, there's only one defined but not used.

PR #1208, opened 2023-12-04
TST: Add tests for 4bit LoftQ

BenjaminBossan:
Add GPU tests for LoftQ with 4bit quantization.

## Notes

Tests for 8bit quantization are already there but not run at the moment, see this comment:

https://github.com/huggingface/peft/pull/1150#issuecomment-1838891499

In my testing, 8bit passes when using `NFQuantizer`, so if the original author is fine with using that, I can make the adjustment.


BenjaminBossan:
@yxli2123 If you could also do a quick review, that would be great.

I ran the GPU tests locally and they passed, so :crossed_fingers: that they also pass on the nightly CI (PR CI does not run GPU tests).

tests/test_gpu_examples.py
```diff
@@ -933,3 +935,138 @@ def test_causal_lm_training_mutli_gpu(self):
...
             # assert loss is not None
             self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
+
+
+@require_torch_gpu
+class LoftQTests(unittest.TestCase):
+    r"""
+    Tests for LoftQ
+    """
+
```
younesbelkada, line 945:
```suggestion
    error_factor = 3
```

tests/test_gpu_examples.py
```diff
@@ -933,3 +935,138 @@ def test_causal_lm_training_mutli_gpu(self):
...
+        # first, sanity check that all errors are > 0.0
+        self.assertTrue(mae_quantized > 0.0)
+        self.assertTrue(mse_quantized > 0.0)
+        self.assertTrue(mae_loftq > 0.0)
+        self.assertTrue(mse_loftq > 0.0)
+
+        # next, check that LoftQ quantization errors are smaller than LoRA errors by a certain margin
+        factor = 3
+        self.assertTrue(mae_loftq < mae_quantized / factor)
+        self.assertTrue(mse_loftq < mse_quantized / factor)
```
younesbelkada, line 1032:
```suggestion
        self.assertTrue(mae_loftq < mae_quantized / self.error_factor)
        self.assertTrue(mse_loftq < mse_quantized / self.error_factor)
```

tests/test_gpu_examples.py
```diff
@@ -933,3 +935,138 @@ def test_causal_lm_training_mutli_gpu(self):
...
```
younesbelkada, line 1052:
```suggestion
        self.assertTrue(mae_loftq < mae_quantized / self.error_factor)
        self.assertTrue(mse_loftq < mse_quantized / self.error_factor)
```

tests/test_gpu_examples.py
```diff
@@ -933,3 +935,138 @@ def test_causal_lm_training_mutli_gpu(self):
...
```
younesbelkada, line 1072:
```suggestion
        self.assertTrue(mae_loftq < mae_quantized / self.error_factor)
        self.assertTrue(mse_loftq < mse_quantized / self.error_factor)
```

younesbelkada (approved):
Thanks a lot for adding the test ! I left one suggestion so that we can easily play with the error tolerance in the future in case the test fail on the runner - what do you think?

PR #1210, opened 2023-12-04
Fix an issue with layer merging for LoHa and OFT

### LINKED ISSUE #1194: Figure out failing stable diffusion tests
### System Info

Tests may only fail on Windows, only with CPU, or only for specific PyTorch versions.

### Who can help?

Anyone.

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

Currently, the merging test for LoHa is being skipped:

https://github.com/huggingface/peft/blob/f0fb9516d8ebc4fed1bacaa11b789be023f003f8/tests/test_stablediffusion.py#L122-L124

Testing on Linux, the test passes with GPU and PyTorch 2.1, but fails on CPU.

Furthermore, the same test causes issues for OFT as well when using CPU:

https://github.com/huggingface/peft/pull/1160#issuecomment-1829890240

### Expected behavior

Merging should work on CPU for LoHa and OFT.

lukaskuhn-lku:
This fixes #1194 partly, as the issue only arises while merging the text-encoder. This excludes the merging of the text-encoder for LoHaConfigs and OFTConfigs, until a full fix is found, but still runs the merging of the unet which is important for test coverage in my opinion.


BenjaminBossan:
Thanks for doing further testing. Let's also adjust the comment, since I don't think it was correct to begin with: The test is not flaky, it fails consistently; it's not only Windows, I can reproduce on Linux; but for me, it's only on CPU, not GPU.

I also did some further testing and it looks like this is not related to merging. E.g. this test fails for me for all adapters:

```python
    @parameterized.expand(
        PeftStableDiffusionTestConfigManager.get_grid_parameters(
            {
                "model_ids": PEFT_DIFFUSERS_SD_MODELS_TO_TEST,
                "lora_kwargs": {"init_lora_weights": [False]},
                "loha_kwargs": {"init_weights": [False]},
                "lokr_kwargs": {"init_weights": [False]},
                "oft_kwargs": {"init_weights": [False]},
            },
        )
    )
    def test_foobar(self, test_name, model_id, config_cls, config_kwargs):
        model = self.instantiate_sd_peft(model_id, config_cls, config_kwargs)
        dummy_input = self.prepare_inputs_for_testing()

        # on GPU
        with temp_seed(seed=42):
            output = np.asarray(model(**dummy_input).images[0])

        # on GPU, again
        with temp_seed(seed=42):
            output2 = np.asarray(model(**dummy_input).images[0])

        # on CPU
        with temp_seed(seed=42):
            cpu_output = np.asarray(model.to("cpu")(**dummy_input).images[0])

        # sanity check: outputs from GPU are the same (i.e. not an RNG issue)
        np.testing.assert_allclose(output, output2)         # PASSES
        # check that outputs from GPU and CPU are the same
        np.testing.assert_allclose(output, cpu_output)      # FAILS
```

`state_dict`s for unet and text encoder are identical but outputs are different, quite strange. I also checked the outputs from the SD model without PEFT adapters, and there everything looks fine.

lukaskuhn-lku:
Exactly what I found as well, the state dicts are the same but the output is different on the CPU. Weird. I will fix the comment

lukaskuhn-lku:
@BenjaminBossan I changed the comment

BenjaminBossan (approved):
Thanks for improving the test coverage, LGTM!

Issue #1093, opened 2023-11-08
Possible issue with prompt tuning with left padding


SumanthRH:
### System Info

peft version 0.7.0

### Who can help?

@pacman100 @BenjaminBossan 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

I believe the current implementation for prompt tuning is suboptimal for causal language models, because of left padding. In the original paper from [Lester _et al_](https://arxiv.org/abs/2104.08691), prompt tuning is proposed as a method to prepend prompt tokens $P$ to an input $X$ to get the best performance for the task of generating $Y$. Note that the paper only implemented prompt tuning for seq2seq models, specifically for T5. Now with causal language models, you typically have a decoder-only architecture and thus left padding. For example, for the example notebook on prompt tuning for BLOOM (`examples/causal_language_modeling/peft_prompt_tuning_clm.ipynb`), here is how one input looks like: 

`<pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad>Tweet text : Couples wallpaper, so cute. :) #BrothersAtHome Label : no complaint</s>`

The current prompt tuning code simply prepends prompt embeddings to the input sequence after padding, so this is equivalent to having something like: 

`Prompt<pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad>Query`

This seems very sub-optimal. Also, this leads to a different kind of prompt conditioning for training time and test time inputs, so it might even be wrong. I was wondering if this has been investigated. I'm happy to take up benchmarking to see what performance you might get if you can somehow prepend the prompt embeddings after padding. I believe PEFT should atleast add a warning right now that prompt tuning with causal LM is not the same as prompt tuning with seq2seq. 

### Expected behavior

Prompt tokens should be prepended right before input query, or PEFT should atleast provide a warning that prompt tuning for causal LMs is suboptimal with padding.

BenjaminBossan:
[...]

PR #1212, opened 2023-12-05
Fix prompt tuning with left padding

SumanthRH:
# What does this PR do?

This is a long pending followup to #1093 . This is meant to be a draft PR, and I'm hoping for comments from maintainers on the best way forward. 

To summarize, the current `PeftModelForCausalLM` code has different behaviour from `PeftModelForSeq2SeqLM` and the original prompt tuning implementation. The prompt embeddings are meant to be placed right before the input query, but with left padding, you end up with something like 
`Prompt<pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad>Query`

The original prompt tuning method was proposed with T5, a seq2seq model, for which this issue isn't there. 

It is very messy to deal with this at the model stage! An ad-hoc fix is to detect the boundary between padding and non-padding tokens via the attention mask, and then split up the input batch of tensors into two. The prompt embeddings get added right before the input embeddings without padding. We can finally add the prompt tokens/embeddings back. I've had to break everything down into lists of tensors, because padding is going to be non-uniform. All the current tests pass for this implementation (more should be added if we decide to use this).

Another solution is to simply add a warning for the user for CausalLM + prompt tuning (this is, not great, because warnings usually get lost when your stack is PEFT + Accelerate + Transformers + PyTorch + NCCL). 

I am yet to do a proper benchmark for performance difference b/w this corrected version and current PEFT implementation. 
( When I ran the example notebook for GPT-2 on Twitter Complaints, there was almost no difference in performance, but then the Twitter Complaints dataset only has 50 train examples)

cc @BenjaminBossan @pacman100


SumanthRH:
No issues at all!

src/peft/utils/other.py
```diff
@@ -481,3 +481,80 @@ def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
...
+      containing the padding token elements
+    """
+    if input_ids is None and inputs_embeds is None:
+        raise ValueError("You have to provide either input_ids or inputs_embeds")
+    ndim = input_ids.ndim if input_ids is not None else inputs_embeds.ndim - 1
+    if not ndim == 2:
+        raise ValueError("input_ids must be a 2D tensor")
+    # find the first index of a non-padding token
+    if attention_mask is not None:
+        first_non_padding_tok_ind = torch.argmax(attention_mask, dim=1)
```
BenjaminBossan, line 505: Clever. I guess it can theoretically fail if we have a row with all zeros (would return 0 for that row), but that should not happen, right? Maybe an alternative is to sum over dim=1 and use the sum as index from the end.

src/peft/utils/other.py
```diff
@@ -481,3 +481,80 @@ def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
...
+    else:
+        batch_size = _get_batch_size(input_ids, inputs_embeds)
+        first_non_padding_tok_ind = torch.zeros(batch_size, dtype=torch.long)
```
BenjaminBossan, line 508: Could you please explain why we can make this assumption? Also, does this mean that if `attention_mask` is `None`, all the special handling could be skipped?

src/peft/utils/other.py
```diff
@@ -481,3 +481,80 @@ def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
...
+
+
+def add_pad_tokens(
+    input_embeds: List[torch.Tensor],
+    attention_mask: List[torch.Tensor],
+    labels: List[torch.Tensor],
+    pad_els: Tuple[List[torch.Tensor], List[torch.Tensor], List[torch.Tensor]],
+):
+    """
+    Adds the padding token elements back to input_ids, attention_mask and labels Returns:
```
BenjaminBossan, line 542:
```suggestion
    Adds the padding token elements back to input_ids, attention_mask and labels

    Returns:
```

BenjaminBossan (changes requested):
Thanks a lot for working on this tricky issue. I have to admit that parts of the implementation are not quite clear to me, but this is more due to the way PEFT handles the different adapter types than due to this PR.

Overall, this feels like it would better be handled outside of PEFT, i.e. when creating the tokens, not after. But then it would require extra work on the user side, which can easily be forgotten, leading to the problem you mentioned. Therefore, I see no better option as what is here.

For this PR, I think it's important to have unit tests for both `separate_pad_tokens` and `add_pad_tokens`, to ensure that all corner cases are taken care off. Would you be willing to add those?

> I am yet to do a proper benchmark for performance difference b/w this corrected version and current PEFT implementation.
> ( When I ran the example notebook for GPT-2 on Twitter Complaints, there was almost no difference in performance, but then the Twitter Complaints dataset only has 50 train examples)

Let us know if you have more results. I think some of the disadvantage of the current implementation could be softened if during inference, the inputs are padded in the same fashion with the gap in the middle.

src/peft/utils/other.py
SumanthRH: This logic is again because this fix is at the model stage. If `attention_mask` is `None`, then, without knowing the padding token id, there's no way to segregate different input tokens in `input_ids`. So the assumption is that there is no padding done. One more alternative to this change is to change the API itself for prompt tuning: ask the user to always pass `pad_token_id`, and raise an error if not. This is a breaking change though so I didn't go for this.

The special handling can indeed be completely skipped if `attention_mask` is `None`. My issue is that it makes the types more unclear that it is already. Currently, if it's prompt tuning, then all the input tensors are internally passed around as `List[torch.Tensor]` until the final `model.forward` call. If I skip the special handling completely, then it will sometimes be `List[torch.Tensor]` and sometimes `torch.Tensor` internally. Let me see what I can do.

src/peft/utils/other.py
SumanthRH: Yes many cornes cases left to handle! This is an interesting one: If the attention_mask is all zeros, it's currently a valid input for HF models. I'm trying to think what this means. This is surely a malformed input, because the attention weights become zeroed out in every transformer block. I think for now I'll have consistency and change this code to make `first_non_padding_tok_ind` to be `attention_mask.shape[1]` i.e length of the sequence. I can do this simply by detecting if the `first_non_padding_tok_ind` has a zero entry. For now, I'll also enforce that input `attention_mask` should only have 0 and 1 entries, with a check in `separate_pad_tokens`. AFAIK `transformers` allows all kinds of values but ignores it internally. 

Also, sum over dim=1 and using that as the index from the end can break sometimes - there can be a corner case where there are padding tokens added at the right end by the user and we end up ignoring some tokens.

BenjaminBossan: Thanks for the detailed analysis. It is unfortunate that we have to rely on what comes down to implementation details of transformers, but I don't see a good way to make progress without making some simplifying assumptions.

> Also, sum over dim=1 and using that as the index from the end can break sometimes - there can be a corner case where there are padding tokens added at the right end by the user and we end up ignoring some tokens.

Ah yes, I didn't think of that one.

In general, I think similar kinds of problems come up again and again, but there really doesn't seem to be any solution to simplify it. PyTorch is working on adding `nested_tensor`, which seems to go in this direction, but unfortunately this feature is experimental and also would currently not support what we need here.

src/peft/utils/other.py
BenjaminBossan: > without knowing the padding token id, there's no way to segregate different input tokens in `input_ids`.

Okay, so this is basically saying: If there is no attention mask, we can't do anything, so we throw our hands in the air :) And not necessarily that this is always the correct behavior.

> This is a breaking change though so I didn't go for this.

Yeah, that's a tricky problem, as we want to indeed avoid breaking existing code. If the existing code were doing a totally incorrect thing, with the result being nonsense, I could see an argument for raising an error. But as you mentioned earlier, it's not a given that the result would be nonsense.

> The special handling can indeed be completely skipped if `attention_mask` is `None`. My issue is that it makes the types more unclear that it is already.

I see the argument. Maybe if we could do something like returning early on the corresponding `PeftModelForCausalLM` methods (as we do for non-prompt-learning methods), we could avoid having a bunch of if-else for list of tensors vs tensors. WDYT?

SumanthRH: Yes that sounds good! I'll refactor that way.

SumanthRH:
TODO:

- [ ] Bypass special handling with `attention_mask` is `None`.
- [ ] Add unit tests and asserts for `separate_pad_tokens` and `add_pad_tokens`.
- [ ] Performance comparison for current prompt tuning impl. and the corrected prompt tuning impl.

BenjaminBossan:
@SumanthRH LMK once this is ready for the next review. Also, please don't forget to run `make style` :)

SumanthRH:
Hi @BenjaminBossan , finally got some numbers! TLDR is again that performance differences on small datasets isn't much (and even the speed differences weren't much). 

I've benchmarked on subsets of the ELI5 and the OpenOrca dataset.

I used GPT2 on ELI5 (hoping this isn't in the training set). Eval numbers with the exact same hyperparameters on the current and the new, fixed implementation are below:

| Implementation    | train_ppl | train_epoch_loss | eval_ppl | eval_epoch_loss |
|--------------------|-----------|------------------|----------|-----------------|
| Old Implementation | 47.9587   | 3.8703           | 39.1261  | 3.6668          |
| New Implementation | 47.8425   | 3.8679           | 39.1270  | 3.6668          |

Not very significant imo. I thought maybe in this case the task was also not hard, so I benchmarked on a harder dataset: OpenOrca. 


I switched to the openlm-research/open_llama_3b model, as this is a harder task overall. Results after finetuning on  a 10K subset: 

| Implementation    | train_ppl | train_epoch_loss | eval_ppl | eval_epoch_loss |
|--------------------|-----------|------------------|----------|-----------------|
| Old Implementation | 3.4938    | 1.2510           | 3.4668   | 1.2432          |
| New Implementation | 3.4785    | 1.2466           | 3.4382   | 1.2349          |

Can see a second decimal point difference now with `eval_ppl` 🙆🏻‍♂️

I believe more benchmarking on larger datasets will make this more significant. From a research perspective, this improvement should still be useful

## Slowdown
I added time.time() statements for the input preprocessing in `PeftModelForCausalLM` . This is the time to make the new input_embeds, attention_mask and labels, right from [this start](https://github.com/huggingface/peft/blob/799420aef18146cffe98db7eeb3e4b34dc566110/src/peft/peft_model.py#L1083) to [this end](https://github.com/huggingface/peft/blob/799420aef18146cffe98db7eeb3e4b34dc566110/src/peft/peft_model.py#L1120) line number. I got the time deltas for the Orca dataset with a batch size of 32 (max sequence length of 512). Tests done on an RTX 2080 Ti. Taking the average over 40 iterations, I see:
Old implementation:  0.00061834 s ( 0.6 ms)
New implementation: 0.00955 s (9 ms)

There's a 15x slowdown in the preprocessing step, but it is still in the milliseconds even with a batch size of 32! From my perspective, this slowdown can be tolerated, but I'm curious what you think.

src/peft/peft_model.py
```diff
@@ -1108,14 +1122,35 @@ def forward(
...
-                prefix_labels = torch.full((batch_size, peft_config.num_virtual_tokens), -100).to(labels.device)
-                kwargs["labels"] = torch.cat((prefix_labels, labels), dim=1)
-            prompts = self.get_prompt(batch_size=batch_size, task_ids=task_ids)
-            prompts = prompts.to(inputs_embeds.dtype)
-            inputs_embeds = torch.cat((prompts, inputs_embeds), dim=1)
+                # index into first element for device info since labels is a list for prompt tuning
+                prefix_labels = torch.full((batch_size, peft_config.num_virtual_tokens), -100).to(labels[0].device)
+                labels = batch_concatenate_with_list_or_tensor(prefix_labels, labels)  # prepend prefix labels
+
+            if pad_els is not None:  # special handling
```
BenjaminBossan, line 1138: Could you please add a comment here about why this whole block is needed?

src/peft/utils/other.py
```diff
@@ -481,3 +481,116 @@ def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
...
+      containing the padding token elements
+    """
+    if input_ids is None and inputs_embeds is None:
+        raise ValueError("You have to provide either input_ids or inputs_embeds")
+    ndim = input_ids.ndim if input_ids is not None else inputs_embeds.ndim - 1
+    if not ndim == 2:
+        raise ValueError("input_ids must be a 2D tensor/ input_embeds must be a 3D tensor")
+
+    if attention_mask is None:
+        return input_ids, inputs_embeds, attention_mask, labels, None  # no special handling needed
```
BenjaminBossan, line 511: > # no special handling needed

Isn't it more like no special handling is possible without the attention mask?

src/peft/utils/other.py
```diff
@@ -481,3 +481,116 @@ def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
...
+    if input_ids is not None:
+        input_id_paddings = [input_ids[i][:first_ind] for i, first_ind in enumerate(first_non_padding_tok_ind)]
+        input_ids = [input_ids[i][first_ind:] for i, first_ind in enumerate(first_non_padding_tok_ind)]
+    else:
+        input_embed_paddings = [inputs_embeds[i][:first_ind] for i, first_ind in enumerate(first_non_padding_tok_ind)]
+        inputs_embeds = [inputs_embeds[i][first_ind:] for i, first_ind in enumerate(first_non_padding_tok_ind)]
+    # get the same slices for attention_mask and labels. Note that all the padding vectors are of different shapes
+    # so we can't stack them into a tensor
+    attention_mask_paddings, labels_paddings = None, None
+    if attention_mask is not None:
```
BenjaminBossan, line 530: Is this check necessary, given that it's already checked earlier?

src/peft/utils/other.py
```diff
@@ -481,3 +481,116 @@ def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
...
+    """
+    if pad_els is None:
+        return inputs_embeds, attention_mask, labels  # no special handling needed
+
+    pad_input_ids, pad_inputs_embeds, pad_mask, pad_labels = pad_els
+    final_inputs_embeds = torch.stack(
+        [torch.cat((pad_embed, input_embed)) for input_embed, pad_embed in zip(inputs_embeds, pad_inputs_embeds)]
+    )
+    final_attention_mask, final_labels = None, None
+    if attention_mask is not None:
```
BenjaminBossan, line 567: Can the attention mask really be `None` here? If yes, the type annotation needs to reflect that.

src/peft/utils/other.py
```diff
@@ -481,3 +481,116 @@ def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
...
+        final_attention_mask = torch.stack(
+            [torch.cat((pad_mask, attention_mask)) for attention_mask, pad_mask in zip(attention_mask, pad_mask)]
+        )
+    if labels is not None:
+        final_labels = torch.stack([torch.cat((pad_label, label)) for label, pad_label in zip(labels, pad_labels)])
+    return final_inputs_embeds, final_attention_mask, final_labels
+
+
+def apply_to_list_or_tensor(f, x: Union[List[torch.Tensor], torch.Tensor]) -> Union[List[torch.Tensor], torch.Tensor]:
```
BenjaminBossan, line 576: Nit: The "more correct" type annotation should be:

```python
from typing import overload

@overload
def apply_to_list_or_tensor(f, x: List[torch.Tensor]) -> List[torch.Tensor]:
   ...

def apply_to_list_or_tensor(f, x: torch.Tensor) -> torch.Tensor:
   # actual code
```

But we're not really strict with type annotations, so it's okay :)

src/peft/utils/other.py
```diff
@@ -481,3 +481,116 @@ def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
...
+    """
+    if isinstance(x, list):
+        return [f(inp) for inp in x]
+    else:
+        return f(x)
+
+
+def batch_concatenate_with_list_or_tensor(
+    tensor: torch.Tensor, inp: Union[List[torch.Tensor], torch.Tensor]
+) -> Union[List[torch.Tensor], torch.Tensor]:
```
BenjaminBossan, line 589: Same nit about the type annotation.

BenjaminBossan (commented):
Thanks so much for running the tests. Honestly, I'm surprised that there is basically no difference in scores, it makes me wonder if it could point to an error somewhere. Did you double-check with these experiments that the inputs look correct? Then again, maybe the model just learns not to attend to the gaps, which explains the parity...

I agree that despite the big relative difference in processing time, the absolute number should not matter much when it comes to total training time (I assume the number you reported is per batch). Maybe there is room for speed ups but this shouldn't be a blocker.

The bigger cost IMO is the cost of maintaining the more complex code. I'm really on the fence when it comes to this, as the code is significantly harder to maintain -- even if correct in theory, there appears to be no impact in practice. Do you happen to know if there are other implementations out there that use this correct approach for causal LMs?

SumanthRH:
Not stale

pacman100 (commented):
Hello, I've an overall comment. When padding happens on the left hand side, the attention mask is supplied with 0s at the beginning to inform the model to ignore these during self-attention. So, the only positions being attended are the soft prompt tokens and the valid input tokens. Maybe, that can be the reason for minimal performance change between this PR with correct implementation and the current codebase.

pacman100:
> The bigger cost IMO is the cost of maintaining the more complex code. I'm really on the fence when it comes to this, as the code is significantly harder to maintain -- even if correct in theory, there appears to be no impact in practice.

I also feel the same.

PR #1218, opened 2023-12-05
DOC: Initialization options for LoRA

BenjaminBossan:
Document the initialization options for LoRA. This is especially important for LoftQ, since otherwise, it may not obvious for users how to make use of it.


MKhalusova (approved):
Very well written and informative, thank you for adding this!

younesbelkada (approved):
Great, thanks!

Issue #559, opened 2023-06-09
just import of peft leads to use of cuda and introduces cuda context that makes forks not possible


pseudotensor:
Older peft did not do this, but newer does, at least since 3714aa2fff158fdfa637b2b65952580801d890b2 and likely earlier.

This means if import in global scope as normal, no forks are possible anymore in python, ruining multiprocessing etc.

The use of cuda should be lazy on-demand and not forced when importing peft.  That is, cuda should only be introduced when model itself is on cuda, not just from an import of peft.

```
jon@pseudotensor:~/h2ogpt$ python
Python 3.10.11 (main, Apr 20 2023, 19:02:41) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from peft import PeftModel

===================================BUG REPORT===================================
Welcome to bitsandbytes. For bug reports, please run

python -m bitsandbytes

 and submit this information together with your error trace to: https://github.com/TimDettmers/bitsandbytes/issues
================================================================================
bin /home/jon/miniconda3/envs/h2ollm/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda121.so
/home/jon/miniconda3/envs/h2ollm/lib/python3.10/site-packages/bitsandbytes/cuda_setup/main.py:149: UserWarning: WARNING: The following directories listed in your path were found to be non-existent: {PosixPath('/usr/lib/jvm/default-java/jre/lib/amd64/server'), PosixPath('/opt/clang+llvm-4.0.0-x86_64-linux-gnu-ubuntu-16.04/lib'), PosixPath('/home/jon/lib'), PosixPath('/opt/rstudio-1.0.136/bin'), PosixPath('/usr/local/cuda/extras/CUPTI/lib64')}
  warn(msg)
CUDA SETUP: CUDA runtime path found: /usr/local/cuda/lib64/libcudart.so
CUDA SETUP: Highest compute capability among GPUs detected: 7.5
CUDA SETUP: Detected CUDA version 121
CUDA SETUP: Loading binary /home/jon/miniconda3/envs/h2ollm/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda121.so...
>>> 
```

Because of this, after such an import, things like this work:

```
from concurrent.futures import ProcessPoolExecutor


def go():
    from transformers import AutoModelForCausalLM
[...]

PR #1230, opened 2023-12-07
Lazy import of bitsandbytes

BenjaminBossan:
Should fix various issues reported here and elsewhere, e.g. #559, https://github.com/huggingface/accelerate/issues/2216.

## Description

Previously, we imported from bitsandbytes eagerly if the package was installed. This caused two major issues:

- Slow loading time of PEFT (~4 sec) when bnb is installed
- Errors with multiprocessing because bnb initializes CUDA -- users had to load PEFT lazily or else would get:

 > RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method

This commit fixes both issues by importing bitsandbytes lazily. PEFT import time is now reduced to ~2sec.

## Notes

Implementation-wise, I use a combination of local imports and module-level `__getattr__`. The latter was introduced in Python 3.7 and should therefore be safe to use.

As for testing, it is not as straightforward as just adding a new unit test. The issue is that when multiple tests are run, depending on the order, bitsandbytes may already be imported once the new test runs. To avoid this issue, I created a completely separate testing script that is run in CI. It is executed on the nightly multi-GPU CI run. Locally, it passes, :crossed_fingers: that it also passes there.

If anyone has a better idea, please let me know.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -34,7 +34,7 @@ jobs:
           source activate peft
           pip install -e . --no-deps
           pip install pytest-reportlog
-      
+
       - name: Run common tests on single GPU
         run: |
           source activate peft
@@ -44,12 +44,12 @@ jobs:
         run: |
           source activate peft
           make tests_examples_single_gpu
-      
+
       - name: Run core tests on single GPU
         run: |
           source activate peft
           make tests_core_single_gpu
-          
+
       - name: Generate Report
         if: always()
         run: |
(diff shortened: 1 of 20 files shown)


pacman100 (approved):
Thank you @BenjaminBossan for fixes the issues as well as making the import faster, LGTM! 🚀

.github/workflows/nightly.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           source activate peft
           pip install -e . --no-deps
           pip install pytest-reportlog
-      
+
```
younesbelkada, line 37: Why this diff is here?

BenjaminBossan: Note that I moved the test to the proper unit tests using a trick that Zach showed, so I undid all the changes here (i.e. trailing whitespaces are back).

muellerzr (approved):
Thanks for fixing this! Great job :)

PR #1232, opened 2023-12-07
Copy default value for rank_pattern from LoraConfig

qmeeus:
This avoids errors like NoneType has no attribute keys, e.g. when calling update_and_allocate

Concretely, I changed the default value from `None` to an empty dictionary (`default_factory=dict`). If the original purpose of the None default was to disallow empty values, then I think that explicitly checking for that and raising an exception in `__post_init__` when necessary is more appropriate

### DIFF
--- a/src/peft/tuners/adalora/config.py
+++ b/src/peft/tuners/adalora/config.py
@@ -47,7 +47,15 @@ class AdaLoraConfig(LoraConfig):
     beta2: float = field(default=0.85, metadata={"help": "Hyperparameter of EMA."})
     orth_reg_weight: float = field(default=0.5, metadata={"help": "The orthogonal regularization coefficient."})
     total_step: Optional[int] = field(default=None, metadata={"help": "The total training steps."})
-    rank_pattern: Optional[dict] = field(default=None, metadata={"help": "The saved rank pattern."})
+    rank_pattern: Optional[dict] = field(
+        default_factory=dict,
+        metadata={
+            "help": (
+                "The mapping from layer names or regexp expression to ranks which are different from the default rank specified by `r`. "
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Thanks for working on this. Could you please provide a minimal code snippet that produces the error you mentioned above?

qmeeus:
This is an example of how it might occur. Note that `total_steps` is 1000, `tfinal` is 200 and provided `global_step` is 0, so we are hitting [this condition](https://github.com/huggingface/peft/blob/b482391b801d5e6846f5bd1826b92ece7113afad/src/peft/tuners/adalora/model.py#L312)
```
peft_config = AdaLoraConfig(peft_type=<PeftType.ADALORA: 'ADALORA'>,
              auto_mapping=None,
              base_model_name_or_path='openai/whisper-large-v2',
              revision=None,
              task_type=None,
              inference_mode=False,
              r=8,
              target_modules='.*decoder.*.(k_proj|q_proj|v_proj|out_proj|fc1|fc2)',
              lora_alpha=32,
              lora_dropout=0.1,
              fan_in_fan_out=False,
              bias='none',
              modules_to_save=None,
              init_lora_weights=True,
              layers_to_transform=None,
              layers_pattern=None,
              rank_pattern=None,
              alpha_pattern={},
              target_r=4,
              init_r=12,
              tinit=200,
              tfinal=800,
              deltaT=10,
              beta1=0.85,
              beta2=0.85,
              orth_reg_weight=0.5,
              total_step=1000)
model = PeftModel(AutoModel.from_pretrained("openai/whisper-tiny"), peft_config)
model.model.update_and_allocate(0)
*** TypeError: unsupported operand type(s) for *: 'Parameter' and 'NoneType'
```

BenjaminBossan:
ping @qmeeus

Issue #1225, opened 2023-12-06
ModulesToSaveWrapper has no 'weight' attritude


zhangsheng377:
### System Info

peft=0.4.0
model=Megatron's GPTModel

### Who can help?

@BenjaminBossan 
We are recently using peft=0.4.0+Megatron to run GPTModel. There is a code of the model `self.language_model.output_layer.weight`, which directly takes the weight of the output_layer, but the output_layer has already been `modules_to_save`, so it reported  "ModulesToSaveWrapper has no 'weight' attribute" error.

We now solve the problem with the following code:
```python
def modules_to_save_getattr_hook(self, name: str):
    try:
        return super(ModulesToSaveWrapper, self).__getattr__(name)  # defer to nn.Module's logic
    except AttributeError:
        if self.active_adapter not in self.modules_to_save:
            return getattr(self.original_module, name)
        return getattr(self.modules_to_save[self.active_adapter], name)


setattr(ModulesToSaveWrapper, '__getattr__', modules_to_save_getattr_hook)
```

But I'm not sure if the new version of peft code still has this problem, so I'm here to consult you.

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

self.language_model.output_layer.weight

### Expected behavior

ModulesToSaveWrapper can obtain the "weight" attribute normally

zhangsheng377:
@BenjaminBossan

BenjaminBossan:
Thanks for reporting. Where does the snippet `self.language_model.output_layer.weight` come from, it would be easiest to make a change there to accommodate for `ModulesToSaveWrapper`.

> But I'm not sure if the new version of peft code still has this problem

Newer PEFT versions will display the same issue. I think we could add an attribute to `ModulesToSaveWrapper` that returns the weight of the currently active adapter or of the original module of the adapter is deactivated. But I'm not sure if having that attribute could have some unwanted side-effects.

zhangsheng377:
> Thanks for reporting. Where does the snippet `self.language_model.output_layer.weight` come from, it would be easiest to make a change there to accommodate for `ModulesToSaveWrapper`.

https://github.com/NVIDIA/Megatron-LM/blob/bcce6f54e075e3c3374ea67adefe54f3f2da2b07/megatron/model/gpt_model.py#L94
It comes from here, but the Megatron-LM has not received PR for a long time.


[...]

PR #1238, opened 2023-12-08
Fix ModulesToSaveWrapper __getattr__

zhangsheng377:
https://github.com/huggingface/peft/issues/1225

We are recently using peft=0.4.0+Megatron to run GPTModel. There is a code of the model [`self.language_model.output_layer.weight`](https://github.com/NVIDIA/Megatron-LM/blob/bcce6f54e075e3c3374ea67adefe54f3f2da2b07/megatron/model/gpt_model.py#L94), which directly takes the `weight` of the `output_layer`, but the `output_layer` has already been `modules_to_save`, so it reported "ModulesToSaveWrapper has no 'weight' attribute" error.

So, I returns the weight of the currently active adapter or of the original module of the adapter is deactivated. 

And `__getattr__` will be called only when its own class does not have this attribute, so this should not destroy the original scene.

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -186,6 +186,14 @@ class ModulesToSaveWrapper(torch.nn.Module):
         # use a property to ensure that active_adapter is not set directly, instead use the set_adapter method
         return self._active_adapter
 
+    def __getattr__(self, name: str):
+        try:
+            return super(ModulesToSaveWrapper, self).__getattr__(name)  # defer to nn.Module's logic
+        except AttributeError:
+            if self.active_adapter not in self.modules_to_save:
+                return getattr(self.original_module, name)
+            return getattr(self.modules_to_save[self.active_adapter], name)
+
     def update(self, adapter_name):
         self.modules_to_save.update(torch.nn.ModuleDict({adapter_name: copy.deepcopy(self.original_module)}))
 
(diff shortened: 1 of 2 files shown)


BenjaminBossan (changes requested):
Thanks for the PR. Thinking a bit more about it, I think the best course of action would be to actually implement this as a `@property`, which allows us to avoid fiddling with `__getattr__`.

This should still not affect stuff like `state_dict`, so would not break anything.

On top of that, would you be up to implement a unit test?

zhangsheng377:
> Thanks for the PR. Thinking a bit more about it, I think the best course of action would be to actually implement this as a `@property`, which allows us to avoid fiddling with `__getattr__`.
> 
> This should still not affect stuff like `state_dict`, so would not break anything.
> 
> On top of that, would you be up to implement a unit test?

ok~

BenjaminBossan (approved):
Thanks, LGTM.

Can one of you @pacman100 or @younesbelkada give this a quick look to ensure it doesn't break anything?

pacman100 (approved):
Thank you @zhangsheng377 for adding the weight property when using `ModulesToSaveWrapper`! This makes sense and I don't see any issues with this approach.

younesbelkada:
Tests are all green! https://github.com/huggingface/peft/actions/runs/7194802375/job/19596105974 🍏

PR #1241, opened 2023-12-08
TST: Add tolerance for regression tests

BenjaminBossan:
Tests currently call `torch.allclose` without any tolerance, which is probably the cause of the CI failure. Now, tolerance is set to 1e-4. When testing on another machine, the tests now pass for me.

[Failing tests](https://github.com/huggingface/peft/actions/runs/7136508584/job/19435046711).


younesbelkada (approved):
Thanks!

PR #1244, opened 2023-12-09
Added the option to use the corrected scaling factor for LoRA, based on new research.

Damjan-Kalajdzievski:
Hi, I am proposing to add an option to use the corrected scaling factor for LoRA, based on the recent paper [A Rank Stabilization Scaling Factor for Fine-Tuning with LoRA](https://doi.org/10.48550/arXiv.2312.03732). Try setting `use_rslora = True`  in your `LoraConfig` for ranks greater than 32 and see the increase in fine-tuning performance (same or better performance for ranks lower than 32 as well).
Please feel free to suggest or change the implementation; I tried to go for the minimum code length change that implements this option.

## Summary of method
For a LoRA adapter of rank $r$, the factor $\frac{\alpha}{r}$ that scales the adapter is too aggressive as a function of $r$, and slows learning for higher ranks so that no fine-tuning performance is gained over lower ranks. The paper [A Rank Stabilization Scaling Factor for Fine-Tuning with LoRA](https://doi.org/10.48550/arXiv.2312.03732) proves theoretically and experimentally that we should be using an adapter scaling factor of $\frac{\alpha}{\sqrt{r}}$. This corrected scaling factor unlocks a compute/performance trade-off where increasing the rank increases the fine-tuning performance. This also corrects for the ongoing misconceptions that very low ranks not greater than 32 suffice for maximal performance, which entails the belief that the intrinsic dimensionality of fine-tuning is very low dimensional.

## Description of changes
Added `use_rslora` bool in `LoraConfig`, which when set to`True`, corrects the scaling factor of adapters created with `_create_and_replace` in `LoraModel`. The variable `use_rslora` is set to `False` by default for backwards consistency.

### DIFF
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -57,6 +57,10 @@ class LoraConfig(PeftConfig):
         bias (`str`): Bias type for Lora. Can be 'none', 'all' or 'lora_only'. If 'all' or 'lora_only', the
             corresponding biases will be updated during training. Be aware that this means that, even when disabling
             the adapters, the model will not produce the same output as the base model would have without adaptation.
+        use_rslora (`bool`): When set to True, uses
+            <a href='https://doi.org/10.48550/arXiv.2312.03732'>Rank-Stabilized LoRA</a>
+            which sets the correct adapter scaling factor (`self.scaling`) of `lora_alpha/math.sqrt(r)`.
+            Otherwise, it will use the original default value of `lora_alpha/r`.
         modules_to_save (`List[str]`):List of modules apart from LoRA layers to be set as trainable
             and saved in the final checkpoint.
         layers_to_transform (`Union[List[int],int]`):
@@ -89,6 +93,18 @@ class LoraConfig(PeftConfig):
         metadata={"help": "Set this to True if the layer to replace stores weight like (fan_in, fan_out)"},
     )
     bias: str = field(default="none", metadata={"help": "Bias type for Lora. Can be 'none', 'all' or 'lora_only'"})
+    use_rslora: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "When set to True, uses "
+                "<a href='https://doi.org/10.48550/arXiv.2312.03732'>Rank-Stabilized LoRA</a> "
+                "which sets the correct adapter scaling factor (`self.scaling`) "
+                "of `lora_alpha/math.sqrt(r)`. Otherwise, it will use the original "
+                "default value of `lora_alpha/r`."
+            )
+        },
+    )
     modules_to_save: Optional[List[str]] = field(
         default=None,
         metadata={
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -59,8 +59,8 @@ class LoraConfig(PeftConfig):
             the adapters, the model will not produce the same output as the base model would have without adaptation.
         use_rslora (`bool`):
             When set to True, uses <a href='https://doi.org/10.48550/arXiv.2312.03732'>Rank-Stabilized LoRA</a> which
-            sets the adapter scaling factor to the correct value of `lora_alpha/math.sqrt(r)`. Otherwise, it will use
-            the original default value of `lora_alpha/r`.
+            sets the adapter scaling factor to `lora_alpha/math.sqrt(r)`, since it was proven to work better.
+            Otherwise, it will use the original default value of `lora_alpha/r`.
         modules_to_save (`List[str]`):List of modules apart from LoRA layers to be set as trainable
             and saved in the final checkpoint.
         layers_to_transform (`Union[List[int],int]`):
@@ -97,11 +97,10 @@ class LoraConfig(PeftConfig):
         default=False,
         metadata={
             "help": (
-                "When set to True, uses "
-                "<a href='https://doi.org/10.48550/arXiv.2312.03732'>Rank-Stabilized LoRA</a> "
-                "which sets the adapter scaling factor to the correct value "
-                "of `lora_alpha/math.sqrt(r)`. Otherwise, it will use the original "
-                "default value of `lora_alpha/r`."
+                "When set to True, uses Rank-Stabilized LoRA doi.org/10.48550/arXiv.2312.03732"
+                " which sets the adapter scaling factor to `lora_alpha/math.sqrt(r)`, since it"
+                " was proven to work better. Otherwise, it will use the original default"
+                " value of `lora_alpha/r`."
             )
         },
     )
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -57,10 +57,10 @@ class LoraConfig(PeftConfig):
         bias (`str`): Bias type for Lora. Can be 'none', 'all' or 'lora_only'. If 'all' or 'lora_only', the
             corresponding biases will be updated during training. Be aware that this means that, even when disabling
             the adapters, the model will not produce the same output as the base model would have without adaptation.
-        use_rslora (`bool`): When set to True, uses
-            <a href='https://doi.org/10.48550/arXiv.2312.03732'>Rank-Stabilized LoRA</a>
-            which sets the adapter scaling factor to the correct value of `lora_alpha/math.sqrt(r)`.
-            Otherwise, it will use the original default value of `lora_alpha/r`.
+        use_rslora (`bool`):
+            When set to True, uses <a href='https://doi.org/10.48550/arXiv.2312.03732'>Rank-Stabilized LoRA</a> which
+            sets the adapter scaling factor to the correct value of `lora_alpha/math.sqrt(r)`. Otherwise, it will use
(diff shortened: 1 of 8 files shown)


Damjan-Kalajdzievski:
> Thanks @Damjan-Kalajdzievski for the PR. I haven't checked it yet, but could you please run `make style` so that the CI can run?

Hi, I have run the command and commited the modification to the files I've changed.

src/peft/tuners/lora/config.py
```diff
@@ -57,6 +57,10 @@ class LoraConfig(PeftConfig):
         bias (`str`): Bias type for Lora. Can be 'none', 'all' or 'lora_only'. If 'all' or 'lora_only', the
             corresponding biases will be updated during training. Be aware that this means that, even when disabling
             the adapters, the model will not produce the same output as the base model would have without adaptation.
+        use_rslora (`bool`):
+            When set to True, uses <a href='https://doi.org/10.48550/arXiv.2312.03732'>Rank-Stabilized LoRA</a> which
```
BenjaminBossan, line 61:
```suggestion
            When set to True, uses <a href='https://doi.org/10.48550/arXiv.2312.03732'> Rank-Stabilized LoRA</a> which
```

For some reason, there is no space before the name, see here:

https://moon-ci-docs.huggingface.co/docs/peft/pr_1244/en/package_reference/lora

Maybe this fixes it.

src/peft/tuners/lora/config.py
```diff
@@ -57,6 +57,10 @@ class LoraConfig(PeftConfig):
...
+            sets the adapter scaling factor to the correct value of `lora_alpha/math.sqrt(r)`. Otherwise, it will use
```
BenjaminBossan, line 62: I wouldn't say "correct value", maybe something like:

> sets the adapter scaling factor to `lora_alpha/math.sqrt(r)`, which was shown to work better.

src/peft/tuners/lora/config.py
```diff
@@ -89,6 +93,18 @@ class LoraConfig(PeftConfig):
         metadata={"help": "Set this to True if the layer to replace stores weight like (fan_in, fan_out)"},
     )
     bias: str = field(default="none", metadata={"help": "Bias type for Lora. Can be 'none', 'all' or 'lora_only'"})
+    use_rslora: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "When set to True, uses "
+                "<a href='https://doi.org/10.48550/arXiv.2312.03732'>Rank-Stabilized LoRA</a> "
```
BenjaminBossan, line 101: Let's not use html syntax in the help.

src/peft/tuners/lora/config.py
```diff
@@ -89,6 +93,18 @@ class LoraConfig(PeftConfig):
...
+                "which sets the adapter scaling factor to the correct value "
```
BenjaminBossan, line 102: Same argument about "correct".

src/peft/tuners/lora/model.py
```diff
@@ -194,6 +195,13 @@ def _create_and_replace(
                 new_module.requires_grad_(False)
             self._replace_module(parent, target_name, new_module, target)
 
+        if lora_config.use_rslora:
```
BenjaminBossan, line 198: I think here is not the right place to control the scaling, as this leads to spreading the initialization of the LoRA parameters into different parts of the code. Instead, `update_layer`, `update_layer_conv2d`, and `update_layer_embedding` in `tuners/lora/layer.py` should be adjusted, since that's where we set the scale initially. This also requires updating the `__init__` method to accept the new argument, as well as the `kwargs` variable [here](https://github.com/huggingface/peft/blob/0f1e9091cc975eb5458cc163bf1843a34fb42b76/src/peft/tuners/lora/model.py#L151-L157).

BenjaminBossan (changes requested):
Thanks a lot for adding this seemingly small but efficient initialization option to LoRA. I only skimmed the paper, but it seems that this options helps across the board, although it's most prominent for large ranks.

Regarding how it's implemented, I think we need to adjust the where we apply the scaling. Check my comment regarding that.

Furthermore, I would like to see two additions:

- An entry in the docs about this new option [here](https://github.com/huggingface/peft/blob/0f1e9091cc975eb5458cc163bf1843a34fb42b76/docs/source/conceptual_guides/lora.md#initialization-options)
- A small unit test that checks the scaling factor after initializing a simple LoRA model. If you need help with that, let us know.

src/peft/tuners/lora/config.py
Damjan-Kalajdzievski: I think I may have already fixed it [in the last commit](https://github.com/huggingface/peft/pull/1244/commits/44e6eabed5598090e3c5be84802b908e73f477d1) by making sure there is a space and not a newline after "uses". My main problem is I can't test the compilation of the docs as I'm still trying to get doc-builder working for me (it compiles the docs with no changes for me for some reason)

src/peft/tuners/lora/config.py
Damjan-Kalajdzievski: I interpreted this to mean that I should still include the url but without the html syntax.

Damjan-Kalajdzievski:
> * An entry in the docs about this new option [here](https://github.com/huggingface/peft/blob/0f1e9091cc975eb5458cc163bf1843a34fb42b76/docs/source/conceptual_guides/lora.md#initialization-options)

I am not sure if it belongs in the description of the initializations since I didn't want to confuse a person unfamiliar with LoRA that the scaling factor is only applied at initialization instead of scaling every adapter output. So [here I commited](https://github.com/huggingface/peft/pull/1244/commits/5b141ae854728b3065be46915293c30d1d25bc36) a small section below here describing the scaling factor and what use_rslora does. Not sure if this is the best place so feel free to suggest a change.

> * A small unit test that checks the scaling factor after initializing a simple LoRA model. If you need help with that, let us know.

I added a test in the test_initialization.py on commit [871ed7d](https://github.com/huggingface/peft/pull/1244/commits/871ed7dc097d6c101c86b1d6ec0ca900ac70944a) . Let me know if this makes sense for what you envisioned the test to be like.

>I think here is not the right place to control the scaling, as this leads to spreading the initialization of the LoRA parameters into different parts of the code. Instead, update_layer, update_layer_conv2d, and update_layer_embedding in tuners/lora/layer.py should be adjusted, since that's where we set the scale initially. This also requires updating the __init__ method to accept the new argument, as well as the kwargs variable [here](https://github.com/huggingface/peft/blob/0f1e9091cc975eb5458cc163bf1843a34fb42b76/src/peft/tuners/lora/model.py#L151-L157).

Makes sense, I had wondered about doing it this way also and will do this now.

tests/test_initialization.py
```diff
@@ -230,3 +230,27 @@ def test_lora_conv2d_false(self):
...
         self.assertFalse(torch.allclose(weight_B, torch.zeros_like(weight_B)))
+
+    def test_lora_use_rslora(self):
+        # default is True
+        torch.manual_seed(0)
+
+        _model = self.get_model()
+
+        # check scaling factor use_rslora=False
+        config = LoraConfig(target_modules=["linear", "embed", "conv2d"], lora_alpha=3, r=16, use_rslora=False)
```
BenjaminBossan, line 241: Thanks for testing the base setting with `use_rslora=False` too. Let's split this into two separate tests, say `test_lora_scaling_default` and `test_rslora_scaling` for better readability. Also, since the `_model` is mutated by `get_peft_model`, it's also the cleaner solution.

On top, could you please also add a test using `rank_pattern` and `alpha_pattern`, just to be sure that this works with rslora as well? So e.g. using `LoraConfig(..., rank_pattern={"embed": 8, "conv2d": 32}, alpha_pattern={"linear": 4, "conv2d": 20})`.

src/peft/tuners/lora/config.py
BenjaminBossan: The docs should be updated automatically (once the "Build PR Documentation" job has finished). But anyway, the space is now there.

BenjaminBossan (changes requested):
Thanks for making the adjustments. From my point of view, there are only a few small changes needed and then the PR should be good to be merged.

> I am not sure if it belongs in the description of the initializations since I didn't want to confuse a person unfamiliar with LoRA that the scaling factor is only applied at initialization, as instead it scales every adapter output.

IMO, we can still consider this to be initialization, as in almost all use cases, scaling is set once, at initialization, and then remains untouched.

> I added a test in the test_initialization.py

Nicely done.

> Not sure exactly what class' __init__ method you mean, but I added the argument use_rslora to all the update_* methods, and then feed it through the kwargs in _create_new_module so that every sublcass of LoraLayer calling update_layer has the use_rslora argument to pass along. Essentially treating use_rslora in the code exactly analogous to init_lora_weights.

Yes, that was exactly what I meant, thanks!

BenjaminBossan (approved):
Thanks, looks very good, nice tests.

pacman100 (approved):
Thank you @Damjan-Kalajdzievski for adding the support for rank-stabilized LoRA scaling, LGTM! 🚀

PR #1251, opened 2023-12-11
FIX: Truncate slack message to not exceed 3000 chars

BenjaminBossan:
Should fix the issue of not receiving slack notifications because the message is too long.

Currently, we get:

> Error: ver responded with: {'ok': False, 'error': 'invalid_blocks', 'errors': ['failed to match all allowed schemas [json-pointer:/blocks/1/text]', 'must be less than 3001 characters [json-pointer:/blocks/1/text/text]'], 'response_metadata': {'messages': ['[ERROR] failed to match all allowed schemas [json-pointer:/blocks/1/text]', '[ERROR] must be less than 3001 characters [json-pointer:/blocks/1/text/text]']}}

https://github.com/huggingface/peft/actions/runs/7148379741/job/19469273483

Fixing the error should also lead to a shorter message, but we should ensure that even if the message is too long, we still get it.

Also: Removed some trailing whitespace.

### DIFF
--- a/scripts/log_reports.py
+++ b/scripts/log_reports.py
@@ -3,6 +3,9 @@ from pathlib import Path
 from datetime import date
 from tabulate import tabulate
 
+
+MAX_LEN_MESSAGE = 2900  # slack endpoint has a limit of 3001 characters
+
 failed = []
 passed = []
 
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Sounds good, thanks!

PR #1255, opened 2023-12-12
FIX: Fix adaption prompt bug with transformers 4.36.0

younesbelkada:
As per title, https://github.com/huggingface/transformers/pull/26681 introduced new cache mechanism, this PR patches the adaption prompt modeling code to handle the new case as well

cc @BenjaminBossan @pacman100


younesbelkada:
Note this PR is not enough to make the CI green , we still need to figure out a fix for the other failing tests

Issue #1260, opened 2023-12-13
This regex is confusing


s-k-yx:
https://github.com/huggingface/peft/blob/c9df262d69b41d36c7e199219605420be03dfbca/src/peft/tuners/tuners_utils.py#L502
It works now, but seems has a wrong format.
`r".*\.{pattern}\.(\d+)\..*".format(pattern=pattern)` this is a better format, if only layer_index is needed `r".*\.{pattern}\.(\d+)".format(pattern=pattern)` is enough.

module name examples:
...
model.layers.0.post_attention_layernorm
model.layers.1  # first expression will escape this
model.layers.1.self_attn
model.layers.1.self_attn.q_proj
model.layers.1.self_attn.k_proj
...

BenjaminBossan:
Do you mean specifically that the 2nd `.` is not escaped and the `*` at the end? I agree that those looks suspicious. We have an extensive test suite here:

https://github.com/huggingface/peft/blob/main/tests/test_tuners_utils.py

Some test cases have dubious results (see the comments). If you want to, feel free to create a PR with a better regex and we can check how the tests are affected.

s-k-yx:
ok, i've made a PR, and  this test case should be false?  https://github.com/huggingface/peft/blob/0f1e9091cc975eb5458cc163bf1843a34fb42b76/tests/test_tuners_utils.py#L65

BenjaminBossan:
Fixed via #1263

PR #1263, opened 2023-12-13
Fix logic in target module finding

s-k-yx:
fix the regex in target module finding, refer to #1260 
modify the logic for empty layer pattern, i don't know whether COMMON_LAYERS_PATTERN has special meaning, new logic is: empty ==> all layer pattern is ok


BenjaminBossan:
Thanks for the PR @s-k-yx. I haven't looked at it yet, but could you please run `make style` so that CI can run successfully?

s-k-yx:
> Thanks for the PR @s-k-yx. I haven't looked at it yet, but could you please run `make style` so that CI can run successfully?

sorry for that, the file is reformatted. And there is a dubious test case: https://github.com/huggingface/peft/blob/0f1e9091cc975eb5458cc163bf1843a34fb42b76/tests/test_tuners_utils.py#L65
maybe this should be checked, thanks

src/peft/tuners/tuners_utils.py
```diff
@@ -486,30 +484,43 @@ def check_target_module_exists(config, key: str) -> bool | re.Match[str] | None:
...
-                        target_module_found = layer_index in config.layers_to_transform
-
-                    break
+            # to keep up with the previous results(if layers_pattern is an empty list, ret is True)
+            # and make the result consistent for all empty layers_pattern: None, [], ""
+            # COMMON_LAYERS_PATTERN is dropped
+            # empty layers_pattern means all layer pattern is ok
+            layer_index = None
+            if layers_pattern is None or len(layers_pattern) == 0:
+                layer_index = re.match(r".*\.[^.]*\.(\d+)\.", key)
```
BenjaminBossan, line 507: Why would you say it's better not to rely on `COMMON_LAYERS_PATTERN` here? If we decide it's really not needed, it should be completely removed from the PEFT code base.

src/peft/tuners/tuners_utils.py
```diff
@@ -486,30 +484,43 @@ def check_target_module_exists(config, key: str) -> bool | re.Match[str] | None:
...
-                layer_index = re.match(f".*.{pattern}\.(\d+)\.*", key)
-                if layer_index is not None:
-                    layer_index = int(layer_index.group(1))
-                    if isinstance(config.layers_to_transform, int):
-                        target_module_found = layer_index == config.layers_to_transform
-                    else:
-                        target_module_found = layer_index in config.layers_to_transform
-
-                    break
+            # to keep up with the previous results(if layers_pattern is an empty list, ret is True)
```
BenjaminBossan, line 501: This whole comment is not clear to me, could you please try to word it differently? Note that the comment should make sense as is, even if the reader is not aware of how the code looked like previously.

src/peft/tuners/tuners_utils.py
```diff
@@ -486,30 +484,43 @@ def check_target_module_exists(config, key: str) -> bool | re.Match[str] | None:
...
+                for pattern in layers_pattern:
+                    layer_index = re.match(r".*\.{layer}\.(\d+)\.".format(layer=pattern), key)
+                    if layer_index is not None:
+                        break
+
+            if layer_index is None:
+                target_module_found = False
+            else:
+                layer_index = int(layer_index.group(1))
+                if isinstance(config.layers_to_transform, int):
```
BenjaminBossan, line 519: Probably, instead of using `config.layers_to_transform`, we should use `layer_indexes`.

src/peft/tuners/tuners_utils.py
s-k-yx: I didn't find any other direct usage of COMMON_LAYERS_PATTERN, so I think it won't be harmful. As for the reason why I drop it, I try to avoid changing the output and keep the logic consistent. In the previous code, if the layers_pattern is an empty list, all layer indexes can be matched, as stated in https://github.com/huggingface/peft/blob/482a2a6d9aaa01d534b1240e8c1ab6d346eb278f/tests/test_tuners_utils.py#L59C9-L59C9
In fact, COMMON_LAYERS_PATTERN also doesn't work here. Maybe this is a bug, but I don't change it to keep the output consistent when layers_pattern is an empty list. And now '[]' makes all layers being matched, I think None(default value) should follow this too, even more suitable than '[]'. So I just make them all consistent, None, [], ""(empty string). Then empty layers_pattern means any layer pattern is ok, it's also reasonable. Maybe there should be a warning after the layers_pattern is set.
And torch.nn.ModuleList where use numeric value to be the module name when add modules, i.e. the layer index, so I don't think others will use numeric value as their own module name. Also there can't be '.' in module name, so I think it's safe to just match "[^.]*\.\d+\.".
How to treat empty list in layers_pattern is controversial, here I just follow previous 'bug'.

src/peft/tuners/tuners_utils.py
s-k-yx: Sorry for that, I explain it above, and it will be changed later

src/peft/tuners/tuners_utils.py
s-k-yx: Your are right, layer_indexes should be better. Here I just keep the previous code, I'll change it later

s-k-yx:
Thanks for the helpful suggestions, I'll change test code and comments later. And I'm sorry I didn't make myself clear. I've made some replies, hope it helps.

s-k-yx:
And there is the same problem in layer_indexes, now the new added test case ("foo.bar.7.baz", ["baz"], [], ["bar"], True) will pass(in previous code, it will fail), to be consistent with layers_pattern. Maybe all the empty list should fail, no matter layers_pattern or  layer_indexes

src/peft/tuners/tuners_utils.py
```diff
@@ -23,8 +23,6 @@
 import torch
 from torch import nn
 
-from peft.utils import COMMON_LAYERS_PATTERN
```
BenjaminBossan, line 26: Could you please remove this constant completely from the code base?

tests/test_tuners_utils.py
```diff
@@ -55,14 +55,16 @@
...
-    ("foo.bar.1.baz", ["baz"], [1], [], True),
-    # TODO: Below test currently fails, again because of empty layers_pattern
-    # layers_to_transform is 0, but layers_pattern is empty, so all layer indexes are matched
-    # ("foo.bar.1.baz", ["baz"], [0], [], False),
-    ("foo.bar.1.baz", ["baz"], [1], ["ar"], True),
+    # empty layers_to_transform
+    ("foo.bar.7.baz", ["baz"], [], ["bar"], True),
+    ("foo.bar.7.baz", ["baz"], None, ["bar"], True),
+    # empty layers_pattern
+    ("foo.whatever.1.baz", ["baz"], [1], [], True),
```
BenjaminBossan, line 62: Thanks for adding these examples.

tests/test_tuners_utils.py
```diff
@@ -55,14 +55,16 @@
...
-    # Currently, an empty layers_pattern leads to all layer indexes being matched,
-    # which means layers_to_transform is ignored.
-    ("foo.bar.1.baz", ["baz"], [1], [], True),
-    # TODO: Below test currently fails, again because of empty layers_pattern
-    # layers_to_transform is 0, but layers_pattern is empty, so all layer indexes are matched
-    # ("foo.bar.1.baz", ["baz"], [0], [], False),
-    ("foo.bar.1.baz", ["baz"], [1], ["ar"], True),
+    # empty layers_to_transform
+    ("foo.bar.7.baz", ["baz"], [], ["bar"], True),
+    ("foo.bar.7.baz", ["baz"], None, ["bar"], True),
```
BenjaminBossan, line 60: Also test `""` for `layers_to_transform`?

src/peft/tuners/tuners_utils.py
```diff
@@ -483,30 +481,41 @@ def check_target_module_exists(config, key: str) -> bool | re.Match[str] | None:
...
-                if layer_index is not None:
-                    layer_index = int(layer_index.group(1))
-                    if isinstance(config.layers_to_transform, int):
-                        target_module_found = layer_index == config.layers_to_transform
-                    else:
-                        target_module_found = layer_index in config.layers_to_transform
-
-                    break
+            layer_index = None
+            # fixme: It's still unclear how empty layers_pattern(None, [], or "") should behave
```
BenjaminBossan, line 499:
```suggestion
            # TODO It's still unclear how empty layers_pattern (None, [], or "") should behave
```

Let's change `fixme` to `TODO`. We use `FIXME` for things that need to be fixed immediately, i.e. before the PR is merged. `TODO` is more long term.

BenjaminBossan (changes requested):
Thanks a lot for improving the PR and your explanations. Those make sense to me. I only have a few smaller change requests left.

> Maybe all the empty list should fail, no matter layers_pattern or layer_indexes

I agree that this could be more sensible, but it could be breaking backwards compatibility of existing code out there, so I'd rather want to avoid it.

tests/test_tuners_utils.py
s-k-yx: Due to the type of this arg `layers_to_transform: Optional[Union[List[int], int]])`, these two should be enough.

s-k-yx:
Thanks for the checking, I've modified them accordingly, glad to be of help.

BenjaminBossan (approved):
Thanks, this LGTM to me now. 

Since this PR has the potential to break existing code, let's wait for another review from @pacman100 or @younesbelkada to ensure that all is good.

src/peft/tuners/tuners_utils.py
```diff
@@ -483,30 +481,41 @@ def check_target_module_exists(config, key: str) -> bool | re.Match[str] | None:
         None if no match found
     """
     if isinstance(config.target_modules, str):
-        target_module_found = re.fullmatch(config.target_modules, key)
+        target_module_found = (config.target_modules == key) or re.fullmatch(config.target_modules, key)
```
younesbelkada, line 484:
```suggestion
        target_module_found = re.fullmatch(config.target_modules, key)
```
those are equivalent no?

younesbelkada (commented):
Thanks ! I left a single comment, what do you think?

src/peft/tuners/tuners_utils.py
s-k-yx: Thanks for the suggestion, indeed the suggestion is the previous code. I change the code because I see `target_module_found = key in config.target_modules or ...` when the type of `target_modules` is list.  And if the `target_modules` is regex it will be fine, but if the `target_modules` is just the complete layer name and contains special characters in regex, it may cause problem. But for the common layer name like `foo.bar.1.baz`, it doesn't matter, although `.` is special in regex.
  In a word, `target_module_found = re.fullmatch(config.target_modules, key)` should be fine in almost all the cases, but it doesn't hurt to add `(config.target_modules == key) or `.

younesbelkada: ok looks good to me, thanks for explaining!

younesbelkada (approved):

s-k-yx:
> Thanks ! I left a single comment, what do you think?

sorry, I found there is the same problem in my change, I shouldn't mix regex and simply compare. It should be better to use `target_module_found = re.fullmatch(config.target_modules, key)` and limit it to regex when `target_modules` is `str`

BenjaminBossan:
> sorry, I found there is the same problem in my change, I shouldn't mix regex and simply compare. It should be better to use `target_module_found = re.fullmatch(config.target_modules, key)` and limit it to regex when `target_modules` is `str`

Just for my understanding, what is the problem exactly? Can you give an example?

BenjaminBossan:
Okay, I think that makes sense. I thought it would be impossible to do this because sub-modules must have a valid Python variable name as identifier, but that's actually not true:

```python
class MyModule(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.register_module("foo-bar", torch.nn.Linear(1, 2))
        self.register_module("foo/bar$baz^!", torch.nn.Linear(1, 2))

m = MyModule()
getattr(m, "foo-bar")  # works
getattr(m, "foo/bar$baz^!")  # works

config = LoraConfig(target_modules=["foo-bar", "foo/bar$baz^!"])
get_peft_model(m, config)  # works
```

Hopefully, nobody does this in the wild, but we can never know :) Do you think we should have a test for this @s-k-yx?

s-k-yx:
It's always hard to guess user's purpose, telling them how to use makes things simple :) The above test case does nothing wrong, so it should work. Just clarify it in parameter description that it's different when the type of  `target_modules` is `str` and `list`, when it's `str` it should be regex to match the full layer name,  and when it's `list`, the element can be the full layer name or the module name but never regex. Will that be fine? @BenjaminBossan

BenjaminBossan:
> Just clarify it in parameter description that it's different when the type of `target_modules` is `str` and `list`, when it's `str` it should be regex to match the full layer name, and when it's `list`, the element can be the full layer name or the module name but never regex. Will that be fine?

This sound excellent.

BenjaminBossan:
@s-k-yx Would you be so kind as to fix the merge conflict?

s-k-yx:
done @BenjaminBossan

PR #1268, opened 2023-12-14
Refactor and a couple of fixes for adapter layer updates

BenjaminBossan:
This started as a simple refactor, but during this, I discovered a few issues too, which should now be fixed. This is probably best reviewed by checking each commit separately.

Here are the descriptions for the individual changes:

## 1. Refactor: Move LoRA `update_layer` to child classes 

For LoRA, so far, we have `update_layer` for `Linear`, `update_layer_embedding` for `Embedding`, and `update_layer_conv2d` for Conv2d, all defined on `LoraLayer`.

We can simplify the code by always using the name `update_layer`, and by moving the layer-specific methods to the subclasses. So e.g. `update_layer_embedding` is moved to the `Embedding` class and renamed to `update_layer`. This way, the caller does not need to differentiate which type of layer it's calling, making it much easier (see the simplification in `LoraModel._create_and_replace`). It also makes the code less error prone (see change 4 described below).

Interestingly, this was already practiced for IA³, so the same change was not necessary there. But I did find the same method implemented twice, once on `IA3Layer` and once on `Linear`, so I removed one of the duplicates. For all other adapter methods, no change was required, as they only implement `Linear`.

*Note: This could technically be backwards incompatible if users do some custom stuff with `LoraLayer`s. I could add a note for next release that all `update_layer_*` methods have been consolidated to use the same `update_layer` name.*

## 2. Systematic handling of `r` (rank) <= 0 

Always raise an error when `r <= 0`, not only for LoRA. Also, removed later check for `r > 0` in LoRA layers, since we already check for `r <= 0`.

*Note: This could technically also be considered backwards incompatible, but `r<=0` should not work correctly anyway, so better to raise an error right away.*

## 3. Fix broken `__repr__` method on `QuantLinear` 

Was indented too deep, thus not being applied.

## 4. Fix bug for updating Lora GPTQ and IA3 bnb layers 

Before this fix, when adding a 2nd adapter to a model, we did not correctly check if there was already an adapter layer in the model when dealing with LoRA GPTQ or IA3 bnb layers. As a consequence, instead of updating the existing layers, we would create a new layer and the existing layer would be set as the `base_layer` of that new layer. Now, we correctly update the existing layer to add the new adapter.

For this fix to work correctly with LoRA and GPTQ, I had to add a check for `qweight` in `update_layer`, since we only checked for `weight` before. This was a mistake that didn't surface so far because of the error described in the previous paragraph.

Tests were extended to check this. They fail with the current main but pass on my machine with this PR.

### DIFF
--- a/src/peft/tuners/adalora/layer.py
+++ b/src/peft/tuners/adalora/layer.py
@@ -37,6 +37,9 @@ class AdaLoraLayer(LoraLayer):
         self.ranknum = nn.ParameterDict({})
 
     def update_layer(self, adapter_name, r, lora_alpha, lora_dropout, init_lora_weights):
+        if r <= 0:
+            raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
+
         self.r[adapter_name] = r
         self.lora_alpha[adapter_name] = lora_alpha
         if lora_dropout > 0.0:
--- a/src/peft/tuners/ia3/layer.py
+++ b/src/peft/tuners/ia3/layer.py
@@ -53,6 +53,7 @@ class IA3Layer(BaseTunerLayer):
         self.out_features = out_features
 
     def update_layer(self, adapter_name, init_ia3_weights):
+        # This code works for linear layers, override for other layer types
         # Actual trainable parameters
         if self.is_feedforward:
             weight = torch.randn((1, self.in_features))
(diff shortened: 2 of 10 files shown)


pacman100 (approved):
Thank you @BenjaminBossan for fixing many bugs while simplify a lot of code. LGTM! 🔥

younesbelkada (approved):
Thanks ! Can you double check integration tests pass, just in case?

BenjaminBossan:
@younesbelkada Integration tests don't work on forks, right? So we can check after merging only.

younesbelkada:
I just ran it here: https://github.com/huggingface/peft/actions/runs/7246319576

younesbelkada:
Tests are green ! Feel free to merge!

PR #1271, opened 2023-12-15
remove a duplicated description in peft BaseTuner

butyuhao:
remove duplicated description for _check_target_module_exists in BaseTuner class


younesbelkada (approved):
thanks!

PR #1274, opened 2023-12-15
TST: Extend LoftQ tests to check CPU initialization

BenjaminBossan:
Tests to complement PR #1256, see [this comment](https://github.com/huggingface/peft/pull/1256#pullrequestreview-1780037331).


younesbelkada (approved):
Thanks a lot!

PR #1275, opened 2023-12-15
[`Tests`] Add bitsandbytes installed from source on new docker images

younesbelkada:
# What does this PR do?

As per title, this will make life much easier for bnb developpers. I propose to create two separate docker images, both with bnb installed from source and run PEFT + transformers slow tests on daily basis, similarly as what we do currently with PEFT.

I can confirm building these docker images were successful on my local VM

cc @LysandreJik @Titus-von-Koeller @pacman100 @BenjaminBossan 
Will first test if the workflow works fine here on GH

### DIFF
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -1,9 +1,10 @@
 name: Build Docker images (scheduled)
 
 on:
-  push:
-    branches:
-      - add-peft-bnb-source-docker
+  workflow_dispatch:
+  workflow_call:
+  schedule:
+    - cron: "0 1 * * *"
 
 concurrency:
   group: docker-image-builds
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -1,10 +1,9 @@
 name: Build Docker images (scheduled)
 
 on:
-  workflow_dispatch:
-  workflow_call:
-  schedule:
-    - cron: "0 1 * * *"
+  push:
+    branches:
+      - add-peft-bnb-source-docker
 
 concurrency:
   group: docker-image-builds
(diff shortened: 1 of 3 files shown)


.github/workflows/build_docker_images.yml
```diff
@@ -1,10 +1,9 @@
 name: Build Docker images (scheduled)
 
 on:
-  workflow_dispatch:
-  workflow_call:
-  schedule:
-    - cron: "0 1 * * *"
+  push:
+    branches:
+      - add-peft-bnb-source-docker
```
younesbelkada, line 6: to remove after testing

younesbelkada:
I can confirm the built image points to the correct commit hash: https://github.com/huggingface/peft/actions/runs/7224973902/job/19687439574?pr=1275#step:6:6528 on bnb main

younesbelkada:
The PR is ready for review, you can find the freshly built docker images [here](https://hub.docker.com/r/huggingface/peft-gpu-bnb-latest) and [here](https://hub.docker.com/r/huggingface/peft-gpu-bnb-source)

.github/workflows/build_docker_images.yml
```diff
@@ -72,3 +72,66 @@ jobs:
           context: ./docker/peft-gpu
           push: true
           tags: huggingface/peft-gpu
+
+  latest-cuda-bnb-source:
+    name: "Latest Peft GPU + bnb source [dev]"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Cleanup disk
```
BenjaminBossan, line 80: I've never seen this before in a workflow file :O

docker/peft-gpu-bnb-latest/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+    accelerate \
+    peft
+
+RUN source activate peft && \ 
+    git clone https://github.com/TimDettmers/bitsandbytes && cd bitsandbytes && \
+    CUDA_VERSION=121 make cuda12x && \
+    python setup.py develop
+
+RUN source activate peft && \ 
+    pip freeze | grep bitsandbytes
```
BenjaminBossan, line 65: Steps like these could be combined to a single one to reduce the footprint.

docker/peft-gpu-bnb-latest/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+
+# Stage 2
+FROM nvidia/cuda:12.2.2-devel-ubuntu22.04 AS build-image
+COPY --from=compile-image /opt/conda /opt/conda
+ENV PATH /opt/conda/bin:$PATH
+
+RUN chsh -s /bin/bash
+SHELL ["/bin/bash", "-c"]
+RUN source activate peft && \ 
+    python3 -m pip install --no-cache-dir optimum auto-gptq
```
BenjaminBossan, line 41: Why a separate install of these packages here?

BenjaminBossan (approved):
Looks pretty good, thanks. There are a few minor things that could be done to improve the Docker image caching, but no blocker IMO.

.github/workflows/build_docker_images.yml
younesbelkada: Yeah, I've copied that from transformers since I used to get some disk issues: https://github.com/huggingface/transformers/blob/main/.github/workflows/build-docker-images.yml#L25

docker/peft-gpu-bnb-latest/Dockerfile
younesbelkada: good catch!

docker/peft-gpu-bnb-latest/Dockerfile
younesbelkada: I've now refactored everthing based on your suggestions !

BenjaminBossan (approved):
Nice.

pacman100 (approved):
Thank you @younesbelkada, LGTM! 🚀

PR #1276, opened 2023-12-15
LoftQ: edit README.md and example files

yxli2123:
Hi,

I would like to update the README.md file and example scripts for LoftQ


examples/loftq_finetuning/quantize_save_load.py
```diff
@@ -184,54 +183,8 @@ def quantize_and_save():
     return base_model_dir, lora_model_dir
 
 
-def load_loftq(base_model_path, lora_adapter_path):
```
younesbelkada, line 187: Why this has been removed?

younesbelkada (commented):
Thank you! Left a single comment

pacman100 (approved):
Thank you @yxli2123 for updating README and examples for the LoftQ method, LGTM!

examples/loftq_finetuning/quantize_save_load.py
pacman100: I think this function was supposed to confirm that everything works fine after the loftq weight initialization step. Hence, not a required step.

younesbelkada: Perfect!

PR #1279, opened 2023-12-18
TST: Enable LoftQ 8bit tests

BenjaminBossan:
Due to PR #1276, the bug that prevented use of LoftQ with 8bit quantization has now been fixed. Therefore, the tests no longer need to be skipped.

Note

I tested locally with GPU and the tests passed.


younesbelkada (approved):
Thank you very much Benjamin !

PR #1281, opened 2023-12-19
Fixed several errors in StableDiffusion adapter conversion script

kovalexal:
Hi!

I noticed and fixed several errors in the Stable Diffusion adapter conversion script:

1. There was an error in LoRA keys which resulted in wrong output checkpoints (no error was raised)
2. Some existing LoRA checkpoints like LCM contain only UNet keys, so the script failed to convert them properly

@BenjaminBossan your review is kindly appreciated.


BenjaminBossan (approved):
Thanks for those fixes, LGTM.

PR #1285, opened 2023-12-20
Only use trainable params in prompt tuning optimizer

### LINKED ISSUE #1284: Prompt Tuning documentation trains all parameters
### System Info

Google colab

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

It seems like the [documentation for prompt tuning](https://huggingface.co/docs/peft/task_guides/clm-prompt-tuning) sets up the optimizer incorrectly
`optimizer = torch.optim.AdamW(model.parameters(), lr=lr)` 

Shouldn't this be only the trainable parameters? Otherwise, you're using a lot of additional memory for params that won't be tuned. In this example, you tune on bloom with 500M parameters. 

AdamW requires 18 bytes per model parameter, so 18B * 500M = 9GB. Is this correct, or is there a reason the optimizer requires all parameters? 

Thanks!

### Expected behavior

Potentially a new function `model.trainable_parameters` that would then be passed in

Ben-Epstein:
Closes #1284

### DIFF
--- a/docs/source/task_guides/clm-prompt-tuning.md
+++ b/docs/source/task_guides/clm-prompt-tuning.md
@@ -188,7 +188,7 @@ print(model.print_trainable_parameters())
 Setup an optimizer and learning rate scheduler:
 
 ```py
-optimizer = torch.optim.AdamW(model.parameters(), lr=lr)
+optimizer = torch.optim.AdamW(filter(lambda p: p.requires_grad, model.parameters()), lr=lr)
 lr_scheduler = get_linear_schedule_with_warmup(
     optimizer=optimizer,
     num_warmup_steps=0,


younesbelkada (approved):
Makes sense, thanks a lot for your contribution!

BenjaminBossan:
Before proceeding with this PR, do you have an example that shows that filtering the params makes any difference? When I ran the [prompt tuning notebook](https://github.com/huggingface/peft/blob/main/examples/causal_language_modeling/peft_prompt_tuning_clm.ipynb), making this change had no impact on GPU memory consumption. If it indeed makes no difference, I'd rather not add the change, since it could give readers the impression that it is necessary.

BenjaminBossan:
@Ben-Epstein WDYT about closing this PR?

Ben-Epstein:
Sure thing, we can close it out if it doesn't make a difference

PR #1286, opened 2023-12-20
Experiment fast lora train

BenjaminBossan:
_WIP, For internal testing_

Experimental branch to check if we can add a faster, feature-incomplete training backend for LoRA. To enable it, pass `fast_train_mode=True` to `LoraConfig`. From the help:

            If this is True, then the model will use specialized layers for faster training, especially in conjunction
            with torch.compile. These layers lack a couple of features, e.g. they cannot be merged or disabled, and
            there can only ever be a single adapter. However, for many training use cases, those features do not matter.
            Later, for example for inference, the adapters can be loaded as normal adapters, enabling all these
            missing features. A good workflow is thus to train the adapter using `fast_train_modle=True`, and then
            load the adapter using `fast_train_modle=False` (the default).

### DIFF
--- a/src/peft/tuners/lora/fast.py
+++ b/src/peft/tuners/lora/fast.py
@@ -1,0 +1,410 @@
+# coding=utf-8
+# Copyright 2023-present the HuggingFace Inc. team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import math
+import warnings
+from typing import Any, Optional, Union
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from transformers.pytorch_utils import Conv1D
+
+from peft.tuners.tuners_utils import BaseTunerLayer
+
+from .config import LoraConfig
+
+
+class FastLoraLayer(BaseTunerLayer):
+    # All names of layers that may contain (trainable) adapter weights
+    adapter_layer_names = ("lora_A", "lora_B", "lora_embedding_A", "lora_embedding_B")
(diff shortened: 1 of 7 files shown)


src/peft/tuners/lora/fast.py
```diff
@@ -0,0 +1,410 @@
...
+        if weight is not None:
+            # the layer is already completely initialized, this is an update
+            self.to(base_layer.weight.device, dtype=weight.dtype)
+        self.set_adapter(self.active_adapters)
+
+    def reset_lora_parameters(self, init_lora_weights):
+        if init_lora_weights is False:
+            return
+
+        nn.init.zeros_(self.lora_B.weight)
```
pacman100, line 269:
```suggestion
```

src/peft/tuners/lora/fast.py
```diff
@@ -0,0 +1,410 @@
...
+            max_norm=base_layer.max_norm,
+            norm_type=base_layer.norm_type,
+            scale_grad_by_freq=base_layer.scale_grad_by_freq,
+            sparse=base_layer.sparse,
+        )
+
+    def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
+        result = self.base_layer(x, *args, **kwargs)
+        after_A = self._embed(x, self.lora_embedding_A)
+        result += (after_A @ self.lora_embedding_B) * self.scaling
```
pacman100, line 305:
```suggestion
        embedding_A = self.lora_embedding_A.T
        embedding_B = self.lora_embedding_B.T
        result = self.base_layer(x, *args, **kwargs)
        after_A = self._embed(x, embedding_A)
        result += (after_A @ embedding_B) * self.scaling
```

pacman100 (commented):
Thank you @BenjaminBossan for putting great work into this! 

Let me know your thoughts based on the experiments done and offline discussion concerning this PR.

BenjaminBossan:
I will be closing this PR after internal discussion.

@pacman100 did a series of insightful experiments and could show that together with `torch.compile`, using this "fast" branch indeed can lead to significant speedups (> 2x) of training transformer models. However, this is only without using Flash Attention 2 -- with FA2, there was no difference. Additionally, `torch.compile` does not work with bitsandbytes (yet), i.e. no option to use QLoRA. Given that this PR would require us to maintain an alternative LoRA implementation at all times, the benefits thus don't outweigh the costs.

We can revisit this decision at a later point, but for now will keep this PR closed.

PR #1282, opened 2023-12-19
[`bnb`] Add bnb nightly workflow

younesbelkada:
As per title, this PR adds the possibility to run peft tests on bitsandbytes compiled from source. It is a follow-up work from #1275

PR #1287, opened 2023-12-20
[`bnb-nightly`] Address final comments

younesbelkada:
# What does this PR do?

Address final comments from @ydshieh on. https://github.com/huggingface/peft/pull/1282#pullrequestreview-1791413593 

I left few questions !

cc @ydshieh

### DIFF
--- a/.github/workflows/nightly-bnb.yml
+++ b/.github/workflows/nightly-bnb.yml
@@ -25,7 +25,7 @@ jobs:
       TEST_TYPE: "single_gpu_${{ matrix.docker-image-name }}"
     container:
       image: ${{ matrix.docker-image-name }}
-      options: --gpus all --shm-size "16gb" --ipc host -v /mnt/cache/.cache/huggingface:/mnt/cache/ -e NVIDIA_DISABLE_REQUIRE=true
+      options: --gpus all --shm-size "16gb" --ipc host -v /mnt/cache/.cache/huggingface:/mnt/cache/
     defaults:
       run:
         shell: bash
@@ -39,16 +39,19 @@ jobs:
           mkdir transformers-clone && git clone https://github.com/huggingface/transformers.git transformers-clone # rename to transformers clone to avoid modules conflict
 
       - name: Run examples on single GPU
+        if: always()
         run: |
           source activate peft
           make tests_examples_single_gpu_bnb
       
       - name: Run core tests on single GPU
+        if: always()
         run: |
           source activate peft
           make tests_core_single_gpu_bnb
 
       - name: Run transformers tests on single GPU
+        if: always()
         run: |
           source activate peft
           make transformers_tests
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-bnb.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       TEST_TYPE: "single_gpu_${{ matrix.docker-image-name }}"
     container:
       image: ${{ matrix.docker-image-name }}
-      options: --gpus all --shm-size "16gb" --ipc host -v /mnt/cache/.cache/huggingface:/mnt/cache/ -e NVIDIA_DISABLE_REQUIRE=true
+      options: --gpus all --shm-size "16gb" --ipc host -v /mnt/cache/.cache/huggingface:/mnt/cache/
```
younesbelkada, line 28: the env variable `NVIDIA_DISABLE_REQUIRE` is set on the very top so these are not needed

.github/workflows/nightly-bnb.yml
```diff
@@ -39,16 +39,19 @@ jobs:
           mkdir transformers-clone && git clone https://github.com/huggingface/transformers.git transformers-clone # rename to transformers clone to avoid modules conflict
 
       - name: Run examples on single GPU
+        if: always()
```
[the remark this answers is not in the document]
ydshieh: The `fail-fast:` (under `jobs.<job_id>.strategy`) applies to the jobs in the defined matrix. But here each job in the matrix has 3 testing steps in each job, but those steps won't be controlled by `jobs.<job_id>.strategy`.

That is my understanding

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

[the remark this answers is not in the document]
ydshieh: well, let me double check maybe!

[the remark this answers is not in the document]
ydshieh: See this example

https://github.com/huggingface/transformers/actions/runs/7286009365/job/19853943708

all 3 jobs (in the matrix) are run. But the 3rd step in each job are not.

.github/workflows/nightly-bnb.yml
```diff
@@ -39,16 +39,19 @@ jobs:
...
         run: |
           source activate peft
           make tests_examples_single_gpu_bnb
       
       - name: Run core tests on single GPU
+        if: always()
```
ydshieh, line 48: This is the quick and simple way. Just want to point out:

- `if: always()` will make the step uncancellable. A better approach would be `if: !cancelled()`

ydshieh: But I believe we only  want to run the 3 tests only if the step `Pip install` succeed. The condition gets a bit more complex in this case. OK for me if you are happy to keep the current version.

ydshieh (approved):
LGTM, just give you some info for you to make the final decision.

.github/workflows/nightly-bnb.yml
younesbelkada: Perfect thanks! I will stick to the current plan then :D

younesbelkada:
Thanks a lot for your review @ydshieh !

PR #1290, opened 2023-12-21
DOC: Improve target modules description

BenjaminBossan:
For LoRA and IA³, it is allowed to not specify a target module, in which case the correct layers are derived from the model architecture. This was not documented so far.


younesbelkada (approved):
Thanks a lot for improving the target modules description docs!

PR #1291, opened 2023-12-22
[BNB] Fix bnb dockerfile for latest version 

SunMarc:
# What does this PR do ?
This PR fixes the nightly docker file for the tests on the latest transformers. Right now, we are installing the latest transformers but we are launching the tests that are on the main branch. We need to checkout the corresponding branch (e.g. v4.36.3 if it is the version of transformers that is installed).

### DIFF
--- a/.github/workflows/nightly-bnb.yml
+++ b/.github/workflows/nightly-bnb.yml
@@ -88,6 +88,7 @@ jobs:
             echo "Checking out tag for Transformers version: v$transformers_version"
             git fetch --tags
             git checkout tags/v$transformers_version
+            cd .. 
 
       - name: Run core GPU tests on multi-gpu
         run: |
--- a/.github/workflows/nightly-bnb.yml
+++ b/.github/workflows/nightly-bnb.yml
@@ -82,6 +82,12 @@ jobs:
           pip install -e . --no-deps
           pip install pytest-reportlog pytest-cov parameterized datasets scipy einops
           mkdir transformers-clone && git clone https://github.com/huggingface/transformers.git transformers-clone
+          if [ "${{ matrix.docker-image-name }}" = "huggingface/peft-gpu-bnb-latest:latest" ]; then
+            cd transformers-clone
+            transformers_version=$(pip show transformers | grep '^Version:' | cut -d ' ' -f2 | sed 's/\.dev0//')
+            echo "Checking out tag for Transformers version: v$transformers_version"
+            git fetch --tags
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Makes a lot of sense to me, thanks a lot for fixing! 🚀

.github/workflows/nightly-bnb.yml
```diff
@@ -82,6 +82,12 @@ jobs:
           pip install -e . --no-deps
           pip install pytest-reportlog pytest-cov parameterized datasets scipy einops
           mkdir transformers-clone && git clone https://github.com/huggingface/transformers.git transformers-clone
+          if [ "${{ matrix.docker-image-name }}" = "huggingface/peft-gpu-bnb-latest:latest" ]; then
```
younesbelkada, line 85:
```suggestion
          if [ "${{ matrix.docker-image-name }}" == "huggingface/peft-gpu-bnb-latest:latest" ]; then
```
reference: https://github.com/huggingface/diffusers/blob/main/.github/workflows/pr_test_peft_backend.yml#L48

.github/workflows/nightly-bnb.yml
```diff
@@ -82,6 +82,13 @@ jobs:
           pip install -e . --no-deps
           pip install pytest-reportlog pytest-cov parameterized datasets scipy einops
           mkdir transformers-clone && git clone https://github.com/huggingface/transformers.git transformers-clone
+          if [ "${{ matrix.docker-image-name }}" == "huggingface/peft-gpu-bnb-latest:latest" ]; then
+            cd transformers-clone
+            transformers_version=$(pip show transformers | grep '^Version:' | cut -d ' ' -f2 | sed 's/\.dev0//')
```
BenjaminBossan, line 87: This assumes that there is never going to be a `.dev1`, right? I wonder if something along the lines of `git checkout $(git tag --list --sort=-v:refname | head -n 1)` would be more robust, but there might be a reason to do it this way that's eluding me, so no blocker.

BenjaminBossan (approved):
LGTM, thanks. Only a quick comment.

.github/workflows/nightly-bnb.yml
SunMarc: Normally, we should not even get a `.dev0` since we are just installing the latest version on pip. As for the tag, I thought about getting the latest tag but I wasn't certain whether the tags had additional uses beyond versioning. Let's see if this holds and,   if we came across issues, we will switch to your command.

PR #1294, opened 2023-12-23
Fix bnb lora layers not setting active adapter

tdrussell:
This appears to be a small oversight introduced with #1106. I was trying to load some loras onto LLMs using text-generation-webui, and noticed that the lora had no effect if the model was loaded with load_in_4bit. But non-quantized models still worked with the lora.

By adding this single line, it matches what is in the standard Linear lora layer. Now the quantized linear lora layers have the lora active by default.

### DIFF
--- a/src/peft/tuners/lora/bnb.py
+++ b/src/peft/tuners/lora/bnb.py
@@ -43,6 +43,7 @@ if is_bnb_available():
             super().__init__()
             LoraLayer.__init__(self, base_layer)
 
+            self._active_adapter = adapter_name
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Makes sense to me! 
Can you elaborate more on: 

> the lora layers had no effect when using `load_in_4bit`

BenjaminBossan (approved):
Thanks a lot for this bugfix, LGTM. @younesbelkada I think it means that there was no active adapter when using bnb layers.

younesbelkada:
Ok perfect, we can probably merge it then

PR #1295, opened 2023-12-24
Add an option 'ALL' to include all linear layers as target modules

SumanthRH:
# What does this PR do?

Adds a shorthand 'ALL' to include all linear/Conv1D layers as target modules. Code adapted [from the QLoRA repository](https://github.com/artidoro/qlora/blob/7f4e95a68dc076bea9b3a413d2b512eca6d004e5/qlora.py#L248).

## Why is this needed?
Currently, when `target_modules` is `None`, then PEFT will choose specific layers for different models based on the original paper for that adapter. For example, this is the query and value layers in the case of LoRA. However, this is far from the best configuration for LoRA. More recently, in QLoRA, it was found that adding LoRA to ALL the layers is important to match full fine-tuning performance. QLoRA-style training is very popular so I believe this feature will be helpful.

## Changes made
I've added a helper function in `tuners_utils.py` which will preprocess the `target_modules` argument specified by the user. I've also added some tests in `test_tuners_utils.py`


src/peft/tuners/tuners_utils.py
```diff
@@ -523,3 +526,33 @@ def inspect_matched_modules(tuner: BaseTuner, adapter_name: str = "default") ->
...
+            linear_classes = [bnb.nn.Linear4bit] + linear_classes if is_loaded_in_4bit else linear_classes
+
+        linear_module_names = set()
+        for name, module in model.named_modules():
+            # match with all linear classes.
+            for cls in linear_classes:
+                if isinstance(module, cls):
+                    names = name.split(".")
+                    linear_module_names.add(names[0] if len(names) == 1 else names[-1])
+        if "lm_head" in linear_module_names:  # needed for 16-bit
```
younesbelkada, line 555: What about using this method: https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/bitsandbytes.py#L267 ?

younesbelkada (commented):
Thanks a lot for this addition! Great tests!
One thing here is that this API won't be generalized for all models as some architectures have different names for `lm_head`. I proposed a suggestion below. 
I suggest to copy over the method `get_keys_to_not_convert` here in PEFT and use that method. 
GPTNeoX uses `embed_out` for the name of the lm head: https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt_neox/modeling_gpt_neox.py#L964 - you can use this model: https://huggingface.co/hf-internal-testing/tiny-random-GPTNeoXForCausalLM for testing

src/peft/tuners/tuners_utils.py
SumanthRH: Great idea! I didn't know about GPTNeoX's output layer! 

Regarding the `get_keys_to_not_convert` method: I don't think the entire method is useful, as it is meant for detecting keys/ child modules that  should not be quantized. With the `_maybe_include_all_linear_layers` function, we just need to make sure only internal transformer blocks are considered without the final classification layer if any.  I've copied over only the part that gets the key for the final output layer (also, I'm assuming only one such key will match - the `get_keys_to_not_convert` [returns a list](https://github.com/huggingface/transformers/blob/3b7675b2b844b02d4821b827871a21ad16dd446c/src/transformers/integrations/bitsandbytes.py#L295C13-L295C13) - I'm not aware of any architecture where there could be a clash - Correct me if i'm wrong).

SumanthRH:
@younesbelkada Do take another look!

Improvements: 
- I've used some code from `get_keys_to_not_convert` so that different key values for the output layer are supported.
-  I added the current shorthand value "ALL" in constants.py - not sure if there's a better way to handle this. 
- Added GPTNeoX to tests with some refactoring. Should be good now!

younesbelkada (approved):
Very nice work, thanks very much for working on this @SumanthRH !

src/peft/tuners/tuners_utils.py
```diff
@@ -523,3 +528,37 @@ def inspect_matched_modules(tuner: BaseTuner, adapter_name: str = "default") ->
...
+            import bitsandbytes as bnb
+
+            linear_classes = [bnb.nn.Linear8bitLt] + linear_classes if is_loaded_in_8bit else linear_classes
+            linear_classes = [bnb.nn.Linear4bit] + linear_classes if is_loaded_in_4bit else linear_classes
+
+        linear_module_names = set()
+        for name, module in model.named_modules():
+            # match with all linear classes.
+            for cls in linear_classes:
+                if isinstance(module, cls):
```
BenjaminBossan, line 554: You can pass multiple classes at once to `isinstance` instead of looping.

BenjaminBossan (commented):
Thanks for this PR. I haven't done a detailed review, but have some comments:

- How useful do you think the feature is. Do you have use cases where all linear layers have to be replaced? Note that we could already do something like `target_modules = [name for name, module in model.named_modules() if isinstance(module, linear_classes)]`.
- As a user, I could be confused by the `"ALL"` option, since it does not target all possible layers, only all linear layers. I assume you didn't want to call the option `"Linear"` because that could be ambiguous if there is a module with the name `"Linear"`. But we could use something like `"<Linear>"` or so, since that would not be a legal name for a module, so it's unambiguous. WDYT?
- IIUC, this method works by modifying the peft config and overriding the value of `target_modules`. I wonder if that is the best way. As a user, if I pass a certain config, I would generally assume that it stays as it is. Instead of overriding the config, would it be better to leave it and then, when we check if we should add the LoRA adapters to a certain layer, we change the logic to match the type instead of name.

src/peft/tuners/tuners_utils.py
SumanthRH: Fixed! Actually didn't know about this :p

SumanthRH:
> * How useful do you think the feature is. Do you have use cases where all linear layers have to be replaced? Note that we could already do something like `target_modules = [name for name, module in model.named_modules() if isinstance(module, linear_classes)]`.

I think this is a popular usecase, because QLoRA-style training is popular and effective. This configuration, as of 2023 atleast, is the best one for LoRA models to match full finetuning. From the QLoRA paper:
"When using the standard practice of applying LoRA to query and value attention projection matrices...we are not able to replicate full finetuning performance for large base models......we find that the most critical LoRA hyperparameter is how many LoRA adapters are used in total and that LoRA on all linear transformer block layers are required to match full finetuning performance."

As such, there are still some experiments to be made to see if we can push this further (for example, adding LoRA to the last output layer as well), but I feel PEFT should support this since it's the best LoRA setting right now.

 In terms of modifications, it isn't as simple as the one-liner you provided because 1) the user needs to provide all linear classes (with `bnb` and `Conv1D` layers if needed), 2) the user should also exclude the final output layer if it exists. This is basically what the current function does right now (with one extra step to get the base name of the module). 

> * As a user, I could be confused by the `"ALL"` option, since it does not target all possible layers, only all linear layers. I assume you didn't want to call the option `"Linear"` because that could be ambiguous if there is a module with the name `"Linear"`. But we could use something like `"<Linear>"` or so, since that would not be a legal name for a module, so it's unambiguous. WDYT?

Right as always :) I was only focused on language models with Linear layers, in which case "ALL" is a good name. I'm not a fan of mixed case or having an uncommon symbol like `<`, so I've changed it to "all_linear" now, hope that works!

> * IIUC, this method works by modifying the peft config and overriding the value of `target_modules`. I wonder if that is the best way. As a user, if I pass a certain config, I would generally assume that it stays as it is. Instead of overriding the config, would it be better to leave it and then, when we check if we should add the LoRA adapters to a certain layer, we change the logic to match the type instead of name.

Hmm this is a good point. If the user passed in "all_linear", it will get secretly rewritten with a new list of target modules, and, for example, when they save the config/ model with `model.save_pretrained`, they might see an unexpected entry. In terms of correctness it is fine though so not sure how problematic it is.

BenjaminBossan (changes requested):
> I think this is a popular usecase, because QLoRA-style training is popular and effective. This configuration, as of 2023 atleast, is the best one for LoRA models to match full finetuning.

Makes sense.

> In terms of modifications, it isn't as simple as the one-liner you provided

Yes, true, the snippet would have to rather be:

```python
linear_classes = ...
target_modules = [name for name, module in model.named_modules() if isinstance(module, linear_classes) and name != <NAME-HEAD>]
```

Still, I'm fine with adding this convenience function.

> I'm not a fan of mixed case or having an uncommon symbol like <, so I've changed it to "all_linear" now, hope that works!

Not a fan of special chars either, but the issue is still that if a model has a module with the name "all_linear", this would be ambiguous, only special chars can prevent this. How about "all-linear"? Off the top of my head, I can't think of any potential conflicts here.

Regarding whether to modify the config or not:

> this change mirrors what _prepare_adapter_config which updates the peft_config directly if target_modules is None

Yes, there is precedence, but I don't like the precedence either :)

Let's think this design through. If we modify the config as proposed in this PR, these points come to mind:

1. There is a bit of a surprise because the saved config is not the same as the one indicated by the user.
2. The saved config transparently shows what layers were adapted.
3. If we fix something in this logic (say, there is a bug with some layers matching incorrectly), a user who loads the config (with the explicit `target_modules`) will get different results from a user who uses a "fresh" config (with implicit `target_modules`). This can be bad (bug is not fixed) or good (reproducibility).
4. It works directly for other tuner methods like IA³ because we don't need to touch `_create_and_replace`, which is different for each tuner method.

In sum, there are pros and cons for both. I'm fine with the current approach (but let's document it).

src/peft/tuners/tuners_utils.py
```diff
@@ -523,3 +528,41 @@ def inspect_matched_modules(tuner: BaseTuner, adapter_name: str = "default") ->
...
+        for name, module in model.named_modules():
+            # match with all linear classes.
+            if isinstance(module, linear_classes):
+                names = name.split(".")
+                linear_module_names.add(names[-1])  # get the base name
+
+        # ignore the last classification head for text generation models
+        output_emb = model.get_output_embeddings()
+        if output_emb is not None:
+            last_module_name = [name for name, module in model.named_modules() if id(module) == id(output_emb)][0]
```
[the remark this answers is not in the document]
SumanthRH: not required indeed!

src/peft/tuners/tuners_utils.py
```diff
@@ -523,3 +528,41 @@ def inspect_matched_modules(tuner: BaseTuner, adapter_name: str = "default") ->
...
+            if last_module_name in linear_module_names:
```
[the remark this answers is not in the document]
SumanthRH: You're right! this is more pythonic :)

tests/test_tuners_utils.py
```diff
@@ -175,3 +225,41 @@ def test_feedforward_matching_ia3(self):
...
+            model_id, model, initial_target_modules, expected_target_modules
+        )
+
+    def _check_match_with_expected_target_modules(
+        self, model_id, model, initial_target_modules, expected_target_modules
+    ):
+        """
+        Helper function for the test for `_maybe_include_all_linear_layers`
+        """
+        lora_config = LoraConfig(base_model_name_or_path=model_id, target_modules=initial_target_modules)
```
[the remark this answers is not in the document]
SumanthRH: Added!

src/peft/tuners/tuners_utils.py
```diff
@@ -523,3 +528,41 @@ def inspect_matched_modules(tuner: BaseTuner, adapter_name: str = "default") ->
...
             module_dict["unmatched"].append(key)
     return module_dict
+
+
+def _maybe_include_all_linear_layers(peft_config: PeftConfig, model: nn.Module) -> PeftConfig:
+    """
+    Helper function to update `target_modules` to all linear/Conv1D layers if provided as 'all_linear'. Adapted from
+    the QLoRA repository: https://github.com/artidoro/qlora/blob/main/qlora.py
+    """
+    # if `target_modules` is a string, forgive case issues and match
```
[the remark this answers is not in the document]
SumanthRH: Have improved the comment!

src/peft/tuners/tuners_utils.py
```diff
@@ -523,3 +528,41 @@ def inspect_matched_modules(tuner: BaseTuner, adapter_name: str = "default") ->
...
+    if (
```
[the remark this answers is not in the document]
SumanthRH: Updated!

tests/test_tuners_utils.py
```diff
@@ -175,3 +225,41 @@ def test_feedforward_matching_ia3(self):
...
+        new_lora_config = _maybe_include_all_linear_layers(lora_config, model)
+        if isinstance(expected_target_modules, list):
```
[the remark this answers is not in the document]
SumanthRH: Hmm I think that does make sense, if in the future we decided to refactor this a bit. I've made this update! However, I still feel there should be some test for corner cases for the internal function `_maybe_include_all_linear_layers`. For example, in the case where we want `target_modules` to remain unchanged. I've separated those cases out now where I specifically look at the internals.

src/peft/tuners/ia3/config.py
```diff
@@ -49,7 +49,7 @@ class IA3Config(PeftConfig):
         metadata={
             "help": (
                 "List of module names or regex expression of the module names to replace with (IA)³."
-                "For example, ['q', 'v'] or '.*decoder.*(SelfAttention|EncDecAttention).*(q|v)$'. "
+                "For example, ['q', 'v'] or '.*decoder.*(SelfAttention|EncDecAttention).*(q|v)$'. Can also be a wildcard 'all_linear' to match all linear layers"
```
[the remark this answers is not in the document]
SumanthRH: Oh this didn't pop up with `make quality`, but I'll make it consistent anyway

src/peft/tuners/ia3/config.py
```diff
@@ -28,9 +28,9 @@ class IA3Config(PeftConfig):
     Args:
         target_modules (`Union[List[str],str]`):
             The names of the modules to apply (IA)³ to. If this is specified, only the modules with the specified names
-            will be replaced. If this is not specified, modules will be chosen according to the model architecture. If
-            the architecture is not known, an error will be raised -- in this case, you should specify the target
-            modules manually.
+            will be replaced. If this is specified as 'all_linear', then all linear/Conv1D modules are chosen. If this
```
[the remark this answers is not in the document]
SumanthRH: Done!

src/peft/tuners/tuners_utils.py
```diff
@@ -523,3 +528,41 @@ def inspect_matched_modules(tuner: BaseTuner, adapter_name: str = "default") ->
...
+
+        linear_module_names = set()
+        for name, module in model.named_modules():
+            # match with all linear classes.
+            if isinstance(module, linear_classes):
+                names = name.split(".")
+                linear_module_names.add(names[-1])  # get the base name
+
+        # ignore the last classification head for text generation models
+        output_emb = model.get_output_embeddings()
```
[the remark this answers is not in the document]
SumanthRH: You're right! However, on closer inspection, it looks like this function makes sense only for `PreTrainedModel` subclasses - firstly, because this is meant to be simple shorthand to mimic QLoRa-like training. Further, the more complicated `ModuleList` structures in `diffusers` models makes the matching function incorrect. In these cases, I'll need to save the full name of the module, instead of the base name, but then I'm also trying to save everything into the PeftConfig, so now the `target_modules` list is very long (can be 1000s) for LLMs. It made more sense for me to raise a ValueError in case the model is not a PreTrainedModel. I've added this now and have also added a test case for a diffusion model where this error is raised. 

This function now doesn't work for simple `nn.Module` models, but then it's trivial to get the same `all-linear` functionality for such models. So I think this should be fine!

LMK what you think!

src/peft/tuners/ia3/config.py
```diff
@@ -28,9 +28,10 @@ class IA3Config(PeftConfig):
     Args:
         target_modules (`Union[List[str],str]`):
             The names of the modules to apply (IA)³ to. If this is specified, only the modules with the specified names
-            will be replaced. If this is not specified, modules will be chosen according to the model architecture. If
-            the architecture is not known, an error will be raised -- in this case, you should specify the target
-            modules manually.
+            will be replaced. If this is specified as 'all-linear', then all linear/Conv1D modules are chosen,
```
BenjaminBossan, line 31: I just noticed that the same update is required also for: LoHa, LoKr, OFT. Sorry for missing it so far.

src/peft/tuners/tuners_utils.py
```diff
@@ -523,3 +529,48 @@ def inspect_matched_modules(tuner: BaseTuner, adapter_name: str = "default") ->
...
+    # if `target_modules` is a string, convert to lower case and check if it matches "all-linear"
+    if not (
+        isinstance(peft_config.target_modules, str)
+        and peft_config.target_modules.lower() == INCLUDE_LINEAR_LAYERS_SHORTHAND
+    ):
+        return peft_config
+
+    if not isinstance(model, PreTrainedModel):
+        raise ValueError(
+            f"Only instances of PreTrainedModel are supported for the '{INCLUDE_LINEAR_LAYERS_SHORTHAND}' flag"
```
BenjaminBossan, line 549:
```suggestion
            f"Only instances of PreTrainedModel support `target_modules={INCLUDE_LINEAR_LAYERS_SHORTHAND!r}`"
```

src/peft/tuners/tuners_utils.py
```diff
@@ -523,3 +529,48 @@ def inspect_matched_modules(tuner: BaseTuner, adapter_name: str = "default") ->
...
+        if isinstance(module, linear_classes):
+            names = name.rsplit(".", 1)[-1]  # get the base name
+            linear_module_names.add(names)
+
+    # ignore the last classification head for text generation models
+    output_emb = model.get_output_embeddings()
+    if output_emb is not None:
+        last_module_name = [name for name, module in model.named_modules() if module is output_emb][0]
+        linear_module_names -= {last_module_name}
+    peft_config.target_modules = list(linear_module_names)
```
BenjaminBossan, line 575:
```suggestion
    peft_config.target_modules = linear_module_names
```

`target_modules` should be a set:

https://github.com/huggingface/peft/blob/cbf346d962121a3d59568a615d6a3813357169d7/src/peft/tuners/lora/config.py#L205-L207

tests/test_tuners_utils.py
```diff
@@ -175,3 +231,80 @@ def test_feedforward_matching_ia3(self):
...
+
+    def _check_match_with_expected_target_modules(
+        self, model_id, model, config_cls, initial_target_modules, expected_target_modules
+    ):
+        """
+        Helper function for the test for `_maybe_include_all_linear_layers`
+        """
+        actual_config = config_cls(base_model_name_or_path=model_id, target_modules=initial_target_modules)
+        expected_config = config_cls(base_model_name_or_path=model_id, target_modules=expected_target_modules)
+        actual_model = get_peft_model(model, peft_config=actual_config)
```
BenjaminBossan, line 269: It would be a bit safer to pass a copy of `model`, since `get_peft_model` will mutate the model, so on 2nd call, it's not the same model anymore.

tests/test_tuners_utils.py
```diff
@@ -175,3 +231,80 @@ def test_feedforward_matching_ia3(self):
...
+            self.assertEqual(type(actual_module), type(expected_module))
+
+    def test_maybe_include_all_linear_layers_ia3_loha(self):
+        model_id, initial_target_modules, expected_target_modules = (
+            "HuggingFaceH4/tiny-random-LlamaForCausalLM",
+            INCLUDE_LINEAR_LAYERS_SHORTHAND,
+            ["k_proj", "v_proj", "q_proj", "o_proj", "down_proj", "up_proj", "gate_proj"],
+        )
+        model = AutoModelForCausalLM.from_pretrained(model_id)
+        config_cls = IA3Config
```
BenjaminBossan, line 284:
```suggestion
```

tests/test_tuners_utils.py
```diff
@@ -175,3 +231,80 @@ def test_feedforward_matching_ia3(self):
...
+    def test_maybe_include_all_linear_layers_lora(
+        self, model_id, model_type, initial_target_modules, expected_target_modules
+    ):
+        model = self.transformers_class_map[model_type].from_pretrained(model_id)
+        config_cls = LoraConfig
+        self._check_match_with_expected_target_modules(
+            model_id, model, config_cls, initial_target_modules, expected_target_modules
+        )
+
+    @parameterized.expand(BNB_TEST_CASES)
```
BenjaminBossan, line 245: Nit: Let's move the `@parameterized.expand` to be the innermost decorator.

BenjaminBossan (changes requested):
I agree with your statements on testing and checking for `PreTrainedModel`. The PR looks pretty good. I found some minor issues still, but hopefully those can be fixed quickly. Thanks for your continued work.

PS: Optionally, having an entry in the docs (for instance [here](https://github.com/huggingface/peft/blob/main/docs/source/developer_guides/lora.md#initialization)) could make this feature easier to discover.

tests/test_tuners_utils.py
SumanthRH: This requires some more work! Currently, what's happening is that the `parameterized.expand` creates multiple test cases with the given arguments, and then the GPU and BNB checks run for each test case. This is fine but a bit wasteful. However, if I add `parameterized.expand` to the end, then the test case isn't skipped! I believe there's some weird interaction between `unittest.skip`/`pytest.mark.skip` and `paramterized` because of which the test case doesn't get skipped (locally, I see errors because I'm using a Macbook). I'll need to look into this a bit more and will make another PR with a fix.

tests/test_tuners_utils.py
SumanthRH: You're right! Have used copies now!

src/peft/tuners/ia3/config.py
SumanthRH: I was saving this for a future PR since I haven't actually looked into LoHa, LoKr, etc,, but have made these changes as well now!

SumanthRH:
@BenjaminBossan thanks for the detailed review again! Hopefully this can be merged now!

tests/test_tuners_utils.py
BenjaminBossan: Weird. Never mind then, it's not super important, that's why I called it a "nit".

SumanthRH:
Following discussions in #1318, I've simplified the logic for bnb layers! Code is much cleaner now. The tests seem like overkill now, although I think the `bnb` tests can stay just to be sure.

BenjaminBossan (approved):
Thanks so much @SumanthRH, the PR looks pretty good, nothing to add from my side. Before merging, let's wait for @pacman100 (he's OOO this week), because of the considerable overlap with his PR.

>  The tests seem like overkill now, although I think the bnb tests can stay just to be sure.

They should stay for sure. This could help us in the future (e.g. if they decide that bnb layers no longer inherit from `Linear`, the tests would catch it).

pacman100 (approved):
Thank you @SumanthRH for adding `all-linear` option to target the linear layers following QLoRA recipe, very useful! Also, comprehensive tests! 🚀

PR #1301, opened 2023-12-26
Add LayerNorm tuning model

DTennant:
Hello, this is to add a PEFT method from this paper: https://arxiv.org/abs/2312.11420

It is basically just tuning the parameters of the LayerNorm layers in a transformer.
Please let me know how I can improve it.
Thanks!


BenjaminBossan:
Thanks for contributing this PR which adds a new fine-tuning method. 

I only did a quick check so far. Before doing a detailed review, could you please:

- Run `make style` on your code? 
- Provide some kind of example, docs or unit tests? (one of the 3 is sufficient)

src/peft/tuners/ln_tuning/model.py
```diff
@@ -0,0 +1,117 @@
...
+        >>> from peft import get_peft_model, TaskType, LNTuningConfig
+
+        >>> peft_config = LNTuningConfig(
+        ...     task_type=TaskType.CAUSAL_LM,
+        ... )
+
+        >>> model = AutoModelForCausalLM.from_pretrained("meta-llama/Llama-2-7b-hf")
+        >>> model = get_peft_model(model, peft_config)
+        >>> model.print_trainable_parameters()
+        ```
```
DTennant, line 65: Hello, I have run `make style` on my code, and these lines are two examples of how to use this new fine-tuning method, is this enough?
Thanks for your time!

DTennant:
Hello @BenjaminBossan, I have run `make style` on my code and list two examples of how to use this new method in the code.
Could you please advise me on how to proceed next?

BenjaminBossan:
Thanks for updating the PR. Before doing a deep dive, I'd like to ask some clarifying questions:

1. You are the first author of the paper, right? Do you have some example scripts that for instance reproduce experiments that you ran for your paper using the implementation in this PR? That would be an ideal example to include.
2. Do I understand correctly that this method consists only of marking a subset of layers specified on `target_modules` as trainable? It could be any layer, there is not even anything specific to layer norm, is that right?
3. Would saving and loading work as is? I don't think there is any logic right now to implement this.

BenjaminBossan:
Friendly ping @DTennant

DTennant:
Hello, sorry for the late reply, 

1. Yes, I am the first author on this paper, I will prepare this as an example.
2. Yes, the method itself is just marking layernorm layers as trainable.
3. Thanks for the reminder, I haven't implement a correct save and loading logic, I will work on implementing this.

BenjaminBossan:
Thanks for clarifying. 

Heads up: Before you put too much additional work into this, could you please wait while we're discussing this method internally? The reason is that this is not _really_ a PEFT method, as it's essentially only about selecting specific layers to fine-tune. In that sense, it's akin to only fine-tuning the last layer of a model. The code is not even specific to `LayerNorm`, in principle any type of layer added to `target_modules` could work.

Maybe it's still useful to have, especially if saving and loading works seamlessly, but we'll have to think about it first.

DTennant:
Thanks for the heads up! I look forward to the results of your internal discussion.

BenjaminBossan:
@DTennant After discussion with @pacman100, I think we can proceed with this PR, as it can be considered a PEFT-method. Thanks for your patience. Some points to consider for the implementation:

1. Let's ensure that saving and loading works correctly and that only the updated weights are saved.
2. How about making the naming more neutral? The paper works with LN layers, but the implementation can work with any layer type, right?
3. Right now, you're using `target_modules` to decide which layers to update. Then the `requires_grad` for these layers is set to `True`. A disadvantage of this approach is that if users want to go back to the original base model after training, it is not possible, since the weights were updated. However, this is a feature we would like to support in PEFT, so that users can call `unload`, `disable_adapters`, or even use multiple adapters at the same time. I haven't thought too deeply about this, but we could consider using `modules_to_save` instead of `target_modules`. That way, we create a copy of the original weights and update those copies, instead of updating the original weights. WDYT?

DTennant:
Hi @BenjaminBossan  Thanks for the update! Glad to hear this PR can proceed.

1. I will work on this to make saving and loading work.
2. Yes this can work with any layer, what are your suggestions about the naming scheme?
3. Thanks for the suggestion! I will work on this.

Hopefully, I can finish these things by the start of next week.

BenjaminBossan:
> what are your suggestions about the naming scheme?

Good question, not sure, naming things is hard :D. Maybe something along the lines of "selective layer tuning" or so would work semantically, but ideally we could rely on established nomenclature. Conceptually, this seems most similar to me the technique of training only the last few layer(s) for fine-tuning, as is for instance common for computer vision models. Anyway, the name should can be changed later when we find something fitting.

Note that you can still mention the idea of tuning the LN layers specifically, perhaps even provide a convenience function to retrieve the names of the `LayerNorm` layers of a given model, similar to [this](https://github.com/huggingface/peft/blob/6451cbd70c187dc7baeadeed701851ac7bdd6ec2/src/peft/tuners/tuners_utils.py#L600).

DTennant:
@BenjaminBossan Hi, I have make the code to work with the correct save and loading, I'm using target module to implement this selective layer tuning. Could you take a look? I'm happy to revise the code further.

BenjaminBossan:
@DTennant Could you please merge with, or rebase on, the latest `main`? If you have issues with some of the merge conflicts, please let me know and I'll take a look.

DTennant:
Hi @BenjaminBossan I have rebased on the lastest main now.

BenjaminBossan:
@DTennant Could you please merge with (or rebase on) the latest main branch and also run `make style`?

DTennant:
Thanks for the careful review! I will try and fix these.

BenjaminBossan:
@DTennant are you still working on this?

DTennant:
Hello, @BenjaminBossan sorry for the late update, I'm still working on this, the progress is blocked by the ECCV deadline in the past weeks (I need to finish my paper first...). I will try and update this pull request as soon as possible.

DTennant:
Hi @BenjaminBossan,
I have added the LNTuning method to the test script and made sure all tests passed for the new method.
Also I've add an example of using this method for causal language modeling and resolved most of your review comment.
Could you please take a look? I am happy to further change the code.

BenjaminBossan:
Thanks for the updates, I'll probably take a closer look next week. Meanwhile, could you please run `make style` to ensure that the quality check passes? You may have to update your ruff version to v0.2.2 to obtain the same results as the CI.

DTennant:
The errors in the CI didn't show up in my local testing... But I will investigate and fix it asap.

BenjaminBossan:
> The errors in the CI didn't show up in my local testing... But I will investigate and fix it asap.

Thanks.

.github/workflows/tests.yml
```diff
@@ -50,4 +50,4 @@ jobs:
           pip install -e .[test]
       - name: Test with pytest
         run: |
-          make test
+          make test_custom_model
```
BenjaminBossan, line 53: Could you please undo this change and the change to the `Makefile`? If it's to speed up your local testing, you can run `python -m pytest tests/ -k LayerNorm` and it should only run your new tests.

ImKeTT:
> The errors in the CI didn't show up in my local testing... But I will investigate and fix it asap.

This is really wired, because I'm facing the same issue on my end. All tests on LayerNorm have passed in my local testing with `pip install -e .[test]` in a new virual env. Do you have any idea what might be causing this? @BenjaminBossan thanks!

image

BenjaminBossan:
> This is really wired, because I'm facing the same issue on my end. All tests on LayerNorm have passed in my local testing with `pip install -e .[test]` in a new virual env. Do you have any idea what might be causing this?

The error seems to occur only on CPU, which is why it passes locally but fails on CI.

Try running this:

```sh
CUDA_VISIBLE_DEVICES='' pytest tests/test_custom_models.py -k test_disable_adapters_70_LayerNorm_1_LNTuning
CUDA_VISIBLE_DEVICES='' pytest tests/test_custom_models.py -k test_disable_adapters_38_Vanilla_MLP_6_LOHA
```

DTennant:
Hi @BenjaminBossan , I have updated the code, and now it should be able to pass the CI tests.

docs/source/package_reference/layernorm_tuning.md
```diff
@@ -0,0 +1,34 @@
...
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# LayerNorm Tuning
+
+LayerNorm Tuning ([LN Tuning](https://arxiv.org/abs/2312.11420)) is a PEFT method that only fine-tunes the parameters of the LayerNorm layers in a model.
```
BenjaminBossan, line 19:
```suggestion
LayerNorm Tuning ([LN Tuning](https://huggingface.co/papers/2312.11420)) is a PEFT method that only fine-tunes the parameters of the LayerNorm layers in a model.
```

src/peft/tuners/ln_tuning/config.py
```diff
@@ -0,0 +1,57 @@
...
+            specified names will be replaced. If this is not specified, modules will be chosen according to the model
+            architecture. If the architecture is not known, an error will be raised -- in this case, you should specify
+            the target modules manually.
+    """
+
+    target_modules: Optional[Union[list[str], str]] = field(
+        default=None,
+        metadata={
+            "help": (
+                "List of module names or regex expression of the module names to replace with LNTuning."
```
BenjaminBossan, line 40: This help text is a bit different from the docstring. Could you please adjust the two so that they're identical?

src/peft/tuners/ln_tuning/layer.py
```diff
@@ -0,0 +1,112 @@
...
+from copy import deepcopy
+from typing import List, Optional
+
+import torch
+import torch.nn as nn
+
+from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
+
+
+class SelectLayer(nn.Module, BaseTunerLayer):
```
BenjaminBossan, line 25: Since we ended up using `LNTuning` as a name, let's rename this layer to `LNTuningLayer`. Same for the `adapter_layer_names` and `select_new_layers` attributes below.

src/peft/tuners/ln_tuning/layer.py
```diff
@@ -0,0 +1,112 @@
...
+        else:
+            if len(self.active_adapters) != 1:
+                raise ValueError(
+                    "You can only use one adapter for SelectLayer Adapter."
+                    "Because SelectLayer selects one set of layers from the original arch."
+                )
+            active_adapter = self.active_adapters[0]
+            result = self.select_new_layers[active_adapter](x, *args, **kwargs)
+
+        return result
```
BenjaminBossan, line 112: Let's add a `__repr__` method for consistency with the other PEFT layers:

```python

    def __repr__(self) -> str:
        rep = super().__repr__()
        return "ln_tuning." + rep
```

src/peft/tuners/ln_tuning/layer.py
```diff
@@ -0,0 +1,112 @@
...
+    def merge(self, adapter_names: Optional[List[str]] = None):
+        adapter_names = check_adapters_to_merge(self, adapter_names)
+        if not adapter_names:
+            # no adapter to merge
+            return
+
+        if len(adapter_names) > 1:
+            raise ValueError(
+                "You can only use one adapter for SelectLayer Adapter. "
+                "Because SelectLayer selects one set of layers from the original arch."
```
BenjaminBossan, line 73: I think the message could be clarified a bit. Maybe something along the lines of `f"Trying to merge {len(adapter_names)} adapters, but LN tuning does not allow merging more than one adapter at a time"`.

src/peft/tuners/ln_tuning/layer.py
```diff
@@ -0,0 +1,112 @@
...
+            self.select_new_layers[adapter_names[0]],
+            self.base_layer,
+        )
+        self.merged_adapters.append(adapter_names[0])
+
+    def unmerge(self):
+        if not self.merged:
+            warnings.warn("Already unmerged. Nothing to do.")
+            return
+        merged_name = self.merged_adapters.pop(0)
```
BenjaminBossan, line 90: Let's add a comment here that popping one element (argument `0` should not be necessary) is sufficient because LN tuning does not allow merging more than one adapter at a time.

src/peft/tuners/ln_tuning/layer.py
```diff
@@ -0,0 +1,112 @@
...
+            if self.merged:
+                self.unmerge()
+            result = self.base_layer(x, *args, **kwargs)
+        elif self.merged:
+            result = self.base_layer(x, *args, **kwargs)
+        else:
+            if len(self.active_adapters) != 1:
+                raise ValueError(
+                    "You can only use one adapter for SelectLayer Adapter."
+                    "Because SelectLayer selects one set of layers from the original arch."
```
BenjaminBossan, line 107: I think the message could be clarified a bit. Maybe something along the lines of `f"Trying to run forward with {len(self.active_adapters)} active adapters, but LN tuning does not allow inference with more than one adapter at a time"`.

docs/source/package_reference/layernorm_tuning.md
```diff
@@ -0,0 +1,34 @@
+<!--Copyright 2023 The HuggingFace Team. All rights reserved.
```
BenjaminBossan, line 1: Let's update the date here and in the other files to 2024 :)

tests/test_custom_models.py
```diff
@@ -702,8 +751,13 @@ def test_disable_adapters(self, test_name, model_id, config_cls, config_kwargs):
...
 
-        assert not torch.allclose(outputs_before, outputs_after)
-        assert torch.allclose(outputs_before, outputs_disabled)
+        if self.torch_device == "cpu":
+            # LayerNorm is running float32 on cpu, so difference in outputs are smaller
+            rtol, atol = 1e-8, 1e-8
+        else:
+            rtol, atol = 1e-5, 1e-8
+        assert not torch.allclose(outputs_before, outputs_after, rtol=rtol, atol=atol)
+        assert torch.allclose(outputs_before, outputs_disabled), f"{config_kwargs}"
```
BenjaminBossan, line 760:
```suggestion
        assert torch.allclose(outputs_before, outputs_disabled)
```

tests/testing_common.py
```diff
@@ -665,8 +666,10 @@ def _test_safe_merge(self, model_id, config_cls, config_kwargs):
         model = get_peft_model(model, config).eval()
         logits_peft = model(**inputs)[0]
 
-        # sanity check that the logits are different
-        assert not torch.allclose(logits_base, logits_peft, atol=1e-6, rtol=1e-6)
+        # LNTuning do not change the logits
```
BenjaminBossan, line 669: To clarify a bit more:

```suggestion
        # Initializing with LN tuning cannot be configured to change the outputs (unlike init_lora_weights=False)
```

BenjaminBossan (changes requested):
Thanks for fixing the test, it now passes. I did another in-depth review and this PR is very close to being finished. I still found some small issues here and there, especially the notebook could be cleaned up a bit, but nothing major. Please check my comments.

ImKeTT:
Hi @BenjaminBossan, I've updated the code according to you review. Would you please take another look when you have time? Thanks!

BenjaminBossan (changes requested):
Thanks so much, this is 99% finished from my point of view. I just added a few smaller comments, please check them out.

Also, I saw that we don't have tests for LN tuning in `RequiresGradTester` yet ([found here](https://github.com/huggingface/peft/blob/56773b9a92b141111d65fe3548d0c30233358868/tests/test_custom_models.py#L1570)). Could you please add them? Just copy the existing tests and adjust them for LN tuning, it should be quite straightforward. Please ask if you have questions.

DTennant:
Hi @BenjaminBossan , I've added the test for `RequiresGradTester`, and fixed the remaining issues.

BenjaminBossan:
@DTennant Sorry, merging the VeRA PR created a bunch of merge conflicts but they should be quite easy to resolve. Could you please take a look? LMK if you have questions.

DTennant:
Hi @BenjaminBossan , I have resolved the conflicts, thanks!

DTennant:
The tests failed because of connection errors, maybe we can run it again?

BenjaminBossan:
> The tests failed because of connection errors, maybe we can run it again?

Yeah, it's especially bad today and might require a few restarts before all tests pass :( Is this PR ready for review otherwise?

DTennant:
> > The tests failed because of connection errors, maybe we can run it again?
> 
> Yeah, it's especially bad today and might require a few restarts before all tests pass :( Is this PR ready for review otherwise?

Yes, it's ready

BenjaminBossan (commented):
Looks pretty good from my perspective, thanks for the latest fixes. I have one small comment left and I think notebook could also be improved a little bit (see my open comments there). Once those are addressed, I'll ask for a 2nd reviewer from another maintainer and then this should be good to merge.

examples/causal_language_modeling/peft_ln_tuning_clm.ipynb
```diff
@@ -0,0 +1,1881 @@
...
+    "    print(outputs)\n",
+    "    print(tokenizer.batch_decode(outputs.detach().cpu().numpy(), skip_special_tokens=True))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c8f35152",
+   "metadata": {},
+   "source": [
+    "You can push model to hub or save model locally. \n",
```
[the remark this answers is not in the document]
DTennant: This cell seems to be rendered correctly on reviewnb?  https://app.reviewnb.com/huggingface/peft/pull/1301/

DTennant:
I've updated the code a bit. For the notebook, I'm not sure which comment I should try and address?

BenjaminBossan (approved):
Thanks for providing the last fixes.

Regarding the notebook, there is no blocker, just some nice to haves, like adding a few more sentences to explain what's happening (assume that the user knows very little or nothing about LN tuning), some references, avoid unnecessary outputs like `next(iter(test_dataloader))`.

@pacman100 @younesbelkada This PR is ready from my point of view, I'd like to have a 2nd review before merging.

DTennant:
Thanks! I will try to find some time to polish the notebook over the weekend.

DTennant:
Hello, is there any updates for the review of the code? anything else we should try to improve on?

DTennant:
@BenjaminBossan Hi, I have merged the code

tests/test_custom_models.py
```diff
@@ -2484,6 +2543,97 @@ def test_requires_grad_boft_same_targets(self):
...
 
+    def test_requires_grad_lntuning_different_targets(self):
+        config0 = LNTuningConfig(
+            target_modules=["layernorm0"],
+        )
+        peft_model = get_peft_model(MLP_LayerNorm(), config0)
+
+        config1 = LNTuningConfig(
+            target_modules=["layernorm1"],
+            inference_mode=True,
```
pacman100, line 2554: if inference mode is enabled, then the adapter shouldn't be trainable, right?

tests/test_custom_models.py
```diff
@@ -2484,6 +2543,97 @@ def test_requires_grad_boft_same_targets(self):
...
+            "base_model.model.layernorm1.ln_tuning_layers.adapter1.bias",
+        )
+
+    def test_requires_grad_lntuning_same_targets(self):
+        config0 = LNTuningConfig(
+            target_modules=["layernorm0"],
+        )
+        peft_model = get_peft_model(MLP_LayerNorm(), config0)
+
+        config1 = LNTuningConfig(target_modules=["layernorm0"], inference_mode=True)
```
pacman100, line 2599: same comment as above.

src/peft/utils/save_and_load.py
```diff
@@ -302,6 +306,7 @@ def set_peft_model_state_dict(
             PeftType.OFT: "oft_",
             PeftType.POLY: "poly_",
             PeftType.BOFT: "boft_",
+            PeftType.LN_TUNING: "ln_tuning",
```
pacman100, line 309:
```suggestion
            PeftType.LN_TUNING: "ln_tuning_",
```

pacman100 (commented):
Thank you, @DTennant, for adding LayerNorm tuning method along with the necessary documentation and example, great job! 🚀

I've left a few comments, Thank you!

DTennant:
Hi @pacman100 , Thanks for the review, I've fixed the save_and_load. 
For the test, I'm following the tests of other custom models which also set `config1` to inference_mode=True, for example, https://github.com/DTennant/peft/blob/main/tests/test_custom_models.py#L2526 and https://github.com/DTennant/peft/blob/main/tests/test_custom_models.py#L2568
If I remove this inference_mode=True, the test will fail. But now the code's logic is the same as the tests for other methods like BOFT. So I'm not sure how I should change this, thanks.

pacman100:
@DTennant, I re-checked the logic of `set_adapter` and it sets the `requires_grad=True` for the active adapter. So, all good on that front.

pacman100 (approved):

BenjaminBossan:
:tada: Thanks a lot @DTennant for your nice work and patience.

DTennant:
Thanks!

PR #1302, opened 2023-12-26
fix fsdp auto wrap policy

pacman100:
### What does this PR do?
1. Fixes the FSDP Auto Wrap policy for PEFT models to use the default `_no_split_modules` when available.


younesbelkada (approved):
Thanks!

PR #1312, opened 2023-12-31
Clone for 8bit

athatheo:


BenjaminBossan:
Hi, I assume you want to add this line because you encounter an error without it? Could you please provide an example that replicates that error?

PR #1314, opened 2024-01-01
fix the embedding saving for adaption prompt

pacman100:
### What does this PR do?
1. For Adaption prompt, the embedding saving feature is currently broken. This PR fixes that.
2. Automate setting save_embedding_layers when embedding layer is resized during finetuning

### DIFF
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -135,7 +135,7 @@ def get_peft_model_state_dict(
     elif (
         save_embedding_layers == "auto"
         and hasattr(model, "config")
-        and config.base_model_name_or_path is not None
+        and config.base_model_name_or_path
         and model.config.vocab_size != AutoConfig.from_pretrained(config.base_model_name_or_path).vocab_size
     ):
         warnings.warn(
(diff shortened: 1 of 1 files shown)


src/peft/utils/save_and_load.py
```diff
@@ -130,14 +131,27 @@ def get_peft_model_state_dict(
     ):
         warnings.warn("Setting `save_embedding_layers` to `True` as embedding layers found in `target_modules`.")
         save_embedding_layers = True
+    # check if the vocab size of the base model is different from the vocab size of the finetuned model
+    elif (
+        save_embedding_layers == "auto"
+        and hasattr(model, "config")
+        and config.base_model_name_or_path
+        and model.config.vocab_size != AutoConfig.from_pretrained(config.base_model_name_or_path).vocab_size
+    ):
```
BenjaminBossan, line 140: I wonder if we should be super defensive here, just in case:

```python
    elif save_embedding_layers == "auto":
        vocab_size = getattr(getattr(model, config, None), "vocab_size", None)
        model_id = getattr(config, "base_model_name_or_path", None)
        if vocab_size and model_id and (vocab_size != AutoConfig.from_pretrained(model_id).vocab_size):
            ...
```

BenjaminBossan (commented):
Nice catch. I have a small suggestion to be on the super safe side. It would also be nice to have a test, but not a must.

src/peft/utils/save_and_load.py
pacman100: Done, Thank you for the suggestion, makes sense.

pacman100:
> It would also be nice to have a test, but not a must.

Will do it in a follow up PR

BenjaminBossan (approved):
Thanks, LGTM

PR #1317, opened 2024-01-03
fix diffusers tests

pacman100:
### What does this PR do?
1. fix diffusers tests 

```
FAILED tests/lora/test_lora_layers_peft.py::StableDiffusionXLLoRATests::test_simple_inference_with_text_unet_lora_save_load - ValueError: The checkpoint you are trying to load has model type `clip_text_model` but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.
```


BenjaminBossan (approved):
Nice find!

PR #1318, opened 2024-01-03
QOL improvements and doc updates

pacman100:
### What does this PR do?
1. Improves doc string for `TaskType`
2. Adds `cast_non_trainable_to_dtype` to cast all non-trainable parameters of the model to the given `dtype`.
    This is meant to reduce the GPU memory usage when using PEFT methods by using half-precision dtype for non-trainable parameters.


src/peft/utils/other.py
```diff
@@ -496,3 +496,39 @@ def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
         unique_id = storage_ptr(tensor)
 
     return tensor.device, unique_id, storage_size(tensor)
+
+
+# adapted from https://github.com/Bavest/fin-llama/blob/8ba68f4b49753358ffbe693d38f2461c031625e3/qlora.py#L229
+def get_linear_layer_names(model):
```
BenjaminBossan, line 502: Given that we have #1295, do you think we still need this option? At the very least, we could try to re-use most of the code between this function and that PR. For instance, this function does not consider bnb layers.

BenjaminBossan (commented):
Thanks for updating the docs and adding the convenience function. Regarding that, I have a question, see the comment.

src/peft/utils/other.py
pacman100: Given that bnb layers inherit from Linear layers, this function is enough to get the intended functionality. Will take a look at the other PR.

SumanthRH: Didn't now about bnb layers inheriting from Linear! Logic in #1295 was initially based on the special handling in the Qlora repo, but it's simplified now!

docs/source/developer_guides/lora.md
```diff
@@ -136,3 +136,16 @@ model.unload()
 # delete adapter
 model.delete_adapter("dpo")
 ```
+
+## Utilities
```
stevhliu, line 140:
```suggestion
## Finding linear layers
```

docs/source/developer_guides/lora.md
```diff
@@ -136,3 +136,16 @@ model.unload()
...
+
+Q-LoRA paper outlines that best performance is achieved by targetting all the Linear layers of the model instead of only targetting the query and value layers. To find all the linear layers to target, we provide a utility function `get_linear_layer_names`. Below is the sample usage of it:
```
stevhliu, line 142: Is targeting all the linear layers for best performance specific to QLoRA, or is it also true for LoRA in general? If it's the latter, maybe we can remove mentioning the QLoRA paper.

```suggestion
In the QLoRA paper, the best performance is achieved by targeting all the linear layers of the model instead of only targeting the query and value layers. To find all the linear layers to target, use the [`~utils.get_linear_layer_names`] function:
```

stevhliu (approved):
Thanks for the updates! I would also add a short Utilities section in the [LoRA API reference](https://huggingface.co/docs/peft/main/en/package_reference/lora) like this:

```md
## Utilities

[[autodoc]] utils.get_linear_layer_names
```

src/peft/utils/other.py
```diff
@@ -498,3 +498,23 @@ def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
         unique_id = storage_ptr(tensor)
 
     return tensor.device, unique_id, storage_size(tensor)
+
+
+def cast_non_trainable_to_dtype(model, dtype):
+    """
+    Cast all non-trainable parameters of the model to the given `dtype`. The trainable parameters are casted to full
```
BenjaminBossan, line 505:
```suggestion
    Cast all non-trainable parameters of the model to the given `dtype`. The trainable parameters are cast to full
```

BenjaminBossan: Maybe worth mentioning that dtype should typically be bf16 or fp16? Are there other dtypes that make sense here?

src/peft/__init__.py
```diff
@@ -83,5 +83,6 @@
     set_peft_model_state_dict,
     shift_tokens_right,
     load_peft_weights,
+    cast_non_trainable_to_dtype,
```
BenjaminBossan, line 86: We could update the section that I just recently added to the docs to use this function now, right?

https://github.com/huggingface/peft/blob/c6b28a22b86b2547cbf4443a9348a9e232f85c9e/docs/source/developer_guides/troubleshooting.md#L52-L54

BenjaminBossan (changes requested):
Thanks a lot for updating the PR. It is almost good to go, just found a typo and have two small comments.

src/peft/utils/other.py
```diff
@@ -498,3 +498,23 @@ def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
         unique_id = storage_ptr(tensor)
 
     return tensor.device, unique_id, storage_size(tensor)
+
+
+def cast_non_trainable_to_dtype(model, dtype):
```
BenjaminBossan, line 503: Just noticing that we may want to use a different name, since the trainable parameters will also be cast. The name could give the impression that trainable parameters are left untouched.

src/peft/__init__.py
pacman100: Done.

src/peft/utils/other.py
pacman100: Updated the PR to use `cast_mixed_precision_params`, let me know if you have better naming suggestions.

BenjaminBossan (approved):
Thanks a lot, LGTM.

PR #1319, opened 2024-01-04
Refactor dispatching logic of LoRA layers

BenjaminBossan:
This PR's goal is to simplify the logic for deciding which LoRA layer backend is being used when LoRA is applied to a target layer.

Originally, this refactor was done in #1286 which was about adding the "fast" backend for LoRA, but since that PR was closed, I moved the refactor to this dedicated PR.

## Motivation

Right, now, the `LoraModel._create_new_module` method has become quite complex and hard to read, spanning >100 lines:

https://github.com/huggingface/peft/blob/8665e2b5719faa4e4b91749ddec09442927b53e0/src/peft/tuners/lora/model.py#L235-L339

The reason for this is that method contains the logic for deciding which LoRA layer backend to use for all the different types of LoRA layers that we have, i.e. normal `Linear` layer, `Conv2d` layer, bnb layer, gptq, etc.

This PR greatly simplifies this method (30 LOC) and should make it easier to prevent bugs. It should also simplify adding further backends in the future.

## Description

I moved the logic for deciding which layer to match to the respective implementation of the layers. For example, in `lora/layer.py`, there is now a function called `dispatch_default`, whose responsibility it is to decide if an `Embedding` layer, `Conv2d` layer or `Linear` layer is the right match. Similarly, in `lora/bnb.py`, there are now the two functions `dispatch_bnb_8bit` and `dispatch_bnb_4bit` to decide what/if any bnb 8bit or 4bit layer should be matched. Same for the gptq and the megatron backend.

This way, the logic to decide what layer to match now resides next to the respective layers. The only thing that `LoraModel` now needs to do is to collect all the dispatching methods and use the first layer that matches.

Note that the logic to decide if a layer matches is 100% the same, just moved to a different place. Therefore, there should be no difference in the LoRA model being created.

Only LoRA was modified because the other tuners don't have different backends and thus this approach was not necessary for them. The only exception is IA³, which has normal and bnb backend. Since those are only two, it's not as complicated as for LoRA, but if this PR is accepted, I can refactor IA³ in a similar fashion.

## Other changes

- Removed the `optional_kwargs` argument from `_create_and_replace`, as it was an unnecessary indirection.
- Removed the `bias` argument from `kwargs`, as it was not used.

## Backwards compatibility

This should be fully backwards compatible, as the constructed LoRA model is 100% the same. If there are users that override `_create_new_module`, their code will probably break, but since this is a private method, we should be fine.

**Edit**: Also ran regression tests and they passed.

### DIFF
--- a/src/peft/tuners/adalora/model.py
+++ b/src/peft/tuners/adalora/model.py
@@ -110,24 +110,22 @@ class AdaLoraModel(LoraModel):
         target,
         target_name,
         parent,
-        **optional_kwargs,
+        current_key,
     ):
-        loaded_in_8bit = optional_kwargs.get("loaded_in_8bit", False)
-        loaded_in_4bit = optional_kwargs.get("loaded_in_4bit", False)
-        if (loaded_in_8bit or loaded_in_4bit) and not is_bnb_available():
-            raise ImportError(
-                "To use AdaLora with 8-bit quantization, please install the `bitsandbytes` package. "
-                "You can install it with `pip install bitsandbytes`."
-            )
         kwargs = {
             "r": lora_config.init_r,
             "lora_alpha": lora_config.lora_alpha,
             "lora_dropout": lora_config.lora_dropout,
             "fan_in_fan_out": lora_config.fan_in_fan_out,
             "init_lora_weights": lora_config.init_lora_weights,
-            "loaded_in_8bit": loaded_in_8bit,
-            "loaded_in_4bit": loaded_in_4bit,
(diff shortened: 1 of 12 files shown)


younesbelkada (approved):
This is much more cleaner indeed, thanks a lot for putting efforts in this refactor @BenjaminBossan and making the codebase cleaner!

pacman100 (approved):
Thank you @BenjaminBossan, super clean refactor! 🧹

PR #1330, opened 2024-01-08
ENH: Add attribute to show targeted module names

BenjaminBossan:
This is just a tiny convenience feature to help users understand which modules are being targeted by the adapter. This can be useful to quickly check if a complex regex works for `target_modules`.

Note: This should work for all adapters that use `BaseTuner`, so not only LoRA but also IA³, LoHa, etc. Only the first two were tested but that should be enough.

Usage:

```python
config = LoraConfig(target_modules=<COMPLICATED-REGEX>)
model = get_peft_mode(base_model, config)
print(model.targeted_module_names)
```


pacman100 (approved):
Thank you @BenjaminBossan for adding this feature to see all the targeted module names, LGTM!

PR #1332, opened 2024-01-09
[docs] Task guides

stevhliu:
This is a proof-of-concept and not ready to merge or review! I'm starting with the prompt-based methods as an example.

I feel like the task guides kind of take the focus away from how to apply a PEFT method and are more of a guide for how to fine-tune a model for a specific task. Most of the guides' content is about preprocessing and getting a dataset/model ready for a specific task and training it. Only a short section of the guide is dedicated to applying the PEFT method. I think we can point users to the Transformers docs for how-to fine for a specific task and/or link to our notebook collection in our PEFT Hub organization (I'll work on getting this updated and ready!). This way, the focus is less on the specific task and more on the specific PEFT method, and users can generalize this to other tasks. Another benefit is it makes it easier to maintain and keep the docs updated since we don't have to add to the docs for every *different task* versus PEFT method. This would keep the docs cleaner and more focused I think.

Let me know what you think @BenjaminBossan @pacman100 @younesbelkada!


docs/source/task_guides/prompt_based_methods.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+Create a preprocessing function that tokenizes the tweet text and labels, pad each label in a batch, concatenate the input text and labels, and create an attention mask. Then you'll loop through each example in a batch to pad the input ids, labels, and attention mask to a predefined maximum length before converting them to PyTorch tensors.
+
+```py
+import torch
+
+max_length = 64
+
+def preprocess_function(examples):
```
BenjaminBossan, line 77: Running this function raised a couple of errors because `text_column` and `label_column`  are undefined. I could fix this by making this change:
```suggestion
def preprocess_function(examples, text_column='Tweet text', label_column='text_label')
```

docs/source/task_guides/prompt_based_methods.md
```diff
@@ -0,0 +1,303 @@
...
+<hfoptions id="configurations">
+<hfoption id="p-tuning">
+
+[P-tuning](../conceptual_guides/prompting#p-tuning) adds a trainable embedding tensor where the prompt tokens can be added anywhere in the input sequence. Create a [`PromptEncoderConfig`] with the task type, the number of virtual tokens to add and learn, and the hidden size of the encoder for learning the prompt parameters.
+
+```py
+from peft import PromptEncoderConfig
+
+peft_config = PromptEncoderConfig(task_type="CAUSAL_LM", num_virtual_tokens=20, encoder_hidden_size=128)
+model = get_peft_model(model, peft_config)
```
BenjaminBossan, line 165: `get_peft_model` is not imported at this point, thus should be added either somewhere before or for each of the methods here and below.

docs/source/task_guides/prompt_based_methods.md
```diff
@@ -0,0 +1,303 @@
...
+    num_training_steps=(len(train_dataloader) * num_epochs),
+)
+```
+
+Move the model to the GPU and create a training loop that reports the loss and perplexity for each epoch.
+
+```py
+from tqdm import tqdm
+
+model = model.to("cuda")
```
BenjaminBossan, line 230:
```suggestion
device = "cuda"
model = model.to(device)
```

Since otherwise, `device` is undefined and raises an error below when calling `batch = {k: v.to(device) for k, v in batch.items()}`.

docs/source/task_guides/prompt_based_methods.md
```diff
@@ -0,0 +1,303 @@
...
+
+## Share your model
+
+Once training is complete, you can upload your model to the Hub with the [`~transformers.PreTrainedModel.push_to_hub`] method. You'll need to login to your Hugging Face account first and enter your token when prompted.
+
+```py
+from huggingface_hub import notebook_login
+
+notebook_login()
+peft_model_id = "your-name/bloomz-560-m-peft-method"
```
BenjaminBossan, line 273: How about doing something like:

```python
account = <your-hf-account-name>
peft_model_id = f"{account}/bloomz-560-m-peft-method"
model.push_to_hub(peft_model_id)
```

Then, further below, instead of `model = AutoPeftModelForCausalLM.from_pretrained("stevhliu/bloomz-560m-p-tuning").to("cuda")`, do `model = AutoPeftModelForCausalLM.from_pretrained(peft_model_id).to("cuda")`

docs/source/task_guides/prompt_based_methods.md
```diff
@@ -0,0 +1,303 @@
...
+
+```py
+from peft import AutoPeftModelForCausalLM
+
+model = AutoPeftModelForCausalLM.from_pretrained("stevhliu/bloomz-560m-p-tuning").to("cuda")
+tokenizer = AutoTokenizer.from_pretrained("bigscience/bloomz-560m")
+
+i = 15
+inputs = tokenizer(f'{text_column} : {dataset["test"][i]["Tweet text"]} Label : ', return_tensors="pt")
+print(dataset["test"][i]["Tweet text"])
```
BenjaminBossan, line 291:
```suggestion
inputs = tokenizer(f'{text_column} : {ds["test"][i]["Tweet text"]} Label : ', return_tensors="pt")
print(ds["test"][i]["Tweet text"])
```

Above, `ds` is used, it would be more consistent to use the same name here (else users get an error)

docs/source/task_guides/prompt_based_methods.md
```diff
@@ -0,0 +1,303 @@
...
+```py
+from peft import PromptEncoderConfig
+
+peft_config = PromptEncoderConfig(task_type="CAUSAL_LM", num_virtual_tokens=20, encoder_hidden_size=128)
+model = get_peft_model(model, peft_config)
+model.print_trainable_parameters()
+"trainable params: 300,288 || all params: 559,514,880 || trainable%: 0.05366935013417338"
+```
+
+</hfoption>
```
BenjaminBossan, line 170: Running the code with prefix-tuning or prompt-tuning resulted in errors for me. I had to add:

```python
classes = [k.replace("_", " ") for k in ds["train"].features["Label"].names]
ds = ds.map(
    lambda x: {"text_label": [classes[label] for label in x["Label"]]},
    batched=True,
    num_proc=1,
)
```

before calling `processed_ds = ds.map(...)`, which I just copied from the original docs. Probably it could be added to `preprocess_function` to avoid mapping twice.

docs/source/task_guides/prompt_based_methods.md
```diff
@@ -0,0 +1,303 @@
...
+<hfoption id="prefix tuning">
+
+[Prefix tuning](../conceptual_guides/prompting#prefix-tuning) adds task-specific parameters in all of the model layers which are optimized by a separate feed-forward network. Create a [`PrefixTuningConfig`] with the task type and number of virtual tokens to add and learn.
```
BenjaminBossan, line 173:
```suggestion
[Prefix tuning](../conceptual_guides/prompting#prefix-tuning) adds task-specific parameters in all of the model layers, which are optimized by a separate feed-forward network. Create a [`PrefixTuningConfig`] with the task type and number of virtual tokens to add and learn.
```

I wonder if we should have a comma here to disambiguate?

docs/source/task_guides/prompt_based_methods.md
```diff
@@ -0,0 +1,303 @@
...
+def preprocess_function(examples):
+    batch_size = len(examples[text_column])
+    inputs = [f"{text_column} : {x} Label : " for x in examples[text_column]]
+    targets = [str(x) for x in examples[label_column]]
+    model_inputs = tokenizer(inputs)
+    labels = tokenizer(targets)
+    for i in range(batch_size):
+        sample_input_ids = model_inputs["input_ids"][i]
+        label_input_ids = labels["input_ids"][i] + [tokenizer.pad_token_id]
+        # print(i, sample_input_ids, label_input_ids)
```
BenjaminBossan, line 86: remove

docs/source/task_guides/prompt_based_methods.md
```diff
@@ -0,0 +1,303 @@
...
+        model_inputs["input_ids"][i] = sample_input_ids + label_input_ids
+        labels["input_ids"][i] = [-100] * len(sample_input_ids) + label_input_ids
+        model_inputs["attention_mask"][i] = [1] * len(model_inputs["input_ids"][i])
+    # print(model_inputs)
```
BenjaminBossan, line 90: remove

BenjaminBossan (changes requested):
Super good, thanks! I like the idea of condensing the individual guides into a single one to highlight the commonalities and differences, and to refer to existing docs to avoid duplication.

I ran into a couple of minor errors, please check out my comments.

docs/source/task_guides/prompt_based_methods.md
stevhliu: Great idea! 👏

docs/source/task_guides/prompt_based_methods.md
```diff
@@ -0,0 +1,303 @@
...
+rendered properly in your Markdown viewer.
+
+-->
+
+# Prompt-based methods
+
+A prompt can describe a task or provide an example of the task you want the model to learn. Instead of manually creating these prompts, soft prompting methods add learnable parameters to the input embeddings that can be optimized for a specific task while keeping the pretrained model's parameters frozen. This makes it both faster and easier to finetune large language models (LLMs) for new downstream tasks.
+
+The PEFT library supports several types of prompting methods (p-tuning, prefix tuning, prompt tuning, multitask prompt tuning) and you can learn more about how these methods work conceptually in the [Soft prompts](../conceptual_guides/prompting) guide. If you're interested in applying these methods to other tasks, take a look at our collection of [notebooks](https://huggingface.co/spaces/PEFT/soft-prompting)!
+
```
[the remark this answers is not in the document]
stevhliu: I think it's ok to not add a new subsection here. Between the intro paragraphs and the concrete guide is the `<Tip>`, which belongs to the intro for setting expectations, and just a short setup to install the necessary libraries. Since it's so short, I don't think we need to break it down so granularly and it could be included in the intro as a transition to the code.

stevhliu:
Thanks for your excellent review as usual @BenjaminBossan! 🔥 I'll wait a bit for the other maintainers to chime in if they want, and if we're all on board, then the next steps would be:

1. Complete the PEFT Hub organization with all the notebooks so we can refer to them from the docs
2. Condense the LoRA guides
3. Add guides for methods that aren't included in the docs yet, but they're in the `peft/examples` folder

BenjaminBossan (approved):
Great. Let's see what the others have to say.

pacman100 (commented):
Thank you @stevhliu for consolidating the soft-prompt based methods in a single task guide. I really like the `hfoptions` based toggle to see the changes required between these methods while trying to solve a given task. 🔥🚀✨

Left a few comments/nits.

pacman100 (approved):
Thank you @stevhliu! 🚀

PR #1333, opened 2024-01-09
Fix bug when load the prompt tuning in inference.

yileld:
Prompt tuning did not need tokenizer when inference.

And it may encounter some bug if load tokenizer in some cases when inference.

(For my example, it will cant find the config.json for tokenizer in cache dir, and my development environment cant connect to the hugginface.)


pacman100 (approved):
Thank you @yileld for the fix, makes sense.

BenjaminBossan (approved):
Thanks.

Issue #1249, opened 2023-12-11
About the dtype of trainable params


hiyouga:
### System Info

peft 0.7.0
transformers 4.34.0
torch 2.0.1

### Who can help?

@pacman100 @younesbelkada @sayakpaul 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [X] My own task or dataset (give details below)

### Reproduction

```python
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer
from peft import PeftModel, TaskType, LoraConfig, get_peft_model

tok = AutoTokenizer.from_pretrained("llama2-7b")
model = AutoModelForCausalLM.from_pretrained("llama2-7b", torch_dtype=torch.float16, low_cpu_mem_usage=True)

lora_config = LoraConfig(
    task_type=TaskType.CAUSAL_LM,
    inference_mode=False,
    r=8,
    lora_alpha=16,
    lora_dropout=0.05,
    target_modules=["q_proj", "v_proj"],
    modules_to_save=["embed_tokens", "lm_head"]
)
peft_model = get_peft_model(model, lora_config)

for name, param in peft_model.named_parameters():
    print(name, param.dtype)
"""
base_model.model.model.layers.31.self_attn.q_proj.base_layer.weight torch.float16
base_model.model.model.layers.31.self_attn.q_proj.lora_A.default.weight torch.float16
base_model.model.model.layers.31.self_attn.q_proj.lora_B.default.weight torch.float16
base_model.model.model.layers.31.self_attn.k_proj.weight torch.float16
base_model.model.model.layers.31.self_attn.v_proj.base_layer.weight torch.float16
base_model.model.model.layers.31.self_attn.v_proj.lora_A.default.weight torch.float16
base_model.model.model.layers.31.self_attn.v_proj.lora_B.default.weight torch.float16
...
base_model.model.lm_head.original_module.weight torch.float16
base_model.model.lm_head.modules_to_save.default.weight torch.float16
"""
```

### Expected behavior

If we load the model with half-precision and use fp16 mixed precision training, it will throw "ValueError: Attempting to unscale FP16 gradients."

Should we manually cast them in float32?

BenjaminBossan:
Could you please provide the training code and the full error?

hiyouga:
I use a modified version of https://github.com/huggingface/transformers/blob/main/examples/pytorch/language-modeling/run_clm.py

[...]

Issue #341, opened 2023-04-20
modules_to_save: "ValueError: Attempting to unscale FP16 gradients"


jonathanasdf:
I'm trying to finetune llama with some expanded tokens using `resize_token_embeddings()` and passing `modules_to_save=['embed_tokens', 'lm_head']`, but it seems there is some misconfiguration

```
Traceback (most recent call last):
  File "/home/jonathanasdf/.local/lib/python3.10/site-packages/transformers/trainer.py", line 1662, in train
    return inner_training_loop(
  File "/home/jonathanasdf/.local/lib/python3.10/site-packages/transformers/trainer.py", line 1962, in _inner_training_loop
    self.scaler.unscale_(self.optimizer)
  File "/home/jonathanasdf/.local/lib/python3.10/site-packages/torch/cuda/amp/grad_scaler.py", line 284, in unscale_
    optimizer_state["found_inf_per_device"] = self._unscale_grads_(optimizer, inv_scale, found_inf, False)
  File "/home/jonathanasdf/.local/lib/python3.10/site-packages/torch/cuda/amp/grad_scaler.py", line 212, in _unscale_grads_
    raise ValueError("Attempting to unscale FP16 gradients.")
```

jonathanasdf:
So I think the problem above is some incompatibility between int8 and modules_to_save. Using float16 instead of int8 is fine.

But actually, it seems that after https://github.com/huggingface/peft/commit/c21afbe868734c0af8bd4577c4c7acdf366b96d1 setting modules_to_save for LoraConfig CAUSAL_LM doesn't do anything anymore. I had to revert to before that commit for the setting to make the modules trainable and saved in the checkpoint.

ZenBuilds:
yes, for now, if use a lower version of peft, it will report the same error as you.
But if upgrade peft to latest version(0.3.0.dev0 for now), I can't save my lora weights for the weight size will became 443 bytes unexpectly.
#https://github.com/oobabooga/text-generation-webui/issues/1270#issue-1669857486

DavidVillaHMD:
I am having the same problem here and not solved yet even I installed peft by the up-to-date repo.
Just like the author of this issue, I'm trying to finetune llama with some expanded tokens using resize_token_embeddings() and passing modules_to_save=['embed_tokens', 'lm_head'].
Here is what I am suffering:
1. The problem solved when I delete 'modules_to_save' in LoraConfig.
[...]

PR #1336, opened 2024-01-09
DOC Troubleshooting for unscaling error with fp16

BenjaminBossan:
Some users ran into the issue of trying to use a model loaded in float16 with mixed precision, e.g. these issues: #341, #1249. This PR documents a workaround to solve the issue.

I also added tests that demonstrate the issue, as well as the workaround.

## Notes

This is not strictly a PEFT issue, but more a general error when using AMP with float16. Still, since PEFT users encounter this sometimes, it is useful to document it.

When we discussed this issue in the past, I think we concluded that it's not as straightforward as PEFT automatically casting the weights to float32, though I cannot remember anymore what the drawbacks were.

In any case, should we ever add an automatic solution for this in PEFT, the added test should fail, which alerts us to the fact that we need to update the documentation.

### DIFF
--- a/docs/source/developer_guides/troubleshooting.md
+++ b/docs/source/developer_guides/troubleshooting.md
@@ -39,6 +39,25 @@ Installing PEFT from source is useful for keeping up with the latest development
 python -m pip install git+https://github.com/huggingface/peft
 ```
 
+## Training errors
+
+### Getting: ValueError: Attempting to unscale FP16 gradients
+
+This error probably occurred because the model was loaded with `torch_dtype=torch.float16` and then used in an automatic mixed precision (AMP) context, e.g. by setting `fp16=True` in the `Trainer` class from 🤗 Transformers. The reason is that when using AMP, trainable weights should never use fp16. To make this work without having to load the whole model in FP32, add the following snippet to your code:
+
+```python
+peft_model = get_peft_model(...)
+
+# add this:
+for param in model.parameters():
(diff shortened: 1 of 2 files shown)


younesbelkada (approved):
Makes sense, thanks for adding these clarifications!

pacman100 (approved):
Thank you @BenjaminBossan for adding this troubleshoot details regarding mixed precision finetuning with PEFT.

hiyouga:
Sorry I have a question, will training a model loaded in fp16 with PEFT lead to bad convergence? I saw an issue about it: https://github.com/huggingface/transformers/pull/28142

BenjaminBossan:
Yes, generally it is better to use AMP if possible.

hiyouga:
Sorry for disturbing you again, I might not properly describe my question. If I enable AMP and use the above workaround to convert the LoRA weights to fp32 while keeping the other parameters in fp16. Will it also lead to the unstable problem mentioned in https://github.com/huggingface/transformers/pull/28142? I am not willing to convert all of the parameters to fp32 since it might double the GPU memory usage.

BenjaminBossan:
I see what you mean. In that case, it should generally be fine, we only want to avoid having the trainable parameters in fp16, the other parameters should work fine. There is still a tiny chance that having fp16 instead of fp32 on those other parameters could lead to a noticeable loss of performance, which is always a risk when using lower precision. But empirically, AFAIK, it works well.

PR #1343, opened 2024-01-10
DOC: Update docstring for the config classes

BenjaminBossan:
Over time, the docstrings of the numerous config classes have not kept up to date with changes in the code. This PR updates the docstrings to reflect the current state of the code.

On top of that, multiple small updates have been made:

- Correct wrong or imprecise type annotations.
- More neutral wording of the docstring. E.g., say "adapter" instead of "LoRA". This makes it easier to copy&paste the docstrings between classes.
- Use same wording for shared arguments.
- Add missing arguments.
- Uniform formatting: Always a line break after the first line of the docstring (not mixed, as that can be confusing).
- Fix line lengths to be consistently at 120 characters.

### DIFF
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -52,8 +52,8 @@ class LoraConfig(PeftConfig):
         target_modules (`Optional[Union[List[str], str]]`):
             The names of the modules to apply the adapter to. If this is specified, only the modules with the specified
             names will be replaced. When passing a string, a regex match will be performed. When passing a list of
-            strings, either an exact match will be performed or it is checked if the name of the module ends with any of
-            the passed strings. If this is specified as 'all-linear', then all linear/Conv1D modules are chosen,
+            strings, either an exact match will be performed or it is checked if the name of the module ends with any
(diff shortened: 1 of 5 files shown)


src/peft/tuners/lora/config.py
```diff
@@ -225,5 +249,5 @@ def __post_init__(self):
                 raise ValueError("`loftq_config` must be specified when `init_lora_weights` is 'loftq'.")
 
         # convert loftq_config to dict
-        if self.loftq_config is not None and not isinstance(self.loftq_config, dict):
+        if self.loftq_config and not isinstance(self.loftq_config, dict):
```
younesbelkada, line 252: Just flagging out this change in case it is not intended but I think it is perfectly ok to have it

younesbelkada (approved):
Thanks for the clean up !

pacman100 (approved):
Thank you @BenjaminBossan for the doc updates!

src/peft/tuners/lora/config.py
BenjaminBossan: This is because `loftq_config` is type `Union[LoftQConfig, dict]` (by default an empty dict), so `None` is not an option. But even if a user passes `None`, the logic would still be sound, so I made this change.

PR #1348, opened 2024-01-11
New transformers caching ETA now v4.38

BenjaminBossan:
See #1252 for more context.

The initial idea was for transformers v4.37 to add the new caching to all architectures, but this was postponed to v4.38. The code needs to be adapted for prompt tuning not to break when transformers v4.37 is released.

Big thanks to Joao for the heads up.


younesbelkada (approved):
Makes sense, thanks!

gante (approved):

PR #1352, opened 2024-01-12
fix `prepare_inputs_for_generation` logic for Prompt Learning methods

pacman100:
### What does this PR do?
1. The current logic from PR https://github.com/huggingface/peft/pull/1252 is incorrect as explained in https://github.com/huggingface/peft/pull/1252#issuecomment-1888604086 and https://github.com/huggingface/peft/pull/1252#issuecomment-1888614334.
2. This PR fixes it.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -1155,7 +1155,7 @@ class PeftModelForCausalLM(PeftModel):
                 # In prompt learning methods, past key values are longer when compared to the `input_ids`.
                 # As such only consider the last input ids in the autogressive generation phase.
                 if model_kwargs["past_key_values"][0][0].shape[-2] >= model_kwargs["input_ids"].shape[1]:
-                    model_kwargs["input_ids"] = model_kwargs["input_ids"][:, -1]
+                    model_kwargs["input_ids"] = model_kwargs["input_ids"][:, -1:]
 
             if model_kwargs.get("attention_mask", None) is not None:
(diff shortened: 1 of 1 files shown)


gante:
(LGTM)

BenjaminBossan (approved):
Thanks for doing the hard work and figuring this out. LGTM.

Do you have an example where the current logic fails. We could add it as a test (in a later PR) to prevent regressions.

pacman100:
> Do you have an example where the current logic fails. 

The current logic doesn't fail in the sense of runtime errors but will produce wrong outputs as the initial input ids are again run through the model and they involve in self-attention with themselves. Will look into test in a follow up PR.

BenjaminBossan:
> The current logic doesn't fail in the sense of runtime errors but will produce wrong outputs as the initial input ids are again run through the model and they involve in self-attention with themselves. Will look into test in a follow up PR.

Ah, got it, thanks. That makes it indeed very tricky to test.

gante:
> The current logic doesn't fail in the sense of runtime errors but will produce wrong outputs as the initial input ids are again run through the model and they involve in self-attention with themselves. 

@pacman100 @BenjaminBossan for these cases, in generate-related tasks in `transformers`, we often test the some of the returned logits against hardcoded expected values (e.g. [here](https://github.com/huggingface/transformers/blob/07bdbebb48a9fe1e748348e4e14ae0b4659e54c4/tests/models/mistral/test_modeling_mistral.py#L480)). In some cases, when the point is to test the logic (and not so much a model architecture), we use a tiny dummy model.

PR #1355, opened 2024-01-12
FIX: Make merging of adapter weights idempotent

BenjaminBossan:
Right now, merging of adapters weights such as LoRA and IA³ is not idempotent. This means that if a user calls `merge` multiple times, the resulting weights will be different each time because the delta weights are added again and again.

This fix checks that only those adapters are merged that are not yet merged. Also, gives a more precise warning:

- Say when there is nothing to merge.
- If there are some adapters to merge, only mention those

This bug is more subtle than it may seem at first, since we sometimes merge implicitly without the user necessarily being aware of it (e.g. when calling `merge_and_unload`). Therefore, this bug can occur quite easily, even if the user does not explicitly call `merge` twice in a row.

### DIFF
--- a/src/peft/tuners/adalora/layer.py
+++ b/src/peft/tuners/adalora/layer.py
@@ -20,6 +20,7 @@ import torch
 from torch import nn
 
 from peft.tuners.lora import LoraLayer
+from peft.tuners.tuners_utils import check_adapters_to_merge
 from peft.utils import transpose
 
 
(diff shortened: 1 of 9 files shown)


pacman100 (approved):
Thank you @BenjaminBossan for the fixes, makes sense. 🤗

younesbelkada (approved):
Very clean fix, thanks so much @BenjaminBossan !

PR #1357, opened 2024-01-12
Improve documentation for the `all-linear` flag

SumanthRH:
A small PR to add the new `all-linear` flag  (#1295) to PEFT docs.

### DIFF
--- a/docs/source/developer_guides/quantization.md
+++ b/docs/source/developer_guides/quantization.md
@@ -125,6 +125,12 @@ lora_config = LoraConfig(
 
 model = get_peft_model(model, lora_config)
 ```
+### QLoRA-style training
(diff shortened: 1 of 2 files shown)


docs/source/developer_guides/quantization.md
```diff
@@ -125,6 +125,12 @@ lora_config = LoraConfig(
 
 model = get_peft_model(model, lora_config)
 ```
+### QLoRA-style training
+QLoRA adds trainable weights to all the linear layers in the transformer architecture. Since the attribute names for these linear layers can vary across architectures, we provide a convenient flag `'all-linear'` for this setting:
```
younesbelkada, line 129:
```suggestion

QLoRA adds trainable weights to all the linear layers in the transformer architecture. Since the attribute names for these linear layers can vary across architectures, we provide a convenient flag `'all-linear'` for this setting:
```

younesbelkada (approved):
Thanks!

PR #1359, opened 2024-01-15
remove unnecessary code `len(unique_adapters)`

TaeWoo21:
A small PR that removes seemingly unnecessary code([peft/src/peft/tuners/lora/layer.py#L220](https://github.com/huggingface/peft/blob/918d4c6997de35c1cdcfe068e69eedda6cc10b80/src/peft/tuners/lora/layer.py#L220)).

https://github.com/huggingface/peft/blob/918d4c6997de35c1cdcfe068e69eedda6cc10b80/src/peft/tuners/lora/layer.py#L220


younesbelkada (approved):
Thanks for the clean up!

PR #1365, opened 2024-01-17
Added missing getattr dunder methods for mixed model

kovalexal:
Hello, PEFT team!

I tried to use the new interface for MixedModel and discovered that the current implementation of PeftMixedModel and MixedModel lacks getattr dunder methods (so we cannot use corresponding underlying implementations for e.g. device).

This PR just adds these missing methods.


kovalexal:
@younesbelkada @pacman100 your review is kindly requested!

pacman100 (approved):
Thank you @kovalexal for adding this, LGTM!

PR #1367, opened 2024-01-17
Handle resizing of embedding layers for AutoPeftModel

pacman100:
### What does this PR do?
1. In few scenarios the vocab of the model is extended to add special tokens like `<|im_start|>` ... In such scenarios during inference, the user needs to load the base model first and resize it the same way they did before training the model as shown below:
```python
peft_model_id = "smangrul/tinyllama_lora_norobots"
config = PeftConfig.from_pretrained(peft_model_id)
model = AutoModelForCausalLM.from_pretrained(config.base_model_name_or_path, load_in_4bit=True, device_map="auto")
tokenizer = AutoTokenizer.from_pretrained(peft_model_id)
model.resize_token_embeddings(len(tokenizer))
model = PeftModel.from_pretrained(model, peft_model_id, adapter_name="norobots")
```

2. This wasn't handled by `AutoPeftModel` classes. This PR automates the resizing of the base model if the tokenizer is saved with the adapter weights. The above code gets simplified to the one shown below:
```diff
from peft import AutoPeftModelForCausalLM, PeftConfig
from transformers import AutoTokenizer

peft_model_id = "smangrul/tinyllama_lora_norobots"
config = PeftConfig.from_pretrained(peft_model_id)
+ model = AutoPeftModelForCausalLM.from_pretrained(peft_model_id, adapter_name="norobots", load_in_4bit=True, device_map="auto")
tokenizer = AutoTokenizer.from_pretrained(peft_model_id)
- model.resize_token_embeddings(len(tokenizer))
- model = PeftModel.from_pretrained(model, peft_model_id, adapter_name="norobots")
```

- [x] Add tests

### DIFF
--- a/src/peft/auto.py
+++ b/src/peft/auto.py
@@ -41,7 +41,7 @@ from .peft_model import (
     PeftModelForSequenceClassification,
     PeftModelForTokenClassification,
 )
-from .utils.other import TOKENIZER_CONFIG_NAME
+from .utils.constants import TOKENIZER_CONFIG_NAME
 
 
 class _BaseAutoPeftModel:
--- a/src/peft/auto.py
+++ b/src/peft/auto.py
@@ -16,8 +16,10 @@
 from __future__ import annotations
 
 import importlib
+import os
 from typing import Optional
 
+from huggingface_hub import file_exists
 from transformers import (
     AutoModel,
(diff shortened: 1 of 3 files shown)


younesbelkada (approved):
Makes sense, thanks!

PR #1368, opened 2024-01-18
Add support for layer replication in LoRA

siddartha-RE:
This PR adds the ability to duplicate layers in a model according to a layer map
and then fine tune separate lora adapters for the layers post duplication. This
allows expanding a model to larger model and then fine tuning that model with
very little extra memory compared to the original smaller model.

### DIFF
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -89,15 +89,13 @@ config = LoraConfig(target_modules="all-linear", ...)
 
 ### Memory efficient Layer Replication with LoRA
 
-One of approach used to improve the performance of models is using model merging techniques is to expand a model by duplicating layers in the model to build a larger model from a pretrained model of a given size.
-For example increasing a 7B model to a 10B model as described in the [SOLAR](https://arxiv.org/abs/2312.15166). PEFT LoRA supports this kind of merge in a memory efficient manner that supports further fine-tuning
-using LoRA adapters attached to the layers post replication of the layers. The replicated layers do not take additional memory as they share the underlying weights so the only additional memory required is the
-memory for the adapter weights. To use this feature you would create a config with the `layer_replication` argument.
+An approach used to improve the performance of models is to expand a model by duplicating layers in the model to build a larger model from a pretrained model of a given size. For example increasing a 7B model to a 10B model as described in the [SOLAR](https://arxiv.org/abs/2312.15166) paper. PEFT LoRA supports this kind of expansion in a memory efficient manner that supports further fine-tuning using LoRA adapters attached to the layers post replication of the layers. The replicated layers do not take additional memory as they share the underlying weights so the only additional memory required is the memory for the adapter weights. To use this feature you would create a config with the `layer_replication` argument.
+
 ```py
 config = LoraConfig(layer_replication=[[0,4], [2,5]], ...)
 ```
-Given the original model had 5 layers `[0, 1, 2 ,3, 4]`, this would create a model with 7 layers arranged as `[0, 1, 2, 3, 2, 3, 4]`. This follows the mergekit pass through merge convention where sequences
-of layers specified as start inclusive and end exclusive tuples are stacked to build the final model. It is important to note that each layer in the final model gets its own distinct set of LoRA adpaters.
+
+Assuming the original model had 5 layers `[0, 1, 2 ,3, 4]`, this would create a model with 7 layers arranged as `[0, 1, 2, 3, 2, 3, 4]`. This follows the [mergekit](https://github.com/arcee-ai/mergekit) pass through merge convention where sequences of layers specified as start inclusive and end exclusive tuples are stacked to build the final model. Each layer in the final model gets its own distinct set of LoRA adpaters.
 
 [Fewshot-Metamath-OrcaVicuna-Mistral-10B](https://huggingface.co/abacusai/Fewshot-Metamath-OrcaVicuna-Mistral-10B) is an example of a model trained using this method on Mistral-7B expanded to 10B. The
 (adapter_config.json)[https://huggingface.co/abacusai/Fewshot-Metamath-OrcaVicuna-Mistral-10B/blob/main/adapter_config.json] shows a sample LoRA adapter config applying this method for fine-tuning.
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -87,6 +87,18 @@ The default LoRA settings in PEFT add trainable weights to the query and value l
 config = LoraConfig(target_modules="all-linear", ...)
 ```
 
+### Memory efficient Layer Replication with LoRA
+
+One of approach used to improve the performance of models is using model merging techniques is to expand a model by duplicating layers in the model to build a larger model from a pretrained model of a given size.
+For example increasing a 7B model to a 10B model as described in the [SOLAR](https://arxiv.org/abs/2312.15166). PEFT LoRA supports this kind of merge in a memory efficient manner that supports further fine-tuning
+using LoRA adapters attached to the layers post replication of the layers. The replicated layers do not take additional memory as they share the underlying weights so the only additional memory required is the
+memory for the adapter weights. To use this feature you would create a config with the `layer_replication` argument.
+```py
+config = LoraConfig(layer_replication=[[0,4], [2,5]], ...)
+```
+Given the original model had 5 layers `[0, 1, 2 ,3, 4]`, this would create a model with 7 layers arranged as `[0, 1, 2, 3, 2, 3, 4]`. This follows the mergekit pass through merge convention where sequences
+of layers specified as start inclusive and end exclusive tuples are stacked to build the final model. It is important to note that each layer in the final model gets its own distinct set of LoRA adpaters.
+
 ## Merge adapters
 
 While LoRA is significantly smaller and faster to train, you may encounter latency issues during inference due to separately loading the base model and the LoRA adapter. To eliminate latency, use the [`~LoraModel.merge_and_unload`] function to merge the adapter weights with the base model. This allows you to use the newly merged model as a standalone model. The [`~LoraModel.merge_and_unload`] function doesn't keep the adapter weights in memory.
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -99,6 +99,9 @@ config = LoraConfig(layer_replication=[[0,4], [2,5]], ...)
 Given the original model had 5 layers `[0, 1, 2 ,3, 4]`, this would create a model with 7 layers arranged as `[0, 1, 2, 3, 2, 3, 4]`. This follows the mergekit pass through merge convention where sequences
 of layers specified as start inclusive and end exclusive tuples are stacked to build the final model. It is important to note that each layer in the final model gets its own distinct set of LoRA adpaters.
 
+[Fewshot-Metamath-OrcaVicuna-Mistral-10B](https://huggingface.co/abacusai/Fewshot-Metamath-OrcaVicuna-Mistral-10B) is an example of a model trained using this method on Mistral-7B expanded to 10B. The
+(adapter_config.json)[https://huggingface.co/abacusai/Fewshot-Metamath-OrcaVicuna-Mistral-10B/blob/main/adapter_config.json] shows a sample LoRA adapter config applying this method for fine-tuning.
+
 ## Merge adapters
 
 While LoRA is significantly smaller and faster to train, you may encounter latency issues during inference due to separately loading the base model and the LoRA adapter. To eliminate latency, use the [`~LoraModel.merge_and_unload`] function to merge the adapter weights with the base model. This allows you to use the newly merged model as a standalone model. The [`~LoraModel.merge_and_unload`] function doesn't keep the adapter weights in memory.
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -108,6 +108,10 @@ class LoraConfig(PeftConfig):
             ranks. Right now, DoRA only supports non-quantized linear layers. DoRA introduces a bigger overhead than
             pure LoRA, so it is recommended to merge weights for inference. For more information, see
             https://arxiv.org/abs/2402.09353.
+        layer_replication(`List[Tuple[int, int]]):
+            Build a new stack of layers by stacking the original model layers according to the ranges specified.
+            This allows expanding (or shrinking) the model without duplicating the base model weights.
+            The new layers will all have separate LoRA adapters attached to them.
     """
 
     r: int = field(default=8, metadata={"help": "Lora attention dimension"})
@@ -249,7 +253,7 @@ class LoraConfig(PeftConfig):
         default=None,
         metadata={
             "help": (
-                "This enable using LoRA to effectively expand a model to a larger size by repeating some layers. "
+                "This enables using LoRA to effectively expand a model to a larger size by repeating some layers. "
                 "Base weights are shared so the memory usage is close to the original model."
                 "The format is a list of (start, end) pairs which specify the layer ranges to stack."
             )
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -108,7 +108,7 @@ class LoraConfig(PeftConfig):
             ranks. Right now, DoRA only supports non-quantized linear layers. DoRA introduces a bigger overhead than
             pure LoRA, so it is recommended to merge weights for inference. For more information, see
             https://arxiv.org/abs/2402.09353.
-        layer_replication(`List[Tuple[int, int]]):
+        layer_replication(`List[Tuple[int, int]]`):
             Build a new stack of layers by stacking the original model layers according to the ranges specified. This
             allows expanding (or shrinking) the model without duplicating the base model weights. The new layers will
             all have separate LoRA adapters attached to them.
@@ -254,7 +254,7 @@ class LoraConfig(PeftConfig):
         metadata={
             "help": (
                 "This enables using LoRA to effectively expand a transformer model to a larger size by repeating some layers. "
-                "The transformation handles models (currently Llama, Bert or Falcon compatible architecutres) with "
+                "The transformation handles models (currently Llama, Bert or Falcon compatible architectures) with "
                 "a module list in the model which it modifies to expand the number of modules. "
                 "Base weights are shared so the memory usage is close to the original model. The intended use is these base weights "
                 "remain fixed during finetuning but each layer has a separate LoRA adapter so the layers can be specialed via "
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -109,9 +109,9 @@ class LoraConfig(PeftConfig):
             pure LoRA, so it is recommended to merge weights for inference. For more information, see
             https://arxiv.org/abs/2402.09353.
         layer_replication(`List[Tuple[int, int]]):
-            Build a new stack of layers by stacking the original model layers according to the ranges specified.
-            This allows expanding (or shrinking) the model without duplicating the base model weights.
-            The new layers will all have separate LoRA adapters attached to them.
+            Build a new stack of layers by stacking the original model layers according to the ranges specified. This
+            allows expanding (or shrinking) the model without duplicating the base model weights. The new layers will
+            all have separate LoRA adapters attached to them.
     """
 
     r: int = field(default=8, metadata={"help": "Lora attention dimension"})
@@ -254,8 +254,9 @@ class LoraConfig(PeftConfig):
         metadata={
             "help": (
                 "This enables using LoRA to effectively expand a transformer model to a larger size by repeating some layers. "
-                "The transformation expects a `layers` module list in the model which it modifies to expand the number of modules. "
-                "Base weights are shared so the memory usage is close to the original model. The inteneded use is these base weights "
+                "The transformation handles models (currently Llama, Bert or Falcon compatible architecutres) with "
+                "a module list in the model which it modifies to expand the number of modules. "
+                "Base weights are shared so the memory usage is close to the original model. The intended use is these base weights "
                 "remain fixed during finetuning but each layer has a separate LoRA adapter so the layers can be specialed via "
                 "the adapter layers fit during fine tuning."
                 "The format is a list of [start, end) pairs which specify the layer ranges to stack. For example:\n"
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import List, Literal, Optional, Tuple, Union
+from typing import Literal, Optional, Union
 
 from peft.config import PeftConfig
 from peft.utils import PeftType
@@ -249,13 +249,17 @@ class LoraConfig(PeftConfig):
         },
     )
     # Enables replicating layers in a model to expand it to a larger model.
-    layer_replication: Optional[List[Tuple[int, int]]] = field(
+    layer_replication: Optional[list[tuple[int, int]]] = field(
         default=None,
         metadata={
             "help": (
-                "This enables using LoRA to effectively expand a model to a larger size by repeating some layers. "
+                "This enables using LoRA to effectively expand a transformer model to a larger size by repeating some layers. "
+                "The transformation expects a `layers` module list in the model which it modifies to expand the number of modules. "
                 "Base weights are shared so the memory usage is close to the original model."
-                "The format is a list of (start, end) pairs which specify the layer ranges to stack."
+                "The format is a list of [start, end) pairs which specify the layer ranges to stack. For example:\n"
+                "   original: `[0, 1, 2, 3, 4]`\n"
+                "   layer_replication: `[[0, 4], [2, 5]]`\n"
+                "   final: `[0, 1, 2, 3, 2, 3, 4]`"
             )
         }
     )
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import Literal, Optional, Union
+from typing import List, Literal, Optional, Tuple, Union
 
 from peft.config import PeftConfig
 from peft.utils import PeftType
@@ -243,6 +243,17 @@ class LoraConfig(PeftConfig):
                 "information, see  https://arxiv.org/abs/2402.09353."
             )
         },
+    )
+    # Enables replicating layers in a model to expand it to a larger model.
+    layer_replication: Optional[List[Tuple[int, int]]] = field(
+        default=None,
+        metadata={
+            "help": (
+                "This enable using LoRA to effectively expand a model to a larger size by repeating some layers. "
+                "Base weights are shared so the memory usage is close to the original model."
+                "The format is a list of (start, end) pairs which specify the layer ranges to stack."
+            )
+        }
     )
 
     def __post_init__(self):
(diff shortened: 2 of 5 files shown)


younesbelkada (commented):
Thanks @siddartha-RE for this feature! Can you elaborate more on its usage by providing some possible snippets of the API? Would you be happy to extend the current documentation and test suite to add that feature as well?

siddartha-RE:
Hi Younes,

definitely, sorry I should have marked it draft. I am in the process of
training and uploading a model to HF to demonstrate its use. I can use the
same config to build out an example of usage and figure out testing. I will
get it done in a day or two. I will also elaborate on the PR description.
If it is not clear the basic idea is to be able to fine tune models with
layers replicated (like has done with SOLAR / Goliath / ...) but with
separate lora adjustments for layers that have been duplicated. This allows
training very large models (for example 120B) with memory usage close to
that of the base 70B model.

- Siddartha

siddartha-RE:
I have created an example model using this PR:
https://huggingface.co/abacusai/Fewshot-Metamath-OrcaVicuna-Mistral-10B
and the model description explains the motivation for this PR.

I have also added a test to verify and added documentation in the code. I am not quite sure where to add additional docs. Should I modify one of the existing task docs that discuss LoRA (did not see an obvious candidate) or should I check a new example under `examples/causal_language_modelling`?

pacman100 (commented):
Thank you @siddartha-RE, I like the idea, the implementation is clear and I went over the POC experiment you have shared. I think this is a good feature to have. I would like to know the thoughts of @BenjaminBossan and @younesbelkada on this.

siddartha-RE:
Following up here with an ablation study:

| Model | Trainable Params | Train Loss | Eval Loss | GSM8K | TruthfulQA |
| :-----| ------: | ---------: | -------: | ----: | ---------: |
| Mistral 7B | 0 | - | - | 0.374 | 0.426 |
| Mistral 10B | 0 | - | - | 0.290 | 0.407 |
| Mistral 7B + LoRA r=12 | 31M | 0.412 | 0.366 | 0.514 | 0.499 |
| Mistral 10B + LoRA r=8 | 31M | 0.401 | 0.363 | 0.663 | 0.540 |

The 10B model was built by layer stacking the 7B model with the following config `[[0, 16], [8, 24], [24, 32]]` for a total of 48 layers initialized from the 32 layers in the original model. We ran LoRA finetuning on the original 32 layer model and 48 layer model (using this PR) and adjusting the `r` parameters so that trainable parameters were exactly the same in both cases. You can see that the 48 layers model clearly achieved better performance.

I think this is a pretty compelling approach for training large models without incurring the full memory cost.

BenjaminBossan (commented):
Thanks for providing this interesting new technique.

From my understanding of reading the code, this effectively modifies the base model by copying a couple of existing layers (with shared weights). Then, by applying different LoRA weights on these copies, it is ensured that these copied layers can learn different things.

From my understanding, the copy-functionality is not tied to LoRA at all. The same principle could be applied to IA³, LoHa, etc. Therefore, I think it would make sense to remove this functionality from `LoraModel` and make it a generic utility function. In practice, the usage would look something like this:

```python
base_model = AutoModelForCausalLM.from_pretrained(...)
layer_replication_map = [(0, 16), (8, 24), (16, 32)]
extended_model = replicate_layers(base_model, layer_replication_map)
config = LoraConfig(...)  # or IA3Config or ...
peft_model = get_peft_model(extended_model, config)
```

The advantage of this approach would be to make it generic and not overburden the existing LoRA implementation with something that is not fundamentally related to LoRA.

The disadvantage would be:

- as a user, I have to run the same extension code again when loading the trained adapter
- as it's not part of the `LoraConfig`, we cannot easily check if merging is allowed

IMO, this is still a better approach, but please tell me what you think about my proposal.

siddartha-RE:
> IMO, this is still a better approach, but please tell me what you think about my proposal.

I did consider how to keep it independent of LoRA. The thought I had was to make it just a separate adapter that can then be used with the multi adapter support that has recently landed to allow mixing with other adapters. The reasons I did not do it is because multi adapter seems to still be evolving and I didn't know how many adapters this would actually realistically work with. Second, I thought it was better to see if this method was generally interesting to the community before trying a more complex implementation.

One definite goal is to allow this to be loaded without any custom code so I would like the solution to work with any code that current just works with PeftConfig.from_pretrained(...)

BenjaminBossan:
> I did consider how to keep it independent of LoRA. The thought I had was to make it just a separate adapter that can then be used with the multi adapter support that has recently landed to allow mixing with other adapters.

I wonder if we even need a full new adapter class for this feature. What if we just make it a function that can be called on the base model before applying LoRA or another method?

siddartha-RE:
But say I want to submit a model to HF. How would I push a config that
would allow people to load the model without custom code to invoke the hook.

BenjaminBossan:
> But say I want to submit a model to HF. How would I push a config that would allow people to load the model without custom code to invoke the hook.

I definitely see your point of wanting to avoid requiring custom code. Loading from HF Hub would still work for the full model state dict, right? But of course, that would be a huge waste of space, so it'd be nice to have another solution. Still, I don't think the solution is to bundle this with LoRA, as the two are independent.

I could see the argument for having a separate adapter type with its own config, which could be combined with LoRA in a separate step, but I'm not sure if that's really better than having custom code, which comes down to a single function call :thinking: 

At the end of the day, if the other maintainers would rather have this bundled with LoRA, I can certainly be convinced :) My personal preference is still a separate function call though.

siddartha-RE:
I can certainly factor this into a separately usable function, but perhaps
also leave it as a functionality in LoRA which I see as one of the primary
ways to leverage this efficiently. Keep in mind its not just the issue of
downloading the weights. This approach even keeps GPU memory usage almost
the same as the base model. Effectively we have trained 120B models using
the memory of a 70B model. Because the lora adapters are distinct for all
the layers in the effectively 120B model it is really like have double the
number of independent layers.

BenjaminBossan:
> I can certainly factor this into a separately usable function, but perhaps also leave it as a functionality in LoRA which I see as one of the primary ways to leverage this efficiently.

This may be a good compromise. As a separate function, it should also make it easier to add it to the other methods. I'm a little wary that we're overburdening the LoRA options, which are already very many, with the features not being very modular or easy to compose, but that's an issue for another day.

> Keep in mind its not just the issue of downloading the weights. This approach even keeps GPU memory usage almost the same as the base model. Effectively we have trained 120B models using the memory of a 70B model. Because the lora adapters are distinct for all the layers in the effectively 120B model it is really like have double the number of independent layers.

Yes, this is great, but it's not directly related to integrating this into `LoraModel`, right?

siddartha-RE:
Ok will follow up and factor this into a separate function.

Yes, the second part was more the issue with limitations of exporting as a
merged base model.

BenjaminBossan:
@siddartha-RE Let us know once this is ready for review.

siddartha-RE:
> @siddartha-RE Let us know once this is ready for review.

I think this is ready. I did the refactoring of the function so it is in utils. LoRA just hooks into it.

src/peft/tuners/tuners_utils.py
```diff
@@ -665,3 +681,32 @@ def check_adapters_to_merge(module: BaseTunerLayer, adapter_names: Optional[list
...
+
+    return clone
+
+
+def replicate_layers(model: nn.Module, layer_map: List[Tuple[int, int]]):
+    new_layers = []
+    for start, end in layer_map:
+        for i in range(start, end):
+            current_idx = len(new_layers)
+            new_layers.append(clone_module(model.base_model.layers[i], share_weights=True))
```
BenjaminBossan, line 705: This assumes that the base model has a very specific architecture, right? I think some work should be done here to make this work with different architectures. If we cannot identify where to find the `ModuleList`, we should raise an error with a helpful message.

src/peft/tuners/tuners_utils.py
```diff
@@ -665,3 +681,32 @@ def check_adapters_to_merge(module: BaseTunerLayer, adapter_names: Optional[list
...
+            # This is a hack needed to work around the layer_idx introduced in HF transformers.
+            for submodule in new_layers[-1].modules():
+                if hasattr(submodule, 'layer_idx'):
```
BenjaminBossan, line 708: What happens if we don't find this attribute on any submodule? Should we raise an error or is it okay to just ignore this?

tests/test_decoder_models.py
```diff
@@ -302,3 +302,22 @@ def test_generate_adalora_no_dropout(self):
...
+            "lora_dropout": 0.0,
+            "layer_replication": [[0, 1], [0, 2], [1, 2]]
+        }
+        model = self.transformers_class.from_pretrained(model_id).to(self.torch_device)
+        config = LoraConfig(
+            base_model_name_or_path=model_id,
+            **config_kwargs,
+        )
+        model = get_peft_model(model, config)
+        self.assertEquals(4, len(model.base_model.model.model.layers), 'Expected 4 layers in adapted model.')
```
BenjaminBossan, line 320: We recently made a change to use plain `assert`s instead of `self.assert`. Could you please adjust these lines?

tests/test_decoder_models.py
```diff
@@ -302,3 +302,22 @@ def test_generate_adalora_no_dropout(self):
...
+        self.assertEquals(8, len([n for n, _ in model.named_parameters() if '.lora_A.' in n]))
```
BenjaminBossan, line 321: Let's add a comment to explain why we expect 8 here.

tests/test_decoder_models.py
```diff
@@ -302,3 +302,22 @@ def test_generate_adalora_no_dropout(self):
...
+            "target_modules": ['down_proj', 'up_proj'],
+            "task_type": "CAUSAL_LM",
+            "lora_dropout": 0.0,
+            "layer_replication": [[0, 1], [0, 2], [1, 2]]
+        }
+        model = self.transformers_class.from_pretrained(model_id).to(self.torch_device)
+        config = LoraConfig(
+            base_model_name_or_path=model_id,
+            **config_kwargs,
+        )
```
BenjaminBossan, line 318: Let's add a sanity check here that the number of layers is only 2 here, so that we can be sure with the check below that really some layers were replicated.

BenjaminBossan (changes requested):
Thanks for the update. I have some comments, please take a look.

src/peft/tuners/lora/config.py
```diff
@@ -224,6 +228,17 @@ class LoraConfig(PeftConfig):
...
         },
     )
+    # Enables replicating layers in a model to expand it to a larger model.
+    layer_replication: Optional[List[Tuple[int, int]]] = field(
+        default=None,
+        metadata={
+            "help": (
+                "This enables using LoRA to effectively expand a model to a larger size by repeating some layers. "
+                "Base weights are shared so the memory usage is close to the original model."
+                "The format is a list of (start, end) pairs which specify the layer ranges to stack."
```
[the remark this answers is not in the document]
siddartha-RE: Clarified that it applies to transformers with a layers attribute and added a concrete example of usage.

src/peft/tuners/lora/model.py
```diff
@@ -327,6 +346,16 @@ def set_adapter(self, adapter_name: str | list[str]) -> None:
...
         self.active_adapter = adapter_name
 
+    def _check_merge_allowed(self):
+        if getattr(self.model, "quantization_method", None) == "gptq":
+            raise ValueError("Cannot merge LORA layers when the model is gptq quantized")
+        if self.peft_config.get('layer_replication'):
+            raise ValueError("Cannot merge LORA layers when base model layers are replicated")
+
+    def merge_adapter(self, adapter_names: Optional[list[str]] = None) -> None:
+        self._check_merge_allowed()
```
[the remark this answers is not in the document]
siddartha-RE: Added to base

src/peft/tuners/lora/model.py
```diff
@@ -327,6 +346,16 @@ def set_adapter(self, adapter_name: str | list[str]) -> None:
                 module.set_adapter(adapter_name)
         self.active_adapter = adapter_name
 
+    def _check_merge_allowed(self):
```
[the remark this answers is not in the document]
siddartha-RE: Done

src/peft/tuners/lora/config.py
```diff
@@ -224,6 +228,17 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    # Enables replicating layers in a model to expand it to a larger model.
+    layer_replication: Optional[List[Tuple[int, int]]] = field(
```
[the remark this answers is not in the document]
siddartha-RE: Switched.

src/peft/tuners/tuners_utils.py
```diff
@@ -665,3 +681,32 @@ def check_adapters_to_merge(module: BaseTunerLayer, adapter_names: Optional[list
...
+            dst.register_parameter(name, param)
+
+    if share_weights:
+        for name, submodule in module.named_modules():
+            _share_weights(submodule, clone.get_submodule(name))
+
+    return clone
+
+
+def replicate_layers(model: nn.Module, layer_map: List[Tuple[int, int]]):
```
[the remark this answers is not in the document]
siddartha-RE: Done.

src/peft/tuners/tuners_utils.py
siddartha-RE: Done

src/peft/tuners/tuners_utils.py
siddartha-RE: I think it is ok to ignore it. I have opened an issue on avoiding this layer_idx attribute so that layers are not tied to their position in the transformers repo. Have not had a chance to send a PR. Some transformers don't need the layer idx. In fact, the layer idx is primarily used for inference optimization with the key value cache. It does not impact training.

tests/test_decoder_models.py
siddartha-RE: Done

tests/test_decoder_models.py
siddartha-RE: Done

src/peft/tuners/tuners_utils.py
```diff
@@ -665,3 +681,32 @@ def check_adapters_to_merge(module: BaseTunerLayer, adapter_names: Optional[list
             warnings.warn("All adapters are already merged, nothing to do.")
 
     return adapter_names
+
+
+def clone_module(module: nn.Module, share_weights=False):
```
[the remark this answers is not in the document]
BenjaminBossan: Could you please add a short description?

tests/test_decoder_models.py
```diff
@@ -302,3 +302,25 @@ def test_generate_adalora_no_dropout(self):
...
+        assert 2 == len(model.model.layers), 'Expected 2 layers in original model.'
```
BenjaminBossan, line 319: Could you please switch the order of the check, i.e. `assert len(model.model.layers) == 2`, I find it confusing otherwise. It's like writing `if "__main__" == __name__` ;-)

src/peft/tuners/lora/config.py
```diff
@@ -224,6 +228,21 @@ class LoraConfig(PeftConfig):
...
     )
+    # Enables replicating layers in a model to expand it to a larger model.
+    layer_replication: Optional[list[tuple[int, int]]] = field(
+        default=None,
+        metadata={
+            "help": (
+                "This enables using LoRA to effectively expand a transformer model to a larger size by repeating some layers. "
+                "The transformation expects a `layers` module list in the model which it modifies to expand the number of modules. "
+                "Base weights are shared so the memory usage is close to the original model."
+                "The format is a list of [start, end) pairs which specify the layer ranges to stack. For example:\n"
```
BenjaminBossan, line 239: I think it should be added here that these layers should be targeted with LoRA, otherwise it makes no sense to replicate the layers. Also, it would be great to link to some kind of resource to find out more about this. Since there is no paper, maybe the repo you posted earlier?

src/peft/tuners/tuners_utils.py
```diff
@@ -665,3 +688,43 @@ def check_adapters_to_merge(module: BaseTunerLayer, adapter_names: Optional[list
...
+    """Replicate layers in a transfomer model with weight sharing.
+    
+    This function looks for a module list attribute at model[(.model)*].layers 
+    and replicates the layers in the module list according to the layer map.
+    For example the map `[[0, 4], [2, 5]]` will take the set of layers `[0, 1, 2, 3, 4]`
+    and replace them with a module list containing `[0, 1, 2, 3, 2, 3, 4]`.
+    """
+    while hasattr(model, 'model'):
+        model = model.model
+    if not hasattr(model, 'layers'):
```
BenjaminBossan, line 717: I wonder if we could put more effort into accommodating more architectures. E.g. for BERT it's called `"layer"`, for bloomz and Falcon, it just seems to be `"h"`, etc.

src/peft/tuners/lora/config.py
```diff
@@ -224,6 +228,21 @@ class LoraConfig(PeftConfig):
...
+                "   original: `[0, 1, 2, 3, 4]`\n"
+                "   layer_replication: `[[0, 4], [2, 5]]`\n"
+                "   final: `[0, 1, 2, 3, 2, 3, 4]`"
```
BenjaminBossan, line 242: I still think this is not very intuitive to understand. My understanding is that this replaces the existing layers by copies of the existing layers, but whether any layers are replicated (i.e. appear more than once) is totally up to the config passed by the user, right? If I pass `[[0, 2], [3, 5]]`, for instance, I would only get `[0, 1, 3, 4]`, which wouldn't even make sense in this context, right?

Also, in this example I get `[0, 1, 2, 3, 2, 3, 4]` but does it make sense to have layer 2 follow after layer 3 at one point? Wouldn't it be better to have `[0, 1, 2, 2, 3, 3, 4]`? Yes, this can be achieved by passing different arguments, but I'd prefer to have an example that is somewhat realistic.

I wonder if we can find better ways to configure the layer replication, e.g. how about passing a dict whose keys indicate what layer to replicate and the value indicates how often to replicate them. So as an example:

```python
original = [0, 1, 2, 3, 4]
layer_replication = {1: 1, 2: 1, 3: 2}
final = [0, 1, 1, 2, 2, 3, 3, 3, 4]
```

This is a bit less flexible then what you proposed, but IMHO more intuitive. It would also ensure that the order of the layers is never switched around and no layer can ever be skipped completely.

Maybe there is another way that is better, I'm open to suggestion.

BenjaminBossan (changes requested):
I think we're getting close to finishing this PR. I have a few comments left, please check them out.

src/peft/tuners/lora/config.py
siddartha-RE: Before I address the other comments and fix the lint (sorry) I wanted to address this in some detail. The format for this argument was actually inspired by:
https://github.com/arcee-ai/mergekit
One style of merges involves this kind of layer interleaving. However, mergekit of course actually duplicates the weights and serializes a new model. I figured it would be familiar to those users. In practice there are models where this kind of layer repetition has worked. One well known example is SOLAR 10.7B which repeats some layers in the center.
https://arxiv.org/abs/2312.15166
One of the things SOLAR did was finetune the final 10.7B model. One of the experiments we have run is to do the same but only with LoRA adapters rather than a full fine tune.
The original model I linked:
https://huggingface.co/abacusai/Fewshot-Metamath-OrcaVicuna-Mistral-10B
does the same reusing layers.
I know it seems crazy that it works and even more surprisingly models built like this are coherent even without fine tuning. For example see this model:
https://huggingface.co/abacusai/TheProfessor-155b
This one interleaves layers from multiple different models to inject knowledge and without fine tuning is capable of generating very good outputs (if not particularly high scoring on eval tasks).

So I do think whatever config scheme is used it should support this kind of repetition.

BenjaminBossan: I didn't know that this format is inspired by mergekit. In that case, I can totally see why you chose that. It could be worth mentioning that the same format is used, and I still think the explanation could be clarified, but I'm fine with keeping this format.

tests/test_decoder_models.py
siddartha-RE: Done. Habit left over from C++. Constant on the left prevent typos like `a = 5` vs `5 = a`.

src/peft/tuners/tuners_utils.py
siddartha-RE: I added support for bert and falcon. But have not figured out testing each of these.  Llama (and mistral) are much more common now so I am not sure how many we want to try and add. I know I should also update the test to cover these. Its a bit of work since I need to figure out dummy models for each. Pushing an update but I will look into the tests.

src/peft/tuners/lora/config.py
siddartha-RE: Done

siddartha-RE:
Sorry about not running the code style checks. Ran the test but forgot to rerun make quality/style. Ran it as part of the latest commit.

BenjaminBossan (changes requested):
Thanks for the updates, I only have a few smaller comments left.

Also, I think it would be nice to add a section to the docs, perhaps [here](https://huggingface.co/docs/peft/v0.8.2/en/developer_guides/lora#initialization). This way, you'll have a better chance that users will discover this feature. Ideally, a nice example could be documented there, or a link to a working example added.

src/peft/tuners/tuners_utils.py
```diff
@@ -665,3 +688,73 @@ def check_adapters_to_merge(module: BaseTunerLayer, adapter_names: Optional[list
...
+                    submodule.layer_idx = current_idx
+    layers = nn.ModuleList(new_layers)
+    if model_type == "llama":
+        model.layers = layers
+    elif model_type == "bert":
+        model.encoder.layer = layers
+    elif model_type == "falcon":
+        model.h = layers
+    else:
+        raise AssertionError("Unexpected model type.")
```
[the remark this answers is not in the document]
siddartha-RE: Missed that I had overridden merged adapters just to apply the check. Good catch. Fixed.

I added that as a helper for people potentially adding support for new models and forgetting to add a condition here. Changed to a value error and updated the message to be helpful to someone modifying the code.

src/peft/tuners/lora/config.py
```diff
@@ -108,6 +108,10 @@ class LoraConfig(PeftConfig):
             ranks. Right now, DoRA only supports non-quantized linear layers. DoRA introduces a bigger overhead than
             pure LoRA, so it is recommended to merge weights for inference. For more information, see
             https://arxiv.org/abs/2402.09353.
+        layer_replication(`List[Tuple[int, int]]):
```
BenjaminBossan, line 111:
```suggestion
        layer_replication(`List[Tuple[int, int]]`):
```

src/peft/tuners/lora/model.py
```diff
@@ -129,6 +135,19 @@ def _check_new_adapter_config(self, config: LoraConfig) -> None:
...
         return check_target_module_exists(lora_config, key)
 
+    def _prepare_model(self, peft_config: LoraConfig, model: nn.Module):
+        r"""
+        A private method to modify the model structure before adapter is applied.
+
+        Args:
+            peft_config (`PeftConfig`):
+                The prepared adapter config.
+            model_config (`nn.Module`):
```
BenjaminBossan, line 145:
```suggestion
            model (`nn.Module`):
```

BenjaminBossan (changes requested):
Thanks a lot for the recent changes. The documentation is much clearer now and users should be able to understand what this does and find further information.

I found a couple of typos etc., otherwise this LGTM.

One question: Would we be able to check in the unit test that the weights are shared  instead of creating a new copy?

docs/source/developer_guides/lora.md
```diff
@@ -87,6 +87,21 @@ The default LoRA settings in PEFT add trainable weights to the query and value l
...
+For example increasing a 7B model to a 10B model as described in the [SOLAR](https://arxiv.org/abs/2312.15166). PEFT LoRA supports this kind of merge in a memory efficient manner that supports further fine-tuning
+using LoRA adapters attached to the layers post replication of the layers. The replicated layers do not take additional memory as they share the underlying weights so the only additional memory required is the
+memory for the adapter weights. To use this feature you would create a config with the `layer_replication` argument.
+```py
+config = LoraConfig(layer_replication=[[0,4], [2,5]], ...)
+```
+Given the original model had 5 layers `[0, 1, 2 ,3, 4]`, this would create a model with 7 layers arranged as `[0, 1, 2, 3, 2, 3, 4]`. This follows the mergekit pass through merge convention where sequences
+of layers specified as start inclusive and end exclusive tuples are stacked to build the final model. It is important to note that each layer in the final model gets its own distinct set of LoRA adpaters.
```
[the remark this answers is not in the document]
siddartha-RE: People with my writing style must have generated training data!

siddartha-RE:
@BenjaminBossan addressed comments and updated the unit test to check that some weights are shared and others are distinct.

src/peft/tuners/tuners_utils.py
```diff
@@ -170,13 +171,27 @@ def _prepare_adapter_config(self, peft_config: PeftConfig, model_config: dict) -
...
+    def _prepare_model(self, peft_config: PeftConfig, model: nn.Module):
+        r"""
+        A private method to modify the model structure before adapter is applied.
+
+        See `peft.tuner.lora.LoraModel._prepare_model` for an example.
+
+        Args:
+            peft_config (`PeftConfig`):
+                The prepared adapter config.
+            model_config (`nn.Module`):
```
BenjaminBossan, line 190:
```suggestion
            model (`nn.Module`):
```

BenjaminBossan (approved):
Thanks a lot, this LGTM.

I found a minor error in a docstring, no big deal.

Let's have a final review by @younesbelkada or @pacman100 before merging.

siddartha-RE:
@BenjaminBossan do you have a sense for when this PR is likely to be merged?

pacman100 (approved):
Wow! Thank you @siddartha-RE for all the work on adding support for memory efficient way of replicating layers and inserting LoRAs on those for efficient fine-tuning! 🔥🚀✨

BenjaminBossan:
I think there is no technical reason why this shouldn't work, but with the given way of specifying the layers, I'm not sure how to indicate if layers in the encoder or decoder part should be replicated. Maybe Siddartha has an idea.

siddartha-RE:
> I think there is no technical reason why this shouldn't work, but with the given way of specifying the layers, I'm not sure how to indicate if layers in the encoder or decoder part should be replicated. Maybe Siddartha has an idea.

I think the easiest option here may be to promote the replication config to `Union[Dict[str, list], list]` if it is a dict then there will be separate entries for `encoder` and `decoder` and architectures that support can do the right thing.

BenjaminBossan:
> I think the easiest option here may be to promote the replication config to `Union[Dict[str, list], list]` if it is a dict then there will be separate entries for `encoder` and `decoder` and architectures that support can do the right thing.

Sounds good, are you interested in adding this? I checked mergekit and AFAICT there is no specification for how to define this for encoder-decoder models. It would be unfortunate if they come up with a different standard, but I guess we could add that later.

xzuyn:
Could this possibly be updated to set the replicated layer's `o_proj` and `down_proj` to zero so that the performance remains the same as the base before training?

I think this idea initially comes from [LLaMa-Pro](https://github.com/TencentARC/LLaMA-Pro/blob/main/scripts/block_expansion.py#L32), and was later tested on [chargoddard's `mistral-11b-slimorca`](https://huggingface.co/chargoddard/mistral-11b-slimorca).

Quoting Charles here:
> The base model for this came from a variation on Undi's [Mistral 11B recipe](https://huggingface.co/Undi95/Mistral-11B-v0.1). The `o_proj` and `down_proj` tensors were set to zero in the added layers, making the output exactly identical to Mistral 7B before training.
>
> ~~Benchmarks look good locally but still evaluating actual usefulness.~~ Update: this turned out great! 10/10 would recommend as a training approach.

BenjaminBossan:
> Could this possibly be updated to optionally set the replicated layer's `o_proj` and `down_proj` to zero so that the performance remains the same as the base before training?

Also described in [this answer.ai blogpost](https://www.answer.ai/posts/2024-04-26-fsdp-qdora-llama3.html).

KimJaehee0725:
@siddartha-RE Hello. First of all, thanks for your brilliant idea. I want to deep dive into the mechanism. For that, could you give me the specific configurations of experiments of the models you trained, https://huggingface.co/abacusai/Fewshot-Metamath-OrcaVicuna-Mistral-10B. 

It will be really helpful for my work. 

Thanks!

PR #1371, opened 2024-01-18
[docs] Lora-like guides

stevhliu:
As discussed in #1332, this PR condenses the LoRA guides by showcasing one specific task type (image classification) and redirecting to other use cases/tasks stored on the PEFT Hub org. Instead, this guide expands on the other LoRA methods such as LoHa, LoKr, and AdaLoRA. The configs of these methods aren't optimized for the best performance, so any feedback you have on that would be super useful!


docs/source/task_guides/lora_based_methods.md
```diff
@@ -0,0 +1,343 @@
...
+)
+model = get_peft_model(model, config)
+model.print_trainable_parameters()
+"trainable params: 667,493 || all params: 86,543,818 || trainable%: 0.7712775047664294"
+```
+
+</hfoption>
+<hfoption id="LoHa">
+
+[LoHa](../conceptual_guides/adapter#low-rank-hadamard-product-loha) decomposes the weight update matrix into *four* smaller matrices and each pair of smaller matrices are combined with the Hadamard product. This allows the weight update matrix to keep the same number of trainable parameters but with a higher rank. The size of the smaller matrices are determined by its *rank* or `r`. You'll also want to specify the `target_modules` which determines where the smaller matrices are inserted. For this guide, you'll target the *query* and *value* matrices of the attention blocks. Other important parameters to set are `alpha` (scaling factor), and `modules_to_save` (the modules apart from the LoHa layers to be trained and saved). All of these parameters - and more - are found in the [`LoHaConfig`].
```
pacman100, line 181:
```suggestion
[LoHa](../conceptual_guides/adapter#low-rank-hadamard-product-loha) decomposes the weight update matrix into *four* smaller matrices and each pair of smaller matrices is combined with the Hadamard product. This allows the weight update matrix to keep the same number of trainable parameters when compared to LoRA but with a higher rank (r^2 for LoHA when compared to 2*r for LoRA). The size of the smaller matrices is determined by its *rank* or `r`. You'll also want to specify the `target_modules` which determines where the smaller matrices are inserted. For this guide, you'll target the *query* and *value* matrices of the attention blocks. Other important parameters to set are `alpha` (scaling factor), and `modules_to_save` (the modules apart from the LoHa layers to be trained and saved). All of these parameters - and more - are found in the [`LoHaConfig`].
```

docs/source/task_guides/lora_based_methods.md
```diff
@@ -0,0 +1,343 @@
...
+)
+model = get_peft_model(model, config)
+model.print_trainable_parameters()
+"trainable params: 116,069 || all params: 87,172,042 || trainable%: 0.13314934162033282"
+```
+
+</hfoption>
+<hfoption id="AdaLoRA">
+
+[AdaLoRA](../conceptual_guides/adapter#adaptive-low-rank-adaptation-adalora) efficiently manages the LoRA parameter budget by assigning important weight matrices more parameters and pruning less important ones. In contrast, LoRA evenly distributes parameters across all modules. You can control the average desired *rank* or `r` of the matrices, and which modules to apply AdaLoRA to with `target_modules`. Other important parameters to set are `lora_alpha` (scaling factor), and `modules_to_save` (the modules apart from the AdaLoRA layers to be trained and saved). All of these parameters - and more - are found in the [`AdaLoraConfig`].
```
pacman100, line 221: The important config params in AdaLoRA are init_r, tinit, tfinal and deltaT. Here, rank < init_r, i.e., the final average rank should be less than that of the initial rank. This config is incorrect in this regard and needs to be updated accordingly.

pacman100 (commented):
Thank you @stevhliu for consolidating all the LoRA like guides into a single guide while focusing on usage of different LoRA variants. Thank you for the HF Space collection of notebook examples.

Left a few suggestions and comments.

pacman100 (approved):
Thank you! 🤗

younesbelkada (approved):
Thanks @stevhliu  !

docs/source/task_guides/lora_based_methods.md
```diff
@@ -0,0 +1,345 @@
...
```
BenjaminBossan, line 221: There is a bit of an issue with the AdaLora example. For AdaLora, we need to call the method `update_and_allocate` during each training step, otherwise, the adaptation step is not performed. This requires to either write a custom training loop or to subclass the `Trainer` to add this step. An example of a custom training loop that calls this method is here:

https://github.com/huggingface/peft/blob/912ad41e96e03652cabf47522cd876076f7a0c4f/examples/conditional_generation/peft_adalora_seq2seq.py#L120

Given this guide, as is, the reader may not be aware that this is required and thus use AdaLora incorrectly. These solutions come to mind:

1. Don't provide the AdaLora example. Instead, refer to the existing examples with custom training loops.
2. Don't use `Trainer`, instead use a custom training loop that calls this method when AdaLora is being used.
3. Subclass `Trainer` to perform this step if AdaLora is being used. I don't know enough about `Trainer` to be really sure how to do that correctly.

BenjaminBossan (changes requested):
Thanks for writing this excellent guide and consolidating the different methods. Overall, this looks really good, I just have a concern regarding the use of AdaLora. Please check the comment and the solutions I suggested.

docs/source/task_guides/lora_based_methods.md
stevhliu: Great point! I mention `update_and_allocate` + link to a custom training loop in a `Warning` in the training section for more visibility. I think we should still keep the example of showing the `AdaLoraConfig` though so users still know how to set it up.

BenjaminBossan (approved):
Happy with the added warning, thanks for adding it.

PR #1380, opened 2024-01-22
add mixtral in LoRA mapping

younesbelkada:
cc @pacman100 @abhishekkrthakur


pacman100 (approved):
Thank you @younesbelkada!

PR #1393, opened 2024-01-24
Add positional args to PeftModelForCausalLM.generate

SumanthRH:
# What does this PR do?
A small PR for adding positional arguments to the generate method for `PeftModelForCausalLM`. The motivation is that `PeftModel` classes like `PeftModelForCausalLM` should feel more like a wrapper for `AutoModel` and shouldn't break existing code which uses positional args. Given that the messy internals of input processing are handled in the `prepare_inputs_for_generation` method, I think this is a safe feature. Also, it looks like `PeftModelForSeq2SeqLM` needs some refactoring for the prompt learning methods (to shift logic from `.generate` to `prepare_inputs_for_generation`), so I'm leaving that refactoring + positional args support to another PR.

### DIFF
--- a/tests/test_tuners_utils.py
+++ b/tests/test_tuners_utils.py
@@ -24,11 +24,11 @@ from transformers import AutoModel, AutoModelForCausalLM, AutoModelForSeq2SeqLM
 
 from peft import IA3Config, LoHaConfig, LoraConfig, get_peft_model
 from peft.tuners.tuners_utils import (
-    INCLUDE_LINEAR_LAYERS_SHORTHAND,
     _maybe_include_all_linear_layers,
     check_target_module_exists,
     inspect_matched_modules,
 )
+from peft.utils import INCLUDE_LINEAR_LAYERS_SHORTHAND
 
 from .testing_utils import require_bitsandbytes, require_torch_gpu
 
--- a/tests/testing_common.py
+++ b/tests/testing_common.py
@@ -650,8 +650,22 @@ class PeftCommonTester:
         # check if `generate` works
         _ = model.generate(**inputs)
 
-        with self.assertRaises(TypeError):
-            # check if `generate` raises an error if no positional arguments are passed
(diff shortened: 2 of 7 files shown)


tests/test_tuners_utils.py
```diff
@@ -24,11 +24,11 @@
 
 from peft import IA3Config, LoHaConfig, LoraConfig, get_peft_model
 from peft.tuners.tuners_utils import (
-    INCLUDE_LINEAR_LAYERS_SHORTHAND,
     _maybe_include_all_linear_layers,
     check_target_module_exists,
     inspect_matched_modules,
 )
+from peft.utils import INCLUDE_LINEAR_LAYERS_SHORTHAND
```
SumanthRH, line 31: This is a minor edit to import the constant from the right place

tests/testing_common.py
```diff
@@ -672,10 +686,6 @@ def _test_generate_half_prec(self, model_id, config_cls, config_kwargs):
         # check if `generate` works
         _ = model.generate(input_ids=input_ids, attention_mask=attention_mask)
 
-        with self.assertRaises(TypeError):
-            # check if `generate` raises an error if no positional arguments are passed
-            _ = model.generate(input_ids, attention_mask=attention_mask)
```
SumanthRH, line 677: This is a redundant check on positional args (behaviour is captured in `test_generate` already) so I've removed it

pacman100 (approved):
Thank you @SumanthRH for adding support for positional args to `generate` method of CausalLM. LGTM! 🤗

>  Given that the messy internals of input processing are handled in the prepare_inputs_for_generation method, I think this is a safe feature.

Yes, while reviewing the PR I was thinking the same, makes sense.

younesbelkada (approved):
This is very nice and fixes un-consistency between `PreTrainedModel` API and `PeftModel` classes ! Thanks a lot @SumanthRH !

PR #1409, opened 2024-01-29
Update custom_models.md

boyufan:
Hi, in the example of timm models, there is a misuse of MLP( ), which should be model.


younesbelkada (approved):
Makes sense, thanks for the fix!

pacman100 (approved):
Thank you @boyufan!

PR #1413, opened 2024-01-30
Fix PEFT breaking change

younesbelkada:
cc @pacman100 @sayakpaul


sayakpaul (approved):
Thanks for the quick resolve. Would it make sense to do a patch release for this?

pacman100 (commented):
Hello Younes, thank you for the fix but is there a failure case because i think `vocab_size` would only be available for transformers model.

pacman100 (approved):
Just saw the issue offline, Thank you for the quick fix! 🚀

src/peft/utils/save_and_load.py
```diff
@@ -136,8 +136,18 @@ def get_peft_model_state_dict(
     elif save_embedding_layers == "auto":
         vocab_size = getattr(getattr(model, "config", None), "vocab_size", None)
         model_id = getattr(config, "base_model_name_or_path", None)
+
+        # For some models e.g. diffusers the text config file is stored in a subfolder
+        # we need to make sure we can donwload that config.
+        has_remote_config = model_id is not None and file_exists(model_id, "config.jon")
```
pacman100, line 142:
```suggestion
        has_remote_config = model_id is not None and file_exists(model_id, "config.json")
```

pacman100 (approved):
Thank you Younes for the fix! ✨ left a suggestion

src/peft/utils/save_and_load.py
```diff
@@ -136,8 +136,18 @@ def get_peft_model_state_dict(
     elif save_embedding_layers == "auto":
         vocab_size = getattr(getattr(model, "config", None), "vocab_size", None)
         model_id = getattr(config, "base_model_name_or_path", None)
+
+        # For some models e.g. diffusers the text config file is stored in a subfolder
+        # we need to make sure we can donwload that config.
```
pacman100, line 141:
```suggestion
        # we need to make sure we can download that config.
```

PR #1420, opened 2024-01-31
DO NOT MERFGE: Attempt to fix memory 

younesbelkada:
Attempt to fix : https://github.com/huggingface/optimum-habana/issues/590

cc @regisss 
Do not merge for now

### DIFF
--- a/src/peft/tuners/lora/bnb.py
+++ b/src/peft/tuners/lora/bnb.py
@@ -153,11 +153,14 @@ if is_bnb_available():
             if self.disable_adapters:
                 if self.merged:
                     self.unmerge()
-                result = self.base_layer(x, *args, **kwargs)
+                with torch.no_grad():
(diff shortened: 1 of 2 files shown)


younesbelkada:
cc @vivekgoe this might fix the bug you tagged me, can you try to have a look 🙏 ?

vivekgoe:
@younesbelkada this PR does reduce memory usage, but with this loss is not converging.  

Can the issue be due to linear layer used in the model being a derived class from nn.linear? 
https://github.com/huggingface/transformers/blob/09eb11a1bd6e77833fde7020efc57e6a804e4e8a/src/transformers/models/falcon/modeling_falcon.py#L75

If I replace all instances of FalconLinear in the model with nn.Linear then memory footprint goes back to numbers before. Also this memory increase is not seen for other models using PEFT.

PR #1421, opened 2024-01-31
Renovate quality tools

akx:
This PR:

* removes old `isort` configuration that hasn't been in use since 99901896ccea794fd012b81fbf747c3476d98388
* replaces `black` formatting with `ruff format` (though `black` is still a hidden dependency via `hf-doc-builder`
* broadens quality controls to everywhere in the repo (since they're instant to run now)
* conservatively enables the Ruff `UP` (upgrade) rules and auto-fixes what can be auto-fixed.

This is similar to e.g. https://github.com/huggingface/transformers/pull/27144 and https://github.com/huggingface/diffusers/pull/5841.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,8 @@ target-version = ['py38']
 [tool.ruff]
 target-version = "py38"
 line-length = 119
+
+[tool.ruff.lint]
 extend-select = [
     "C", # Complexity
     "E", # PEP8 errors
@@ -22,7 +24,7 @@ ignore = [
     "W605", # Invalid escape sequence
 ]
 
-[tool.ruff.isort]
+[tool.ruff.lint.isort]
 lines-after-imports = 2
 known-first-party = ["peft"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,6 @@ extend-select = [
 ignore = [
     "C901", # Function too complex
     "E501", # Line length (handled by ruff-format)
-    "E741", # Ambiguous variable name
     "UP007", # X | Y style Unions
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,15 +11,14 @@ extend-select = [
     "E", # PEP8 errors
     "F", # PEP8 formatting
     "I", # Import sorting
-    "UP006", # use dict[...] and list[...] instead typing.*
-    "UP009", # UTF-8 declaration not needed in py3
-    "UP032", # f-string instead of format call
+    "UP", # Pyupgrade upgrades
     "W", # PEP8 warnings
 ]
 ignore = [
     "C901", # Function too complex
     "E501", # Line length (handled by ruff-format)
     "E741", # Ambiguous variable name
+    "UP007", # X | Y style Unions
     "W605", # Invalid escape sequence
 ]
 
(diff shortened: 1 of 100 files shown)


BenjaminBossan (changes requested):
Thanks a lot for this update, which looks indeed very useful.

Doing a full review is a bit difficult because the diff is so large. Could we maybe add some exceptions to the ruff checks for the time being? Specifically, I'm thinking of:

1. https://docs.astral.sh/ruff/rules/redundant-open-modes/
2. https://docs.astral.sh/ruff/rules/utf8-encoding-declaration/
3. https://docs.astral.sh/ruff/rules/f-string/

This should take care of the majority of changes and thus lead to a much smaller diff and make the review easier. Then I can discuss with my co-maintainers if we want to have these exceptions or not, and if we don't, remove them in a later PR.

akx:
@BenjaminBossan Sure! I split the single `UP` autofix commit into multiple, that can be reviewed in GitHub one by one ✨

src/peft/auto.py
```diff
@@ -50,7 +49,7 @@ class _BaseAutoPeftModel:
 
     def __init__(self, *args, **kwargs):
         # For consistency with transformers: https://github.com/huggingface/transformers/blob/91d7df58b6537d385e90578dac40204cb550f706/src/transformers/models/auto/auto_factory.py#L400
-        raise EnvironmentError(
+        raise OSError(
```
BenjaminBossan, line 52: As the comment above suggests, let's stick with `EnvironmentError`, even if it's not the correct error type here.

BenjaminBossan (commented):
Thanks for splitting up the commits. I'll have a discussion with the co-maintainers of the specific rules added here and we'll come back with a conclusion.

src/peft/auto.py
akx: I believe the "consistency" there refers to raising an error in `__init__`, not the particular error type, but sure, added a commit to `noqa` that instead.

pyproject.toml
```diff
@@ -1,33 +1,33 @@
...
+    "I", # Import sorting
+    "UP", # Pyupgrade upgrades
+    "W", # PEP8 warnings
+]
+ignore = [
+    "C901", # Function too complex
+    "E501", # Line length (handled by ruff-format)
+    "E741", # Ambiguous variable name
+    "UP007", # X | Y style Unions
+    "W605", # Invalid escape sequence
```
BenjaminBossan, line 24: What is the reasoning for ignoring this?

src/peft/auto.py
BenjaminBossan: No, this refers to the type of error. I know because I flagged the error type initially :)

BenjaminBossan (commented):
Thanks a lot, the individual changes look sound and after discussing with the co-maintainers, we agree with the rules and exceptions chosen here.

Could you please check the two scripts that have yet to be reformatted according to CI? Thanks.

pyproject.toml
akx: It was ignored in the preimage. I'll check if any of the original ignores can be safely removed.

akx: Okay, E741 and W605 were easy enough to fix, so did that.

akx:
@BenjaminBossan Done, rebased, removed a couple of ignores by either autofixing or noqa'ing (there are 29 functions that are too complex according to default C901, so not touching those).

BenjaminBossan (approved):
Thanks a lot for moving the code quality checks over completely to ruff. LGTM.

PR #1422, opened 2024-01-31
[docs] Polytropon API

stevhliu:
Adds Polytropon to the API docs.


younesbelkada (approved):
Very nice work ! thanks a lot @stevhliu !

PR #1364, opened 2024-01-16
Add new merging methods

pacman100:
### What does this PR do?
1. Add new model merging methods for LoRA based on the papers [TIES-MERGING: Resolving Interference When Merging Models](https://arxiv.org/abs/2306.01708) and [Language Models are Super Mario: Absorbing Abilities from Homologous Models as a Free Lunch](https://arxiv.org/abs/2311.03099). The methods are `ties`, `dare_linear`, `dare_ties`, `ties_svd`, `dare_linear_svd`, `dare_ties_svd`.
3. The inspiration for the implementation of these methods are from https://github.com/yule-BUAA/MergeLM/tree/main and https://github.com/cg123/mergekit/tree/main.

Example of `ties_svd` is shown below (https://github.com/pacman100/peft-dreambooth-ui/blob/main/lora_merging.ipynb):
Screenshot 2024-01-17 at 10 45 07 AM

LLM LoRA merging example:
Screenshot 2024-01-17 at 6 12 49 PM

To do:
- [x] Add tests
- [x] Add documentation
- [x] Add example notebook

PR #1423, opened 2024-01-31
[docs] Model merging

stevhliu:
A guide to new model merging methods introduced in #1364.

todo:

- [x] add API reference for merging utilities (once the other PR is merged, I'll rerun the `build_pr_documentation` test and it should pass) 
- [ ] test and run code examples

### DIFF
--- a/docs/source/developer_guides/model_merging.md
+++ b/docs/source/developer_guides/model_merging.md
@@ -18,16 +18,16 @@ rendered properly in your Markdown viewer.
 
 Training a model for each task can be costly, take up storage space, and the models aren't able to learn new information to improve their performance. Multitask learning can overcome some of these limitations by training a model to learn several tasks, but it is expensive to train and designing a dataset for it is challenging. *Model merging* offers a solution to these challenges by combining multiple pretrained models into one model, giving it the combined abilities of each individual model without any additional training.
 
-PEFT provides two methods for model merging:
+PEFT provides several methods for merging models like a linear or SVD combination. This guide focuses on two methods that are more efficient for merging LoRA adapters by eliminating redundant parameters:
 
 * [TIES](https://hf.co/papers/2306.01708) - TrIm, Elect, and Merge (TIES) is a three-step method for merging models. First, redundant parameters are trimmed, then conflicting signs are resolved into an aggregated vector, and finally the parameters whose signs are the same as the aggregate sign are averaged. This method takes into account that some values (redundant and sign disagreement) can degrade performance in the merged model.
 * [DARE](https://hf.co/papers/2311.03099) - Drop And REscale is a method that can be used to prepare for other model merging methods like TIES. It works by randomly dropping parameters according to a drop rate and rescaling the remaining parameters. This helps to reduce the number of redundant and potentially interfering parameters among multiple models.
 
-Models are merged with the [`~LoraModel.add_weighted_adapter`] method, and the specific model merging method is specified in the `combination_type` parameter. This guide will show you how to merge models with TIES and DARE.
+Models are merged with the [`~LoraModel.add_weighted_adapter`] method, and the specific model merging method is specified in the `combination_type` parameter.
 
 ## Merge method
 
-With PEFT, merging is enabled by setting `combination_type` and `ties_density` to a value of the weights to keep from the individual models. For example, let's merge three finetuned [TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T](https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T) models: [tinyllama_lora_nobots](https://huggingface.co/smangrul/tinyllama_lora_norobots), [tinyllama_lora_sql](https://huggingface.co/smangrul/tinyllama_lora_sql), and [tinyllama_lora_adcopy](https://huggingface.co/smangrul/tinyllama_lora_adcopy).
+With TIES and DARE, merging is enabled by setting `combination_type` and `density` to a value of the weights to keep from the individual models. For example, let's merge three finetuned [TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T](https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T) models: [tinyllama_lora_nobots](https://huggingface.co/smangrul/tinyllama_lora_norobots), [tinyllama_lora_sql](https://huggingface.co/smangrul/tinyllama_lora_sql), and [tinyllama_lora_adcopy](https://huggingface.co/smangrul/tinyllama_lora_adcopy).
 
 <Tip warninig={true}>
 
@@ -35,7 +35,7 @@ When you're attempting to merge fully trained models with TIES, you should be aw
 
 <br>
 
-This shouldn't be an issue if you're only merging PEFT modules trained from a shared base model.
+This shouldn't be an issue if you're only merging LoRA adapters trained from the same base model.
 
 </Tip>
 
@@ -47,7 +47,7 @@ from transformers import AutoModelForCausalLM, AutoTokenizer
 import torch
 
 config = PeftConfig.from_pretrained("smangrul/tinyllama_lora_norobots")
-model = AutoModelForCausalLM.from_pretrained(config.base_model_name_or_path, load_in_4bit=True, device_map="auto")
+model = AutoModelForCausalLM.from_pretrained(config.base_model_name_or_path, load_in_4bit=True, device_map="auto").eval()
 tokenizer = AutoTokenizer.from_pretrained("smangrul/tinyllama_lora_norobots")
 
 model = PeftModel.from_pretrained(model, "smangrul/tinyllama_lora_norobots", adapter_name="norobots")
@@ -55,7 +55,7 @@ _ = model.load_adapter("smangrul/tinyllama_lora_sql", adapter_name="sql")
 _ = model.load_adapter("smangrul/tinyllama_lora_adcopy", adapter_name="adcopy")
 ```
 
-Set the adapters, weights, `adapter_name`, `combination_type`, and `ties_density` with the [`~LoraModel.add_weighted_adapter`] method.
+Set the adapters, weights, `adapter_name`, `combination_type`, and `density` with the [`~LoraModel.add_weighted_adapter`] method.
 
 <hfoptions id="merge-method">
 <hfoption id="TIES">
@@ -87,7 +87,6 @@ model.add_weighted_adapter(adapters, weights, adapter_name, combination_type="da
 Set the newly merged model as the active model with the [`~LoraModel.set_adapter`] method.
 
 ```py
-model.eval()
 model.set_adapter("merge")
 ```
 
--- a/docs/source/developer_guides/model_merging.md
+++ b/docs/source/developer_guides/model_merging.md
@@ -16,37 +16,53 @@ rendered properly in your Markdown viewer.
 
 # Model merging
 
-Training a model for each task can be costly and take up storage, and these models aren't able to learn new information to improve performance. Multitask learning can train a model to learn multiple tasks, but this is costly to train and designing a dataset for it can be challenging. *Model merging* offers a solution to these challenges by combining multiple pretrained models into one model, giving it the combined abilities of each individual model, without any additional training.
+Training a model for each task can be costly, take up storage space, and the models aren't able to learn new information to improve their performance. Multitask learning can overcome some of these limitations by training a model to learn several tasks, but this is expensive to train and designing a dataset for it can be challenging. *Model merging* offers a solution to these challenges by combining multiple pretrained models into one model, giving it the combined abilities of each individual model, without any additional training.
 
 PEFT provides two methods for model merging:
 
-* [TIES-Merging](https://hf.co/papers/2306.01708) - TrIm, Elect, and Merge (TIES) is a three-step method for merging models. First, redundant parameters are trimmed, then conflicting signs are resolved into an aggregated vector, and finally the parameters whose signs are the same as the aggregate sign are averaged. This method takes into account that some values (redundant and sign disagreement) can reduce model performance in the merged model.
-* [DARE](https://hf.co/papers/2311.03099) - Drop And REscale is a method that can be used to prepare model merging methods like TIES-Merging. It works by randomly dropping parameters according to a drop rate, and rescaling the remaining parameters. This helps to reduce the number of redundant and potentially interfering parameters among multiple models.
+* [TIES-Merging](https://hf.co/papers/2306.01708) - TrIm, Elect, and Merge (TIES) is a three-step method for merging models. First, redundant parameters are trimmed, then conflicting signs are resolved into an aggregated vector, and finally the parameters whose signs are the same as the aggregate sign are averaged. This method takes into account that some values (redundant and sign disagreement) can degrade performance in the merged model.
+* [DARE](https://hf.co/papers/2311.03099) - Drop And REscale is a method that can be used to prepare for model merging methods like TIES. It works by randomly dropping parameters according to a drop rate and rescaling the remaining parameters. This helps to reduce the number of redundant and potentially interfering parameters among multiple models.
 
 Models are merged with the [`~LoraModel.add_weighted_adapter`] method, and the specific model merging method is specified in the `combination_type` parameter. This guide will show you how to merge models with TIES, DARE, and a combination of both TIES and DARE.
 
 ## TIES
 
-The [`~utils.ties`] method uses a [`~utils.magnitude_based_pruning`] approach to trim redundant parameters such that only the top-k percent of values are kept from each task vector. The number of values to keep are specified by the `density` parameter. The task tensors are weighted, and the [`~utils.calculate_majority_sign_mask`] *elects* the sign vector which means calculating the total magnitude for each parameter across all models. Lastly, the [`~utils.disjoint_merge`] function calculates the mean of the parameter values whose sign is the same as the *elected sign vector*.
+The [`~utils.ties`] method uses a [`~utils.magnitude_based_pruning`] approach to trim redundant parameters such that only the top-k percent of values are kept from each task vector. The number of values to keep are specified by the `density` parameter. The task tensors are weighted, and the [`~utils.calculate_majority_sign_mask`] *elects* the sign vector. This means calculating the total magnitude for each parameter across all models. Lastly, the [`~utils.disjoint_merge`] function calculates the mean of the parameter values whose sign is the same as the *elected sign vector*.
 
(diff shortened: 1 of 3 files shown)


docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,178 @@
...
+vae = AutoencoderKL.from_pretrained(
+        "madebyollin/sdxl-vae-fp16-fix",
+)
+pipeline = StableDiffusionXLPipeline.from_pretrained(
+    "stabilityai/stable-diffusion-xl-base-1.0",
+    vae=vae,
+    torch_dtype=torch.float16,
+).to("cuda")
+
+pipeline.load_lora_weights(
```
stevhliu, line 143: @pacman100 is this possible or do we need to import the `load_combine_checkpoint` method from `train_dreambooth_peft` as shown in this [notebook](https://github.com/pacman100/peft-dreambooth-ui/blob/main/lora_merging.ipynb)?

pacman100: Hello, for Diffusers, the merging methods can't be used as they follow a ckpt format that is different from the one I use and as Diffusers use low-level API of PEFT, it doesn't have access to the `add_weighted_adapter` method. The repo that I have uses bespoke ckpt format and uses normal PEFT API to get access to the `add_weighted_adapter` method.

docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,178 @@
...
+
+PEFT provides two methods for model merging:
+
+* [TIES-Merging](https://hf.co/papers/2306.01708) - TrIm, Elect, and Merge (TIES) is a three-step method for merging models. First, redundant parameters are trimmed, then conflicting signs are resolved into an aggregated vector, and finally the parameters whose signs are the same as the aggregate sign are averaged. This method takes into account that some values (redundant and sign disagreement) can degrade performance in the merged model.
+* [DARE](https://hf.co/papers/2311.03099) - Drop And REscale is a method that can be used to prepare for model merging methods like TIES. It works by randomly dropping parameters according to a drop rate and rescaling the remaining parameters. This helps to reduce the number of redundant and potentially interfering parameters among multiple models.
+
+Models are merged with the [`~LoraModel.add_weighted_adapter`] method, and the specific model merging method is specified in the `combination_type` parameter. This guide will show you how to merge models with TIES, DARE, and a combination of both TIES and DARE.
+
+## TIES
+
```
prateeky2806, line 29: It might be worth adding a note of caution for TIES which might not be an issue here but for the full model, it can be an issue. I discussed this in my previous comment on the pull request which adds these merging features. 

"As far as I remember, there were no special tricks when using TIES. Typically, it is important to take care of the shared parameters when merging full models. Sometimes the embedding layers of different finetunes have a few additional tokens added and they can be different across different finetunes. Say a math model adds "=" and a coding model adds "{" as additional tokens that will occupy the V+1 index in vocabulary but so in such cases the embedding layers should be merged carefully because different token embedding can be merged together. I feel like this is something people typically miss but should not be an issue with PEFT methods. For merging Lora I have tried merging BA and then decomposing it back which seems to work fine. However, I have not tried merging A and B matrices separately but I feel that it should not be much worse. How big is the performance gap between these two ways of merging Lora's in your experience?"

docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,178 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# Model merging
```
prateeky2806, line 17: One more comment, I am not sure if these readme guides/documentation typically provide citations to the original papers or not but that would be good to have the bottom of the page.

docs/source/developer_guides/model_merging.md
stevhliu: Ah I see! Then maybe we shouldn't show an example of TIES/DARE model merging with Diffusers yet then since it's not supported and `peft_dreambooth_ui` is still in development. WDYT @BenjaminBossan?

docs/source/developer_guides/model_merging.md
stevhliu: Awesome, added a note about this! 👍

docs/source/developer_guides/model_merging.md
BenjaminBossan: Yes, I agree, it would be confusing otherwise.

prateeky2806:
Hi, is there an estimated timeline about by when this would be merged ?

docs/source/developer_guides/model_merging.md
stevhliu: Generally, we provide a link to the original paper that implements the method as a reference/citation for users who are more interested. I don’t think it’s as common outside of academia to have Citation sections in documentation. For user-facing guides like these, users are typically only interested in quickly copying/pasting the code to get it to run 🙂 

We'll add a link to the blog post @sayakpaul is working on once it's ready where it'll be more appropriate to acknowledge!

stevhliu:
> Hi, is there an estimated timeline about by when this would be merged ?

Hi, it should be ready in the next few days and at the end of the week by the latest if there are no major issues!

docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Merge method
+
+With PEFT, merging is enabled by setting `combination_type` and `ties_density` to a value of the weights to keep from the individual models. For example, let's merge three finetuned [TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T](https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T) models: [tinyllama_lora_nobots](https://huggingface.co/smangrul/tinyllama_lora_norobots), [tinyllama_lora_sql](https://huggingface.co/smangrul/tinyllama_lora_sql), and [tinyllama_lora_adcopy](https://huggingface.co/smangrul/tinyllama_lora_adcopy).
+
+<Tip>
+
+If you're merging with TIES, use the [`~transformers.PreTrainedModel.resize_token_embeddings`] method to account for special tokens added to the embedding layer for each finetuned model. This method ensures the special tokens and initialization of the embedding layers are consistent.
+
+</Tip>
```
prateeky2806, line 36: I am not sure if I understand this correctly. Just to clarify what I meant in my earlier comment, if we are merging say three full Models like, WizardMath, WizardCode which are based on the common LLaMA checkpoints which has vocab of say 32k. Say when finetuning on LLaMA model on code data to get WizardCode model some new special tokens might be added which are not in the LLaMA tokenizer, say "{", "}" are the two special tokens added. Similarly for WizardMath we might have added say "=" as a special token.  Now in the embedding layer of WizardCode the 32001 embedding correspond to "{" and 32002 corresponds to "}". However, in WizardMath the 32001 embedding corresponds to "=". Hence, we need to merge the embedding layer such that we do not merge the learned representation of "{" from WizardCode with "=" from WizardMath which both have index 32001. 

However, this issue can only occur when merging full model and should not be an issue when merging PEFT modules trained on a shared base model. I am not sure if this is clear from the tip that has been added. It is also possible that I am misunderstanding something.

docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,136 @@
...
+```py
+from peft import PeftConfig, PeftModel
+from transformers import AutoModelForCausalLM, AutoTokenizer
+import torch
+
+config = PeftConfig.from_pretrained("smangrul/tinyllama_lora_norobots")
+model = AutoModelForCausalLM.from_pretrained(config.base_model_name_or_path, load_in_4bit=True, device_map="auto")
+tokenizer = AutoTokenizer.from_pretrained("smangrul/tinyllama_lora_norobots")
+
+# model.resize_token_embeddings(len(tokenizer)) if using TIES method
```
prateeky2806, line 49: Given my last comment, I am not sure if this is needed or if it makes sense. Based on my understanding this should not be needed for PEFT modules merging but I might be wrong.

docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+</hfoption>
+<hfoption id="DARE">
+
+```py
+adapters = ["norobots", "adcopy", "sql"]
+weights = [2.0, 0.3, 0.7]
+adapter_name = "merge"
+density = 0.2
```
prateeky2806, line 75: I am not sure how this density parameter works in dare_ties, I am assuming it is used to keep 20% params and then rescale as in DARE. However, I am not sure if we do TIES on top of it then will it again only keep the 20% params of the pruned and rescaled checkpoint essentially leading to 0.2*0.2 *100 = 4% remaining parameter or if it will keep 20% of the parameters. This is not a comment on the documentation but this behaviour is not very clear.

docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,136 @@
...
+* [DARE](https://hf.co/papers/2311.03099) - Drop And REscale is a method that can be used to prepare for other model merging methods like TIES. It works by randomly dropping parameters according to a drop rate and rescaling the remaining parameters. This helps to reduce the number of redundant and potentially interfering parameters among multiple models.
+
+Models are merged with the [`~LoraModel.add_weighted_adapter`] method, and the specific model merging method is specified in the `combination_type` parameter. This guide will show you how to merge models with TIES and DARE.
+
+## Merge method
+
+With PEFT, merging is enabled by setting `combination_type` and `ties_density` to a value of the weights to keep from the individual models. For example, let's merge three finetuned [TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T](https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T) models: [tinyllama_lora_nobots](https://huggingface.co/smangrul/tinyllama_lora_norobots), [tinyllama_lora_sql](https://huggingface.co/smangrul/tinyllama_lora_sql), and [tinyllama_lora_adcopy](https://huggingface.co/smangrul/tinyllama_lora_adcopy).
```
prateeky2806, line 30: Moreover, I am not sure what is the right place to say this, given that ties do disjoint mean hence typically values >=1 for the weights works best because this preserves the correct scale. Moreover, all ones are a good place to start if it is hard to select the weight values. Adding a tip for this can encourage people to explore values >=1 which in my experience works best.

docs/source/developer_guides/model_merging.md
stevhliu: Thanks, I tried to clarify it a bit more based on your comment

docs/source/developer_guides/model_merging.md
stevhliu: Added a note in the TIES section to try using weight values >= 1.0 for better results.

prateeky2806 (approved):

docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,141 @@
...
+
+PEFT provides two methods for model merging:
+
+* [TIES](https://hf.co/papers/2306.01708) - TrIm, Elect, and Merge (TIES) is a three-step method for merging models. First, redundant parameters are trimmed, then conflicting signs are resolved into an aggregated vector, and finally the parameters whose signs are the same as the aggregate sign are averaged. This method takes into account that some values (redundant and sign disagreement) can degrade performance in the merged model.
+* [DARE](https://hf.co/papers/2311.03099) - Drop And REscale is a method that can be used to prepare for other model merging methods like TIES. It works by randomly dropping parameters according to a drop rate and rescaling the remaining parameters. This helps to reduce the number of redundant and potentially interfering parameters among multiple models.
+
+Models are merged with the [`~LoraModel.add_weighted_adapter`] method, and the specific model merging method is specified in the `combination_type` parameter. This guide will show you how to merge models with TIES and DARE.
```
sayakpaul, line 26: @pacman100 should we clarify that these merging methods, for now, only work with LoRA adapters?

docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Merge method
+
+With PEFT, merging is enabled by setting `combination_type` and `ties_density` to a value of the weights to keep from the individual models. For example, let's merge three finetuned [TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T](https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T) models: [tinyllama_lora_nobots](https://huggingface.co/smangrul/tinyllama_lora_norobots), [tinyllama_lora_sql](https://huggingface.co/smangrul/tinyllama_lora_sql), and [tinyllama_lora_adcopy](https://huggingface.co/smangrul/tinyllama_lora_adcopy).
```
sayakpaul, line 30: `ties_density` should be optional, no? Would one need it even when the `combination_type` is something like "linear"?

docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,141 @@
...
+
+<Tip warninig={true}>
+
+When you're attempting to merge fully trained models with TIES, you should be aware of any special tokens each model may have added to the embedding layer which are not a part of the original checkpoint's vocabulary. This may cause an issue because each model may have added a special token to the same embedding position. If this is the case, you should use the [`~transformers.PreTrainedModel.resize_token_embeddings`] method to avoid merging the special tokens at the same embedding index.
+
+<br>
+
+This shouldn't be an issue if you're only merging PEFT modules trained from a shared base model.
```
sayakpaul, line 38: What's a "shared" base model?

docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,141 @@
...
+from transformers import AutoModelForCausalLM, AutoTokenizer
+import torch
+
+config = PeftConfig.from_pretrained("smangrul/tinyllama_lora_norobots")
+model = AutoModelForCausalLM.from_pretrained(config.base_model_name_or_path, load_in_4bit=True, device_map="auto")
+tokenizer = AutoTokenizer.from_pretrained("smangrul/tinyllama_lora_norobots")
+
+model = PeftModel.from_pretrained(model, "smangrul/tinyllama_lora_norobots", adapter_name="norobots")
+_ = model.load_adapter("smangrul/tinyllama_lora_sql", adapter_name="sql")
+_ = model.load_adapter("smangrul/tinyllama_lora_adcopy", adapter_name="adcopy")
```
sayakpaul, line 55: @pacman100 @BenjaminBossan makes sense to have copies of these checkpoints under the PEFT testing org?

docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,141 @@
...
+model.add_weighted_adapter(adapters, weights, adapter_name, combination_type="dare_ties", density=density)
+```
+
+</hfoption>
+</hfoptions>
+
+Set the newly merged model as the active model with the [`~LoraModel.set_adapter`] method.
+
+```py
+model.eval()
```
sayakpaul, line 90: I think it's better to call `eval()` when loading the base model.

```python
model = AutoModelForCausalLM.from_pretrained(config.base_model_name_or_path, load_in_4bit=True, device_map="auto")
```

docs/source/package_reference/merge_utils.md
```diff
@@ -0,0 +1,33 @@
...
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# Model merge
+
+PEFT provides several internal utilities for [model merging](../developer_guides/model_merging) with the TIES and DARE methods.
```
sayakpaul, line 19: If these merging methods are only supported for LoRA, let's add a disclaimer at the top.

sayakpaul (approved):
Looking solid!

docs/source/developer_guides/model_merging.md
BenjaminBossan: Yes, let's try to put those artifacts on https://huggingface.co/peft-internal-testing.

docs/source/developer_guides/model_merging.md
pacman100: Hello, in `dare_ties`, first random pruning happens based on density followed by scaling. After this, the `majority_sign_mask` and `disjoint_merge` are performed similar to the ties method. So, pruning is taken from dare which is random and rescaled followed by majority sign and disjoint merge from ties.

docs/source/developer_guides/model_merging.md
pacman100: I'm having different experience here. In my experiments based on LoRAs for diffusers/LLms, I find that `ties` with `majority_sign_method="frequency"` instead of the default `majority_sign_method="total"` is working better. Also, the weights of the adapters <=1 based on the relative weightage that we want to give is working better. For `ties` with `majority_sign_method="total"`, weighting has to be done carefully.

docs/source/developer_guides/model_merging.md
BenjaminBossan: Yes, I agree that this should be mentioned prominently.

docs/source/developer_guides/model_merging.md
```diff
@@ -0,0 +1,141 @@
...
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# Model merging
+
+Training a model for each task can be costly, take up storage space, and the models aren't able to learn new information to improve their performance. Multitask learning can overcome some of these limitations by training a model to learn several tasks, but it is expensive to train and designing a dataset for it is challenging. *Model merging* offers a solution to these challenges by combining multiple pretrained models into one model, giving it the combined abilities of each individual model without any additional training.
+
+PEFT provides two methods for model merging:
```
BenjaminBossan, line 21: There are a few more options than these two. I don't think we necessarily need to describe all of them in detail here, focusing on these two should be good. But I would use a different wording to not create the impression that these are the only ones.

BenjaminBossan (approved):
Nicely done. On top of what has already been mentioned, I just have one comment about there actually being more than 2 methods. Otherwise, this LGTM.

docs/source/developer_guides/model_merging.md
stevhliu: Good catch, that'll be an important distinction to add :)

docs/source/developer_guides/model_merging.md
stevhliu: Clarified that it's mainly used for TIES/DARE

docs/source/developer_guides/model_merging.md
stevhliu: Clarified to say it shouldn't be an issue if you're merging LoRA adapters trained from the same base model. Is that clearer?

docs/source/developer_guides/model_merging.md
prateeky2806: I have not tested TIES on diffusion models, but I could believe that to be true. When we were writing the TIES merging paper, we had many other methods of selecting signs based on normalized movement, frequency, etc which led to a few points improvements over TIES on average across many tasks but in the end, we did not add them because the optimal params depends on the distribution of the task vectors. So to make it simple we went with total everywhere. Maybe then its better to just set the default to weights = 1 with 20% unpruned parameters and then let the user try out what values they like most.

PR #1427, opened 2024-02-01
[`core`] fix critical bug in diffusers

younesbelkada:
Fixes https://github.com/huggingface/diffusers/issues/6809

Indeeed `copy` does not exist in PyTorch 🤯 one needs to use `clone()` .. that code path was never tested so I added some CI tests for that so that it'll never happen in the future

cc @pacman100 @apolinario


pacman100 (approved):
Thank you Younes for the quick fixes and nice test! ⚡

PR #1431, opened 2024-02-04
Extend PeftModel.from_pretrained() to models with disk-offloaded modules

blbadger:
This PR extends PeftModel.from_pretrained() to be compatible with disk offloading.  The approach is to re-make the `offload_index` map to the original (offloaded) model's safetensors files and then save the LoRA parameters with renamed model module parameters to new safetensors folders, updating the `offload_index` accordingly. 

This is a complement to PRs #1190  and `transformers` #[27412](https://github.com/huggingface/transformers/pull/27412), and is designed to allow for the loading of PeftModels using only the memory equivalent of one safetensors file.

TODO: merging compatibility for offloading was not designed for use with safetensors in #1190, so the merging code will be modified to support safetensors-saved models.


younesbelkada (commented):
The approach sounds great, thanks very much @blbadger for putting efforts into this ! I would like first @SunMarc to have a look as he is familiar with that 🙏

src/peft/peft_model.py
```diff
@@ -721,16 +748,82 @@ def load_adapter(self, model_id: str, adapter_name: str, is_trainable: bool = Fa
...
             if isinstance(device_map, str):
                 device_map = infer_auto_device_map(
                     self, max_memory=max_memory, no_split_module_classes=no_split_module_classes
                 )
+
+            prefix = "base_model.model."
+            # rename offload index weight and model names
+            target_modules = self.peft_config["default"].target_modules
+            if offload_index:
+                keys = list(offload_index.keys())
```
BenjaminBossan, line 761: Since the `load_adapter` method is already quite long, I wonder if we could move the offloading code inside a sub-method. WDYT?

src/peft/peft_model.py
```diff
@@ -721,16 +748,82 @@ def load_adapter(self, model_id: str, adapter_name: str, is_trainable: bool = Fa
...
+                    offload_index[new_key] = offload_index[key]
+                    del offload_index[key]
+
+                # rename safetensors for dispatch
+                file_seen = set()
+                for new_key in list(offload_index.keys()):
+                    fname = offload_index[new_key]["safetensors_file"]
+
+                    # make a new file name
+                    new_fname_list = list(fname.split("/"))
```
BenjaminBossan, line 780: Here and below, let's not hard-code `"/"` but use `os.path.sep`.

src/peft/peft_model.py
```diff
@@ -721,16 +748,82 @@ def load_adapter(self, model_id: str, adapter_name: str, is_trainable: bool = Fa
...
+                            new_fname_list[i] += "-peft"
+                    new_fname = "/".join(new_fname_list)
+
+                    if fname not in file_seen:
+                        safe_dict = {}
+                        with safe_open(fname, framework="pt") as f:
+                            for safe_key in f.keys():
+                                safe_tensor = f.get_tensor(safe_key)
+                                metadata = f.metadata()
+                                if any(module in safe_key for module in target_modules):
```
BenjaminBossan, line 792: Same comment as above about `target_modules` possibly being a str.

src/peft/peft_model.py
```diff
@@ -337,6 +338,32 @@ def from_pretrained(
...
+        index = None
+        for name, module in model.named_modules():
+            if hasattr(module, "_hf_hook") and hasattr(module._hf_hook, "original_devices"):
+                index = module._hf_hook.weights_map.dataset.index
+                for key in module._hf_hook.original_devices.keys():
+                    if dict(module._hf_hook.original_devices)[key] == torch.device("meta"):
+                        disk_modules.append(str(name) + "." + str(key))
+
+        if disk_modules and hasattr(kwargs, "use_safetensors") and not kwargs["use_safetensors"]:
+            raise AssertionError("Disk offloading not currently only supported for safetensors")
```
BenjaminBossan, line 353:
```suggestion
            raise ValueError("Disk offloading currently only supported for safetensors")
```

That's what you mean, right?

BenjaminBossan (changes requested):
Thanks for working on disk-offloading. I'm not knowledgeable enough about safetensors and disk offloading to fully judge the implementation, so I hope someone else can provide a review.

I still found a few places which I think can be improved, please take a look at my comments.

src/peft/peft_model.py
```diff
@@ -721,16 +748,82 @@ def load_adapter(self, model_id: str, adapter_name: str, is_trainable: bool = Fa
...
+                                    extended_prefix = prefix + block_id + safe_key[:suffix_pos]
+                                    final_key = extended_prefix + ".base_layer" + safe_key[suffix_pos:]
+                                    lora_dict = {
+                                        key: val for key, val in adapters_weights.items() if extended_prefix in key
+                                    }
+
+                                    # add LoRA keys and values to disk offload
+                                    for lora_key, lora_val in lora_dict.items():
+                                        divide = lora_key.rfind(".")
+                                        new_key = lora_key[:divide] + ".default" + lora_key[divide:]
```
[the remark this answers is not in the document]
blbadger: Good idea! I have swapped all 'default' strings to variables assigned to the `self.peft_config.keys()` elements, and this allows us to iterate through all the keys that might be present and save the appropriate values.

src/peft/peft_model.py
```diff
@@ -721,16 +748,82 @@ def load_adapter(self, model_id: str, adapter_name: str, is_trainable: bool = Fa
...
+
+            prefix = "base_model.model."
+            # rename offload index weight and model names
+            target_modules = self.peft_config["default"].target_modules
+            if offload_index:
+                keys = list(offload_index.keys())
+                block_id = keys[0].split(".")[0] + "."  # for writing safetensors key
+
+                for key in keys:
+                    if any(module in key for module in target_modules):
```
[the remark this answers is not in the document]
blbadger: I had not realized that strings could be supplied instead, thanks for the tip on the method to substitute. I have tried substituting in this method but it does not seem to find the match : for example, 

`self._check_target_module_exists(self.peft_config[adapter_name], key)` 

always returns False even when the key should match the adapter's target modules. I will take a look at this some more, however.

blbadger:
No problem @younesbelkada! Thanks very much for the review @BenjaminBossan.  I have gone ahead and implemented most of the suggestions and will finish with the rest soon, and am also working on making the `PeftModel` merging compatible with safetensors disk-offloading too.

There is currently an issue in this PR where some modules are being saved twice, and I am working on sorting out as well.

BenjaminBossan:
> I have gone ahead and implemented most of the suggestions and will finish with the rest soon

Great, thanks a lot, please ping us once the PR is ready for the next review round.

src/peft/peft_model.py
blbadger: Replaced with `os` functions

blbadger:
@BenjaminBossan Finished implementing your suggestions and fixed the double-saving bug so I think we are good to go.  I am still having trouble with swapping in `_check_target_module_exists` for identifying if a layer is a Peft layer, currently done via `any(module in safe_key for module in target_modules)` and any pointers would be much appreciated.

cc @younesbelkada

src/peft/peft_model.py
```diff
@@ -650,6 +677,84 @@ def _split_kwargs(cls, kwargs: Dict[str, Any]):
...
+            adapters_weights (Dict[str: torch.tensor]):
+                Dictionary of Peft adapter module names and weights
+        """
+
+        prefix = "base_model.model."
+        # rename offload index weight and model names
+        adapter_names = list(self.peft_config.keys())
+        for adapter_name in adapter_names:
+            target_modules = self.peft_config[adapter_name].target_modules
+            if offload_index:
```
BenjaminBossan, line 696: So IIUC, if we don't have `offline_index`, nothing really needs to happen. So can we add

```python
if not offload_index:
    return offload_index
```

to the start of the function and remove the `if offload_index:` check here? This should make the code easier to read and save one level of indentation.

src/peft/peft_model.py
```diff
@@ -650,6 +677,84 @@ def _split_kwargs(cls, kwargs: Dict[str, Any]):
...
+                files_seen = set()
+                # rename safetensors for dispatch
+                for new_key in list(offload_index.keys()):
+                    fname = offload_index[new_key]["safetensors_file"]
+
+                    # make a new file name
+                    new_fname_list = list(fname.split(os.sep))
+                    for i, name in enumerate(new_fname_list):
+                        if "--" in name:
+                            new_fname_list[i] += "-peft"
```
BenjaminBossan, line 720: We can `break` after the first match, right?

src/peft/peft_model.py
```diff
@@ -650,6 +677,84 @@ def _split_kwargs(cls, kwargs: Dict[str, Any]):
...
+                    new_fname = os.path.join(*new_fname_list)
+
+                    if fname not in files_seen:
```
BenjaminBossan, line 723: Can we do:

```python
if fname in files_seen:
    continue
safe_dict = {}
...
```

again, for readability and to unindent one level?

src/peft/peft_model.py
```diff
@@ -650,6 +677,84 @@ def _split_kwargs(cls, kwargs: Dict[str, Any]):
...
+        Update the offload_index and safetensors files for loading and mergine PeftModels with disk-offloaded modules.
+
+        Args:
+            offload_index (Dict[str: str]):
+                Dictionary of disk-offloaded modules with their metadata and safetensors filenames
+            adapters_weights (Dict[str: torch.tensor]):
+                Dictionary of Peft adapter module names and weights
+        """
+
+        prefix = "base_model.model."
```
BenjaminBossan, line 691: I wonder: Should we create a shallow copy of `offload_index`? This would be to avoid having a function that both returns a value and mutates an input argument. Alternatively, we could make the function not return anything but mutate the `offload_index`. So below we would have:

```diff
- offload_index = self._update_offload(offload_index, adapters_weights)
+ self._update_offload(offload_index, adapters_weights)
```

I personally prefer the latter suggestion. WDYT?

src/peft/tuners/tuners_utils.py
```diff
@@ -66,9 +67,21 @@ def onload_layer(layer):
...
         and layer.base_layer._hf_hook.offload
     ):
-        if torch.device("meta") in layer.base_layer._hf_hook.original_devices.values():
-            # retrieve the name of the original disk-offload directory
-            offload_folder = layer.base_layer._hf_hook.weights_map.dataset.save_folder
+        if torch.device("meta") in layer.base_layer._hf_hook.original_devices.values() and hasattr(
+            layer.base_layer._hf_hook.weights_map, "dataset"
+        ):
+            index = layer.base_layer._hf_hook.weights_map.dataset.index
+            module_name = list(dict(layer.base_layer._hf_hook.weights_map.dataset).keys())[0]  # any module will do
```
BenjaminBossan, line 74: Could you give me a pointer under what cirumstances `_hf_hook.weights_map` can have a `dataset` argument? Let's also add a comment to explain it.

BenjaminBossan (changes requested):
Thanks so much for the updates. I did another review, please check my comments.

> I am still having trouble with swapping in _check_target_module_exists for identifying if a layer is a Peft layer

So I think the reason why this fails is because the check is supposed to be on the base model and not on the `PeftModel`. The latter has different parameter names, since layers have been replaced, resulting in there being no match.

I wonder: Could we do a check like `if isinstance(module, BaseTunerLayer)` instead?

src/peft/peft_model.py
```diff
@@ -337,6 +338,32 @@ def from_pretrained(
...
+        disk_modules = []
+        index = None
+        for name, module in model.named_modules():
+            if hasattr(module, "_hf_hook") and hasattr(module._hf_hook, "original_devices"):
+                index = module._hf_hook.weights_map.dataset.index
+                for key in module._hf_hook.original_devices.keys():
+                    if dict(module._hf_hook.original_devices)[key] == torch.device("meta"):
+                        disk_modules.append(str(name) + "." + str(key))
+
+        if disk_modules and hasattr(kwargs, "use_safetensors") and not kwargs["use_safetensors"]:
```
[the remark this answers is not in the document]
blbadger: Nice one, looks it should indeed

src/peft/peft_model.py
```diff
@@ -337,6 +338,32 @@ def from_pretrained(
...
 
+        weight_map = dict(named_module_tensors(model, recurse=True))
+
+        disk_modules = []
+        index = None
+        for name, module in model.named_modules():
+            if hasattr(module, "_hf_hook") and hasattr(module._hf_hook, "original_devices"):
+                index = module._hf_hook.weights_map.dataset.index
+                for key in module._hf_hook.original_devices.keys():
+                    if dict(module._hf_hook.original_devices)[key] == torch.device("meta"):
```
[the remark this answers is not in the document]
blbadger: Good catch, it is indeed so the `dict()` has been removed

src/peft/peft_model.py
```diff
@@ -337,6 +338,32 @@ def from_pretrained(
         else:
             raise ValueError(f"The input config must be a PeftConfig, got {config.__class__}")
 
+        weight_map = dict(named_module_tensors(model, recurse=True))
+
+        disk_modules = []
```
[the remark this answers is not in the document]
blbadger: It should indeed, changed to a `set`

src/peft/peft_model.py
blbadger: Very true, and it does make everything a bit more readable

src/peft/peft_model.py
blbadger: Good idea and yes indeed, only the `models--xxxxx` dirname is changed

src/peft/peft_model.py
blbadger: LGTM

src/peft/peft_model.py
blbadger: I think the second should be cleaner, and you are right that we don't really need to return the `offload_index` anyways

src/peft/tuners/tuners_utils.py
blbadger: My understanding is that the `dataset` argument should be present for every module that has one or more offloaded sub-modules, although is not necessarily absent if cpu offloading is performed. The relevant code is in `accelerate/hooks.py` in `attach_align_device_hook()`, where the `dataset` is the `PrefixedDataset` object which is called if there is offloading done via buffers (len(list(directs)) > 0),

```python
    if len(list(directs)) > 0 or full_offload:
        if weights_map is not None:
            prefix = f"{module_name}." if len(module_name) > 0 else ""
            prefixed_weights_map = PrefixedDataset(weights_map, prefix)
        else:
            prefixed_weights_map = None
        hook = AlignDevicesHook(
            execution_device=execution_device,
            offload=offload,
            weights_map=prefixed_weights_map,
```

the `dataset` contains the offload `index`, which contains the map of each parameter to the safetensors file of interest. I have added a comment:)

src/peft/peft_model.py
```diff
@@ -650,6 +677,86 @@ def _split_kwargs(cls, kwargs: Dict[str, Any]):
...
+                    # avoid overwriting original safetensors
+                    for key in safe_dict.keys():
+                        offload_index[key] = {"safetensors_file": new_fname, "weight_name": key}
+
+                    base_name = os.path.dirname(new_fname)
+                    if not os.path.exists(base_name):
+                        os.makedirs(base_name)
+                    safe_save_file(safe_dict, new_fname, metadata=metadata)
+
+        return
```
BenjaminBossan, line 758: Can be removed

src/peft/tuners/tuners_utils.py
BenjaminBossan: Thanks for explaining, I didn't know about this. Regarding `layer.base_layer._hf_hook.weights_map.dataset.index`, from my understanding, `dataset` is just a `Mapping` (e.g. a `state_dict`), where does the `index` attribute come from?

src/peft/peft_model.py
```diff
@@ -337,6 +338,32 @@ def from_pretrained(
         else:
             raise ValueError(f"The input config must be a PeftConfig, got {config.__class__}")
 
+        weight_map = dict(named_module_tensors(model, recurse=True))
+
+        disk_modules = set()
+        index = None
+        for name, module in model.named_modules():
+            if hasattr(module, "_hf_hook") and hasattr(module._hf_hook, "original_devices"):
+                index = module._hf_hook.weights_map.dataset.index
```
BenjaminBossan, line 347: Same question as below about whether we can assume that the `index` attribute always exists.

BenjaminBossan (commented):
Thanks for the updates. I have some more comments, please take a look.

Regarding the merge conflict, we recently switched to plain `assert`s in test instead of using unittest's `self.assertFoo` style.

src/peft/peft_model.py
blbadger: Removed the now-superfluous return

blbadger:
Thanks very much for the review @BenjaminBossan ! With some tweaks, `if isinstance(module, BaseTunerLayer)` did indeed do the trick, nice one! I will finish working through the comments and will resolve the merge conflict too, appreciate having the heads up about switching to `assert`

src/peft/tuners/tuners_utils.py
blbadger: The tricky thing here is that the `dataset` is actually an instance of `class OffloadedWeightsLoader(Mapping)` once the mapping is offloaded (in `accelerate.utils.offload.py`), and can have optional args `state_dict`, `save_folder`, and `index`. The `index` can be read from JSON in the `save_folder` if it is not supplied (this is why we have to give our own `index` while renaming safetensors).

BenjaminBossan:
A more general comment on this PR: I'm a little bit anxious that we have to rely heavily on implementation details of accelerate, e.g. [here](https://github.com/huggingface/peft/pull/1431#discussion_r1490262613). Although we of course try to ensure that accelerate and PEFT work well together, these types of details are easy to overlook. I would be more comfortable if accelerate provided some kind of utility function to achieve this goal, but this use case is so entangle with PEFT-specifics that it would not make sense.

As a next best safeguard, I wonder if we should have a CI workflow for accelerate-main + PEFT-main. I couldn't find an existing one on accelerate or PEFT at first glance.

@muellerzr Do you have a good idea?

src/peft/peft_model.py
blbadger: AFAIK every deployed `dataset` should have an `index`, but perhaps we should check for the `dataset`. I have modified for that check now

src/peft/tuners/tuners_utils.py
blbadger: Agreed, and this certainly wasn't obvious to me either so I will add a comment

blbadger:
@BenjaminBossan just finished with the last comments, thanks for the thorough review! Happy to address anything else that comes up too.

BenjaminBossan (approved):
Thanks a lot for your great work and digging deep into the issue. The PR LGTM.

Before merging, it would be great if someone from accelerate could take a look at this, especially with regard to what I mentioned in my previous comment about relying heavily on accelerate implementation details here. Maybe @SunMarc or @muellerzr when he's back in office.

Perhaps there is no better way and we just need to accept it as is, but I'd like to hear some more opinions first :)

BenjaminBossan:
Sorry for the delay, I'm still waiting for another reviewer.

SunMarc (approved):
Thanks for contributing @blbadger 🔥 LGTM ! We can't offload much of the work done here in accelerate since most of the code here relies heavily on peft details (e.g. when updating the `offload_index`). 
One thing that could be done is to add a attribute to know on which device the module is offloaded to the `cpu` or `disk`, hence we won't have to check for "meta" device + and the `dataset` each time. We can get this information from the `hf_device_map` but it is only accessible at the level of the model.
 As for your question about the `dataset` attribute in `weights_map`, we will always have it for offloaded modules to the "cpu" or "disk". This is linked to the `dispatch_model` where we define [`weights_map`](https://github.com/huggingface/accelerate/blob/e8aaee5d9b9333979546f13e245e34380010fa54/src/accelerate/big_modeling.py#L393) in the offload case + it will call the following [function](https://github.com/huggingface/accelerate/blob/e8aaee5d9b9333979546f13e245e34380010fa54/src/accelerate/hooks.py#L434). It is also inside the `dispatch_model` function where we define `hf_device_map`. So the code proposed here should work fine.

src/peft/peft_model.py
```diff
@@ -336,6 +338,33 @@ def from_pretrained(
         else:
             raise ValueError(f"The input config must be a PeftConfig, got {config.__class__}")
```
SunMarc, line 340: maybe add a check for `hf_device_map` here and that we indeed that modules offloaded to `disk`

src/peft/peft_model.py
```diff
@@ -336,6 +338,33 @@ def from_pretrained(
...
+                    if module._hf_hook.original_devices[key] == torch.device("meta"):
+                        disk_modules.add(str(name) + "." + str(key))
+
+        if disk_modules and not kwargs.get("use_safetensors", True):
+            raise ValueError("Disk offloading currently only supported for safetensors")
+
+        start_prefix = ""
+        if index:
+            offload_index = {
+                p[len(start_prefix) :]: {
```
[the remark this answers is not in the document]
blbadger: Good catch! I don't think so, it has been removed and this allows the offload_index to be simplified too

src/peft/peft_model.py
```diff
@@ -336,6 +338,33 @@ def from_pretrained(
         else:
             raise ValueError(f"The input config must be a PeftConfig, got {config.__class__}")
 
+        weight_map = dict(named_module_tensors(model, recurse=True))
+
+        disk_modules = set()
```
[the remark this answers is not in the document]
blbadger: Sure thing, added a comment for clarity

src/peft/peft_model.py
```diff
@@ -336,6 +338,33 @@ def from_pretrained(
...
+        index = None
+        for name, module in model.named_modules():
+            if hasattr(module, "_hf_hook") and hasattr(module._hf_hook, "original_devices"):
+                if hasattr(module._hf_hook.weights_map, "dataset"):
```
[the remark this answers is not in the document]
blbadger: I was not aware of `expand_device_map`, thanks for the tip! I might prefer keeping this as-is if you don't have a preference

src/peft/peft_model.py
```diff
@@ -649,6 +678,86 @@ def _split_kwargs(cls, kwargs: dict[str, Any]):
...
+                    files_seen.add(new_fname)
+
+                    # avoid overwriting original safetensors
+                    for key in safe_dict.keys():
+                        offload_index[key] = {"safetensors_file": new_fname, "weight_name": key}
+
+                    base_name = os.path.dirname(new_fname)
+                    if not os.path.exists(base_name):
+                        os.makedirs(base_name)
+                    safe_save_file(safe_dict, new_fname, metadata=metadata)
```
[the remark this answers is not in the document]
blbadger: That is correct: the original idea was to avoid using extra storage by saving the new safetensors to the original file name (modifying the metadata if necessary), but I ran into some problems with that approach that were difficult to debug. Hence this approach of creating new safetensors files and saving those instead.

blbadger:
@SunMarc Thanks very much for the review! Agreed that it would be a good idea to check for a `hf_device_map` before iterating through modules to look for disk-offloaded ones, and I have added that as a condition before the module iteration.

As for using the `hf_device_map` together with `expand_device_map` to get disk-offload name information, if either of you think we should make that change I would be happy to implement! If no one has a strong opinion I would be inclined to leave as-is, as it seems we would need to map the parameters to modules before iterating through the modules in order to use `expand_device_map`

BenjaminBossan:
> As for using the `hf_device_map` together with `expand_device_map` to get disk-offload name information, if either of you think we should make that change I would be happy to implement! If no one has a strong opinion I would be inclined to leave as-is, as it seems we would need to map the parameters to modules before iterating through the modules in order to use `expand_device_map`

I'm fine with keeping it as is.

If Marc is fine with the last updates you made, we can go ahead and merge.

blbadger:
> > As for using the `hf_device_map` together with `expand_device_map` to get disk-offload name information, if either of you think we should make that change I would be happy to implement! If no one has a strong opinion I would be inclined to leave as-is, as it seems we would need to map the parameters to modules before iterating through the modules in order to use `expand_device_map`
> 
> I'm fine with keeping it as is.
> 
> If Marc is fine with the last updates you made, we can go ahead and merge.

Awesome, sounds good to me!

SunMarc (approved):
LGTM ! The changes looks good to me @blbadger !

BenjaminBossan:
Thanks so much @blbadger for this big PR, as well as Marc for your review.

blbadger:
My pleasure, thanks very much for all your help @BenjaminBossan @SunMarc !

PR #1433, opened 2024-02-05
Add Support for Mistral Model in Llama-Adapter Method

PrakharSaxena24:
Hello PEFT team,
**Purpose of This PR:**
 Add support for Mistral model for llama-adapter method.
 
 **Background:**
I wanted to test how does the method in [this paper](https://arxiv.org/abs/2303.16199) works with Mistral based models. compared to Lora. Initially I though that since the architecture of Llama and Mistral are almost the same, this could be achieved by just changing the config, however I found out that the mistral models k_proj and v_proj dimensions are different from that of Llama. 
Hence I added the model support for Mistral in the llama-adapter method (the naming is confusing).
I hope it will be useful for anyone else willing to experiment with different methods.

**Request for Review:**

Please provide review and let me know if my implementation makes sense.

Thank you for all your hardwork!


src/peft/tuners/adaption_prompt/layer.py
```diff
@@ -75,21 +75,25 @@ def forward(self, **kwargs):
         k_proj_layer = TRANSFORMERS_MODEL_CONFIG[self.model_type].k_proj_layer
         v_proj_layer = TRANSFORMERS_MODEL_CONFIG[self.model_type].v_proj_layer
         o_proj_layer = TRANSFORMERS_MODEL_CONFIG[self.model_type].o_proj_layer
+        factor = (
```
BenjaminBossan, line 78: I wonder if it would be more cautious to check that the two values divide without remainder. Or should it be impossible to have a remainder?

factor, remainder = divmod(self.model.k_proj.in_features // self.model.k_proj.out_features)
if remainder:
    raise ValueError(...)

src/peft/tuners/adaption_prompt/layer.py
```diff
@@ -100,6 +104,15 @@ def forward(self, **kwargs):
         query_states = compute_query_states(model=self.model, **kwargs)
 
         previous_dtype = query_states.dtype
+
+        # Reshape and average the extra tensors
+        query_states_reshaped = query_states.reshape(
+            bsz, self.model.num_heads, -1, (self.model.head_dim // factor), factor
+        )
+
+        # Take the mean along the last dimension to get [bsz, 32, X, 32]
```
BenjaminBossan, line 113: Just for my understanding, what is the intuition behind taking the mean here?

src/peft/tuners/adaption_prompt/layer.py
```diff
@@ -109,6 +122,10 @@ def forward(self, **kwargs):
         scores = self.adaption_gate * F.softmax(scores, dim=-1, dtype=torch.float32).to(previous_dtype)
         # (bsz, q_len, num_heads * head_dim)
         adapter_output = torch.matmul(scores, adapter_v).transpose(1, 2).reshape(bsz, q_len, -1)
+
+        adapter_output = torch.repeat_interleave(
+            adapter_output, repeats=factor, dim=2
+        )  # https://github.com/huggingface/transformers/blob/e547458c43dfdbbb8f6a7757237e234c44e20a8f/src/transformers/models/mistral/modeling_mistral.py#L181
```
BenjaminBossan, line 128: Could you please put the comment on a separate line, this should make the code more readable.

BenjaminBossan (commented):
Thanks for adding support for Mistral. This looks very promising. I only have a couple of comments, please check them out. Also, would it be possible to add a unit test to `tests/test_adaption_prompt.py` involving a small mistral model?

src/peft/tuners/adaption_prompt/layer.py
PrakharSaxena24: The shape of query states is not compatible of that of adapter_k, which is required to calculate the adapter_output. Hence, to reshape it I took the average. Another way would be to drop the extra states, however averaging can prevent loss of information(to some extent). As a result I took the average. 
Please let me know if there is a better way!

src/peft/tuners/adaption_prompt/layer.py
PrakharSaxena24: Sure!

src/peft/tuners/adaption_prompt/layer.py
PrakharSaxena24: Thank you for the prompt reply!
Upon checking [modeling_mistral.py](https://github.com/huggingface/transformers/blob/1c31b7aa3bb4e7ef24c77596d2a76f45a770159f/src/transformers/models/mistral/modeling_mistral.py#L214), it seems like it should not be possible to have a remainder. 
Please let me know your thoughts.

src/peft/tuners/adaption_prompt/layer.py
PrakharSaxena24: Will also add the unit test!

src/peft/tuners/adaption_prompt/layer.py
```diff
@@ -100,6 +104,15 @@ def forward(self, **kwargs):
         query_states = compute_query_states(model=self.model, **kwargs)
 
         previous_dtype = query_states.dtype
+
+        # Reshape and average the extra tensors
```
pacman100, line 108: No need to reshape and avg query states as the above key shape is (bsz, adapter_seq_len, num_kv_heads, head_dim), the value shape is (bsz, adapter_seq_len, num_kv_heads, head_dim) and query shape is (bsz, adapter_seq_len, num_heads, head_dim). Now, you would need to repeat the `num_kv_heads` to match `num_heads` as done in https://github.com/huggingface/transformers/blob/1c31b7aa3bb4e7ef24c77596d2a76f45a770159f/src/transformers/models/mistral/modeling_mistral.py#L193. After that the attn computation is same as normal MHA case.

src/peft/tuners/adaption_prompt/layer.py
```diff
@@ -75,21 +75,25 @@ def forward(self, **kwargs):
...
         if k_proj_layer == v_proj_layer:
             _, key, value = getattr(self.model, k_proj_layer)(self.adaption_prompt).split(embed_dim, dim=2)
         else:
             key = getattr(self.model, k_proj_layer)(self.adaption_prompt)
             value = getattr(self.model, v_proj_layer)(self.adaption_prompt)
         # (bsz, num_heads, adapter_len, head_dim)
+
         adapter_k = (
-            key.view(1, self.adapter_len, self.model.num_heads, self.model.head_dim)
+            key.view(1, self.adapter_len, self.model.num_heads, (self.model.head_dim // factor))
```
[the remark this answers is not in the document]
PrakharSaxena24: I see! Thanks a lot, this seems correct.
Will edit this.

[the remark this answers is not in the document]
PrakharSaxena24: Also I think I will need to do the same in utils.py

src/peft/tuners/adaption_prompt/layer.py
PrakharSaxena24: Thanks a lot, so rather than repeating the adapter output, I should repeat adapter_k and adapter_v.
`
        adapter_k = torch.repeat_interleave(
            adapter_k, repeats=factor, dim=1
        ) 
        adapter_v = torch.repeat_interleave(
            adapter_v, repeats=factor, dim=1
        )
`
Does this makes sense?

PrakharSaxena24:
@BenjaminBossan @pacman100  thank you for you kind comments.
, I have corrected the logic and added test. 
I did not add test for `test_bf16_inference` as I could not find any mistral model in [here](https://huggingface.co/trl-internal-testing). 
Please have a look.
Thanks for your hard work!

tests/test_adaption_prompt.py
```diff
@@ -409,6 +774,32 @@ def test_use_cache(self) -> None:
...
+                MistralConfig(
+                    vocab_size=16,
+                    hidden_size=8,
+                    intermediate_size=8,
+                    num_hidden_layers=8,
+                    num_attention_heads=4,
+                    num_key_value_heads=2,
+                    use_cache=False,
+                )
+        ).eval()
```
BenjaminBossan, line 789: Wrong indentation.

tests/test_adaption_prompt.py
```diff
@@ -78,7 +106,19 @@ def test_attributes(self) -> None:
         self.assertTrue(hasattr(model, "from_pretrained"))
         self.assertTrue(hasattr(model, "push_to_hub"))
 
+        #Test Mistral
+        if self.mistral_available:
```
BenjaminBossan, line 110: Instead of attaching the mistral tests to the llama tests, could you please create a separate test for each? You can decorate them with `@unittest.skipIf(not is_mistral_available())` to avoid the `if self.mistral_available:` line.

src/peft/tuners/adaption_prompt/config.py
```diff
@@ -55,6 +55,13 @@ def is_adaption_prompt(self) -> bool:
         v_proj_layer="v_proj",
         o_proj_layer="o_proj",
     ),
+    "mistral": ModelTypeConfig(  # same for Mistral,
```
BenjaminBossan, line 58: Regarding the comment: Do you mean "same as llama"?

BenjaminBossan (changes requested):
Thanks a lot. From my side, there are only a few comments left, please take a look.

src/peft/tuners/adaption_prompt/config.py
PrakharSaxena24: Yes I mean same as llama. 
I can change the comment.

tests/test_adaption_prompt.py
PrakharSaxena24: Thank for pointing out.
Will make it correct!

tests/test_adaption_prompt.py
PrakharSaxena24: Thank you for the advice! I will do it. Good feedback is very helpful for me to learn and apply the best practices :)

BenjaminBossan (approved):
Thanks a lot for this PR, it looks good now. I also appreciate extending the existing tests.

Let's wait for @pacman100's final review before merging.

PrakharSaxena24:
@pacman100 please let me know it there is something to change in the PR.

BenjaminBossan:
@PrakharSaxena24 Heads up, there is now a merge conflict, which stems from a recent PR where we switched from unittest-style `self.assertFoo(...)` to pytest-style plain `assert`s. Could you please fix the conflict?

PrakharSaxena24:
@BenjaminBossan Changed the assert to pytest style. Moreover there was a conflict in utils.py which was also resolved. Please have a look!

BenjaminBossan (approved):
Thanks a lot, the PR LGTM. Let's wait for a final review by @pacman100 before merging.

PrakharSaxena24:
@BenjaminBossan 
Current pr and [this PR](https://github.com/huggingface/peft/pull/1451) do very similar things, current PR adds Mistral and also Llama2 34b and 70b (GQA).
The PR above adds GQA(Llama2 34b and 70b).
I think that the there might be conflict if both are merged to the main. 
How do you think we should proceed?

BenjaminBossan:
> I think that the there might be conflict if both are merged to the main.
> How do you think we should proceed?

Yes, there will be conflicts, so whoever comes last will have to resolve them :)

I don't think it's a huge issue. Since both PRs have tests, we should hopefully have the guard rails to ensure that resolving the merge conflict won't lead to a regression in the other PR.

PrakharSaxena24:
> I don't think it's a huge issue. Since both PRs have tests, we should hopefully have the guard rails to ensure that resolving the merge conflict won't lead to a regression in the other PR.

Thank you for the reply!

pacman100 (approved):
Thank you @PrakharSaxena24 for supporting Mistral with Adaptation Prompt and the detailed tests! ✨

Issue #1392, opened 2024-01-24
[Q] Does support for embedding and conv2d layers in bnb.py at tuners/lora make sense?


GM-git-dotcom:
### Feature request

Currently, `bnb.py` at `peft/tuners/lora` has classes for `Linear4bit` and `Linear8bit`. But `layers.py` at `peft/tuners/lora` defines implementations for LoRA in `Linear`, `Embedding`, and `Conv2d` (I believe in the future, it would also include `Conv3d` as per #1193).

I was wondering if it makes sense to extend quantization support for `Embedding` and `Conv2d` layers? Could they be a thing in `bnb.py`? (possibly in other tuners as well). Sorry if this is obvious.

### Motivation

I was just curious about `bnb.py` only having support for Linear when LoRA can be implemented through other layer types (please correct me if I'm wrong).

### Your contribution

With some help, I would be happy to contribute a PR towards this, starting with bnb for `Embedding`.

BenjaminBossan:
Yes, in principle, it should be possible add support for the `Embedding` layer from bnb to PEFT. Feel free to open a PR and ping us so that we can give feedback and help with open questions.

Regarding `Conv2d`, I don't think it's supported by bnb and thus could not be easily added to PEFT.

asr-sheep1:
Thank you very much for making such a great project open source, I learned a lot of new things by using PEFT, thank you again. I have a puzzle of  `class Linear8bitLt` in `peft/src/peft/tuners/lora/bnb.py`. By reading the code I understand that it is different from the definition in bitsanbytes because the latter uses int8 precision during calculation. I used PEFT in Alpaca-lora finetune. In the transfoemer model obtained with `is_load_8bit`, the parts (q, v) replaced by lora still use `Linear8bitLt `on the surface, but actually use `torch.nn.Linear` in `peft/src/peft/tuners/lora/layer.py update_layer()`.  Lora layer does not use int8 to participate in the calculation, it is for the sake of accuracy. Is that right, or did I not understand the whole calling program correctly?

younesbelkada:
Hi @asr-sheep1 

Below is what happens under the hood if you initialize a `Linear8bitLt` class from peft

![lora_animated](https://github.com/huggingface/peft/assets/49240599/36bccc04-130c-482b-b3e8-870bf6cd4136)

In blue you have the original linear layer (which is in 8bit) and we simply attach the LoRA adapters next to it and calculate the sum of both hidden states.

[...]

PR #1436, opened 2024-02-05
[WIP] Adding support for Embedding4bit in bnb.py at tuners/lora

GM-git-dotcom:
This adds `bitsandbytes` support for embedding layers by adding an `Embedding4bit` class at `peft/tuners/lora/bnb.py` as discussed in this issue #1392. I will keep working and updating this PR as time allows. 

### Update 14 Feb 24 
- _As discussed_: Removed overloaded `update_layer` as we do not need to quantize LoRA params.
- Removed resolved todo in comments.
- Planned unit tests.
- Updated current todo: `Embedding4bit` imports are handled, and all unit tests will come under `test_common_gpu`.

### Current todo
- [x] Make necessary changes to `dispatch_bnb_4bit` to incorporate instances of `bnb.nn.Embedding` class.
- [x] Complete ~~update_layer~~, `merge`, `unmerge`, `get_delta_weight`, `_embed`, `forward`.
- [x] Inspect `bitsandbytes` for an implementation for 8-bit version, as for `Linear8bitLt`.
- [x] Check files where `Embedding4bit` would be imported.
- [ ] Add doc to demonstrate manually replacing `Embedding` layers with `bnb.nn.Embedding` for everything to work.
- [x] Check from tests: `test_common_gpu`, `test_gpu_examples`, `test_tuners_utils`, `test_regression` for writing potential tests to verify implementation.
- [ ] Make unit tests: 
  - `test_lora_bnb_4bit_quantization`
  - `test_embedding4bit_merge_lora`
  - `test_embedding4bit_merge_and_disable_lora`

____________
_Older updates below..._
**11 Feb 24.**
- **Completed all methods;** added `unmerge`, `forward`, ~~update_layer~~, and `dispatch_bnb_4bit`.
- _As discussed_: Added `**kwargs**` arguments to the superclass `LoyaLayer` initialization calls in `Linear8bitLt`, `Linear4bit`, and `Embedding4bit`.
- _As discussed_: Removed `transpose` (and thus requirement of `fan_in_fan_out`) from `get_delta_weight` in `Embedding4bit`.
- Added new imports in `bnb.py`: `math` and `torch.nn as nn` for `update_layer`.
- Added `Embedding4bit` import to `__init__.py` at `tuners/lora` (avoiding eager import).

**07 Feb 24.** 
- Completed methods `get_delta_weight`, `_embed`, and `merge`.
- Import added for `torch.nn.functional as F` as it is used for `_embed`.

**05 Feb 24.** Added skeleton for `Embedding4bit` class in `bnb.py`. 
_Remarks:_ You may notice the last commit undoes the first two commits. By error, I had these commits from another branch that I was working on for documentation. It ultimately did not need to be merged, and can be found here: #1386. 
____________

Any thoughts/comments/help as I continue would be much appreciated! Thanks to @BenjaminBossan


BenjaminBossan:
Thanks a lot @GM-git-dotcom, this PR looks like it's on the right track. Tip: Adding a test may help you to iterate faster, as you can quickly check if your current implementation works.

Could you please remove the package-lock.json file?

If you have specific questions, feel free to ask them.

GM-git-dotcom:
I've added the 07 Feb 24 update in the PR. Some questions I had while working on it -

1. I'm not sure if we have to pass `**kwargs` for `LoraLayer` here.
```py
class Embedding4bit(torch.nn.Module, LoraLayer):  # WORK IN PROGRESS
        # LoRA implemented in an Embedding layer
        def __init__(
            self,
            ...
            **kwargs,
        ) -> None:
            super().__init__()
            LoraLayer.__init__(self, base_layer) # <--- possibly LoraLayer.__init__(self, base_layer, **kwargs)?
            ...
```
It wasn't passed in the init method of `Linear4bit` nor `Linear8bitLt` but is passed in the `Linear` at `tuners/lora/layer.py` (but not for `Embedding` there).

2. The `merge` method for `Linear` in `bnb.py` has this warning:
```py
...
    if active_adapter not in self.lora_A.keys():
        continue
    warnings.warn(
    "Merge lora module to 4-bit linear may get different generations due to rounding errors."
     )
     # Refer to https://gist.github.com/ChrisHayduk/1a53463331f52dca205e55982baf9930
     ...
```
Would this behavior be consistent with the `Embedding4bit` merge you think? If so, I will add a warning.

3. The `get_delta_weight` method for `Linear4bit` was defined as follows:
```py
def get_delta_weight(self, adapter):
    return (
        transpose(
                self.lora_B[adapter].weight @ self.lora_A[adapter].weight,
                False,
            )
            * self.scaling[adapter]
    )
```
Currently, I have kept it the same for `Embedding4bit`, but I was wondering if we should use `False` for the `fan_in_fan_out` argument in `transpose()` above. Keeping it `False` makes `Transpose` simply return the product, so why even use transpose here? Interestingly, the `get_delta_weight` method for `Embedding` under `tuners/lora/layer.py` uses `True` for the `fan_in_fan_out`, while for `Linear` it is set to take `self.fan_in_fan_out`. What should I do?

GM-git-dotcom:
I've added the 07 Feb 24 update in the PR. Some questions I had while working on it -

1. I'm not sure if we have to pass `**kwargs` for `LoraLayer` here.
```py
class Embedding4bit(torch.nn.Module, LoraLayer):  # WORK IN PROGRESS
        # LoRA implemented in an Embedding layer
        def __init__(
            self,
            ...
            **kwargs,
        ) -> None:
            super().__init__()
            LoraLayer.__init__(self, base_layer) # <--- possibly LoraLayer.__init__(self, base_layer, **kwargs)?
            ...
```
It wasn't passed in the init method of `Linear4bit` nor `Linear8bitLt` but is passed in the `Linear` at `tuners/lora/layer.py` (but not for `Embedding` there).

2. The `merge` method for `Linear` in `bnb.py` has this warning:
```py
...
    if active_adapter not in self.lora_A.keys():
        continue
    warnings.warn(
    "Merge lora module to 4-bit linear may get different generations due to rounding errors."
     )
     # Refer to https://gist.github.com/ChrisHayduk/1a53463331f52dca205e55982baf9930
     ...
```
Would this behavior be consistent with the `Embedding4bit` merge you think? If so, I will add a warning.

3. The `get_delta_weight` method for `Linear4bit` was defined as follows:
```py
def get_delta_weight(self, adapter):
    return (
        transpose(
                self.lora_B[adapter].weight @ self.lora_A[adapter].weight,
                False,
            )
            * self.scaling[adapter]
    )
```
Currently, I have kept it the same for `Embedding4bit`, but I was wondering if we should use `False` for the `fan_in_fan_out` argument in `transpose()` above. Keeping it `False` makes `Transpose` simply return the product, so why even use transpose here? Interestingly, the `get_delta_weight` method for `Embedding` under `tuners/lora/layer.py` uses `True` for the `fan_in_fan_out`, while for `Linear` it is set to take `self.fan_in_fan_out`. What should I do? 

Thanks in advance. @BenjaminBossan

BenjaminBossan:
Thanks for your continued work on this. Regarding your questions:

> 1. I'm not sure if we have to pass `**kwargs` for `LoraLayer` here.

Good catch. Yeah, it's better if we pass the `kwargs`, even if they're not used at the moment, for consistency. Feel free to update the bnb linear layers as well.

> 2\. The `merge` method for `Linear` in `bnb.py` has this warning:
> Would this behavior be consistent with the Embedding4bit merge you think? If so, I will add a warning.

Yes, we should expect the same here. Instead of just making assumptions, ideally, we can test for this. For example, check out this test:

https://github.com/huggingface/peft/blob/055e4dbe1e9f82df44e7aaac990752338219efeb/tests/test_common_gpu.py#L685

There we have relatively high tolerances because of the error being introduced.

> Currently, I have kept it the same for `Embedding4bit`, but I was wondering if we should use `False` for the `fan_in_fan_out` argument in `transpose()` above.

It shouldn't be necessary to use `transpose` and `fan_in_fan_out` here.

GM-git-dotcom:
Added the **11 Feb update**. Some concerns I had -

1. The `forward` method in `Linear4bit` has defensive cloning:

https://github.com/huggingface/peft/blob/a1c472f08f39a4b95a228dd436944bc3a75406ea/src/peft/tuners/lora/bnb.py#L302-L316

I've done the same for `Embedding4bit` for now, as I could not find more info about it online apart from the comment made. Ideally, should we test if defensive cloning is needed for `Embedding4bit`? How does one test for this?

2. The `forward` method in `Embedding` at `lora/layer.py` does not convert dtype:

https://github.com/huggingface/peft/blob/a1c472f08f39a4b95a228dd436944bc3a75406ea/src/peft/tuners/lora/layer.py#L473-L474

Per this, I did not convert dtypes for `Embedding4bit`'s `forward`. In case it is not required for `Embedding` either, let me know and I will remove that comment. Otherwise, I will work on checking for dtype conversion in `forward` for `Embedding4bit` (and `Embedding` in `layer.py` if needed).

3. Currently, I have `update_layer` set up like this:
```py
def update_layer(self, adapter_name, r, lora_alpha, lora_dropout, init_lora_weights, use_rslora):
    # TODO: Check update_layer for correctness
    if r <= 0:
        raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")

    self.r[adapter_name] = r
    self.lora_alpha[adapter_name] = lora_alpha
    if lora_dropout > 0.0:
        lora_dropout_layer = nn.Dropout(p=lora_dropout)
    else:

    weight = getattr(base_layer, "weight", None)
    # Actual trainable parameters
    weight_A = torch.randn((r, self.in_features))
    weight_B = torch.randn((self.out_features, r))
    # Quantize to 4 bit
    weight_A, _ = bnb.functional.quantize_4bit(weight_A)  # TODO: Quantize using Params4bit instead?
    weight_B, quant_state = bnb.functional.quantize_4bit(weight_B)

    self.lora_embedding_A[adapter_name] = nn.Parameter(weight_A)
    self.lora_embedding_B[adapter_name] = nn.Parameter(weight_B)

    if use_rslora:
        self.scaling[adapter_name] = lora_alpha / math.sqrt(r)
    else:
        self.scaling[adapter_name] = lora_alpha / r
...
```
Specifically these lines from above, where I quantize params to 4 bits -
```py
weight_A, _ = bnb.functional.quantize_4bit(weight_A)  # TODO: Quantize using Params4bit instead?
weight_B, quant_state = bnb.functional.quantize_4bit(weight_B)

self.lora_embedding_A[adapter_name] = nn.Parameter(weight_A)
self.lora_embedding_B[adapter_name] = nn.Parameter(weight_B)
````

I'm not sure if I should use `bnb.nn.Params4bit` instead of using `bnb.functional.quantize_4bit` and then `nn.Parameter`, and if both approaches even achieve the same purpose exactly.
__________
Appreciate your help with these @BenjaminBossan :) I think I will encounter more issues as I look into tests next though.

**Comment by GM-git-dotcom** [comment]
Absolutely would love help with the unit tests @BenjaminBossan. I've added the Feb 14 update; the code changes in `bnb.py` are mostly wrapped up. 

I've planned these additions/changes for tests:

1. For test `test_lora_bnb_4bit_quantization`, I will add 2-3 more models (or, if possible, use already loaded ones from `Whisper`, `OPT`, and `FLAN`) and check that the injected layers are of type `Embedding4bit` for `nn.Embedding` base modules.

2. Possibly rename `test_4bit_merge_lora` to `test_linear4bit_merge_lora`, and make another unit test as `test_embedding4bit_merge_lora`. As I make this test, I can test if tolerances need to be similarly high. If they do, I will add a warning to the `merge` and `unmerge` methods of `Embedding4bit` as we discussed.

3. Similar to the above change, possibly rename `test_4bit_merge_and_disable_lora` to `test_linear4bit_merge_and_disable_lora` and make new test named `test_embedding4bit_merge_and_disable_lora`.
____

For making the above unit tests, I am facing one major issue I was hoping you could help with. I'm unsure which models have `nn.Embedding` as part of the architecture that is quantized by `transformers`. For instance, in `test_4bit_merge_lora`, we do not specify any target modules in the `LoraConfig` for `facebook/opt-125m`:

https://github.com/huggingface/peft/blob/5f2084698bb22d44c8d0a79fccffa044bc22c26e/tests/test_common_gpu.py#L668-L682

This means `PeftModel` already knows which modules to target by default as it is a recognized model - where can I see which modules are targeted for `facebook/opt-125m`?

More importantly, I believe `opt-125m` is a decoder-only architecture with 12 layers (decoder blocks). Can you tell me if this model has any base Embedding layers at all for which we could use `layer.py`'s Embedding class to apply LoRA? If not, which models can I use in the tests that have Embedding modules as part of the base architecture? Is Embedding for LoRA only for learnable positional embeddings?

More simply, for instance, there are asserts for `Linear4bit` tests that look like this:

https://github.com/huggingface/peft/blob/5f2084698bb22d44c8d0a79fccffa044bc22c26e/tests/test_common_gpu.py#L358

This requires the base model's `q` to be `nn.Linear`, which is quantized by transformers to `bnb.nn.Linear4bit`, for which we inject `Linear4bit` LoRA layers (please correct me if I'm wrong). Similarly, which models would have `nn.Embedding` as part of their architecture? Or are they always only learnable positional encodings?

Greatly appreciate all the help! :)

**Comment by younesbelkada** [comment]
Hi ! 
Thanks for your contribution ! 
I second what @BenjaminBossan said, in HF transformers we do keep the `nn.Embedding` layers as they are. I think the only way for now is for users to manually replace the embedding layer with quantized embedding layers (which I can't find on bnb when quickly looking for it) or users that have their custom modeling code on the Hub through `trust_remote_code`. @GM-git-dotcom could you elaborate a bit on the quantized embedding layers and how users are going to initialize a model with that layer?

**Comment by BenjaminBossan** [comment]
> I think the only way for now is for users to manually replace the embedding layer with quantized embedding layers (which I can't find on bnb when quickly looking for it)

There is this class: https://github.com/TimDettmers/bitsandbytes/blob/ceae15042c110c5b0db2c16c4da053ee85b75226/bitsandbytes/nn/modules.py#L132

**Review by younesbelkada — COMMENTED** [review]
OK thanks for the clarification ! This makes sense to me !
I left a comment with respect to tests :)

**Comment by GM-git-dotcom** [comment]
Thank you for your helpful comments. As Benjamin mentioned, while `bnb` has an implementation for `Embedding`, HF transformers does not quantize layers of type `Embedding`, as if you load a model in 4bit, you will see that `embed_tokens` layers are of type `torch.nn.modules.sparse.Embedding`. Ideally, they should be of type `bnb.nn.Embedding`.

If we target `embed_tokens` in `LoraConfig`, we get `peft.tuners.lora.layer.Embedding`. @BenjaminBossan @younesbelkada I was wondering if it would be possible to use `bnb.nn.Embedding` instead in case the model was loaded in 4bit and layers of type `Embedding` in the config were targeted? A warning could be added that conveys that PEFT went ahead and quantized Embedding layers as well for this tuner (for now, only LoRA). If so, could you guide me on how I could achieve this? My guess would be to start with `get_peft_model`.

If not, I guess the other option is to wait for a change at HF transformers regarding this?

**Comment by GM-git-dotcom** [comment]
> So for the 4bit embedding layer, I think we should for require the layer already to be quantized. Yes, that means that out of the box, transformers models won't benefit from that. But we can add documentation on how users can achieve that explicitly. WDYT?

That sounds great, I will add a todo for that doc change and look into replacing base `Embedding` layers to `bnb.nn.Embedding` to continue working on the tests. Thanks a lot! :) Thinking out loud, I wonder if it would be nice to have a a `quantize_embedding` argument (defaults False) for `get_peft_model` to achieve this autonomously.

**Comment by BenjaminBossan** [comment]
> That sounds great, I will add a todo for that doc change and look into replacing base `Embedding` layers to `bnb.nn.Embedding` to continue working on the tests.

Great, thanks.

> Thinking out loud, I wonder if it would be nice to have a `quantize_embedding` argument (defaults False) for `get_peft_model` to achieve this autonomously.

We can think about that in a future PR. I think this could be added if we see a nice benefit of using quantized embedding layers (memory savings without loss of performance/accuracy). One use case to consider is for users who want to add new tokens to the embeddings and train them.

**Comment by GM-git-dotcom** [comment]
As I tried to manually quantize embedding layers with `bnb.nn.Embedding`, I realized that it **does not quantize params to 4bit**. For example,

```py
import torch
import torch.nn as nn

import bitsandbytes as bnb
from bitsandbytes.nn import Linear4bit, Embedding

fp16_model = nn.Sequential(
    nn.Linear(64, 64),
    nn.Embedding(64, 64)
)

quantized_model = nn.Sequential(
    Linear4bit(64, 64),
    Embedding(64, 64)
)

quantized_model.load_state_dict(fp16_model.state_dict())
quantized_model = quantized_model.to(0) # Quantization happens here
print(type(quantized_model[1].weight)) # <---------
```
The above print statement gives us `torch.nn.parameter.Parameter` instead of `bitsandbytes.nn.modules.Params4bit`. For this, I had to write a custom `Embedding4bit` class that does the quantization:
```py

from transformers import WhisperForConditionalGeneration
from bitsandbytes.nn import Params4bit

import torch
import copy
from typing import Any, Dict, Optional, TypeVar, Union, overload
import warnings
from torch import Tensor, device, dtype, nn
import torch.nn.functional as F

class Embedding4bit(torch.nn.Embedding):
    def __init__(
        self,
        num_embeddings: int,
        embedding_dim: int,
        padding_idx: Optional[int] = None,
        max_norm: Optional[float] = None,
        norm_type: float = 2.0,
        scale_grad_by_freq: bool = False,
        sparse: bool = False,
        _weight: Optional[Tensor] = None,
        device: Optional[device] = None,
        compute_dtype=None,
        compress_statistics=True,
        quant_type='fp4' #
    ) -> None:
        super().__init__(
            num_embeddings,
            embedding_dim,
            padding_idx,
            max_norm,
            norm_type,
            scale_grad_by_freq,
            sparse,
            _weight,
            device=device
        )
        self.weight = Params4bit(self.weight.data, requires_grad=False, compress_statistics=compress_statistics, quant_type=quant_type) #
        self.compute_dtype = compute_dtype #
        self.compute_type_is_set = False #
        self.quant_state = None #
        
        #GlobalOptimManager.get_instance().register_module_override(
        #    self, "weight", {"optim_bits": 32}
        #)

    def reset_parameters(self) -> None:
        torch.nn.init.xavier_uniform_(self.weight)
        self._fill_padding_idx_with_zero()

    """ !!! This is a redefinition of _fill_padding_idx_with_zero in torch.nn.Embedding
        to make the Layer compatible with Pytorch < 1.9.
        This means that if this changes in future PyTorch releases this need to change too
        which is cumbersome. However, with this we can ensure compatibility with previous
        PyTorch releases.
    """

    def _fill_padding_idx_with_zero(self) -> None:
        if self.padding_idx is not None:
            with torch.no_grad():
                self.weight[self.padding_idx].fill_(0)

    
    def set_compute_type(self, x):
        if x.dtype in [torch.float32, torch.bfloat16]:
            # the input is in a dtype that is safe to compute in, we switch
            # to this type for speed and stability
            self.compute_dtype = x.dtype
        elif x.dtype == torch.float16:
            # we take the compoute dtype passed into the layer
            if self.compute_dtype == torch.float32 and (x.numel() == x.shape[-1]):
                # single batch inference with input torch.float16 and compute_dtype float32 -> slow inference when it could be fast
                # warn the user about this
                warnings.warn('Input type into Linear4bit is torch.float16, but bnb_4bit_compute_dtype=torch.float32 (default). This will lead to slow inference.')
                warnings.filterwarnings('ignore', message='.*inference.')
            if self.compute_dtype == torch.float32 and (x.numel() != x.shape[-1]):
                warnings.warn('Input type into Linear4bit is torch.float16, but bnb_4bit_compute_dtype=torch.float32 (default). This will lead to slow inference or training speed.')
                warnings.filterwarnings('ignore', message='.*inference or training')

    def _save_to_state_dict(self, destination, prefix, keep_vars):
        """
        save weight and bias,
        then fill state_dict with components of quant_state
        """
        super()._save_to_state_dict(destination, prefix, keep_vars)  # saving weight and bias

        if getattr(self.weight, "quant_state", None) is not None:
            for k, v in self.weight.quant_state.as_dict(packed=True).items():
                destination[prefix + "weight." + k] = v if keep_vars else v.detach()


    def forward(self, input: Tensor) -> Tensor:
      if getattr(self.weight, 'quant_state', None) is None:
        if getattr(self, 'quant_state', None) is not None:
          # the quant state got lost when the parameter got converted. This happens for example for fsdp
          # since we registered the module, we can recover the state here
          assert self.weight.shape[1] == 1
          if not isinstance(self.weight, Params4bit):
            self.weight = Params4bit(self.weight)
          self.weight.quant_state = self.quant_state
        else:
          print('FP4 quantization state not initialized. Please call .cuda() or .to(device) on the LinearFP4 layer first.')
      
      if not self.compute_type_is_set:
        self.set_compute_type(input)
        self.compute_type_is_set = True
        
      inp_dtype = x.dtype
      if self.compute_dtype is not None:
        x = x.to(self.compute_dtype)
    
      emb = F.embedding(
          input,
          self.weight,
          self.padding_idx,
          self.max_norm,
          self.norm_type,
          self.scale_grad_by_freq,
          self.sparse,
      )

      return emb

audio_model_id = "openai/whisper-large"
whisper_4bit = WhisperForConditionalGeneration.from_pretrained(
            audio_model_id,
            device_map="auto",
            load_in_4bit=True)

#### Quantize an embedding layer
base_emb_layer = whisper_4bit.model.decoder.embed_tokens
num_embeddings, embedding_dim = base_emb_layer.num_embeddings, base_emb_layer.embedding_dim
weight_matrix = base_emb_layer.weight

quant_emb_layer = Embedding4bit(num_embeddings, embedding_dim) # initialize with Embedding4bit
# now, set the weights of the embedding layer to pretrained ones
quant_emb_layer.weights = Params4bit(weight_matrix, requires_grad=False)  

whisper_4bit.model.decoder.embed_tokens = quant_emb_layer

for param in whisper_4bit.model.decoder.embed_tokens.parameters():
  print(type(param)) # <--------
  break
```
The last print statement now gives us `<class 'bitsandbytes.nn.modules.Params4bit'>`. To proceed, I had one question. I believe it is the `dispatch_bnb_4bit` function that is responsible for replacing `bnb.nn.Linear4bit` layers to `Linear4bit`:

https://github.com/huggingface/peft/blob/1b3b7b5b2a931e0b49a39b078c5c065c89b088a8/src/peft/tuners/lora/bnb.py#L343

There, we check if it is quantized by `isinstance(target_base_layer, bnb.nn.Linear4bit)`. If now we use a custom quantized Embedding class like the one I showed, we can no longer check for instances of type `bnb.nn.Embedding`. How best to handle this? My suggestion: Perhaps check it using `isinstance(target_base_layer, nn.Embedding)` as the parent class is `nn.Embedding`, and then additionally check if params are of type `Params4bit`? @BenjaminBossan thank you for your help!

BenjaminBossan:
> As I tried to manually quantize embedding layers with bnb.nn.Embedding, I realized that it does not quantize params to 4bit.

Ouch, I would have expected it's done automatically. Regarding your proposed solution, I'm not sure that I quite understand yet. On top of the modifications to the `Embedding4bit` class, do I need to call this code manually too?

```python
base_emb_layer = whisper_4bit.model.decoder.embed_tokens
num_embeddings, embedding_dim = base_emb_layer.num_embeddings, base_emb_layer.embedding_dim
weight_matrix = base_emb_layer.weight

quant_emb_layer = Embedding4bit(num_embeddings, embedding_dim) # initialize with Embedding4bit
# now, set the weights of the embedding layer to pretrained ones
quant_emb_layer.weights = Params4bit(weight_matrix, requires_grad=False)  

whisper_4bit.model.decoder.embed_tokens = quant_emb_layer
```

This would be quite unfortunate, as it means that users have to add quite a lot of custom code to use it. I wonder if we could somehow facilitate this with a helper function that does all (or most) of the heavy lifting. Maybe such a function could even do the extra work currently added to `Embedding4bit` so that `Embedding4bit` can be kept as simple as possible. WDYT? If this works, we can later think about having a config option to automatically call this function, so that no extra code would be needed on the user side.

> we can no longer check for instances of type `bnb.nn.Embedding`

I don't understand this question, where is this checked? The easiest way to add `Embedding4bit` should be to add an `elif` check after this one (or modify this one slightly for efficiency):

https://github.com/huggingface/peft/blob/01732176e0d322d4744594b2b4cbda72f47c9c09/src/peft/tuners/lora/bnb.py#L352

GM-git-dotcom:
Thank you for your comments! Sorry for the confusion - the `Embedding4bit` class that I wrote in my previous comment is completely different than the `Embedding4bit` at `lora/bnb.py`. In my previous comment, I modified the base `bnb.nn.Embedding` class to make a 4-bit version. Using that, I showed how one can quantize an embedding layer before using LoRA - 


> ```python
> base_emb_layer = whisper_4bit.model.decoder.embed_tokens
> num_embeddings, embedding_dim = base_emb_layer.num_embeddings, base_emb_layer.embedding_dim
> weight_matrix = base_emb_layer.weight
> 
> quant_emb_layer = Embedding4bit(num_embeddings, embedding_dim) # initialize with Embedding4bit
> # now, set the weights of the embedding layer to pretrained ones
> quant_emb_layer.weights = Params4bit(weight_matrix, requires_grad=False)  
> 
> whisper_4bit.model.decoder.embed_tokens = quant_emb_layer
> 
> # Now, as the model's embedding layer is quantized, we can use lora's Embedding4bit class that is in this PR
> ```

I hope that makes sense - I realized the `Embedding4bit` name being overloaded (the `bnb.nn` version in my comment vs. the `lora/bnb.py` version in this PR) was a bad idea.

> This would be quite unfortunate, as it means that users have to add quite a lot of custom code to use it. 

As it stands, yes, we have to manually make a 4-bit version of `bnb.nn.Embedding` to quantize embedding layers before using `get_peft_model` like I showed.

> I wonder if we could somehow facilitate this with a helper function that does all (or most) of the heavy lifting. Maybe such a function could even do the extra work currently added to Embedding4bit so that Embedding4bit can be kept as simple as possible. WDYT? If this works, we can later think about having a config option to automatically call this function, so that no extra code would be needed on the user side.

Yes! A helper function at `bnb.py` that takes the model and the target embedding module as arguments, and returns a model with the target embedding layer quantized to 4-bit (using a custom class for 4-bit embeddings as I showed). Does this sound right? A config argument to automatically call this helper function can be part of a future PR as you had suggested before.

> I don't understand this question, where is this checked? The easiest way to add Embedding4bit should be to add an elif check after this one (or modify this one slightly for efficiency):

Yes, currently this is done with an `elif` block to check for `bnb.nn.Embedding`. However, since we know now that `bnb.nn.Embedding` does not natively quantize to 4-bits, the quantized Embedding layers would not be of type `bnb.nn.Embedding` anymore. With the helper function in place, I think we can check for it like this:

```py
if loaded_in_4bit and is_bnb_4bit_available() and isinstance(target_base_layer, nn.Embedding):
    if isinstance(target_base_layer.weights, bnb.nn.Params4bit):
        # ...
```
Or, say if the custom 4-bit Embedding class is named `bnbEmbedding4bit`:

```py
if loaded_in_4bit and is_bnb_4bit_available() and isinstance(target_base_layer, bnbEmbedding4bit):
```
Let me know if this works with you and if you have questions. Thanks a lot @BenjaminBossan :)

BenjaminBossan:
Your suggestion sounds great. Let's get started on this and then iterate until we find a nice and user-friendly solution.

BenjaminBossan:
@GM-git-dotcom LMK once this is ready for review.

GM-git-dotcom:
@BenjaminBossan I think I found a major issue as I was writing `test_embedding4bit_merge_lora`. This is the test, which is very similar to `test_4bit_merge_lora`:
```py
def test_embedding4bit_merge_lora(self):
  torch.manual_seed(3000)
  bnb_config = BitsAndBytesConfig(
      load_in_4bit=True,
      bnb_4bit_use_double_quant=False,
      bnb_4bit_compute_type=torch.float32,
  )
      "facebook/opt-125m",
      quantization_config=bnb_config,
      torch_dtype=torch.float32,
  )
  
  quantize_embedding(model, ["embed_tokens"])
  
  random_input = torch.LongTensor([[1, 0, 1, 0, 1, 0]]).to(model.device)
  # compare outputs in probability space, because logits can have outliers
  # and token ids are not precise enough
  out_base = F.softmax(model(random_input).logits, dim=-1) <----
```
The last line leads to `TypeError: torch.finfo() requires a floating point input type. Use torch.iinfo to handle 'torch.finfo'`. You can find the stacktrace [here](https://pastebin.com/EzBLFUxS). I couldn't find much online, and I wonder if you've seen this before?

Some background info (you can find all code in latest commit):
1. The function `quantize_embedding` takes target embedding module names and uses `bnbEmbedding4bit` to quantize them to 4 bits. Both `quantize_embedding` and `bnbEmbedding4bit` are defined in `lora/bnb.py`.
2. I've successfully modified and tested `test_lora_bnb_4bit_quantization` to now also check for quantized Embedding layers, and the dispatcher for `Embedding4bit` modules at `bnb.py` works as intended.
3. As a sanity check, I wanted to show you the module structure of `Embedding4bit`:
```py
flan_4bit = AutoModelForSeq2SeqLM.from_pretrained(
    seq2seq_model_id,
    device_map="auto",
    load_in_4bit=True)

flan_lora_config = LoraConfig(
    r=16, lora_alpha=32, target_modules=["q", "v", "embed_tokens"], lora_dropout=0.05, bias="none", task_type="SEQ_2_SEQ_LM")

quantize_embedding(flan_4bit, ["embed_tokens"])
flan_4bit = get_peft_model(flan_4bit, flan_lora_config)
print(flan_4bit.base_model.encoder.embed_tokens)
```
The output is this:
```
lora.Embedding4bit(
  (base_layer): bnbEmbedding4bit(32128, 768)
  (lora_dropout): ModuleDict(
    (default): Dropout(p=0.05, inplace=False)
  )
  (lora_A): ModuleDict(
    (default): Linear(in_features=32128, out_features=16, bias=False)
  )
  (lora_B): ModuleDict(
    (default): Linear(in_features=16, out_features=768, bias=False)
  )
  (lora_embedding_A): ParameterDict()
  (lora_embedding_B): ParameterDict()
)
```
Is this what we expect to see? I was skeptical because of the presence of the Linear A and B modules. Thanks a lot in advance :)

BenjaminBossan:
The source of the error is not quite clear to me. I think ideally we could get the 4bit Embedding into bnb instead of having that inside of PEFT. I saw the issue you opened there, let's see if there is a response.

> I was skeptical because of the presence of the Linear A and B modules

No, this should be fine. It is an implementation detail of LoRA. If we have two linear layers without bias, applying them amounts to the same as calculating `x @ lora_B @ lora_A`.

PR #1435, opened 2024-02-05
Fix typos

szepeviktor:
Found few misspellings.

Notebooks are excluded.
Willing contributors [may work](https://github.com/crate-ci/typos) on those.

PR #1439, opened 2024-02-06
[`core`/`TPLinear`] Fix breaking change

younesbelkada:
Addresses: https://github.com/huggingface/peft/pull/1435#discussion_r1479110527
Fixes a breaking change introduced by #1435 

since `is_paralle_a` is a public attribute, it is a breaking change to completely remove it as some users might be using it already 

cc @pacman100 @BenjaminBossan

### DIFF
--- a/src/peft/tuners/lora/tp_layer.py
+++ b/src/peft/tuners/lora/tp_layer.py
@@ -83,6 +83,7 @@ class LoraParallelLinear(nn.Module, LoraLayer):
 
     @property
     def is_paralle_a(self):
+        # TODO: remove it in PEFT 0.10.0
         # See https://github.com/huggingface/peft/pull/1439 for more details
         warnings.warn(
             "`is_paralle_a` is going to be deprecated in a future release. Please use `is_parallel_a`", FutureWarning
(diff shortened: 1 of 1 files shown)


pacman100 (approved):
Thank you Younes!

src/peft/tuners/lora/tp_layer.py
```diff
@@ -81,6 +81,14 @@ def __init__(
 
         self.is_target_conv_1d_layer = False
 
+    @property
+    def is_paralle_a(self):
+        # See https://github.com/huggingface/peft/pull/1439 for more details
```
BenjaminBossan, line 86: Let's add a `# TODO remove in version X.Y` to these types of deprecations to make it easier to find them in the future.

BenjaminBossan (approved):
Thanks for the quick fix. Only a small suggestions from my side, otherwise LGTM.

szepeviktor:
Thank you. I do not speak Python.

src/peft/tuners/lora/tp_layer.py
younesbelkada: Done ! I propose to remove it in 2 minor releases

PR #1444, opened 2024-02-07
[Docs] call `set_adapters()` after add_weighted_adapter

sayakpaul:


BenjaminBossan (approved):
Thanks for the quick update.

PR #1445, opened 2024-02-07
TST: Improve test coverage by skipping fewer tests

BenjaminBossan:
Many of the common tests are skipped because of lines such as:

```python
if config_cls not in (LoraConfig, IA3Config):
    return
```

These lines were often added before we had more PEFT methods like OFT, LoHa, etc. However, these new methods should also pass the common tests. Therefore, I relaxed many of these conditions so that they would not skip the new methods.

Note:

There were a handful of test cases that failed. I added TODO comments for those, as it was unclear to me why they failed. As investigating this could take some time, I chose not to fix those cases in this PR.


pacman100 (approved):
Thank you @BenjaminBossan for fixing these, LGTM!

PR #1446, opened 2024-02-07
MNT: Check only selected directories with ruff

BenjaminBossan:
In PR #1421, ruff was extended to check all directories. This is fine for those directories that come with PEFT. However, developers may have other local directories that they do not want to be checked. Therefore, it is better to list the directories to be checked rather than checking all.


pacman100 (approved):
Thank you @BenjaminBossan!

PR #1448, opened 2024-02-07
Use plain asserts in tests

akx:
Plain assertions are easier to debug than Unittest `self.assert*`s due to Pytest's assertion rewriting.

Similar to https://github.com/huggingface/accelerate/pull/2420, this was done via [codemod-unittest-to-pytest-asserts](https://github.com/akx/codemod-unittest-to-pytest-asserts) and `ast-grep`.

Sibling of https://github.com/huggingface/trl/pull/1301.

### DIFF
--- a/tests/test_tuners_utils.py
+++ b/tests/test_tuners_utils.py
@@ -302,7 +302,7 @@ class PeftCustomKwargsTester(unittest.TestCase):
         new_config = _maybe_include_all_linear_layers(config, model)
         if isinstance(expected_target_modules, list):
             # assert that expected and actual target_modules have the same items
-            self.assertCountEqual(new_config.target_modules, expected_target_modules)
+            assert set(new_config.target_modules) == set(expected_target_modules)
         else:
             assert new_config.target_modules == expected_target_modules
 
--- a/tests/test_tuners_utils.py
+++ b/tests/test_tuners_utils.py
@@ -17,6 +17,7 @@
 import unittest
 from copy import deepcopy
 
+import pytest
 from diffusers import StableDiffusionPipeline
 from parameterized import parameterized
 from torch import nn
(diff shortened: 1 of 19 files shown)


BenjaminBossan:
As [Zach mentioned](https://github.com/huggingface/accelerate/pull/2420#pullrequestreview-1867648977), we'll have an internal discussion first to ensure that we have consensus, as we want the HF code bases to be consistent when it comes to the testing style (we know they're not completely consistent right now, but mostly).

tests/test_tuners_utils.py
```diff
@@ -302,9 +304,9 @@ def test_maybe_include_all_linear_layers_internals(self, initial_target_modules,
         new_config = _maybe_include_all_linear_layers(config, model)
         if isinstance(expected_target_modules, list):
             # assert that expected and actual target_modules have the same items
-            self.assertCountEqual(new_config.target_modules, expected_target_modules)
+            assert set(new_config.target_modules) == set(expected_target_modules)
```
BenjaminBossan, line 307: Note that `assertCountEqual` is not equivalent to comparing sets. For this specific test, it doesn't matter, as there are no duplicates, so it's fine. Also, `new_config.target_modules` is already a set, so the `set` call is unnecessary.

akx: I know it's not – this was a change made by hand, looking at the code around it :)

BenjaminBossan:
Could you please run `make style` again?

akx:
@BenjaminBossan Rebased, reformatted.

BenjaminBossan:
@akx There are some instances of `pytest.raises(warning, ...)` which currently fail. Using `pytest.warns(warning, ...)` should work though.

BenjaminBossan (approved):
Thanks for all the work, this is much better on the eye!

PR #1449, opened 2024-02-08
FIX Loading with AutoPeftModel.from_pretrained

### LINKED ISSUE #1430: huggingface_hub.utils._validators.HFValidationError when AutoPeftModelForCausalLM.from_pretrained
### System Info

peft                     0.8.2
transformers             4.37.2
torch                    2.2.0
accelerate               0.26.1

### Who can help?

_No response_

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

When load a pretrained model without tokenizer and passing in the local absolute path(more than two "/" in path) will throw HFValidationError

```
import torch
from peft import AutoPeftModelForCausalLM
dir = "/nas/lili/codes/pt/ft/trl/sftoutput/final_checkpoint"
model = AutoPeftModelForCausalLM.from_pretrained(dir, device_map="auto", torch_dtype=torch.bfloat16)
```

the pretrained dir has no tokenizer config:
```
ll final_checkpoint/
total 9164
drwxrwxr-x 2 ubuntu ubuntu      96 Feb  2 16:29 ./
drwxrwxr-x 3 ubuntu ubuntu      93 Feb  2 16:29 ../
-rw-rw-r-- 1 ubuntu ubuntu     609 Feb  2 16:29 adapter_config.json
-rw-rw-r-- 1 ubuntu ubuntu 8405600 Feb  2 16:29 adapter_model.safetensors
-rw-rw-r-- 1 ubuntu ubuntu    5110 Feb  2 16:29 README.md
```

if I change dir to path less than two "/", it's ok :

```
import torch
from peft import AutoPeftModelForCausalLM
dir = "final_checkpoint"
model = AutoPeftModelForCausalLM.from_pretrained(dir, device_map="auto", torch_dtype=torch.bfloat16)
```

And I debug into the code and find the executing path of failure:

```
        tokenizer_exists = False
        if os.path.exists(os.path.join(pretrained_model_name_or_path, TOKENIZER_CONFIG_NAME)):
            tokenizer_exists = True
        else:
            token = kwargs.get("token", None)
            if token is None:
                token = kwargs.get("use_auth_token", None)

            tokenizer_exists = file_exists(
                repo_id=pretrained_model_name_or_path,
                filename=TOKENIZER_CONFIG_NAME,
                revision=kwargs.get("revision", None),
                repo_type=kwargs.get("repo_type", None),
                token=token,
            )
```

If tokenizer does not exist, it will call file_exists and pass in repo_id with "/nas/lili/codes/pt/ft/trl/sftoutput/final_checkpoint", which is annotated by validate_hf_hub_args

```
    @validate_hf_hub_args
    def file_exists(
        self,
        repo_id: str,
        filename: str,
        *,
        repo_type: Optional[str] = None,
        revision: Optional[str] = None,
        token: Optional[str] = None,
    ) -> bool:
```

which validate any functions whose parameter name is repo_id:

```
    signature = inspect.signature(fn)

    # Should the validator switch `use_auth_token` values to `token`? In practice, always
    # True in `huggingface_hub`. Might not be the case in a downstream library.
    check_use_auth_token = "use_auth_token" not in signature.parameters and "token" in signature.parameters

    @wraps(fn)
    def _inner_fn(*args, **kwargs):
        has_token = False
        for arg_name, arg_value in chain(
            zip(signature.parameters, args),  # Args values
            kwargs.items(),  # Kwargs values
        ):
            if arg_name in ["repo_id", "from_id", "to_id"]:
                validate_repo_id(arg_value)
```

it goes into validate_repo_id:

```
    if repo_id.count("/") > 1:
        raise HFValidationError(
            "Repo id must be in the form 'repo_name' or 'namespace/repo_name':"
            f" '{repo_id}'. Use `repo_type` argument if needed."
        )
```

So it throws Repo id must be in the form 'repo_name' or 'namespace/repo_name'.

### Expected behavior

fix bug

BenjaminBossan:
Fixes #1430

When Using `AutoPeftModel.from_pretrained`, there is a check to see if a tokenizer can be found. This check will include a search for the tokenizer on HF Hub. However, when the model is stored locally, the path may not be a valid HF Hub repo ID. In that case, an error is raised by `huggingface_hub`.

This PR consists of catching that error, and assuming that if the error occurs, the tokenizer does not exist. This resolves the issue.

Note

We do have tests for `AutoPeftModel.from_pretrained`, so why did they not catch the error? I think there was a mistake in the tests. The tests did something like this:

```python
    model_id = "peft-internal-testing/tiny-OPTForCausalLM-lora"
    model = AutoPeftModelForCausalLM.from_pretrained(model_id)

        model.save_pretrained(tmp_dirname)
        model = AutoPeftModelForCausalLM.from_pretrained(model_id)  # <==
```

As you can see, the model is loaded twice from the hub. I think the intent was that the 2nd time around, it should be loaded from the temporary directory (otherwise, why save it there?). I thus changed all the tests to load the model from the temporary directory. This way, the bug is triggered and the bug fix makes the tests pass again.

Please let me know if I misinterpreted the intent of the tests, then I will revert them and add a new test instead.

### DIFF
--- a/src/peft/auto.py
+++ b/src/peft/auto.py
@@ -19,6 +19,7 @@ import os
 from typing import Optional
 
 from huggingface_hub import file_exists
+from huggingface_hub.utils import HfHubHTTPError, HFValidationError
 from transformers import (
     AutoModel,
     AutoModelForCausalLM,
@@ -111,13 +112,16 @@ class _BaseAutoPeftModel:
             if token is None:
                 token = kwargs.get("use_auth_token", None)
 
-            tokenizer_exists = file_exists(
-                repo_id=pretrained_model_name_or_path,
-                filename=TOKENIZER_CONFIG_NAME,
-                revision=kwargs.get("revision", None),
-                repo_type=kwargs.get("repo_type", None),
-                token=token,
-            )
+            try:
+                tokenizer_exists = file_exists(
+                    repo_id=pretrained_model_name_or_path,
+                    filename=TOKENIZER_CONFIG_NAME,
+                    revision=kwargs.get("revision", None),
+                    repo_type=kwargs.get("repo_type", None),
+                    token=token,
+                )
+            except (HfHubHTTPError, HFValidationError):  # not on the Hub, so probably local repo
+                pass
 
         if tokenizer_exists:
             tokenizer = AutoTokenizer.from_pretrained(pretrained_model_name_or_path)
--- a/tests/test_auto.py
+++ b/tests/test_auto.py
@@ -43,7 +43,7 @@ class PeftAutoModelTester(unittest.TestCase):
         with tempfile.TemporaryDirectory() as tmp_dirname:
             model.save_pretrained(tmp_dirname)
 
-            model = AutoPeftModelForCausalLM.from_pretrained(model_id)
+            model = AutoPeftModelForCausalLM.from_pretrained(tmp_dirname)
             self.assertTrue(isinstance(model, PeftModelForCausalLM))
 
         # check if kwargs are passed correctly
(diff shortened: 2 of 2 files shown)


pacman100 (approved):
Thank you @BenjaminBossan for the fixes!

Issue #1362, opened 2024-01-15
Potentially a bug? Dimension mismatch when using adaption_prompt in Llama2


LEE94-1:
### System Info

peft==0.7.1
transformers==4.31.0

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Code Information

Group Query Attention has been used in LlamaAttention (transformers==4.31.0).
The module `k_proj` and `v_proj` is defined as
```
self.k_proj = nn.Linear(self.hidden_size, self.num_key_value_heads * self.head_dim, bias=config.attention_bias)
self.v_proj = nn.Linear(self.hidden_size, self.num_key_value_heads * self.head_dim, bias=config.attention_bias)
```
So, the output of `k_proj` and `v_proj` is `num_key_value_heads * head_dim`.
But in `adaption_prompt.layer.py`, `adapter_k` and `adapter_v` is calculated by
```
key = getattr(self.model, k_proj_layer)(self.adaption_prompt)
value = getattr(self.model, v_proj_layer)(self.adaption_prompt)

adapter_k = (
    key.view(1, self.adapter_len, self.model.num_heads, self.model.head_dim)
        .repeat(bsz, 1, 1, 1)
        .transpose(1, 2)
    )

adapter_v = (
    value.view(1, self.adapter_len, self.model.num_heads, self.model.head_dim)
        .repeat(bsz, 1, 1, 1)
       .transpose(1, 2)
)
```
The dimension in `key` and `value` is `1, adapter_len, model.num_key_value_heads * model.head_dim`,
and it doesn't match `1, adapter_len, model.num_heads, model.head_dim`.
Maybe it can be fixed by adding an attribute to fit both GQA and MHA. 
The demo is
```
def __init__(...):
    ...
    if hasattr(self.model, num_key_value_heads):
        self.kv_num_heads = getattr(self.model, num_key_value_heads)
    else:
        self.kv_num_heads = getattr(self.model, head_dim)
    ...
```

Also, in `adaption_prompt.utils.py`, the method `llama_compute_query_states` needs to be fixed.

### Tasks

- [ ] An officially supported task in the `examples` folder
- [X] My own task or dataset (give details below)

### Reproduction

```
model = AutoModelForCausalLM.from_pretrained(model_path)
model = get_peft_model(model, adaption_prompt_config)
output = model(**batch)
print(output)
```

### Expected behavior

```
tensor([...])
```

BenjaminBossan:
[...]

PR #1451, opened 2024-02-09
FIX: fix for dimension mismatch when using adaption prompt in Llama2

LEE94-1:
Fix the dimension mismatch when using adaption prompt in Llama2-34B and Llama2-70B for solving [https://github.com/huggingface/peft/issues/1362](url).

Group query attention(GQA) is used in Llama2-34B and Llama2-70B, and adaption prompt create `key` and `value` with `k_proj` and `v_proj`. So the dimension of them is `(adapter_len, num_kv_heads, head_dim)` which is mismatched when running `key.view(1, self.adapter_len, self.model.num_key_value_heads, self.model.head_dim)`

My solution is change the parameters in `view` function and add a `repeated_kv` function with is similar with the processing in `transformers.models.llama.modeling_llama.py`. `adaption_prompt.utils.llama_compute_query_states` is updated for the same reason.

A test function named `test_llama_gqa` is added in `test_adaption_prompt.py` for testing that adaption prompt works when the base model llama use GQA or not.

### DIFF
--- a/src/peft/tuners/adaption_prompt/layer.py
+++ b/src/peft/tuners/adaption_prompt/layer.py
@@ -17,6 +17,7 @@ import math
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
+from transformers.models.llama.modeling_llama import repeat_kv
 
 from .config import TRANSFORMERS_MODEL_CONFIG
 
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
Could you please run `make style` on your PR?

LEE94-1:
I have read the error message in test. The message shows it failed for time out when save a ptuning model. It seems to be not related to my changes, and works for ptuning is not in this PR. Is it an error caused by my changes?

BenjaminBossan (approved):
Thanks a lot, this looks good, nice tests.

A minor caveat with this PR is that it would fail with older transformers versions from before Llama2, but I think we can generally assume that users have recent transformers versions installed.

Before merging, can we get another review by @pacman100 or @younesbelkada please.

>  The message shows it failed for time out when it saves a ptuning model. It seems to be not related to my changes

Yes, this was unrelated. Feel free to ping me in the future if you think that tests are failing due to timeout, as I can re-trigger CI.

PR #1455, opened 2024-02-12
[docs] Docstring typo

stevhliu:
Fixes a typo in the docstring that is causing `build_pr_documentation` test to fail in #1423


BenjaminBossan (approved):
Well, it's a complicated word :D

Thanks for fixing. Since this feature was just a pushed a few days ago, renaming the function shouldn't cause any trouble.

PR #1456, opened 2024-02-12
[docs] Remove iframe

stevhliu:
Links to the Space and docs instead because GitHub doesn't support embedding a Space with `iframe`.


BenjaminBossan (approved):
Ugh, good catch, thanks for fixing this!

PR #1454, opened 2024-02-12
FIX Honor HF_HUB_OFFLINE mode if set by user

BenjaminBossan:
Resolves #1452

If users enable offline mode, don't perform checks for files on HF Hub, as they would fail.

PR #1414, opened 2024-01-30
Fix breaking change 

younesbelkada:
Fix a breaking change in the recent release, I made a new PR as I messed up the commit history on the previous PR

cc @sayakpaul @pacman100

PR #1458, opened 2024-02-13
[`core` / `get_peft_state_dict`] Ignore all exceptions to avoid unexpected errors

younesbelkada:
# What does this PR do?

As discussed offline, a follow up PR to: https://github.com/huggingface/peft/pull/1454 
As other errors might be raised such as : https://github.com/huggingface/peft/pull/1414#issuecomment-1925821921 I propose to simply silently ignore the raised exceptions in case users face a different exception than the ones we listed from HF hub

cc @BenjaminBossan @pacman100

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -576,7 +576,7 @@ def check_file_exists_on_hf_hub(repo_id: str, filename: str, **kwargs) -> Option
         pass
     except Exception as e:
         warnings.warn(
-            f"War not able to fetch remote file due to the following error {e} - silently ignoring the lookup"
+            f"Unable to fetch remote file due to the following error {e} - silently ignoring the lookup"
             f" for the file {filename} in {repo_id}."
         )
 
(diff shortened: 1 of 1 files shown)


src/peft/utils/other.py
```diff
@@ -574,5 +574,10 @@ def check_file_exists_on_hf_hub(repo_id: str, filename: str, **kwargs) -> Option
     except (HFValidationError, EntryNotFoundError):
         # error, exists stays None
         pass
+    except Exception as e:
+        warnings.warn(
+            f"War not able to fetch remote file due to the following error {e} - silently ignoring the lookup"
```
pacman100, line 579:
```suggestion
            f"Unable to fetch remote file due to the following error {e} - silently ignoring the lookup"
```

pacman100 (commented):
Thank you Younes!

pacman100 (approved):

BenjaminBossan (approved):
I'm +/- 1 on this change. There could be "false negatives" from intercepting all types of exceptions, i.e. when it would indeed be better to raise an error. OTOH I understand that we don't want to add a new exception each time a user reports a new problem.

younesbelkada:
Thanks both for your review!

powerjohnnyli:
@pacman100 Thanks for this change, can we get a patch release with this included for easy installation using pip? What's the expected timeline on this?

PR #1459, opened 2024-02-13
[ `Adaptation Prompt`] Fix llama rotary embedding issue with transformers main

younesbelkada:
Currently the adaptation prompt slow tests are failing on transformers main due to a recent refactor to add compatibility with static cache. https://github.com/huggingface/transformers/pull/27931 

That PR broke the adaptation prompt forward pass, as `module.rotary_emb` forward pass now have `position_ids` as a required argument, thus breaking the current impmelementation of llama roatary embedding as we don't pass position_ids in PEFT.

In addition, cos / sin are now directly indexed through that method, therefore we don't need to call our method that also indexes cos / sin with `positon_ids`. As computing the query states given cos / sin is a one-liner, I slighlty modified the logic in `llama_compute_query_states` to make sure the forward pass works fine with transformers main

cc @pacman100 @BenjaminBossan

### DIFF
--- a/src/peft/tuners/adaption_prompt/utils.py
+++ b/src/peft/tuners/adaption_prompt/utils.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import inspect
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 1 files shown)


BenjaminBossan (approved):
Thanks for fixing this, LGTM.

Too bad that the signature changed and we have to resort to `inspect`, but I guess there isn't really a better way.

PR #1461, opened 2024-02-14
[`CI`] Add CI tests on transformers main to catch early bugs

younesbelkada:
As per title and as discussed offline, let's run CIs on transformers main, only on push PEFT main branch in order to check bugs such as https://github.com/huggingface/peft/pull/1459 early.

IMO for that workflow, no need to make things complicated, let's just test it on a speciifc python / OS version

cc @BenjaminBossan @pacman100

### DIFF
--- a/.github/workflows/tests-main.yml
+++ b/.github/workflows/tests-main.yml
@@ -1,0 +1,27 @@
+name: tests on transformers main
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  tests:
+    needs: check_code_quality
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.11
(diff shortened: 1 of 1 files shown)


.github/workflows/tests-main.yml
```diff
@@ -0,0 +1,27 @@
+name: tests on transformers main
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  tests:
+    needs: check_code_quality
```
younesbelkada, line 9:
```suggestion
```

pacman100 (approved):
Thank you @younesbelkada for adding workflow to run tests wrt transformers main branch on push to main! ✨

BenjaminBossan (approved):
Thanks a lot, LGTM.

I think we should get a notification from GH if this action fails, right?

younesbelkada:
Thanks everyone! 
@BenjaminBossan hmm not sure, as disucssed offline, maybe that has somehow to be enabled from your GH account but not sure

PR #1463, opened 2024-02-14
Make tests succeed more on MPS

akx:
Some tests would not succeed on MPS devices because `bfloat16` is not supported. This makes them get skipped instead.

On a similar note, tests that are not applicable (e.g. generated via a `parametrized` matrix and should have been skipped) no longer seem to quietly pass, but are instead marked `skip`.

It also fixes up the rtype for `infer_device` to always be a string – previously, it would return a string, or in the case of `mps`, a `torch.device()`.

Sibling of #1448, since I got frustrated trying to run tests on my Apple Silicon device and figuring which ones of them are actual failures.

### DIFF
--- a/tests/test_auto.py
+++ b/tests/test_auto.py
@@ -32,9 +32,12 @@ from peft import (
     PeftModelForSequenceClassification,
     PeftModelForTokenClassification,
 )
+from peft.utils import infer_device
 
 
 class PeftAutoModelTester(unittest.TestCase):
+    dtype = torch.float16 if infer_device() == "mps" else torch.bfloat16
+
     def test_peft_causal_lm(self):
         model_id = "peft-internal-testing/tiny-OPTForCausalLM-lora"
         model = AutoPeftModelForCausalLM.from_pretrained(model_id)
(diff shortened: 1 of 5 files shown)


akx:
cc @BenjaminBossan :)

tests/test_auto.py
```diff
@@ -32,9 +33,14 @@
     PeftModelForSequenceClassification,
     PeftModelForTokenClassification,
 )
+from peft.utils import infer_device
 
 
 class PeftAutoModelTester(unittest.TestCase):
+    def setUp(self):
+        if infer_device() == "mps":
+            return pytest.skip("bfloat16 is not supported on MPS")
```
BenjaminBossan, line 42: Instead of skipping all the tests in this class, I wonder if it wouldn't be better to remove the `torch_dtype=torch.bfloat16` setting here. The tests aren't really about bf16. Maybe this was added for memory reasons, but we could still use float16 if that were the case.

Pinging @younesbelkada since he added the tests.

younesbelkada: Since we are not performing any inference here they can indeed be in float16 - @akx can you try that out and see if it works?

akx: Yeah, that works. I'll make it do `float16` on MPS and `bfloat16` otherwise.

younesbelkada (commented):
Thanks a lot @akx ! We should be able to merge after fixing the styling checks ! 🙏 Can you format your changes ?

akx:
@younesbelkada Done... Can you maybe take a peek at #1467 so I don't need to manually remember to run `make style` and `make quality`? 😂

younesbelkada (approved):
Thanks for making these tests device-agnostic and addressing follow up enhancements on our testing suite ! 🤩

tests/test_auto.py
```diff
@@ -32,9 +32,12 @@
     PeftModelForSequenceClassification,
     PeftModelForTokenClassification,
 )
+from peft.utils import infer_device
 
 
 class PeftAutoModelTester(unittest.TestCase):
+    dtype = torch.float16 if infer_device() == "mps" else torch.bfloat16
```
BenjaminBossan, line 39: Would it be possible to check `if torch.cuda.is_bf16_supported()` instead or does that not work on Mac?

akx: I wouldn't trust anything that says `cuda` on the tin to work with Macs 😁

akx: Maybe that could be held off for later, anyway?

BenjaminBossan (approved):

BenjaminBossan:
Thanks @akx for helping out all the Mac users out there :)

PR #1466, opened 2024-02-15
add `magnitude_prune` merging method

pacman100:
### What does this PR do?
1. In my experiments, I found that pruning task tensors based on magnitude followed by weighted sum to perform better in comparison to dare_linear or ties. This PR add a new merging method `magnitude_prune` which first prunes the task tensors to remove small values and retain top_k values based on the param `density`. After this, a weighted sum is computed and returned.

### DIFF
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -36,7 +36,7 @@ from peft.utils import (
     _get_submodules,
     get_quantization_config,
 )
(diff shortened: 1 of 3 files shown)


BenjaminBossan (approved):
Thanks for adding this new adapter merging method. Implementation-wise, this LGTM. Since your experiments show that it works well, it's a great addition.

Let's ensure that it is documented by updating the docs added via #1423, depending on which PR lands first.

Also, as mentioned in a previous PR, it would be good to add unit tests for `merge_utils.py` eventually.

PR #1467, opened 2024-02-15
Add pre-commit configuration

akx:
# What does this PR do?

This PR adds a [pre-commit](https://pre-commit.com/) configuration, so developers may opt in to running `ruff` and `ruff-format` as a pre-commit hook.

Should reduce the need for "fix lint" commits and `check_code_quality` CI failures.

Sibling of https://github.com/huggingface/accelerate/pull/2451

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,13 @@
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.2.1
+    hooks:
+      - id: ruff
+        args:
+          - --fix
+      - id: ruff-format
+  - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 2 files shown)


younesbelkada (commented):
Thanks ! In principle I am ok with this PR as we have it on other libs as well, e.g. TRL ! 
I would like to hear what @BenjaminBossan & @pacman100 think, I agree pre-commit is quite convenient

BenjaminBossan:
I had some issues with pre-commit in the past. Some probably won't apply here (e.g. it being slow) but in the end, everyone is free to choose if they want to activate them or not, so I see no real harm (there is a small maintenance cost of keeping the config in sync).

BenjaminBossan (approved):

younesbelkada (approved):

PR #1471, opened 2024-02-16
Adds an example notebook for showing multi-adapter weighted inference

sayakpaul:
For the blog post.


pacman100 (approved):
Thank you @sayakpaul for adding the example for SD LoRAs! ✨

PR #1474, opened 2024-02-16
Implement DoRA

BenjaminBossan:
https://arxiv.org/abs/2402.09353

State:

- Only `Linear` layer supported, not conv or embedding
- Quantized layers not supported


Nikhil-Paleti:
https://github.com/catid/dora

nbasyl:
@BenjaminBossan Hi Benjamin, may I have your email, I am the author of DoRA, and would like to help with integrating DoRA into PEFT.

BenjaminBossan:
Thanks @nbasyl. You can contact me at `benjamin@` + HF domain.

nbasyl:
@BenjaminBossan  Sorry for the late reply, was consulting with my manager regrading this, I just sent out the email, please check!

RonanKMcGovern:
Great this is being added, is the PR at a point where I can test it out? Seems like I just add use_dora=True in the Lora config?

(btw, does LoRA alpha become redundant then if use_dora is True)?

BenjaminBossan:
@RonanKMcGovern There are still a few smaller kinks to work out, so don't test unless you're ready to test again once the next commits roll in.

LoRA alpha should still be relevant when DoRA is being used.

RonanKMcGovern:
Just gave this a quick spin and seems to do better on ppl than LoRA with the same r. Thanks for the nice work.

BenjaminBossan:
> Just gave this a quick spin and seems to do better on ppl than LoRA with the same r. Thanks for the nice work.

Thanks, sounds fantastic. If you can share anything like scripts, that would be great.

aliencaocao:
Hi @nbasyl do you think the concept of differential LR by Lora+ (https://arxiv.org/abs/2402.12354) can be integrated into DoRa? Both paper came out almost exact same time and they dont seem mutually exclusive.

BenjaminBossan:
@aliencaocao I skimmed the paper and code and I think there is no technical limitation to combining the two. I don't know if the gains from the two approaches would be additive though.

RonanKMcGovern:
> https://arxiv.org/abs/2402.12354

very nice, is there a pr for this yet for HF?

RonanKMcGovern:
> > Just gave this a quick spin and seems to do better on ppl than LoRA with the same r. Thanks for the nice work.
> 
> Thanks, sounds fantastic. If you can share anything like scripts, that would be great.

I'm using the huggingface trainer and this LoRA config for Qwen models:
```

# Initialize LoRA configuration
config = LoraConfig(
    r=8, 
    lora_alpha=32, 
    target_modules=[
      "q_proj",
      "k_proj",
      "v_proj",
      "o_proj",
      # "self_attn.rotary_emb.inv_freq",
      "gate_proj",
      "up_proj",
      "down_proj",
      # "input_layernorm.weight",
      # "post_attention_layernorm.weight",
      # "model.norm.weight",
      # "lm_head.weight"
    ],
    lora_dropout=0.1,
    bias="none",
    task_type="CAUSAL_LM",
    use_dora=True
)
```

BenjaminBossan:
> very nice, is there a pr for this yet for HF?

The code is published, there is a function here that you can simply call on your LoRA model and optimizer class:

https://github.com/nikhil-ghosh-berkeley/loraplus/blob/c8d30388f5d1ec6d8c93ff67330e4d36937d48dc/loraplus.py#L31

So far, the PEFT repo does not contain any code related to training directly, as we want to keep it agnostic with regard to the training method. If there is a big interest though, we could think about adding this function as a helper function for convenience. But that's a different topic than this PR ;)

src/peft/tuners/lora/config.py
```diff
@@ -101,6 +101,8 @@ class LoraConfig(PeftConfig):
             The configuration of LoftQ. If this is not None, then LoftQ will be used to quantize the backbone weights
             and initialize Lora layers. Also pass `init_lora_weights='loftq'`. Note that you should not pass a
             quantized model in this case, as LoftQ will quantize the model itself.
+        use_dora (`bool`):
+            TODO https://arxiv.org/abs/2402.09353
```
younesbelkada, line 105:
```suggestion
            Enables the DoRA decomposition as explain in the DoRA paper: https://arxiv.org/abs/2402.09353
```

src/peft/tuners/lora/config.py
```diff
@@ -224,6 +226,19 @@ class LoraConfig(PeftConfig):
...
+                "weights into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the "
+                "magnitude is handled by a separate learnable parameter. This can improve the performance of LoRA, "
+                "especially at low ranks. Right now, DoRA only supports non-quantized linear layers. DoRA introduces "
+                "a bigger overhead than pure LoRA, so it is recommended to merge weights for inference. For more "
+                "information, see  https://arxiv.org/abs/2402.09353."
+            )
+        },
+    )
 
     def __post_init__(self):
```
younesbelkada, line 243: Should we block some mis-intended usage ? e.g. if one passes `use_dora` & `loft_config` etc.

docs/source/developer_guides/lora.md
```diff
@@ -69,6 +69,16 @@ from peft import LoraConfig
...
+### Weight-Decomposed Low-Rank Adaptation (DoRA)
+
+This technique decomposes the updates of the weights into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the magnitude is handled by a separate learnable parameter. This can improve the performance of LoRA, especially at low ranks. Right now, DoRA only supports non-quantized linear layers. DoRA introduces a bigger overhead than pure LoRA, so it is recommended to merge weights for inference. For more information, see  https://arxiv.org/abs/2402.09353.
+
+```py
+from peft import LoraConfig
+
+config = LoraConfig(use_dora=True, ...)
+```
+
```
younesbelkada, line 81: Maybe add a mention that DoRA is not supported for quantized models 🙏

src/peft/tuners/lora/layer.py
```diff
@@ -47,6 +47,9 @@ def __init__(self, base_layer: nn.Module, **kwargs) -> None:
         # Mark the weight as unmerged
         self._disable_adapters = False
         self.merged_adapters = []
+        self.use_dora: dict[str, bool] = {}
+        self.lora_magnitude_vector: torch.nn.ParameterDict = torch.nn.ParameterDict({})  # for DoRA
```
younesbelkada, line 51: Hmmm this means now for all LoRA models we save this extra param. in the state dict (even if it's empty), do you know if that's the case? If not i think it should be all good, otherwise I think we should init it to None - what do you think?

src/peft/tuners/lora/layer.py
```diff
@@ -111,6 +116,13 @@ def update_layer(self, adapter_name, r, lora_alpha, lora_dropout, init_lora_weig
                 else:
                     self.to(weight.device)
                 break
+
+        if use_dora:
+            self.dora_init(adapter_name)
+            self.use_dora[adapter_name] = True
```
younesbelkada, line 122: this technically means you can have multiple adapters, with some of them using DoRA and some not?

younesbelkada (commented):
Thanks for your great work and leading this @BenjaminBossan ! IMO the PR is already great as it is
I left some initial open questions ! What do you think?

src/peft/tuners/lora/layer.py
BenjaminBossan: I think it's not included in the `state_dict`. E.g. we already have `lora_A` vs `lora_embedding_A` etc., one of which is always empty depending on layer type. But I still changed the default to be `None`, as it makes the repr smaller.

src/peft/tuners/lora/layer.py
BenjaminBossan: I guess so, although it wouldn't currently be possible to configure it as such.

BenjaminBossan:
@younesbelkada I addressed your comments, please check again.

src/peft/tuners/lora/layer.py
younesbelkada: even if you do `add_adapter` witha LoraConfig that has `use_dora=True` ?

RonanKMcGovern:
I have been running some fine-tunes of LoRA vs DoRA. The only change to my script is turning on use_dora here:
```
model.gradient_checkpointing_enable()

peft_config = LoraConfig(
    r=8,
    lora_alpha=32,
    target_modules=[
        "q_proj",
        "k_proj",
        "v_proj",
        "o_proj",
        "gate_proj",
        "up_proj",
        "down_proj",
        "lora_magnitude_vector", #required for DoRA
    ],
    lora_dropout=0.1,
    bias="none",
    task_type="CAUSAL_LM",
    use_dora=True # only for DoRA
)

model = get_peft_model(model, peft_config) #move to a peft model
```

I'm finding that DoRA is taking 2-2.5X longer on a 7B model.

Obviously with DoRA, 'lora_magnitude_vector' creates an added set of trainable parameters, but this is a small increase in params (total of 74,164,224 vs 73,654,272 in my specific case).

Does this slow doesn't seem intuitive? Seems rather large.

docs/source/developer_guides/lora.md
```diff
@@ -69,6 +69,16 @@ from peft import LoraConfig
 config = LoraConfig(use_rslora=True, ...)
 ```
 
+### Weight-Decomposed Low-Rank Adaptation (DoRA)
+
+This technique decomposes the updates of the weights into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the magnitude is handled by a separate learnable parameter. This can improve the performance of LoRA, especially at low ranks. Right now, DoRA only supports non-quantized linear layers. DoRA introduces a bigger overhead than pure LoRA, so it is recommended to merge weights for inference. For more information, see  https://arxiv.org/abs/2402.09353.
```
sayakpaul, line 74: If we support conv layers for regular LoRA, would it be very non-trivial to support conv layers with DoRA?

sayakpaul: I would also hyperlink "merge weights for inference" so that users know what we're talking about here.

src/peft/tuners/lora/config.py
```diff
@@ -101,6 +101,13 @@ class LoraConfig(PeftConfig):
             The configuration of LoftQ. If this is not None, then LoftQ will be used to quantize the backbone weights
             and initialize Lora layers. Also pass `init_lora_weights='loftq'`. Note that you should not pass a
             quantized model in this case, as LoftQ will quantize the model itself.
+        use_dora (`bool`):
```
sayakpaul, line 104: Nice docstrings.

src/peft/tuners/lora/layer.py
```diff
@@ -47,6 +48,9 @@ def __init__(self, base_layer: nn.Module, **kwargs) -> None:
         # Mark the weight as unmerged
         self._disable_adapters = False
         self.merged_adapters = []
+        self.use_dora: dict[str, bool] = {}
```
sayakpaul, line 51: I thought `use_dora` is a boolean flag? What are the instances of it being a `dict`?

sayakpaul, line 51: (nit): `use_dora` where the variable is supposed to be a `dict` -- doesn't seem like a good name. Could we go for something else?

tests/test_initialization.py
```diff
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import unittest
-
+import pytest
```
sayakpaul, line 15: Haven't checked thoroughly but have we started migrating to `pytest` in `peft`?

sayakpaul (approved):
Thanks for working on this so quickly. 

I left a few minor nits and questions. My biggest question is - is it possible to do weighted inference with a non-DoRA LoRA and a DoRA-LoRA? The current PR seems to suggest it's possible.

nbasyl:
Hi @Benjamin, just some minor suggestions to further reduce the training time and GPU cost. 
I think it would be slightly more efficient to modify the apply_dora function to 
```
def apply_dora(self, basemodel_result, x, lora_weight, lora_A, lora_B, active_adapter):
        scaling = self.scaling[active_adapter]
        magnitude = self.lora_magnitude_vector[active_adapter]
        weight_norm = self._get_weight_norm(weight, lora_weight, scaling)
        # see section 4.3 of DoRA (https://arxiv.org/abs/2402.09353)
        # "[...] we suggest treating ||V +∆V ||_c in
        # Eq. (5) as a constant, thereby detaching it from the gradient
        # graph. This means that while ||V + ∆V ||_c dynamically
        # reflects the updates of ∆V , it won’t receive any gradient
        # during backpropagation"
        dora_weight = transpose(weight + scaling * lora_weight, self.fan_in_fan_out)
        # original implementation: 
        # return (magnitude / weight_norm - 1).view(1, -1) * F.linear(x, dora_weight)
        # new
        mag_norm_scale = (magnitude / weight_norm - 1).view(1, -1)
        return mag_norm_scale * basemodel_result + mag_norm_scale * lora_B(lora_A(x)) * scaling
```
This way you don't need to recompute the basemodel_result. Besides, I also found that calculating the dora_weight then pass it to F.linear takes more GPU memory than `mag_norm_scale * basemodel_result + mag_norm_scale * lora_B(lora_A(x)) * scaling`.

nbasyl:
> Hi @benjamin, just some minor suggestions to further reduce the training time and GPU cost. I think it would be slightly more efficient to modify the apply_dora function to
> 
> ```
> def apply_dora(self, basemodel_result, x, lora_weight, lora_A, lora_B, active_adapter):
>         scaling = self.scaling[active_adapter]
>         magnitude = self.lora_magnitude_vector[active_adapter]
>         weight = self.get_base_layer().weight
>         weight_norm = self._get_weight_norm(weight, lora_weight, scaling)
>         # see section 4.3 of DoRA (https://arxiv.org/abs/2402.09353)
>         # "[...] we suggest treating ||V +∆V ||_c in
>         # Eq. (5) as a constant, thereby detaching it from the gradient
>         # graph. This means that while ||V + ∆V ||_c dynamically
>         # reflects the updates of ∆V , it won’t receive any gradient
>         # during backpropagation"
>         weight_norm = weight_norm.detach()
>         dora_weight = transpose(weight + scaling * lora_weight, self.fan_in_fan_out)
>         # original implementation: 
>         # return (magnitude / weight_norm - 1).view(1, -1) * F.linear(x, dora_weight)
>         # new
>         mag_norm_scale = (magnitude / weight_norm - 1).view(1, -1)
>         return mag_norm_scale * basemodel_result + mag_norm_scale * lora_B(lora_A(x)) * scaling
> ```
> 
> This way you don't need to recompute the basemodel_result. Besides, I also found that calculating the dora_weight then pass it to F.linear takes more GPU memory than `mag_norm_scale * basemodel_result + mag_norm_scale * lora_B(lora_A(x)) * scaling`.

Just realize that the code can be even more efficient by reusing `lora_B(lora_A(x)) * scaling` which is already computed outside self.apply_dora and pass it as a function argument to self.apply_dora. 

Something like:
`return mag_norm_scale * basemodel_result + mag_norm_scale *  lora_result`

BenjaminBossan:
@nbasyl I tried to integrate your suggestion. Maybe I was missing a bit of context but the results didn't quite line up and I got failing tests. But based on this, I tried another way which I think should be more efficient, please check this out:

```python
                # ...
                lora_A = self.lora_A[active_adapter]
                lora_B = self.lora_B[active_adapter]
                dropout = self.lora_dropout[active_adapter]
                scaling = self.scaling[active_adapter]
                x = x.to(lora_A.weight.dtype)

                else:
                    x = dropout(x)
                    lora_weight = lora_B.weight @ lora_A.weight
                    magnitude = self.lora_magnitude_vector[active_adapter]
                    weight_norm = self._get_weight_norm(weight, lora_weight, scaling)
                    # see section 4.3 of DoRA (https://arxiv.org/abs/2402.09353)
                    mag_norm_scale = (magnitude / weight_norm).view(1, -1)
                    lora_result = lora_B(lora_A(x)) * scaling
                    result = result - self.get_base_layer().bias
                    result = mag_norm_scale * result + mag_norm_scale * lora_result
                    result = result + self.get_base_layer().bias
```

I in-lined the `apply_dora` call to have all code in one place. WDYT of this approach, I believe it should be more readable and also faster. It is a bit unfortunate that I had to remove and re-add the bias, I'm not quite sure if this would be valid in all circumstances. It might not quite work if the base model is not a pure vanilla `Linear` layer, but then again the whole DoRA approach is already making that assumption, so we could be fine.

A first few tests didn't show conclusive results when it came to speed, probably the tested workloads are not big enough, but I think it should be faster because we do not need to compute anything twice (except the bias part but that should be fast).

src/peft/tuners/lora/layer.py
BenjaminBossan: I agree in principle, but I wanted to be consistent with the name of the other model params, which are also dicts to support multiple adapters. E.g. `lora_alpha`, `scaling`, `lora_A`, `lora_B` are all dicts. If I would rename this to something like `use_dora_mapping`, it would be inconsistent.

src/peft/tuners/lora/layer.py
BenjaminBossan: I guess we could have multiple adapters, some of which use DoRA and some don't. I believe this would be fine or can you think of an issue?

tests/test_initialization.py
BenjaminBossan: From my understanding, we don't really plan on migrating. Instead, we can consider using more pytest features if we add new tests or are updating existing tests. Here I thought it would only minimally change the code, so doing this change should be fine.

BenjaminBossan:
> My biggest question is - is it possible to do weighted inference with a non-DoRA LoRA and a DoRA-LoRA?

You mean via `add_weighted_adapter`? No, I don't think that would work correctly as of now. The issue is that if we take something like `current_adapter_lora_A = target.lora_A[adapter].weight`, we haven't taken the new DoRA parameters into account yet. We would have to integrate them but I would have to think about how to do that correctly. It's probably enough work for a different PR.

docs/source/developer_guides/lora.md
nbasyl: I think DoRA can definitely be applied on conv2d similar to how LoRA is applied to conv2d. And the normalization could be calculated per-input channel.

sayakpaul: Thanks! We will loop you in when we add the support! 

Cc: @BenjaminBossan

BenjaminBossan: Great. Let's add Conv2d in a follow-up PR once this one is merged.

nbasyl:
@BenjaminBossan 
I found that my last suggestion is actually wrong, which explain why the results didn't quite line up.
Below is the correct implementation:
```
                # ...
                lora_A = self.lora_A[active_adapter]
                lora_B = self.lora_B[active_adapter]
                dropout = self.lora_dropout[active_adapter]
                scaling = self.scaling[active_adapter]
                x = x.to(lora_A.weight.dtype)

                else:
                    x = dropout(x)
                    lora_weight = lora_B.weight @ lora_A.weight
                    magnitude = self.lora_magnitude_vector[active_adapter]
                    weight_norm = self._get_weight_norm(weight, lora_weight, scaling)
                    # see section 4.3 of DoRA (https://arxiv.org/abs/2402.09353)
                    mag_norm_scale = (magnitude / weight_norm).view(1, -1)
                    result += (mag_norm_scale - 1) *  (F.linear(x, transpose(weight, self.fan_in_fan_out))) + mag_norm_scale * lora_B(lora_A(x)) * scaling
                    
```
you can refer to the following formulation
image

Note that `(mag_norm_scale - 1) *  (F.linear(x, transpose(weight, self.fan_in_fan_out)))` must be included to properly apply dropout; otherwise, the outcome would be inaccurate. Therefore, we can not simply reuse the base_result as suggested before.

nbasyl:
> @BenjaminBossan I found that my last suggestion is actually wrong, which explain why the results didn't quite line up. Below is the correct implementation:
> 
> ```
>     def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
>                 # ...
>                 lora_A = self.lora_A[active_adapter]
>                 lora_B = self.lora_B[active_adapter]
>                 dropout = self.lora_dropout[active_adapter]
>                 scaling = self.scaling[active_adapter]
>                 x = x.to(lora_A.weight.dtype)
> 
>                 if not self.use_dora[active_adapter]:
>                     result = result + lora_B(lora_A(dropout(x))) * scaling
>                 else:
>                     x = dropout(x)
>                     lora_weight = lora_B.weight @ lora_A.weight
>                     magnitude = self.lora_magnitude_vector[active_adapter]
>                     weight = self.get_base_layer().weight
>                     weight_norm = self._get_weight_norm(weight, lora_weight, scaling)
>                     # see section 4.3 of DoRA (https://arxiv.org/abs/2402.09353)
>                     weight_norm = weight_norm.detach()
>                     mag_norm_scale = (magnitude / weight_norm).view(1, -1)
>                     result += (mag_norm_scale - 1) *  (F.linear(x, transpose(weight, self.fan_in_fan_out))) + mag_norm_scale * lora_B(lora_A(x)) * scaling
>                     
> ```
> 
> you can refer to the following formulation image
> 
> Note that `(mag_norm_scale - 1) * (F.linear(x, transpose(weight, self.fan_in_fan_out)))` must be included to properly apply dropout; otherwise, the outcome would be inaccurate. Therefore, we can not simply reuse the base_result as suggested before.

The only difference is `mag_norm_scale * lora_B(lora_A(x)) * scaling` which require less GPU memory compared to the original implementation where you explicitly calculate the dora_weight and apply F.linear.

nbasyl:
I believe that in your implementation, where you reuse the result by removing and then adding back the bias, there may be issues when updating the magnitude vector. This is because the magnitude vector should only consider dropout input, but with your approach, it also interacts with pre-dropout input. While this might not seem like a big difference, I've actually tested both methods, and they yield noticeably different training results.

nbasyl:
For your implementation, where you reuse the result by removing and then adding back the bias, there may be issues when updating the magnitude vector. This is because the magnitude vector should only consider dropout input, but with your approach, it also interacts with pre-dropout input. While this might not seem like a big difference, I've actually already tested both implementation when I first came up with DoRA, and they yield noticeably different training results.

nbasyl (approved):

BenjaminBossan:
> This is because the magnitude vector should only consider dropout input, but with your approach, it also interacts with pre-dropout input.

Ah I see now what you mean, true. The other method could still be used when we have `dropout=0`, right? But I agree, let's stick with the more accurate implementation.

nbasyl:
Right, two are the same when dropout = 0

BenjaminBossan:
> Right, two are the same when dropout = 0

Thanks for confirming, also thanks a lot for you valuable feedback all throughout this PR.

nbasyl:
Your welcome! Appreciate your effort on making DoRA available on PEFT this fast! Can't wait for more people to try DoRA!

src/peft/tuners/lora/layer.py
```diff
@@ -153,6 +166,65 @@ def loftq_init(self, adapter_name):
...
+        # calculate L2 norm of weight matrix, column-wise
+        weight = weight + scaling * lora_weight
+        weight_norm = torch.linalg.norm(weight, dim=1)
+        return weight_norm
+
+    def dora_init(self, adapter_name: str) -> None:
+        lora_A = self.lora_A[adapter_name]
+        lora_B = self.lora_B[adapter_name]
+        scaling = self.scaling[adapter_name]
+        with gather_params_ctx(self.get_base_layer()):
```
younesbelkada, line 179: Do you think there is a way we could always avoid calling that context manager? I wonder if this might create some weird interactions if one does not use DeepSpeed Zero-3 with Trainer as `is_deepspeed_zero_3_available()` seems quite specific to `Trainer`: https://github.com/huggingface/transformers/blob/a3f9221a449e9b949e71d9b047c66186f023481f/src/transformers/integrations/deepspeed.py#L286

BenjaminBossan: I'm not sure I understand the question. We have to call the context manager or else this will break with DeepSpeed. If a user does not use DS but uses `Trainer`, should `is_deepspeed_zero3_enabled()` not return `False`? In this case the context manager just does nothing and should still work.

BenjaminBossan (commented):
Thanks @younesbelkada for catching the transformers version dependency for `is_deepspeed_zero3_enabled`, I added a check for the transformers version now.

src/peft/tuners/lora/layer.py
younesbelkada: ok yes makes sense!

younesbelkada (approved):
Thanks a lot @BenjaminBossan !

RonanKMcGovern:
Well done @BenjaminBossan

aliencaocao:
@BenjaminBossan On a RTX 3090 finetuning llava1.6 mistral 7b, I get half the throughput as without dora but with lora.

sayakpaul:
And with DoRA?

aliencaocao:
What i meant was, with dora is half the speed as without dora but with lora

sayakpaul:
Oh okay. On the same rank?

sayakpaul:
If not, maybe try with a lower rank which should hopefully compensate for the throughput without disturbing the performance.

aliencaocao:
Yes same rank

nbasyl:
Hi, one way to further reduce the GPU cost (# of trainable parameters) as well as the training time is to only finetune the magnitude component of certain modules while finetuning both the magnitude and directional component for the remaining modules. You can refer to Sec.5.6 of the paper for more details. This feature can be added in another PR to give user more flexibility.

aliencaocao:
so is double the cost to be expected here because there is double the trainable params?

nbasyl:
If the same rank is utilized, the number of trainable parameters is only slightly higher, approximately 0.01%, compared to LoRA. However, the current computation overhead arises from the need to calculate the weight's norm as well as the calculation of the second term in the given formulation (can simply reuse the first time due to dropout alignment issue).
image

nbasyl:
> Hi, one way to further reduce the GPU cost (# of trainable parameters) as well as the training time is to only finetune the magnitude component of certain modules while finetuning both the magnitude and directional component for the remaining modules. You can refer to Sec.5.6 of the paper for more details. This feature can be added in another PR to give user more flexibility.

@BenjaminBossan, I have already implemented this feature. Once our code is released (should be released soon), you can take a look at it and see if you can integrate this functionality with another pull request (PR). Or we can discuss via email to start working on this new PR earlier.

BenjaminBossan:
> On a RTX 3090 finetuning llava1.6 mistral 7b, I get half the throughput as without dora but with lora.

Thanks for the feedback. On my puny 2060 with bloomz-560m and fp16, I got 15-20% slowdown during training with DoRA enabled, same rank. So there seem to be pretty big differences based on the exact model or settings. @nbasyl do you have some numbers that you could share?

> one way to further reduce the GPU cost (# of trainable parameters) as well as the training time is to only finetune the magnitude component of certain modules while finetuning both the magnitude and directional component for the remaining modules. You can refer to Sec.5.6 of the paper for more details.

It would make the whole thing more complex. Are the gains big enough to be worth it?

I'll definitely take a closer look and check what else we can do to improve runtime performance while keeping the code flexible and maintainable.

RonanKMcGovern:
BTW, this merge is causing some breaking changes on packages using the code of peft, such as unsloth: https://github.com/unslothai/unsloth/issues/201

```
TypeError                                 Traceback (most recent call last)
Cell In[23], line 2
      1 # Do model patching and add fast LoRA weights
----> 2 model = FastLanguageModel.get_peft_model(
      3     model,
      4     r = 8,
      5     target_modules = ["q_proj", "k_proj", "v_proj", "o_proj",
      6                       "gate_proj", "up_proj", "down_proj",],
      7     lora_alpha = 32,
      8     lora_dropout = 0, # Dropout = 0 is currently optimized
      9     bias = "none",    # Bias = "none" is currently optimized
     10     use_gradient_checkpointing = True,
     11     random_state = 3407,
     12 )

File /usr/local/lib/python3.10/dist-packages/unsloth/models/llama.py:1313, in FastLlamaModel.get_peft_model(model, r, target_modules, lora_alpha, lora_dropout, bias, layers_to_transform, layers_pattern, use_gradient_checkpointing, random_state, max_seq_length, use_rslora, init_lora_weights, loftq_config, **kwargs)
   1310 if not SUPPORTS_RSLORA: del arguments["use_rslora"]
   1312 lora_config = LoraConfig(**arguments)
-> 1313 model = _get_peft_model(model, lora_config)
   1315 model = FastLlamaModel.patch_peft_model(model, use_gradient_checkpointing)
   1316 return model

File /usr/local/lib/python3.10/dist-packages/peft/mapping.py:136, in get_peft_model(model, peft_config, adapter_name, mixed)
    134 if peft_config.is_prompt_learning:
    135     peft_config = _prepare_prompt_learning_config(peft_config, model_config)
--> 136 return MODEL_TYPE_TO_PEFT_MODEL_MAPPING[peft_config.task_type](model, peft_config, adapter_name=adapter_name)

File /usr/local/lib/python3.10/dist-packages/peft/peft_model.py:1059, in PeftModelForCausalLM.__init__(self, model, peft_config, adapter_name)
   1058 def __init__(self, model: torch.nn.Module, peft_config: PeftConfig, adapter_name: str = "default") -> None:
-> 1059     super().__init__(model, peft_config, adapter_name)
   1060     self.base_model_prepare_inputs_for_generation = self.base_model.prepare_inputs_for_generation

File /usr/local/lib/python3.10/dist-packages/peft/peft_model.py:126, in PeftModel.__init__(self, model, peft_config, adapter_name)
    124     self._peft_config = None
    125     cls = PEFT_TYPE_TO_MODEL_MAPPING[peft_config.peft_type]
--> 126     self.base_model = cls(model, {adapter_name: peft_config}, adapter_name)
    127     self.set_additional_trainable_modules(peft_config, adapter_name)
    129 if getattr(model, "is_gradient_checkpointing", True):

File /usr/local/lib/python3.10/dist-packages/peft/tuners/lora/model.py:111, in LoraModel.__init__(self, model, config, adapter_name)
    110 def __init__(self, model, config, adapter_name) -> None:
--> 111     super().__init__(model, config, adapter_name)

File /usr/local/lib/python3.10/dist-packages/peft/tuners/tuners_utils.py:147, in BaseTuner.__init__(self, model, peft_config, adapter_name)
    144         self.peft_config.update(peft_config)
    146 self.active_adapter = adapter_name
--> 147 self.inject_adapter(self.model, adapter_name)
    149 # Copy the peft_config in the injected model.
    150 self.model.peft_config = self.peft_config

File /usr/local/lib/python3.10/dist-packages/peft/tuners/tuners_utils.py:302, in BaseTuner.inject_adapter(self, model, adapter_name)
    300     is_target_modules_in_base_model = True
    301     parent, target, target_name = _get_submodules(model, key)
--> 302     self._create_and_replace(peft_config, adapter_name, target, target_name, parent, current_key=key)
    304 if not is_target_modules_in_base_model:
    305     raise ValueError(
    306         f"Target modules {peft_config.target_modules} not found in the base model. "
    307         f"Please check the target modules and try again."
    308     )

File /usr/local/lib/python3.10/dist-packages/peft/tuners/lora/model.py:182, in LoraModel._create_and_replace(self, lora_config, adapter_name, target, target_name, parent, current_key)
    172     target.update_layer(
    173         adapter_name,
    174         r,
   (...)
    179         use_dora=lora_config.use_dora,
    180     )
    181 else:
--> 182     new_module = self._create_new_module(lora_config, adapter_name, target, **kwargs)
    183     if adapter_name != self.active_adapter:
    184         # adding an additional adapter: it is not automatically trainable
    185         new_module.requires_grad_(False)

File /usr/local/lib/python3.10/dist-packages/peft/tuners/lora/model.py:257, in LoraModel._create_new_module(lora_config, adapter_name, target, **kwargs)
    255 new_module = None
    256 for dispatcher in dispatchers:
--> 257     new_module = dispatcher(target, adapter_name, lora_config=lora_config, **kwargs)
    258     if new_module is not None:  # first match wins
    259         break

File /usr/local/lib/python3.10/dist-packages/peft/tuners/lora/layer.py:858, in dispatch_default(target, adapter_name, lora_config, **kwargs)
    856         kwargs["fan_in_fan_out"] = lora_config.fan_in_fan_out = False
    857     kwargs.update(lora_config.loftq_config)
--> 858     new_module = Linear(target, adapter_name, **kwargs)
    859 elif isinstance(target_base_layer, Conv1D):
    860     if not kwargs["fan_in_fan_out"]:

File /usr/local/lib/python3.10/dist-packages/peft/tuners/lora/layer.py:289, in Linear.__init__(self, base_layer, adapter_name, r, lora_alpha, lora_dropout, fan_in_fan_out, is_target_conv_1d_layer, init_lora_weights, use_rslora, use_dora, **kwargs)
    286 self.fan_in_fan_out = fan_in_fan_out
    288 self._active_adapter = adapter_name
--> 289 self.update_layer(
    290     adapter_name,
    291     r,
    292     lora_alpha=lora_alpha,
    293     lora_dropout=lora_dropout,
    294     init_lora_weights=init_lora_weights,
    295     use_rslora=use_rslora,
    296     use_dora=use_dora,
    297 )
    298 self.is_target_conv_1d_layer = is_target_conv_1d_layer

TypeError: LoraLayer_update_layer() got an unexpected keyword argument 'use_dora'
```

nbasyl:
> > On a RTX 3090 finetuning llava1.6 mistral 7b, I get half the throughput as without dora but with lora.
> 
> Thanks for the feedback. On my puny 2060 with bloomz-560m and fp16, I got 15-20% slowdown during training with DoRA enabled, same rank. So there seem to be pretty big differences based on the exact model or settings. @nbasyl do you have some numbers that you could share?
> 
> > one way to further reduce the GPU cost (# of trainable parameters) as well as the training time is to only finetune the magnitude component of certain modules while finetuning both the magnitude and directional component for the remaining modules. You can refer to Sec.5.6 of the paper for more details.
> 
> It would make the whole thing more complex. Are the gains big enough to be worth it?
> 
> I'll definitely take a closer look and check what else we can do to improve runtime performance while keeping the code flexible and maintainable.

I have tried finetuning LLaMA-7B/13B on 3090/4090/V100/A100 and only also got around 20% slow down. I am assuming that such drastic slowdown for @aliencaocao is probably caused by the optimization problem of deepspeed which is used by the LLaVA code base.

nbasyl:
> > On a RTX 3090 finetuning llava1.6 mistral 7b, I get half the throughput as without dora but with lora.
> 
> Thanks for the feedback. On my puny 2060 with bloomz-560m and fp16, I got 15-20% slowdown during training with DoRA enabled, same rank. So there seem to be pretty big differences based on the exact model or settings. @nbasyl do you have some numbers that you could share?
> 
> > one way to further reduce the GPU cost (# of trainable parameters) as well as the training time is to only finetune the magnitude component of certain modules while finetuning both the magnitude and directional component for the remaining modules. You can refer to Sec.5.6 of the paper for more details.
> 
> It would make the whole thing more complex. Are the gains big enough to be worth it?
> 
> I'll definitely take a closer look and check what else we can do to improve runtime performance while keeping the code flexible and maintainable.

The gain here is not the accuracy improvement, but in the reduction of trainable parameters. You can refer to this table.
image
Besides. for my case, I implemented this feature under the PEFT package framework and didn't modify the code too much, so I think it wouldn't be as complicated as you thought.

peterjc123:
According to my experiment, DoRA is ~2x slower than LoRA, and I am able to achieve lower loss with DoRA.
Some numbers & settings
GPU: 1x Nvidia 3080 (10GB)
LoRA rank: 64
Model: Qwen 1.5 - 1.8B
Max Length: 512
Packing: True
Gradient Checkpointing: True
Batch: 2
Gradient Accumulation Steps: 8
Optim: Adam8Bit
GPU mem usage: ~6800MB
Use Flash Attention2: True
LoRA timings: 2.7 s/Iter
DoRA timings: 5.5 s/Iter
DoRA timings optimized(https://github.com/huggingface/peft/pull/1474#issuecomment-1964699829): 4.5 s/Iter

BenjaminBossan:
@nbasyl Okay, this suggestion sounds good for the purpose of saving on the number of trainable parameters (even if runtime would probably not change much). I'm not sure yet how to configure this so that it would work with most or all model architectures, we'd probably have to add a new config argument for that. If your code is to be released soon, we can wait for that and add it to PEFT afterwards.

When it comes to VeRA, would DVoRA help with reducing the overhead of calculating the weight norm (since W0 and BA are fixed) or is it the same cost as DoRA?

@peterjc123 Thanks for providing your settings. I tried to replicate (using Qwen1.5-0.5B for memory reasons) and got ~30% (2060) to ~40% (T4) slower training with DoRA activated.

> it is a little bit not worth it compared to training with QLoRA on a larger model

I'll investigate if we can make DoRA work with bnb. It could be a bit tricky when it comes to calculating the weight norm, let's see.

BenjaminBossan:
I created a PR to support DoRA with bnb (QDoRA): #1518 

I tested it on a small use case and it worked. If someone wants to give it a spin and report the results, that would be fantastic.

peterjc123:
> I created a PR to support DoRA with bnb (QDoRA): #1518
> 
> I tested it on a small use case and it worked. If someone wants to give it a spin and report the results, that would be fantastic.

@BenjaminBossan 

With bf16=True, your implementation OOM when batch = 2 and the configuration I posted above. It also throws the following warning.
```
The input hidden states seems to be silently casted in float32, this might be related to the fact you have upcasted embedding or layer norm layers in float32. We will cast back the input in bfloat16.
```
And then, I debugged your code and found out that the dtype of `weight_norm` in `apply_dora` is `torch.float32`. So the dtype of the output of the LoRA layer is `torch.float32`, too.
```py
-> weight_norm = self._get_weight_norm(weight, lora_weight, scaling)
(Pdb) s
--Call--
> peft/src/peft/tuners/lora/layer.py(172)_get_weight_norm()
-> def _get_weight_norm(self, weight, lora_weight, scaling) -> torch.Tensor:
(Pdb) n
> peft/src/peft/tuners/lora/layer.py(174)_get_weight_norm()
-> weight = weight + scaling * lora_weight
(Pdb) p weight.dtype
torch.bfloat16
(Pdb) p scaling
0.25
(Pdb) p lora_weight.dtype
torch.bfloat16
(Pdb) n
> peft/src/peft/tuners/lora/layer.py(175)_get_weight_norm()
-> weight_norm = torch.linalg.norm(weight, dim=1)
(Pdb) p weight.dtype
torch.bfloat16
(Pdb) n
> peft/tuners/lora/layer.py(176)_get_weight_norm()
-> return weight_norm
(Pdb) p weight_norm.dtype
torch.float32
```
As you can see, `torch.linalg.norm` returns a float32 `tensor` regardless of the bfloat16 input.

BenjaminBossan:
> when launching [llava training](https://github.com/haotian-liu/LLaVA/blob/main/scripts/v1_5/finetune_task_lora.sh) (using `deepspeed --include 'localhost:0,1,3,4'`) with `use_dora=True` added [here](https://github.com/haotian-liu/LLaVA/blob/main/llava/train/train.py#L869), dora works perfectly on ZeRO3, but mysteriously deadlocks on ZeRO2.

Very strange, typically PEFT has issues with ZeRO3, not ZeRO2. I don't know enough about DS to tell what could be the cause of the deadlock here.

> So I have to change that line to

Thanks for investigating, I pushed this change to the PR.

> After fixing this, it is still super slow, yields about `12.96s/it`.

That's unfortunate, but not unexpected. The issue is that we need to have an additional dequantization step for DoRA, as we have to calculate the weight norm of the quantized weight + LoRA. I couldn't come up with a way to avoid this or somehow cache the results. If you or someone else has an idea, please let me know.

PR #1475, opened 2024-02-16
Update docstring at peft_types.py

eduardozamudio:
Docstring shows a missing underscore at TaskType "Causal LM"


younesbelkada (approved):
Thanks for fixing the typo!

PR #1480, opened 2024-02-19
FIX [`CI` / `bnb`] Fix failing bnb workflow

younesbelkada:
This PR fixes the current bnb workflow that is broken due to an incompatible pytest version with transformers leading to transformers tests being silently skipped: https://github.com/huggingface/peft/actions/runs/7945902843/job/21693054056

cc @pacman100 @BenjaminBossan @Titus-von-Koeller FYI

### DIFF
--- a/.github/workflows/nightly-bnb.yml
+++ b/.github/workflows/nightly-bnb.yml
@@ -36,6 +36,7 @@ jobs:
           source activate peft
           pip install -e . --no-deps
           pip install pytest-reportlog pytest-cov parameterized datasets scipy einops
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Is this something that only affects bnb or should we pin pytest in general?

younesbelkada:
@BenjaminBossan it only affect bnb as we run the bnb transformers tests on the transformers repo

BenjaminBossan (approved):
> it only affect bnb as we run the bnb transformers tests on the transformers repo

I see, so this is an issue only with transformers tests + pytest>=8.0.0.

younesbelkada:
Thanks for the review @BenjaminBossan ! 🙏

PR #1481, opened 2024-02-19
ENH: [`CI` / `Docker`]: Create a workflow to temporarly build docker images in case dockerfiles are modified

younesbelkada:
Hi !

I would like to enhance the current way of testing docker build workflow, right now I need to manually change the rules for triggering the build docker image job to test if a PR leads to potential failure on docker build. In this PR I propose to simply trigger a simple workflow that just builds (without pushing) docker images that are modified by a PR

cc @pacman100 @BenjaminBossan

### DIFF
--- a/.github/workflows/test-docker-build.yml
+++ b/.github/workflows/test-docker-build.yml
@@ -31,4 +31,12 @@ jobs:
         for file in ${CHANGED_FILES}; do
           echo "$file was changed"
         done
+    - name: Build Docker images
+      strategy:
+        matrix:
+          docker-file: ${{ steps.changed-files.outputs.all_changed_files}}
+      uses: docker/build-push-action@v4
+      with:
+        context: ${{ matrix.docker-file }}
+        push: False
 
--- a/.github/workflows/test-docker-build.yml
+++ b/.github/workflows/test-docker-build.yml
@@ -1,0 +1,34 @@
+name: Test Docker images (scheduled)
+
+on:
+  pull_request:
+  paths:
+    # Run only when python files are modified
+    - "docker/**"
+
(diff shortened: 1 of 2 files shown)


younesbelkada:
@glegendre01 do you think it's ok to move forward with this new workflow?

.github/workflows/test-docker-build.yml
```diff
@@ -0,0 +1,42 @@
+name: Test Docker images (scheduled)
+
+on:
+  pull_request:
+  paths:
+    # Run only when python files are modified
```
younesbelkada, line 6:
```suggestion
    # Run only when DockerFile files are modified
```

glegendre01 (changes requested):

glegendre01 (approved):

BenjaminBossan (approved):
In general, LGTM, thanks for adding this useful feature.

A minor concern I have is about adding this new GH action. Could it be a possible security concern? Maybe it's a non-issue, it was just something that crossed my mind.

younesbelkada:
Thanks for the review! 
I think that it should be all good as we don't push any docker image using that workflow and we use a GHA pinned on a specific commit (from @glegendre01 comment), in any case all PRs needs a maintainer approval before running the CI so we should be careful when reviewing PRs where we have some doubts (by pinging @glegendre01 if we have some strong doubts 🙏 )

younesbelkada:
Also I need to merge this PR before properly testing it as GH does not allow to add new workflow from PRs :D

pacman100 (approved):
Thank you @younesbelkada, nice QoL update for contributors! 🤗

PR #1482, opened 2024-02-19
ENH [`CI`] Run tests only when relevant files are modified

younesbelkada:
Let's run the whole tests CI only when relevant files are changed, for minor PRs such as doc PR, there should be no need to run the testing suite IMO

Reference: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

cc @pacman100 @BenjaminBossan

### DIFF
--- a/.github/workflows/tests-main.yml
+++ b/.github/workflows/tests-main.yml
@@ -3,6 +3,8 @@ name: tests on transformers main
 on:
   push:
     branches: [main]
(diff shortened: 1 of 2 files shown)


BenjaminBossan (approved):
Thanks, makes sense.

PR #1493, opened 2024-02-21
FIX: [`CI` / `Adaptation Prompt`] Fix CI on transformers main

younesbelkada:
# What does this PR do?

https://github.com/huggingface/transformers/pull/29109 changed the output shape of cos / sin generated by the rotary embedding, by considering the case of batched inference. This PR fixes it

cc @pacman100 @BenjaminBossan


pacman100 (approved):
Thank you @younesbelkada for the super quick fixes! ⚡

PR #1496, opened 2024-02-21
Raise error on wrong type for to modules_to_save

### LINKED ISSUE #1492: ModulesToSaveWrapper not working with ModulesDict dictionary methods
### System Info

transformers 4.37.2
peft 0.8.2
accelerate 0.27.2
torch  2.2.0

        

### Who can help?

_No response_

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

Error:
```
Exception has occurred: TypeError
'ModulesToSaveWrapper' object is not subscriptable
  File "/home/sam_work/code/tag_classification/test.py", line 14, in forward
    return self.foo_dict["foo"](torch.rand(1,10))
           ~~~~~~~~~~~~~^^^^^^^
  File "/home/sam_work/code/tag_classification/test.py", line 25, in <module>
    lora_model()
TypeError: 'ModulesToSaveWrapper' object is not subscriptable
```

```python
from peft import *
from torch import nn, Tensor
import torch


class ModelFoo(nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.other_layer = nn.Linear(10, 10)
        self.foo_dict = nn.ModuleDict({"foo": nn.Linear(10, 10)})

    def forward(self) -> Tensor:
        return self.foo_dict["foo"](torch.rand(1, 10))


model = ModelFoo()

peft_config = LoraConfig(
    target_modules=["other_layer"],
    modules_to_save=["foo_dict"],
)
model()
lora_model = get_peft_model(model=model, peft_config=peft_config)

lora_model()
```

### Expected behavior

The dictionary methods are promoted to wrapper - can use modulesdict without modifying code.

Seems like only forward is getting promoted to the wrapper and treating the modulesdict as a dictionary doesn't work. . Similar errors for items keys values indexing with key etc. Also same issues with iterating over ModulesList. Can look into working on it myself if there isn't some reason this hasn't been implemented,

BenjaminBossan:
Resolves #1492

This PR is for user convenience. When they try to pass the wrong type to `modules_to_save`, they will now get an early error message, instead of getting an obscure error when calling `forward` later.

Note:

The reason why `ModulesToSaveWrapper` cannot support `ModuleDict` et al. is because it tries to call forward on the original or copied module, but these modules don't implement a `forward` method.

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -178,6 +178,17 @@ class ModulesToSaveWrapper(torch.nn.Module):
         self._active_adapter = adapter_name
         self._disable_adapters = False
         self.update(adapter_name)
+        self.check_module()
+
+    def check_module(self):
+        """Perform some sanity checks on the module to ensure that it works"""
+        # Try to anticipate some modules that users could try to target that would not work.
+        # Note: It's not possible to check hasattr(module, "forward"), since that returns True for ModuleDict and
+        # ModuleList, even though their forward methods cannot be called
+        forbidden_classes = (torch.nn.ModuleDict, torch.nn.ModuleList, torch.nn.ParameterDict, torch.nn.ParameterList)
+        if isinstance(self.original_module, forbidden_classes):
+            cls_name = self.original_module.__class__.__name__
+            raise TypeError(f"modules_to_save cannot be applied to modules of type {cls_name}")
 
     @property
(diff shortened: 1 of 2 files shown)


younesbelkada (approved):
Very nice! Thanks @BenjaminBossan for working on this !

Issue #1485, opened 2024-02-19
all-linear + classification models have double-wrapped linear layers


nbroad1881:
### System Info

all

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

```python
m = AutoModelForTokenClassification.from_pretrained("microsoft/deberta-v3-xsmall", num_labels=10)

config = LoraConfig(
        r=16,
        lora_alpha=32,
        lora_dropout=0.1,
        task_type="TOKEN_CLS",
        target_modules="all-linear",
        bias="none",
    )

pmod = get_peft_model(m, config)

merged = pmod.merge_and_unload()
```

`merged` will still have lora layers in the classifier because it gets double-wrapped due to `"all-linear"`

### Expected behavior

Merged shouldn't have lora layers still in the classifier because calling `merged.save_pretrained` will result in uninitialized weights upon loading again

BenjaminBossan:
Thanks a lot Nicholas. Adding more context from our internal discussion: With the given `LoraConfig`, we get a PEFT model that has a `ModulesToSaveWrapper` on top of a `lora.Linear` layer:

```
      ...
      (classifier): ModulesToSaveWrapper(
        (original_module): lora.Linear(
          (base_layer): Linear(in_features=384, out_features=10, bias=True)
          (lora_dropout): ModuleDict(
            (default): Dropout(p=0.1, inplace=False)
          )
          (lora_A): ModuleDict(
            (default): Linear(in_features=384, out_features=16, bias=False)
          )
          (lora_B): ModuleDict(
            (default): Linear(in_features=16, out_features=10, bias=False)
          )
          (lora_embedding_A): ParameterDict()
          (lora_embedding_B): ParameterDict()
        )
        (modules_to_save): ModuleDict(
          (default): lora.Linear(
            (base_layer): Linear(in_features=384, out_features=10, bias=True)
            (lora_dropout): ModuleDict(
              (default): Dropout(p=0.1, inplace=False)
            )
            (lora_A): ModuleDict(
              (default): Linear(in_features=384, out_features=16, bias=False)
            )
            (lora_B): ModuleDict(
              (default): Linear(in_features=16, out_features=10, bias=False)
            )
            (lora_embedding_A): ParameterDict()
            (lora_embedding_B): ParameterDict()
          )
        )
      )
```

[...]

PR #1498, opened 2024-02-21
make `all-linear` as default for `target_modules`

pacman100:
### What does this PR do?
1. Make `all-linear` as default for `target_modules`. This should make LoRA and variants work for any new model out-of-the-box.

### DIFF
--- a/src/peft/tuners/loha/config.py
+++ b/src/peft/tuners/loha/config.py
@@ -42,7 +42,7 @@ class LoHaConfig(LycorisConfig):
             of the passed strings. If this is specified as 'all-linear', then all linear/Conv1D modules are chosen,
             excluding the output layer. If this is not specified, modules will be chosen according to the model
             architecture. If the architecture is not known, an error will be raised -- in this case, you should specify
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
Hmm, I'm not sure, this would break backwards compatibility, right? If we have a supported model architecture that does not target all linear layers but only a subset, using the `LoraConfig` defaults would suddenly target different layers. I think this would be very unexpected for users.

We could consider to leave `None` as default but to fall back on `all-linear` in case that the model architecture is not supported. But even then there can be problems. For instance, `all-linear` currently only works for transformers models. Also, as we saw in #1485, the function to detect all linear layers is still not quite robust. I thus don't think it's mature enough to be used as default.

My personal preference would be to leave the default the same. We could, however, improve the error message in case that the architecture is not supported by adding something like "When using a transformers model, consider setting `target_layers="all-linear"` to automatically add LoRA to all Linear layers".

pacman100:
Thank you benjamin, makes sense, closing the PR.

PR #1503, opened 2024-02-22
ENH: [`Docker`] Notify us when docker build pass or fail

younesbelkada:
As discussed offline cc @glegendre01

### DIFF
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: false
 
 env:
-  CI_SLACK_CHANNEL: "C06L7KU9WQ4"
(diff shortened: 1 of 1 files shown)


younesbelkada:
We could use re-usable workflows here cc @glegendre01

younesbelkada:
cc @glegendre01 @pacman100 @BenjaminBossan I tested out everything and seems to work great, this PR is ready for review!

pacman100 (approved):
Thank you Younes for adding this to alert when the Docker build fails!

BenjaminBossan (approved):
LGTM, thanks.

There is a lot of code duplication which could probably be avoided with a matrix build, but if it works as is, I'm fine.

younesbelkada:
Yeah I agree, we could also put it in a re-usable workflow. Will do it as a follow up PR !

PR #1505, opened 2024-02-23
FIX Safe merging with LoHa and LoKr

BenjaminBossan:
There was a small bug when merging the LoHa and LoKr tuners with `safe_merge=True` due to a missing `clone` call. This is now fixed.

Furthermore, the test coverage for merging with LoHa and LoKr has been extended, as there were a few tests where these methods were excluded unnecessarily.


younesbelkada (approved):
Makes sense, thanks !

PR #1512, opened 2024-02-27
Bump versions for release 0.9.0

BenjaminBossan:
Note that we did not set a dev version for 0.8.2, so this PR goes directly from 0.8.2 to 0.9.0.

No deprecated code or similar to change.


younesbelkada (approved):
Thanks  benjamin ! 🚀

PR #1516, opened 2024-02-28
Feat: add support for Conv2D DoRA

sayakpaul:
TODO

- [x] Incorporate `_apply_dora()` in the `forward()` of Conv2D
- [x] `merge()`
- [x] `unmerge()`
- [x] Tests
- [x] Docs

### DIFF
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -105,7 +105,7 @@ class LoraConfig(PeftConfig):
             Enable 'Weight-Decomposed Low-Rank Adaptation' (DoRA). This technique decomposes the updates of the weights
             into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the magnitude is
             handled by a separate learnable parameter. This can improve the performance of LoRA, especially at low
-            ranks. Right now, DoRA only supports non-quantized linear layers. DoRA introduces a bigger overhead than
+            ranks. Right now, DoRA only supports non-quantized linear and Conv2D layers. DoRA introduces a bigger overhead than
             pure LoRA, so it is recommended to merge weights for inference. For more information, see
             https://arxiv.org/abs/2402.09353.
     """
@@ -238,7 +238,7 @@ class LoraConfig(PeftConfig):
                 "Enable 'Weight-Decomposed Low-Rank Adaptation' (DoRA). This technique decomposes the updates of the "
                 "weights into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the "
                 "magnitude is handled by a separate learnable parameter. This can improve the performance of LoRA, "
-                "especially at low ranks. Right now, DoRA only supports non-quantized linear layers. DoRA introduces "
+                "especially at low ranks. Right now, DoRA only supports non-quantized linear and Conv2D layers. DoRA introduces "
                 "a bigger overhead than pure LoRA, so it is recommended to merge weights for inference. For more "
                 "information, see  https://arxiv.org/abs/2402.09353."
             )
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -105,8 +105,8 @@ class LoraConfig(PeftConfig):
             Enable 'Weight-Decomposed Low-Rank Adaptation' (DoRA). This technique decomposes the updates of the weights
             into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the magnitude is
             handled by a separate learnable parameter. This can improve the performance of LoRA, especially at low
-            ranks. Right now, DoRA only supports non-quantized linear and Conv2D layers. DoRA introduces a bigger overhead than
-            pure LoRA, so it is recommended to merge weights for inference. For more information, see
+            ranks. Right now, DoRA only supports non-quantized linear and Conv2D layers. DoRA introduces a bigger
+            overhead than pure LoRA, so it is recommended to merge weights for inference. For more information, see
             https://arxiv.org/abs/2402.09353.
     """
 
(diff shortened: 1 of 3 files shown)


sayakpaul:
@BenjaminBossan would love to get your thoughts on the changes I have introduced so far here.

BenjaminBossan:
Thanks a lot Sayak, this already looks pretty good. Probably the main thing to get right is the dimensions across which to normalize, which does look correct to me. Maybe there is also some room to speed up the application of the norm, not sure.

As a next step, let's add a test case for Conv2d + DoRA. This should be as easy as adding an example or two here:

https://github.com/huggingface/peft/blob/7e84dec20b3106bdd0a90ba8e80187f0aec835b7/tests/test_custom_models.py#L58-L70

Let's do this and check if the tests pass, then plan the next steps.

sayakpaul:
@BenjaminBossan I have added tests. However, 4 tests related to "merging" are currently failing. I would appreciate another round of review; maybe something pesky will come out that's making those tests fail. LMK.

sayakpaul:
@BenjaminBossan I think this is ready for a review now.

sayakpaul:
Cc: @nbasyl if you want to give this a look as well.

src/peft/tuners/lora/layer.py
```diff
@@ -797,6 +838,38 @@ def get_delta_weight(self, adapter) -> torch.Tensor:
 
         return output_tensor
 
+    def _get_weight_norm(self, weight, lora_weight, scaling) -> torch.Tensor:
+        # calculate L2 norm of weight matrix, channel-wise
+        weight = weight + scaling * lora_weight
+        weight_norm = weight.norm(p=2, dim=(1, 2, 3))
+        weight_norm = weight_norm.reshape((1, self.get_base_layer().out_channels, 1, 1))
```
BenjaminBossan, line 845: Would this work instead:

`weight_norm = weight.norm(p=2, dim=(1, 2, 3), keepdim=True).transpose(1, 0)`

src/peft/tuners/lora/layer.py
```diff
@@ -797,6 +838,38 @@ def get_delta_weight(self, adapter) -> torch.Tensor:
...
+        # see section 4.3 of DoRA (https://arxiv.org/abs/2402.09353)
+        # "[...] we suggest treating ||V +∆V ||_c in
+        # Eq. (5) as a constant, thereby detaching it from the gradient
+        # graph. This means that while ||V + ∆V ||_c dynamically
+        # reflects the updates of ∆V , it won’t receive any gradient
+        # during backpropagation"
+        weight_norm = weight_norm.detach()
+        mag_norm_scale = magnitude / weight_norm
+        result_dora = (mag_norm_scale - 1) * (
+            F.conv2d(x, weight, None, base_layer.stride, base_layer.padding, base_layer.dilation, base_layer.groups)
```
BenjaminBossan, line 868: Can we pass all args after `weight` as keyword args please?

tests/test_custom_models.py
```diff
@@ -646,7 +648,10 @@ def test_parameters_after_loading_model(self, test_name, model_id, config_cls, c
     @parameterized.expand(TEST_CASES)
     def test_disable_adapters(self, test_name, model_id, config_cls, config_kwargs):
         X = self.prepare_inputs_for_testing()
+        print(f"Class name: {self.transformers_class.__name__}")
         model = self.transformers_class.from_pretrained(model_id).to(self.torch_device).eval()
+        for k in X:
+            print(k, X[k].shape)
```
BenjaminBossan, line 654: Please remove.

BenjaminBossan (changes requested):
This looks pretty good, thanks a lot for enabling DoRA with Conv2d. Curious to see if we can get nice improvements for diffusion models.

Implementation-wise, I have only a few minor comments. On top of those, could you please document the new changes here:

- The docstring of `use_dora` inside of `tuners/lora/config.py` still mentions that only linear layers are supported.
- Same for the help of `use_dora` in the same file.
- `docs/source/developer_guide/lora.md` also needs updating with regard to the supported layers.

Before merging, it would be nice to have another review.

Edit: Please also merge with recent main, as there could be merge conflicts with #1518 otherwise.

src/peft/tuners/lora/layer.py
sayakpaul: Could work but I think the code is slightly more readable in the current form.

sayakpaul:
@BenjaminBossan resolved your comments. Thank you!

src/peft/tuners/lora/layer.py
BenjaminBossan: We can leave the reshape (though view is mostly used otherwise), but I think we should avoid the call to `self.get_base_layer().out_channels`, as we have all the information we need without it. Maybe a comment that documents that shapes could be added instead.

sayakpaul: Fixed in 641c5b5.

src/peft/tuners/lora/config.py
```diff
@@ -104,7 +104,10 @@ class LoraConfig(PeftConfig):
         use_dora (`bool`):
             Enable 'Weight-Decomposed Low-Rank Adaptation' (DoRA). This technique decomposes the updates of the weights
             into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the magnitude is
-            handled by a separate learnable parameter. This can improve the performance of LoRA, especially at low
+            handled by a separate learnable parameter. This can improve the performance of LoRA especially at low
+            ranks. Right now, DoRA only supports linear and Conv2D layers. DoRA introduces a bigger overhead than pure
```
BenjaminBossan, line 108: Oops, I must have broken this with a merge resolution. Thanks for fixing!

BenjaminBossan (approved):
Thanks a lot, Sayak, the PR looks great. Let's see if Shih-yang also has time to take a look. Otherwise, I'd be glad to have a second pair of eyes by @younesbelkada or @pacman100.

BenjaminBossan:
Thanks so much Sayak.

PR #1518, opened 2024-02-29
QDoRA: Support DoRA with BnB quantization

BenjaminBossan:
Adds support for DoRA on 4bit and 8bit quantized models with bitsandbytes. Merging also works, with the usual caveats for quantized weights (results are not 100% identical), but it's not worse than vanialla LoRA.

I did some quick tests and could see the expected memory savings with bnb. Same as with DoRA on non-quantized layers, using DoRA on quantized layers leads to a moderate increase in runtime.

### DIFF
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -71,7 +71,7 @@ config = LoraConfig(use_rslora=True, ...)
 
 ### Weight-Decomposed Low-Rank Adaptation (DoRA)
 
-This technique decomposes the updates of the weights into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the magnitude is handled by a separate learnable parameter. This can improve the performance of LoRA, especially at low ranks. Right now, DoRA only supports linear layers. DoRA introduces a bigger overhead than pure LoRA, so it is recommended to merge weights for inference, see [`LoraModel.merge_and_unload`]. For more information on DoRA, see  https://arxiv.org/abs/2402.09353.
+This technique decomposes the updates of the weights into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the magnitude is handled by a separate learnable parameter. This can improve the performance of LoRA, especially at low ranks. For more information on DoRA, see  https://arxiv.org/abs/2402.09353.
 
 ```py
 from peft import LoraConfig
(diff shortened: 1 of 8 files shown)


younesbelkada (approved):
Looks great to me thank you @BenjaminBossan for adding quantization support for DoRA !

docs/source/developer_guides/lora.md
```diff
@@ -71,14 +71,16 @@ config = LoraConfig(use_rslora=True, ...)
...
+This technique decomposes the updates of the weights into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the magnitude is handled by a separate learnable parameter. This can improve the performance of LoRA, especially at low ranks. Right now, DoRA only supports linear layers. DoRA introduces a bigger overhead than pure LoRA, so it is recommended to merge weights for inference, see [`LoraModel.merge_and_unload`]. For more information on DoRA, see  https://arxiv.org/abs/2402.09353.
 
 ```py
 from peft import LoraConfig
 
 config = LoraConfig(use_dora=True, ...)
 ```
 
+DoRA should work with weights quantized with bitsandbytes ("QDoRA"). Issues have been reported when using QDoRA with DeepSpeed Zero2.
```
pacman100, line 82: this should be in a `Caveats` section wherein all such notes can be collated in one place.

pacman100 (approved):
Thank you @BenjaminBossan for adding support to use DoRA with bnb quantized layers and the thorough tests! 🤗

docs/source/developer_guides/lora.md
BenjaminBossan: I added a caveats section directly below, within the DoRA section. Is this what you had in mind?

pacman100 (approved):
Thank you @BenjaminBossan!  ✨

mallorbc:
Qdora seems to be working for me.  However, I am noticing a large slowdown comparing Qlora and Qdora of around 2x.  I am not sure if that is expected or not.  But this seems like a good place as any to share this finding.

BenjaminBossan:
> I am noticing a large slowdown comparing Qlora and Qdora of around 2x. I am not sure if that is expected or not.

Yes, QDoRA unfortunately requires an additional dequantization step on the quantized weights to calculate the weight norm. I wouldn't expect this to slow down training by 2x, but a significant slowdown is expected. Maybe you can run some profiler to check further if you think it's worth investigating.

> But this seems like a good place as any to share this finding.

You can also create new issues or discussions (if there aren't already existing ones) for this type of question.

PR #1519, opened 2024-03-01
fix: fail when required args not passed when prompt_tuning_init==TEXT

kmehant:
When `prompt_tuning_init==TEXT` is set, there are some args needed to be provided. For instance, `tokenizer_name_or_path` should be set to some tokenizer name or path, failing which would result in errors such as

```
HTTPError: 401 Client Error: Unauthorized for url: https://huggingface.co/None/resolve/main/tokenizer_config.json
```
tokenizer_name_or_path is taken as `None`. Such errors can be handled and reported meaningfully post config initialization and this PR makes that contribution.

### DIFF
--- a/src/peft/tuners/prompt_tuning/config.py
+++ b/src/peft/tuners/prompt_tuning/config.py
@@ -75,9 +75,9 @@ class PromptTuningConfig(PromptLearningConfig):
                 f"When prompt_tuning_init='{PromptTuningInit.TEXT.value}', "
                 f"tokenizer_name_or_path can't be {self.tokenizer_name_or_path}."
             )
-        if (self.prompt_tuning_init == PromptTuningInit.TEXT) and not self.prompt_tuning_init_text:
+        if (self.prompt_tuning_init == PromptTuningInit.TEXT) and self.prompt_tuning_init_text is None:
             raise ValueError(
-                f"when prompt_tuning_init='{PromptTuningInit.TEXT.value}', "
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Thanks for adding this quality of life feature. Two points:

- Could you please break the long lines? We want to keep a 120 char limit.
- Would you mind adding tests to check that the error is raised? We can help with that if needed.

kmehant:
@BenjaminBossan 

> Would you mind adding tests to check that the error is raised? We can help with that if needed.


Sure @BenjaminBossan, I tried looking for tests for prompt tuning implementation, but I could not find them, can you point me to that?

BenjaminBossan:
> I tried looking for tests for prompt tuning implementation, but I could not find them, can you point me to that?

The tests are a bit scattered around, as we group by model architecture and not by method. But a good fit would be to add them to `tests/test_initialization.py`.

kmehant:
@BenjaminBossan Have added the tests and made required changes.

Screenshot 2024-03-01 at 4 35 15 PM

BenjaminBossan:
Thanks a lot for adding the test, could you please run `make style` and push the changes?

kmehant:
https://github.com/huggingface/peft/pull/1519#issuecomment-1973305566

@BenjaminBossan I have run `make style` and pushed the changes thanks.

BenjaminBossan (changes requested):
Thanks for the update. I found some very small things still, after that we should be good to go.

PR #1521, opened 2024-03-02
Use peft_doc as notebook path as doc-builder expects <package>_doc for the Google Colab and AWS Studio links

DriesVerachtert:
At the moment, the peft notebooks are generated in a path containing 'peft_docs', with the 's' at the end. This results in broken links in the online documentation.

For example: https://huggingface.co/docs/peft/main/en/task_guides/clm-prompt-tuning : the link to the pytorch version of the google colab notebook is https://colab.research.google.com/github/huggingface/notebooks/blob/main/peft_doc/en/pytorch/clm-prompt-tuning.ipynb , which uses 'peft_doc' without the trailing 's'.

The 'resolv_open_in_colab' method in doc-builder has the path <package>_doc hardcoded at https://github.com/huggingface/doc-builder/blob/3543e0aca2029201051373e3b6ec871ef44d0d93/src/doc_builder/build_doc.py#L52 . This means it always uses a 'peft_doc' path for peft when generating the google colab and aws studio links. There's even a test which checks that the links use _doc for another package at https://github.com/huggingface/doc-builder/blob/main/tests/test_build_doc.py

It's likely the easiest to simply use peft_doc instead of peft_docs so that the automatically generated links are correct.

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -13,7 +13,7 @@ jobs:
     with:
       commit_sha: ${{ github.sha }}
       package: peft
-      notebook_folder: peft_docs
+      notebook_folder: peft_doc
     secrets:
       token: ${{ secrets.HUGGINGFACE_PUSH }}
       hf_token: ${{ secrets.HF_DOC_BUILD_PUSH }}


DriesVerachtert:
This would fix https://github.com/huggingface/notebooks/issues/417

.github/workflows/build_documentation.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     with:
       commit_sha: ${{ github.sha }}
       package: peft
-      notebook_folder: peft_docs
+      notebook_folder: peft_doc
```
stevhliu, line 16: Ok for me to try this to remain consistent with the other libraries.

However, I'm not sure if this'll fix it. For example, Diffusers is also set up to `diffusers_doc` but when you click on a `Open in Colab` button you'll get the same error because either the doc wasn't converted to a notebook or it failed to push it to the https://github.com/huggingface/notebooks repo.

cc @mishig25 @coyotte508

DriesVerachtert: I'm definitely not 100% sure if it will fix the issue or not :-) I've just been trying to understand all the github CI/CD code and this is what I would try. Perfectly fine to revert this commit if it doesn't fix the issue of course.

DriesVerachtert: @stevhliu could you maybe give an example of such a diffuser documentation page which has a broken link to the google Colab? I tried some random ones and they worked for me. If you could provide an example broken link, that would be great.

stevhliu: Sure, thanks for trying to troubleshoot the issue as well! ❤️ 

* this [Inference with PEFT](https://colab.research.google.com/github/huggingface/notebooks/blob/main/diffusers_doc/en/pytorch/using_peft_for_inference.ipynb) doc from Diffusers was added ~5 months ago
* this [Generation with LLMs](https://colab.research.google.com/github/huggingface/notebooks/blob/main/transformers_doc/en/pytorch/llm_tutorial.ipynb) tutorial from Transformers was added ~7 months ago

BenjaminBossan:
@DriesVerachtert @stevhliu what is the state of this PR?

stevhliu:
I'm not sure this PR will fix the issue, and it is being internally investigated by @mishig25

PR #1527, opened 2024-03-04
Optimize levenshtein_distance algorithm in peft_lora_seq2seq_accelera…

SUNGOD3:
This commit refines the levenshtein_distance algorithm implemented in peft_lora_seq2seq_accelerate_ds_zero3_offload.py to improve its space complexity from O(n^2) to O(n). Additionally, thorough testing has been conducted to ensure the correctness and reliability of the revised implementation.

### DIFF
--- a/examples/causal_language_modeling/peft_lora_clm_accelerate_ds_zero3_offload.py
+++ b/examples/causal_language_modeling/peft_lora_clm_accelerate_ds_zero3_offload.py
@@ -28,7 +28,7 @@ def levenshtein_distance(str1, str2):
         return 0
     num_rows = len(str1) + 1
     num_cols = len(str2) + 1
-    dp_matrix = [i for i in range(num_cols)]
+    dp_matrix = list(range(num_cols))
     for i in range(1, num_rows):
         prev = dp_matrix[0]
         dp_matrix[0] = i
--- a/examples/causal_language_modeling/peft_lora_clm_accelerate_ds_zero3_offload.py
+++ b/examples/causal_language_modeling/peft_lora_clm_accelerate_ds_zero3_offload.py
@@ -3,7 +3,6 @@ import os
 import sys
 import threading
 
-import numpy as np
 import psutil
 import torch
 from accelerate import Accelerator
--- a/examples/causal_language_modeling/peft_lora_clm_accelerate_ds_zero3_offload.py
+++ b/examples/causal_language_modeling/peft_lora_clm_accelerate_ds_zero3_offload.py
@@ -23,23 +23,23 @@ from peft import LoraConfig, TaskType, get_peft_model
 
 def levenshtein_distance(str1, str2):
     # TC: O(N^2)
-    # SC: O(N^2)
+    # SC: O(N)
     if str1 == str2:
         return 0
     num_rows = len(str1) + 1
     num_cols = len(str2) + 1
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Hi, thanks for providing a more efficient implementation. From what I can tell, this should have very marginal effects on this script's runtime, if any, given how little time is spent on this vs training the net. Do you think it's really worth it to make this update?

SUNGOD3:
Hi BenjaminBossan,
Thanks for your feedback. As you mentioned, the improvement in the levenshtein_distance algorithm has minimal impact on the overall effectiveness. In fact, I stumbled upon this file while tracing a [bug](https://github.com/huggingface/transformers/issues/29047). However, based on my experience, state compression in dynamic programming (DP) typically has minimal side effects. In comparison to the original code, it generally performs better in terms of both time and space complexity in nearly all cases. Furthermore, it hardly compromises the readability of the code. So, why not go for it?

BenjaminBossan:
In general, I agree with your argument. From a maintainer's point of view, I'm just a bit reluctant because:

1. Is there an edge case where this returns different results? Leading to a user later opening an issue asking us why their results changed after updating PEFT.
2. What was the intent when this function was originally added. Is it maybe a 1:1 copy of somewhere else and is purposefully like this?

Those are not big issues, but then again the advantage is also minimal, which is why I'm hesitating.

SUNGOD3:
I understand, your hesitation is entirely reasonable. However, perhaps the following analysis might change your perspective?

Firstly, regarding the first point, the probability of encountering edge test cases resulting in different outcomes is exceedingly low. This is simply a function of a classic problem. Furthermore, under the condition where both input and output types remain unchanged, this update not only passed the 1146 test cases on [Leetcode](https://leetcode.com/problems/edit-distance/) but also passed my scripted tests. (Contains special characters, long strings, empty strings, special arrangements)

image


As for the second point, I've gained some understanding of the context in which this function is utilized. It serves as a classification algorithm to predict which class the model's output belongs to. The source code might have been written by the original author themselves, as I couldn't find any similar implementations in the relevant documentation.

Additionally, it's worth mentioning that the impact of this optimization on time might not be negligible. The reason it seems to have minimal effect currently is simply because during testing, only a few classes (just 3) were utilized, and the names of these classes were also very short (less than 10 characters). In reality, as the number and length of classes increase, this improvement could easily exceed 1 second in overall acceleration and scales linearly with the number of classes and output length. Moreover, when the tested model becomes smaller or the training epoches decreases, its overall impact will gradually become apparent.

https://colab.research.google.com/drive/1JUznpk2OAGCu5ZVAddLtKdPKyGKMp2ed?usp=sharing

SUNGOD3:
It seems that there should be no problem after the last commit? (Unit test has also been updated)
Screenshot 2024-03-06 at 9 21 42 PM

BenjaminBossan:
Strange that it passes for you, CI shows this error:

```
examples/causal_language_modeling/peft_lora_clm_accelerate_ds_zero3_offload.py:6:17: F401 [*] `numpy` imported but unused
examples/conditional_generation/peft_lora_seq2seq_accelerate_ds_zero3_offload.py:6:17: F401 [*] `numpy` imported but unused
```

SUNGOD3:
Maybe make quality is something I should run too? 
BTW, I seem to get no errors after executing make style repeatedly. Maybe this is a bug?

Issue #1524, opened 2024-03-04
`lora.Linear.bias` should point to `lora.Linear.base_layer.bias` as the `lora.Linear.weight` does


ArthurZucker:
### System Info

- `transformers` version: 4.39.0.dev0
- Platform: Linux-5.4.0-166-generic-x86_64-with-glibc2.31
- Python version: 3.9.17
- Huggingface_hub version: 0.21.1
- Safetensors version: 0.4.2
- Accelerate version: 0.27.2
- Accelerate config:    not found
- PyTorch version (GPU?): 2.2.0+cu121 (True)
- Tensorflow version (GPU?): 2.13.0 (True)
- Flax version (CPU?/GPU?/TPU?): 0.7.0 (cpu)
- Jax version: 0.4.13
- JaxLib version: 0.4.13
- Using GPU in script?: <fill in>
- Using distributed or parallel set-up in script?: <fill in>
- Peft version: 0.9.0
- Trl version: 0.7.11

### Who can help?

@pacman100 @BenjaminBossan 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

I am running the following: 

```python
from datasets import load_dataset
from trl import SFTTrainer
from peft import LoraConfig
from transformers import AutoTokenizer, AutoModelForCausalLM, TrainingArguments
import torch

model_id = "ArthurZ/mamba-130m"
tokenizer = AutoTokenizer.from_pretrained(model_id, pad_token ="<|padding|>")
model = AutoModelForCausalLM.from_pretrained(model_id, pad_token_id = 1, torch_dtype = torch.float32, residual_in_fp32 =True, device_map = {"":1})

dataset = load_dataset("Abirate/english_quotes", split="train")


training_args = TrainingArguments(
    output_dir="./results",
    num_train_epochs=5,
    per_device_train_batch_size=4,
    logging_dir='./logs',
    logging_steps=10,
    learning_rate=1e-3
)
lora_config =  LoraConfig(
        r=8,
        target_modules=["x_proj", "in_proj", "embeddings"],
        task_type="CAUSAL_LM",
        bias="none"
)
lora_config = None
trainer = SFTTrainer(
    model=model,
    tokenizer=tokenizer,
    args=training_args,
    peft_config=lora_config,
    train_dataset=dataset,
    dataset_text_field="quote",
)

trainer.train()
```

[...]

PR #1530, opened 2024-03-04
Expose bias attribute on tuner layers

BenjaminBossan:
See #1524

This is similar to how we already expose the weight attribute.


younesbelkada (approved):
Makes sense, thanks !

ArthurZucker:
Thanks a lot for the prompt fix both! 🤗

StarCycle:
Great! This PR just solved my problem!!!

Issue #1526, opened 2024-03-04
Inconsistency between get_nb_trainable_parameters and num_parameters(only_trainable=True) for prompt tuning


kmehant:
### System Info

peft==0.8.2
transformers==4.37.2

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

```python
from peft import get_peft_model, PromptTuningConfig

from transformers import AutoConfig, AutoModelForCausalLM

model = AutoModelForCausalLM.from_config(AutoConfig.from_pretrained("bigcode/starcoderbase-1b"))

pt_config = PromptTuningConfig(
    peft_type="PROMPT_TUNING",
    task_type="CAUSAL_LM",
    num_virtual_tokens=20,
    prompt_tuning_init="TEXT",
    prompt_tuning_init_text="some prompt tuning init text",
    tokenizer_name_or_path="path to tokenizer",
)

model_pt = get_peft_model(model, pt_config)

model_pt.get_nb_trainable_parameters() # output 1 (first element in the tuple)
model_pt.num_parameters(only_trainable=True) # output 2

# output 1 does not match output 2

```

### Expected behavior

[UPDATED]

we should have trainable paramters > 0 this is consistent with LoRA. For CodeLlama 7B I get

```
model_lora.get_nb_trainable_parameters() # (8388608, 6746935296)

model_lora.num_parameters(only_trainable=True) # 8388608
```

However, for promptuning, 

```
model_lora.get_nb_trainable_parameters() # (81920, 6738628608)

model_lora.num_parameters(only_trainable=True) # 0
```

I agree with the outputs from `get_nb_trainable_parameters()` However, trying to understand this inconsistent behaviour of `num_parameters(only_trainable=True)` for LoRA and prompt tuning techniques.

BenjaminBossan:
When I tried, I got 40960 and 0 parameters, respectively. But that isn't too surprising. The `num_parameters` method is not defined in the `PeftModel` (i.e. the thing you get back after calling `get_peft_model`). Therefore, the method call is delegated to the base model itself, i.e. the starcoderbase. As the base model is frozen, we get back 0. As for the `PeftModel`, it adds new trainable parameters, or else fine-tuning wouldn't work, so that's why we get a number > 0 back.

kmehant:
[...]

PR #1531, opened 2024-03-04
docs: highlight difference between num_parameters() and get_nb_trainable_parameters() in PEFT

kmehant:
related to #1526

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -516,6 +516,13 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
     def print_trainable_parameters(self) -> None:
         """
         Prints the number of trainable parameters in the model.
+
+        Note: print_trainable_parameters() uses get_nb_trainable_parameters() which is different from
+        num_parameters(only_trainable=True) from huggingface/transformers. get_nb_trainable_parameters() returns
+        (trainable parameters, all parameters) of the Peft Model which includes modified backbone transformer model.
+        For techniques like LoRA, the backbone transformer model is modified in place with LoRA modules. However, for
+        prompt tuning, the backbone transformer model is unmodified. num_parameters(only_trainable=True) returns number
+        of trainable parameters of the backbone transformer model which can be different.
         """
         trainable_params, all_param = self.get_nb_trainable_parameters()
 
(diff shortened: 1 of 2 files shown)


src/peft/peft_model.py
```diff
@@ -515,7 +515,12 @@ def get_nb_trainable_parameters(self) -> tuple[int, int]:
 
     def print_trainable_parameters(self) -> None:
         """
-        Prints the number of trainable parameters in the model.
+        Prints the number of trainable parameters in the model. Note: print_trainable_parameters() uses
```
BenjaminBossan, line 518: Could you please add 1 empty line after the first sentence (before "Note")?

src/peft/peft_model.py
```diff
@@ -515,7 +515,12 @@ def get_nb_trainable_parameters(self) -> tuple[int, int]:
...
+        get_nb_trainable_parameters() which is different from num_parameters(only_trainable=True).
```
BenjaminBossan, line 519: Let's mention that `num_parameters` is a method from transformers.

src/peft/peft_model.py
```diff
@@ -515,7 +515,12 @@ def get_nb_trainable_parameters(self) -> tuple[int, int]:
...
+        get_nb_trainable_parameters() returns (trainable parameters, all paramters) of the Peft Model which includes
```
BenjaminBossan, line 520: paramters -> parameters

src/peft/peft_model.py
```diff
@@ -515,7 +515,12 @@ def get_nb_trainable_parameters(self) -> tuple[int, int]:
...
+        modified backbone transformer model. For techniques like LoRA, the backbone transformer model is modified in
+        place with LoRA modules. However, for prompt tuning, the backbone transformer model is unmodified.
+        num_parameters(only_trainable=True) returns number of trainable parameters of the backbone transformer model.
```
BenjaminBossan, line 523: Just to make it totally clear, let's add here that this can result in a different number :)

BenjaminBossan (changes requested):
Thanks for adding the docstrings. I made some comments where I think it can be improved. Please check them out.

BenjaminBossan (approved):
Thanks for adding this informative docstring!

Issue #1525, opened 2024-03-04
LoftQ does not seem to quantify the base model


Mr-KenLee:
### System Info

transformers version: 4.37.2
Platform: Ubuntu 18.04.6 LTS
GPU: RTX GeForce 3090 x 2
Python version: 3.10.13
Huggingface_hub version: 0.20.3
Safetensors version: 0.4.2
Accelerate version: 0.26.1
Accelerate config: not found
PyTorch version (GPU?): 2.2.1+cu121 (True)
Tensorflow version (GPU?):  not found
Flax version (CPU?/GPU?/TPU?):  not found
Jax version:  not found
JaxLib version:  not found
Using GPU in script?: yes
Using distributed or parallel set-up in script?: no
Peft version: 0.9.0
Trl version: 0.7.11

### Who can help?

@pacman100 @stevhliu

### Information

- [X] The official example scripts
- [X] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder
- [X] My own task or dataset (give details below)

### Reproduction

I am attempting to fine-tune my model (Baichuan2-chat-7B) using LoftQ, but the results seem to differ from my expectations. Due to the computational resources available in our lab, I am using two 3090 GPUs for fine-tuning. I followed the method described at https://huggingface.co/docs/peft/en/developer_guides/lora#loftq:

```python
from transformers import AutoModelForCausalLM, BitsAndBytesConfig
from peft import LoftQConfig, LoraConfig, get_peft_model

base_model = AutoModelForCausalLM.from_pretrained("Baichuan2-7B-Chat", trust_remote_code=True, device_map="auto", torch_dtype=torch.bfloat16)  # don't quantize here
loftq_config = LoftQConfig(loftq_bits=4)           # set 4bit quantization
lora_config = LoraConfig(
    init_lora_weights="loftq",
    loftq_config=loftq_config,
    r=16,
    lora_alpha=8,
    target_modules=["W_pack"],
    lora_dropout=0.05,
    bias="none",
    task_type="CAUSAL_LM"
)
peft_model = get_peft_model(base_model, lora_config)
```

However, I have found that LoftQ does not seem to quantize my LLM as expected. The actual usage of GPU memory is roughly the same as when directly loading the model (about 8GB on each GPU), with only an additional 2GB for LoRA used.

Could you please help me understand why this is happening? Is there something I am doing incorrectly?

Additionally, there is a discussion thread with a similar issue encountered by someone using the Llama-2-7b model:
[...]

PR #1532, opened 2024-03-04
Fix LoftQ docs and tests

BenjaminBossan:
Relates to #1525

_Don't merge this, some GPU tests are failing_

Unfortunately, the docs I wrote about how to use LoftQ were incorrect, based on a misunderstanding I had. In reality, it is quite a bit more involved to get LoftQ working, requiring a complete roundtrip first loading a non-quantized model with LoftQ, saving the LoRA weights and the modified base model, loading the just stored base model again but this time with quantization, and finally loading the LoftQ-initialized adapter on top. The docs now link to the example which demonstrates how to move through these steps.

The unit tests have been adjusted to go through these same steps but now most of them fail, i.e. the quantization error is _greater_ with LoftQ than without. This needs to be investigated.

### DIFF
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -42,16 +42,7 @@ config = LoraConfig(init_lora_weights=False, ...)
 
 ### LoftQ
 
-When quantizing the base model for QLoRA training, consider using the [LoftQ initialization](https://arxiv.org/abs/2310.08659), which has been shown to improve performance when training quantized models. The idea is that the LoRA weights are initialized such that the quantization error is minimized. If you're using LoftQ, *do not* quantize the base model. You should set up a [`LoftQConfig`] instead:
-
-```python
-from peft import LoftQConfig, LoraConfig, get_peft_model
-
-base_model = AutoModelForCausalLM.from_pretrained(...)  # don't quantize here
-loftq_config = LoftQConfig(loftq_bits=4, ...)           # set 4bit quantization
-lora_config = LoraConfig(..., init_lora_weights="loftq", loftq_config=loftq_config)
-peft_model = get_peft_model(base_model, lora_config)
-```
+When quantizing the base model for QLoRA training, consider using the [LoftQ initialization](https://arxiv.org/abs/2310.08659), which has been shown to improve performance when training quantized models. The idea is that the LoRA weights are initialized such that the quantization error is minimized. To use LoftQ, follow [these instructions](https://github.com/huggingface/peft/tree/main/examples/loftq_finetuning).
 
 <Tip>
 
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -44,6 +44,8 @@ config = LoraConfig(init_lora_weights=False, ...)
 
 When quantizing the base model for QLoRA training, consider using the [LoftQ initialization](https://arxiv.org/abs/2310.08659), which has been shown to improve performance when training quantized models. The idea is that the LoRA weights are initialized such that the quantization error is minimized. To use LoftQ, follow [these instructions](https://github.com/huggingface/peft/tree/main/examples/loftq_finetuning).
 
+In general, for LoftQ to work best, it is recommended to target as many layers with LoRA as possible, since those not targeted cannot have LoftQ applied. This means that passing `LoraConfig(..., target_modules="all-linear")` will most likely give the best results. Also, you should use `nf4` as quant type in your quantization config when using 4bit quantization, i.e. `BitsAndBytesConfig(load_in_4bit=True, bnb_4bit_quant_type="nf4")`.
+
 <Tip>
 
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
@yxli2123 As described above, I made a mistake in how LoftQ needs to be used. This PR adjust the docs and the unit tests to apply LoftQ correctly (from my understanding).

However, after I changed the unit tests, 8 of them started failing, only 4 passed:

```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃  File                        ┃  Function                                       ┃  Function Line  ┃  Error Line  ┃  Error           ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│  tests/test_gpu_examples.py  │  TestLoftQ.test_bloomz_loftq_4bit[cuda]         │  1244           │  1260        │  AssertionError  │
│  tests/test_gpu_examples.py  │  TestLoftQ.test_bloomz_loftq_4bit[cpu]          │  1244           │  1260        │  AssertionError  │
│  tests/test_gpu_examples.py  │  TestLoftQ.test_bloomz_loftq_4bit_iter_5[cuda]  │  1263           │  1275        │  AssertionError  │
│  tests/test_gpu_examples.py  │  TestLoftQ.test_bloomz_loftq_4bit_iter_5[cpu]   │  1263           │  1275        │  AssertionError  │
│  tests/test_gpu_examples.py  │  TestLoftQ.test_bloomz_loftq_8bit_iter_5[cuda]  │  1293           │  1305        │  AssertionError  │
│  tests/test_gpu_examples.py  │  TestLoftQ.test_bloomz_loftq_8bit_iter_5[cpu]   │  1293           │  1305        │  AssertionError  │
│  tests/test_gpu_examples.py  │  TestLoftQ.test_t5_loftq_4bit[cuda]             │  1308           │  1320        │  AssertionError  │
│  tests/test_gpu_examples.py  │  TestLoftQ.test_t5_loftq_4bit[cpu]              │  1308           │  1320        │  AssertionError  │
└──────────────────────────────┴─────────────────────────────────────────────────┴─────────────────┴──────────────┴──────────────────┘
```

In these unit tests, I measure the model output from the base model without any quantization, then the model output of the bnb-quantized model, and finally the model output of the quantized model using LoftQ. The expectation is that the model output with LoftQ should be closer to the base model output than the bnb-quantized model output (without LoftQ). 

(Note that I'm only testing the LoftQ should lead to closer results, even if it's only marginally closer. Ideally, however, we want it be X% better, i.e. there should be a certain margin)

I'm not sure why the error with LoftQ is greater and not smaller. Am I still applying LoftQ incorrectly? Please check the unit test if I make some mistake there. For debugging purposes, I added a check that on each individual layer, the [residual error](https://github.com/huggingface/peft/blob/34f3fba2b3f76d62a55491524301baca83dbc22b/src/peft/utils/loftq_utils.py#L220) does indeed decrease when applying the LoRA weights initialized with LoftQ, so this part seems to be correct.

Moreover, while working on this, some questions came up:

1. Why do we need to override the weights with the dequantized weights ([here](https://github.com/huggingface/peft/blob/34f3fba2b3f76d62a55491524301baca83dbc22b/src/peft/tuners/lora/layer.py#L170)). Why can we not just load the quantized base model with the LoftQ-LoRA weights, i.e. skip [this step](https://github.com/huggingface/peft/blob/34f3fba2b3f76d62a55491524301baca83dbc22b/examples/loftq_finetuning/quantize_save_load.py#L176-L177)? I tried that as well but test are failing no matter what.
2. Is it really working correctly when we pass `num_iter > 1`? I feel like [this line](https://github.com/huggingface/peft/blob/34f3fba2b3f76d62a55491524301baca83dbc22b/src/peft/utils/loftq_utils.py#L225) is not correct, because we are iteratively changing the residual but `lora_A` and `lora_B` are only determined by the very last step. Maybe those need to be incremented iteratively as well?

pacman100 (commented):
Thank you Benjamin. After going through the tests, the examples for LoftQ. I believe that the discrepancy and failing tests happen due to the casting of activation to `float16` by bitsandbytes when loading the loftq base and adapters.

pacman100:
So, the difference between original weight W and quantized+adapters weights (Q+BA) would be less when compared to just using Q, it isn't reflecting in logits because activations at each layer are being casted to `float16` by bnb.

pacman100 (approved):
Thank you @BenjaminBossan for all the work on rectifying LoftQ tests and uncovering best practices for applying it!

Issue #1462, opened 2024-02-14
Ability to add new tokens and keep existing token embeddings frozen?


DreamGenX:
### Feature request

Today, when you add new tokens to the vocabulary (e.g. `<|im_start|>` and `<|im_end|>`), you need to also add `embed_tokens` and `lm_head` to the `modules_to_save` kwarg. This, as far as I can tell, unfreezes all token embeddings.

Would it be possible to unfreeze only the newly added tokens?

### Motivation

This would allow us to expand the vocabulary, and not fine tune too many extra params.

### Your contribution

Seeking feedback

BenjaminBossan:
> you need to also add `embed_tokens` and `lm_head` to the `tokens_to_save` kwarg

I assume you mean `modules_to_save` and not `tokens_to_save`. Indeed, this makes the whole embedding layer learnable. AFAIK, there is no easy way in PyTorch to make only a part of the parameters learnable and freeze the rest. This would probably require a custom module class to work.

A possible solution to avoid having to update the possibly huge embedding weights is to apply LoRA (assuming you use LoRA) to that layer. More on this can be found in the docs [here](https://huggingface.co/docs/peft/developer_guides/troubleshooting#extending-the-vocabulary).

DreamGenX:
> I assume you mean `modules_to_save` and not `tokens_to_save`

Oh yes, updated.

> A possible solution to avoid having to update the possibly huge embedding weights is to apply LoRA (assuming you use LoRA) to that layer.

Yeah, that would certainly help reduce the parameter count.

marcusinthesky:
I wonder if this could be done, using [sparse matrices](https://pytorch.org/docs/stable/sparse.html#supported-operations)? You could initial LoRA's A matrix as a sparse matrix, or `modules_to_save` as a sparse matrix. My only concern would by [safetensor's support](https://github.com/huggingface/safetensors/blob/b947b59079a6197d7930dfb535818ac4896113e8/bindings/python/py_src/safetensors/torch.py#L397).

BenjaminBossan:
@marcusinthesky I'm not sure how sparse matrices would help here, could you please elaborate? LoRA weights are already sparse in the sense that they're a low rank approximation of a much bigger matrix. When it comes to `modules_to_save`, we create a copy of the original module, so the parameter type has to be the same.

marcusinthesky:
[...]

PR #1541, opened 2024-03-06
CUSTOM_TOKEN Tuner

marcusinthesky:
We may want to train just a subset of our token's while leaving the rest frozen. This may be done when we have token's for named entities, or special tokens for chat assistants. While you can train only certain token's by registering a custom gradient hook, to 'zero-out' the frozen token embedding, this still requires us to store in memory and on-disk a large embedding matrix, which can take up ~30% of the total model size. A more efficient approach, would be to wrap embedding weight deltas for trainable token's in a Sparse Matrix, making merging, unmerging and delta operations fast. 

Please see https://github.com/huggingface/peft/issues/1462. 

ToDo:

- [ ] Tests
- [ ] Documentation
- [ ] Examples

### DIFF
--- a/src/peft/tuners/custom_tokens/config.py
+++ b/src/peft/tuners/custom_tokens/config.py
@@ -1,0 +1,22 @@
+from dataclasses import asdict, dataclass, field
+from typing import List, Optional, Union
+from peft import PeftConfig
+from peft.utils import PeftType
+
+
+@dataclass
+class CustomTokensConfig(PeftConfig):
+    token_indices: List[int] = field(default_factory=list)
+    target_modules: Optional[Union[list[str], str]] = field(
+        default='embedding',
+        metadata={
+            "help": (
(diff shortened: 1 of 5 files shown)


BenjaminBossan:
Thanks for this PR. I have just skimmed it so far. Could you give an example of how to use this adapter? How would we combine it with, say, LoRA?

If you have specific questions regarding tests and documentation, feel free to ask me.

BenjaminBossan:
@marcusinthesky Do you still plan on working on this?

marcusinthesky:
Yes. I do. It may be two weeks.

marcusinthesky:
@ds5423 thanks. I got fairly far, but did not have time for tests, docs, examples and save/load config. 

I do think this is the optimal approach, which would really benefit the community. Would love to hear how it works out for you?

PR #1543, opened 2024-03-07
More convenient way to initialize LoftQ

BenjaminBossan:
Related to #1532

At the moment, using LoftQ is quite cumbersome, as shown in this example:

https://github.com/huggingface/peft/tree/7e84dec20b3106bdd0a90ba8e80187f0aec835b7/examples/loftq_finetuning

Essentially, users have to:

1. Load the non-quantized model with LoftQ (which can be quite huge)
2. Modify the PEFT config
3. Save the adapter
4. Unwrap the base model with custom functions
5. Save the base model with modified weights (i.e. a whole copy of the base model)
6. Load the base model from step 5 with bnb quantization
7. Load the adapter from step 3

Yes, there is a helper script to do this, but this still has the advantage that we need to load the non-quantized model and that we have to create a completely new model checkpoint with the modified weights.

This PR aims to make this process more convenient by adding a single function `replace_lora_weights_loftq`. This function takes the bnb-quantized LoRA model as input. Then it goes through each module with LoRA weights, lazily loads the corresponding non-quantized weights one at a time using safetensors, computes the quantization error, and replaces the LoRA weights with LoftQ-initialized LoRA weights.

This is much more convenient because we only require very little extra memory thanks to lazy loading, and we don't have to keep an extra copy of the whole model weights.

```python
bnb_config = BitsAndBytesConfig(
    load_in_4bit=True,
)
model = AutoModelForCausalLM.from_pretrained(model_id, quantization_config=bnb_config)
model = get_peft_model(model, LoraConfig(task_type="CAUSAL_LM"))
replace_lora_weights_loftq(model, model_path=...)
```

While working on this, I still found that LoftQ initialization often did not seem to help a lot, as mentioned in #1532. I measured this by creating (1) logits with the base model, (2) logits with the quantized+LoRA model, and (3) logits with the quantized+LoRA+LoftQ model. The expectation is that (1) should be closer to (3) than to (2). This was often not the case.

I therefore added the possibility to run a check each time that we replace a LoRA weight with the LoftQ weights. If this check returns True, we keep the change and proceed to the next weight, otherwise we discard the change before proceeding. That way, we only make the replacement with LoftQ weights if we see a real improvement. Of course, this is only a form of greedy optimization, but it seems to work in practice. And since it's optional, users can choose not to use it.

This PR is not yet finished since I ran into an issue with matching the key names from safetensors not matching.

Furthermore, for now this doesn't support 8bit quantization and the `num_iter` arguments of LoftQ, which I'm not sure is really working. However, I guess the `replace_lora_weights_loftq` function could be called multiple times in a row.

ping @yxli2123

### DIFF
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -42,11 +42,15 @@ config = LoraConfig(init_lora_weights=False, ...)
 
 ### LoftQ
 
+#### Standard approach
+
 When quantizing the base model for QLoRA training, consider using the [LoftQ initialization](https://arxiv.org/abs/2310.08659), which has been shown to improve performance when training quantized models. The idea is that the LoRA weights are initialized such that the quantization error is minimized. To use LoftQ, follow [these instructions](https://github.com/huggingface/peft/tree/main/examples/loftq_finetuning).
 
 In general, for LoftQ to work best, it is recommended to target as many layers with LoRA as possible, since those not targeted cannot have LoftQ applied. This means that passing `LoraConfig(..., target_modules="all-linear")` will most likely give the best results. Also, you should use `nf4` as quant type in your quantization config when using 4bit quantization, i.e. `BitsAndBytesConfig(load_in_4bit=True, bnb_4bit_quant_type="nf4")`.
 
-An easier way to apply LoftQ initialization is to use the convenience function `replace_lora_weights_loftq`. This takes the quantized PEFT model as input and replaces the LoRA weights in-place with their LoftQ-initialized counterparts.
+#### A more convienient way
+
+An easier but more limited way to apply LoftQ initialization is to use the convenience function `replace_lora_weights_loftq`. This takes the quantized PEFT model as input and replaces the LoRA weights in-place with their LoftQ-initialized counterparts.
 
 ```python
 from peft import replace_lora_weights_loftq
@@ -60,12 +64,14 @@ peft_model = get_peft_model(base_model, lora_config)
 replace_lora_weights_loft(peft_model)
 ```
 
-`replace_lora_weights_loftq` also allows you to pass a `callback` argument to give you more control over which layers should be modified or not. To see a more elaborate example of this, check out [this notebook](https://github.com/huggingface/peft/blob/main/examples/loftq_finetuning/LoftQ_weight_replacement.ipynb).
+`replace_lora_weights_loftq` also allows you to pass a `callback` argument to give you more control over which layers should be modified or not, which empirically can improve the results quite a lot. To see a more elaborate example of this, check out [this notebook](https://github.com/huggingface/peft/blob/main/examples/loftq_finetuning/LoftQ_weight_replacement.ipynb).
 
-At the moment, `replace_lora_weights_loftq` only supports:
+`replace_lora_weights_loftq` implements only one iteration step of LoftQ. This means that only the LoRA weights are updated, instead of iteratevily updating LoRA weights and quantized base model weights. This may lead to lower performance but has the advantage that we can use the original quantized weights derived from the base model, instead of having to keep an extra copy of modified quantized weights. Whether this tradeoff is worthwhile depends on the use case.
 
-- Model files stored as a `safetensors` file.
-- bitsandbytes 4bit quantization.
+At the moment, `replace_lora_weights_loftq` has these additional limitations:
+
+- Model files must be stored as a `safetensors` file.
+- Only bitsandbytes 4bit quantization is supported.
 
 <Tip>
 


docs/source/developer_guides/lora.md
```diff
@@ -53,6 +53,26 @@ lora_config = LoraConfig(..., init_lora_weights="loftq", loftq_config=loftq_conf
 peft_model = get_peft_model(base_model, lora_config)
 ```
 
+An easier way to apply LoftQ initialization is to use the convenience function `replace_lora_weights_loftq` from PEFT. This takes the quantized PEFT model as input and replaces the LoRA weights in-place with their LoftQ-initialized counterparts.
```
stevhliu, line 56: I would also maybe explain the code snippet below a bit more. For example, adding a sentence explaining the `BitsAndBytesConfig` a bit more (ie, it comes from the Transformers library and include a link to the config class or to the [docs](https://huggingface.co/docs/transformers/quantization#8-bit).

```suggestion
An easier way to apply LoftQ initialization is to use the convenience function `replace_lora_weights_loftq`. This takes the quantized PEFT model as input and replaces the LoRA weights in-place with their LoftQ-initialized counterparts.
```

docs/source/developer_guides/lora.md
```diff
@@ -53,6 +53,26 @@ lora_config = LoraConfig(..., init_lora_weights="loftq", loftq_config=loftq_conf
...
+
+```python
+from peft import replace_lora_weights_loftq
```
stevhliu, line 59:
```suggestion
from peft import replace_lora_weights_loftq
from transformers import BitsAndBytesConfig
```

docs/source/developer_guides/lora.md
```diff
@@ -53,6 +53,26 @@ lora_config = LoraConfig(..., init_lora_weights="loftq", loftq_config=loftq_conf
...
+
+bnb_config = BitsAndBytesConfig(load_in_4bit, ...)
+base_model = AutoModelForCausalLM.from_pretrained(..., quantization_config=bnb_config)
+# note: don't pass init_lora_weights="loftq" or loftq_config!
+lora_config = LoraConfig(task_type="CAUSAL_LM")
+peft_model = get_peft_model(base_model, lora_config)
+replace_lora_weights_loft(peft_model)
+```
+
+`replace_lora_weights_loftq` also allows to pass the argument `callback` to give you more fine-grained control on which layers should be modified or not. To see a more elaborate example of this, check out [this notebook](https://github.com/huggingface/peft/blob/main/examples/loftq_finetuning/LoftQ_weight_replacement.ipynb).
```
stevhliu, line 69:
```suggestion
`replace_lora_weights_loftq` also allows you to pass a `callback` argument to give you more control over which layers should be modified or not. To see a more elaborate example of this, check out [this notebook](https://github.com/huggingface/peft/blob/main/examples/loftq_finetuning/LoftQ_weight_replacement.ipynb).
```

docs/source/developer_guides/lora.md
```diff
@@ -53,6 +53,26 @@ lora_config = LoraConfig(..., init_lora_weights="loftq", loftq_config=loftq_conf
...
+
+At the moment, using this convenience function has the following restrictions:
+
+- The model file must be stored as a `safetensors` file.
+- Only bitsandbytes 4bit quantization is supported.
```
stevhliu, line 74:
```suggestion
At the moment, `replace_lora_weights_loftq` only supports:

- Model files stored as a `safetensors` file.
- bitsandbytes 4bit quantization.
```

stevhliu (approved):
Really cool function! 👏 

I think we should either remove the section above or provide some clarification for when you should use each method, other than the `replace_lora_weights_loftq` is easier, in which case most users would probably just pick that way. 😛

src/peft/utils/loftq_utils.py
```diff
@@ -15,10 +15,16 @@
 # Reference code: https://github.com/yxli2123/LoftQ/blob/main/utils.py
 # Reference paper: https://arxiv.org/abs/2310.08659
 
+from __future__ import annotations
```
younesbelkada, line 18: Why this import?

src/peft/utils/loftq_utils.py
```diff
@@ -227,3 +233,139 @@ def loftq_init(weight: Union[torch.Tensor, torch.nn.Parameter], num_bits: int, r
...
+
+def _check_model_path_loftq(model_path, peft_model):
+    """Check if there is a local model file corresponding to the given model, then format it properly."""
+    if model_path is None:
+        try:
+            model_path = snapshot_download(peft_model.base_model.config._name_or_path, local_files_only=True)
+        except AttributeError as exc:
+            raise ValueError(
+                "The provided model does not appear to be a transformers model. In this case, you must pass the "
+                "model_path to the safetensors file."
```
younesbelkada, line 246:
```suggestion
                "model_path to the safetensors or bin file."
```

src/peft/utils/loftq_utils.py
```diff
@@ -227,3 +233,139 @@ def loftq_init(weight: Union[torch.Tensor, torch.nn.Parameter], num_bits: int, r
...
+            ) from exc
+        except LocalEntryNotFoundError as exc:
+            raise ValueError("The model.safetensors file must be present on disk, but it could not be found.") from exc
```
younesbelkada, line 249:
```suggestion
            raise ValueError("The adapter_model.safetensors / adapter_model.bin file must be present on disk, but it could not be found.") from exc
```
Assuming you're looking for the adapter model file?

src/peft/utils/loftq_utils.py
```diff
@@ -227,3 +233,139 @@ def loftq_init(weight: Union[torch.Tensor, torch.nn.Parameter], num_bits: int, r
...
+
+    suffix = "model.safetensors"
```
younesbelkada, line 251: should we support `.bin` as well for BC - weights that are pushed on the hub with .bin extension?

src/peft/utils/loftq_utils.py
```diff
@@ -227,3 +233,139 @@ def loftq_init(weight: Union[torch.Tensor, torch.nn.Parameter], num_bits: int, r
...
+    if not is_bnb_4bit_available():
+        raise ValueError("bitsandbytes must be installed and the model must be quantized in 4bits.")
+
+    from peft.tuners.lora import Linear4bit
+
+    model_path = _check_model_path_loftq(model_path, peft_model)
+    prefix = "base_model.model."
+    any_match = False
+
+    with safe_open(model_path, framework="pt", device="cpu") as f:
```
younesbelkada, line 326: Same wuestion above about supporting .bin

younesbelkada (approved):
Thanks a lot @BenjaminBossan ! 
The API looks really great ! Thinking about it we should probably not worry about supporting `.bin` files for this API, but I left them as an open question, wdyt?

src/peft/utils/loftq_utils.py
BenjaminBossan: Unfortunately, I don't think this can work. Or at least it would require a lot of extra effort to make .bin work. The problem is that I want to load one weight at a time to reduce memory overhead, which works with safetensors but not with pickle. For the latter, we would load a whole new copy of the model into memory.

younesbelkada: oh ok, then sounds good !

src/peft/utils/loftq_utils.py
```diff
@@ -227,3 +233,139 @@ def loftq_init(weight: Union[torch.Tensor, torch.nn.Parameter], num_bits: int, r
...
+    residual = weight - dequantized_weight
+    torch.cuda.empty_cache()
+    # Decompose the residualidual by SVD
+    output = _low_rank_decomposition(residual, reduced_rank=reduced_rank)
+    L, R, reduced_rank = output["L"], output["R"], output["reduced_rank"]
+    return R, L
+
+
+@torch.no_grad()
+def replace_lora_weights_loftq(
```
stevhliu, line 282: Let's also document this method somewhere in the API section so users can easily refer to it. Maybe put it in a `## Utility` section with the [LoRA API](https://huggingface.co/docs/peft/package_reference/lora)?

BenjaminBossan: Thanks for the suggestion, I added an entry here: https://moon-ci-docs.huggingface.co/docs/peft/pr_1543/en/package_reference/lora#peft.replace_lora_weights_loftq

src/peft/utils/loftq_utils.py
pacman100, line 326: will this apply for model that have state dict sharded across multiple files?

src/peft/utils/loftq_utils.py
```diff
@@ -227,3 +233,139 @@ def loftq_init(weight: Union[torch.Tensor, torch.nn.Parameter], num_bits: int, r
...
+    if not os.path.exists(model_path):
+        raise FileNotFoundError(
+            f"Could not find file {model_path}, ensure that there is a safetensors file of the model."
+        )
+
+    return model_path
+
+
+@torch.no_grad()
+def _loftq_init_new(qweight, weight, num_bits: int, reduced_rank: int):
```
pacman100, line 263: Asd per the loftq algo, there should be alternating steps to quantize Qt = W-B(t-1)A(t-1) and At,Bt ← SVD(W − Qt) for T steps. here, we only do a single step of A1,B1 ← SVD(W − Qbnb)

Screenshot 2024-03-19 at 12 36 57 PM

pacman100 (commented):
Hello Banjamin, thank you For making it easier to use loftQ, however, I have left a couple comments.

src/peft/utils/loftq_utils.py
pacman100: For example, https://huggingface.co/meta-llama/Llama-2-70b-hf

src/peft/utils/loftq_utils.py
BenjaminBossan: Yeah, I guess with the type annotations used in this file, it wouldn't be necessary, but I also don't think it hurts.

src/peft/utils/loftq_utils.py
BenjaminBossan: As discussed internally, since this approach intends not to adapt the quantized weights (to avoid having an extra copy of all these weights), only single step is implemented.

BenjaminBossan (commented):
The reviewer comments should be addressed.

@pacman100 loading sharded models should now work. It's maybe not the most elegant solution, but this is what I gleaned from combing through transformers code. I tested it on [`gemma-2b`](https://huggingface.co/google/gemma-2b/tree/main), which has 2 shards, and it worked.

Moreover, I adjusted the test to use `all-linear`, now the margin is much larger (>1.5 for MAE and > 2.5 for MSE).

Please check again.

pacman100 (approved):
Thank you @BenjaminBossan for iterating to support larger sharded models and using `all-linear` inline with reducing quantization error for all layers targetted by `bitsandbytes`. Much better results in tests and notebook example! 🔥🚀✨

PR #1548, opened 2024-03-10
Update prompt_based_methods.md

insist93:
remove duplicate commas


younesbelkada (approved):
Thanks for removing the duplicated commas!

PR #1552, opened 2024-03-11
MNT: Use BitsAndBytesConfig as load_in_* is deprecated

BenjaminBossan:
Don't pass `load_in_4bit` or `load_in_8bit` to `AutoModel*.from_pretrained`, as it is deprecated. Instead, pass the appropriate `BitsAndBytesConfig` to the `quantization_config` argument of `from_pretrained`.


younesbelkada (approved):
Great clean up !

PR #903, opened 2023-09-05
support multiple LoRAs in batched inference scenario

pacman100:
### What does this PR do?
1. Support multiple LoRAs in batched inference setting. Here, we create sub-batches via logic `groupby(adapter_name)`. 

### How to use:
[To Do]

ToDos:
- [x] Tests
- [x] Example Notebook
- [ ] Documentation
- [ ] Support all the layers types such as Conv, Embedding, Quantized 8-bit and 4-bit layers.

PR #1558, opened 2024-03-13
FEAT Mixing different LoRA adapters in same batch

BenjaminBossan:
This PR tries to revive the work by Sourab in #903. The core logic is the same between the two PRs. This one should be more complete.

The main idea is to allow the user to mix different LoRA adapters in the same batch. This is useful when the user wants perform inference with a batch that uses different LoRA adapters. Without this, each batch would have to be restricted to the same LoRA adapter(s).

This PR should encompass:

- all task types
- all LoRA layer types
- bnb layers

Extensive tests were added, as well as documentation.

### DIFF
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -274,9 +274,9 @@ Note that the order does not matter here, i.e. the samples in the batch don't ne
 Using this features has some drawbacks, namely:
 
 - It only works for inference, not for training.
-- Disabling adapters using the `with model.disable_adapter` context takes precedence over `adapter_names`.
+- Disabling adapters using the `with model.disable_adapter()` context takes precedence over `adapter_names`.
 - You cannot pass `adapter_names` when some adapter weights where merged with base weight using the `merge_adapter` method. Please unmerge all adapters first by calling `model.unmerge_adapter()`.
-- For obvious reasons, this cannot be used after calling `merge_and_unload`, since all the LoRA adapters will be merged into the base weights in this case.
+- For obvious reasons, this cannot be used after calling `merge_and_unload()`, since all the LoRA adapters will be merged into the base weights in this case.
 - This feature does not currently work with DoRA, so set `use_dora=False` in your `LoraConfig` if you want to use it.
 - There is an expected overhead for inference with `adapter_names`, especially if the amount of different adapters in the batch is high. This is because the batch size is effectively reduced to the number of samples per adapter. If runtime performance is your top priority, try the following:
   - Increase the batch size.
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -281,4 +281,4 @@ Using this features has some drawbacks, namely:
 - There is an expected overhead for inference with `adapter_names`, especially if the amount of different adapters in the batch is high. This is because the batch size is effectively reduced to the number of samples per adapter. If runtime performance is your top priority, try the following:
   - Increase the batch size.
   - Try to avoid having a large number of different adapters in the same batch, prefer homogeneous batches. This can be achieved by buffering samples with the same adapter and only perform inference with a small handfull of different adapters.
-  - Take a look at alternative implementations such as [punica](https://github.com/punica-ai/punica) or [S-LoRA](https://github.com/S-LoRA/S-LoRA), which are specialized to work with a large number of different adapters.
+  - Take a look at alternative implementations such as [LoRAX](https://github.com/predibase/lorax), [punica](https://github.com/punica-ai/punica), or [S-LoRA](https://github.com/S-LoRA/S-LoRA), which are specialized to work with a large number of different adapters.
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -219,3 +219,66 @@ model.unload()
 # delete adapter
 model.delete_adapter("dpo")
 ```
+
+## Inference with different LoRA adapters in the same batch
+
+Normally, each inference batch has to use the same adapter(s) in PEFT. This can sometimes be annoying, because we may have batches that contain samples intended to be used with different LoRA adapters. For example, we could have a base model that works well in English and two more LoRA adapters, one for French and one for German. Usually, we would have to split our batches such that each batch only contains samples of one of the languages, we cannot combine different languages in the same batch.
+
+Thankfully, it is possible to mix different LoRA adapters in the same batch using the `adapter_name` argument. Below, we show an examle of how this works in practice. First, let's load the base model, English, and the two adapters, French and German, like this:
+
+```python
+from transformers import AutoTokenizer, AutoModelForCausalLM
+from peft import PeftModel
+
+model_id = ...
+tokenizer = AutoTokenizer.from_pretrained(model_id)
+
+model = AutoModelForCausalLM.from_pretrained(model_id)
+# load the LoRA adapter for French
+peft_model = PeftModel.from_pretrained(model, <path>, adapter_name="adapter_fr")
+# next, load the LoRA adapter for German
+peft_model.load_adapter(<path>, adapter_name="adapter_de")
+```
+
+Now, we want to generate text on a sample that contains all three languages: The first three samples are in English, the next three are in French, and the last three are in German. We can use the `adapter_names` argument to specify which adapter to use for each sample. Since our base model is used for English, we use the special string `"__base__"` for these samples. For the next three samples, we indicate the adapter name of the French LoRA fine-tune, in this case `"adapter_fr"`. For the last three samples, we indicate the adapter name of the German LoRA fine-tune, in this case `"adapter_de"`. This way, we can use the base model and the two adapters in a single batch.
+
+```python
+inputs = tokenizer(
+    [
+        "Hello, my dog is cute",
+        "Hello, my cat is awesome",
+        "Hello, my fish is great",
+        "Salut, mon chien est mignon",
+        "Salut, mon chat est génial",
+        "Salut, mon poisson est super",
+        "Hallo, mein Hund ist süß",
+        "Hallo, meine Katze ist toll",
+        "Hallo, mein Fisch ist großartig",
+    ],
+    return_tensors="pt",
+    padding=True,
+)
+
+adapter_names = [
+    "__base__", "__base__", "__base__",
+    "adapter_fr", "adapter_fr", "adapter_fr",
+    "adapter_de", "adapter_de", "adapter_de",
+]
+output = peft_model.generate(**inputs, adapter_names=adapter_names, max_new_tokens=20)
+```
+
+Note that the order does not matter here, i.e. the samples in the batch don't need to be grouped by adapter as in the example above. We just need to ensure that the `adapter_names` argument is aligned correctly with the samples.
+
+### Caveats
+
+Using this features has some drawbacks, namely:
+
+- It only works for inference, not for training.
+- Disabling adapters using the `with model.disable_adapter` context takes precedence over `adapter_names`.
+- You cannot pass `adapter_names` when some adapter weights where merged with base weight using the `merge_adapter` method. Please unmerge all adapters first by calling `model.unmerge_adapter()`.
+- For obvious reasons, this cannot be used after calling `merge_and_unload`, since all the LoRA adapters will be merged into the base weights in this case.
+- This feature does not currently work with DoRA, so set `use_dora=False` in your `LoraConfig` if you want to use it.
+- There is an expected overhead for inference with `adapter_names`, especially if the amount of different adapters in the batch is high. This is because the batch size is effectively reduced to the number of samples per adapter. If runtime performance is your top priority, try the following:
+  - Increase the batch size.
+  - Try to avoid having a large number of different adapters in the same batch, prefer homogeneous batches. This can be achieved by buffering samples with the same adapter and only perform inference with a small handfull of different adapters.
+  - Take a look at alternative implementations such as [punica](https://github.com/punica-ai/punica) or [S-LoRA](https://github.com/S-LoRA/S-LoRA), which are specialized to work with a large number of different adapters.
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -537,11 +537,33 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
         except AttributeError:
             return getattr(self.base_model, name)
 
+    @contextmanager
+    def _enable_peft_forward_hooks(self, *args, **kwargs):
+        # If the base model has a method called _enable_peft_forward_hooks, it is invoked as a context. Otherwise, this
+        # runs without any changes
+        if hasattr(self.base_model, "_enable_peft_forward_hooks"):
+            with self.base_model._enable_peft_forward_hooks(*args, **kwargs):
+                yield
+            return
+        else:
+            # nothing to enable
+            yield
+            return
+
     def forward(self, *args: Any, **kwargs: Any):
         """
         Forward pass of the model.
         """
-        return self.get_base_model()(*args, **kwargs)
+        with self._enable_peft_forward_hooks(*args, **kwargs):
+            special_peft_args = {"adapter_names"}
+            kwargs = {k: v for k, v in kwargs.items() if k not in special_peft_args}
+            return self.get_base_model()(*args, **kwargs)
+
+    def generate(self, *args, **kwargs):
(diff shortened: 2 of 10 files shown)


src/peft/peft_model.py
```diff
@@ -537,11 +537,33 @@ def __getattr__(self, name: str):
...
+            yield
+            return
+
     def forward(self, *args: Any, **kwargs: Any):
         """
         Forward pass of the model.
         """
-        return self.get_base_model()(*args, **kwargs)
+        with self._enable_peft_forward_hooks(*args, **kwargs):
+            special_peft_args = {"adapter_names"}
```
younesbelkada, line 558: Should we make that as a class attribute ?

src/peft/peft_model.py
```diff
@@ -537,11 +537,33 @@ def __getattr__(self, name: str):
...
+            kwargs = {k: v for k, v in kwargs.items() if k not in special_peft_args}
+            return self.get_base_model()(*args, **kwargs)
+
+    def generate(self, *args, **kwargs):
+        with self._enable_peft_forward_hooks(*args, **kwargs):
+            special_peft_args = {"adapter_names"}
```
younesbelkada, line 564: same comment as above

src/peft/peft_model.py
```diff
@@ -903,19 +925,22 @@ def forward(
     ):
         return_dict = return_dict if return_dict is not None else self.config.use_return_dict
         peft_config = self.active_peft_config
+        special_peft_args = {"adapter_names"}
```
younesbelkada, line 928: same

src/peft/peft_model.py
```diff
@@ -1396,7 +1436,10 @@ def generate(self, **kwargs):
         )
         try:
             if not peft_config.is_prompt_learning:
-                outputs = self.base_model.generate(**kwargs)
+                special_peft_args = {"adapter_names"}
```
younesbelkada, line 1439: same comment as above

src/peft/tuners/lora/model.py
```diff
@@ -364,6 +371,32 @@ def set_adapter(self, adapter_name: str | list[str]) -> None:
...
+            raise ValueError("Cannot pass `adapter_names` when the model is in training mode.")
+
+        hook_handles = []
+        key_list = [key for key, _ in self.model.named_modules() if self.prefix not in key]
+        for key in key_list:
+            _, target, _ = _get_submodules(self.model, key)
+            if isinstance(target, LoraLayer):
+                pre_forward = partial(_adapter_names_pre_forward_hook, adapter_names=adapter_names)
+                handle = target.register_forward_pre_hook(pre_forward, with_kwargs=True)
+                hook_handles.append(handle)
```
younesbelkada, line 393: Can't we just loop here over the named_modules directly?

docs/source/developer_guides/lora.md
```diff
@@ -219,3 +219,66 @@ model.unload()
...
+```
+
+Note that the order does not matter here, i.e. the samples in the batch don't need to be grouped by adapter as in the example above. We just need to ensure that the `adapter_names` argument is aligned correctly with the samples.
+
+### Caveats
+
+Using this features has some drawbacks, namely:
+
+- It only works for inference, not for training.
+- Disabling adapters using the `with model.disable_adapter` context takes precedence over `adapter_names`.
```
younesbelkada, line 277:
```suggestion
- Disabling adapters using the `with model.disable_adapter()` context takes precedence over `adapter_names`.
```

docs/source/developer_guides/lora.md
```diff
@@ -219,3 +219,66 @@ model.unload()
...
+- You cannot pass `adapter_names` when some adapter weights where merged with base weight using the `merge_adapter` method. Please unmerge all adapters first by calling `model.unmerge_adapter()`.
+- For obvious reasons, this cannot be used after calling `merge_and_unload`, since all the LoRA adapters will be merged into the base weights in this case.
```
younesbelkada, line 279:
```suggestion
- For obvious reasons, this cannot be used after calling `merge_and_unload()`, since all the LoRA adapters will be merged into the base weights in this case.
```

docs/source/developer_guides/lora.md
```diff
@@ -219,3 +219,66 @@ model.unload()
...
+tokenizer = AutoTokenizer.from_pretrained(model_id)
+
+model = AutoModelForCausalLM.from_pretrained(model_id)
+# load the LoRA adapter for French
+peft_model = PeftModel.from_pretrained(model, <path>, adapter_name="adapter_fr")
+# next, load the LoRA adapter for German
+peft_model.load_adapter(<path>, adapter_name="adapter_de")
+```
+
+Now, we want to generate text on a sample that contains all three languages: The first three samples are in English, the next three are in French, and the last three are in German. We can use the `adapter_names` argument to specify which adapter to use for each sample. Since our base model is used for English, we use the special string `"__base__"` for these samples. For the next three samples, we indicate the adapter name of the French LoRA fine-tune, in this case `"adapter_fr"`. For the last three samples, we indicate the adapter name of the German LoRA fine-tune, in this case `"adapter_de"`. This way, we can use the base model and the two adapters in a single batch.
```
younesbelkada, line 243: Shall we use `"default"` instead of `"__base__"` for consistency with the default peft adapter name?

younesbelkada (commented):
This is very clean and cool - thanks so much @BenjaminBossan ! I left few comments that are not blockers - what do you think ?

docs/source/developer_guides/lora.md
BenjaminBossan: `"__base__"` refers to the actual base model, without any LoRA adapter. `"default"` corresponds to the base model + default LoRA adapter, so they're not the same. LMK if I need to clarify the text here or change the example.

src/peft/peft_model.py
BenjaminBossan: done

src/peft/peft_model.py
BenjaminBossan: done

src/peft/peft_model.py
BenjaminBossan: done

src/peft/peft_model.py
BenjaminBossan: done

src/peft/tuners/lora/model.py
BenjaminBossan: done

BenjaminBossan (commented):
Thanks for the feedback, your comments should be addressed now.

docs/source/developer_guides/lora.md
younesbelkada: I see thanks, no all is clear !

younesbelkada (approved):
Thanks so much @BenjaminBossan !

docs/source/developer_guides/lora.md
```diff
@@ -219,3 +219,66 @@ model.unload()
...
+- For obvious reasons, this cannot be used after calling `merge_and_unload()`, since all the LoRA adapters will be merged into the base weights in this case.
+- This feature does not currently work with DoRA, so set `use_dora=False` in your `LoraConfig` if you want to use it.
+- There is an expected overhead for inference with `adapter_names`, especially if the amount of different adapters in the batch is high. This is because the batch size is effectively reduced to the number of samples per adapter. If runtime performance is your top priority, try the following:
+  - Increase the batch size.
+  - Try to avoid having a large number of different adapters in the same batch, prefer homogeneous batches. This can be achieved by buffering samples with the same adapter and only perform inference with a small handfull of different adapters.
+  - Take a look at alternative implementations such as [punica](https://github.com/punica-ai/punica) or [S-LoRA](https://github.com/S-LoRA/S-LoRA), which are specialized to work with a large number of different adapters.
```
Noezor, line 284: Do you plan on having some benchmarking of this feature? (may not even need vs other methods, but at least what throughput per batch size looks like on a transformer with different LoRAs)

src/peft/tuners/lora/bnb.py
```diff
@@ -160,11 +161,56 @@ def get_delta_weight(self, adapter):
...
+                scaling = self.scaling[active_adapter]
+
+                requires_conversion = not torch.is_autocast_enabled()
+                if requires_conversion:
+                    expected_dtype = result.dtype
+                    compute_dtype = lora_A.weight.dtype
+                    if x.dtype != compute_dtype:
+                        x = x.to(compute_dtype)
+
+                # getting the sub-batch, passing it ot LoRA layers and updating the corresponding indices of the linear
```
Noezor, line 194:
```suggestion
                # getting the sub-batch, passing it to LoRA layers and updating the corresponding indices of the linear
```

src/peft/tuners/lora/bnb.py
```diff
@@ -352,11 +398,54 @@ def get_delta_weight(self, adapter):
...
+                lora_B = self.lora_B[active_adapter]
+                dropout = self.lora_dropout[active_adapter]
+                scaling = self.scaling[active_adapter]
+
+                requires_conversion = not torch.is_autocast_enabled()
+                if requires_conversion:
+                    expected_dtype = result.dtype
+                    x = x.to(lora_A.weight.dtype)
+
+                # getting the sub-batch, passing it ot LoRA layers and updating the corresponding indices of the linear
```
Noezor, line 429:
```suggestion
                # getting the sub-batch, passing it to LoRA layers and updating the corresponding indices of the linear
```
I think the code is v clean, but if this comment exists, maybe we could add a comment on the sub-batching logic?

src/peft/tuners/lora/layer.py
```diff
@@ -600,6 +664,36 @@ def get_delta_weight(self, adapter) -> torch.Tensor:
...
+            if active_adapter == "__base__":
+                continue
+            if active_adapter not in self.lora_embedding_A.keys():
+                continue
+
+            embedding_A = self.lora_embedding_A[active_adapter].T
+            embedding_B = self.lora_embedding_B[active_adapter].T
+            scaling = self.scaling[active_adapter]
+
+            # getting the sub-batch, passing it ot LoRA layers and updating the corresponding indices of the linear
```
Noezor, line 689:
```suggestion
            # getting the sub-batch, passing it to LoRA layers and updating the corresponding indices of the linear
```

tests/test_custom_models.py
```diff
@@ -2025,3 +2028,216 @@ def test_requires_grad_oft_same_targets(self):
...
+            output_mixed = peft_model.forward(**inputs).logits
+        with timed():
+            output_mixed = peft_model.forward(**inputs).logits
+
+        time_base, time0, time1, *time_mixed = logs
+        time_non_mixed = (time_base + time0 + time1) / 3
+        time_mixed = min(time_mixed)
+
+        factor = 2.0
+        assert time_mixed < factor * time_non_mixed
```
Noezor, line 2243: I'm a bit confused by this test as I feel like this is not the situation users would be interested in when using this method. I think what we want is to compare that the total computation without batching with 3 different LoRAs is slower than the total computation with batching of 3 different LoRAs and thus having a `factor` < 1

Something like instead
```
        with timed():
            output_base = base_model(**inputs[::3]).logits
```

```
        with timed():
            output0 = peft_model(**inputs[1::3]).logits
```

```
        with timed():
            output1 = peft_model(**inputs[2::3]).logits
```
and finally the comparison being 
```
        time_base, time0, time1, *time_mixed = logs
        time_non_mixed = (time_base + time0 + time1)
        time_mixed = min(time_mixed)

        factor = 1.0
        assert time_mixed < factor * time_non_mixed
```

WDYT? (bonus is that it does a free benchmark to include in the changelog)

Noezor (commented):
Cool feature ! Left 2 significant comments about use-case and comparison, and 3 nits on typo

pacman100 (commented):
Thank you @BenjaminBossan for improving upon the previous PR for enabling batched inference with mixed LoRA adapters, make it feature complete with docs and tests! ✨

I agree with comment by @Noezor on the timing comparison on the timing test.

docs/source/developer_guides/lora.md
BenjaminBossan: Added the reference.

src/peft/tuners/lora/bnb.py
BenjaminBossan: done

src/peft/tuners/lora/bnb.py
BenjaminBossan: Not sure what you mean, could you please elaborate?

src/peft/tuners/lora/layer.py
BenjaminBossan: Done

tests/test_custom_models.py
BenjaminBossan: So the purpose of this check was not really to ensure that this is faster than if the user did the calls separately. In fact, if the users are buffering the same adapter into batches of the same size, I would actually expect that to be faster. In my mind, this feature is more of a convenience feature to avoid that type of extra work. Therefore, this test is more of a regression test to ensure we don't do something unreasonably slow in the implementation.

What you suggest is to benchmark this new feature vs a "naive" alternative which effectively runs at a lower batch size and is thus expected to be slower. I agree that this is still useful to test in the sense that some users might otherwise do exactly that, but I don't see it as a replacement to the existing test. Therefore, I added it to the test but kept the existing benchmark.

pacman100 (approved):

src/peft/tuners/lora/bnb.py
Noezor: The sub-batch logic may not be trivial, and I feel that if we detail what is in this comment, we may also want to add a comment on the batching?

tests/test_custom_models.py
Noezor: >  if the users are buffering the same adapter into batches of the same size, I would actually expect that to be faster

I don't think so? If you have 3 batches with 3 different adapters running separatly (as suggested in my comment above) vs using the method implemented here, the "base" will run 3x, vs just once for the new method. Note that this is also the results that @pacman100 is showing under. This is super powerful.

The way I'm reading the test, its purpose is to make sure that optimized batched Lora inference with 3 different adapters on batch size n is roughly smaller than the average of 3 inferences with 3 different LoRAs of batch size n, but I don't know why this is relevant. The quantities computed are quite different.

src/peft/tuners/lora/bnb.py
BenjaminBossan: I see. Yes, it may take a minute to comprehend, but I have a hard time thinking of a comment that would make it easier vs just reading the code. Just paraphrasing what the code does would not really help IMO.

tests/test_custom_models.py
BenjaminBossan: > I don't think so? If you have 3 batches with 3 different adapters running separatly (as suggested in my comment above) vs using the method implemented here, the "base" will run 3x, vs just once for the new method

Yes, you are correct here, sorry for my misunderstanding. I still think there is a tradeoff, especially if we have large batch size + large number of LoRA adapters, which would result in an effectively very small batch size for the LoRA-part of the forwards call, where this approach could actually be slower, but this may be more of an edge case.

> but I don't know why this is relevant. The quantities computed are quite different.

Yes, this is not comparing the same output. As mentioned, this is not a test from a user perspective, but it's purely there to ensure that we don't have a speed regression of the implementation. The thought is that we expect mixed batches to be slower than pure batches, but we don't want them to be too much slower. Does that make sense?

BenjaminBossan:
The newly added benchmark turned out to be flaky. I'm not quite sure why, but on CPU, the timing differences are quite small. Therefore, I only run the test on GPU. Moreover, I run the timing 3 times and take an average to reduce variance.

docs/source/developer_guides/lora.md
```diff
@@ -219,3 +219,66 @@ model.unload()
 # delete adapter
 model.delete_adapter("dpo")
 ```
+
+## Inference with different LoRA adapters in the same batch
+
+Normally, each inference batch has to use the same adapter(s) in PEFT. This can sometimes be annoying, because we may have batches that contain samples intended to be used with different LoRA adapters. For example, we could have a base model that works well in English and two more LoRA adapters, one for French and one for German. Usually, we would have to split our batches such that each batch only contains samples of one of the languages, we cannot combine different languages in the same batch.
+
+Thankfully, it is possible to mix different LoRA adapters in the same batch using the `adapter_name` argument. Below, we show an examle of how this works in practice. First, let's load the base model, English, and the two adapters, French and German, like this:
```
AngledLuffa, line 227: examle -> example

PR #1582, opened 2024-03-22
FIX / Docs: Fix doc link for layer replication

younesbelkada:
as per title

cc @pacman100 @BenjaminBossan


pacman100 (approved):
Thank you @younesbelkada!

PR #1584, opened 2024-03-25
[feat] Add `lru_cache` to `import_utils` calls that did not previously have it

### LINKED ISSUE #1576: Loading LORA weights in `diffusers` with a `peft` backend increases in latency as more paths are added to `PYTHONPATH`
### System Info

```
accelerate==0.21.0
diffusers==0.26.3
peft==0.9.0
safetensors==0.3.3
tokenizers==0.15.2
torch==2.2.1
transformers==4.36.2
```

### Who can help?

@sayakpaul 

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

```python
from diffusers import DiffusionPipeline
import time
import torch
import sys
import os
import shutil

pipe_id = "stabilityai/stable-diffusion-xl-base-1.0"
pipe = DiffusionPipeline.from_pretrained(pipe_id, torch_dtype=torch.float16).to("cuda")

loras = [
    {
        "adapter_name": "anime"
        "location": "./anime",
        "weight_name": "anime.safetensors",
        "token": "my_anime"
    },
]

def run_dynamic_lora_inference(lora):
    start_load_time = time.time()
    pipe.load_lora_weights(lora["location"], weight_name=lora["weight_name"], adapter_name=lora["adapter_name"])
    end_load_time = time.time()
    prompt = f"Illustration of a dog in the style of {lora["token"]}"

    start_fuse_time = time.time()
    pipe.fuse_lora()
    end_fuse_time = time.time()

    start_set_adapter_time = time.time()
    pipe.set_adapters(lora_name)
    end_set_adapter_time = time.time()

    start_inference_time = time.time()
    image = pipe(
        prompt, num_inference_steps=30, generator=torch.manual_seed(0)
    ).images[0]
    end_inference_time = time.time()

    start_unfuse_time = time.time()
    pipe.unfuse_lora()
    end_unfuse_time = time.time()

    start_unload_time = time.time()
    pipe.unload_lora_weights()
    end_unload_time = time.time()

    image.save(f"./{lora_name}.png")

    print("Load time:", end_load_time - start_load_time)
    print("Fuse time:", end_fuse_time - start_fuse_time)
    print("Set adapter time", end_set_adapter_time - start_set_adapter_time)
    print("Inference time:", end_inference_time - start_inference_time)
    print("Unfuse time:", end_unfuse_time - start_unfuse_time)
    print("Unload time:", end_unload_time - start_unload_time)

def add_to_python_path():
    root_path = "./folders"
    shutil.rmtree(root_path)
    os.mkdir(root_path)

    folders =  [f"folder_{x}" for x in range(0, 10000)]
    for folder in folders:
        os.mkdir(os.path.join(root_path, folder))
        sys.path.append(os.path.join(root_path, folder))

def main():
    for lora in loras:
        run_dynamic_lora_inference(lora)

main()
```

Flamegraph: 
profile

### Expected behavior

I run a system with a somewhat large `PYTHONPATH` that we can't truncate, and we are currently blocked from upgrading `diffusers` to any version that uses `peft` for LORA inference. 

It's loosely based on this post: https://huggingface.co/blog/lora-adapters-dynamic-loading

We've observed a behavior where the time taken for `load_lora_weights` increases significantly with more paths added to `PYTHONPATH`. This can be reproduced in the example provided - with 10,000 folders added to PYTHONPATH, we get the following latencies:

```
Load time: 291.78441095352173
Fuse time: 0.12406659126281738
Set adapter time 0.06171250343322754
Inference time: 9.685987710952759
Unfuse time: 0.08063459396362305
Unload time: 0.15737533569335938
```

Benchmarking against 1, 10, 100, 1000, 10000 and 50000 entries in the `PYTHONPATH`, we get a pretty astounding increase in load latency: 

image

Even at 100 entries, we're looking at an extra 4 seconds per `load` call which is a pretty significant increase.

We looked briefly at it and came to the conclusion that it's something to do with the way `peft` runs module imports, particularly repetitive calls to import modules, where the imports are not cached, eg, `importlib.util.find_spec` doesn't cache imports. 

Instead of this behaviour, we'd expect that `load_lora_weights` retains a relatively constant load time, regardless of the length of our python path.

tisles:
Fixes #1576 

Adding @cache to the functions in `import_utils` results in a constant time for LORA loading, regardless of the length of python path. Have verified this using a python path with 50,000 directories - previously the lora load time was 259 seconds, now it's down to a constant time of 1.3 seconds 👍🏻 

With some flamegraphs to prove! Before:
profile

After: 
profile_fixed

### DIFF
--- a/src/peft/import_utils.py
+++ b/src/peft/import_utils.py
@@ -18,12 +18,12 @@ from functools import lru_cache
 import packaging.version
 
 
-@lru_cache(maxsize=None)
+@lru_cache
 def is_bnb_available() -> bool:
     return importlib.util.find_spec("bitsandbytes") is not None
 
 
-@lru_cache(maxsize=None)
+@lru_cache
 def is_bnb_4bit_available() -> bool:
     if not is_bnb_available():
         return False
@@ -33,7 +33,7 @@ def is_bnb_4bit_available() -> bool:
     return hasattr(bnb.nn, "Linear4bit")
 
 
-@lru_cache(maxsize=None)
+@lru_cache
 def is_auto_gptq_available():
     if importlib.util.find_spec("auto_gptq") is not None:
         AUTOGPTQ_MINIMUM_VERSION = packaging.version.parse("0.5.0")
@@ -47,7 +47,7 @@ def is_auto_gptq_available():
             )
 
 
-@lru_cache(maxsize=None)
+@lru_cache
 def is_optimum_available() -> bool:
     return importlib.util.find_spec("optimum") is not None
 
@@ -69,11 +69,11 @@ def is_torch_tpu_available(check_device=True):
     return False
 
 
-@lru_cache(maxsize=None)
+@lru_cache
 def is_aqlm_available():
     return importlib.util.find_spec("aqlm") is not None
 
 
-@lru_cache(maxsize=None)
+@lru_cache
 def is_auto_awq_available():
     return importlib.util.find_spec("awq") is not None
--- a/src/peft/import_utils.py
+++ b/src/peft/import_utils.py
@@ -13,15 +13,15 @@
 # limitations under the License.
 import importlib
 import importlib.metadata as importlib_metadata
-from functools import lru_cache
+from functools import cache, lru_cache
 
 import packaging.version
 
-
+@cache
 def is_bnb_available() -> bool:
     return importlib.util.find_spec("bitsandbytes") is not None
 
-
+@cache
 def is_bnb_4bit_available() -> bool:
     if not is_bnb_available():
         return False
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Thanks a lot for the PR and testing the solution, great that it solves the issue. Just out of curiosity, is there a specific reason for setting `maxsize=None`? From my understanding, just using the default should work equally well here.

tisles:
> Thanks a lot for the PR and testing the solution, great that it solves the issue. Just out of curiosity, is there a specific reason for setting maxsize=None? From my understanding, just using the default should work equally well here.

Yep, good point - you're right. I've updated it to remove the `None` setting.

BenjaminBossan (approved):
Thanks for identifying the issue and providing a fix, LGTM.

PR #1587, opened 2024-03-25
DOC: Short section on using transformers pipeline

BenjaminBossan:
Add a short section on the usage of `pipeline` from transformers.

Note that for some reason, the diff looks much bigger on GH than it actually is -- in the last paragraph, only a single word was removed.


sayakpaul (approved):
Sweet!

PR #1591, opened 2024-03-27
fix deepspeed zero3+prompt tuning bug. word_embeddings.weight shape i…

sywangyi:
…s torch.Size([0])

### DIFF
--- a/src/peft/tuners/prompt_tuning/model.py
+++ b/src/peft/tuners/prompt_tuning/model.py
@@ -16,6 +16,8 @@ import math
 
 import torch
 
+from peft.utils.integrations import gather_params_ctx
+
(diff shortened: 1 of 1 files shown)


sywangyi:
@younesbelkada please help review

BenjaminBossan:
Thanks for this PR. Just to confirm, you tested this fix with DeepSpeed and it was necessary and sufficient to allow training the model?

sywangyi:
yes. I have tested this fixed with deepspeed. and with the fix. the training could continue without error.

BenjaminBossan (approved):
Thanks a lot for identifying this issue with prompt tuning + DeepSpeed and providing the fix!

PR #1603, opened 2024-03-28
FIX: bnb config wrong argument names

BenjaminBossan:
Several tests were using `bnb_4bit_compute_type` but the argument should be called `bnb_4bit_compute_dtype`. Now the 
correct name is used.

This change should not affect the tests, because they were passing the default value anyway. Therefore, the fact that this argument was passed incorrectly (and thus, presumably, ignored) should not affect the results.

Also, in some tests `quantization_config=BitsAndBytesConfig(load_in_4bit=True)` was passed as argument to `BitsAndBytesConfig` due to an overeager search and replace in #1552. Fixing those changes the test outcome but locally they still passed for me.

It is a bit unfortunate that this sort of typo passes completely silently.


younesbelkada (approved):
Thanks for fixing the wrong argument names for bnb !

Issue #1598, opened 2024-03-28
examples/sft/run_peft.sh model load dtype error 


galtay:
### System Info

peft installed from main
c4c826c0c0dfe06620db8b8adcfa55f22acfed61

- `transformers` version: 4.39.1
- Platform: Linux-6.5.0-26-generic-x86_64-with-glibc2.35
- Python version: 3.11.8
- Huggingface_hub version: 0.22.0
- Safetensors version: 0.4.2
- Accelerate version: 0.28.0
- PyTorch version (GPU?): 2.2.1+cu121 (True)
GPU RTX3090

### Who can help?
@pacman100 @younesbelkada 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

```shell
CUDA_VISIBLE_DEVICES=0 python examples/sft/train.py \
--seed 100 \
--model_name_or_path "mistralai/Mistral-7B-v0.1" \
--dataset_name "smangrul/ultrachat-10k-chatml" \
--chat_template_format "chatml" \
--add_special_tokens False \
--append_concat_token False \
--splits "train,test" \
--max_seq_len 2048 \
--num_train_epochs 1 \
--logging_steps 5 \
--log_level "info" \
--logging_strategy "steps" \
--evaluation_strategy "epoch" \
--save_strategy "epoch" \
--bf16 True \
--packing True \
--learning_rate 1e-4 \
--lr_scheduler_type "cosine" \
--weight_decay 1e-4 \
--warmup_ratio 0.0 \
--max_grad_norm 1.0 \
--output_dir "mistral-sft-lora" \
--per_device_train_batch_size 8 \
--per_device_eval_batch_size 8 \
--gradient_accumulation_steps 8 \
--gradient_checkpointing True \
--use_reentrant True \
--dataset_text_field "content" \
--use_peft_lora True \
--lora_r 8 \
--lora_alpha 16 \
--lora_dropout 0.1 \
--lora_target_modules "all-linear" \
--use_4bit_quantization True \
--use_nested_quant True \
--bnb_4bit_compute_dtype "bfloat16" \
--use_flash_attn True
```

it seems that `--bf16` may be having a strange interaction with the bitsandbytes config
adding 
```
--bnb_4bit_quant_storage_dtype "bfloat16" 
```
allows it to run, but that seems like a hack b/c the default for that is typically `uint8`

[...]

PR #1614, opened 2024-04-02
fix the torch_dtype and quant_storage_dtype

pacman100:
### What does this PR do?
1. Fixes https://github.com/huggingface/peft/issues/1598

### DIFF
--- a/examples/sft/utils.py
+++ b/examples/sft/utils.py
@@ -125,12 +125,13 @@ def create_and_prepare_model(args, data_args, training_args):
             load_in_4bit=args.use_4bit_quantization,
         )
     else:
+        torch_dtype = quant_storage_dtype if quant_storage_dtype and quant_storage_dtype.is_floating_point else torch.float32
         model = AutoModelForCausalLM.from_pretrained(
             args.model_name_or_path,
             quantization_config=bnb_config,
             trust_remote_code=True,
             attn_implementation="flash_attention_2" if args.use_flash_attn else "eager",
-            torch_dtype=quant_storage_dtype or torch.float32,
+            torch_dtype=torch_dtype,
         )
 
     peft_config = None
--- a/examples/sft/utils.py
+++ b/examples/sft/utils.py
@@ -125,7 +125,9 @@ def create_and_prepare_model(args, data_args, training_args):
             load_in_4bit=args.use_4bit_quantization,
         )
     else:
-        torch_dtype = quant_storage_dtype if quant_storage_dtype and quant_storage_dtype.is_floating_point else torch.float32
+        torch_dtype = (
+            quant_storage_dtype if quant_storage_dtype and quant_storage_dtype.is_floating_point else torch.float32
+        )
         model = AutoModelForCausalLM.from_pretrained(
             args.model_name_or_path,
             quantization_config=bnb_config,
--- a/examples/sft/train.py
+++ b/examples/sft/train.py
@@ -40,7 +40,7 @@ class ModelArguments:
         metadata={"help": "Compute dtype for 4bit base models"},
     )
     bnb_4bit_quant_storage_dtype: Optional[str] = field(
-        default="float32",
+        default="uint8",
         metadata={"help": "Quantization storage dtype for 4bit base models"},
     )
     bnb_4bit_quant_type: Optional[str] = field(


BenjaminBossan:
Could you please fix the code quality issue?

BenjaminBossan (approved):
Thanks for adding this fix, Sourab, and thanks @galtay for the investigation and proposal. From my perspective, this can be merged if @galtay can confirm that this solves the issue.

examples/sft/utils.py
```diff
@@ -125,12 +125,15 @@ def create_and_prepare_model(args, data_args, training_args):
             load_in_4bit=args.use_4bit_quantization,
         )
     else:
+        torch_dtype = (
+            quant_storage_dtype if quant_storage_dtype and quant_storage_dtype.is_floating_point else torch.float32
+        )
```
galtay, line 130: I think this should be something like, 
```
if training_args.fp16:
    torch_dtype = torch.float16
elif training_args.bf16:
    torch_dtype = torch.bfloat16
else:
    torch_dtype = torch.float32
```

galtay: with the `quant_storage_dtype=uint8` the current code will cause `torch_dtype=torch.float32` and produce, 
```
Flash Attention 2.0 only supports torch.float16 and torch.bfloat16 dtypes, but the current dype in MistralForCausalLM is torch.float32. You should run training or inference using Automatic Mixed-Precision via the `with torch.autocast(device_type='torch_device'):` decorator, or load the model with the `torch_dtype` argument. Example: `model = AutoModel.from_pretrained("openai/whisper-tiny", attn_implementation="flash_attention_2", torch_dtype=torch.float16)`
```

galtay: this is a potentially useful code snippet that shows that with 4bit quantization most layers will be torch.uint8, but some will still be torch_dtype (e.g. embedding layers and layernorms), 
```python
dtc = Counter([p.dtype for n,p in model.named_parameters()])
print(dtc)
```

galtay: with the torch_dtype changes above i can confirm that on my system the `run_peft.sh` script fits on a single 3090 (using just under 24GB of RAM) with `quant_storage_dtype` in (torch.uint8, torch.bfloat16, torch.float16)

pacman100: > with quant_storage_dtype=uint8 the current code will cause torch_dtype=torch.float32 and produce, Flash Attention 2.0 only supports torch.float16 and torch.bfloat16 dtypes, but the current dype in MistralForCausalLM is torch.float32. You should run training or inference using Automatic Mixed-Precision via the `with torch.autocast(device_type='torch_device'):` decorator, or load the model with the `torch_dtype` argument. Example: `model = AutoModel.from_pretrained("openai/whisper-tiny", attn_implementation="flash_attention_2", torch_dtype=torch.float16)`
the example here (https://github.com/huggingface/peft/blob/main/examples/sft/run_peft.sh) includes the --bf16 flag, but it is not currently used to set the torch_dtype

This is a warning and should be ignored as this is the correct behaviour when training. See https://github.com/huggingface/transformers/pull/28142 for more information of why loading the model in half-precision leads to training instability.

pacman100: > I think this should be something like,
> 
> ```python
> if training_args.fp16:
>     torch_dtype = torch.float16
> elif training_args.bf16:
>     torch_dtype = torch.bfloat16
> else:
>     torch_dtype = torch.float32
> ```

Hello, this shouldn't be the case. pretrained model should mostly be loaded in FP32 with the exception when using FSDP+QLoRA or DeepSpeed Stage3+QLoRA wherein the precision of the loaded pretrained model should be inline with bitsandbytes quant storage datatype.

pacman100: Also, as the TRL calls `prepare_model_for_kbit_training` for bnb 8-bit/4-bit quantization, and the below snippet will upcast all the non-quantized layers/unint8 layers to `float32`. The only scenario in which again a casting to half-precision happens is for `torch.bfloat16` as per the findings of QLoRA paper via TRL's function `peft_module_casting_to_bf16`. Hence, for `float16`, you would still find that non bnb-layers/uint8-layers are casted to `float32`. 

https://github.com/huggingface/peft/blob/02b5aeddf9c1ea11451f10a8a26da7e5df8cca4a/src/peft/utils/other.py#L105-L109

> with the torch_dtype changes above i can confirm that on my system the run_peft.sh script fits on a single 3090 (using just under 24GB of RAM) with quant_storage_dtype in (torch.uint8, torch.bfloat16, torch.float16)

Can you confirm if this is still true with the only changes of this PR?

galtay: ok, I see, thanks for the links! makes sense. with just the changes in this PR and specifying `bnb_4bit_quant_storage`=`uint8` in the script (https://github.com/huggingface/peft/blob/main/examples/sft/run_peft.sh) I get layer dtypes like the following after the SFTTrainer is initialized, 
```
Using auto half precision backend
trainable params: 20,971,520 || all params: 7,262,769,152 || trainable%: 0.2887537736790784
model.embed_tokens.weight torch.Size([32008, 4096]) torch.bfloat16
model.layers.0.self_attn.q_proj.base_layer.weight torch.Size([8388608, 1]) torch.uint8
model.layers.0.self_attn.q_proj.lora_A.default.weight torch.Size([8, 4096]) torch.bfloat16
model.layers.0.self_attn.q_proj.lora_B.default.weight torch.Size([4096, 8]) torch.bfloat16
model.layers.0.self_attn.k_proj.base_layer.weight torch.Size([2097152, 1]) torch.uint8
model.layers.0.self_attn.k_proj.lora_A.default.weight torch.Size([8, 4096]) torch.bfloat16
model.layers.0.self_attn.k_proj.lora_B.default.weight torch.Size([1024, 8]) torch.bfloat16
model.layers.0.self_attn.v_proj.base_layer.weight torch.Size([2097152, 1]) torch.uint8
model.layers.0.self_attn.v_proj.lora_A.default.weight torch.Size([8, 4096]) torch.bfloat16
model.layers.0.self_attn.v_proj.lora_B.default.weight torch.Size([1024, 8]) torch.bfloat16
model.layers.0.self_attn.o_proj.base_layer.weight torch.Size([8388608, 1]) torch.uint8
model.layers.0.self_attn.o_proj.lora_A.default.weight torch.Size([8, 4096]) torch.bfloat16
model.layers.0.self_attn.o_proj.lora_B.default.weight torch.Size([4096, 8]) torch.bfloat16
model.layers.0.mlp.gate_proj.base_layer.weight torch.Size([29360128, 1]) torch.uint8
model.layers.0.mlp.gate_proj.lora_A.default.weight torch.Size([8, 4096]) torch.bfloat16
model.layers.0.mlp.gate_proj.lora_B.default.weight torch.Size([14336, 8]) torch.bfloat16
model.layers.0.mlp.up_proj.base_layer.weight torch.Size([29360128, 1]) torch.uint8
model.layers.0.mlp.up_proj.lora_A.default.weight torch.Size([8, 4096]) torch.bfloat16
model.layers.0.mlp.up_proj.lora_B.default.weight torch.Size([14336, 8]) torch.bfloat16
model.layers.0.mlp.down_proj.base_layer.weight torch.Size([29360128, 1]) torch.uint8
model.layers.0.mlp.down_proj.lora_A.default.weight torch.Size([8, 14336]) torch.bfloat16
model.layers.0.mlp.down_proj.lora_B.default.weight torch.Size([4096, 8]) torch.bfloat16
model.layers.0.input_layernorm.weight torch.Size([4096]) torch.float32
model.layers.0.post_attention_layernorm.weight torch.Size([4096]) torch.float32
```
and it still runs on a single 3090. maybe worthwhile to remove the `--bf16` flag from the script if it is not used.

galtay (approved):

examples/sft/utils.py
pacman100: > maybe worthwhile to remove the --bf16 flag from the script if it is not used.

It is required as the fine-tuning is using automatic mixed precision otherwise, one would get float/half mismatch errors.

PR #1615, opened 2024-04-02
FIX Multiple adapters and modules_to_save

### LINKED ISSUE #1574: 'set_adapter()' throws "ValueError: Adapter not found in odict_keys" after 'load_adapter()'
### System Info

peft-0.9.0

### Who can help?

@pacman100 

### Information

- [X] My own modified scripts

### Tasks

### Reproduction
Saved two finetuned models to `peft_model_id_1` and `peft_model_id_2` in this way:

```
model = AutoModelForSequenceClassification.from_pretrained("bert-base-uncased")

if training_args.do_train:
    """ Add PEFT """
    peft_config = LoraConfig(
        task_type=TaskType.SEQ_CLS, 
        r=1, 
        lora_alpha=1, 
        lora_dropout=0.1
    )

    trainer = Trainer(
            model=model,
            args=training_args,
            train_dataset=train_dataset if training_args.do_train else None,
            eval_dataset=eval_dataset if training_args.do_eval else None,
            tokenizer=tokenizer,
            data_collator=data_collator,
        )
    
    train_result = trainer.train()

    peft_model_id = training_args.output_dir
    trainer.model.save_pretrained(peft_model_id)
    tokenizer.save_pretrained(peft_model_id)
```

Load the saved two adapters:
```
##################  load ####################
base_model = AutoModelForSequenceClassification.from_pretrained(
        "bert-base-uncased"
    )

merged_model = PeftModel.from_pretrained(base_model, peft_model_id_1, adapter_name="id_1")
merged_model.load_adapter(peft_model_id_2, adapter_name="id_2")
```

When trying to set adapter:
```
merged_model.set_adapter("id_2")
```

Got error:
```
ValueError: Adapter id_2 not found in odict_keys(['id_1'])
```

### Expected behavior

- I have checked `merged_model.peft_config` and got:
```
{'id_1': LoraConfig(peft_type...ora=False), 'id_2': LoraConfig(peft_type...ora=False)}
```

and both two adapters occur in `merged_mode`l if ```print(merged_model)```.
- `merged_model.add_weighted_adapter` can also work. 
- Only `merged_model.set_adapter("id_2")` has problems: `ValueError: Adapter id_2 not found in odict_keys(['id_1'])`. It seems that only the first adapter can be seen in `odict_keys`

BenjaminBossan:
Resolves #1574.

Previously, we had the bug that if we had multiple adapters, some with
`modules_to_save` and others without, when trying to switch to an adapter
without `modules_to_save`, the `ModulesToSaveWrapper` would raise an error
because it cannot find that adapter. Now, when it detects this, it is
just disabled (so it uses the original weight).

Moreover, we had the issue that when we were using classes such as
`PeftModelForSequenceClassification`, we implicitly added the classifier
layers to `model.modules_to_save`. However, this would only add a new
`ModulesToSaveWrapper` instance for the first adapter being initialized.
When initializing a 2nd adapter via `add_adapter`, this information was
ignored. To fix this, I now update the `peft_config.modules_to_save` to
explicitly add the classifier layers. This is a departure from how this
worked previously, but I'm couldn't find a better way to ensure that
this bug was fixed (LMK if you have a suggestion).

Finally, there was a bug in `add_weighted_adapters` when we were merging
multiple adapters with `modules_to_save`. Previously, when we called
`model.add_weighted_adapter`, the LoRA weights were merged and a new
`ModulesToSaveWrapper` was added for the new adapter based on the first
`LoraConfig` of the two adapters. This `ModulesToSaveWrapper` is just a copy
of the original weights. Thus, when we switch to the newly merged
adapter, we just use the original weights for `modules_to_save`. This
doesn't make a lot of sense and is probably surprising for the user.
Now, we raise an error when we detect this to alert the user to this
fact.

Note that when only one of the adapters to be merged has a
`modules_to_save`, this does not raise an error, instead that module is
being used.

Edit: If this is merged, we should add a note to the next release notes
about possible changes, as this results in a different model for the same
config (though it's a rather specific edge case). We should clarify that the
previous behavior was erroneous and new one is correct.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -698,8 +698,6 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
                 self.modules_to_save = set(peft_config.modules_to_save)
             else:
                 self.modules_to_save.update(peft_config.modules_to_save)
-
-        if self.modules_to_save:
             _set_trainable(self, adapter_name)  # this may add a new ModulesToSaveWrapper
 
     @classmethod
@@ -716,7 +714,7 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
 
         return hf_hub_download_kwargs, other_kwargs
 
-    def _update_offload(self, offload_index: dict[dict[str:str]], adapters_weights: dict[str : torch.tensor]):
+    def _update_offload(self, offload_index: dict[str, dict[str, str]], adapters_weights: dict[str, torch.tensor]):
         """
         Update the offload_index and safetensors files for loading and mergine PeftModels with disk-offloaded modules.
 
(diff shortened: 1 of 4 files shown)


BenjaminBossan:
@pacman100 @younesbelkada I updated this PR after the internal discussion (check the updated PR description).

PR #1620, opened 2024-04-04
FEAT: Allow ignoring mismatched sizes when loading

### LINKED ISSUE #1605: Is there a choice to use `ignore_mismatch_sizes` in PeftModel.from_pretrained like AutoModel in Transformers?
### Feature request

Add ignore_mismatch_sizes in PeftModel.from_pretrained.

### Motivation

When I try to add some special tokens into an LM and use an adapter, I find though I can set ignore_mismatch_sizes = True when I load the base model, I have no similar choice to ignore the mismatch parameters when I load the adapter.
I have searched a possible way to first load the adapter and then add the base model. However, in my case, my base model has been some trained parameters that I cannot resize the parameters after loading the adapter.
Have I miss some args? Or is it a necessary choice?

### Your contribution

I don't know what can I do.

BenjaminBossan:
Resolves #1605

When users pass `ignore_mismatched_sizes=True` to `PeftModel.from_pretrained`, the mismatched tensors will be ignored instead of raising an error. This is in line with how transformers handles this argument.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -839,7 +839,10 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
         adapters_weights = load_peft_weights(model_id, device=torch_device, **hf_hub_download_kwargs)
 
         # load the weights into the model
-        load_result = set_peft_model_state_dict(self, adapters_weights, adapter_name=adapter_name)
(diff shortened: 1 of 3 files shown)


pacman100 (approved):
Thank you @BenjaminBossan for supporting `ignore_mismatched_sizes` flag when loading PEFT models!

PR #1624, opened 2024-04-05
FIX In the image classification example, Change the model to the LoRA…

changhwa:
- FIX In the image classification example, Change the model to the LoRA model in Train.
- Referred to "https://huggingface.co/docs/peft/main/en/task_guides/image_classification_lora#image-classification-using-lora"


BenjaminBossan (approved):
Thanks for fixing this error in the notebook.

PR #1628, opened 2024-04-08
FIX: Send results to correct channel

younesbelkada:
The channel name has changed

cc @Titus-von-Koeller @BenjaminBossan @pacman100


BenjaminBossan (approved):
Good catch

PR #1631, opened 2024-04-08
fix: allow load_adapter to use different device

yhZhai:
### Problem
In the multi-GPU DDP setting, the `load_adapter` function will always map the state_dict to "cuda", This causes all processes load the parameters to the first GPU, potentially causing OOM error.

### Changes
Add an optional "torch_device" argument to the `load_adapter` function, so that by assigning it to "cpu", the parameters will not occupy memory on the first GPU.


BenjaminBossan:
Could you please run `make style` on your code?

BenjaminBossan (approved):
Great PR, thanks for adding this useful parameter.

PR #1634, opened 2024-04-09
fix bf16 model type issue for ia3

sywangyi:


sywangyi:
@BenjaminBossan please help review.

sywangyi:
if set torch_dtype is bfloat16 in model.from_pretained(), all adapter layers after get_peft_model(xxx) should be bfloat16 as well. fix issue in ia3 and adalora, also ia3 forward return dtype should be aligned with the base_layer.forward() output.

BenjaminBossan (approved):
Thanks, those changes to correctly enable bf16 look good. I'll request another review by the co-maintainers just to be sure.

src/peft/tuners/adalora/layer.py
```diff
@@ -66,9 +66,14 @@ def update_layer(self, adapter_name, r, lora_alpha, lora_dropout, init_lora_weig
...
         if hasattr(self.get_base_layer(), "qweight"):
             # QuantLinear
-            self.to(self.get_base_layer().qweight.device)
+            weight = self.get_base_layer().qweight
         else:
-            self.to(self.get_base_layer().weight.device)
+            weight = self.get_base_layer().weight
+
+        if weight.dtype.is_floating_point or weight.dtype.is_complex:
+            self.to(weight.device, dtype=weight.dtype)
```
pacman100, line 74: this would cast the trainable lora weights too in half precision if `torch_dtype` is a half precision which is not good for stable training. We are discussion this internally as this has been used at other places too.

sywangyi: so I would update the PR and focus on IA3 forward return type bug. also IA3 merge weight has type issue in torch_dtype half precision case. if IA3's weight is float32 and base layer weight is bfloat16, after weight merged, it should be bfloat16 instead of float32, am I right?

BenjaminBossan:
@pacman100 Could you please check again?

pacman100 (approved):
Thank you, @sywangyi, for the fixes!

PR #1630, opened 2024-04-08
itemsize is torch>=2.1, use element_size()

winglian:
PEFT lists torch version as >= 1.13.0 but `.itemsize` is only torch 2.1+. This is unexpectedly breaking for users on torch 2.0.

https://github.com/huggingface/peft/blob/main/setup.py#L63

PR #1635, opened 2024-04-09
FIX Correctly call element_size

BenjaminBossan:
Should fix the error introduced by #1630.

AFAICT, `element_size` should be called on the parameter, not the `dtype`. Unfortunately, I had issues getting older PyTorch versions to work with bnb, so I haven't tested the initial issue.

To be safe, I also re-added the previous code path using `itemsize`, although it might be unnecessary (we would have to check the PyTorch code to verify when the different attributes/methods were added).


BenjaminBossan:
Ping @winglian if you could please check if this PR still resolves your initial issue.

younesbelkada (approved):
thanks for the fix!

winglian:
@BenjaminBossan yup, this works for me. thanks!

PR #1648, opened 2024-04-13
print_trainable_parameters - format `%` to be sensible

stas00:
This PR replaces `trainable%: 0.5916145025956931` with `trainable%: 0.5916` as it's done already in `src/peft/mixed_model.py`

Probably `.2f` is more than enough, but since the other one uses `.4f` using that.


BenjaminBossan (approved):
Thanks for this small improvement. This has indeed bothered me too in the past. I think `.2f` wouldn't be enough, as there are some methods with really low numbers of trainable params, but `.4f` should be good.

Issue #1654, opened 2024-04-15
Integrating ReFT


raven38:
Hello all, and thank you for your great work!

[ReFT](https://arxiv.org/abs/2404.03592),  a representation finetuning framework more parameter-efficient than popular PEFTs like LoRA, is announced earlier this month.

I implemented LoReFT, a instance of ReFT, up on the your `PEFT` library referencing to [the original implementation](https://github.com/stanfordnlp/pyreft).
My implementation is available at https://github.com/raven38/peft 

Would you be interested to perhaps integrate ReFT into peft? I would be happy to work on this if there is interest from you and the community.

BenjaminBossan:
Hi, thanks for bringing this to our attention. We had already looked at (Lo)ReFT internally and had some discussion about whether it would be a good addition to PEFT. IIRC, the ReFT repo was heavily relying on pyvene, does your fork do that too or are you integrating the pyvene code? Maybe you can open a draft PR so that we can more easily discuss what has been changed.

Regarding the paper itself, I've only skimmed it, so I don't have the full picture. I'll quote myself on what I had to say internally:

> One thing that makes me wonder is how much of the benefits of ReFT can be attributed to the fact that they only apply it selectively to certain layers and token positions. At first glance, I didn't see an ablation study to check how well and efficiently ReFT works without this.
This makes me wonder if similar gains cannot be obtained by applying LoRA/DoRA/VeRA to only selective layers. Maybe that's something that we can add to PEFT and test out. All this reminds me also of [LISA](https://arxiv.org/abs/2403.17919), which works by just training a random subset of layers (randomized at each step).

Do you have any further insights into that?

raven38:
Thank you for the feedback.
My code does not depend on `pyvene` and works with just integrating few codes in `pyreft` and `pyvene`. I think ReFT mainly consists of the overwrite of layers, additional weight and selection of intervention tokens. In my implementation, I use `PEFT` library for the overwrite of layers, borrow weight operations from `pyreft` and `pyvene`, and use Pytorch's gather and scatter_ for the selection of tokens for intervention.

A draft PR is prepared for discussion.

frankaging:
@raven38 @BenjaminBossan hey! i was randomly browsing on GitHub and found this ticket, super exciting to see the PEFT library potentially supporting ReFT.

[...]

PR #1657, opened 2024-04-16
Integrating ReFT

raven38:
Paper link: https://arxiv.org/abs/2404.03592

This PR integrates [ReFT](https://github.com/stanfordnlp/pyreft) by creating a new tuner model type. Please see  #1654.

# Changes
I defined a new class `ReFTModel` as a subclass of `LycorisTuner`.
`src/peft/tuners/reft/layer.py` has the actual operations for ReFT methods.

# Status
- [ ] Discuss specification
- [ ] Refactor implementation
- [ ] Test implementation
- [ ] Add documentation for methods.


src/peft/peft_model.py
```diff
@@ -252,7 +254,7 @@ def save_pretrained(
                         output_state_dict[shared_tensor_name] = output_state_dict[shared_tensor_name].clone()
 
                 safe_save_file(
-                    output_state_dict,
+                    {k:v.contiguous() for k,v in output_state_dict.items()},
```
BenjaminBossan, line 257: Why do we need the `contiguous` calls here? There should be a comment to explain this. Do you know if this could affect other models negatively?

src/peft/tuners/reft/config.py
```diff
@@ -0,0 +1,106 @@
...
+from peft.utils import PeftType
+
+@dataclass
+class ReftConfig(LycorisConfig):
+    """
+    This is the configuration class to store the configuration of a [`REFTModel`].
+
+    Args:
+        r (`int`): REFT rank.
+        module_dropout (`int`): The dropout probability for disabling REFT modules during training.
```
BenjaminBossan, line 28: Let's add line breaks after `:` for better readability.

src/peft/tuners/reft/config.py
```diff
@@ -0,0 +1,106 @@
...
+            The mapping from layer names or regexp expression to ranks which are different from the default rank
+            specified by `r`.
+        modules_to_save (`List[str]`):
+            List of modules apart from adapter layers to be set as trainable and saved in the final checkpoint.
+        block_share (`bool`):
+            Whether to share the REFT parameters between blocks or not. This is `False` by default.
+    """
+    r: int = field(default=8, metadata={"help": "REFT rank"})
+    alpha: int = field(default=8, metadata={"help": "Reft alpha"})
+    loc: Optional[Union[List[int], int]] = field(default=None, metadata={"help": "token locations are applied ReFT"})
```
BenjaminBossan, line 55: These two don't have a docstring entry. Also, the description "token locations are applied ReFT" needs to be explained better.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,262 @@
...
+        self.out_features = out_features
+
+
+    @property
+    def _available_adapters(self) -> Set[str]:
+        return {*self.reft_A, *self.reft_R}
+
+    def create_adapter_parameters(self, adapter_name: str, r: int, shape: Tuple[int, ...]):
+        rotate_layer = LowRankRotateLayer(self.out_features, r)
+        self.reft_R[adapter_name] = torch.nn.utils.parametrizations.orthogonal(rotate_layer, orthogonal_map='cayley')
```
BenjaminBossan, line 66: Is this redundant with `torch.nn.init.orthogonal_` above?

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,262 @@
...
+            use_effective_conv2d = use_effective_conv2d and base_layer.kernel_size != (1, 1)
+            if use_effective_conv2d:
+                shape = (base_layer.out_channels, base_layer.in_channels, *base_layer.kernel_size)
+            else:
+                shape = (
+                    base_layer.out_channels,
+                    base_layer.in_channels * base_layer.kernel_size[0] * base_layer.kernel_size[1],
+                )
+        else:
+            raise TypeError(f"LoHa is not implemented for base layers of type {type(base_layer).__name__}")
```
BenjaminBossan, line 127:
```suggestion
            raise TypeError(f"REFT is not implemented for base layers of type {type(base_layer).__name__}")
```

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,262 @@
...
+
+        self.in_features = in_features
+        self.out_features = out_features
+
+
+    @property
+    def _available_adapters(self) -> Set[str]:
+        return {*self.reft_A, *self.reft_R}
+
+    def create_adapter_parameters(self, adapter_name: str, r: int, shape: Tuple[int, ...]):
```
BenjaminBossan, line 64: `shape` is not used.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,262 @@
...
+            raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
+
+        self.r[adapter_name] = r
+        self.loc[adapter_name] = loc
+        self.alpha[adapter_name] = alpha
+        self.scaling[adapter_name] = alpha / r
+        self.rank_dropout[adapter_name] = rank_dropout
+        self.module_dropout[adapter_name] = module_dropout
+
+        # Determine shape of ReFT weights
```
BenjaminBossan, line 113: Since `shape` is not used in `create_adapter_parameters`, all this can be removed, right?

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,262 @@
...
+                    base_layer.in_channels * base_layer.kernel_size[0] * base_layer.kernel_size[1],
+                )
+        else:
+            raise TypeError(f"LoHa is not implemented for base layers of type {type(base_layer).__name__}")
+
+        # Create weights with provided shape
+        self.create_adapter_parameters(adapter_name, r, shape)
+
+        # Initialize weights
+        if init_weights:
```
BenjaminBossan, line 133: Normally, when we have `init_weights=True`, the adapter is initialized as a no-op, e.g. `lora_B` is initialized as zeros. Is the same thing happening with REFT?

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,262 @@
...
+                rotate_layer = self.reft_R[active_adapter]
+                learned_source = self.reft_A[active_adapter]
+                module_dropout = self.module_dropout[active_adapter]
+                if self.loc[active_adapter] is None:
+                    rotated_base = rotate_layer(result)
+                    offset = torch.matmul((learned_source(result) - rotated_base), rotate_layer.weight.T)
+                    output = result + offset
+                else:
+                    loc = self.loc[active_adapter]
+                    selected_results = tensor.gather(result, 1, loc)
```
BenjaminBossan, line 175: Wait, there is `tensor` coming from? Do you mean `torch.gather` or `result.gather`?

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,262 @@
...
+                    rotated_base = rotate_layer(selected_results)
+                    offset = torch.matmul((learned_source(selected_results) - rotated_base), rotate_layer.weight.T)
+                    tensor.scatter_(output, 1, loc, offset)
```
BenjaminBossan, line 178: Same.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,262 @@
...
+import math
+import warnings
+from typing import Any, List, Optional, Set, Tuple
+
+import torch
+import torch.nn as nn
+
+from peft.tuners.lycoris_utils import LycorisLayer, check_adapters_to_merge
+
+class LowRankRotateLayer(torch.nn.Module):
```
BenjaminBossan, line 24: Could this be replaced by an `nn.Linear` with `bias=False` and then just apply the orthogonal initialization on its weight?

Alternatively, we could modify the `forward` to take `rotated_base` and `offset` and perform both operations.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,262 @@
...
+            for active_adapter in self.active_adapters:
+                if active_adapter not in self._available_adapters:
+                    continue
+
+                rotate_layer = self.reft_R[active_adapter]
+                learned_source = self.reft_A[active_adapter]
+                module_dropout = self.module_dropout[active_adapter]
+                if self.loc[active_adapter] is None:
+                    rotated_base = rotate_layer(result)
+                    offset = torch.matmul((learned_source(result) - rotated_base), rotate_layer.weight.T)
```
BenjaminBossan, line 171: So IIUC, for `rotated_base`, we use `reft_A[activate_adapter].weight` and for `offset` we use `reft_A[activate_adapter].weight.T`?

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,262 @@
...
+        init_weights: bool = True,
+        **kwargs,
+    ):
+        super().__init__(base_layer)
+
+        # Create adapter and set it active
+        self._active_adapter = adapter_name
+        self.update_layer(adapter_name, r, alpha, rank_dropout, module_dropout, init_weights, **kwargs)
+
+    def _get_delta_activations(
```
BenjaminBossan, line 204: Is this used anywhere?

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,262 @@
...
+    ):
+        super().__init__(base_layer)
+
+        # Create adapter and set it active
+        self._active_adapter = adapter_name
+        self.update_layer(
+            adapter_name, r, alpha, rank_dropout, module_dropout, init_weights, use_effective_conv2d, **kwargs
+        )
+
+    def _get_delta_activations(
```
BenjaminBossan, line 241: Is this used anywhere?

src/peft/tuners/reft/config.py
```diff
@@ -0,0 +1,106 @@
...
+from dataclasses import dataclass, field
+from typing import List, Optional, Union
+
+from peft.tuners.lycoris_utils import LycorisConfig
+from peft.utils import PeftType
+
+@dataclass
+class ReftConfig(LycorisConfig):
+    """
+    This is the configuration class to store the configuration of a [`REFTModel`].
```
BenjaminBossan, line 24:
```suggestion
    This is the configuration class to store the configuration of a [`ReftModel`].
```

BenjaminBossan (changes requested):
Thanks a lot for providing this PR to add REFT to PEFT. This is already looking really good, well done!

I added a few comments, please take a look.

## General comments

- Please run `make style` on your code.
- We'll have to add some tests. Would you mind looking at the tests that other methods have and add the REFT tests in a similar manner? E.g., check out the tests added by #1301.
- Ideally, we have some docs and examples before merging this.
- In case that you have copied code from pyreft/pyvene, please add comments with references.
- IIUC, the implemented method is `LoReFT` specifically. This should be mentioned. We may also want to rename the names to `LoReFT`, except if you think this could be extended in the future to include more ReFT methods.
- Speaking of names, in the descriptions, you write "REFT" but the authors spell it as "ReFT", let's adjust this. When it comes to variable names, there is no need to change those (so `ReftLayer` etc. can stay as is).

## Hyper params

The authors describe 4 important hyper parameters in 4.1:

1. number of prefix positions
2. number of suffix positions
3. set of layers to target
4. whether to tie intervention parameters

IIUC, 3 exists via `target_modules`. What about 1, 2, and 4?

src/peft/tuners/reft/layer.py
raven38: This is not used. Since LycorisTuner defines `_get_delta_activations` as `abstractmethod`, I define  `_get_delta_activations` to avoid raising errors.

src/peft/tuners/reft/layer.py
raven38: `torch.nn.init.orthogonal_` ensures that the tensor is orthogonal at once, and `torch.nn.utils.parametrizations.orthogonal` ensures that the tensor is orthogonal during the training. 
I checked the impelementation of `torch.nn.utils.register_parametrization`. It calls the a parameterization function (`orthogonal_`) in the initilization process. Thus, `torch.nn.init.orthogonal_` is redundant.

src/peft/tuners/reft/layer.py
raven38: Yes, we can remove them.

src/peft/peft_model.py
raven38: To avoid side effects, I chaged codes to call `contiguous` in `get_peft_model_state_dict`.

raven38:
The authors describe 4 important hyper parameters in 4.1:
Number of prefix positions and suffix positions exists via `loc` parameters. I borrow the pyreft implementation for parsing the positions.
As you say, set of layers to target exists via `target_modules`.
We can implement untied intervention by defining two ReFT models same with pyreft. The default ReFT model is tied intervention.

src/peft/tuners/reft/layer.py
BenjaminBossan: I see. In that case, it should probably not be an abstract method on `LycorsLayer` to begin with, but let's not bother with this here. I'd just raise a `NotImplementedError` then, instead of having dead code.

src/peft/tuners/reft/layer.py
BenjaminBossan: So IIUC, we can remove one of the invocations?

BenjaminBossan (commented):
> Number of prefix positions and suffix positions exists via loc parameters. I borrow the pyreft implementation for parsing the positions.

That makes sense, let's stick close to the original implementation.

> We can implement untied intervention by defining two ReFT models same with pyreft. The default ReFT model is tied intervention.

Sounds good.

There is a merge conflict now in `src/peft/utils/save_and_load.py` but it should be easy to fix.

Let me know if you need more input from my side to work further on this PR.

raven38:
Sorry, I took the wrong operation, which fork sync with discards my commits, closing this pull request. Can I reopen it?

raven38:
I reopen it

src/peft/tuners/reft/layer.py
raven38: Yes, we can remove one of them.

src/peft/tuners/reft/layer.py
raven38: LoReFT always initializes the parameters randomly

raven38:
I think the code is ready for the review.

BenjaminBossan:
@raven38 Thanks for the updates, I'm currently reviewing the PR, but it'll take some time. Meanwhile, could you please run `make style` on your code so that the CI passes?

raven38:
I modified codes so that make style passes, but make style still raises the following error. Should it be fixed by this PR?
```
examples/boft_controlnet/test_controlnet.py:40:1: E402 Module level import not at top of file
Found 1 error.
make: *** [style] Error 1
```

BenjaminBossan:
Hmm, code quality checks still fail:

> ruff src tests examples docs scripts docker
ruff format --check src tests examples docs scripts docker
Would reformat: src/peft/tuners/reft/config.py
Would reformat: src/peft/tuners/reft/layer.py
Would reformat: src/peft/tuners/reft/model.py
Would reformat: tests/test_custom_models.py
Would reformat: tests/test_stablediffusion.py
Would reformat: tests/testing_common.py
6 files would be reformatted, 159 files already formatted

This does not appear to be related to anything in `test_controlnet`, which is independent of your PR. Is it possible that you use a different ruff version? The CI uses v0.2.2.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,195 @@
...
+    def _available_adapters(self) -> Set[str]:
+        return {*self.reft_A, *self.reft_R}
+
+    def create_adapter_parameters(self, adapter_name: str, r: int):
+        rotate_layer = torch.nn.Linear(self.out_features, r, bias=False)
+        self.reft_R[adapter_name] = torch.nn.utils.parametrizations.orthogonal(rotate_layer, orthogonal_map='cayley')
+        self.reft_A[adapter_name] =  torch.nn.Linear(self.out_features, r)
+
+    def merge(self, safe_merge: bool = False, adapter_names: Optional[list[str]] = None) -> None:
+        pass
```
BenjaminBossan, line 73: So merging is not supported, right? If we just pass when the user calls `model.merge()`, they might get the impression that it worked. Instead, we should raise a `ValueError` that merging is not possible with LoReFT. Same for `unmerge`.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,195 @@
...
+        if r <= 0:
+            raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
+
+        self.r[adapter_name] = r
+        first_n, last_n = parse_positions(loc) if loc else (0, 0)
+        self.first_n[adapter_name] = first_n
+        self.last_n[adapter_name] = last_n
+        self.alpha[adapter_name] = alpha
+        self.scaling[adapter_name] = alpha / r
+        self.dropout[adapter_name] = torch.nn.Dropout(dropout)
```
BenjaminBossan, line 104: Let's use `nn.Identity` if dropout is 0.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,195 @@
...
+            nn.init.kaiming_uniform_(self.reft_A[adapter_name].weight, a=math.sqrt(5))
+
+        # Move new weights to device
+        weight = getattr(self.get_base_layer(), "weight", None)
+        if weight is not None:
+            # the layer is already completely initialized, this is an update
+            if weight.dtype.is_floating_point or weight.dtype.is_complex:
+                self.to(weight.device, dtype=weight.dtype)
+            else:
+                self.to(weight.device)
```
BenjaminBossan, line 120: We have a new convenience method for this: `self._move_adapter_to_device_of_base_layer(adapter_name)`.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,195 @@
...
+        self.set_adapter(self.active_adapters)
+
+    def get_delta_weight(self, adapter_name: str) -> torch.Tensor:
+        raise NotImplementedError
```
BenjaminBossan, line 124: Let's add a comment: `# Not needed`.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
+        previous_dtype = x.dtype
+
+        if self.disable_adapters:
+            if self.merged:
+                self.unmerge()
+            output = self.base_layer(x, *args, **kwargs)
+        elif self.merged:
```
BenjaminBossan, line 133: Not really needed, as merging is not supported.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,195 @@
...
+                if active_adapter not in self._available_adapters:
+                    continue
+
+                rotate_layer = self.reft_R[active_adapter]
+                learned_source = self.reft_A[active_adapter]
+                dropout = self.dropout[active_adapter]
+
+                result = result.to(torch.float32)
+                if self.first_n[active_adapter] == 0 and self.last_n[active_adapter] == 0:
+                    rotated_base = rotate_layer(result)
```
BenjaminBossan, line 149: This is eq. 2 from the paper, right? Let's add a comment here to facilitate the understanding.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,195 @@
...
+    def __init__(self, base_layer: nn.Module):
+        super().__init__()
+        LycorisLayer.__init__(self, base_layer)
+
+        # ReFT info
+        self.reft_A = nn.ModuleDict({})
+        self.reft_R = nn.ModuleDict({})
+        self.loc = {}
+        self.first_n = {}
+        self.last_n = {}
```
BenjaminBossan, line 51: I think it's not intuitive what these attributes are, especially since some values seem to have special meaning. Let's document this.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,195 @@
...
+                    continue
+
+                rotate_layer = self.reft_R[active_adapter]
+                learned_source = self.reft_A[active_adapter]
+                dropout = self.dropout[active_adapter]
+
+                result = result.to(torch.float32)
+                if self.first_n[active_adapter] == 0 and self.last_n[active_adapter] == 0:
+                    rotated_base = rotate_layer(result)
+                    offset = torch.matmul((learned_source(result) - rotated_base), rotate_layer.weight)
```
BenjaminBossan, line 150: Instead of calling `torch.matmul`, can we not call `rotate_layer(learned_source(result) - rotated_base)`? Generally, it's better to use `forward` than to use the parameters directly.

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,195 @@
...
+                    output = result + offset
+                else:
+                    first_n = self.first_n[active_adapter]
+                    last_n = self.last_n[active_adapter]
+                    loc = torch.cat([torch.arange(first_n), torch.arange(result.shape[1]-last_n, result.shape[1])])
+                    selected_results = torch.gather(result, 1, loc)
+                    rotated_base = rotate_layer(selected_results)
+                    offset = torch.matmul((learned_source(selected_results) - rotated_base), rotate_layer.weight)
```
BenjaminBossan, line 158: Same argument as above regarding `forward`.

src/peft/tuners/reft/config.py
```diff
@@ -0,0 +1,100 @@
...
+@dataclass
+class LoReftConfig(LycorisConfig):
+    """
+    This is the configuration class to store the configuration of a [`LoReftModel`].
+
+    Args:
+        r (`int`): LoReFT rank.
+        alpha (`int`): LoReFT alpha.
+        loc (`Optional[str]`):
+            Token locations are applied LoReFT.
```
BenjaminBossan, line 31: This needs much better explanation, ideally with a few examples. Also, it should be mentioned that this only really makes sense for language models, right?

src/peft/tuners/reft/layer.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def _get_delta_activations(
+        self, adapter_name: str, input: torch.Tensor, *args: Any, **kwargs: Any
+    ) -> torch.Tensor:
+        raise NotImplementedError
+
+
+    def __repr__(self) -> str:
+        rep = super().__repr__()
+        return "reft." + rep
```
BenjaminBossan, line 192:
```suggestion
        return "loreft." + rep
```

In case we add more ReFT layers in the future. This is not consistent with the module name, but I think that's not crucial.

src/peft/tuners/reft/model.py
```diff
@@ -0,0 +1,115 @@
...
+from peft.utils import (
+    TRANSFORMERS_MODELS_TO_LOREFT_TARGET_MODULES_MAPPING,
+)
+
+from .layer import Linear, LoReftLayer
+
+
+class LoReftModel(LycorisTuner):
+    """
+    LoReFT model from a pretrained model.
```
BenjaminBossan, line 30: Let's add links to the paper and the reference repo.

src/peft/tuners/reft/config.py
```diff
@@ -0,0 +1,100 @@
...
+            "For example, in Sequence Classification or Token Classification tasks, "
+            "the final layer `classifier/score` are randomly initialized and as such need to be trainable and saved."
+        },
+    )
+
+    block_share: bool = field(
+        default=False,
+        metadata={"help": "Whether to share the ReFT parameters between blocks or not."},
+    )
+
```
BenjaminBossan, line 95: So right now, LoReFT has no `init_weights` argument, unlike almost all other adapters. I think it would be useful to have this. I think you mentioned somewhere that by default, LoReFT is _not_ initialized to be the identity transform. In that case, we could set the default value to `False` (which in PEFT means _not_ identity), but it would still be nice for users to have the option to initialize as identity by passing `init_weights=True`.

BenjaminBossan (changes requested):
Thanks for all the updates. This looks very promising, and I don't think we need any major changes.

Before merging, we should add examples and docs, but we can tackle this after we have streamlined the implementation.

Regarding the intervention locations, your implementation departs from the pyreft implementation. Here, it is basically a hyper-parameter that is passed to the config. That means that for a given model, these values are fixed, right? In pyreft, it is part of the dataset, so it can vary from sample to sample.

I wonder if we should not stick with the pyreft implementation. This means that we can remove this parameter completely from PEFT. Instead, we can check the arguments passed to `forward`. IIUC, a pyreft dataset would contain an additional key, `intervention_locations`. If this is passed, we could use it, otherwise intervene on all positions. This would also mean that the PEFT implementation could use the datasets from pyreft, right? Maybe I'm missing something and there is a disadvantage in my suggestion, please LMK.

Moreover, in the paper, the authors discuss a variant with tied weights. AFAICT, there is currently no option to configure tied vs untied. Would it be possible to implement this?

frankaging:
hey! thanks again for the PR (i am one of the authors of the ReFT paper). 

a quick question: with the current peft config, could we target any residual stream with loreft? if so, what will the config look like? thanks!

BenjaminBossan:
@frankaging Thanks a lot for taking a look. Regarding your questions, I'll let @raven38 provide the final answer. Personally, I think 2 is right and I don't quite understand 1, maybe you can give an example?

frankaging:
> @frankaging Thanks a lot for taking a look. Regarding your questions, I'll let @raven38 provide the final answer. Personally, I think 2 is right and I don't quite understand 1, maybe you can give an example?

Thanks for the update! For (1), i was wondering: for example, when targeting the attention output matrix for LoRA of a Llama model, we might have a config like this:
```py

peft_config = LoraConfig(
    r=4, lora_alpha=32, target_modules=["o_proj"], layers_to_transform=[15],
    use_rslora=True, lora_dropout=0.05, bias="none", task_type="CAUSAL_LM"
)
```
where `target_modules=["o_proj"]`. Since for LoReFT, we want to target the residual stream (transformer block/layer output), what should we put for the `target_modules` in that case? Thanks!

BenjaminBossan:
> Since for LoReFT, we want to target the residual stream (transformer block/layer output), what should we put for the `target_modules` in that case? Thanks!

I see, I think I get what you mean. Good question. I think as is, it would not work. We would have to target the attention module itself (the type depending on what kind of attention is being used), targeting a `Linear` layer would not work. I think this is what you're getting at, right?

If you have some pointers on how this is implemented in pyreft/pyvene, please share.

frankaging:
> > Since for LoReFT, we want to target the residual stream (transformer block/layer output), what should we put for the `target_modules` in that case? Thanks!
> 
> I see, I think I get what you mean. Good question. I think as is, it would not work. We would have to target the attention module itself (the type depending on what kind of attention is being used), targeting a `Linear` layer would not work. I think this is what you're getting at, right?
> 
> If you have some pointers on how this is implemented in pyreft/pyvene, please share.

Thanks for your reply. Yes, we need to target the whole block module itself (e.g., `model.layers[15]`). Will this be doable? If not, I think it is still valuable to have the current form where we target submodules (e.g., attention output or MLP output).


With `pyreft` or `pyvene`, we use [`register_module_forward_hook`](https://pytorch.org/docs/stable/generated/torch.nn.modules.module.register_module_forward_hook.html) to place a callback to the existing computation graph, instead of overwriting current modules with the additional adapter modules. This is quite different, so it might not be that helpful here.

BenjaminBossan:
> Thanks for your reply. Yes, we need to target the whole block module itself (e.g., `model.layers[15]`). Will this be doable?

Yes, I don't see why not. We would have to add a new layer type, similar to `class Linear(LoReftLayer)`, but for the whole attention block layer. The issue there is probably how well this generalizes across different attention types and architectures. With `Linear`, we know they're used everywhere, but there are quite a few different attention layers implementations. So we'd probably have to focus on a few popular ones, like `LlamaAttention`.

> With `pyreft` or `pyvene`, we use [`register_module_forward_hook`](https://pytorch.org/docs/stable/generated/torch.nn.modules.module.register_module_forward_hook.html) to place a callback to the existing computation graph, instead of overwriting current modules with the additional adapter modules.

We do use forward hooks in PEFT when necessary but if there is an easier way, we generally prefer that. Do you have a pointer in your repos on how it's implemented?

frankaging:
@BenjaminBossan Thanks! It would be great if we could support LlamaAttention on residual stream intervention.

On hook-based impl: ReFT is one of the use-cases of our `pyvene` library (and actually a simple use-case). It might be hard for me to lay out the full details of `pyvene`. But, we essentially, get a intervention config from the user (e.g., pointing out which layer, which stream, what rank, etc..), and then we manage these interventions for users by registering hooks (and removing hooks) onto the model, and memorizing activations (e.g., https://github.com/stanfordnlp/pyvene/blob/main/pyvene/models/intervenable_base.py#L963 <- this is line of registering our hooks).

For a residual stream intervention with `r=4` on layer 15 of a llama-based LM, the config will be something like:
```py
reft_config = pyreft.ReftConfig(representations={
    "component": "model.layers[15].output",
    "intervention": pyreft.LoreftIntervention(embed_dim=model.config.hidden_size,
    low_rank_dimension=4)})
reft_model = pyreft.get_reft_model(model, reft_config)
```

And then internally, we basically call `model.layers[15].register_module_forward_hook(...)` with the intervention torch module.

BenjaminBossan:
Thanks for the pointer (here is the [permalink](https://github.com/stanfordnlp/pyvene/blob/ffad51a169d0dc76f020da0d62fc14ee82732746/pyvene/models/intervenable_base.py#L963) just in case). Let's wait for @raven38's opinion is on this.

aryamanarora:
@BenjaminBossan (ReFT author here) just saw that this was never merged. what exactly is missing? happy to help get the PR ready

BenjaminBossan:
Thanks for the offer @aryamanarora. It appears that the original author did not have time to finish this PR (or lost interest in it). As for the state of the PR, please check the discussion above with @frankaging. The PR would probably also need a refresh, as I'll expect there will be merge conflicts.

If you want to pick this up, it's probably easiest if you create a new PR and copy the changes over. I'd also be open to an implementation that uses pyvene as a dependency under the hood if it makes life easier.

Issue #1567, opened 2024-03-16
Base Model Revision


mnoukhov:
### Feature request

Currently, the `adapter_config` only stores the `base_model_name_or_path` but not the `revision`. This means that if a `PeftModel` is trained on top of a model with a `revision` then loading `PeftModel.from_pretrained` will load the `main` revision of the base model, not the correct revision. 

We should add a `base_model_revision` to the `adapter_config` so that we can correctly load the base model 

### Motivation

This will cause a bug if using `PeftModel.from_pretrained` when the base model has a revision.  

### Your contribution

I can try to submit a PR if I have time in 2-3 weeks

BenjaminBossan:
Thanks for this suggestion, it sounds good. A PR would be great, but no need to rush it.

mnoukhov:
I looked into it and found that there's already actually a `revision` parameter 
https://github.com/huggingface/peft/blob/56773b9a92b141111d65fe3548d0c30233358868/src/peft/config.py#L235

It's just not being used when calling `from_pretrained` here https://github.com/huggingface/peft/blob/56773b9a92b141111d65fe3548d0c30233358868/src/peft/auto.py#L104

I've opened the PR #1658 but in order to do unit tests, I need a model with a base model with revision. If you don't mind this not being on `hf-internal-testing`, I can upload it to my personal account, but maybe you'd prefer the former. Let me know @BenjaminBossan

BenjaminBossan:
_not stale_

mnoukhov:
resolved by #1658

PR #1658, opened 2024-04-16
save and load base model with revision

mnoukhov:
addresses #1567 

changed name of parameter from `revision` to `base_model_revision` for clarity

can add unit tests if someone gives pointers to similar ones to extend
may need to add a model to `hf-internal-testing` with a base model with revision for these unit tests

### DIFF
--- a/src/peft/auto.py
+++ b/src/peft/auto.py
@@ -62,7 +62,7 @@ class _BaseAutoPeftModel:
         adapter_name: str = "default",
         is_trainable: bool = False,
         config: Optional[PeftConfig] = None,
-        revision: Optional[str] = "main",
+        revision: Optional[str] = None,
         **kwargs,
     ):
         r"""
--- a/src/peft/auto.py
+++ b/src/peft/auto.py
@@ -71,7 +71,7 @@ class _BaseAutoPeftModel:
         """
         peft_config = PeftConfig.from_pretrained(pretrained_model_name_or_path, **kwargs)
         base_model_path = peft_config.base_model_name_or_path
-        base_model_revision = peft_config.base_model_revision
+        base_model_revision = peft_config.revision
 
         task_type = getattr(peft_config, "task_type", None)
 
--- a/src/peft/auto.py
+++ b/src/peft/auto.py
@@ -71,6 +71,7 @@ class _BaseAutoPeftModel:
         """
         peft_config = PeftConfig.from_pretrained(pretrained_model_name_or_path, **kwargs)
         base_model_path = peft_config.base_model_name_or_path
+        base_model_revision = peft_config.base_model_revision
 
         task_type = getattr(peft_config, "task_type", None)
 
@@ -101,7 +102,7 @@ class _BaseAutoPeftModel:
                 "Cannot infer the auto class from the config, please make sure that you are loading the correct model for your task type."
             )
 
-        base_model = target_class.from_pretrained(base_model_path, **kwargs)
+        base_model = target_class.from_pretrained(base_model_path, revision=base_model_revision, **kwargs)
 
         tokenizer_exists = False
         if os.path.exists(os.path.join(pretrained_model_name_or_path, TOKENIZER_CONFIG_NAME)):
--- a/src/peft/auto.py
+++ b/src/peft/auto.py
@@ -62,6 +62,7 @@ class _BaseAutoPeftModel:
         adapter_name: str = "default",
         is_trainable: bool = False,
         config: Optional[PeftConfig] = None,
+        revision: Optional[str] = "main",
         **kwargs,
     ):
         r"""
@@ -69,7 +70,7 @@ class _BaseAutoPeftModel:
         are passed along to `PeftConfig` that automatically takes care of filtering the kwargs of the Hub methods and
         the config object init.
         """
-        peft_config = PeftConfig.from_pretrained(pretrained_model_name_or_path, **kwargs)
+        peft_config = PeftConfig.from_pretrained(pretrained_model_name_or_path, revision=revision, **kwargs)
         base_model_path = peft_config.base_model_name_or_path
         base_model_revision = peft_config.revision
 
@@ -115,7 +116,7 @@ class _BaseAutoPeftModel:
             tokenizer_exists = check_file_exists_on_hf_hub(
                 repo_id=pretrained_model_name_or_path,
                 filename=TOKENIZER_CONFIG_NAME,
-                revision=kwargs.get("revision", None),
+                revision=revision,
                 repo_type=kwargs.get("repo_type", None),
                 token=token,
             )
--- a/src/peft/config.py
+++ b/src/peft/config.py
@@ -97,6 +97,7 @@ class PeftConfigMixin(PushToHubMixin):
         # TODO: this hack is needed to fix the following issue (on commit 702f937):
         # if someone saves a default config and loads it back with `PeftConfig` class it yields to
         # not loading the correct config class.
+        #
         # from peft import AdaLoraConfig, PeftConfig
         # peft_config = AdaLoraConfig()
         # print(peft_config)
--- a/src/peft/config.py
+++ b/src/peft/config.py
@@ -97,7 +97,6 @@ class PeftConfigMixin(PushToHubMixin):
         # TODO: this hack is needed to fix the following issue (on commit 702f937):
         # if someone saves a default config and loads it back with `PeftConfig` class it yields to
         # not loading the correct config class.
-
         # from peft import AdaLoraConfig, PeftConfig
         # peft_config = AdaLoraConfig()
         # print(peft_config)
@@ -232,7 +231,9 @@ class PeftConfig(PeftConfigMixin):
     base_model_name_or_path: Optional[str] = field(
         default=None, metadata={"help": "The name of the base model to use."}
     )
-    revision: Optional[str] = field(default=None, metadata={"help": "The specific model version to use."})
+    base_model_revision: Optional[str] = field(
+        default=None, metadata={"help": "The specific base model version to use."}
+    )
     peft_type: Optional[Union[str, PeftType]] = field(default=None, metadata={"help": "Peft type"})
     task_type: Optional[Union[str, TaskType]] = field(default=None, metadata={"help": "Task type"})
     inference_mode: bool = field(default=False, metadata={"help": "Whether to use inference mode"})
--- a/src/peft/config.py
+++ b/src/peft/config.py
@@ -231,7 +231,7 @@ class PeftConfig(PeftConfigMixin):
     base_model_name_or_path: Optional[str] = field(
         default=None, metadata={"help": "The name of the base model to use."}
     )
-    revision: Optional[str] = field(default=None, metadata={"help": "The specific base model version to use."})
+    revision: Optional[str] = field(default="main", metadata={"help": "The specific base model version to use."})
     peft_type: Optional[Union[str, PeftType]] = field(default=None, metadata={"help": "Peft type"})
     task_type: Optional[Union[str, TaskType]] = field(default=None, metadata={"help": "Task type"})
     inference_mode: bool = field(default=False, metadata={"help": "Whether to use inference mode"})
--- a/src/peft/config.py
+++ b/src/peft/config.py
@@ -232,7 +232,7 @@ class PeftConfig(PeftConfigMixin):
     base_model_name_or_path: Optional[str] = field(
         default=None, metadata={"help": "The name of the base model to use."}
     )
-    revision: Optional[str] = field(default="main", metadata={"help": "The specific base model version to use."})
+    revision: Optional[str] = field(default=None, metadata={"help": "The specific base model version to use."})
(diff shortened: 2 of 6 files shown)


BenjaminBossan:
Thanks a lot for this PR.

> can add unit tests if someone gives pointers to similar ones to extend

We don't really have a single place where we test this functionality, I think it's fine to add the tests to the `tests/test_hub_features` file even if it's not a perfect fit. When it comes to what to test, maybe you can proceed as follows:

1. Load the model with revision
2. Add an adapter on top, like LoRA (use `init_lora_weights=False`)
3. Save the adapter in a temp directory (use the `tmp_path` fixture from pytest)
4. Load the base model using the revision from `adapter_config`
5. Load the LoRA adapter
6. Show that the outputs are the same
7. As a sanity check, also load the adapter with the base model without revision and show that the outputs are different

> may need to add a model to `hf-internal-testing` with a base model with revision for these unit tests

I added a model here: `peft-internal-testing/tiny-random-BertModel`. There is a revision `v2.0.0` with updated weights:

```python
>>> model = AutoModelForCausalLM.from_pretrained("peft-internal-testing/tiny-random-BertModel").eval()
>>> model(torch.arange(10).reshape(-1, 1)).logits.sum()
tensor(27.5802, grad_fn=<SumBackward0>)
>>> model_rev = AutoModelForCausalLM.from_pretrained("peft-internal-testing/tiny-random-BertModel", revision="v2.0.0").eval()
model_rev(torch.arange(10).reshape(-1, 1)).logits.sum()
tensor(-166.8932, grad_fn=<SumBackward0>)
```

BenjaminBossan:
@mnoukhov Let me know when the PR is ready for review.

mnoukhov:
I ended up changing `base_model_revision` back to `revision` to maintain backwards compatibility as some models were already uploaded with it, including some models for testing. 

I can change it back and add a little code to account for having `revision` instead of `base_model_revision`, if you think that would be better.

tests/test_hub_features.py
```diff
@@ -13,13 +13,16 @@
...
+import torch
 from transformers import AutoModelForCausalLM
 
-from peft import PeftConfig, PeftModel
+from peft import AutoPeftModelForCausalLM, LoraConfig, PeftConfig, PeftModel, get_peft_model
 
 
 PEFT_MODELS_TO_TEST = [("peft-internal-testing/test-lora-subfolder", "test")]
 
+BASE_REVISION_MODELS_TO_TEST = [("peft-internal-testing/tiny-random-BertModel", "v2.0.0")]
```
BenjaminBossan, line 24: I know you wanted to be consistent here, but since it's only one test, I think there is no need to have a list here and loop below.

tests/test_hub_features.py
```diff
@@ -35,3 +38,34 @@ def test_subfolder(self):
...
 
             assert isinstance(model, PeftModel)
+
+
+class TestBaseModelRevision:
+    def test_save_and_load_base_model_revision(self, tmp_path):
+        r"""
+        Test if subfolder argument works as expected
+        """
+        lora_config = LoraConfig(r=8, lora_alpha=16, lora_dropout=0.0, init_lora_weights=False)
```
BenjaminBossan, line 48: I tried this test on the current main and it succeeds. It's thus not testing what it's supposed to test.

One issue is that without setting a seed, calling `get_peft_model` twice does not yield the same model, so that part of the test passes trivially.

I tried rewriting the test like this:

```python
        lora_config = LoraConfig(r=8, lora_alpha=16, lora_dropout=0.0, init_lora_weights=False)
        test_inputs = torch.arange(10).reshape(-1, 1)

        base_model_revision = AutoModelForCausalLM.from_pretrained("peft-internal-testing/tiny-random-BertModel", revision="v2.0.0").eval()
        torch.manual_seed(0)
        peft_model_revision = get_peft_model(base_model_revision, lora_config)
        output = peft_model_revision(test_inputs).logits

        # sanity check: the model without revision should be different
        base_model_no_revision = AutoModelForCausalLM.from_pretrained("peft-internal-testing/tiny-random-BertModel").eval()
        torch.manual_seed(0)
        peft_model_no_revision = get_peft_model(base_model_no_revision, lora_config)
        output_no_revision = peft_model_no_revision(test_inputs).logits
        assert not torch.allclose(output, output_no_revision)

        # check that if we save and load the model, the output corresponds to the one with revision
        peft_model_revision.save_pretrained(tmp_path / "base_model_revision")
        peft_model_loaded = AutoPeftModelForCausalLM.from_pretrained(tmp_path / "base_model_revision").eval()
        assert peft_model_loaded.peft_config["default"].revision == "v2.0.0"

        output_loaded = peft_model_loaded(test_inputs).logits
        assert torch.allclose(output, output_loaded)

```

Does this better capture what we want to test, WDYT? Note, however, that this test still fails for me even on your branch, because the stored `adapter_config.json` does not seem to contain the revision.

tests/test_hub_features.py
```diff
@@ -35,3 +38,34 @@ def test_subfolder(self):
...
+        test_inputs = torch.arange(10).reshape(-1, 1)
+
+        for model_id, revision in BASE_REVISION_MODELS_TO_TEST:
+            original_base_model = AutoModelForCausalLM.from_pretrained(model_id, revision="main").eval()
+            original_peft_model = get_peft_model(original_base_model, lora_config)
+            original_peft_sum = original_peft_model(test_inputs).logits.sum()
```
BenjaminBossan, line 54: Not really necessary to take the sum here, we can compare the logits directly.

BenjaminBossan (changes requested):
Thanks for the updates. I agree that keeping `revision` at this point makes more sense for backwards compatibility.

I think we have to work a bit on the test, please check out my comment there.

tests/test_hub_features.py
mnoukhov: Those are good points, thank you, I almost all the suggestions here.

I decided to remove `init_lora_weights=False` so it results in an identity transform and then we don't need to do `torch.manual_seed` as the peft model is the same as the base model, removing stochasticity

tests/test_hub_features.py
mnoukhov: done

mnoukhov:
I changed the tests but more importantly, realized there is a big problem.

For some reason, I assumed that `revision` was a property of `AutoModel` or `PretrainedConfig` but it turns out it isn't. 

`name_or_path` is stored https://github.com/huggingface/transformers/blob/8c12690cecbb97e187861e386f7a0ac790e4236c/src/transformers/configuration_utils.py#L351

but `revision` is only used to load the model from the hub.

This feature / fix therefore requires either
1. the user passing in the `base_model_revision` to `get_peft_config` or `PeftModel.from_pretrained` or 
2. adding `revision` to  `PretrainedConfig` in transformers

I think the second makes more sense, but that means this is blocked until this is added to transformers. What do you think we should do?

BenjaminBossan:
> * the user passing in the `base_model_revision` to `get_peft_config` or `PeftModel.from_pretrained` or
> * adding `revision` to  `PretrainedConfig` in transformers

I don't think that the latter would make sense. The revision is not really a model config, the same config will typically apply to all revisions. So that only really leaves the first option.

mnoukhov:
Added the `revision` argument to `get_peft_model` and tests are now working as intended

One small thing I've noticed is that the `peft_config` passed to `get_peft_model` is modified in place and then set in the config so maybe we should be doing `copy.copy` for the config in `get_peft_model` to avoid issues like using the config twice and overwriting the revision the second time?

src/peft/mapping.py
```diff
@@ -118,13 +123,23 @@ def get_peft_model(
...
+            The revision of the base model. If this isn't set, the saved peft model will load the `main` revision for
+            the base model
     """
     model_config = getattr(model, "config", {"model_type": "custom"})
     if hasattr(model_config, "to_dict"):
         model_config = model_config.to_dict()
 
     peft_config.base_model_name_or_path = model.__dict__.get("name_or_path", None)
 
+    if revision is not None:
```
BenjaminBossan, line 136: How about adding a test to check for this warning?

src/peft/auto.py
```diff
@@ -101,7 +102,7 @@ def from_pretrained(
                 "Cannot infer the auto class from the config, please make sure that you are loading the correct model for your task type."
             )
 
-        base_model = target_class.from_pretrained(base_model_path, **kwargs)
+        base_model = target_class.from_pretrained(base_model_path, revision=base_model_revision, **kwargs)
```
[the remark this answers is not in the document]
mnoukhov: That's a good point. It shouldn't affect anything but I'll change all revision defaults from `None` to `"main"` for consistency

src/peft/config.py
```diff
@@ -97,7 +97,6 @@ def from_peft_type(cls, **kwargs):
         # TODO: this hack is needed to fix the following issue (on commit 702f937):
         # if someone saves a default config and loads it back with `PeftConfig` class it yields to
         # not loading the correct config class.
-
```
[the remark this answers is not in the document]
mnoukhov: I must be doing something wrong, but I kept having this autoformat, so I added a comment and it fixed it

src/peft/mapping.py
```diff
@@ -104,7 +105,11 @@ def get_peft_config(config_dict: dict[str, Any]) -> PeftConfig:
 
 
 def get_peft_model(
-    model: PreTrainedModel, peft_config: PeftConfig, adapter_name: str = "default", mixed: bool = False
+    model: PreTrainedModel,
+    peft_config: PeftConfig,
+    adapter_name: str = "default",
+    mixed: bool = False,
+    revision: str = None,
```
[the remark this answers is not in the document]
mnoukhov: done

src/peft/mapping.py
mnoukhov: added

mnoukhov:
In doing the last changes, I discovered there was a problem if both the base model and the peft model have a revision. In that case, call `AutoPeftModel.from_pretrained(revision="foo")` while also having `peft_config.revision="bar"` 

would lead to an error in `base_model = target_class.from_pretrained(base_model_path, revision=base_model_revision, **kwargs)` as `revision` is both explicitly an argument and part of `kwargs`. 

I fixed it, but in order to have a test for this, I would probably need a LoRA model on the hub with a revision different to that of its base model. An example would be an adapter with a `main` revision on top of the base model `peft-internal-testing/tiny-random-BertModel` with revision `v2.0.0`

BenjaminBossan:
Thanks for discovering and fixing that issue. I created a LoRA adapter with and without revision:

```python
model = AutoModelForCausalLM.from_pretrained("hf-internal-testing/tiny-random-BertModel").eval()
# without revision
model = PeftModel.from_pretrained(model, "peft-internal-testing/tiny-random-BertModel-lora")
# with revision
model = PeftModel.from_pretrained(model, "peft-internal-testing/tiny-random-BertModel-lora", revision="v1.2.3")
```

I don't think we have any way of loading a PEFT model with revision directly using `AutoModelForXxx.from_pretrained("peft-internal-testing/tiny-random-BertModel-lora", revision=...)` as the revision would be interpreted as the base model revision, not the PEFT adapter revision.

Regarding what I said earlier:

> I have a small concern here, in PEFT configs, the revision parameter defaults to None but in transformers, it defaults to main

I'm doubting now if the proposed change is better. It is still true, but so far, all adapters have `revision: null` by default in their `adapter_config.json`. With this change, it would be `revision: main`. My concern is that this could have some unforeseen consequences and would rather stick with `null/None` for consistency. WDYT?

mnoukhov:
I've added the test with differing peft and base model revisions. The one caveat is that the peft config you've uploaded has a `revision: null`. If you decide to change this, we should update the test.
  
> I'm doubting now if the proposed change is better. It is still true, but so far, all adapters have revision: null by default in their adapter_config.json. With this change, it would be revision: main. My concern is that this could have some unforeseen consequences and would rather stick with null/None for consistency. WDYT?

The only issue I can forsee is if the default branch name changes in the future i.e. github changing from `master` to `main`. It would be annoying to deal with. 

On a more conceptual basis, I think I prefer `null` since it wasn't specified by the user, but it ends up being inconsistent with transformers since the default in methods is now `revision=main` so either way works

BenjaminBossan:
> The only issue I can forsee is if the default branch name changes in the future i.e. github changing from `master` to `main`. It would be annoying to deal with.
> 
> On a more conceptual basis, I think I prefer `null` since it wasn't specified by the user, but it ends up being inconsistent with transformers since the default in methods is now `revision=main` so either way works

I agree with your assessment. In the end, I think I'd also prefer `null`, even if the scenario you mention could get annoying. At least it will be _consistently_ annoying for all PEFT models, instead of having two different situations.

mnoukhov:
Default reverted to `None` and I also caught an issue where we threw the overwrite warning when `revision` and `peft_config.revision` were the same

BenjaminBossan (approved):
Thanks for all the changes and your patience, this now looks good to be merged.

I'll wait with the merge for after the next PEFT release (likely very soon), just to be 100% sure, since these types of changes always have a small chance to miss some edge case.

BenjaminBossan:
@mnoukhov PEFT release is out, merging this now. Again, thanks for the PR.

PR #1660, opened 2024-04-17
Remove `dreambooth` Git link

charliermarsh:
## Summary

`examples/boft_dreambooth/data/dreambooth` is declared as a submodule, but there's no `.gitmodules` mapping for it, so various Git operations will fail on the repo.

`data` is actually already `.gitignored`, so it seems reasonable to just remove this submodule entirely, since cloning the repo (which is handled by the training script) won't lead to a dirty Git history anyway.

The alternative solution is to add a `.gitmodules` entry:

```
[submodule "dreambooth"]
    path = examples/boft_dreambooth/data/dreambooth
    url = https://github.com/google/dreambooth
```

The downside here is that you'll now be downloading the dataset whenever someone performs a recursive clone on the repo, which seems _not_ to be the intent, since you're downloading the repo out-of-band from Git via the training script anyway.


BenjaminBossan (approved):
Thanks for fixing this. This example was originally structured differently so that it used the sub-module. Later, that was changed, but this wasn't cleaned up.

PR #1663, opened 2024-04-18
FEAT Show adapter layer and model status

BenjaminBossan:
This PR adds a new feature to PEFT models that allows to better understand the status of adapter(s) on the model. Quoting from the doc entry that I added:

> Sometimes, the PEFT model can end up in a bad state, especially when handling multiple adapters. There can be some confusion around what adapters exist, which one is active, which one is merged, etc. To help investigate this issue, you can call the `peft.PeftModel.get_layer_status` and the `peft.PeftModel.get_model_status` methods. The first one gives you a detailed overview about the adapters for each targeted layer. The latter one gives you a high-level overview about the model status.

**Example PEFT**:

```python
>>> from transformers import AutoModel
>>> from peft import get_peft_model, LoraConfig

>>> model_id = "google/flan-t5-small"
>>> model = AutoModel.from_pretrained(model_id)
>>> model = get_peft_model(model, LoraConfig())

>>> model.get_layer_status()
[TunerLayerStatus(name='model.encoder.block.0.layer.0.SelfAttention.q',
                  module_type='lora.Linear',
                  enabled=True,
                  active_adapters=['default'],
                  merged_adapters=[],
                  requires_grad={'default': True},
                  available_adapters=['default']),
 TunerLayerStatus(name='model.encoder.block.0.layer.0.SelfAttention.v',
                  module_type='lora.Linear',
                  enabled=True,
                  active_adapters=['default'],
                  merged_adapters=[],
                  requires_grad={'default': True},
                  available_adapters=['default']),
...]

>>> model.get_model_status()
TunerModelStatus(
    base_model_type='T5Model',
    adapter_model_type='LoraModel',
    peft_types={'default': 'LORA'},
    trainable_params=344064,
    total_params=60855680,
    num_adapter_layers=48,
    enabled=True,
    active_adapters=['default'],
    merged_adapters=[],
    available_adapters=['default'],
)

>>> model.add_adapter("other", LoraConfig())
>>> model.merge_adapter(["default"])
>>> model.get_model_status()
TunerModelStatus(
    base_model_type='T5Model',
    adapter_model_type='LoraModel',
    peft_types={'default': 'LORA', 'other': 'LORA'},
    trainable_params=344064,
    total_params=61199744,
    num_adapter_layers=48,
    enabled=True,
    active_adapters=['default'],
    merged_adapters=['default'],
    requires_grad={'default': True},
    available_adapters=['default', 'other'],
)
```

**Example transformers:**

This also works when loading a transformers model with PEFT adapters injected directly. However, that requires importing the "private" `get_model_status`  and `get_layer_status` functions:

```python
>>> from peft.peft_model import get_model_status, get_layer_status

>>> peft_model_id = "ybelkada/opt-350m-lora"
>>> model = AutoModelForCausalLM.from_pretrained(peft_model_id)
>>> get_layer_status(model)
[TunerLayerStatus(name='model.decoder.layers.0.self_attn.v_proj',
                  module_type='lora.Linear',
                  enabled=True,
                  active_adapters=['default'],
                  merged_adapters=[],
                  requires_grad={'default': False},
                  available_adapters=['default']),
 TunerLayerStatus(name='model.decoder.layers.0.self_attn.q_proj',
                  module_type='lora.Linear',
                  enabled=True,
                  active_adapters=['default'],
                  merged_adapters=[],
                  requires_grad={'default': False},
                  available_adapters=['default']),
...]
>>> get_model_status(model)
TunerModelStatus(
    base_model_type='OPTForCausalLM',
    adapter_model_type='None',
    peft_types={},
    trainable_params=0,
    total_params=332769280,
    num_adapter_layers=48,
    enabled=True,
    active_adapters=['default'],
    merged_adapters=[],
    requires_grad={'default': False},
    available_adapters=['default'],
)
```

Maybe if we find this useful enough, we can make these functions "public".

**Also changed in this PR**:

- Fix a bug that when adapters where disabled, when entering and exiting  the `disable_adapter` context, suddenly all these adapters are enabled. This is a changing behavior and should be highlighted in the next release notes. Fixing this bug was the initial motivation that led me to work on this this feature.
- Add a warning when using the `disable_adapter` context and having irregular activation of adapters, as the correct status cannot be restored (but behavior stays the same as currently).
- Add abstract methods for `disable_adapter_layers` and `enable_adapter_layers` (those are required for subclasses)
- Add a check in `check_adapters_to_merge` for `str`.
- Fix a bug when adding a 2nd adapter that targets different layers than the existing adapter, resulting in those layers being active. As an example, say I create the "default" LoRA adapter and it targets layers 0 and 1. Then I call `add_adapter("other", config2)` targeting layers 1 and 2. The "other" adapter should be inactive. This is true for layer 1, where both "default" and "other" exist, but for layer 2, where only "other" exists, it would currently be active. With the fix, it will be inactive.

### DIFF
--- a/docs/source/developer_guides/troubleshooting.md
+++ b/docs/source/developer_guides/troubleshooting.md
@@ -156,12 +156,14 @@ The [`~peft.PeftModel.get_layer_status`] method gives you a detailed overview of
                   enabled=True,
                   active_adapters=['default'],
                   merged_adapters=[],
+                  requires_grad={'default': True},
                   available_adapters=['default']),
  TunerLayerStatus(name='model.encoder.block.0.layer.0.SelfAttention.v',
                   module_type='lora.Linear',
                   enabled=True,
                   active_adapters=['default'],
                   merged_adapters=[],
+                  requires_grad={'default': True},
                   available_adapters=['default']),
 ...]
 
@@ -176,6 +178,7 @@ TunerModelStatus(
     enabled=True,
     active_adapters=['default'],
     merged_adapters=[],
+    requires_grad={'default': True},
     available_adapters=['default'],
 )
 ```
--- a/docs/source/developer_guides/troubleshooting.md
+++ b/docs/source/developer_guides/troubleshooting.md
@@ -136,9 +136,11 @@ For inference, load the base model first and resize it the same way you did befo
 
 For a complete example, please check out [this notebook](https://github.com/huggingface/peft/blob/main/examples/causal_language_modeling/peft_lora_clm_with_additional_tokens.ipynb).
 
-### Investigating the status of the PEFT model
+### Check layer and model status
 
-Sometimes, the PEFT model can end up in a bad state, especially when handling multiple adapters. There can be some confusion around what adapters exist, which one is active, which one is merged, etc. To help investigate this issue, you can call the [`~peft.PeftModel.get_layer_status`] and the [`~peft.PeftModel.get_model_status`] methods. The first one gives you a detailed overview about the adapters for each targeted layer. The latter one gives you a high-level overview about the model status.
+Sometimes a PEFT model can end up in a bad state, especially when handling multiple adapters. There can be some confusion around what adapters exist, which one is active, which one is merged, etc. To help investigate this issue, call the [`~peft.PeftModel.get_layer_status`] and the [`~peft.PeftModel.get_model_status`] methods. 
+
+The [`~peft.PeftModel.get_layer_status`] method gives you a detailed overview of each targeted layer's active, merged, and available adapters.
 
 ```python
 >>> from transformers import AutoModel
@@ -178,11 +180,11 @@ TunerModelStatus(
 )
 ```
 
-In the model state output, you should look out especially for entries that say `"irregular"`. This means that PEFT has detected an inconsistent state in the model. For instance, when it says `merged_adapters="irregular"`, it means that for at least one adapter, it was found that it was merged on some target modules but not on others that it targets. If you see this, the inference results will most likely be incorrect.
+In the model state output, you should look out for entries that say `"irregular"`. This means PEFT detected an inconsistent state in the model. For instance, if `merged_adapters="irregular"`, it means that for at least one adapter, it was merged on some target modules but not on others. The inference results will most likely be incorrect as a result.
 
-The best way to resolve this issue is to reload the whole model and the adapter checkpoint(s). Ensure that you don't perform any incorrect operations on the model, e.g. manually merging adapters on some modules but not others.
+The best way to resolve this issue is to reload the whole model and adapter checkpoint(s). Ensure that you don't perform any incorrect operations on the model, e.g. manually merging adapters on some modules but not others.
 
-Tip: You can easily convert the layer status into a pandas `DataFrame` for easier inspection:
+Convert the layer status into a pandas `DataFrame` for an easier visual inspection.
 
 ```python
 from dataclasses import asdict
--- a/docs/source/developer_guides/troubleshooting.md
+++ b/docs/source/developer_guides/troubleshooting.md
@@ -195,3 +195,48 @@ import pandas as pd
 
 df = pd.DataFrame(asdict(layer) for layer in model.get_layer_status())
 ```
+
+It is possible to get this information for non-PEFT models if they are using PEFT layers under the hood, but some information like the `base_model_type` or the `peft_types` cannot be determined in that case. As an example, you can call this on a [diffusers](https://huggingface.co/docs/diffusers/index) model like so:
+
+```python
+>>> import torch
+>>> from diffusers import StableDiffusionPipeline
+>>> from peft import get_model_status, get_layer_status
+
+>>> path = "runwayml/stable-diffusion-v1-5"
+>>> lora_id = "takuma104/lora-test-text-encoder-lora-target"
+>>> pipe = StableDiffusionPipeline.from_pretrained(path, torch_dtype=torch.float16)
+>>> pipe.load_lora_weights(lora_id, adapter_name="adapter-1")
+>>> pipe.load_lora_weights(lora_id, adapter_name="adapter-2")
+>>> get_layer_status(pipe.text_encoder)
+[TunerLayerStatus(name='text_model.encoder.layers.0.self_attn.k_proj',
+                  module_type='lora.Linear',
+                  enabled=True,
+                  active_adapters=['adapter-2'],
+                  merged_adapters=[],
+                  requires_grad={'adapter-1': False, 'adapter-2': True},
+                  available_adapters=['adapter-1', 'adapter-2']),
+ TunerLayerStatus(name='text_model.encoder.layers.0.self_attn.v_proj',
+                  module_type='lora.Linear',
+                  enabled=True,
+                  active_adapters=['adapter-2'],
+                  merged_adapters=[],
+                  requires_grad={'adapter-1': False, 'adapter-2': True},
+                  available_adapters=['adapter-1', 'adapter-2']),
+...]
+
+>>> get_model_status(pipe.unet)
+TunerModelStatus(
+    base_model_type='other',
+    adapter_model_type='None',
+    peft_types={},
+    trainable_params=797184,
+    total_params=861115332,
+    num_adapter_layers=128,
+    enabled=True,
+    active_adapters=['adapter-2'],
+    merged_adapters=[],
+    requires_grad={'adapter-1': False, 'adapter-2': True},
+    available_adapters=['adapter-1', 'adapter-2'],
+)
+```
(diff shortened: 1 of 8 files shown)


docs/source/developer_guides/troubleshooting.md
```diff
@@ -135,3 +135,58 @@ model.save_adapter("my_adapter", save_embedding_layers=True)
 For inference, load the base model first and resize it the same way you did before you trained the model. After you've resized the base model, you can load the PEFT checkpoint.
 
 For a complete example, please check out [this notebook](https://github.com/huggingface/peft/blob/main/examples/causal_language_modeling/peft_lora_clm_with_additional_tokens.ipynb).
+
+### Investigating the status of the PEFT model
```
stevhliu, line 139: I think this is clearer about what exactly you're investigating (layer/model status)

```suggestion
### Check layer and model status
```

docs/source/developer_guides/troubleshooting.md
```diff
@@ -135,3 +135,58 @@ model.save_adapter("my_adapter", save_embedding_layers=True)
...
+
+Sometimes, the PEFT model can end up in a bad state, especially when handling multiple adapters. There can be some confusion around what adapters exist, which one is active, which one is merged, etc. To help investigate this issue, you can call the [`~peft.PeftModel.get_layer_status`] and the [`~peft.PeftModel.get_model_status`] methods. The first one gives you a detailed overview about the adapters for each targeted layer. The latter one gives you a high-level overview about the model status.
```
stevhliu, line 141: It'd be nice to break these two methods up a little bit so users don't necessarily have to copy/paste both methods from the code snippet if they only wanted one of them. So I would do something like:

```md
intro about `get_layer_status`
code snippet for `get_layer_stuats`

intro about `get_model_status`
code snippet for `get_model_status`
```

```suggestion
Sometimes a PEFT model can end up in a bad state, especially when handling multiple adapters. There can be some confusion around what adapters exist, which one is active, which one is merged, etc. To help investigate this issue, call the [`~peft.PeftModel.get_layer_status`] and the [`~peft.PeftModel.get_model_status`] methods. 

The [`~peft.PeftModel.get_layer_status`] method gives you a detailed overview of each targeted layer's active, merged, and available adapters.
```

docs/source/developer_guides/troubleshooting.md
```diff
@@ -135,3 +135,58 @@ model.save_adapter("my_adapter", save_embedding_layers=True)
...
+    total_params=60855680,
+    num_adapter_layers=48,
+    enabled=True,
+    active_adapters=['default'],
+    merged_adapters=[],
+    available_adapters=['default'],
+)
+```
+
+In the model state output, you should look out especially for entries that say `"irregular"`. This means that PEFT has detected an inconsistent state in the model. For instance, when it says `merged_adapters="irregular"`, it means that for at least one adapter, it was found that it was merged on some target modules but not on others that it targets. If you see this, the inference results will most likely be incorrect.
```
stevhliu, line 181:
```suggestion
In the model state output, you should look out for entries that say `"irregular"`. This means PEFT detected an inconsistent state in the model. For instance, if `merged_adapters="irregular"`, it means that for at least one adapter, it was merged on some target modules but not on others. The inference results will most likely be incorrect as a result.
```

docs/source/developer_guides/troubleshooting.md
```diff
@@ -135,3 +135,58 @@ model.save_adapter("my_adapter", save_embedding_layers=True)
...
+
+The best way to resolve this issue is to reload the whole model and the adapter checkpoint(s). Ensure that you don't perform any incorrect operations on the model, e.g. manually merging adapters on some modules but not others.
```
stevhliu, line 183:
```suggestion
The best way to resolve this issue is to reload the whole model and adapter checkpoint(s). Ensure that you don't perform any incorrect operations on the model, e.g. manually merging adapters on some modules but not others.
```

docs/source/developer_guides/troubleshooting.md
```diff
@@ -135,3 +135,58 @@ model.save_adapter("my_adapter", save_embedding_layers=True)
...
+
+Tip: You can easily convert the layer status into a pandas `DataFrame` for easier inspection:
```
stevhliu, line 185: Super cool! 🤩 

```suggestion
Convert the layer status into a pandas `DataFrame` for an easier visual inspection.
```

stevhliu (approved):
Very nice pair of tools! Maybe consider putting this as the first subsection under "Bad results from a loaded PEFT model" since this seems like a good first troubleshooting step.

BenjaminBossan:
I added information on `requires_grad` to the outputs as well.

src/peft/peft_model.py
```diff
@@ -2227,3 +2316,255 @@ def forward(
...
+    name: str
+    module_type: str
+    enabled: bool
+    active_adapters: list[str]
+    merged_adapters: list[str]
+    requires_grad: dict[str, bool | Literal["irregular"]]
+    available_adapters: list[str]
+
+
+def get_layer_status(model: torch.nn.Module) -> list[TunerLayerStatus]:
```
younesbelkada, line 2332: I would put these methods in the main init, as people can also use them for diffusers or custom models

src/peft/peft_model.py
```diff
@@ -2227,3 +2316,255 @@ def forward(
...
+       The names of the merged adapters, if any, e.g. `["default"]`.
+    - requires_grad : dict[str, bool | Literal["irregular"]]
+       The requires_grad status of the parameters for each adapter module. Ideally, it should be either `True` or
+       `False`. If the requires_grad status is not consistent across all parameters, the value will be set to
+       `"irregular"`.
+    - `available_adapters` (`list[str]`):
+       The names of the available adapters, e.g. `["default"]`.
+
+    Args:
+        model ([`~PeftModel`]):
```
younesbelkada, line 2356:
```suggestion
        model ([Union[`~PeftModel`, `nn.Module`]]):
```

docs/source/developer_guides/troubleshooting.md
```diff
@@ -135,3 +135,63 @@ model.save_adapter("my_adapter", save_embedding_layers=True)
...
+The best way to resolve this issue is to reload the whole model and adapter checkpoint(s). Ensure that you don't perform any incorrect operations on the model, e.g. manually merging adapters on some modules but not others.
+
+Convert the layer status into a pandas `DataFrame` for an easier visual inspection.
+
+```python
+from dataclasses import asdict
+import pandas as pd
+
+df = pd.DataFrame(asdict(layer) for layer in model.get_layer_status())
+```
```
younesbelkada, line 197: Maybe mention also that you can use the public methods to examine diffusers and / or custom models

src/peft/peft_model.py
```diff
@@ -2227,3 +2316,255 @@ def forward(
...
+            enabled=not module.disable_adapters,
+            active_adapters=module.active_adapters,
+            merged_adapters=module.merged_adapters,
+            requires_grad=requires_grad,
+            available_adapters=sorted(module._get_available_adapters()),
+        )
+        layer_status.append(status)
+
+    if not layer_status:
+        raise ValueError("No adapter layers found in the model, please ensure that it's a PEFT model")
```
younesbelkada, line 2415:
```suggestion
        raise ValueError("No adapter layers found in the model, please ensure that it's a PEFT model or that you have PEFT adapters injected in the model")
```

src/peft/peft_model.py
```diff
@@ -2227,3 +2316,255 @@ def forward(
...
+        base_model = model.base_model
+        peft_types = {key: str(config.peft_type).partition(".")[-1] for key, config in base_model.peft_config.items()}
+        adapter_model_type = base_model.__class__.__name__
+    elif isinstance(model, PreTrainedModel):
+        base_model_type = model.__class__.__name__
+        trainable_params, total_params = PeftModel.get_nb_trainable_parameters(model)
+        base_model = model
+        peft_types = {}
+        adapter_model_type = "None"
+    else:
```
younesbelkada, line 2494: if can be easily extended to diffusers / custom models here no? we could just force-assign `base_model_type` to `"other"`

younesbelkada, line 2494:
```suggestion
    else:
            base_model_type = "other"
            trainable_params, total_params = PeftModel.get_nb_trainable_parameters(model)
            base_model = model
            peft_types = {}
            adapter_model_type = "None"
```

src/peft/peft_model.py
```diff
@@ -2227,3 +2316,255 @@ def forward(
...
+        raise TypeError(
+            f"get_model_status() expects a PeftModel or a PreTrainedModel instance; got {model.__class__.__name__} "
+            "instead."
+        )
```
younesbelkada, line 2498:
```suggestion
```

younesbelkada (commented):
Thanks a lot for this great work ! I think we can easily extend it to all pytorch modules, I left few comments in that regard, what do you think?

src/peft/peft_model.py
BenjaminBossan: I was first hesitant to do that, as they were very tied to PEFT models, but with your suggestion, these functions are much more general now, so I agree and made that change.

src/peft/peft_model.py
BenjaminBossan: Right, good idea.

BenjaminBossan: Done

docs/source/developer_guides/troubleshooting.md
BenjaminBossan: I added a diffusers example below.

BenjaminBossan (commented):
Great feedback Younes, I implemented your changes to allow the functions to be called on non-PEFT models without raising an error. Please review again.

younesbelkada (approved):
Great feature ! Thanks so much @BenjaminBossan !

Issue #1655, opened 2024-04-15
Learning the combination weights of pre-trained LoRA Modules


mahdibeit:
### Feature request

PEFT can combine pre-trained LoRA modules by averaging them or providing custom weights for weighted averaging. This [paper ](https://arxiv.org/abs/2402.15414) showed that learning these weights is better than naive averaging in few-shot adaption settings.
<img width="737" alt="WLoRA" src="https://github.com/huggingface/peft/assets/96784135/667c969a-fbc7-42e4-8aa7-52582a2851b4">

### Motivation

Learning the combination weights allows users to utilize already available pre-trained LoRA modules in the Hugging Face Models. Also, it is very parameter efficient since we are only learning the combination weights. More importantly, it can surpass learning a LoRA from scratch in settings where the number of training samples is limited.

### Your contribution

I can submit a PR. Then, PEFT can combine any pre-trained LoRA using the following format:

`wlora_config = WLoraConfig(skilled_loras = [PATH_TO_UPSTREAM_1, PATH_TO_UPSTREAM_2, ])`

`model = get_peft_model(llama2, wlora_config )`

BenjaminBossan:
Hi, thanks you for proposing to add this method.

I only skimmed the paper but IIUC, we assume that the user has a couple of already trained LoRA adapters and now wants to combine them for a new task. The idea is that by learning the weights used for the weighted average (the `weights` argument for `add_weighted_adapter`) can lead to better results than naive uniform weights. (Note that we offer many combination types, not just averaging, maybe that's worth looking into for the paper.)

To learn these weights, I assume we have to load all the LoRA adapters at training time, freeze their weights, then add an extra scaling factor to [this line](https://github.com/huggingface/peft/blob/56773b9a92b141111d65fe3548d0c30233358868/src/peft/tuners/lora/layer.py#L516), is that right?

I haven't thought through the overall design of this, but I think it should be possible to add this to the existing LoRA code without too many additions. Feel free to open a draft PR where we can discuss the design.

mahdibeit:
Hi @BenjaminBossan , thanks for taking the time to read the paper.

Thank you for your great suggestion. We will evaluate other combination methods in the paper.

To answer your question, yes, you are absolutely right. We can just use the existing LoRA code with a Boolean like `learn_combination_weights` to configure the training process. Overall, I need to do the following:

1. Instantiate a trainable tensor named `combination_weights`  that learns the scaling factor for each pre-trained, frozen LoRA. 
[...]

PR #1666, opened 2024-04-22
supports learning the combination weights of pre-trained LoRA modules

mahdibeit:
Based on #1655

Adds a use_wlora config to `LoraLayer` that allows learning the combination weights (i.e. `wlora_weights) of pre-trained LoRAs.

### DIFF
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -126,6 +126,48 @@ Assuming the original model had 5 layers `[0, 1, 2 ,3, 4]`, this would create a 
 [Fewshot-Metamath-OrcaVicuna-Mistral-10B](https://huggingface.co/abacusai/Fewshot-Metamath-OrcaVicuna-Mistral-10B) is an example of a model trained using this method on Mistral-7B expanded to 10B. The
 [adapter_config.json](https://huggingface.co/abacusai/Fewshot-Metamath-OrcaVicuna-Mistral-10B/blob/main/adapter_config.json) shows a sample LoRA adapter config applying this method for fine-tuning.
 
+### Learning the scaling weights of pre-trained LoRAs
+
+This technique allows learning the scaling weights of multiple pre-trained LoRAs. This method adds a learnable weight to each of the pre-trained LoRA modules to learn the weighted sum of those modules. This allows us to learn the best combination of pre-trained LoRA for few-shot adaptation. For more information, see https://arxiv.org/abs/2402.15414.
+
+To use this feature you would create a config for each LoRA with the `wlora=True`: 
+```python
+from transformers import AutoModelForCausalLM, AutoTokenizer
+from peft import PeftConfig
+
+# Load base model
+base_model = "facebook/opt-350m"
+tokenizer = AutoTokenizer.from_pretrained(base_model)
+model = AutoModelForCausalLM.from_pretrained(base_model)
+
+# Add the first LoRA with learnable weight to the base model
+lora_1 = "varun-v-rao/opt-350m-lora-1.57M-squad-model3"
+lora_1_config = PeftConfig.from_pretrained(lora_1)
+lora_1_config.use_wlora = True
+model.add_adapter(adapter_config=lora_1_config, adapter_name="lora_1")
+
+# Add the second LoRA
+lora_2 = "varun-v-rao/opt-350m-lora-1.57M-squad-model3"
+lora_2_config = PeftConfig.from_pretrained(lora_2)
+lora_2_config.use_wlora = True
+model.add_adapter(adapter_config=lora_2_config, adapter_name="lora_2")
+
+# Activate LoRA modules as trainable
+model.set_adapter(["lora_1", "lora_2"])
+
+# Freeze lora_A and lora_B
+for name, param in model.named_parameters():
+    if "lora_A" in name or "lora_B" in name:
+        param.requires_grad = False
+```
+
+#### Caveats
+
+- WLoRA only supports linear layers at the moment.
+- WLoRA works better than training a LoRA from scratch in few-shot adaptation settings. However, when the training data increases, training a LoRA from scratch outperforms WLoRA since it has more trainable parameters.
+
+
+
 ## Merge adapters
 
 While LoRA is significantly smaller and faster to train, you may encounter latency issues during inference due to separately loading the base model and the LoRA adapter. To eliminate latency, use the [`~LoraModel.merge_and_unload`] function to merge the adapter weights with the base model. This allows you to use the newly merged model as a standalone model. The [`~LoraModel.merge_and_unload`] function doesn't keep the adapter weights in memory.
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -253,18 +253,23 @@ class LoraConfig(PeftConfig):
         default=None,
         metadata={
             "help": (
-                "This enables using LoRA to effectively expand a transformer model to a larger size by repeating some layers. "
-                "The transformation handles models (currently Llama, Bert or Falcon compatible architectures) with "
-                "a module list in the model which it modifies to expand the number of modules. "
-                "Base weights are shared so the memory usage is close to the original model. The intended use is these base weights "
-                "remain fixed during finetuning but each layer has a separate LoRA adapter so the layers can be specialed via "
-                "the adapter layers fit during fine tuning."
-                "The format is a list of [start, end) pairs which specify the layer ranges to stack. For example:\n"
-                "   Original model has 5 layers labelled by their position in the model: `[0, 1, 2, 3, 4]`\n"
-                "   layer_replication: `[[0, 4], [2, 5]]`\n"
-                "   Final model will have this arrangement of original layers: `[0, 1, 2, 3, 2, 3, 4]`\n"
-                "This format is based on what is used for pass-through merges in mergekit. It makes it simple to select sequential "
-                "ranges of a model and stack them while reusing layers at either end of each sequence."
+                "Enable 'Weight-Decomposed Low-Rank Adaptation' (DoRA). This technique decomposes the updates of the "
+                "weights into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the "
+                "magnitude is handled by a separate learnable parameter. This can improve the performance of LoRA, "
+                "especially at low ranks. Right now, DoRA only supports linear and Conv2D layers. DoRA introduces a bigger"
+                "overhead than pure LoRA, so it is recommended to merge weights for inference. For more information, "
+                "see  https://arxiv.org/abs/2402.09353."
+            )
+        },
+    )
+    use_wlora: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Enable learning the combination weights of pre-trained LoRAs. This technique adds a learnable weight "
+                "to each of the pre-trained LoRA module in each layer to learn the weighted sum of those moduels. "
+                "This can allow to learn the best combination of pre-trained LoRA for few-shot adaptation. Right now,"
+                "WLoRA only supports linear layers. For more information, see https://arxiv.org/abs/2402.15414."
             )
         },
     )
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -112,6 +112,11 @@ class LoraConfig(PeftConfig):
             Build a new stack of layers by stacking the original model layers according to the ranges specified. This
             allows expanding (or shrinking) the model without duplicating the base model weights. The new layers will
             all have separate LoRA adapters attached to them.
+        use_wlora (`bool`):
+            Adds a learnable scaling weight to a pre-trained LoRA. This technique can add a learnable weight to each
+            of the pre-trained LoRA module in each layer to learn the weighted sum of those modules. This can allow
+            to learn the best weighted combination of pre-trained LoRA for few-shot adaptation. For more information,
+            see https://arxiv.org/abs/2402.15414.
     """
 
     r: int = field(default=8, metadata={"help": "Lora attention dimension"})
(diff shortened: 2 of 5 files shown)


src/peft/tuners/lora/layer.py
```diff
@@ -125,9 +135,25 @@ def update_layer(
...
+            # initialize wlora_weights
+            if not self.wlora_weights:
+                self.wlora_weights = nn.ParameterDict()
+            self.wlora_weights[adapter_name] = nn.Parameter(torch.tensor([1.0]))
+            # add `wlora_weights`` to the list of learnable parameters
+            self.adapter_layer_names = self.adapter_layer_names[:] + ("wlora_weights",)
+            self.use_wlora[adapter_name] = True
+            # remove `lora_A` and `lora_B` from the list of trainable parameters
+            self.adapter_layer_names = tuple(
+                layer for layer in self.adapter_layer_names if layer != "lora_A" and layer != "lora_B"
```
mahdibeit, line 152: I am not sure if this is the best way to freeze lora_A and loraB

BenjaminBossan:
Thanks for the PR. For me to be able to review it, could you provide an example of how it should be used?

mahdibeit:
@BenjaminBossan Yes, for sure. 

The code implements the learned composition in  *Does Combining Parameter-efficient Modules Improve Few-shot Transfer Accuracy?* (Asadi et al., 2024). More specifically, it learns the $v$ for the weighted sum of LoRA modules as follows.

```math
\hat{\mathbf{W}} = \mathbf{W}_{base} + \sum_{n=1}^{N} \hat{v}_n \left( \frac{\alpha_n}{r_n} \mathbf{A}_n \mathbf{B}_n\right), 
```
```math
\sum_{n=1}^{N}\hat{v}_n=1,
```
where $\hat{v}$ is the softmax operation applied on the weighting vector $v$, i.e.,
```math
\hat{v}_n=e^{v_n}/\left(\sum_{j=1}^{N}e^{v_j}\right)
```
We named the parameter $v$ as `wlora_weights` in the model parameters. 



## Usage example
The following script is an example of how to load two pre-trained LoRA modules and learn the combination weights for LLMs. 

First, we load the base model
```python
from peft import PeftConfig

# Load base model
base_model = "facebook/opt-350m"
tokenizer = AutoTokenizer.from_pretrained(base_model)
model = AutoModelForCausalLM.from_pretrained(base_model)
```

Then, we add the two LoRAs and make their weight trainable.
```python
# Add the first LoRA with learnable weight to the base model
lora_1 = "varun-v-rao/opt-350m-lora-1.57M-squad-model3"
lora_1_config = PeftConfig.from_pretrained(lora_1)
lora_1_config.use_wlora =True
model.add_adapter(adapter_config=lora_1_config, adapter_name='lora_1')

# Add the second LoRA
lora_2 = "varun-v-rao/opt-350m-lora-1.57M-squad-model3"
lora_2_config = PeftConfig.from_pretrained(lora_2)
lora_2_config.use_wlora =True
model.add_adapter(adapter_config=lora_2_config, adapter_name='lora_2')

# Activate LoRA modules as trainable
model.set_adapter(['lora_1', 'lora_2'])
```

Modules are successfully loaded and you can treat the model as any HuggingFace or `torch.nn.Module` and use any training method. Following is an example of using the HuggingFace Trainer.
```python
# Train the wights of LoRA modules
from transformers import TrainingArguments, Trainer

    output_dir="wlora-model",
    evaluation_strategy="epoch",
    learning_rate=1e-4,
    weight_decay=0.01,
    push_to_hub=False,
)

trainer = Trainer(
    model=model,
    args=training_args,
    train_dataset=lm_dataset["train"],
    eval_dataset=lm_dataset["test"],
    data_collator=data_collator,
)

trainer.train()
```

src/peft/tuners/lora/config.py
```diff
@@ -253,18 +253,23 @@ class LoraConfig(PeftConfig):
...
+                "overhead than pure LoRA, so it is recommended to merge weights for inference. For more information, "
+                "see  https://arxiv.org/abs/2402.09353."
+            )
+        },
+    )
+    use_wlora: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Enable learning the combination weights of pre-trained LoRAs. This technique adds a learnable weight "
```
BenjaminBossan, line 269: The docstring also needs to be updated.

src/peft/tuners/lora/config.py
```diff
@@ -253,18 +253,23 @@ class LoraConfig(PeftConfig):
...
+                "Enable 'Weight-Decomposed Low-Rank Adaptation' (DoRA). This technique decomposes the updates of the "
+                "weights into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the "
+                "magnitude is handled by a separate learnable parameter. This can improve the performance of LoRA, "
+                "especially at low ranks. Right now, DoRA only supports linear and Conv2D layers. DoRA introduces a bigger"
+                "overhead than pure LoRA, so it is recommended to merge weights for inference. For more information, "
+                "see  https://arxiv.org/abs/2402.09353."
+            )
+        },
+    )
+    use_wlora: bool = field(
```
BenjaminBossan, line 265: About the name: Was `wlora` used in the paper?

src/peft/tuners/lora/config.py
```diff
@@ -253,18 +253,23 @@ class LoraConfig(PeftConfig):
...
+        default=False,
+        metadata={
+            "help": (
+                "Enable learning the combination weights of pre-trained LoRAs. This technique adds a learnable weight "
+                "to each of the pre-trained LoRA module in each layer to learn the weighted sum of those moduels. "
```
BenjaminBossan, line 270:
```suggestion
                "to each of the pre-trained LoRA module in each layer to learn the weighted sum of those modules. "
```

BenjaminBossan (changes requested):
Thanks for working on this PR and for providing an example. I also see now that you're one of the paper authors :)

I left a couple of comments on this PR. On top of that, we should probably also add a section to the docs ([here](https://huggingface.co/docs/peft/developer_guides/lora#initialization)) because it is not quite trivial to figure out for a user how to use this.

Moreover, I tried to come up with a test for this method. When I tried something based on the example you provided, I ran into an error though. Could you please check?

```python
import torch

from peft import PeftConfig, LoraConfig, get_peft_model, PeftModel

torch.manual_seed(0)
base_model_id = "facebook/opt-125m"
tokenizer = AutoTokenizer.from_pretrained(base_model_id)
model = AutoModelForCausalLM.from_pretrained(base_model_id)

config = LoraConfig(init_lora_weights=False, use_wlora=True)
model.base_model.set_adapter(['lora_1', 'lora_2'])

optimizer = torch.optim.SGD(model.parameters(), lr=0.1)
optimizer.zero_grad()
output = model(torch.arange(10).reshape(-1, 1))
loss = output.logits.sum()
loss.backward()
# this causes: RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

Maybe this is related to the comment about how to set `requires_grad`, not sure.

mahdibeit:
> Thanks for working on this PR and for providing an example. I also see now that you're one of the paper authors :)
> 
> I left a couple of comments on this PR. On top of that, we should probably also add a section to the docs ([here](https://huggingface.co/docs/peft/developer_guides/lora#initialization)) because it is not quite trivial to figure out for a user how to use this.
> 
> Moreover, I tried to come up with a test for this method. When I tried something based on the example you provided, I ran into an error though. Could you please check?
> 
> ```python
> import torch
> 
> from transformers import AutoModelForCausalLM, AutoTokenizer
> from peft import PeftConfig, LoraConfig, get_peft_model, PeftModel
> 
> torch.manual_seed(0)
> base_model_id = "facebook/opt-125m"
> tokenizer = AutoTokenizer.from_pretrained(base_model_id)
> model = AutoModelForCausalLM.from_pretrained(base_model_id)
> 
> config = LoraConfig(init_lora_weights=False, use_wlora=True)
> model = get_peft_model(model, config)
> model.add_adapter("other", config)
> model.base_model.set_adapter(['lora_1', 'lora_2'])
> 
> optimizer = torch.optim.SGD(model.parameters(), lr=0.1)
> optimizer.zero_grad()
> output = model(torch.arange(10).reshape(-1, 1))
> loss = output.logits.sum()
> loss.backward()
> # this causes: RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
> ```
> 
> Maybe this is related to the comment about how to set `requires_grad`, not sure.

Thanks for taking the time to read the PR. Yes, I am one of the authors. I was hoping to create an easy method for the community to combine pre-trained LoRAs :)

This is a great test script. Yes, there was an issue regarding the require grad and I changed the method as you mentioned. Also, `model.base_model.set_adapter(['lora_1', 'lora_2'])`

should contain the names of the layers so it should be :
`model.base_model.set_adapter(['default', 'other'])`

src/peft/tuners/lora/config.py
```diff
@@ -253,18 +253,23 @@ class LoraConfig(PeftConfig):
...
-                "   layer_replication: `[[0, 4], [2, 5]]`\n"
-                "   Final model will have this arrangement of original layers: `[0, 1, 2, 3, 2, 3, 4]`\n"
-                "This format is based on what is used for pass-through merges in mergekit. It makes it simple to select sequential "
-                "ranges of a model and stack them while reusing layers at either end of each sequence."
+                "Enable 'Weight-Decomposed Low-Rank Adaptation' (DoRA). This technique decomposes the updates of the "
+                "weights into two parts, magnitude and direction. Direction is handled by normal LoRA, whereas the "
+                "magnitude is handled by a separate learnable parameter. This can improve the performance of LoRA, "
+                "especially at low ranks. Right now, DoRA only supports linear and Conv2D layers. DoRA introduces a bigger"
+                "overhead than pure LoRA, so it is recommended to merge weights for inference. For more information, "
+                "see  https://arxiv.org/abs/2402.09353."
```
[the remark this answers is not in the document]
mahdibeit: Yes, I fixed it

src/peft/tuners/lora/config.py
mahdibeit: Updated

src/peft/tuners/lora/config.py
mahdibeit: No, we did not use any names. I am not sure what is the best way to describe this method in the config. We can use WeightedLoRATrainer, AdaptiveScaling, or LearnableScaling. Do you have any suggestions?

src/peft/tuners/lora/config.py
mahdibeit: Fixed

src/peft/tuners/lora/layer.py
```diff
@@ -512,11 +560,15 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
...
                 x = x.to(lora_A.weight.dtype)
 
-                if not self.use_dora[active_adapter]:
-                    result = result + lora_B(lora_A(dropout(x))) * scaling
-                else:
+                if self.use_dora[active_adapter]:
                     x = dropout(x)
                     result = result + self._apply_dora(x, lora_A, lora_B, scaling, active_adapter)
+                elif self.use_wlora[active_adapter]:
+                    wlora_scale = self._cal_wlora_scale(active_adapter)
```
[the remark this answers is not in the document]
mahdibeit: I think I understand your suggestion. So there are two ways we can combine dlora with wlora:

### wlora and dlora working separately

So in this case (this is the case that we already support in the PR), we can have let's say three Lora modules, e.g., lora_1, lora_2, and lora_3. We can set lora_1 as dlora and lora_2 and lora_3 as wlora. So the forward formulation looks like this:
```python
result = result
        + self._apply_dora(x, lora_1_A, lora_1_B, scaling_1, lora_1)
        + lora_2_B(lora_2_A(dropout_2(x))) * scaling_2 * wlora_scale_2
        + lora_3_B(lora_3_A(dropout_3(x))) * scaling_3 * wlora_scale_3
```
where `wlora_scale_2 + wlora_scale_3 = 1`

### wlora and dlora working togther
In this case (that we do not currently support and I it helps dlora at all) we use two LoRAs (e.g. lora_1 and lora_2) inside dlora and we need to change _apply_dora function to accept wlora parameters like this:
```python
result = result
        + self._apply_dora_and_wlora(x, lora_1_A, lora_1_B, scaling_1, wlora_scale_1, lora_2_A, lora_2_B, scaling_2, wlora_scale_2,)
```

Overall, I think with the current implementation, wlora works with other modules and we don't have to worry about conflicts since it can be trained and combined with other PEFT methods.

src/peft/tuners/lora/config.py
BenjaminBossan: I don't have any specific suggestion other than probably not `wlora` if it's not used in the paper. E.g. imagine that a new method comes out with the same name (which, given the dozens of existing LoRA variants, is not unlikely), this could cause a lot of confusion. Your latter 2 suggestions sound good though, choose whatever fits the paper better.

src/peft/tuners/lora/layer.py
```diff
@@ -239,6 +261,22 @@ def _apply_dora(self, x, lora_A, lora_B, scaling, active_adapter):
 
         return result_dora
 
+    def _cal_wlora_scale(self, active_adapter: str) -> torch.Tensor:
```
BenjaminBossan, line 264: Let's rename this to `_calculate_wlora_scale` or `_get_wlora_scale` (and replace "wlora" with whatever other name you choose).

src/peft/tuners/lora/layer.py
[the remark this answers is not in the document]
BenjaminBossan: Not quite sure if I understand your suggestion completely. Probably you mean if we "unroll" the for loop for the different adapters inside of `forward`, this is what it would look like?

Would it not work if we just apply the wlora scale to the existing `scaling` argument and the rest takes care of itself?

docs/source/developer_guides/lora.md
```diff
@@ -126,6 +126,48 @@ Assuming the original model had 5 layers `[0, 1, 2 ,3, 4]`, this would create a
...
+from transformers import AutoModelForCausalLM, AutoTokenizer
+from peft import PeftConfig
+
+# Load base model
+base_model = "facebook/opt-350m"
+tokenizer = AutoTokenizer.from_pretrained(base_model)
+model = AutoModelForCausalLM.from_pretrained(base_model)
+
+# Add the first LoRA with learnable weight to the base model
+lora_1 = "varun-v-rao/opt-350m-lora-1.57M-squad-model3"
```
BenjaminBossan, line 144: Ideally, we can move these examples over to our testing repo https://huggingface.co/peft-internal-testing. I can do that. Are these weights actually trained on something and create interesting outputs, or are those just dummy weights?

docs/source/developer_guides/lora.md
```diff
@@ -126,6 +126,48 @@ Assuming the original model had 5 layers `[0, 1, 2 ,3, 4]`, this would create a
...
+lora_2_config.use_wlora = True
+model.add_adapter(adapter_config=lora_2_config, adapter_name="lora_2")
+
+# Activate LoRA modules as trainable
+model.set_adapter(["lora_1", "lora_2"])
+
+# Freeze lora_A and lora_B
+for name, param in model.named_parameters():
+    if "lora_A" in name or "lora_B" in name:
+        param.requires_grad = False
```
BenjaminBossan, line 161: It would be really great if we could automate this step. Check my other comment below.

src/peft/tuners/lora/layer.py
```diff
@@ -125,9 +135,21 @@ def update_layer(
...
             self.use_dora[adapter_name] = False
 
+        if use_wlora:
+            # initialize wlora_weights
+            if not self.wlora_weights:
+                self.wlora_weights = nn.ParameterDict()
+            self.wlora_weights[adapter_name] = nn.Parameter(torch.tensor([1.0]))
+            # add `wlora_weights`` to the list of learnable parameters
+            self.adapter_layer_names = self.adapter_layer_names[:] + ("wlora_weights",)
+            self.use_wlora[adapter_name] = True
```
BenjaminBossan, line 149: I wonder if we could not automatically deactivate the gradients for the normal LoRA weights if we use wlora. So basically: `self.lora_A[adapter_name].requires_grad_(False)` etc. WDYT?

src/peft/tuners/lora/layer.py
```diff
@@ -125,9 +135,21 @@ def update_layer(
...
             self.use_dora[adapter_name] = True
+
         else:
             self.use_dora[adapter_name] = False
 
+        if use_wlora:
+            # initialize wlora_weights
+            if not self.wlora_weights:
+                self.wlora_weights = nn.ParameterDict()
+            self.wlora_weights[adapter_name] = nn.Parameter(torch.tensor([1.0]))
```
BenjaminBossan, line 146: When this is renamed, please ensure that the parameter name still contains the substring (`lora_`), since this makes it much easier for users to quickly identify all LoRA-related parameters.

BenjaminBossan (changes requested):
Great, thanks a lot for the updates.

> So right now using this update we can write the following scripts as test and example usage. I added the example usage to the docs.

Next step is now to add a unit tests (or a few) based on the examples. We should ensure in these tests that training works (can be a very simple training loop). At the end, we should ideally see that only the wlora weights were updated and the LoRA weights stay the same. For this test, we should also generate LoRA adapters with `init_lora_weights=False`.

LMK if you need help with adding these unit tests.

mahdibeit:
Hi @BenjaminBossan, sorry I was busy in the last three weeks. I will apply your comments and push it this week.

PR #1668, opened 2024-04-22
Fix warning messages about `config.json` when the base `model_id` is local.

elementary-particle:
It should be possible for the user to specify a local directory as the base model in the library.
However, currently the library only checks for remote presence of `config.json`, and fails to check the actual `config.json` when using a local repo.
This PR adds check for a local `model_id` and fixes the behavior.


BenjaminBossan:
Thanks for the PR, what you describe sounds reasonable. Do you have a small example where this change would apply? Ideally, we can use that to create a unit test.

elementary-particle:
Sure, a simple example is to create a LoRA adapter for a local base model and saving it.
for example, create a `PeftModel` for a local snapshot of `mistralai/Mistral-7B-v0.1`, and saving the adapter issues a warning saying
```python
warnings.warn(
    f"Could not find a config file in {model_id} - will assume that the vocabulary was not modified."
)
```
And the configuration is not checked therefore.

elementary-particle:
```python
from peft import LoraConfig, PeftModel
from peft import prepare_model_for_kbit_training, get_peft_model

local_dir = 'path/to/model'

base_model = AutoModelForCausalLM.from_pretrained(local_dir)
peft_config = LoraConfig(
    lora_alpha=16,
    lora_dropout=0.1,
    r=64,
    bias="none",
    task_type="CAUSAL_LM",
    target_modules=["q_proj", "k_proj", "v_proj", "o_proj","gate_proj"]
)
peft_model = get_peft_model(base_model, peft_config)
peft_model.save_pretrained("test")
```

elementary-particle:
Is this sufficient? @BenjaminBossan

BenjaminBossan:
Nice, thanks for providing the example. I used it to create a test:

```python
class TestLocalModel:
    def test_local_model_saving_no_warning(self, recwarn, tmp_path):
        local_dir = tmp_path / model_id
        model.save_pretrained(local_dir)
        del model

        base_model = AutoModelForCausalLM.from_pretrained(local_dir)
        peft_config = LoraConfig(
            lora_alpha=16,
            lora_dropout=0.1,
            r=64,
            bias="none",
            task_type="CAUSAL_LM",
            target_modules=["q_proj", "k_proj", "v_proj", "o_proj","gate_proj"]
        )
        peft_model = get_peft_model(base_model, peft_config)
        peft_model.save_pretrained(local_dir)

        for warning in recwarn.list:
            assert "Could not find a config file" not in warning.message.args[0]
```

We could for instance put it into `tests/test_hub_features.py`, WDYT? Running it locally on main, it currently fails, but on your branch, it should pass.

elementary-particle:
Certainly, I followed the syntax in `testing_common.py` and created a test unit for the issue.
Maybe some further checks? @BenjaminBossan

BenjaminBossan:
> Certainly, I followed the syntax in `testing_common.py` and created a test unit for the issue.

The way you added the test, it's not executed. You would have to add corresponding methods that call this method in `test_decoder_models.py`, `test_encoder_decoder_models.py`, etc. But this is overkill, we don't really need to check this with all kind of different model architectures. Instead, as I suggested earlier, just add this test to `tests/test_hub_features.py` and it should be good. Let's also add a comment that explains why we need this test.

elementary-particle:
The test case is fixed as advised and comments are added to explain the issue.
Please review the changes, thanks.

tests/test_hub_features.py
```diff
@@ -35,3 +35,25 @@ def test_subfolder(self):
...
+        # If `save_pretrained` could not find the file, it will issue a warning. 
+        model_id = "facebook/opt-125m"
+        model = AutoModelForCausalLM.from_pretrained(model_id)
+        local_dir = tmp_path / model_id
+        model.save_pretrained(local_dir)
+        del model
+
+        base_model = AutoModelForCausalLM.from_pretrained(local_dir)
+        peft_config = LoraConfig()
+        peft_model = get_peft_model(base_model, peft_config)
```
BenjaminBossan, line 55: `LoraConfig` and `get_peft_model` need to be imported.

BenjaminBossan (changes requested):
Thanks, the test will now be run as part of the test suite. However, you forgot some imports for the test.

tests/test_hub_features.py
elementary-particle: Resolved.

BenjaminBossan:
@elementary-particle Thanks for the update. Could you please run `make style`?

elementary-particle:
The merge conflict is resolved.

BenjaminBossan (approved):
Thanks a lot for fixing this warning, LGTM.

PR #1681, opened 2024-04-25
FIX Use trl version of tiny random llama

BenjaminBossan:
Using the version from HuggingFaceM4 broke our tests because it was updated. Although the update is reverted, we still better switch to this version, which is explicitly for testing and should be stable.


younesbelkada (approved):
Merci !

PR #1682, opened 2024-04-25
FIX / Workflow: Always notify on slack for docker image workflows

younesbelkada:
# What does this PR do?

As per title, we were not posting on slack when the docker image build fails, this PR enables it

cc @pacman100 @BenjaminBossan


BenjaminBossan:
@younesbelkada Could you please merge main so that the tests can pass?

BenjaminBossan (approved):
LGTM, thanks.

PR #1683, opened 2024-04-25
FIX: Don't eagerly import bnb for LoftQ

### LINKED ISSUE #1665: `bitsandbytes` is imported eagerly in `peft/utils/loftq_utils.py`
https://github.com/huggingface/peft/pull/1230 makes the import of `bitsandbytes` lazy, however, the fix seems to be imcomplete.
https://github.com/huggingface/peft/blob/5a4b9cade64bac8afdff5006ee9dd815c90b5469/src/peft/utils/loftq_utils.py#L34-L35
Perhaps there can be a test case ensuring that the CUDA context is not initialized after importing `peft`, e.g.:
```python
import torch
import peft
assert not torch.cuda.is_initialized()
```
@BenjaminBossan

BenjaminBossan:
Resolves #1665.

We accidentally added code in `loftq_utils.py` that eagerly imports bnb, which we want to avoid to prevent CUDA from being initialized too early.

### DIFF
--- a/src/peft/utils/loftq_utils.py
+++ b/src/peft/utils/loftq_utils.py
@@ -31,10 +31,6 @@ from transformers.utils.hub import get_checkpoint_shard_files
 from peft.import_utils import is_bnb_4bit_available, is_bnb_available
 
 
-if is_bnb_available():
-    import bitsandbytes as bnb
-
-
 class NFQuantizer:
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Thanks !

BenjaminBossan:
As I explained [in this comment](https://github.com/huggingface/peft/issues/1665#issuecomment-2077504313), it is unfortunately not easily possible to write a test for this.

PR #1684, opened 2024-04-26
FIX: upgrade autoawq to latest version

younesbelkada:
cc @BenjaminBossan

### DIFF
--- a/docker/peft-gpu/Dockerfile
+++ b/docker/peft-gpu/Dockerfile
@@ -42,9 +42,9 @@ RUN source activate peft && \
 
 # Add autoawq for quantization testing
 RUN source activate peft && \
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Would it be possible to use `pip install autoawq` and `pip install autoawq-kernels` to always stay on the latest release?

younesbelkada:
I think this is not possible since the wheels pushed on pypi are specific to some python and cuda version that are diffferent from the one in our docker image

BenjaminBossan (approved):
Thanks for fixing, hopefully we have a :heavy_check_mark: in our CI tomorrow.

> I think this is not possible since the wheels pushed on pypi are specific to some python and cuda version that are diffferent from the one in our docker image

That's too bad, this means we have to constantly keep watch to update the version in the Dockerfile :-/

BenjaminBossan:
Hmm, [still getting](https://github.com/huggingface/peft/actions/runs/8872334776/job/24356524248)
>FAILED tests/test_gpu_examples.py::PeftAwqGPUTests::test_causal_lm_training_awq - NameError: name 'awq_ext' is not defined

PR #1690, opened 2024-04-29
TST: Skiping AWQ tests for now ..

younesbelkada:
As per title and discussed offline, let's skip the awq tests until: https://github.com/casper-hansen/AutoAWQ/issues/466 gets resolved

cc @BenjaminBossan

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -1930,6 +1930,7 @@ class PeftAqlmGPUTests(unittest.TestCase):
             assert trainer.state.log_history[-1]["train_loss"] is not None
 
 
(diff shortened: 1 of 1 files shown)


BenjaminBossan (approved):
Thanks. Let's maybe add a `# TODO` comment so that we don't forget about this?

tests/test_gpu_examples.py
```diff
@@ -1932,6 +1932,7 @@ def test_causal_lm_training_aqlm(self):
 
 @require_torch_gpu
```
younesbelkada, line 1933:
```suggestion
# TODO: unskip the tests once https://github.com/casper-hansen/AutoAWQ/issues/466 is fixed
@require_torch_gpu
```

younesbelkada:
Sure!

PR #1691, opened 2024-04-29
FIX: Fix eetq + peft

younesbelkada:
Needs to be merged with: https://github.com/NetEase-FuXi/EETQ/pull/16 

https://github.com/NetEase-FuXi/EETQ/pull/15 being merged, we can properly use the correct kernels for backward, 

https://github.com/NetEase-FuXi/EETQ/pull/16  fixes the failing tests on our CI 

cc @BenjaminBossan

### DIFF
--- a/src/peft/tuners/lora/eetq.py
+++ b/src/peft/tuners/lora/eetq.py
@@ -96,6 +96,7 @@ def dispatch_eetq(
 
     if is_eetq_available() and isinstance(target_base_layer, EetqLinear):
         new_module = EetqLoraLinear(target, adapter_name, **kwargs)
+        new_module.train()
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Nice, thanks a lot, Younes, for investigating and addressing this issue. It's a bit odd that we have to call `train()` here, as the convention with `Module`s is that it's the default. Let's wait for the other PR to be merged first.

younesbelkada:
Oh wait if it's the default this is not needed then, let me confirm that

younesbelkada:
```
>>> EetqLinear(1, 1).training
True
```
Closing the PR, let's just wait the above PR is merged

PR #1694, opened 2024-04-30
fix the fsdp peft autowrap policy

pacman100:
### What does this PR do?
1. In PR https://github.com/huggingface/accelerate/pull/2662, the static method was removed as other classes also required it and hence was made a function to avoid code duplication in each class. This PR updates the usage of that function.


BenjaminBossan (commented):
This would only work with the latest accelerate version, right? So this would require all PEFT users to upgrade accelerate. Can we instead make this backwards compatible?

pacman100:
> This would only work with the latest accelerate version, right? So this would require all PEFT users to upgrade accelerate. Can we instead make this backwards compatible?

Done

BenjaminBossan:
@pacman100 After merging #1697, the doc built should be fixed, so merging main should fix the CI.

BenjaminBossan (approved):
Thanks for the fix, LGTM.

PR #1697, opened 2024-04-30
FIX Use different doc builder docker image

BenjaminBossan:
Same as in:

https://github.com/huggingface/transformers/blob/c712d05aa8fc8ba3ebe465079bd377d2dc9c2e07/.github/workflows/build_documentation.yml#L19


BenjaminBossan:
Doc built is currently failing, as e.g. here: https://github.com/huggingface/peft/actions/runs/8893174861/job/24418810402

PR #1698, opened 2024-04-30
Set experimental dynamo config for compile tests

BenjaminBossan:
Right now, a lot of tests fail when applying torch.compile to PEFT models. One of the main reasons is that attribute checks (e.g. `if self.foo`) on `nn.Module`s are not correctly considered.

This PR sets an experimental flag that should fix this. However, this is not public PyTorch API (yet) and incurs a performance penalty. Still, it's interesting to see how this affects our tests.

More context:
https://github.com/pytorch/pytorch/issues/124717#issuecomment-2083235776

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -10,6 +10,11 @@ if os.environ.get("PEFT_DEBUG_WITH_TORCH_COMPILE") == "1":
     import peft
     from peft.mapping import get_peft_model as get_peft_model_original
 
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Nice, thanks ! Maybe we should add this trick somewhere in the docs (although might be not a good practice to publicise experimental features, so I would say good as is)

BenjaminBossan:
Yeah, let's first see how this affects the number of failing tests. Also, I have no idea how big the performance hit is, if it's slower than without compile there is little incentive to use it.

Issue #1688, opened 2024-04-29
Add Support for IA3 Adapters in add_weighted_adapter Method, Currently facing issue that 'IA3Model' object has no attribute 'add_weighted_adapter'


Abdullah-kwl:
### Feature request

I propose adding support for IA3 adapters in the add_weighted_adapter method used within the PEFT module of the transformers library. IA3 adapters enhance model adaptability with minimal parameter increase, offering significant benefits for efficient and effective model fine-tuning across various tasks. This feature would allow users to integrate IA3 adapters seamlessly, thereby expanding the functional capabilities of the library.

### Motivation

The motivation behind this proposal is to address the current limitation in the PEFT module of the transformers library, where IA3 adapters are supported but the merging of IA3 adapters using the add_weighted_adapter method is not implemented. This gap leads to errors such as 'IA3Model' object has no attribute 'add_weighted_adapter' when attempting to merge IA3 adapters.

The ability to merge IA3 adapters is crucial for users who rely on the PEFT module for efficient model adaptation and fine-tuning. Without this functionality, users are unable to leverage the benefits of IA3 adapters, such as reduced memory footprint and customizable model tuning, within the PEFT architecture.the 

### Your contribution

I do not have a complete implementation in mind, but I suggest starting with an evaluation of the current add_weighted_adapter function to determine the necessary modifications for supporting IA3 adapters. Collaboration with researchers familiar with IA3 and PEFT could provide insights into feasible approaches.

BenjaminBossan:
We had a PR once in #980 but there were a few non-trivial decisions to be made. If you want to work on a PR, please check out the discussion there. Pinging @alexrs also to see if there are any updates.

Abdullah-kwl:
I am using your branch [alexrs:multi-ia3](https://github.com/alexrs/peft/tree/multi-ia3) to test the addd_weighted_adapter for ia3

FIRST_ADAPTER_PATH = "/content/drive/MyDrive/FedMl_test_llm/TrainedModels/WestLake_SFT_IA3_1/Weights/Epoch_1/Job_1"
SECOND_ADAPTER_PATH = "/content/drive/MyDrive/FedMl_test_llm/TrainedModels/WestLake_SFT_IA3_2/Weights/Epoch_1/Job_1"

FIRST_ADAPTER_NAME = "first"
SECOND_ADAPTER_NAME = "second"

[...]

PR #1701, opened 2024-05-01
Add add_weighted_adapter to IA3 adapters

alexrs:
(Partially) Resolves: https://github.com/huggingface/peft/issues/1688
See https://github.com/huggingface/peft/pull/980 for context.

### What
$(IA)^3$ adapters can't be combined. This option, however, is available for other PEFT adapters such as LoRA.

### Solution
Implement `add_weighted_adapter` method for $(IA)^3$ models supporting only weighted average of adapters for now.

### DIFF
--- a/docs/source/developer_guides/model_merging.md
+++ b/docs/source/developer_guides/model_merging.md
@@ -141,11 +141,17 @@ print(tokenizer.decode(outputs[0]))
 
 
 ## Merging (IA)³ Models
-(IA)³ models support linear model merging. To merge (IA)³ models, you can use the [`~IA3Model.add_weighted_adapter`] method. This method is similar to the [`~LoraModel.add_weighted_adapter`] method, but it doesn't accept the `combination_type` parameter. Assuming we have a PEFT model and three (IA)³ adapters, we can merge them as follows:
+The (IA)³ models facilitate linear merging of adapters. To merge adapters in an (IA)³ model, utilize the `add_weighted_adapter` method from the `IA3Model` class. This method is analogous to the `add_weighted_adapter` method used in `LoraModel`, with the key difference being the absence of the `combination_type` parameter. For example, to merge three (IA)³ adapters into a PEFT model, you would proceed as follows:
 
 ```py
 adapters = ["adapter1", "adapter2", "adapter3"]
 weights = [0.4, 0.3, 0.3]
 adapter_name = "merge"
 model.add_weighted_adapter(adapters, weights, adapter_name)
+```
+
+It is recommended that the weights sum to 1.0 to preserve the scale of the model. The merged model can then be set as the active model using the `set_adapter` method:
+
+```py
+model.set_adapter("merge")
 ```
--- a/docs/source/developer_guides/model_merging.md
+++ b/docs/source/developer_guides/model_merging.md
@@ -141,7 +141,7 @@ print(tokenizer.decode(outputs[0]))
 
 
 ## Merging (IA)³ Models
-$(IA)^3$ models support linear model merging. To merge $(IA)^3$ models, you can use the `~IA3Model.add_weighted_adapter` method. This method is similar to the `~LoraModel.add_weighted_adapter` method, but it doesn't accept the `combination_type` parameter. Assuming we have a PEFT model and three $(IA)^3$ adapters, we can merge them as follows:
+(IA)³ models support linear model merging. To merge (IA)³ models, you can use the [`~IA3Model.add_weighted_adapter`] method. This method is similar to the [`~LoraModel.add_weighted_adapter`] method, but it doesn't accept the `combination_type` parameter. Assuming we have a PEFT model and three (IA)³ adapters, we can merge them as follows:
 
 ```py
 adapters = ["adapter1", "adapter2", "adapter3"]
--- a/docs/source/developer_guides/model_merging.md
+++ b/docs/source/developer_guides/model_merging.md
@@ -138,3 +138,14 @@ print(tokenizer.decode(outputs[0]))
 
 </hfoption>
 </hfoptions>
+
+
+## Merging $(IA)^3$ Models
+$(IA)^3$ models support linear model merging. To merge $(IA)^3$ models, you can use the `~IA3Model.add_weighted_adapter` method. This method is similar to the `~LoraModel.add_weighted_adapter` method, but it doesn't accept the `combination_type` parameter. Assuming we have a PEFT model and three $(IA)^3$ adapters, we can merge them as follows:
+
+```py
+adapters = ["adapter1", "adapter2", "adapter3"]
+weights = [0.4, 0.3, 0.3]
+adapter_name = "merge"
+model.add_weighted_adapter(adapters, weights, adapter_name)
+```
--- a/docs/source/developer_guides/model_merging.md
+++ b/docs/source/developer_guides/model_merging.md
@@ -140,7 +140,7 @@ print(tokenizer.decode(outputs[0]))
 </hfoptions>
 
 
-## Merging $(IA)^3$ Models
+## Merging (IA)³ Models
 $(IA)^3$ models support linear model merging. To merge $(IA)^3$ models, you can use the `~IA3Model.add_weighted_adapter` method. This method is similar to the `~LoraModel.add_weighted_adapter` method, but it doesn't accept the `combination_type` parameter. Assuming we have a PEFT model and three $(IA)^3$ adapters, we can merge them as follows:
 
 ```py
--- a/src/peft/tuners/ia3/model.py
+++ b/src/peft/tuners/ia3/model.py
@@ -13,10 +13,12 @@
 # limitations under the License.
 from __future__ import annotations
 
+import operator
 import re
 import warnings
-from dataclasses import asdict
+from dataclasses import asdict, replace
 from enum import Enum
+from functools import reduce
 from typing import Optional
 
 import torch
@@ -29,6 +31,7 @@ from peft.utils import (
     TRANSFORMERS_MODELS_TO_IA3_FEEDFORWARD_MODULES_MAPPING,
     TRANSFORMERS_MODELS_TO_IA3_TARGET_MODULES_MAPPING,
     ModulesToSaveWrapper,
+    _freeze_adapter,
     _get_submodules,
 )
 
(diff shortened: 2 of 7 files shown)


BenjaminBossan:
@alexrs Thanks a lot for reviving the feature and updating it so quickly. LMK when you think it's ready for review.

@Abdullah-kwl If you could give this a try, that would be great.

src/peft/tuners/ia3/model.py
```diff
@@ -279,17 +280,20 @@ def set_adapter(self, adapter_name: str | list[str]) -> None:
...
         self.active_adapter = adapter_name
 
-    def _prepare_adapter_config(self, peft_config, model_config):
+    @staticmethod
+    def _prepare_adapter_config(peft_config, model_config):
         if peft_config.target_modules is None:
             if model_config["model_type"] not in TRANSFORMERS_MODELS_TO_IA3_TARGET_MODULES_MAPPING:
                 raise ValueError("Please specify `target_modules` in `peft_config`")
-            peft_config.target_modules = TRANSFORMERS_MODELS_TO_IA3_TARGET_MODULES_MAPPING[model_config["model_type"]]
+            peft_config.target_modules = set(
```
alexrs, line 288: As discussed in https://github.com/huggingface/peft/pull/980#discussion_r1356720557, this seemed to be a bug.

src/peft/tuners/lora/model.py
```diff
@@ -575,9 +575,6 @@ def add_weighted_adapter(
 
         if adapter_name in list(self.peft_config.keys()):
             return
-        for adapter in adapters:
-            if adapter not in list(self.peft_config.keys()):
-                raise ValueError(f"Adapter {adapter} does not exist")
```
alexrs, line 580: This is done in `_check_add_weighted_adapter ` https://github.com/huggingface/peft/blob/77b7238b907c85f21ec756176351d85faf2aeaad/src/peft/tuners/lora/model.py#L462-L464, which is called right after

src/peft/tuners/ia3/model.py
```diff
@@ -393,3 +397,94 @@ def delete_adapter(self, adapter_name: str) -> None:
...
+        key_list = [key for key, _ in self.model.named_modules() if self.prefix not in key]
+        for key in key_list:
+            _, target, _ = _get_submodules(self.model, key)
+            if isinstance(target, IA3Layer):
+                if adapter_name in target.ia3_l:
+                    target_ia3_l = target.ia3_l[adapter_name]
+                else:
+                    continue
+
+                target_ia3_l.data = target_ia3_l.data * 0.0
```
alexrs, line 484: I guess here we could use:

```suggestion
                target_ia3_l.data = target_ia3_l.data.zero_()
```

but I tried to follow the code style used in LoRA: https://github.com/huggingface/peft/blob/77b7238b907c85f21ec756176351d85faf2aeaad/src/peft/tuners/lora/model.py#L612

https://pytorch.org/docs/stable/generated/torch.Tensor.zero_.html#torch-tensor-zero

alexrs:
@BenjaminBossan Ready for review! 👀

src/peft/tuners/ia3/model.py
```diff
@@ -279,17 +280,20 @@ def set_adapter(self, adapter_name: str | list[str]) -> None:
                 module.set_adapter(adapter_name)
         self.active_adapter = adapter_name
 
-    def _prepare_adapter_config(self, peft_config, model_config):
+    @staticmethod
```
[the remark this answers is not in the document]
BenjaminBossan: Thanks for fixing this and the set type below.

src/peft/tuners/ia3/model.py
```diff
@@ -393,3 +397,94 @@ def delete_adapter(self, adapter_name: str) -> None:
...
+        target_module_types = {type(self.peft_config[adapter].target_modules) for adapter in adapters}
+        feedforward_module_types = {type(self.peft_config[adapter].feedforward_modules) for adapter in adapters}
+        if len(target_module_types) > 1 or len(feedforward_module_types) > 1:
+            raise ValueError("All adapter configs should have the same type for target and feedforward modules.")
+
+        # Combine target and feedforward modules
+        if str in target_module_types:
+            new_target_modules = "|".join(f"({self.peft_config[adapter].target_modules})" for adapter in adapters)
+        else:
+            new_target_modules = set.union(*(self.peft_config[adapter].target_modules for adapter in adapters))
```
BenjaminBossan, line 428: Oh yeah, simpler than the reduce op we have in LoRA :+1:

src/peft/tuners/ia3/model.py
BenjaminBossan: Probably `.zero_()` is more efficient (or has the potential to be so), so I'd be fine with this change.

BenjaminBossan (changes requested):
Thanks a lot for your work. This already looks quite good. I have a few smaller comments, and then 3 more general points:

1. I haven't reviewed the tests yet. I see that you refactored `_test_weighted_combination_of_adapters` to work with both LoRA and IA³. However, all the changes make the review quite hard and there is also the risk to accidentally break the LoRA tests (e.g. skipping to test something that is currently tested). Instead, it would be better IMO if `_test_weighted_combination_of_adapters` would call a submethod `_test_weighted_combination_of_adapters_lora`, `_test_weighted_combination_of_adapters_ia3`, etc., with `_test_weighted_combination_of_adapters` being the current function body. Then you can copy and modify it for IA³. Yes, this will result in more code duplication, but for tests that doesn't matter as much and it will be more maintainable, especially since the different methods will always be quite different and thus tests will differ.
2. IIUC, right now, if I merge, say, 10 adapters, all with `weight=1`, the resulting merged adapter will possibly have very large values, right? So I should probably assign `weight=1/10` so that the norm does not shift. Maybe this should be documented.
3. How about adding a section to https://github.com/huggingface/peft/blob/main/docs/source/developer_guides/model_merging.md?

alexrs:
Hey @BenjaminBossan. I made some changes:
- Re-factored the tests to include `_test_weighted_combination_of_adapters_lora` and `_test_weighted_combination_of_adapters_ia3`. These methods are called from `_test_weighted_combination_of_adapters`.
- Added a short section to the docs. Let me know if I should expand it.

With respect to whether we should normalize the adapter weights or not, I'm not sure what is the best approach. On the one hand, I agree that having a combination of adapters with the sum of weights > 1 might not return the best results. On the other hand, I also think it is interesting for users to specify how they want their adapter weights to be normalized. In your example, `weight=1/10` is a totally valid approach, but also is taking a softmax of the weights or passing some weights as 0 if we want to do some sort of Top K combination. I'm not sure this method should be in charge of handling all of that.

Maybe we should just specify in the docs that we recommend `sum(weights) == 1`? We can also print a warning if that condition is not fulfilled.

docs/source/developer_guides/model_merging.md
```diff
@@ -138,3 +138,14 @@ print(tokenizer.decode(outputs[0]))
 
 </hfoption>
 </hfoptions>
+
+
+## Merging $(IA)^3$ Models
```
BenjaminBossan, line 143:
```suggestion
## Merging (IA)³ Models
```

LaTeX syntax doesn't seem to be supported: https://moon-ci-docs.huggingface.co/docs/peft/pr_1701/en/developer_guides/model_merging#merging-ia3-models

docs/source/developer_guides/model_merging.md
```diff
@@ -138,3 +138,14 @@ print(tokenizer.decode(outputs[0]))
...
+$(IA)^3$ models support linear model merging. To merge $(IA)^3$ models, you can use the `~IA3Model.add_weighted_adapter` method. This method is similar to the `~LoraModel.add_weighted_adapter` method, but it doesn't accept the `combination_type` parameter. Assuming we have a PEFT model and three $(IA)^3$ adapters, we can merge them as follows:
```
BenjaminBossan, line 144:
```suggestion
(IA)³ models support linear model merging. To merge (IA)³ models, you can use the [`~IA3Model.add_weighted_adapter`] method. This method is similar to the [`~LoraModel.add_weighted_adapter`] method, but it doesn't accept the `combination_type` parameter. Assuming we have a PEFT model and three (IA)³ adapters, we can merge them as follows:
```

tests/testing_common.py
```diff
@@ -1338,6 +1327,63 @@ def _test_weighted_combination_of_adapters(self, model_id, config_cls, config_kw
...
+
+        adapter_list = ["adapter1", "adapter_2", "adapter_3"]
+        weight_list = [0.5, 1.5, 1.5]
+        # Initialize the config
+        config = config_cls(
+            base_model_name_or_path=model_id,
+            **config_kwargs,
+        )
+
+        # Define a dictionary to map config types to their respective test functions
```
BenjaminBossan, line 1370: I think it's more straightforward to read (though less elegant) if we do:

```python
if isinstance(config, LoraConfig):
    self._test_weighted_combination_of_adapters_lora(...)
elif ...
```

WDYT?

tests/testing_common.py
```diff
@@ -1213,7 +1198,11 @@ def _test_weighted_combination_of_adapters(self, model_id, config_cls, config_kw
 
         # test ties re-weighting with multiple adapters
         model.add_weighted_adapter(
-            adapter_list[:2], weight_list[:2], "multi_adapter_ties_reweighting", combination_type="ties", density=0.5
+            adapter_list[:2],
```
BenjaminBossan, line 1201: This line is 117 chars, so there should be no need to break it (we have 120 limit).

BenjaminBossan (changes requested):
Thanks for the updates, I think we're almost good, I only had a few minor points.

> Maybe we should just specify in the docs that we recommend sum(weights) == 1?

Yes, my suggestion is to document that users should take care that the weights should sum up to 1 or at least be close to 1.

tests/testing_common.py
BenjaminBossan: I see what you mean but honestly, for a test I don't care so much about such issues, it's more important to me that I can quickly read from top to bottom and understand what's going on.

BenjaminBossan (approved):
Thanks a lot Alejandro, LGTM, good to have this feature finally.

@pacman100 do you also want to review?

pacman100 (approved):
Thank you, @alexrs, for adding support for combining adapters when usiung IA3!  ✨

PR #1705, opened 2024-05-03
Chunk adapter for fused qkv_proj

yxli2123:
As `microsoft/Phi-3-mini-4k-instruct` is using fused `qkv_proj` and `up_gate_proj`, LoRA adapters should be applied to each chunk instead of the entire fused matrix. 

The change lies in (1) adapter initialization and (2) forward computing. 

For (1), we create `n_chunk` adapters for one fused matrix, `default-chunk-0`, `default-chunk-1`, ... for both lora_A and lora_B.

For (2), we iterate the adapter by `adapter_name`. Once we find its chunk, we only update the associated dense result.

We have run the `make style` and `make quality`.

### DIFF
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -112,6 +112,10 @@ class LoraConfig(PeftConfig):
             Build a new stack of layers by stacking the original model layers according to the ranges specified. This
             allows expanding (or shrinking) the model without duplicating the base model weights. The new layers will
             all have separate LoRA adapters attached to them.
+        chunk_pattern(`Optional[dict]`):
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
Thanks for providing this PR. Indeed, as of now, we don't have any special provision for LoRA to deal with fused QKV layers (or any fused layers). For some models that used fused QKV, we just apply the LoRA adapter on the whole fused layer. I haven't thought this through completely, but I wonder if this should not be good enough. Do we really need to chunk the LoRA adapter?

Where there is definitely a problem is if we have fused QKV but we want to apply LoRA only to Q and V, for instance. This is currently not possible, but AFAICT, your PR wouldn't change that situation, is that right?

PR #1706, opened 2024-05-03
Autocast adapter weights if fp16/bf16

BenjaminBossan:
As discussed internally, we want to automatically cast the weights of the adapter to `float32` when using `float16`. `float16` is not conducive to stable training and raises errors when used with AMP.

Previously, we had to recommend to users to manually cast the weights if they loaded the base model in `float16`, because PEFT would choose the same `dtype` for the adapter as for the base weights. Forgetting this is a common source of errors, so we choose to automate this.

If this causes trouble, users can prevent the behavior by passing `autocast_adapter_dtype=False` to `get_peft_model`,
`PeftModel.from_pretrained`, or `PeftModel.load_adapter`.

I decided to implement the method that does the actual casting on the adapter model level, i.e. `LoraModel` et al., instead of the level of `PeftModel`. The idea was to be more flexible on how different tuner methods deal with the `dtype` casting.

Right now, I have implemented the casting only for `LoraModel`. If we think this should be applied more generally, we can also lift this to the level of `BaseTuner`, in which case LoHa, OFT, etc. would all inherit the same behavior. LMK what you think.

This PR should be reviewed carefully, as it has the potential to break existing code if something important was missed. We also need to add a note for the upcoming release text about this change in behavior.

### DIFF
--- a/docs/source/developer_guides/troubleshooting.md
+++ b/docs/source/developer_guides/troubleshooting.md
@@ -69,6 +69,12 @@ trainer = Trainer(model=peft_model, fp16=True, ...)
 trainer.train()
 ```
 
+<Tip>
+
+Starting from PEFT verion v0.11.0, PEFT automatically promotes the dtype of adapter weightes from `torch.float16` to `torch.float32` where appropriate. To _prevent_ this behavior, you can pass `autocast_adapter_dtype=False` to [`~get_peft_model`], to [`~PeftModel.from_pretrained`], and to [`~PeftModel.load_adapter`].
+
+</Tip>
+
 ## Bad results from a loaded PEFT model
 
 There can be several reasons for getting a poor result from a loaded PEFT model which are listed below. If you're still unable to troubleshoot the problem, see if anyone else had a similar [issue](https://github.com/huggingface/peft/issues) on GitHub, and if you can't find any, open a new issue.
--- a/docs/source/developer_guides/troubleshooting.md
+++ b/docs/source/developer_guides/troubleshooting.md
@@ -71,7 +71,7 @@ trainer.train()
 
 <Tip>
 
-Starting from PEFT verion v0.11.0, PEFT automatically promotes the dtype of adapter weightes from `torch.float16` to `torch.float32` where appropriate. To _prevent_ this behavior, you can pass `autocast_adapter_dtype=False` to [`~get_peft_model`], to [`~PeftModel.from_pretrained`], and to [`~PeftModel.load_adapter`].
+Starting from PEFT verion v0.11.0, PEFT automatically promotes the dtype of adapter weights from `torch.float16` and `torch.bfloat16` to `torch.float32` where appropriate. To _prevent_ this behavior, you can pass `autocast_adapter_dtype=False` to [`~get_peft_model`], to [`~PeftModel.from_pretrained`], and to [`~PeftModel.load_adapter`].
 
 </Tip>
 
--- a/docs/source/developer_guides/troubleshooting.md
+++ b/docs/source/developer_guides/troubleshooting.md
@@ -135,7 +135,7 @@ If the model's embedding layer doesn't follow the Transformer's naming scheme, y
 ```python
 model = get_peft_model(...)
 # train the model
-model.save_adapter("my_adapter", save_embedding_layers=True)
+model.save_pretrained("my_adapter", save_embedding_layers=True)
 ```
 
 For inference, load the base model first and resize it the same way you did before you trained the model. After you've resized the base model, you can load the PEFT checkpoint.
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -192,6 +192,7 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
         selected_adapters: Optional[list[str]] = None,
         save_embedding_layers: Union[str, bool] = "auto",
         is_main_process: bool = True,
+        convert_pissa_to_lora: Optional[str] = None,
         **kwargs: Any,
     ) -> None:
         r"""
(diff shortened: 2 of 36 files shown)


docs/source/developer_guides/troubleshooting.md
```diff
@@ -69,6 +69,12 @@ trainer = Trainer(model=peft_model, fp16=True, ...)
 trainer.train()
 ```
 
+<Tip>
+
+Starting from PEFT verion v0.11.0, PEFT automatically promotes the dtype of adapter weightes from `torch.float16` to `torch.float32` where appropriate. To _prevent_ this behavior, you can pass `autocast_adapter_dtype=False` to [`~get_peft_model`], to [`~PeftModel.from_pretrained`], and to [`~PeftModel.load_adapter`].
```
pacman100, line 74:
```suggestion
Starting from PEFT verion v0.11.0, PEFT automatically promotes the dtype of adapter weights from `torch.float16` to `torch.float32` where appropriate. To _prevent_ this behavior, you can pass `autocast_adapter_dtype=False` to [`~get_peft_model`], to [`~PeftModel.from_pretrained`], and to [`~PeftModel.load_adapter`].
```

src/peft/peft_model.py
```diff
@@ -391,8 +409,13 @@ def from_pretrained(
         if config.task_type not in MODEL_TYPE_TO_PEFT_MODEL_MAPPING.keys():
             model = cls(model, config, adapter_name)
         else:
-            model = MODEL_TYPE_TO_PEFT_MODEL_MAPPING[config.task_type](model, config, adapter_name)
-        model.load_adapter(model_id, adapter_name, is_trainable=is_trainable, **kwargs)
+            model = MODEL_TYPE_TO_PEFT_MODEL_MAPPING[config.task_type](
+                model, config, adapter_name, autocast_adapter_dtype=autocast_adapter_dtype
```
pacman100, line 413: `autocast_adapter_dtype` should also be passed to line 410 of the `if` conditional.

pacman100 (commented):
Thank you, @BenjaminBossan, for these crucial fixes related to stable amp fine-tuning! One major comment is that the same handling should also happen for `bfloat16` when using amp training. Everything else looks good and should be extended to all the other peft approaches!

src/peft/peft_model.py
BenjaminBossan: done

BenjaminBossan:
@pacman100 Thanks for the review, I have addressed your comments

- Fixed the typo
- Also cast bfloat16
- Also cast for other tuners, not just LoRA (i.e. LoHa, OFT, etc.)

pacman100:
> I tend to think we should do a release first and only then merge this PR. That way, it has a bit of time to be tested by folks using the main branch.

Agreed.

younesbelkada (approved):
Thanks ! This _might_ be breaking (maybe some training setups will OOM since we cast adapter weights in fp32), I would advocate to leave this feature on main for few weeks (maybe ~2 weeks) and sync with our power users to see to what extend this is breaking or not before doing a release, what do you think?

younesbelkada:
https://github.com/huggingface/peft/pull/1706#issuecomment-2101163837

Agreed with that too

BenjaminBossan:
_Status: This is ready to be merged after the PEFT v0.11 release_

notoookay:
Hi, I trained qwen2-1.5b model (with `bfloat16` dtype) via Lora, I found the dtype of trained model (merged) becomes `float32`. Is this normal? How do I cast it back to `bfloat16`?

BenjaminBossan:
Could you please show the code to reproduce this?

notoookay:
Thank you for helping. I use the code base from [open-instruct](https://github.com/allenai/open-instruct/tree/main), the part of training codes are:

```python
# construct the peft model
                args.model_name_or_path,
                from_tf=bool(".ckpt" in args.model_name_or_path),
                config=config,
                trust_remote_code=args.trust_remote_code,
                low_cpu_mem_usage=args.low_cpu_mem_usage,
                torch_dtype=torch.bfloat16,
                attn_implementation="flash_attention_2" if args.use_flash_attn else "eager",
                revision=args.model_revision,
                token=os.getenv("HF_TOKEN", None),
            )
peft_config = LoraConfig(
            task_type=TaskType.CAUSAL_LM,
            inference_mode=False,
            r=args.lora_rank,
            lora_alpha=args.lora_alpha,
            lora_dropout=args.lora_dropout,
            target_modules=["q_proj", "o_proj", "v_proj", "k_proj", "gate_proj", "up_proj", "down_proj"],
        )

save_with_accelerate(
            accelerator,
            model,
            tokenizer,
            args.output_dir,
            args.use_lora,
        )

def save_with_accelerate(
    accelerator: Accelerator,
    model: torch.nn.Module,
    tokenizer: PreTrainedTokenizer,
    output_dir: str,
    use_lora: bool = False,
) -> None:
    # set the generation config to an empty setting to be safe.
    # we usually do greedy decoding for generation, so this should be okay.
    # otherwise, we get an error thrown at save time.
    model.generation_config = transformers.GenerationConfig(
        temperature=None, top_p=None, eos_token_id=tokenizer.eos_token_id, bos_token_id=tokenizer.bos_token_id
    )

    unwrapped_model: PreTrainedModel = accelerator.unwrap_model(model)
    # When doing multi-gpu training, we need to use accelerator.get_state_dict(model) to get the state_dict.
    # Otherwise, sometimes the model will be saved with only part of the parameters.
    # Also, accelerator needs to use the wrapped model to get the state_dict.
    state_dict = accelerator.get_state_dict(model)
    if use_lora:
        # When using lora, the unwrapped model is a PeftModel, which doesn't support the is_main_process
        # and has its own save_pretrained function for only saving lora modules.
        # We have to manually specify the is_main_process outside the save_pretrained function.
        if accelerator.is_main_process:
            unwrapped_model.save_pretrained(output_dir, state_dict=state_dict)
    else:
        # don't use safetensors for saving for now
        unwrapped_model.save_pretrained(
            output_dir,
            is_main_process=accelerator.is_main_process,
            save_function=accelerator.save,
            state_dict=state_dict,
            safe_serialization=False,
        )

    if accelerator.is_main_process:
        tokenizer.save_pretrained(output_dir)
        
# merge base model and adapter

peft_config = PeftConfig.from_pretrained(args.lora_model_name_or_path)
base_model = AutoModelForCausalLM.from_pretrained(
            args.base_model_name_or_path if args.base_model_name_or_path else peft_config.base_model_name_or_path,
        )
print("Loading the lora model...")
lora_model = PeftModel.from_pretrained(base_model, args.lora_model_name_or_path)
print("Merging the lora modules...")
merged_model = lora_model.merge_and_unload()

output_dir = args.output_dir if args.output_dir else args.lora_model_name_or_path
os.makedirs(output_dir, exist_ok=True)

print(f"Saving merged model to {output_dir}...")
merged_model.save_pretrained(output_dir)
```

Codes above are from their `open_instruct/finetune.py` and `open_instruct/merge_lora.py`. Please tell me if anything is unclear.

BenjaminBossan:
When you load the base model before merging, i.e. here:

```python
peft_config = PeftConfig.from_pretrained(args.lora_model_name_or_path)
base_model = AutoModelForCausalLM.from_pretrained(
            args.base_model_name_or_path if args.base_model_name_or_path else peft_config.base_model_name_or_path,
        )
```

you are not passing `torch_dtype=torch.bfloat16` as you did the first time. Could this be the reason why you get float32 weights? Please always ensure that you load the base model in the same way.

PR #1713, opened 2024-05-06
FEAT Helper to check if a model is a PEFT model

BenjaminBossan:
After internal discussion, we decided to add this convenience function.

```python
>>> from peft.helpers import check_is_peft_model
>>> check_is_peft_model("gpt2")
False
>>> check_is_peft_model("peft-internal-testing/gpt2-lora-random")
True
>>> check_is_peft_model("peft-internal-testing/non-existing-model")
False
>>> check_is_peft_model("path/to/local/model/also/works")
```

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -106,6 +106,8 @@
       title: Layernorm tuning
     - local: package_reference/vera
       title: VeRA
+    - local: package_reference/helpers
+      title: Helpers
     title: Adapters
   - sections:
     - local: package_reference/merge_utils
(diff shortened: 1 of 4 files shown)


BenjaminBossan:
Ping @stas00

stas00:
Excellent, thank you @BenjaminBossan.

The only possible suggestion is for a different naming as `check` and `is` are both verbs, so perhaps either `is_peft_model` or `check_if_peft_model`?

I'm not sure if you have a precedent for other similar check-if utils to follow the same convention.

younesbelkada (approved):
Nice API ! Thanks !

BenjaminBossan:
> The only possible suggestion is for a different naming as `check` and `is` are both verbs, so perhaps either `is_peft_model` or `check_if_peft_model`?

Good point, I renamed it to `check_if_peft_model`. My reasoning for not using `is_peft_model` is that I often want to call the variable that name in similar situations, e.g. `is_peft_model = check_if_peft_model(...); if is_peft_model: ...`

stas00:
yes, I totally agree that:
```
is_peft_model = check_if_peft_model()
```
sounds perfect to me.

PR #1714, opened 2024-05-06
Some small cleanups in docstrings, copyright note

BenjaminBossan:


younesbelkada (approved):
Thanks for the clean up and for adding a copyright note in the helpers file !

PR #1719, opened 2024-05-09
Fix docs typo

NielsRogge:
This PR fixes the guide that explains saving embedding layers besides adapter parameters. The `save_adapter` method doesn't exist, should be `save_pretrained`.


BenjaminBossan (approved):
Thanks for fixing this mistake.

PR #1722, opened 2024-05-10
revise run_peft_multigpu.sh

abzb1:
the blank after \ makes HFparser to confuse.

with the blank, the example makes an error that the blank itself remains untreated as a argument.


BenjaminBossan (approved):
Thanks for removing the trailing whitespace.

PR #1723, opened 2024-05-10
Workflow: Add slack messages workflow

younesbelkada:
Uses the new workflow from `hf-workflows` in order to post nice messages on slack with emojis when the workflow fails !

### DIFF
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -45,31 +45,14 @@ jobs:
           push: true
           tags: huggingface/peft-cpu
 
-      - name: Post to a Slack channel
+      - name: Post to Slack
         if: always()
-        id: slack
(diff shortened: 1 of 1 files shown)


.github/workflows/build_docker_images.yml
```diff
@@ -101,32 +84,15 @@ jobs:
...
           tags: huggingface/peft-gpu
-
-      - name: Post to a Slack channel
+ 
+      - name: Post to Slack
         if: always()
-        id: slack
-        #uses: slackapi/slack-github-action@v1.25.0
-        uses: slackapi/slack-github-action@6c661ce58804a1a20f6dc5fbee7f0381b469e001
+        uses: huggingface/transformers/.github/actions/post-slack@main
```
younesbelkada, line 90:
```suggestion
        uses: huggingface/hf-workflows/.github/actions/post-slack@main
```

PR #1725, opened 2024-05-10
TST: torch compile tests

BenjaminBossan:
Right now, we don't have specific tests for torch.compile. Instead, we have a "hack" that allows to run _all_ tests with `torch.compile` if we set the environment variable `PEFT_DEBUG_WITH_TORCH_COMPILE=1`.

This is not very practical because it takes a lot of time to run all these tests with compilation enabled. Also, currently hundreds of tests are failing, which makes it impossible to understand more closely what does or does not work.

This PR removes the aforementioned "hack" and instead replaces it with a list of explicit torch.compile tests. Currently, these tests cover training/inference with a bunch of different tuner types, as well as more advanced features with LoRA (e.g. quantization, multiple adapters, etc.).

Some of these tests pass and some of them fail. This is documented now, so that users can quickly look up if their use case would be compatible with `torch.compile`. This is very useful to have, because sometimes `torch.compile` may appear to work but actually returns the wrong result. For users, it's not immediately obvious when this happens.

The test suite is not exhaustive, there are many combinations of features that could be added. However, it should be a good starting point and can be extended later.

The test suite does _not_ cover whether `torch.compile` actually accelerates the code. This may not be the case even if it works correctly (e.g. because of graph breaks). Testing this would require bigger models and more data, which is prohibitively slow to test.


docs/source/developer_guides/torch_compile.md
```diff
@@ -0,0 +1,84 @@
...
+
+Note that just because there is no error, it doesn't mean that `torch.compile` works correctly. It might give you an output but the output is incorrect. Therefore, if you're unsure if you can use `torch.compile` in your PEFT use case, please consult this document and make sure that your case is covered.
+
+Note: Unless indicated otherwise, the default `torch.compile` settings were used.
+
+## Features that work with `torch.compile`
+
+These features work with `torch.compile`. Everything listed below was tested with a causal LM:
+
+### Training and inference
```
stevhliu, line 29:
```suggestion
## Training and inference

torch.compile is supported for training with Transformers [`~transformers.Trainer`] class or a custom PyTorch loop. You could also use torch.compile for inference and generation. The table below shows which adapters are compatible and incompatible for training with torch.compile.

| torch.compile compatible | torch.compile incompatible     |
|--------------------------|--------------------------------|
| AdaLoRA                  | LoKr                           |
| BOFT                     | LoRA targeted embedding layers |
| IA3                      |                                |
| Layer Norm Tuning        |                                |
| LoHa                     |                                |
| LoRA                     |                                |
| LoRA + DoRA              |                                |
| OFT                      |                                |
| VeRA                     |                                |
```

docs/source/developer_guides/torch_compile.md
```diff
@@ -0,0 +1,84 @@
...
+
+- AdaLoRA
+- BOFT
+- IA³
+- Layer Norm Tuning
+- LoHa
+- LoRA
+- LoRA + DoRA
+- OFT
+- VeRA
```
stevhliu, line 46:
```suggestion
```

docs/source/developer_guides/torch_compile.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Advanced PEFT features
+
+These more advanced PEFT features don't work in conjunction with `torch.compile`. Tests were run with LoRA:
+
+- Using PEFT adapters with quantization (bitsandbytes)
+- Inference with multiple adapters
+- Unloading (i.e. calling `model.merge_and_unload()`)
+- Disabling adapters (i.e. using `with model.disable_adapter()`)
+- Mixed adapter batches (i.e. calling `model(batch, adapter_names=["__base__", "default", "other", ...])`)
```
stevhliu, line 75:
```suggestion
```

stevhliu, line 75:
```suggestion
## Advanced PEFT features

PEFT supports a number of features for working with adapters, such as merging and saving, but not all of them work with torch.compile. Check the table below to find out which features are compatible with torch.compile and which ones aren't.

| torch.compile compatible                                                                                   | torch.compile incompatible                                                                                                                         |
|------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
| [modules_to_save](https://huggingface.co/docs/peft/package_reference/lora#peft.LoraConfig.modules_to_save) | usage with quantization (bitsandbytes)                                                                                                             |
| merging one or more adapters                                                                               | inference with multiple adapters                                                                                                                   |
| merging multiple adapters into one adapter                                                                 | unloading weights with [merge_and_unload](https://huggingface.co/docs/peft/v0.10.0/en/package_reference/lora#peft.LoraModel.merge_and_unload)      |
|                                                                                                            | disabling adapters with [disable_adapter](https://huggingface.co/docs/peft/v0.10.0/en/package_reference/peft_model#peft.PeftModel.disable_adapter) |
|                                                                                                            | mixing adapter batches (`model(batch, adapter_names=["__base__", "default", "other"])`)                                                            |
```

docs/source/developer_guides/torch_compile.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Test cases
+
+All the use cases listed above are tested inside of `peft/tests/test_torch_compile.py`. If you want to check in more detail how we tested a certain feature, please go to that file and check the test that corresponds to your use case.
```
stevhliu, line 79: I would add a link directly to the file so it's easier for users to access.

docs/source/developer_guides/torch_compile.md
```diff
@@ -0,0 +1,84 @@
...
+<Tip>
+
+If you have another use case where you know that `torch.compile` does or does not work with PEFT, please contribute to PEFT by letting us know about this or by opening a PR to add this use case to the covered test cases.
```
stevhliu, line 82:
```suggestion
If you have another use case where you know that `torch.compile` does or does not work with PEFT, please contribute by letting us know or by opening a PR to add this use case to the covered test cases.
```

stevhliu (approved):
Nice, thanks for adding! I think it'd be more compact and easier to compare if we presented what works and doesn't side by side in a table

Makefile
```diff
@@ -53,3 +53,6 @@ transformers_tests:
 
 tests_regression:
 	python -m pytest -s --regression tests/regression/ $(if $(IS_GITHUB_CI),--report-log "regression_tests.log",)
+
+tests_torch_compile:
+	python -m pytest tests/test_torch_compile.py $(if $(IS_GITHUB_CI),--report-log "core_multi_gpu.log",)
```
younesbelkada, line 58:
```suggestion
	python -m pytest tests/test_torch_compile.py $(if $(IS_GITHUB_CI),--report-log "compile_tests.log",)
```

tests/test_torch_compile.py
```diff
@@ -0,0 +1,643 @@
...
+        assert not torch.allclose(output_base.logits, output_other.logits, atol=atol, rtol=rtol)
+        assert not torch.allclose(output_default.logits, output_other.logits, atol=atol, rtol=rtol)
+        # outputs of combined adapter != all others
+        assert not torch.allclose(output_base.logits, output_combined.logits, atol=atol, rtol=rtol)
+        assert not torch.allclose(output_default.logits, output_combined.logits, atol=atol, rtol=rtol)
+        assert not torch.allclose(output_other.logits, output_combined.logits, atol=atol, rtol=rtol)
+
+        # TODO: more tests, including tests that are expected to fail
+
+    # @pytest.mark.single_gpu_tests
```
younesbelkada, line 573: to remove?

younesbelkada (approved):
Thanks a lot for this and for explaining ! LGTM with few nits

BenjaminBossan:
@stevhliu Thanks for the feedback, I added most of your suggestions. I'm not sure about the tables though. Using tables was also my first instinct, but what I dislike about them is 1) there are many empty entries and 2) there is no correspondence between items in the same row, which is typically expected in a table. WDYT?

stevhliu:
The empty entries don't really bother me, but #2 is a good point. How about just reorganizing the sections but keeping the lists?

```md
## Training and inference
- list adapters that work and don't work for training and inference
## Advanced PEFT features
- list features that work and don't work
```

docs/source/developer_guides/torch_compile.md
```diff
@@ -0,0 +1,83 @@
...
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# torch.compile
+
+In PEFT, [torch.compile](https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html) works for some but not all features. The reason why it won't always work is because PEFT is highly dynamic in certain places (loading and switching between multiple adapters, for instance), which can cause trouble for `torch.compile`. In other places, `torch.compile` may work, but won't be as fast as expected because of graph breaks.
+
+If you don't see an error, it doesn't necessarily mean that `torch.compile` worked correctly. It might give you an output, but the output is incorrect. Check the table below to see which features work and which ones don't with torch.compile.
```
stevhliu, line 21:
```suggestion
If you don't see an error, it doesn't necessarily mean that `torch.compile` worked correctly. It might give you an output, but the output is incorrect. This guide describes what works with torch.compile and what doesn't.
```

docs/source/developer_guides/torch_compile.md
```diff
@@ -0,0 +1,83 @@
...
+
+> [!TIP]
+> Unless indicated otherwise, the default `torch.compile` settings were used.
+
+## Training and inference
+
+### Features that work with `torch.compile`
```
stevhliu, line 28: I think we can even remove these smaller subsections (Features that work with torch.compile and Features that don't work with torch.compile) to keep things cleaner and avoid duplicate subsection titles

```suggestion
```

stevhliu (approved):
Just a few more minor suggestions but LGTM!

If you do keep the smaller subsection titles, make sure you format them like `## Features that work with torch.compile[[training]]` to differentiate it from the identical title in the Advanced PEFT features section. Otherwise, clicking on the `## Features that work with torch.compile` in the Advanced PEFT features section will link to the first one.

docs/source/developer_guides/torch_compile.md
BenjaminBossan: Done, could you please do a (hopefully) last check?

stevhliu: Looks good, thanks for iterating! ❤️

PR #1734, opened 2024-05-15
DOC TST Document and test reproducibility with models using batch norm

### LINKED ISSUE #1732: Reproducibility when using a model with batch norm
### System Info

Latest version of PEFT

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

```python
model_id = "microsoft/resnet-18"

@pytest.fixture
def image_processor():
    image_processor = AutoImageProcessor.from_pretrained(model_id)
    return image_processor

@pytest.fixture
def data(image_processor):
    dataset = load_dataset("huggingface/cats-image")
    image = dataset["test"]["image"][0]
    return image_processor(image, return_tensors="pt")

def test_model_with_batchnorm(tmp_path, data):
    torch.manual_seed(0)
    model = AutoModelForImageClassification.from_pretrained(model_id)
    config = LoraConfig(target_modules=["convolution"], modules_to_save=["classifier"])

    # record outputs before training
    model.eval()
    with torch.inference_mode():
        output_before = model(**data)
    model.train()

    optimizer = torch.optim.AdamW(model.parameters(), lr=1e-3)
    batch_size = 4
    max_steps = 5 * batch_size
    labels = torch.zeros(1, 1000)
    labels[0, 283] = 1
    for i in range(0, max_steps, batch_size):
        optimizer.zero_grad()
        outputs = model(**data, labels=labels)
        loss = outputs.loss
        loss.backward()
        optimizer.step()

    model.eval()
    with torch.inference_mode():
        output_after = model(**data)
    assert torch.isfinite(output_after.logits).all()
    atol, rtol = 1e-4, 1e-4
    # sanity check: model was updated
    assert not torch.allclose(output_before.logits, output_after.logits, atol=atol, rtol=rtol)

    # check saving the model and loading it
    del model
    torch.manual_seed(0)
    model = AutoModelForImageClassification.from_pretrained(model_id)
    model = PeftModel.from_pretrained(model, tmp_path).eval()
    with torch.inference_mode():
        output_loaded = model(**data)
    # THIS FAILS
    assert torch.allclose(output_after.logits, output_loaded.logits, atol=atol, rtol=rtol)
```

### Expected behavior

After loading a model that was trained with PEFT on a base model with some kind of batch norm layer, the loaded model should produce the same output. Right now, this does not happen.

The reason is that during training, buffers for running mean etc. are updated, but they are not saved when calling `save_pretrained` on the `PeftModel` instance. Normally in PEFT, we assume that during training, the base model parameters are kept constant, which is not the case with batch norm. We only save the PEFT parameters and assume that when the user loads the base model, all parameters are restored exactly. That way, the information in the buffers is lost completely.

One possible solution would be to try to include the buffers in the PEFT adapter, which is not very pretty. For this to work, we would need to have a way to identify buffers that were updated vs those that are static. If someone knows a way to achieve this, or has a better idea how to fix this, please let us know.

**Edit**: Best suggestion so far by @kashif: Check for the `track_running_stats` and if it's `True`, save the module's buffer. This will not cover all possible corner cases, but hopefully most.

BenjaminBossan:
Fixes #1732

After loading a model that was trained with PEFT on a base model with some kind of batch norm layer, the loaded model should produce the same output. Right now, this does not happen.

The reason is that during training, buffers for running mean etc. are updated, but they are not saved when calling `save_pretrained` on the `PeftModel` instance. Normally in PEFT, we assume that during training, the base model parameters are kept constant, which is not the case with batch norm. We only save the PEFT parameters and assume that when the user loads the base model, all parameters are restored exactly. That way, the information in the buffers is lost completely.

This PR fixes this issue by saving the buffers of the batch norm layers. They are identified by checking for the presence of the `track_running_stats` attribute.

Note: One test for BOFT is currently failing, see the comment in the test file.

### DIFF
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -45,6 +45,35 @@ def get_embedding_layer_name(model, layer, is_embedding_in_target_modules):
     return None
 
 
+def get_buffers_to_save(model: torch.nn.Module, config) -> dict[str, torch.Tensor]:
+    """
+    Get the buffers to save for the given model and config.
+
+    Although the base model weights are not updated when using PEFT, some buffers may still be updated and therefore
+    need to be saved as part of the PEFT checkpoint. An example of this are the running stats of batch norm layers.
+
+    Args:
+        model (torch.nn.Module):
+            The PEFT model.
+        config (PeftConfig):
+            The PEFT config.
+
+    Returns:
+        dict[str, torch.Tensor]:
+            The buffers to save.
+    """
+    # note: config is currently not used but that may change in the future
+    buffers_to_save = {}
+    for module_name, module in model.named_modules():
+        # currently, we only deal with running stats from BatchNorm* modules
+        if not hasattr(module, "track_running_stats"):
+            continue
+        for buffer_name, buffer in module.named_buffers():
+            if buffer is not None:
+                buffers_to_save[module_name + "." + buffer_name] = buffer
+    return buffers_to_save
+
+
 def get_peft_model_state_dict(
     model, state_dict=None, adapter_name="default", unwrap_compiled=False, save_embedding_layers="auto"
 ):
@@ -71,6 +100,8 @@ def get_peft_model_state_dict(
     config = model.peft_config[adapter_name]
     if state_dict is None:
         state_dict = model.state_dict()
+
+    # TUNER SPECIFIC CODE
     if config.peft_type in (PeftType.LORA, PeftType.ADALORA):
         # to_return = lora_state_dict(model, bias=model.peft_config.bias)
         # adapted from `https://github.com/microsoft/LoRA/blob/main/loralib/utils.py`
@@ -165,11 +196,13 @@ def get_peft_model_state_dict(
     else:
         raise ValueError(f"Unknown PEFT type passed: {config.peft_type}")
 
+    # MODULES TO SAVE
     if getattr(model, "modules_to_save", None) is not None:
         for key, value in state_dict.items():
             if any(f"{module_name}.modules_to_save.{adapter_name}" in key for module_name in model.modules_to_save):
                 to_return[key.replace("modules_to_save.", "")] = value
 
+    # DEAL WITH EMBEDDINGS
     # check the common embedding layers in `target_modules` to reset `save_embedding_layers` if necessary
     is_embedding_in_target_modules = False
     if (
@@ -223,6 +256,11 @@ def get_peft_model_state_dict(
     elif save_embedding_layers:
         warnings.warn("Could not identify embedding layer(s) because the model is not a 🤗 transformers model.")
 
+    # DEAL WITH BUFFERS
+    buffers_to_save = get_buffers_to_save(model, config)
+    to_return.update(buffers_to_save)
+
+    # REMOVE ADAPTER NAME
     to_return = {k.replace(f".{adapter_name}", ""): v for k, v in to_return.items()}
     return to_return
 
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -45,35 +45,6 @@ def get_embedding_layer_name(model, layer, is_embedding_in_target_modules):
     return None
 
 
-def get_buffers_to_save(model: torch.nn.Module, config) -> dict[str, torch.Tensor]:
-    """
-    Get the buffers to save for the given model and config.
-
-    Although the base model weights are not updated when using PEFT, some buffers may still be updated and therefore
-    need to be saved as part of the PEFT checkpoint. An example of this are the running stats of batch norm layers.
-
-    Args:
-        model (torch.nn.Module):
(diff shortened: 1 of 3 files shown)


kashif (approved):

src/peft/utils/save_and_load.py
```diff
@@ -45,6 +45,35 @@ def get_embedding_layer_name(model, layer, is_embedding_in_target_modules):
...
+        dict[str, torch.Tensor]:
+            The buffers to save.
+    """
+    # note: config is currently not used but that may change in the future
+    buffers_to_save = {}
+    for module_name, module in model.named_modules():
+        # currently, we only deal with running stats from BatchNorm* modules
+        if not hasattr(module, "track_running_stats"):
+            continue
+        for buffer_name, buffer in module.named_buffers():
```
kashif, line 71: do we wanna check if `track_running_stats` is set to `True`?

BenjaminBossan: Good point. I considered this, but was hesitant. Say someone trains a model with tracking enabled, and then turns it off for some reason before saving. Then we would not save these buffers even though they're needed, do I see that right? Ideally, we would have a check if they changed vis-à-vis the base model, but I don't see a way to monitor this except for storing a copy of all these buffers, requiring extra memory.

kashif: yes... true.. ok its safer to save them you are right yeah someone can then turn it off and it starts to use the batch's summary statistics in inference mode...

BenjaminBossan:
> when merging the weights... I suppose currently it will work for `nn.Params` but now it also needs to do the same for the buffers... how is that handled?

Hmm, not sure if I understand. If I train a LoRA adapter and then merge it, its weights will be fused with the base model weights. When we load the LoRA adapter, the base model's running stats are replaced by whatever is saved in the LoRA checkpoint. As the running stats buffers are part of the base model and no LoRA is applied to them, they are not further affected by merging.

Based on your comment, I could think of another problematic case though: A user adds 2 adapters, first adapter `A`, then `B`. Let's call the running stats buffers `rs`. They train `A` first, updating the running stats to `rs_A`. Then they switch to `B` and train, which updates the running stats further to `rs_A_B`. When they now safe the adapter, we will store `rs_A_B` for both adapters, when in reality we want `rs_A` and `rs_B`.

There are probably more edge cases that result from the fact that we kinda assume that only the PEFT parameters ever change, whereas the base model parameters are fixed. I think for this particular one, we can accept this failure case for now, as the scenario I describe should be very rare (users would typically train `A` and `B` separately, not in the same process).

BenjaminBossan:
Oh, now I wonder if there isn't a different solution: Ask users to add the batch norm layers to `modules_to_save`. I haven't thought this through completely, but this may solve all the problems we discussed with batch norm. The disadvantage is that users need to explicitly set `modules_to_save`.

kashif:
yes that was my initial solution i believe...

BenjaminBossan:
Just tested the `modules_to_save` solution: When using this config: `LoraConfig(target_modules=["convolution"], modules_to_save=["classifier", "normalization"])`, the test passes even without changing the code to explicitly add buffers to the `state_dict`.

pacman100 (commented):
Thank you @BenjaminBossan for the nice minimal reproduction of the issue with batch norm and this PR to fix it. After going through the thread I see that passing batch norm layers to `modules_to_save` fixes this issue, right?

BenjaminBossan:
@pacman100 I removed the new functionality and instead changed the test to add batch norm layers to `modules_to_save` and the tests pass. I also added a troubleshooting section about this.

pacman100 (approved):
Thank you @BenjaminBossan for adding detailed docs on handling BatchNorm, adding comments on different sections of save checkpoint utility and extensive tests! ✨

tests/test_vision_models.py
```diff
@@ -0,0 +1,117 @@
...
+
+CONFIGS = {
+    "lora": LoraConfig(target_modules=["convolution"], modules_to_save=["classifier", "normalization"]),
+    "loha": LoHaConfig(target_modules=["convolution"], modules_to_save=["classifier", "normalization"]),
+    "lokr": LoKrConfig(target_modules=["convolution"], modules_to_save=["classifier", "normalization"]),
+    "oft": OFTConfig(target_modules=["convolution"], modules_to_save=["classifier", "normalization"]),
+    # TODO: cannot use BOFT because some convolutional kernel dimensions are even (64) and others odd (147). There is no
+    # common denominator for the boft_block_size except 1, but using 1 results in an error in the fbd_cuda kernel:
+    # > Error in forward_fast_block_diag_cuda_kernel: an illegal memory access was encountered
+    # "boft": BOFTConfig(target_modules=["convolution"], modules_to_save=["classifier", "normalization"], boft_block_size=2),
```
BenjaminBossan, line 36: @yfeng95 @Zeju1997 @YuliangXiu Any idea how I could fix the described issue?

PR #1742, opened 2024-05-17
FIX Allow same layer adapters on different devices

### LINKED ISSUE #1639: GPU memory degradation despite offloading to CPU (on 50+ LoRAs for Stable Diffusion)
### System Info

accelerate==0.28.0
diffusers==0.27.0
peft==0.9.0
transformers==4.38.2

Amazon EC2 instance (g5.2xlarge)
Deep Learning OSS Nvidia Driver AMI GPU PyTorch 2.0.1 (Amazon Linux 2)

### Who can help?

@pacman100 @younesbelkada

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

Problem Setup
---
I am using the `peft` library to run SDXL + LoRA inference. We aim to support thousands of LoRAs. Here is the high-level algorithm that we are using on each query.

At model setup time, we load an SDXL model:
```
t2i_pipe = StableDiffusionXLPipeline.from_pretrained(
    "stabilityai/stable-diffusion-xl-base-1.0",
    torch_dtype=torch.float16,
    variant="fp16").to("cuda")
```

Then, for each query:

**Step 1**: Load the adapter into GPU, either from disk or CPU.
```
ensure_lora_is_on_gpu(request.adapter)
t2i_pipe.set_adapters(request.adapter)
assert len(t2i_pipe.get_active_adapters()) == 1
```
where `ensure_lora_is_on_gpu` is implemented like so:
```
def ensure_lora_is_on_gpu(lora_name: str):
    global t2i_pipe

    loaded_loras = t2i_pipe.get_list_adapters().get("unet", [])
    if lora_name in loaded_loras:
        t2i_pipe.set_lora_device([lora_name], "cuda")
    else:
        t2i_pipe.load_lora_weights(os.path.join(LORA_MODELS_PATH, lora_name),
                                   weight_name="lora.safetensors",
                                   local_files_only=True,
                                   adapter_name=lora_name)
        print(f"Successfully loaded LoRA {lora_name}. {len(loaded_loras) + 1} LoRAs are currently in memory.")
```

**Step 2**: Run inference
```
images = t2i_pipe(...)
```

**Step 3**: Offload the adapter to CPU
```
t2i_pipe.set_lora_device([request.adapter], "cpu")
t2i_pipe.disable_lora()
```

Problem Symptoms
---
To test our setup, we make ~60 sequential calls to the model, each with a different LoRA. In this log file, you can see the latency gradually degrading from ~20s/query to ~43s/query.
[cpu_offloading_sagemaker_calls.txt](https://github.com/huggingface/peft/files/14938974/cpu_offloading_sagemaker_calls.txt)

Looking at the GPU memory utilization, you can see it gradually getting close to 100% (which must be the cause of degradation in latency):
cpu_offloading_gpu_memory_utilization

This is surprising to me, since we are always offloading the LoRA onto CPU after running inference.

Debugging attempt:
---
Instead of offloading to CPU in Step 3, we unloaded after every inference:

**Step 3**: Completely unload LoRA
```
t2i_pipe.unload_lora_weights()
```

Repeating the same experiment as above (60 sequential requests with different LoRAs), we see that this time latency stays constant across calls:
[unload_loras_sagemaker_calls.txt](https://github.com/huggingface/peft/files/14939001/unload_loras_sagemaker_calls.txt)

And GPU memory consumption stays constant, as expected:
unload_loras_gpu_memory_utilization

### Expected behavior

I would expect GPU memory utilization to stay constant when we offload LoRAs to the CPU (similarly to what happens when we unload them completely). It seems like our Step 3 (`t2i_pipe.set_lora_device([request.adapter], "cpu")`) has virtually no effect on GPU memory consumption.

Hypothesis 1:
---
We are missing something obvious.

Hypothesis 2 (debunked I think):
---
Maybe the GPU keeps a copy of the tensor (in addition to the one that is copied to the CPU)?

I tried modifying [this](https://github.com/huggingface/diffusers/blob/b2323aa2b76ffa90a71507f09c18792d1dba2523/src/diffusers/loaders/lora.py#L1268) line to explicitly delete the GPU tensor, but it did not help.

Before:
```
unet_module.lora_A[adapter_name].to(device)
```

After (doesn't help):
```
gpu_tensor = unet_module.lora_A[adapter_name]
gpu_tensor.to("cpu")
del gpu_tensor
```

Your help would be much appreciated!

BenjaminBossan:
Resolves #1639

The issue is that so far, we made the assumption in PEFT that all adapter weights of a specific layer are on the same device. There can be cases where it is useful to have adapters on different devices. E.g. when a user loads a lot of LoRA adapters and wants to offload those not currently in use to CPU, they would not currently be able to do so.

With this PR, we add this possibility. To achieve this, when we update an adapter layer with a new adapter, we only move that specific adapter to the device of the base layer, will not touching the other loaded adapters.

While working on this, I discovered a small bug in VeRA when adding multiple adapters, which is now also fixed.

This PR has the potential to lead to unforeseen issues, so careful review is required. After merging this, let's keep it out of releases for a while to ensure it doesn't break anything.

### DIFF
--- a/src/peft/tuners/adalora/layer.py
+++ b/src/peft/tuners/adalora/layer.py
@@ -72,11 +72,7 @@ class AdaLoraLayer(LoraLayer):
         if init_lora_weights:
             self.reset_lora_parameters(adapter_name)
 
-        if hasattr(self.get_base_layer(), "qweight"):
-            # QuantLinear
-            self.to(self.get_base_layer().qweight.device)
-        else:
-            self.to(self.get_base_layer().weight.device)
+        self._move_adapter_to_device_of_base_layer(adapter_name)
         self.set_adapter(self.active_adapters)
 
     def reset_lora_parameters(self, adapter_name):
--- a/src/peft/tuners/boft/layer.py
+++ b/src/peft/tuners/boft/layer.py
@@ -310,18 +310,11 @@ class BOFTLayer(BaseTunerLayer):
 
         self.reset_boft_parameters(adapter_name, init_weights)
 
-        weight = getattr(self, "weight", None)
-        if weight is not None:
-            # the layer is already completely initialized, this is an update
-            if weight.dtype.is_floating_point or weight.dtype.is_complex:
-                self.to(weight.device, dtype=weight.dtype)
-            else:
-                self.to(weight.device)
-
         # set the boft block size and number
         self.boft_block_size[adapter_name] = boft_block_size
         self.boft_block_num[adapter_name] = boft_block_num
 
+        self._move_adapter_to_device_of_base_layer(adapter_name)
         self.set_adapter(self.active_adapters)
 
     def reset_boft_parameters(self, adapter_name, init_weights):
@@ -742,18 +735,12 @@ class Conv2d(nn.Module, BOFTLayer):
 
         self.reset_boft_parameters(adapter_name, init_weights)
 
-        weight = getattr(self, "weight", None)
-        if weight is not None:
-            # the layer is already completely initialized, this is an update
-            if weight.dtype.is_floating_point or weight.dtype.is_complex:
-                self.to(weight.device, dtype=weight.dtype)
-            else:
-                self.to(weight.device)
-        self.set_adapter(self.active_adapters)
-
         # set the boft block size and number
         self.boft_block_size[adapter_name] = boft_block_size
         self.boft_block_num[adapter_name] = boft_block_num
+
+        self._move_adapter_to_device_of_base_layer(adapter_name)
+        self.set_adapter(self.active_adapters)
 
     def merge(self, safe_merge: bool = False, adapter_names: Optional[list[str]] = None) -> None:
         """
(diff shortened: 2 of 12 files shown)


pacman100 (approved):
Thank you @BenjaminBossan for refactoring out the logic for setting the device of the adapters without disturbing the devices of the prior added adapters! 🚀  Very useful for offloading when working with a vast numbers of adapters.

younesbelkada (approved):
Thanks a lot for this !

BenjaminBossan:
Thanks for the reviews. I'll wait a bit in case @iuliaturc has time to check if this PR fixes the initial issue. If we don't hear back, I'll merge in a few days.

We should let this PR "simmer" for a bit since there is a small probability that this will break some edge case we haven't thought of.

iuliaturc:
Thanks so much for the PR! I left a comment [here](https://github.com/huggingface/peft/issues/1639#issuecomment-2125950827).

TL;DR is that, indeed, only one LoRA seems to be loaded at a time, but the fix doesn't seem to address the original problem (that latency keeps creeping up the more calls we make).

BenjaminBossan:
Thanks for the confirmation.

PR #1747, opened 2024-05-20
Patch for Cambricon MLUs test 

huismiling:
Cambricon MLUs has been supported by peft ([huggingface/peft/pull/1687](https://github.com/huggingface/peft/pull/1687)).
But the pytest have some failed cases. It's mostly about tolerances. 
This patch fix the failed case of tolerances.


tests/testing_common.py
```diff
@@ -464,13 +464,13 @@ def _test_merge_layers_fp16(self, model_id, config_cls, config_kwargs):
         if ("gpt2" in model_id.lower()) and (config_cls != LoraConfig):
             self.skipTest("Merging GPT2 adapters not supported for IA³ (yet)")
 
-        model = self.transformers_class.from_pretrained(model_id, torch_dtype=torch.float16)
+        model = self.transformers_class.from_pretrained(model_id)
         config = config_cls(
```
huismiling, line 468: `torch_dtype=torch.float16` leads to an error.
`RuntimeError: "addmm_impl_cpu_" not implemented for 'Half'`

huismiling:
@BenjaminBossan Hi, could you help to reivew this PR? Thx.

BenjaminBossan:
Thanks @huismiling, could you please run `make style` on this PR?

huismiling:
`make style` has been done. Thx. @BenjaminBossan

tests/testing_common.py
BenjaminBossan: Hmm, I cannot replicate this, whether with or without GPU. The idea of this test is exactly to check that this error does not occur with fp16, so not using this dtype is counter-productive. Is this only occurring with MLU devices?

BenjaminBossan (changes requested):
Thanks for these adjustments. Are the results from the tests on MLU visible somewhere?

Overall, the changes LGTM, I just have one concern about the float16 test. Please check my comment.

huismiling:
The results from the tests on MLU are as follows. More details is in attachment.  
[mlu_pytest.txt](https://github.com/huggingface/peft/files/15412432/mlu_pytest.txt)

`== 4368 passed, 1483 skipped, 1 xfailed, 2981 warnings in 5991.03s (1:39:51) ===
`

tests/testing_common.py
BenjaminBossan: Okay, so instead of changing the dtype, how about skipping the test if an old pytorch version is detected?

huismiling: Hmm, maybe we can use fp16 with pt>=2.3, and fp32 with pt<2.3 ?

BenjaminBossan: We really don't need to test merging with fp32 here, as it's tested extensively in other tests. This test is very specifically for merging with fp16, so if we don't use fp16, we can skip it.

huismiling: Hmm, I found that it is the device "cpu" leads error. 
When device is changed to self.torch_device as I fixed, MLU tests is OK with torch.float16 .
@BenjaminBossan Will this test use "cpu" device? If not, it isn't need to skip test for pt2.1 .
https://github.com/huggingface/peft/blob/cb0bf077744d11524ec6f68d920f4cfe4ef3e8f3/tests/testing_common.py#L473

BenjaminBossan: So IIRC, there is an error when using CPU + float16 + old PyTorch. If we change either of those variables, there is no error. On CI, we have a new PyTorch version, so it passes, despite using CPU.

If we switch to `self.torch_device`, it depends, because the device is auto-inferred based on what devices are available. So on our CI, this would still be CPU. On yours, it might not, but then we don't really test what was the original intent, namely that float16 works on CPU.

I assume this fails on your CI because it uses an older PyTorch version. This is why I suggested to just skip the test with older PyTorch versions. If you want, you could add a specific test for merging float16 with MLU, which would be skipped if the device is not available.

huismiling: Got it! Skipping test_merge_layers_fp16 for pt2.1 when cpu device, this should be OK.

tests/testing_common.py
```diff
@@ -464,13 +465,16 @@ def _test_merge_layers_fp16(self, model_id, config_cls, config_kwargs):
         if ("gpt2" in model_id.lower()) and (config_cls != LoraConfig):
             self.skipTest("Merging GPT2 adapters not supported for IA³ (yet)")
 
+        if (self.torch_device in ["cpu"]) and (version.parse(torch.__version__) <= version.parse(2.1)):
```
BenjaminBossan, line 468:
```suggestion
        if (self.torch_device in ["cpu"]) and (version.parse(torch.__version__) <= version.parse("2.1")):
```

BenjaminBossan (changes requested):
Thanks for adding the skipping logic. There's just a small typo there, the rest should be good.

tests/testing_common.py
huismiling: Sorry for that. Fixed !

BenjaminBossan (approved):
Fantastic, thanks for the PR. Keep us in the loop if tests should start failing for Cambricon MLUs.

PR #1791, opened 2024-05-22
FIX Use correct attribute name for HQQ in merge

BenjaminBossan:
Without this fix, [test_hqq_lora_model_outputs](https://github.com/huggingface/peft/actions/runs/9184294503/job/25256361781) currently fails.


younesbelkada (approved):
Thanks a lot @BenjaminBossan

BenjaminBossan:
Thanks for the review, will merge once CI is green (these timeout errors...).

PR #1792, opened 2024-05-22
TST Add regression test for DoRA, VeRA, BOFT, LN Tuning

BenjaminBossan:
These new methods were added but the regression tests were not extended yet. This PR adds regression tests for these methods. The regression artifacts have been pushed based on PEFT v0.11.1. The new tests pass locally.


younesbelkada (approved):
Thanks a lot for adding these regression tests !

PR #1799, opened 2024-05-24
FIX BOFT device error after PR 1742

BenjaminBossan:
Fixes [failing CI](https://github.com/huggingface/peft/actions/runs/9217614705/job/25359847525).

PR #1742 introduced the feature that adapters of the same layer can be on different devices. A new method was introduced that is responsible for moving the parameters related to a specific adapter in a consistent way.

In BOFT, however, one parameter was overlooked, `boft_P`. This parameter is not stored inside a `ParameterDict` or `ModuleDict`, hence it was not moved. The reason is (presumably) that this parameter is shared between all BOFT adapters, as it's always identical. However, this clashes with having different adapters on different devices.

To solve this, the parameter is now moved on the fly to the correct device during the forward pass.

Alternatively, we could store this parameter inside a `ParameterDict` or `BufferDict` and have one copy per adapter. This would probably be faster at the cost of more memory.

I ran the BOFT tests locally with GPU and they passed, so this should fix the CI.

### DIFF
--- a/src/peft/tuners/boft/layer.py
+++ b/src/peft/tuners/boft/layer.py
@@ -572,8 +572,9 @@ class Linear(nn.Module, BOFTLayer):
             block_diagonal_butterfly = torch.block_diag(*torch.unbind(orth_rotate_butterfly))
             block_diagonal_butterfly = block_diagonal_butterfly.unsqueeze(0)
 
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Ping @yfeng95, @Zeju1997, @YuliangXiu

younesbelkada (approved):
Thanks for fixing !

PR #1808, opened 2024-05-29
CI Make torch compile tests run on GPU

BenjaminBossan:
Many of these tests require a GPU to run, so use custom runners.

Code was mostly copied from existing workflows, I hope that this works.

### DIFF
--- a/.github/workflows/torch_compile_tests.yml
+++ b/.github/workflows/torch_compile_tests.yml
@@ -19,17 +19,13 @@ env:
 
 jobs:
   run_tests_with_compile:
-    runs-on: [self-hosted, single-gpu, nvidia-gpu, t4, ci]
-    strategy:
-      fail-fast: false
-      matrix:
-          docker-image-name: ["huggingface/peft-gpu-bnb-latest:latest"]
+    runs-on: [self-hosted, single-gpu, nvidia-gpu, a10, ci]
     env:
       PEFT_DEBUG_WITH_TORCH_COMPILE: 1
       CUDA_VISIBLE_DEVICES: "0"
-      TEST_TYPE: "single_gpu_${{ matrix.docker-image-name }}"
+      TEST_TYPE: "single_gpu_huggingface/peft-gpu-bnb-latest:latest"
     container:
-      image: ${{ matrix.docker-image-name }}
+      image: "huggingface/peft-gpu-bnb-latest:latest"
       options: --gpus all --shm-size "16gb" --ipc host -v /mnt/cache/.cache/huggingface:/mnt/cache/
     defaults:
       run:
--- a/.github/workflows/torch_compile_tests.yml
+++ b/.github/workflows/torch_compile_tests.yml
@@ -11,32 +11,46 @@ on:
         required: false
         default: false
 
+env:
+  RUN_SLOW: "yes"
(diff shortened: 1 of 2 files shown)


.github/workflows/torch_compile_tests.yml
```diff
@@ -11,32 +11,46 @@ on:
...
+env:
+  RUN_SLOW: "yes"
+  IS_GITHUB_CI: "1"
+  # To be able to run tests on CUDA 12.2
+  NVIDIA_DISABLE_REQUIRE: "1"
+
 jobs:
   run_tests_with_compile:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, single-gpu, nvidia-gpu, t4, ci]
```
younesbelkada, line 22:
```suggestion
    runs-on: [self-hosted, single-gpu, nvidia-gpu, a10, ci]
```

younesbelkada (approved):
Thanks ! Note we do have access to A10 GPUs in case you want to run them on more recent hardware

.github/workflows/torch_compile_tests.yml
```diff
@@ -11,32 +11,46 @@ on:
...
+    strategy:
+      fail-fast: false
+      matrix:
+          docker-image-name: ["huggingface/peft-gpu-bnb-latest:latest"]
```
younesbelkada, line 26: nit: we don't really need matrix here

younesbelkada, line 26:
```suggestion
```

.github/workflows/torch_compile_tests.yml
```diff
@@ -11,32 +11,46 @@ on:
...
     env:
       PEFT_DEBUG_WITH_TORCH_COMPILE: 1
+      CUDA_VISIBLE_DEVICES: "0"
+      TEST_TYPE: "single_gpu_${{ matrix.docker-image-name }}"
+    container:
+      image: ${{ matrix.docker-image-name }}
```
younesbelkada, line 32:
```suggestion
      image: "huggingface/peft-gpu-bnb-latest:latest"
```

.github/workflows/torch_compile_tests.yml
```diff
@@ -11,32 +11,46 @@ on:
...
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, single-gpu, nvidia-gpu, t4, ci]
+    strategy:
+      fail-fast: false
+      matrix:
+          docker-image-name: ["huggingface/peft-gpu-bnb-latest:latest"]
     env:
       PEFT_DEBUG_WITH_TORCH_COMPILE: 1
+      CUDA_VISIBLE_DEVICES: "0"
+      TEST_TYPE: "single_gpu_${{ matrix.docker-image-name }}"
```
younesbelkada, line 30:
```suggestion
      TEST_TYPE: "single_gpu_huggingface/peft-gpu-bnb-latest:latest"
```

younesbelkada (commented):
Left also small nits to make the workflow file clearer

BenjaminBossan:
@younesbelkada Thanks for the review. I kept the test matrix with a single entry to make it easier to extend in the future but as is, it's indeed not needed, so it's removed now.

PR #1811, opened 2024-05-30
MNT Remove deprecated use of load_in_8bit

BenjaminBossan:
Don't pass `load_in_8bit` to `AutoModel.from_pretrained`, instead use `BitsAndBytesConfig`.

There was already a PR to clean this up (#1552) but a slightly later PR (#1518) re-added this usage.

### DIFF
--- a/tests/test_common_gpu.py
+++ b/tests/test_common_gpu.py
@@ -826,7 +826,7 @@ class PeftGPUCommonTests(unittest.TestCase):
         # check that we can pass mixed adapter names to the model
         # note that with 8bit, we have quite a bit of imprecision, therefore we use softmax and higher tolerances
         torch.manual_seed(3000)
-        bnb_config = BitsAndBytesConfig(quantization_config=BitsAndBytesConfig(load_in_8bit=True))
+        bnb_config = BitsAndBytesConfig(load_in_8bit=True)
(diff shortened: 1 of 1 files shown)


tests/test_common_gpu.py
```diff
@@ -826,11 +826,7 @@ def test_8bit_lora_mixed_adapter_batches_lora(self):
         # check that we can pass mixed adapter names to the model
         # note that with 8bit, we have quite a bit of imprecision, therefore we use softmax and higher tolerances
         torch.manual_seed(3000)
-        bnb_config = BitsAndBytesConfig(
-            load_in_8bit=True,
-            bnb_4bit_use_double_quant=False,
-            bnb_4bit_compute_dtype=torch.float32,
-        )
+        bnb_config = BitsAndBytesConfig(quantization_config=BitsAndBytesConfig(load_in_8bit=True))
```
younesbelkada, line 829:
```suggestion
        bnb_config = BitsAndBytesConfig(load_in_8bit=True)
```
Maybe a regex error 🤯

younesbelkada (approved):
Makes sense, thanks !

tests/test_common_gpu.py
BenjaminBossan: Good catch, yeah, overeager search and replace.

PR #1813, opened 2024-06-01
MoRA support

winglian:
Copied from the official implementation at https://github.com/kongds/MoRA by @kongds

Paper: https://arxiv.org/abs/2405.12130

### DIFF
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -275,7 +275,8 @@ class LoraConfig(PeftConfig):
         default=1,
         metadata={
             "help": (
-                "Enable MoRA"
+                "type 1 (Sharing) for large lora ranks, Eq. 6 in paper. "
+                "type 6 (RoPE based) for small lora ranks, Eq. 9 in paper."
             )
         },
     )
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -259,6 +259,26 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    use_mora: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Enable MoRA: High-Rank Updating for Parameter-Efﬁcient Fine-Tuning. This technique employs a square "
+                "matrix to achieve high-rank updating while maintaining the same number of trainable parameters. "
+                "To achieve it, MoRA introduces the corresponding non-parameter operators to reduce the input dimension "
+                "and increase the output dimension for the square matrix. Furthermore, these operators ensure that the "
+                "weight can be merged back into LLMs, which makes the method can be deployed like LoRA."
+            )
+        },
+    )
+    mora_type: int = field(
+        default=1,
+        metadata={
+            "help": (
+                "Enable MoRA"
+            )
+        },
+    )
     # Enables replicating layers in a model to expand it to a larger model.
     layer_replication: Optional[list[tuple[int, int]]] = field(
         default=None,
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -52,6 +52,8 @@ class LoraLayer(BaseTunerLayer):
         self.merged_adapters = []
         self.use_dora: dict[str, bool] = {}
         self.lora_magnitude_vector = torch.nn.ModuleDict()  # for DoRA
+        self.use_mora: dict[str, bool] = {}
+        self.mora_type: dict[str, int] = {}
         self._caches: dict[str, Any] = {}
         self.kwargs = kwargs
 
@@ -91,7 +93,8 @@ class LoraLayer(BaseTunerLayer):
         self.out_features = out_features
 
     def update_layer(
-        self, adapter_name, r, lora_alpha, lora_dropout, init_lora_weights, use_rslora, use_dora: bool = False
+        self, adapter_name, r, lora_alpha, lora_dropout, init_lora_weights, use_rslora, use_dora: bool = False,
+        use_mora: bool = False, mora_type: int = 1,
     ):
         # This code works for linear layers, override for other layer types
         if r <= 0:
@@ -105,20 +108,35 @@ class LoraLayer(BaseTunerLayer):
             lora_dropout_layer = nn.Identity()
 
         self.lora_dropout.update(nn.ModuleDict({adapter_name: lora_dropout_layer}))
-        # Actual trainable parameters
-        self.lora_A[adapter_name] = nn.Linear(self.in_features, r, bias=False)
-        self.lora_B[adapter_name] = nn.Linear(r, self.out_features, bias=False)
-        if use_rslora:
-            self.scaling[adapter_name] = lora_alpha / math.sqrt(r)
-        else:
-            self.scaling[adapter_name] = lora_alpha / r
 
-        if isinstance(init_lora_weights, str) and init_lora_weights.startswith("pissa"):
-            self.pissa_init(adapter_name, init_lora_weights)
-        elif init_lora_weights == "loftq":
-            self.loftq_init(adapter_name)
-        elif init_lora_weights:
-            self.reset_lora_parameters(adapter_name, init_lora_weights)
+        if use_mora:
+            new_r = int(math.sqrt((self.in_features + self.out_features)*r)+0.5)
+            if mora_type == 6:
+                # type 6 require new_r to be even for RoPE
+                new_r = new_r//2*2
+
+            self.lora_A[adapter_name] = nn.Linear(new_r, new_r, bias=False)
+            self.r[adapter_name] = new_r
+
+            nn.init.zeros_(self.lora_A[adapter_name].weight)
+            self.lora_B[adapter_name] = self.lora_A[adapter_name]
+            self.use_mora[adapter_name] = True
+            self.scaling[adapter_name] = 1.0
+        else:
+            # Actual trainable parameters
+            self.lora_A[adapter_name] = nn.Linear(self.in_features, r, bias=False)
+            self.lora_B[adapter_name] = nn.Linear(r, self.out_features, bias=False)
+            if use_rslora:
+                self.scaling[adapter_name] = lora_alpha / math.sqrt(r)
+            else:
+                self.scaling[adapter_name] = lora_alpha / r
+
+            if isinstance(init_lora_weights, str) and init_lora_weights.startswith("pissa"):
+                self.pissa_init(adapter_name, init_lora_weights)
+            elif init_lora_weights == "loftq":
+                self.loftq_init(adapter_name)
+            elif init_lora_weights:
+                self.reset_lora_parameters(adapter_name, init_lora_weights)
 
         # call this before dora_init
         self._move_adapter_to_device_of_base_layer(adapter_name)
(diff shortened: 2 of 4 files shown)


src/peft/tuners/lora/layer.py
```diff
@@ -219,6 +244,76 @@ def dora_init(self, adapter_name: str) -> None:
...
+                pad_size = r - in_f % r
+                x = torch.cat([x, x[..., :pad_size]], dim=-1)
+                sum_inter += 1
+            in_x = x.view(*x.shape[:-1], sum_inter, r).sum(dim=-2)
+        elif mora_type == 2 or mora_type == 3:
+            mr, nr = in_f//r+1, in_f//r
+            m, n = in_f - r*nr, r*mr - in_f
+            mm, nn_ = m*mr, n * nr
+            if m > 0:
+                x_m, x_n = x[..., :mm], x[..., nn_:]
```
winglian, line 267: the original implementation had:
```
x_m, x_n = x[..., :mm], x[..., mm:]
```

but I'm guessing @kongds intended to use nn_ as the index instead?

src/peft/tuners/lora/config.py
```diff
@@ -259,6 +259,27 @@ class LoraConfig(PeftConfig):
...
+    )
+    mora_type: int = field(
+        default=1,
+        metadata={
+            "help": (
+                "type 1 (Sharing) for large lora ranks, Eq. 6 in paper. "
+                "type 6 (RoPE based) for small lora ranks, Eq. 9 in paper."
+            )
+        },
+    )
```
winglian, line 282: I would prefer this somehow be an enum, because the original code references possible values for `mora_type` as 1, 2, 3, 4, 6. However only 1 and 6 are ever described anywhere.

winglian: I think mora_type might be best defined as 3 different flags. where 1, 2, 3,& 4 are permutations of 2 of the flags, and 6 is the rope implementation. @kongds

src/peft/tuners/lora/layer.py
kongds: The original implementation should be correct. `mm` is used to split `x` into two parts.
 Detailed explanations can be found https://github.com/kongds/MoRA/issues/7#issuecomment-2143690518.

src/peft/tuners/lora/config.py
kongds: Thanks for your advice.

For MoRA, we only use the `1` and `6` flags.
The flags like `2`, `3`, and `4` are internally used for ReMoRA to change types of sharing groups during training.

src/peft/tuners/lora/layer.py
winglian: Thanks, so I can drop the `n * nr` part from this?

kongds: Yes. Thanks for the PR. However,  if merging it into peft, I think it’s better to separate MoRA from LoRA by using new classes like `MoraConfig` and `MoraModel`, since MoRA does not use low-rank matrices in LoRA (for example, the current implementation only uses matrix A in LoRA).

src/peft/tuners/lora/config.py
BenjaminBossan: > I would prefer this somehow be an enum

Even better would be a string enum with some illustrative names. As a user, I would appreciate not having to look up each time what type 1 or type 6 is. The descriptions in the help could also be elaborated.

src/peft/tuners/lora/config.py
```diff
@@ -259,6 +259,27 @@ class LoraConfig(PeftConfig):
...
     )
+    use_mora: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Enable MoRA: High-Rank Updating for Parameter-Efﬁcient Fine-Tuning. This technique employs a square "
+                "matrix to achieve high-rank updating while maintaining the same number of trainable parameters. "
+                "To achieve it, MoRA introduces the corresponding non-parameter operators to reduce the input dimension "
+                "and increase the output dimension for the square matrix. Furthermore, these operators ensure that the "
+                "weight can be merged back into LLMs, which makes the method can be deployed like LoRA."
```
BenjaminBossan, line 270: Nice description. The last sentence is probably missing a word?

> which makes the method can be deployed like LoRA

Also, I would add that when using MoRA, IIUC the rank `r` is basically inferred such that the number of trainable parameters corresponds to that rank for LoRA.

src/peft/tuners/lora/layer.py
```diff
@@ -105,20 +108,35 @@ def update_layer(
...
+            new_r = int(math.sqrt((self.in_features + self.out_features)*r)+0.5)
+            if mora_type == 6:
+                # type 6 require new_r to be even for RoPE
+                new_r = new_r//2*2
+
+            self.lora_A[adapter_name] = nn.Linear(new_r, new_r, bias=False)
+            self.r[adapter_name] = new_r
+
+            nn.init.zeros_(self.lora_A[adapter_name].weight)
+            self.lora_B[adapter_name] = self.lora_A[adapter_name]
```
BenjaminBossan, line 122: Would this not result in redundant entries in the `state_dict`?

BenjaminBossan (changes requested):
Thanks a lot for creating this PR to add MoRA to PEFT. I actually had a look at the paper and code last week and considered adding it to PEFT but was still busy with other things :)

Before I do an in-depth review, I wanted to discuss some bigger design decisions. As in the original repo, you're adding this on top of LoRA. My hunch would be to create a separate adapter though. Here are my thoughts:

- From a technical point of view, it is sufficiently different. E.g. there is only one adapter matrix instead of 2 for LoRA. This line is a strong indicator: `self.lora_B[adapter_name] = self.lora_A[adapter_name]`.
- If we add it to LoRA, we probably have to add a bunch of checks for stuff that does or does not work with MoRA (e.g. MoRA + DoRA, MoRA + LoftQ, `mora_type` + anything other than MoRA). If a separate method, we just don't add the options that are incompatible.
- In terms of code complexity, I think it would be okay as is, but over time, more options make the code harder to maintain.

Of course, there are disadvantages:

- More differences when compared to original repo.
- Probably a lot of code duplication.
- When LoRA is updated, we might have to port the change to MoRA.

When it comes to code duplication, we might consider using the [`lycoris` abstractions](https://github.com/huggingface/peft/blob/main/src/peft/tuners/lycoris_utils.py) that are providing most of the required boiler plate (check for instance the LoHa, LoKr, OFT). But copy-pasting a bunch of LoRA code would also be fine.

Overall, I'm leaning in favor of having a separate method, but LMK what you think.

Apart from this bigger design decision, here are some more points we'd need for this PR to be complete:

- Please always run `make quality`
- If code is copied (almost) verbatim, it's useful to add a comment with a reference
- There should be unit tests, check the tests for the other methods. It can be useful to check other PRs on how to test new method (e.g. [VeRA](https://github.com/huggingface/peft/pull/1564/files) for a completely new method or [DoRA](https://github.com/huggingface/peft/pull/1474/files) for a new LoRA option). If you have questions, don't hesitate to ask.
- For usability and discoverability, it's really important to add docs and examples.

src/peft/tuners/lora/config.py
winglian: Agreed. I'm not exactly sure what the two flags for the sharing groups would be called though.

kallewoof:
I just tried training a model using this code. Admittedly a bit of patchwork needed to get it to work and I may have fatfingered something somewhere.

The resulting adapter was, however, broken.

Here is an example of it merged with failspy's l3-8b abliterated v3.0:
```
./main --temp 0.5 -t 11 -ngl 25 -c 2048 -n 512 -m ./mora-output-q8_0.gguf -p "<|begin_of_text|><|start_header_id|>user<|end_header_id|>

What are some fun facts about pigs?<|eot_id|><|start_header_id|>assistant<|end_header_id|>

"
Log start
main: build = 3083 (adc9ff38)
main: built with cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0 for x86_64-linux-gnu
main: seed  = 1718342921
[...]
▅ <|begin_of_text|><|start_header_id|>user<|end_header_id|>

What are some fun facts about pigs?<|eot_id|><|start_header_id|>assistant<|end_header_id|>

What are some fun▅▅▅ [end of text]
```
In an existing Silly Tavern chat, when doing a regeneration, it outputted "Thelisten" and then stopped.

I initially did a training run with constant lr 1e-4, but eval loss did not drop as I wanted so I dropped to 6e-5 and it got a bit better. Are there other gotchas with doing MoRA training compared to LoRA that could have caused this?

kongds:
> I just tried training a model using this code. Admittedly a bit of patchwork needed to get it to work and I may have fatfingered something somewhere.
> 
> The resulting adapter was, however, broken.
> 
> Here is an example of it merged with failspy's l3-8b abliterated v3.0:
> 
> ```
> $ ./main --temp 0.5 -t 11 -ngl 25 -c 2048 -n 512 -m ./mora-output-q8_0.gguf -p "<|begin_of_text|><|start_header_id|>user<|end_header_id|>
> 
> What are some fun facts about pigs?<|eot_id|><|start_header_id|>assistant<|end_header_id|>
> 
> "
> Log start
> main: build = 3083 (adc9ff38)
> main: built with cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0 for x86_64-linux-gnu
> main: seed  = 1718342921
> [...]
> ▅ <|begin_of_text|><|start_header_id|>user<|end_header_id|>
> 
> What are some fun facts about pigs?<|eot_id|><|start_header_id|>assistant<|end_header_id|>
> 
> What are some fun▅▅▅ [end of text]
> ```
> 
> No matter what I do, the output ends up repeating my prompt and ending at some point a few tokens in.
> 
> In an existing Silly Tavern chat, when doing a regeneration, it outputted "Thelisten" and then stopped.
> 
> I initially did a training run with constant lr 1e-4, but eval loss did not drop as I wanted so I dropped to 6e-5 and it got a bit better. Are there other gotchas with doing MoRA training compared to LoRA that could have caused this?
> 
> Here's another example where I forgo the llama 3 template entirely:
> 
> ```
> $ ./main --temp 2.5 -t 11 -ngl 25 -c 2048 -n 512 -m ./mora-output-q8_0.gguf -p "What are some fun facts about pigs?
> [...]
> ▅ What are some fun facts about pigs?
> 
>  activityapeuntoape лет lowprocess▅anaprocess overse H activity activity
> ```

Hello, thanks for your interesting to our works. But this code has some bugs in implementation like this line.
```
                x_m, x_n = x[..., :mm], x[..., nn_:]
```
which should be 
```
                x_m, x_n = x[..., :mm], x[..., mm:]
```
Could you try the original version?

kallewoof:
> Hello, thanks for your interesting to our work. But this code has some bugs in implementation like this line.
> 
> ```
>                 x_m, x_n = x[..., :mm], x[..., nn_:]
> ```
> 
> which should be
> 
> ```
>                 x_m, x_n = x[..., :mm], x[..., mm:]
> ```
> 
> Could you try the original version?

This line does not affect me, as I did not change the mora_type from its default (1). The above code only applies to mora_type=2 or 3. Correct me if I'm wrong. (I did see this and put in a breakpoint to see if it would reach the code in question but it did not.)

kongds:
> > Hello, thanks for your interesting to our work. But this code has some bugs in implementation like this line.
> > ```
> >                 x_m, x_n = x[..., :mm], x[..., nn_:]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which should be
> > ```
> >                 x_m, x_n = x[..., :mm], x[..., mm:]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could you try the original version?
> 
> This line does not affect me, as I did not change the mora_type from its default (1). The above code only applies to mora_type=2 or 3. Correct me if I'm wrong. (I did see this and put in a breakpoint to see if it would reach the code in question but it did not.)

Yes, it is for mora_type=2 or 3. Could you provide the settings of MoRA?

kallewoof:
I think my comments above can be disregarded for the purposes of this PR. The dataset I used was most likely not suitable for MoRA training. I intend to play with it more but haven't had time yet.

It would be a pity if this dragged out indefinitely.

Issue #1816, opened 2024-06-03
Support for different layer shapes for VeRA


dkopi:
### Feature request

Current implementation of VeRA does not allow to adapt layers of different shapes, as the basis matrices A and B are shared across all adapted layers and need to have compatible shapes. Solution to that would be creating the largest required A & B matrices, and slicing it accordingly for each adapted layer.

### Motivation

The current constraint is significantly limiting the scope of applications for VeRA.

### Your contribution

I'll do the PR.

PR #1817, opened 2024-06-03
Support for different layer shapes for VeRA

dkopi:
PR for #1816.


BenjaminBossan:
Thanks for the PR!

> Solution to that would be creating the largest required A & B matrices, and slicing it accordingly for each adapted layer.

I haven't looked into the details yet, but I wondered about the reasoning for this approach. You probably also considered to save one set of weights for each unique shape (perhaps transposition invariant). What would you think is the advantage of the proposed solution, does it save on average on the number of parameters, does it work better? How would you handle parameters with different dimensions, like `Conv2d`?

dkopi:
Slicing creates a view of the original tensor so in most cases it *should* be more memory-efficient than creating separate sets of A&B, and it better fits the current approach which is - one pair of A&B for the whole adapter. For layers like Conv2d, we could slice the large A&B first, and then create a view to give it proper dimensions.

Orthogonally to that, we could have an option in the config to have separate A&B for *every* adapted layer, that would be the equivalent of the ablation from the paper (table 7). I don't see any benefit of having something in between - a separate pair per different shape.

BenjaminBossan:
Yes, my expectation would also be that slicing from a single big matrix should result in fewer parameters overall than having one weight for each unique shape. There could be some exceptions, e.g. say we have a shape 100x10 and a shape 10x100, the total shape would be 100x100, which is much bigger than 100x10 + 10x100. But I'm not sure if there would be any practical examples of that, since `lora_A` and `lora_B` should generally have somewhat similar shapes for different layers (disregarding higher dimensional weights for now). I was just curious if the proposed implementation is based on some empirical finding (I haven't seen it described in the paper).

Even if this is better for overall parameter count, one potential disadvantage of this approach is, however, that it means we need to use the weights directly:

```python
                sliced_A = vera_A[:, : self.in_features]
                sliced_B = vera_B[: self.out_features, :]

                dropout = self.vera_dropout[active_adapter]
                x = x.to(lambda_d.dtype)
                result = result + lambda_b * F.linear(lambda_d * F.linear(dropout(x), sliced_A), sliced_B)
```

(https://github.com/huggingface/peft/pull/1817/files#diff-681ccefeb9c0abc7bcde2c387ed7be1a5271713483d6871018579c8d0b9a96c9R255-R260)

I.e. we cannot go through the forward call of the adapter layers (something like `lambda_b(lora_B(lambda_d(lora_A(dropout(x)))))`). I know that the current VeRA code also doesn't do that, but I think it could be changed to make it happen.

Why would want to go through the `forward` call? In some circumstances, this is important because the module is wrapped or hooks are being added such that some important functionality is executed when `module.forward` or `module.__call__` are being called. This is something that I came across recently when working with FSDP. My concern is now that if we go with the slicing approach, we could end up in a situation where we cannot possibly make this work. In contrast, if we create one layer per shape, we leave the door open to refactor such that  `forward` is called on these modules.

> Orthogonally to that, we could have an option in the config to have separate A&B for every adapted layer, that would be the equivalent of the ablation from the paper (table 7).

This to me seems to be an inferior approach compared to having one set of weights per shape (though easier to implement), as I think VeRA is all about minimizing the number of trainable parameters as much as possible.

Anyway, this is a all bit hypothetical at the moment and I just wanted to put write down my thoughts and hear your opinion. One step forward would be to try out VeRA in different settings that involve these types of wrapping and hooks like offloading with accelerate or using DeepSpeed/FSDP and check if it works.

dkopi:
Regarding the edge case of two "long and narrow" tensors of different "orientation", e.g. 100x10 and 10x100 - resulting A & B would have shapes (r, 100) & (100, r), not (100x100). So it would be 200\*r vs 220\*r.

>I was just curious if the proposed implementation is based on some empirical finding (I haven't seen it described in the paper).

I didn't compare two implementations, but imo both should be equivalent in outcomes, with negligible differences in memory usage.

>Even if this is better for overall parameter count, one potential disadvantage of this approach is, however, that it means we need to use the weights directly

Actually, my first version of implementation was temporarily swapping .data of tensors, so it could be changed back to that.
Then it would be something like:

```python
                sliced_A = vera_A.data[:, : self.in_features]
                sliced_B = vera_B.data[: self.out_features, :]
                init_A = vera_A.data
                init_B = vera_B.data
                vera_A.data = sliced_A
                vera_B.data = sliced_B

                dropout = self.vera_dropout[active_adapter]
                x = x.to(lambda_d.dtype)
                result = result + lambda_b * F.linear(lambda_d * F.linear(dropout(x), sliced_A), sliced_B)
                vera_A.data = init_A
                vera_B.data = init_B
```
But done on the weights of adapted layers. Still, I'm not sure if it would work with stuff like FSDP.

>This to me seems to be an inferior approach compared to having one set of weights per shape (though easier to implement), as I think VeRA is all about minimizing the number of trainable parameters as much as possible.

True, it would make sense only in combination with the regeneration from a seed. Then we still store only one seed per adapted layer.

The most flexible and "future proof" approach would be to give an option to have basis A&B matrices per:
- model
- adapted layer (ablation showed slight advantage of it, so would be good to have)
- shape of adapted layers

However, it would require the most work and could be confusing for new users of VeRA.

---

_On a side note_ (completely unrelated to this PR), A & B could be potentially quantized as they are not trainable, leading to lower memory usage. I haven't tested it yet tho.

tests/test_vera.py
```diff
@@ -265,13 +265,18 @@ def test_vera_lambda_dont_share_memory(self, mlp_same_prng):
...
-        # It is not possible (currently) to have vera_A and vera_B for different shapes, as they cannot be shared if
-        # their shapes are not identical. lin0 and lin1 have different shapes.
-        config = VeraConfig(target_modules=["lin0", "lin1"], init_weights=False)
-        msg = re.escape(
-            "Multiple target layers with different dimensions were specified. VeRA only supports a single dimension "
-            "size. Expected shape (20, 10), got (20, 20)."
-        )
-        with pytest.raises(ValueError, match=msg):
-            get_peft_model(mlp, config)
+    def test_vera_different_shapes(self, mlp):
```
BenjaminBossan, line 268: On top of adding this rather specific test here, I would suggest to add a new row to the test matrix here:

https://github.com/huggingface/peft/blob/ad8f7cb59ee7ca4b9ca1c9048711038ac36b31b8/tests/test_custom_models.py#L336

That way, a VeRA model with different shapes will be checked on a whole bunch of tests.

docs/source/package_reference/vera.md
```diff
@@ -22,7 +22,6 @@ When saving the adapter parameters, it's possible to eschew storing the low rank
 
 VeRA currently has the following constraints:
 
-- All targeted parameters must have the same shape.
```
BenjaminBossan, line 25: Let's also add a sentence or two about how VeRA now handles different shapes.

src/peft/tuners/vera/layer.py
```diff
@@ -252,9 +252,12 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
                 vera_A = self.vera_A[active_adapter]
                 vera_B = self.vera_B[active_adapter]
 
+                sliced_A = vera_A[:, : self.in_features]
```
BenjaminBossan, line 255: Let's just add a comment on why this is needed.

tests/test_vera.py
```diff
@@ -265,13 +265,18 @@ def test_vera_lambda_dont_share_memory(self, mlp_same_prng):
...
+        config = VeraConfig(target_modules=["lin0", "lin3"], init_weights=False)
+        mlp_different_shapes = get_peft_model(mlp, config)
+
+        vera_A = mlp_different_shapes.vera_A["default"]
+        vera_B = mlp_different_shapes.vera_B["default"]
+
+        # lin0 has the largest output dimension, lin3 has the largest input dimension
+        # vera_A should have the shape of (rank, largest_in), vera_B should have the shape of (largest_out, rank)
+        assert vera_A.shape == (config.r, mlp.lin3.in_features)
+        assert vera_B.shape == (mlp.lin0.out_features, config.r)
```
BenjaminBossan, line 278: Let's add a sanity check that `lin0` and `lin3` do not actually have the same shape, as otherwise, these asserts would pass trivially.

BenjaminBossan (changes requested):
Thanks for your replies. At the end of the day, I think it's okay to go ahead as is. If we do run into the issue that users want to use VeRA with DS or FSDP and it doesn't work, I'm sure we can figure out a way. This PR shouldn't break anything in that regard which currently works.

The implementation is super sweet, nice that only a few lines needed to be changed. I only have a few small comments, please check.

> Actually, my first version of implementation was temporarily swapping .data of tensors, so it could be changed back to that.

That probably would also run into trouble, so let's not do it that way.

> the most flexible and "future proof" approach would be to give an option to have basis A&B matrices per: ...
However, it would require the most work and could be confusing for new users of VeRA.

I agree.

> On a side note (completely unrelated to this PR), A & B could be potentially quantized as they are not trainable, leading to lower memory usage. I haven't tested it yet tho.

Cool idea, if you have some time to test it and find that it works, that would be great. My intuition is, however, that it won't be quite trivial to implement and of course it will introduce more quantization error.

dkopi:
I've addressed the requested changes and fixed merging of the adapter - adding the new test to `test_custom_models.py` revealed that merging also required an update.

BenjaminBossan:
Thanks @dkopi. Could you please run `make style` so that CI will pass?

dkopi:
@BenjaminBossan Done. I just skipped two files modified by "make style" - lycoris_utils.py and router.py, to avoid potential merge conflicts.

dkopi:
@BenjaminBossan One of the tests failed with `requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='huggingface.co', port=443): Read timed out. (read timeout=10)`.

Is it an issue with HF servers? Could we rerun it?

BenjaminBossan:
> Is it an issue with HF servers? Could we rerun it?

Yes, no worries, I'll get a notification and re-run those tests.

BenjaminBossan (approved):
Everything looks fantastic, thanks. Updating the notebook was a nice touch.

PR #1819, opened 2024-06-03
Update peft_integrations.md

rodrigomeireles:
The dependency tree has changed therefore the import in the tutorial has been updated accordingly.

### DIFF
--- a/docs/source/tutorial/peft_integrations.md
+++ b/docs/source/tutorial/peft_integrations.md
@@ -28,7 +28,7 @@ For example, load a base model and then load the [artificialguybr/3DRedmond-V1](
 
 ```py
 import torch
-from diffusers import DiffusionPipeline
+from diffusers.pipelines.pipeline_utils import DiffusionPipeline
 
 pipeline = DiffusionPipeline.from_pretrained(
     "stabilityai/stable-diffusion-xl-base-1.0", torch_dtype=torch.float16


younesbelkada:
cc @sayakpaul

docs/source/tutorial/peft_integrations.md
```diff
@@ -28,7 +28,7 @@ For example, load a base model and then load the [artificialguybr/3DRedmond-V1](
 
 ```py
 import torch
-from diffusers import DiffusionPipeline
+from diffusers.pipelines.pipeline_utils import DiffusionPipeline
```
sayakpaul, line 31: No this is not required. Could you help us with a Colab that reproduces this issue?

rodrigomeireles: Hi, thank you for the attention. Indeed it does not seem necessary. I'm sorry for the oversight. However I get this from Pyright language server when using the original import:

image

This kind of warning makes it impossible to pass the training script through CI. Do you know why that happens?

BenjaminBossan: The reason why pyright complains is probably because it does not understand the lazy import mechanism used in diffusers: https://github.com/huggingface/diffusers/blob/main/src/diffusers/__init__.py

This should probably be solved by telling pyright to ignore this line (probably `# type: ignore`). Anyway, since this is just relevant for your script, I would suggest to close this PR (still thanks for trying to fix this!).

rodrigomeireles: Thank you all for the support.

PR #1821, opened 2024-06-04
ENH / Workflow: Notify on slack about peft + transformers main test results

younesbelkada:
As per title and discussed offline, this will send the results in our daily CI channel 

cc @BenjaminBossan


BenjaminBossan (approved):
Thanks for adding this notification

PR #1822, opened 2024-06-04
FIX CI: Install pytest-reportlog package

BenjaminBossan:
To fix:

> `UsageError: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: --report-log compile_tests.log`


younesbelkada (approved):
Thanks !

PR #1626, opened 2024-04-07
Adding PiSSA as an optional initialization method of LoRA

fxmeng:
In paper "https://arxiv.org/pdf/2404.02948.pdf", we introduce a parameter-efficient fine-tuning (PEFT) method, **P**r**i**ncipal **S**ingular values and **S**ingular vectors **A**daptation (PiSSA), which optimizes a significantly reduced parameter space while achieving or surpassing the performance of full-parameter fine-tuning. 
image

PiSSA is inspired by Intrinsic SAID, which suggests that pre-trained, over-parametrized models inhabit a space of low intrinsic dimension. Consequently, PiSSA represents a matrix $W\in\mathbb{R}^{m\times n}$ within the model by the product of two trainable matrices $A \in \mathbb{R}^{m\times r}$ and $B \in \mathbb{R}^{r\times n}$, where $r \ll \min(m, n)$, plus a residual matrix $W^{res}\in\mathbb{R}^{m\times n}$ for error correction. Singular value decomposition (SVD) is employed to factorize $W$, and the principal singular values and vectors of $W$ are utilized to initialize $A$ and $B$. The residual singular values and vectors initialize the residual matrix $W^{res}$, which keeps frozen during fine-tuning. Notably, PiSSA shares the same architecture with Low-Rank Adaptation (LoRA), which hypothesizes that changes in model parameters $\Delta W$ form a low-rank matrix. However, LoRA approximates $\Delta W$ through the product of two matrices, $A$, initialized with Gaussian noise, and \(B\), initialized with zeros, while PiSSA initializes $A$ and $B$ with principal singular values and singular vectors of the original matrix $W$. Given that the principal singular values and vectors capture the essence of a low-rank matrix, PiSSA can better approximate the outcomes of full-parameter fine-tuning at the beginning by changing the essential parts while freezing the "noisy" parts. In comparison, LoRA freezes the original matrix and updates the "noise". This distinction enables PiSSA to convergence much faster than LoRA and also achieve better performance in the end. On five common benchmarks, PiSSA outperforms LoRA on all of them using exactly the same setups except for a different initialization. On GSM8K, Mistral-7B fine-tuned with PiSSA achieves an accuracy of 72.86\%, outperforming LoRA's 67.7\% by 5.16\%.
image
image

[...]

PR #1828, opened 2024-06-05
Adding support for an optional initialization strategy OLoRA

tokenizer-decode:
In this [paper](https://arxiv.org/abs/2406.01775) we introduced OLoRA, an optional initialization strategy that converges significantly faster and ultimately achieves better performance than the standard LoRA implementation. It can be implemented with a single function. I closely followed other init methods so this implementation should be in PEFT style. If you'd lke this to be an optional method,  I can go ahead and implement examples and tests. 

image

### DIFF
--- a/examples/olora_finetuning/README.md
+++ b/examples/olora_finetuning/README.md
@@ -49,24 +49,27 @@ from peft import PeftModel
 model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m")
 tokenizer = AutoTokenizer.from_pretrained("facebook/opt-350m")
 olora_model = PeftModel.from_pretrained(model, "olora-opt-350m")
-print(olora_model)
 ```
 
 ## OLoRA and LoRA
-Unlike LoRA, OLoRA mutates the original weights. If you want to use multiple adapters simultaneously you can use `path_initial_model_for_weight_conversion` option. First save your model that was initialized with OLoRA **before** performing any training:
+OLoRA differs from LoRA in that it mutates the original weights. To utilize multiple adapters simultaneously, you can leverage the `path_initial_model_for_weight_conversion` option. Below is a simple template illustrating how to convert OLoRA to conventional LoRA:
 ```python
+base_model = AutoModel.from_pretrained("facebook/opt-350m")
+olora_config = LoraConfig(
+    ...
+    init_lora_weights = "olora" # Initialize the model with OLoRA
+)
+olora_model = get_peft_model(base_model, olora_config)
 init_path = <path-to-untrained-olora-model>
-olora_model.save_pretrained(init_path)
+olora_model.save_pretrained(init_path) # Save the model *before* performing any training
+
+# Train the model
+train(olora_model) # Your training loop
+
+#Save the model after training
+olora_model.save_pretrained(output_dir, path_initial_model_for_weight_conversion=init_path) 
 ```
-# After training
-Then you can specify the path of the initialized adapter in the `save_pretrained` method:
-```python
-from transformers import AutoModelForCausalLM
-model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m")
-olora_model = get_peft_model(model, lora_config)
-olora_model.save_pretrained(save_dir, path_initial_model_for_weight_conversion=init_path)
-```
-Thus converting OLoRA to LoRA to use with multiple adapters.
+After completing training, you can save and convert your OLoRA model to a conventional LoRA model by setting `path_initial_model_for_weight_conversion` to `init_path`, that is the path of your untrained OLoRA model. This conversion enables you to use multiple adapters with your LoRA model.
 
 ## Citation
 ```
--- a/examples/olora_finetuning/README.md
+++ b/examples/olora_finetuning/README.md
@@ -1,7 +1,7 @@
 # OLoRA: Orthonormal Low Rank Adaptation of Large Language Models
 
 ## Introduction
-[OLoRA](https://arxiv.org/abs/2406.01775) is a novel approach that leverages orthonormal low rank adaptation through QR decomposition. Unlike the default LoRA implementation, OLoRA decomposes original weights into their Q and R parts, and then uses the first r columns of Q and the first r rows of R to initialize $\mathbf{A}$ and $\mathbf{B}$, respectively. This results in significantly faster convergence, more stable training, and superior performance.
+[OLoRA](https://arxiv.org/abs/2406.01775) is a novel approach that leverages orthonormal low rank adaptation through QR decomposition. Unlike the default LoRA implementation, OLoRA decomposes original weights into their $\mathbf{Q}$ and $\mathbf{R}$ parts, and then uses the first `rank` rows of $\mathbf{R}$ and the first `rank` columns of $\mathbf{Q}$ to initialize $\mathbf{A}$ and $\mathbf{B}$, respectively. This results in significantly faster convergence, more stable training, and superior performance.
 
 ## Quick start
 ```python
@@ -29,6 +29,19 @@ trainer.train()
 peft_model.save_pretrained("olora-opt-350m")
 ```
 
+There is no additional change needed to your standard LoRA procedure, except for specifying `init_lora_weights = "olora"` option in your lora configuration.
+
+Additionally you can refer to olora finetuning script.
+Run the script simply by running:
+```bash
+python3 examples/olora_finetuning/olora_finetuning.py --base_model facebook/opt-350m
+```
+OLoRA also supports quantization. To use 4-bit quantization try:
+```bash
+python3 examples/olora_finetuning/olora_finetuning.py --base_model facebook/opt-350m --quantize
+```
+
+
 ## Use the model
 You can load and use the model as any other 🤗 PEFT model
 ```python
@@ -38,6 +51,22 @@ tokenizer = AutoTokenizer.from_pretrained("facebook/opt-350m")
 olora_model = PeftModel.from_pretrained(model, "olora-opt-350m")
 print(olora_model)
 ```
+
+## OLoRA and LoRA
+Unlike LoRA, OLoRA mutates the original weights. If you want to use multiple adapters simultaneously you can use `path_initial_model_for_weight_conversion` option. First save your model that initialized by using OLoRA before performing any training:
+```python
+olora_model.save_pretrained(init_path)
+```
+
+Then you can specify the path of the initialized adapter in the `save_pretrained` method:
+```python
+from transformers import AutoModelForCausalLM
+model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m")
+olora_model = get_peft_model(model, lora_config)
+olora_model.save_pretrained(save_dir, path_initial_model_for_weight_conversion=init_path)
+```
+Thus converting OLoRA to LoRA to use with multiple adapters.
+
 ## Citation
 ```
 @misc{büyükakyüz2024olora,
--- a/examples/olora_finetuning/README.md
+++ b/examples/olora_finetuning/README.md
@@ -22,7 +22,7 @@ trainer = SFTTrainer(
     model=peft_model,
     train_dataset=dataset,
     dataset_text_field="text",
-    max_seq_length=32,
+    max_seq_length=512,
     tokenizer=tokenizer,
 )
 trainer.train()
@@ -31,7 +31,7 @@ peft_model.save_pretrained("olora-opt-350m")
 
 ## Use the model
 You can load and use the model as any other 🤗 PEFT model
-```
+```python
 from peft import PeftModel
 model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m")
 tokenizer = AutoTokenizer.from_pretrained("facebook/opt-350m")
--- a/examples/olora_finetuning/README.md
+++ b/examples/olora_finetuning/README.md
@@ -1,0 +1,51 @@
+# OLoRA: Orthonormal Low Rank Adaptation of Large Language Models
+
+## Introduction
+[OLoRA](https://arxiv.org/abs/2406.01775) is a novel approach that leverages orthonormal low rank adaptation through QR decomposition. Unlike the default LoRA implementation, OLoRA decomposes original weights into their Q and R parts, and then uses the first r columns of Q and the first r rows of R to initialize $\mathbf{A}$ and $\mathbf{B}$, respectively. This results in significantly faster convergence, more stable training, and superior performance.
+
+## Quick start
+```python
+import torch
+from peft import LoraConfig, get_peft_model
+from transformers import AutoTokenizer, AutoModelForCausalLM
+from trl import SFTTrainer
+from datasets import load_dataset
+
+model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m", torch_dtype=torch.bfloat16, device_map="auto")
+tokenizer = AutoTokenizer.from_pretrained("facebook/opt-350m")
+dataset = load_dataset("imdb", split="train[:1%]")
+lora_config = LoraConfig(
+    init_lora_weights="olora"
+)
+peft_model = get_peft_model(model, lora_config)
+trainer = SFTTrainer(
+    model=peft_model,
+    train_dataset=dataset,
+    dataset_text_field="text",
+    max_seq_length=32,
+    tokenizer=tokenizer,
+)
+trainer.train()
+peft_model.save_pretrained("olora-opt-350m")
+```
+
+## Use the model
+You can load and use the model as any other 🤗 PEFT model
+```
+from peft import PeftModel
+model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m")
+tokenizer = AutoTokenizer.from_pretrained("facebook/opt-350m")
+olora_model = PeftModel.from_pretrained(model, "olora-opt-350m")
+print(olora_model)
+```
+## Citation
+```
+@misc{büyükakyüz2024olora,
+      title={OLoRA: Orthonormal Low-Rank Adaptation of Large Language Models}, 
+      author={Kerim Büyükakyüz},
+      year={2024},
+      eprint={2406.01775},
+      archivePrefix={arXiv},
+      primaryClass={cs.CL}
+}
+```
--- a/examples/olora_finetuning/README.md
+++ b/examples/olora_finetuning/README.md
@@ -53,11 +53,12 @@ print(olora_model)
 ```
 
 ## OLoRA and LoRA
-Unlike LoRA, OLoRA mutates the original weights. If you want to use multiple adapters simultaneously you can use `path_initial_model_for_weight_conversion` option. First save your model that initialized by using OLoRA before performing any training:
+Unlike LoRA, OLoRA mutates the original weights. If you want to use multiple adapters simultaneously you can use `path_initial_model_for_weight_conversion` option. First save your model that was initialized with OLoRA **before** performing any training:
 ```python
+init_path = <path-to-untrained-olora-model>
 olora_model.save_pretrained(init_path)
 ```
-
+# After training
 Then you can specify the path of the initialized adapter in the `save_pretrained` method:
 ```python
 from transformers import AutoModelForCausalLM
--- a/examples/olora_finetuning/olora_finetuning.py
+++ b/examples/olora_finetuning/olora_finetuning.py
@@ -1,4 +1,4 @@
-# Copyright 2023-present the HuggingFace Inc. team.
+# Copyright 2024-present the HuggingFace Inc. team.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -15,7 +15,6 @@
 
 from typing import List
 
-import fire
 import torch
 import transformers
 from datasets import load_dataset
@@ -32,7 +31,6 @@ def train(
     data_path: str = "yahma/alpaca-cleaned",
     output_dir: str = "olora",
     batch_size: int = 16,
-    micro_batch_size: int = 4,
     num_epochs: int = 1,
     learning_rate: float = 3e-4,
     cutoff_len: int = 256,
@@ -47,7 +45,6 @@ def train(
     lora_target_modules: List[str] = None,
     init_lora_weights="olora",
 ):
-    gradient_accumulation_steps = batch_size // micro_batch_size
     model = AutoModelForCausalLM.from_pretrained(
         base_model,
         device_map=device_map,
@@ -111,8 +108,7 @@ def train(
         train_dataset=train_data,
         eval_dataset=val_data,
         args=transformers.TrainingArguments(
-            per_device_train_batch_size=micro_batch_size,
-            gradient_accumulation_steps=gradient_accumulation_steps,
+            per_device_train_batch_size=batch_size,
             warmup_steps=100,
             num_train_epochs=num_epochs,
             learning_rate=learning_rate,
@@ -144,5 +140,45 @@ def generate_prompt(example):
 
 
 if __name__ == "__main__":
-    torch.manual_seed(42)
-    fire.Fire(train)
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--base_model", type=str, default="path/to/model")
+    parser.add_argument("--data_path", type=str, default="yahma/alpaca-cleaned")
+    parser.add_argument("--output_dir", type=str, default="olora")
+    parser.add_argument("--batch_size", type=int, default=16)
+    parser.add_argument("--num_epochs", type=int, default=1)
+    parser.add_argument("--learning_rate", type=float, default=3e-4)
+    parser.add_argument("--cutoff_len", type=int, default=256)
+    parser.add_argument("--val_set_size", type=int, default=16)
+    parser.add_argument("--quantize", action="store_true")
+    parser.add_argument("--eval_step", type=int, default=100)
+    parser.add_argument("--save_step", type=int, default=100)
+    parser.add_argument("--device_map", type=str, default="auto")
+    parser.add_argument("--lora_r", type=int, default=32)
+    parser.add_argument("--lora_alpha", type=int, default=16)
+    parser.add_argument("--lora_dropout", type=float, default=0.05)
+    parser.add_argument("--lora_target_modules", type=str, default=None)
+    parser.add_argument("--init_lora_weights", type=str, default="olora")
+
+    args = parser.parse_args()
+
+    train(
+        base_model=args.base_model,
+        data_path=args.data_path,
+        output_dir=args.output_dir,
+        batch_size=args.batch_size,
+        num_epochs=args.num_epochs,
+        learning_rate=args.learning_rate,
+        cutoff_len=args.cutoff_len,
+        val_set_size=args.val_set_size,
+        quantize=args.quantize,
+        eval_step=args.eval_step,
+        save_step=args.save_step,
+        device_map=args.device_map,
+        lora_r=args.lora_r,
+        lora_alpha=args.lora_alpha,
+        lora_dropout=args.lora_dropout,
+        lora_target_modules=args.lora_target_modules,
+        init_lora_weights=args.init_lora_weights,
+    )
--- a/examples/olora_finetuning/olora_finetuning.py
+++ b/examples/olora_finetuning/olora_finetuning.py
@@ -13,13 +13,13 @@
 # limitations under the License.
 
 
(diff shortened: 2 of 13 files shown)


BenjaminBossan:
Thanks for proposing this new method for initializing the LoRA weights. In general, I think we can for sure add it to PEFT. As you mentioned, having some tests and documentation would be appreciated.

Implementation-wise, I only had a quick look at the paper and code for now. One part could make things a bit more complicated:

> weight_tensor.data -= scale_factor * self.lora_B[adapter_name].weight @ self.lora_A[adapter_name].weight

Here the base weights of the model are mutated, which can have some negative side effects. In fact, this is basically the same problem we also faced when adding PiSSA, so I'll just link to my [comment there](https://github.com/huggingface/peft/pull/1626#issuecomment-2049485948). If you could share your insights on this, that would be great.

Furthermore, as is, I think this would not work with QLoRA, right? But I think not much would be needed to make it work. Do you think this extra functionality could be added?

tokenizer-decode:
Regarding mutation, I wouldn't expect significant side effects if we're meticulous about checking the types of scale factor inputs, weights etc.. And given types are correct, mathematically speaking, translating **A** by QR decomp of **A** should not move **A** far away from its initial point. However, in the worst-case scenario, we can adopt a similar approach to PiSSA.

Regarding quantization, it's been a significant challenge. I'd really like this to be quantization friendly but firstly, Torch QR doesn't support anything lower than 16-bit. We could implement a custom QR decomposition to support lower bit depths, but I'm skeptical about its feasibility and potential performance. As it stands, it's not compatible with QLoRA.

Additionally, based on my experiments, I've found that this doesn't work with Torch.compile. When I enable compilation, I have to disable safe serialization. I'm unsure whether this is a problem with Torch compile or safe serialization, but I don't think it's an issue with OLoRA. 

So I'll start with adding tests, checks for mutation issue, documentation, and we build from there?

BenjaminBossan:
> Regarding mutation, I wouldn't expect significant side effects if we're meticulous about checking the types of scale factor inputs, weights etc.

This is not so much about the LoRA A and B weights, but about the weights of the base model, which are mutated in the cited line. Let's say I want to load to LoRA adapters, one normal one and one trained with OLoRA. As OLoRA changes the base weights, my normal LoRA adapter may suddenly stop working correctly, as it assumes that the base weights are not touched. This is what I was referring to in my comment.

Also, I wonder if we save and load an OLoRA model, do we have to ensure that the random seed is exactly the same so that the base weights are modified exactly identically? If yes, this could be an easy source of errors.

Did you run tests to check how much of a difference it makes if the base weights were not changed? If we could avoid it, it would make things a lot easier.

> Regarding quantization, it's been a significant challenge. I'd really like this to be quantization friendly but firstly, Torch QR doesn't support anything lower than 16-bit. We could implement a custom QR decomposition to support lower bit depths, but I'm skeptical about its feasibility and potential performance. As it stands, it's not compatible with QLoRA.

I was wondering if we could temporarily dequantize the quantized weights, that way, all operations could be performed on floats. It's not ideal but it would be better than not offering this at all. Of course, this can still be problematic when the base weights have to be modified.

> Additionally, based on my experiments, I've found that this doesn't work with Torch.compile.

This is fine, we don't support torch.compile across the board, as [documented here](https://huggingface.co/docs/peft/main/en/developer_guides/torch_compile).

> When I enable compilation, I have to disable safe serialization.

This sounds strange, do you mean in general or just with OLoRA? If in general, please open an issue and show us the error.

> So I'll start with adding tests, checks for mutation issue, documentation, and we build from there?

Exactly, thanks.

examples/olora_finetuning/olora_finetuning.py
```diff
@@ -0,0 +1,130 @@
+from typing import List
```
BenjaminBossan, line 1: This is missing the copyright notice that we have on each file. Also, would you mind adding a README.md to this example with a quick explanation and/or reference and an instruction on how to call the script?

src/peft/peft_model.py
```diff
@@ -215,12 +215,12 @@ def save_pretrained(
             is_main_process (`bool`, *optional*):
                 Whether the process calling this is the main process or not. Will default to `True`. Will not save the
                 checkpoint if not on the main process, which is important for multi device setups (e.g. DDP).
-            convert_pissa_to_lora (`str`):
-                The path to the initialized PiSSA adapter, which is obtained after initializing the model with PiSSA
-                and before performing any training. When `convert_pissa_to_lora` is not None, the difference in PISSA
+            convert_mutated_to_lora (`Literal["olora", "pissa"]`):
```
BenjaminBossan, line 218: Regarding the name, I think we have to find something better. Of course, we cannot keep the old name but at the same time, I'm afraid that the new name could confuse users. I don't have really good ideas, but something like `path_initial_model_for_weight_conversion` could be more fitting, WDYT?

Moreover, the type annotation is wrong, isn't it? The argument is the path to the base model, not the initialization scheme, or `None`, so `Optional[str]` would be correct.

src/peft/peft_model.py
```diff
@@ -240,19 +240,24 @@ def save_pretrained(
                     f" {list(self.peft_config.keys())} - got {selected_adapters}."
                 )
 
-        def save_pissa_as_lora(peft_config, convert_pissa_to_lora, output_state_dict, kwargs):
-            if not str(peft_config.init_lora_weights).startswith("pissa"):
-                warnings.warn("`convert_pissa_to_lora` only works for converting a PiSSA adapter to a LoRA adapter")
-            initial_adapter = os.path.basename(convert_pissa_to_lora)
+        def save_mutated_as_lora(peft_config, convert_mutated_to_lora, output_state_dict, kwargs):
+            if not any(str(peft_config.init_lora_weights).startswith(prefix) for prefix in ["pissa", "olora"]):
```
BenjaminBossan, line 244: There is some inconsistency with regard to how the argument is called. Here it is assumed that it's `"olora"`, but in the example in the docs, it is `"OLoRA"`. How about making it case insensitive, so that both work? I.e. here we should call `.lower()` before comparing.

BenjaminBossan (changes requested):
Thanks for the updates to the PR. I haven't done an in depth review yet, as I think we still have to flesh out some designs.

I agree that if the mechanism for updating the weights can be shared with PiSSA, it should be done so. However, we should not remove the `convert_pissa_to_lora` argument, as this would be a breaking change.

Instead, please leave that argument and add the new one also. Then add a check if `convert_pissa_to_lora is not None`, in which case a deprecation warning should be given with the instruction to use the new argument in the future, then set the new argument to the value of `convert_pissa_to_lora`. That way, existing code continues to work but users are nudged towards using the new argument.

Regarding OLoRA + quantized weights, if it's not so straightforward, just skip it for now. We can always come back to this in a future PR. I would just add an extra error message if the weights are quantized to explicitly say that OLoRA + quantization is not yet supported.

src/peft/peft_model.py
tokenizer-decode: That was my mistake. Everything should be lower case.

src/peft/peft_model.py
tokenizer-decode: I also had really hard time coming up with a name. Just don't know what to call it. And yes type should be fixed.

examples/olora_finetuning/olora_finetuning.py
tokenizer-decode: Sure

tokenizer-decode:
I believe it should be good to review now.

examples/olora_finetuning/README.md
```diff
@@ -0,0 +1,51 @@
...
+from datasets import load_dataset
+
+model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m", torch_dtype=torch.bfloat16, device_map="auto")
+tokenizer = AutoTokenizer.from_pretrained("facebook/opt-350m")
+dataset = load_dataset("imdb", split="train[:1%]")
+lora_config = LoraConfig(
+    init_lora_weights="olora"
+)
+peft_model = get_peft_model(model, lora_config)
+trainer = SFTTrainer(
```
tokenizer-decode, line 21: This is straight from [huggingface](https://huggingface.co/docs/trl/en/sft_trainer#quickstart). Do you want me to add a reference?

examples/olora_finetuning/README.md
```diff
@@ -0,0 +1,51 @@
...
+```python
+from peft import PeftModel
+model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m")
+tokenizer = AutoTokenizer.from_pretrained("facebook/opt-350m")
+olora_model = PeftModel.from_pretrained(model, "olora-opt-350m")
+print(olora_model)
+```
+## Citation
+```
+@misc{büyükakyüz2024olora,
```
tokenizer-decode, line 43: You have emojis so UTF-8 should be ok?

src/peft/peft_model.py
```diff
@@ -239,20 +242,33 @@ def save_pretrained(
...
+                "`convert_pissa_to_lora` is deprecated and will be removed in a future version. "
+                "Use `path_initial_model_for_weight_conversion` instead."
+            )
+            path_initial_model_for_weight_conversion = convert_pissa_to_lora
 
-        def save_pissa_as_lora(peft_config, convert_pissa_to_lora, output_state_dict, kwargs):
-            if not str(peft_config.init_lora_weights).startswith("pissa"):
-                warnings.warn("`convert_pissa_to_lora` only works for converting a PiSSA adapter to a LoRA adapter")
-            initial_adapter = os.path.basename(convert_pissa_to_lora)
+        def save_mutated_as_lora(peft_config, path_initial_model_for_weight_conversion, output_state_dict, kwargs):
```
tokenizer-decode, line 252: I think this could be named better too. But at least its internal and should be ok for now.

src/peft/tuners/lora/layer.py
```diff
@@ -150,6 +153,30 @@ def reset_lora_parameters(self, adapter_name, init_lora_weights):
             nn.init.zeros_(self.lora_embedding_A[adapter_name])
             nn.init.normal_(self.lora_embedding_B[adapter_name])
 
+    def olora_init(self, adapter_name):
+        dtype = self.base_layer.weight.dtype
+        if dtype in [torch.int8, torch.uint8]:
+            warnings.warn("You are using OLoRA with quantized weights. This is still experimental.")
+            base_layer_copy = deepcopy(self.base_layer)  # Make FSDP happy
+            weight_tensor = dequantize_module_weight(base_layer_copy)
```
tokenizer-decode, line 161: Experimental quantization support. I am not familiar with how peft internally dequantizes so I came up with this. Maybe you can help me out on this one. It works ok with `load_in_4bit` and `load_in_8bit` for now.

tests/test_initialization.py
```diff
@@ -300,11 +309,86 @@ def test_lora_pissa_conversion_same_output_after_loading(self, data, tmp_path):
...
+        peft_model = get_peft_model(deepcopy(model), config)
+        peft_model.peft_config["default"].init_lora_weights = True
+        peft_model.save_pretrained(tmp_path / "init-model")
+        peft_model.peft_config["default"].init_lora_weights = "pissa"
+
+        tol = 1e-06
+        peft_model.base_model.linear.lora_B["default"].weight.data *= 2.0
+        output_pissa = peft_model(data)[0]
+
+        # test the warning
```
BenjaminBossan, line 339: Hmm, I don't quite get this test. From the name and the comment, it looks like it's supposed to check the deprecation warning, but I don't see any such check. Instead, the test seems to be basically a copy of the previous test.

examples/olora_finetuning/olora_finetuning.py
```diff
@@ -0,0 +1,140 @@
...
+    return f"""Below is an instruction that describes a task. Write a response that appropriately completes the request.
+            ### Instruction:
+            {example["instruction"]}
+            ### Response:
+            {example["output"]}"""
+
+
+if __name__ == "__main__":
+    torch.manual_seed(42)
+    train()
```
BenjaminBossan, line 140: Not a must, but a nice to have: An option for the user to pass `path_initial_model_for_weight_conversion` via CLI.

BenjaminBossan (changes requested):
Thanks for the updates, the code looks quite good and the PR is on a good path to be merged. I still saw a few issues, please check my comments.

src/peft/tuners/lora/layer.py
```diff
@@ -150,6 +153,30 @@ def reset_lora_parameters(self, adapter_name, init_lora_weights):
             nn.init.zeros_(self.lora_embedding_A[adapter_name])
             nn.init.normal_(self.lora_embedding_B[adapter_name])
 
+    def olora_init(self, adapter_name):
+        dtype = self.base_layer.weight.dtype
+        if dtype in [torch.int8, torch.uint8]:
+            warnings.warn("You are using OLoRA with quantized weights. This is still experimental.")
+            base_layer_copy = deepcopy(self.base_layer)  # Make FSDP happy
```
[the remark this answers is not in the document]
tokenizer-decode: I actually did not, I don't have access to multiple machines at the moment. This is adapted from peft/dora. Ultimately I decided to remove it. Copying also caused problems with uint8. Now it works quite well.

tests/test_initialization.py
tokenizer-decode: Maybe my comment is misleading. I am actually testing that when we specify the deprecated option is the behaviour same with the new option (i.e did we correctly re-assign and update the option).

src/peft/tuners/lora/layer.py
```diff
@@ -150,6 +153,30 @@ def reset_lora_parameters(self, adapter_name, init_lora_weights):
             nn.init.zeros_(self.lora_embedding_A[adapter_name])
             nn.init.normal_(self.lora_embedding_B[adapter_name])
 
+    def olora_init(self, adapter_name):
+        dtype = self.base_layer.weight.dtype
+        if dtype in [torch.int8, torch.uint8]:
+            warnings.warn("You are using OLoRA with quantized weights. This is still experimental.")
```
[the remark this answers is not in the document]
tokenizer-decode: I simply put that to remind that I did not run any benchmarks with quantization yet. We can remove it. its ok to use quantization IMO.

tokenizer-decode:
So okay, I improved the docs and examples. float and int8 is a good place to start to support quantization.

examples/olora_finetuning/README.md
```diff
@@ -0,0 +1,80 @@
...
+```python
+from peft import PeftModel
+model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m")
+tokenizer = AutoTokenizer.from_pretrained("facebook/opt-350m")
+olora_model = PeftModel.from_pretrained(model, "olora-opt-350m")
+print(olora_model)
+```
+
+## OLoRA and LoRA
+Unlike LoRA, OLoRA mutates the original weights. If you want to use multiple adapters simultaneously you can use `path_initial_model_for_weight_conversion` option. First save your model that initialized by using OLoRA before performing any training:
```
BenjaminBossan, line 56:
```suggestion
Unlike LoRA, OLoRA mutates the original weights. If you want to use multiple adapters simultaneously you can use `path_initial_model_for_weight_conversion` option. First save your model that was initialized with OLoRA **before** performing any training:
```

examples/olora_finetuning/README.md
```diff
@@ -0,0 +1,80 @@
...
+```python
+olora_model.save_pretrained(init_path)
+```
+
+Then you can specify the path of the initialized adapter in the `save_pretrained` method:
+```python
+from transformers import AutoModelForCausalLM
+model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m")
+olora_model = get_peft_model(model, lora_config)
```
BenjaminBossan, line 65:
```suggestion
# after training
```

examples/olora_finetuning/README.md
```diff
@@ -0,0 +1,80 @@
...
+model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m")
+tokenizer = AutoTokenizer.from_pretrained("facebook/opt-350m")
+olora_model = PeftModel.from_pretrained(model, "olora-opt-350m")
+print(olora_model)
+```
+
+## OLoRA and LoRA
+Unlike LoRA, OLoRA mutates the original weights. If you want to use multiple adapters simultaneously you can use `path_initial_model_for_weight_conversion` option. First save your model that initialized by using OLoRA before performing any training:
+```python
+olora_model.save_pretrained(init_path)
```
BenjaminBossan, line 58:
```suggestion
init_path = <path-to-untrained-olora-model>
olora_model.save_pretrained(init_path)
```

src/peft/tuners/lora/config.py
```diff
@@ -73,12 +73,12 @@ class LoraConfig(PeftConfig):
...
         modules_to_save (`List[str]`):
             List of modules apart from adapter layers to be set as trainable and saved in the final checkpoint.
-        init_lora_weights (`bool` | `Literal["gaussian", "pissa", "pissa_niter_[number of iters]", "loftq"]`):
+        init_lora_weights (`bool` | `Literal["gaussian", "olora", "pissa", "pissa_niter_[number of iters]", "loftq"]`):
             How to initialize the weights of the adapter layers. Passing True (default) results in the default
             initialization from the reference implementation from Microsoft. Passing 'gaussian' results in Gaussian
             initialization scaled by the LoRA rank for linear and layers. Setting the initialization to False leads to
-            completely random initialization and is discouraged. Pass `'loftq'` to use LoftQ initialization. Passing
-            'pissa' results in the initialization of PiSSA, which converge more rapidly than LoRA and ultimately
+            completely random initialization and is discouraged. Pass `'loftq'` to use LoftQ initialization. Pass `'olora'` to use OLoRA initialization.
```
BenjaminBossan, line 80: Please format to fit the char limit (ruff doesn't automatically adjust strings).

src/peft/tuners/lora/layer.py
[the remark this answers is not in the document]
BenjaminBossan: Yeah, let's remove it, unless we have a strong suspicion that this will change for the user in the past. The more warnings there are, the more users tend to ignore them.

To be very sure that this works, consider adding GPU tests in `test_gpu_examples.py` using OLoRA. You can copy one of the existing tests and adjust the `LoraConfig`.

tests/test_initialization.py
BenjaminBossan: Okay I see. In this case, let's remove this comment. Let's rename the test function to `test_lora_pissa_conversion_same_output_after_loading_with_deprecated_arg` and also add a `# TODO: remove test for deprecated arg in PEFT v0.14.0` comment.

Furthermore, let's add a test below to check the warning itself (which was what I initially thought this test was doing). You can use `with pytest.warns` or the `recwarn` fixture for this. Also add the `TODO` comment there.

src/peft/tuners/lora/layer.py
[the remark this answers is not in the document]
BenjaminBossan: Yeah, let's not add this type of code without testing. If this does not yet work with FSDP, it's fine. We can test it at a later point and then add fixes if it should fail. The deepcopy-issue should be resolved in the next bnb release.

BenjaminBossan (changes requested):
Thanks for the new additions and adjustments. This looks quite good, only a few smaller comments left. Please take a look.

tokenizer-decode:
Okay done. Additional quant and warning tests added. Doc improved. Quant warning fixed. Please take a look at the final commit.

tokenizer-decode:
Also something's just been itching me. We have an option reserved for OLoRA and PiSSA in `save_pretrained` method. For some reason (I'm also not sure why, I only have empirical evidence for this) mutating base weights results in pretty good performance increase compared to its non-mutating counterpart. Since we mutate the base weights we do have to perform some tricks to mitigate the negative effects that you mentioned. I tried to make this approach as uniform as possible, so that in the future if someone comes up with a new init strategy and wants to mutate the base weights as we do, he/she can use our method easily with relatively little hustle.

However, from a strictly design perspective, I think `save_pretrained` is too generic to have an option reserved strictly for something as niche as our fine-tuning strategy. First of all, I think very few people will use multiple adapters, most of them will use a single adapter and call it a day (or that's just my observation), but of course it's good that we're supporting this and we should do so. But there should be a cleaner way to do this without polluting `save_pretrained` method. I just don't know yet. Maybe this option can be carried to `LoraConfig` and since we know the user is using OLoRA, we can automatically do conversion if we're asked to do so. I don't know. Even if we should do this, that's for another PR. Does it make sense to you? Wdyt?

tests/test_gpu_examples.py
```diff
@@ -1671,6 +1671,117 @@ def forward(self, x):
...
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="test requires a GPU")
+class TestOLoRA:
+    r"""
+    Tests for OLoRA to ensure that it reduces the quantization error compared to normal LoRA quantization.
+    """
+
+    # The error factor indicates by how much the quantization error should be decreased when using OLoRA compared to
+    # quantization without OLoRA. Thus 1.03 means that the error should be decreased by 3% at least. This is a very
+    # conservative value to prevent flakiness, in practice most gains are > 1.5
+    error_factor = 1.03
```
BenjaminBossan, line 1683: I checked the errors for all the tests and found that 1.03 is very conservative. Something like 1.3 would also work, i.e. a 10x increase in margin (though not 1.5 as the comment suggests, which probably was just copied from PiSSA?). What did you observe on your machine?

We should probably not go too high here to prevent flakiness, but if you can confirm my observations, how about going to 1.2?

BenjaminBossan (changes requested):
The latest changes look quite good. CI is still failing because of line lengths, could you please run `make style` on the code?

tokenizer-decode:
Weird that I don't get any errors from make style. It did change some irrelevant functions though. Please check the latest commit.

BenjaminBossan:
> I tried to make this approach as uniform as possible, so that in the future if someone comes up with a new init strategy and wants to mutate the base weights as we do, he/she can use our method easily with relatively little hustle.

Yes, that is a great addition.

> I think `save_pretrained` is too generic to have an option reserved strictly for something as niche as our fine-tuning strategy.

It is a bit specific, but I don't think the cost of adding this option there is too high.

> First of all, I think very few people will use multiple adapters, most of them will use a single adapter and call it a day

Probably a single adapter is the most common case, but I would not underestimate the number of users who are using more advanced features. Of course I have no way of really knowing, but based on the issues that are being opened, a significant number of users uses multiple adapters.

> Maybe this option can be carried to `LoraConfig` and since we know the user is using OLoRA, we can automatically do conversion if we're asked to do so.

> This also eliminates manual saving in the first place. This is just a sketch. Main idea is to not mess with `save_pretrained`.
> I don't know. Even if we should do this, that's for another PR. Does it make sense to you? Wdyt?

Thanks for thinking about and proposing an alternative design. If I understand your proposal correctly, what I think could is that it might create the impression that passing this to the config is all it takes to take care of this issue. But once training is over, we would still require the user to call `save_pretrained` and we would need to add logic there to perform the conversion, right? If the user just loads another LoRA adapter after training, it won't work properly.

We could try to call this conversion automatically after training, but this is also problematic. First of all, PEFT does not provide training code, so how would we know when training is over? Second, even if training is over, we don't know if the user may not want to continue training, in which case we should not convert.

Furthermore, if we automated this, it would mean that we need to save a copy of the model when we initialize the PEFT model. This could be quite surprising to users.

What I like about the current approach, even if it's cumbersome, is that is very explicit. Therefore, it is less likely that the user will be surprised by anything that happens.

What I could imagine that we could do is to factor out the conversion into its own function. That way, the user could explicitly call it once they're finished with training. The `save_pretrained` method could use this same function under the hood if the `path_initial_model_for_weight_conversion` is passed. But then we should ideally have a check to ensure that there is no double conversion.

tokenizer-decode:
My bad that I didn't have `doc-builder`. Now it should work I suppose.

Thanks a lot for your insights about `save_pretrained` issue. I might create a new PR in the future and we can talk about improving user-experience there. 

Also you've been super helpful in the process. Thanks.

examples/olora_finetuning/README.md
```diff
@@ -0,0 +1,81 @@
...
+print(olora_model)
+```
+
+## OLoRA and LoRA
+Unlike LoRA, OLoRA mutates the original weights. If you want to use multiple adapters simultaneously you can use `path_initial_model_for_weight_conversion` option. First save your model that was initialized with OLoRA **before** performing any training:
+```python
+init_path = <path-to-untrained-olora-model>
+olora_model.save_pretrained(init_path)
+```
+# After training
```
BenjaminBossan, line 61: You added `# After training` as a new section, but my intent was to add it below where my comment was, above the `olora_model.save_pretrained` call. The reason is to clarify to the user that the order is:

1. Load the base model and apply OLoRA
2. Save initial model with `init_path`
3. Train the model
4. Save the final model with `path_initial_model_for_weight_conversion=init_path`

Let's try to make this super clear. If you have different proposal to achieve this, feel free to do that.

examples/olora_finetuning/olora_finetuning.py
```diff
@@ -0,0 +1,148 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan, line 1: Let's update the year to 2024

examples/olora_finetuning/README.md
```diff
@@ -0,0 +1,81 @@
...
+    max_seq_length=512,
+    tokenizer=tokenizer,
+)
+trainer.train()
+peft_model.save_pretrained("olora-opt-350m")
+```
+
+There is no additional change needed to your standard LoRA procedure, except for specifying `init_lora_weights = "olora"` option in your lora configuration.
+
+Additionally you can refer to olora finetuning script.
```
BenjaminBossan, line 34: Let's mention that users need to install `fire`, which is not a PEFT dependency.

BenjaminBossan (changes requested):
We're almost good to go, I only found a few minor issues, please take a look.

tokenizer-decode:
Nice catch, thanks. Please see my final commit.

BenjaminBossan (commented):
This looks excellent, thanks for your contribution and ensuring that OLoRA fits well into the PEFT framework.

I was about to approve when I saw something I can't quite explain:

During the tests, when we want to save the initial model, we toggle `init_lora_weights` temporarily to `True` and then switch back to `"olora"`:

https://github.com/huggingface/peft/pull/1828/files#diff-df9ecc7077bee932f56e76161ada47693d73acd3ed175a5b9a9158cfe03ec381R372-R374

I was about to comment that this should be added to the example that you provided. But out of curiosity, I removed the toggle from the test and found that it still passes. Now I wonder if those lines are unnecessary or if the test is not quite correct. Do you have an idea?

Btw. the same is true for the PiSSA tests, i.e. we can remove the toggle there and they also pass, so whatever this is is also affecting PiSSA.

tokenizer-decode:
Hmm, what part did you remove? Because removing the toggle i.e `peft_model.peft_config["default"].init_lora_weights = "olora"` fails the test on my machine. If you removed the whole section, I mean this part:
```
        peft_model.peft_config["default"].init_lora_weights = True
        peft_model.save_pretrained(tmp_path / "init-model")
        peft_model.peft_config["default"].init_lora_weights = "olora"
```
That would break the test.

BenjaminBossan:
Sorry, I should have been more precise, I meant this:

```diff
-       peft_model.peft_config["default"].init_lora_weights = True
        peft_model.save_pretrained(tmp_path / "init-model")
-       peft_model.peft_config["default"].init_lora_weights = "olora"
```

tokenizer-decode:
Nice! That still passes because I internally set this to `True` with my last commit. If you remove my last commit from `peft_model` you'd see that the test fails. PiSSA authors needed that because they did not correctly set this field, and we adapted tests from them. Now we don't need those lines. I think let's just keep those lines so that they can see in the future and they may have comments.

BenjaminBossan:
Ah I see, that's a nice addition. Let's remove the toggle from the OLoRA tests then, as they are just adding code noise. As for PiSSA, this should also be removed from the tests and from the docs, but that can be done in a later PR.

BenjaminBossan:
Yes, nothing to worry about, for some reasons we get many timeouts lately. I'll check in on occasion and restart :)

tokenizer-decode:
Tests look good. Anything left on my end? @BenjaminBossan

BenjaminBossan (approved):
Thanks for the ping, I was distracted by other work and forgot to check back on this PR.

Everything looks super good. Thanks for adding the OLoRA initialization method and refactoring the existing code to make similar additions easier in the future. Also good discussions on different designs.

tokenizer-decode:
Thanks a lot for helping me with the process. Great work.

PR #1830, opened 2024-06-05
Workflow / Bnb: Add a mechanism to inform us if the import fails

younesbelkada:
As per title and to avoid our CI silently passing

cc @Titus-von-Koeller @BenjaminBossan

### DIFF
--- a/.github/workflows/nightly-bnb.yml
+++ b/.github/workflows/nightly-bnb.yml
@@ -46,6 +46,22 @@ jobs:
             git checkout tags/v$transformers_version
             cd .. 
           fi
+
+      - name: Test bnb import
+        id: import
+        if: always()
+        run: |
+          python3 -c "import bitsandbytes as bnb"
+
+      - name: Post to Slack
+        if: always()
+        uses: huggingface/hf-workflows/.github/actions/post-slack@main
+        with:
+          slack_channel: ${{ env.BNB_SLACK_CHANNEL_ID }}
+          title: 🤗 Results of bitsandbytes import
+          status: ${{ steps.import.status }}
+          slack_token: ${{ secrets.SLACK_CIFEEDBACK_BOT_TOKEN }}
+          
       - name: Run examples on single GPU
         if: always()
         run: |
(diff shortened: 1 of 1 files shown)


BenjaminBossan (commented):
Just for my understanding, this is basically an extra step that checks if bnb import works and reports the results to slack (I wonder if checking `python -m bitsandbytes` would also make sense).

This does not address the issue that the [failing nightly bnb runs](https://github.com/huggingface/peft/actions/runs/9377120709) are not correctly reported on slack, right? Could we also use the `post-slack` workflow for those to report them correctly?

younesbelkada:
Makes sense, added them as suggested ! LMK wdyt !

BenjaminBossan (approved):
Thanks for working on this. LGTM.

Titus-von-Koeller:
Thanks @younesbelkada for working on this and @BenjaminBossan for raising this and following up!

Everything LGTM as well, but agreed that we should definitely add `python -m bitsandbytes` as a check. I did a quick text search, but didn't find it yet, sorry if I missed it somehow.

.github/workflows/nightly-bnb.yml
```diff
@@ -46,29 +46,85 @@ jobs:
             git checkout tags/v$transformers_version
             cd .. 
           fi
+
+      - name: Test bnb import
+        id: import
+        if: always()
+        run: |
+          python3 -c "import bitsandbytes as bnb"
```
younesbelkada, line 54:
```suggestion
          python -m bitsandbytes
```

younesbelkada:
The issue with `python -m bitsandbytes` is that this will not explictly throw an error if the setup is broken, it will simply display a message saying that bnb is broken, on the other hand importing bnb on the command line will fail and we would be able to catch it with `jobs.import.status`

younesbelkada:
LMK what do you think and I will merge if you think what I said makes sense

BenjaminBossan:
> The issue with `python -m bitsandbytes` is that this will not explictly throw an error if the setup is broken, it will simply display a message saying that bnb is broken, on the other hand importing bnb on the command line will fail and we would be able to catch it with `jobs.import.status`

Can we not have both, first `python -m bitsandbytes` and then importing? That way, we have a bit more information in the logs if something breaks.

.github/workflows/nightly-bnb.yml
younesbelkada, line 54:
```suggestion
          python3 -c "import bitsandbytes as bnb"
          python3 -m bitsandbytes
```

younesbelkada:
Done !

BenjaminBossan:
Thanks. Just a small note: Would it make sense to call `python3 -m bitsandbytes` first, in case the import fails the next line is not executed, or is it?

.github/workflows/nightly-bnb.yml
```diff
@@ -46,29 +46,86 @@ jobs:
...
+          python3 -m bitsandbytes
```
younesbelkada, line 55:
```suggestion
          python3 -m bitsandbytes
          python3 -c "import bitsandbytes as bnb"
```

younesbelkada:
makes sense!

BenjaminBossan:
From my point of view, this can be merged, not sure if we want to wait for another look by Titus.

younesbelkada:
just forgot to merge! :D

PR #1832, opened 2024-06-06
FIX Failing Llama tests due to new kv cache

BenjaminBossan:
Requires to install transformers from source (so normal CI won't pick this up).

To replicate, run these tests:

```
pytest tests/test_multitask_prompt_tuning.py
pytest tests/test_decoder_models.py -k
"test_trl_internal_testing_tiny_random_LlamaForCausalLM and (prefix or prompt) and (generate or disable)"
```

This should result in 17 failing tests.

The issue is that `past_key_values` can now be an instance of `DynamicCache`. Therefore, just indexing into it won't work anymore. The solution is to check the type and if it's not a tuple, use the methods on the cache object instead.


younesbelkada (approved):
Thanks a lot for fixing ! 
Importing `Cache` class from `transformers.cache_utils` might make our code bloated since that object exists only on newer transformers versions. This solution looks great

PR #1842, opened 2024-06-10
Workflow: Fix broken messages

younesbelkada:
We should use `secrets` instead of `envs` otherwise we get empty strings

cc @BenjaminBossan


BenjaminBossan (approved):
Thanks for figuring this out!

PR #1845, opened 2024-06-11
CI Activate env to prevent bnb import error

BenjaminBossan:
All bitsandbytes nightly CI runs are currently [failing](https://github.com/huggingface/peft/actions/runs/9458512932/job/26054124989) with:

> Run python3 -m bitsandbytes
> /opt/conda/bin/python3: No module named bitsandbytes

This fix should hopefully solve this, but it's untested.

### DIFF
--- a/.github/workflows/nightly-bnb.yml
+++ b/.github/workflows/nightly-bnb.yml
@@ -51,6 +51,7 @@ jobs:
         id: import
         if: always()
         run: |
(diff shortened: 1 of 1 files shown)


younesbelkada (commented):
Thanks ! Can you also add it in the line 174 as well ?

younesbelkada (approved):
Thanks !

PR #1851, opened 2024-06-12
MNT Upgrade ruff version to ~0.4.8

BenjaminBossan:
We currently use ruff v0.2.2, which is quite far behind the latest version. This has the disadvantage that new contributors will often install the latest version of ruff and then get CI errors, even though they ran `make style`.

With this update, the discrepancy should go away. But if we merge this, all contributors may have to update their PRs. This is annoying but I think we need to do this sooner or later.

Here is the full list of changes:

- bump ruff version to ~0.4.8
- update the ruff commands in Makefile (ruff foo/ -> ruff check foo/)
- update coding style of two files that changed with the new ruff version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,13 +6,13 @@ check_dirs := src tests examples docs scripts docker
 
 # this target runs checks on all files
 quality:
-	ruff $(check_dirs)
+	ruff check $(check_dirs)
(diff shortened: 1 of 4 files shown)


younesbelkada (approved):
Thanks for updating `ruff` version !

PR #1852, opened 2024-06-12
FIX: Adalora ranknum loaded on wrong device

BenjaminBossan:
Locally, multiple AdaLoRA-related tests are failing. We did not catch this in the nightly run because the tests were missing the necessary pytest marker.

The issue is related to the change in #1742, which made it possible to load different adapters on different devices. Although that PR itself was sound, the problem is that for AdaLoRA, one of its parameters, `ranknum`, was not listed in the `other_param_names` and was thus not moved to the correct device. This oversight is now fixed and the GPU tests are now passing locally for me.

Note that this bug only occurred in the tests when loading a 2nd AdaLoRA adapter. At first I was confused because I thought that this is not supported (there is a check for that), but actually it's supported as long as the additional adapters are not trainable, so the check is fine.

This PR also adds the missing pytest marker to the test class that was missing it, so that these errors should be caught by our nightly CI in the future.


younesbelkada (approved):
Thanks a lot for fixing !

PR #1854, opened 2024-06-13
Workflow / FIX: Fix red status on our CI

younesbelkada:
# What does this PR do?

According to: https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context when manipulating steps statuses, one should always use `steps.<step_id>.outputs` to retrieve the status. See also: https://stackoverflow.com/a/58003436 

cc @BenjaminBossan


BenjaminBossan (approved):
Interesting, let's hope that this works.

younesbelkada:
Failing CI is unrelated, merging !

PR #1856, opened 2024-06-13
DOC Move helpers section to dev developer guide

BenjaminBossan:
It was in the "Adapters" section, which doesn't really fit.


younesbelkada (approved):
Thanks !

PR #1857, opened 2024-06-13
ephemeral GPU offload support

kallewoof:
Adds the concept of ephemeral GPU offloading, i.e. where data in compute intense operations is copied onto the GPU before the operation is performed, after which the result is put back on CPU memory.

This PR adds support in the dora initialization code, but the approach can be applied in a number of places: when the size of the data compared to the time to perform the operation on CPU memory is heavily time dominant, using ephemeral transfers has a fairly small VRAM overhead (depending on the size of the model/adapter) with orders of magnitude speed-up in certain operations.

For example, a Llama3-8B DoRA adapter with r=64 would put an overhead of 2 x (64 x 4096 x 2 + 4096 x 4096) bytes (assuming fp16), i.e. 33 MB or so. A Llama3-70B adapter with r=32 would have 2 x (32 x 8192 x 2 + 8192 x 8192) bytes =130 MB.

By making use of ephemeral GPU offloading, more efficient juggling of data between GPU and CPU may become possible, i.e. where instead of always loading as much as we can onto the GPU and then endure the CPU slowness for whatever happens to not fit in there, we intentionally leave a (modest) chunk of VRAM for optimizations like these, and the end result is a much (MUCH) faster experience.

See `examples/ephemeral_gpu_offload/load_with_dora.py` for an example script demonstrating this feature. Example output using the defaults of that script are:
```
Example outputs:
$ python load_with_dora.py
-------------------- Loading model --------------------
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:04<00:00,  1.03s/it]
-------------------- Loading PeftModel --------------------
-------------------- Done --------------------
Model loading time: 4.83s
PeftModel loading time: 28.14s
Use ephemeral GPU offloading: False

(Note: if this was the first time you ran the script, or if your cache was cleared, the times shown above are invalid, due to the time taken to download the model and DoRA files. Just re-run the script in this case.)

$ python load_with_dora.py --ephemeral_gpu_offload
-------------------- Loading model --------------------
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:03<00:00,  1.11it/s]
-------------------- Loading PeftModel --------------------
-------------------- Done --------------------
Model loading time: 4.28s
PeftModel loading time: 16.59s
Use ephemeral GPU offloading: True

(Note: if this was the first time you ran the script, or if your cache was cleared, the times shown above are invalid, due to the time taken to download the model and DoRA files. Just re-run the script in this case.)
```
In this case, using ephemeral GPU offloading, we finished in 2/3rd of the time taken doing it purely on CPU.

I verified that the resulting merged models were identical using `cmp` on the model-000*.safetensors files for the variants using and not using ephemeral transfers.

### DIFF
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -122,12 +122,12 @@ from peft import LoraConfig
 config = LoraConfig(use_dora=True, ...)
 ```
 
-If parts of the model or the DoRA adapter are offloaded to CPU you can get a significant speedup at the cost of some temporary (ephemeral) VRAM overhead by using `ephemeral_transfers=True` in `config.runtime`.
+If parts of the model or the DoRA adapter are offloaded to CPU you can get a significant speedup at the cost of some temporary (ephemeral) VRAM overhead by using `ephemeral_transfers=True` in `config.runtime_config`.
 
 ```py
-from peft import LoraConfig, LoraConfigRuntime
+from peft import LoraConfig, LoraRuntimeConfig
 
-config = LoraConfig(use_dora=True, runtime=LoraConfigRuntime(ephemeral_transfers=True), ...)
+config = LoraConfig(use_dora=True, runtime_config=LoraRuntimeConfig(ephemeral_transfers=True), ...)
 ```
 
 A `PeftModel` with a DoRA adapter can also be loaded with `ephemeral_transfers=True` flag using the `from_pretrained` method as well as the `load_adapter` method.
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -122,20 +122,20 @@ from peft import LoraConfig
 config = LoraConfig(use_dora=True, ...)
 ```
 
-If parts of the model or the DoRA adapter are offloaded to CPU you can get a significant speedup at the cost of some temporary (ephemeral) VRAM overhead by using `ephemeral_transfers=True` in `config.runtime_config`.
+If parts of the model or the DoRA adapter are offloaded to CPU you can get a significant speedup at the cost of some temporary (ephemeral) VRAM overhead by using `ephemeral_gpu_offload=True` in `config.runtime_config`.
 
 ```py
 from peft import LoraConfig, LoraRuntimeConfig
 
-config = LoraConfig(use_dora=True, runtime_config=LoraRuntimeConfig(ephemeral_transfers=True), ...)
+config = LoraConfig(use_dora=True, runtime_config=LoraRuntimeConfig(ephemeral_gpu_offload=True), ...)
 ```
 
-A `PeftModel` with a DoRA adapter can also be loaded with `ephemeral_transfers=True` flag using the `from_pretrained` method as well as the `load_adapter` method.
+A `PeftModel` with a DoRA adapter can also be loaded with `ephemeral_gpu_offload=True` flag using the `from_pretrained` method as well as the `load_adapter` method.
 
 ```py
 from peft import PeftModel
 
-model = PeftModel.from_pretrained(base_model, peft_model_id, ephemeral_transfers=True)
+model = PeftModel.from_pretrained(base_model, peft_model_id, ephemeral_gpu_offload=True)
 ```
 
 #### Caveats
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -122,6 +122,22 @@ from peft import LoraConfig
 config = LoraConfig(use_dora=True, ...)
 ```
 
+If parts of the model or the DoRA adapter are offloaded to CPU you can get a significant speedup at the cost of some temporary (ephemeral) VRAM overhead by using `ephemeral_transfers=True` in `config.runtime`.
+
+```py
+from peft import LoraConfig, LoraConfigRuntime
+
+config = LoraConfig(use_dora=True, runtime=LoraConfigRuntime(ephemeral_transfers=True), ...)
+```
+
+A `PeftModel` with a DoRA adapter can also be loaded with `ephemeral_transfers=True` flag using the `from_pretrained` method as well as the `load_adapter` method.
+
+```py
+from peft import PeftModel
+
+model = PeftModel.from_pretrained(base_model, peft_model_id, ephemeral_transfers=True)
+```
+
 #### Caveats
 
 - DoRA only supports linear and Conv2d layers at the momement.
--- a/examples/ephemeral_gpu_offloading/load_with_dora.py
+++ b/examples/ephemeral_gpu_offloading/load_with_dora.py
@@ -17,10 +17,10 @@ Example script demonstrating the time difference loading a model with a DoRA usi
 
 Example outputs:
 $ python load_with_dora.py
--------------------- Loading model --------------------
+--- Loading model ---
 Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:04<00:00,  1.03s/it]
--------------------- Loading PeftModel --------------------
--------------------- Done --------------------
+--- Loading PeftModel ---
+--- Done ---
 Model loading time: 4.83s
 PeftModel loading time: 28.14s
 Use ephemeral GPU offloading: False
@@ -28,10 +28,10 @@ Use ephemeral GPU offloading: False
 (Note: if this was the first time you ran the script, or if your cache was cleared, the times shown above are invalid, due to the time taken to download the model and DoRA files. Just re-run the script in this case.)
 
 $ python load_with_dora.py --ephemeral_gpu_offload
--------------------- Loading model --------------------
+--- Loading model ---
 Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:03<00:00,  1.11it/s]
--------------------- Loading PeftModel --------------------
--------------------- Done --------------------
+--- Loading PeftModel ---
+--- Done ---
 Model loading time: 4.28s
 PeftModel loading time: 16.59s
 Use ephemeral GPU offloading: True
--- a/examples/ephemeral_gpu_offloading/load_with_dora.py
+++ b/examples/ephemeral_gpu_offloading/load_with_dora.py
@@ -4,6 +4,7 @@
 import argparse
 import time
 
+from huggingface_hub import snapshot_download
 from transformers import AutoModelForCausalLM
 
 from peft import PeftModel
@@ -30,6 +31,18 @@ def main():
         "device_map": {"": "cpu"},
     }
 
+    # Predownload
+    try:
+        snapshot_download(repo_id=args.model)
+    except Exception as e:
+        print(f"Failed to download model: {e}")
+        # We continue anyway as this might be e.g. a local directory or something
+    try:
+        snapshot_download(repo_id=args.dora)
+    except Exception as e:
+        print(f"Failed to download DoRA: {e}")
+        # We continue anyway as this might be e.g. a local directory or something
+
     start = time.perf_counter()
     print("-" * 20 + " Loading model " + "-" * 20)
     model = AutoModelForCausalLM.from_pretrained(args.model)
@@ -42,9 +55,6 @@ def main():
     print(f"Model loading time: {model_time:.2f}s")
     print(f"PeftModel loading time: {peft_model_time:.2f}s")
     print(f"Use ephemeral GPU offloading: {args.ephemeral_gpu_offload}")
-    print(
-        "\n(Note: if this was the first time you ran the script, or if your cache was cleared, the times shown above are invalid, due to the time taken to download the model and DoRA files. Just re-run the script in this case.)\n"
-    )
 
     if args.merge_model:
         merged_model = peft_model.merge_and_unload(progressbar=True)
--- a/examples/ephemeral_gpu_offloading/load_with_dora.py
+++ b/examples/ephemeral_gpu_offloading/load_with_dora.py
@@ -1,0 +1,80 @@
+# Example script demonstrating the time difference loading a model with a DoRA using ephemeral GPU offloading
+# vs doing it purely on the CPU.
+
+import argparse
+import time
+
+from transformers import AutoModelForCausalLM
+
+from peft import PeftModel
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Load a model with DoRA using ephemeral GPU offloading")
+    parser.add_argument("--model", type=str, default="NousResearch/Hermes-2-Pro-Mistral-7B", help="Model to load")
+    parser.add_argument(
+        "--dora",
+        type=str,
+        default="anamikac2708/Mistral-7B-DORA-finetuned-investopedia-Lora-Adapters",
+        help="DoRA to use",
+    )
+    parser.add_argument("--ephemeral_gpu_offload", action="store_true", help="Use ephemeral GPU offloading")
+    parser.add_argument(
+        "--merge_model", action="store_true", help="Merge the model with the DoRA model and save to disk"
+    )
+    args = parser.parse_args()
+
+    peft_model_kwargs = {
+        "ephemeral_gpu_offload": args.ephemeral_gpu_offload,
+        "max_memory": {"cpu": "256GiB"},
+        "device_map": {"": "cpu"},
+    }
+
+    start = time.perf_counter()
+    print("-" * 20 + " Loading model " + "-" * 20)
+    model = AutoModelForCausalLM.from_pretrained(args.model)
+    model_time = time.perf_counter() - start
+    print("-" * 20 + " Loading PeftModel " + "-" * 20)
+    peft_model = PeftModel.from_pretrained(model, args.dora, **peft_model_kwargs)
+    print("-" * 20 + " Done " + "-" * 20)
+    peft_model_time = time.perf_counter() - start
+
+    print(f"Model loading time: {model_time:.2f}s")
+    print(f"PeftModel loading time: {peft_model_time:.2f}s")
+    print(f"Use ephemeral GPU offloading: {args.ephemeral_gpu_offload}")
+    print(
+        "\n(Note: if this was the first time you ran the script, or if your cache was cleared, the times shown above are invalid, due to the time taken to download the model and DoRA files. Just re-run the script in this case.)\n"
+    )
+
+    if args.merge_model:
+        merged_model = peft_model.merge_and_unload(progressbar=True)
+        merged_model.save_pretrained("merged_model")
+
+
+if __name__ == "__main__":
+    main()
+
+"""
+Example outputs:
+$ python load_with_dora.py
+-------------------- Loading model --------------------
+Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:04<00:00,  1.03s/it]
+-------------------- Loading PeftModel --------------------
+-------------------- Done --------------------
+Model loading time: 4.83s
+PeftModel loading time: 28.14s
+Use ephemeral GPU offloading: False
+
+(Note: if this was the first time you ran the script, or if your cache was cleared, the times shown above are invalid, due to the time taken to download the model and DoRA files. Just re-run the script in this case.)
+
+$ python load_with_dora.py --ephemeral_gpu_offload
+-------------------- Loading model --------------------
+Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:03<00:00,  1.11it/s]
+-------------------- Loading PeftModel --------------------
+-------------------- Done --------------------
+Model loading time: 4.28s
+PeftModel loading time: 16.59s
+Use ephemeral GPU offloading: True
+
+(Note: if this was the first time you ran the script, or if your cache was cleared, the times shown above are invalid, due to the time taken to download the model and DoRA files. Just re-run the script in this case.)
+"""
--- a/examples/ephemeral_gpu_offloading/load_with_dora.py
+++ b/examples/ephemeral_gpu_offloading/load_with_dora.py
@@ -1,70 +1,20 @@
-# Example script demonstrating the time difference loading a model with a DoRA using ephemeral GPU offloading
-# vs doing it purely on the CPU.
-
-import argparse
-import time
-
-from huggingface_hub import snapshot_download
-from transformers import AutoModelForCausalLM
-
-from peft import PeftModel
-
-
-def main():
-    parser = argparse.ArgumentParser(description="Load a model with DoRA using ephemeral GPU offloading")
-    parser.add_argument("--model", type=str, default="NousResearch/Hermes-2-Pro-Mistral-7B", help="Model to load")
-    parser.add_argument(
-        "--dora",
-        type=str,
-        default="anamikac2708/Mistral-7B-DORA-finetuned-investopedia-Lora-Adapters",
-        help="DoRA to use",
-    )
-    parser.add_argument("--ephemeral_gpu_offload", action="store_true", help="Use ephemeral GPU offloading")
-    parser.add_argument(
-        "--merge_model", action="store_true", help="Merge the model with the DoRA model and save to disk"
-    )
-    args = parser.parse_args()
-
-    peft_model_kwargs = {
-        "ephemeral_gpu_offload": args.ephemeral_gpu_offload,
-        "max_memory": {"cpu": "256GiB"},
-        "device_map": {"": "cpu"},
-    }
-
-    # Predownload
-    try:
-        snapshot_download(repo_id=args.model)
-    except Exception as e:
-        print(f"Failed to download model: {e}")
-        # We continue anyway as this might be e.g. a local directory or something
-    try:
-        snapshot_download(repo_id=args.dora)
-    except Exception as e:
-        print(f"Failed to download DoRA: {e}")
-        # We continue anyway as this might be e.g. a local directory or something
-
-    start = time.perf_counter()
-    print("-" * 20 + " Loading model " + "-" * 20)
-    model = AutoModelForCausalLM.from_pretrained(args.model)
-    model_time = time.perf_counter() - start
-    print("-" * 20 + " Loading PeftModel " + "-" * 20)
-    peft_model = PeftModel.from_pretrained(model, args.dora, **peft_model_kwargs)
-    print("-" * 20 + " Done " + "-" * 20)
-    peft_model_time = time.perf_counter() - start
-
-    print(f"Model loading time: {model_time:.2f}s")
-    print(f"PeftModel loading time: {peft_model_time:.2f}s")
-    print(f"Use ephemeral GPU offloading: {args.ephemeral_gpu_offload}")
-
-    if args.merge_model:
-        merged_model = peft_model.merge_and_unload(progressbar=True)
-        merged_model.save_pretrained("merged_model")
-
-
-if __name__ == "__main__":
-    main()
+# Copyright 2024-present the HuggingFace Inc. team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
 """
+Example script demonstrating the time difference loading a model with a DoRA using ephemeral GPU offloading vs doing it purely on the CPU.
+
 Example outputs:
 $ python load_with_dora.py
 -------------------- Loading model --------------------
@@ -88,3 +38,66 @@ Use ephemeral GPU offloading: True
 
 (Note: if this was the first time you ran the script, or if your cache was cleared, the times shown above are invalid, due to the time taken to download the model and DoRA files. Just re-run the script in this case.)
 """
+
+import argparse
+import time
+
+from huggingface_hub import snapshot_download
+from transformers import AutoModelForCausalLM
+
+from peft import PeftModel
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Load a model with DoRA using ephemeral GPU offloading")
+    parser.add_argument("--model", type=str, default="NousResearch/Hermes-2-Pro-Mistral-7B", help="Model to load")
+    parser.add_argument(
+        "--dora",
+        type=str,
+        default="peft-internal-testing/DoRA-Hermes-2-Pro-Mistral-7B",
+        help="DoRA to use",
+    )
+    parser.add_argument("--ephemeral_gpu_offload", action="store_true", help="Use ephemeral GPU offloading")
+    parser.add_argument(
+        "--merge_model_path", type="str", help="Merge the model with the DoRA model and save to the given path"
+    )
+    args = parser.parse_args()
+
+    peft_model_kwargs = {
+        "ephemeral_gpu_offload": args.ephemeral_gpu_offload,
+        "max_memory": {"cpu": "256GiB"},
+        "device_map": {"": "cpu"},
+    }
+
+    # Predownload
+    try:
+        snapshot_download(repo_id=args.model)
+    except Exception as e:
+        print(f"Failed to download model: {e}")
+        # We continue anyway as this might be e.g. a local directory or something
+    try:
+        snapshot_download(repo_id=args.dora)
+    except Exception as e:
+        print(f"Failed to download DoRA: {e}")
+        # We continue anyway as this might be e.g. a local directory or something
+
+    start = time.perf_counter()
+    print("--- Loading model ---")
+    model = AutoModelForCausalLM.from_pretrained(args.model)
+    model_time = time.perf_counter() - start
+    print("--- Loading PeftModel ---")
+    peft_model = PeftModel.from_pretrained(model, args.dora, **peft_model_kwargs)
+    print("--- Done ---")
+    peft_model_time = time.perf_counter() - start
+
+    print(f"Model loading time: {model_time:.2f}s")
+    print(f"PeftModel loading time: {peft_model_time:.2f}s")
+    print(f"Use ephemeral GPU offloading: {args.ephemeral_gpu_offload}")
+
+    if args.merge_model_path is not None:
+        merged_model = peft_model.merge_and_unload(progressbar=True)
+        merged_model.save_pretrained(args.merge_model_path)
+
+
+if __name__ == "__main__":
+    main()
--- a/src/peft/config.py
+++ b/src/peft/config.py
@@ -14,6 +14,7 @@
 import inspect
 import json
 import os
+import warnings
 from dataclasses import asdict, dataclass, field
 from typing import Dict, Optional, Union
 
@@ -161,6 +162,13 @@ class PeftConfigMixin(PushToHubMixin):
         """
         with open(path_json_file) as file:
             json_object = json.load(file)
+
+        # Sanity check that config does not contain a runtime_config
+        if "runtime_config" in json_object:
+            warnings.warn(
+                "The configuration file contains a `runtime_config` key. This is ignored. Runtime configurations are only valid at runtime."
+            )
+            del json_object["runtime_config"]
 
         return json_object
 
--- a/src/peft/config.py
+++ b/src/peft/config.py
@@ -90,7 +90,7 @@ class PeftConfigMixin(PushToHubMixin):
         os.makedirs(save_directory, exist_ok=True)
         auto_mapping_dict = kwargs.pop("auto_mapping_dict", None)
 
-        output_dict = asdict(self)
+        output_dict = self.to_dict()
         # converting set type to list
         for key, value in output_dict.items():
             if isinstance(value, set):
@@ -101,10 +101,6 @@ class PeftConfigMixin(PushToHubMixin):
         # Add auto mapping details for custom models.
         if auto_mapping_dict is not None:
             output_dict["auto_mapping"] = auto_mapping_dict
-
-        # Drop runtime key
-        if "runtime" in output_dict:
-            output_dict.pop("runtime")
 
         # save it
         with open(output_path, "w") as writer:
--- a/src/peft/config.py
+++ b/src/peft/config.py
@@ -24,6 +24,30 @@ from .utils import CONFIG_NAME, PeftType, TaskType
 
 
 @dataclass
+class PeftConfigRuntime:
+    """
+    This is the sub-configuration class to store the runtime configurations for the model.
+
+    Args:
+        ephemeral_transfers (`bool`): Whether to use ephemeral transfers for models partially kept in CPU memory.
+    """
+    ephemeral_transfers: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Whether to use ephemeral transfers for models partially kept in CPU memory. Ephemeral transfers result in"
+                "the data involved in intense operations being momentarily copied over to the GPU, and the results copied"
+                "back to CPU. There is a momentary VRAM overhead, but operations are generally orders of magnitude faster"
+                "compared to performing them on the CPU. This is useful when parts of the model and/or components (such"
+                "as adapters) are kept in CPU memory until they are needed. Rather than perform expensive operations on"
+                "small data, the data is transferred to the GPU on-demand, the operation(s) performed, and the results"
+                "moved back to CPU memory. This brings a slight momentary VRAM overhead but gives orders of magnitude"
+                "speedup in certain cases."
+            )
+        },
+    )
+
+@dataclass
 class PeftConfigMixin(PushToHubMixin):
     r"""
     This is the base configuration class for PEFT adapter models. It contains all the methods that are common to all
@@ -74,6 +98,10 @@ class PeftConfigMixin(PushToHubMixin):
         # Add auto mapping details for custom models.
         if auto_mapping_dict is not None:
             output_dict["auto_mapping"] = auto_mapping_dict
+
+        # Drop runtime key
+        if "runtime" in output_dict:
+            output_dict.pop("runtime")
 
         # save it
         with open(output_path, "w") as writer:
@@ -236,6 +264,7 @@ class PeftConfig(PeftConfigMixin):
     peft_type: Optional[Union[str, PeftType]] = field(default=None, metadata={"help": "Peft type"})
     task_type: Optional[Union[str, TaskType]] = field(default=None, metadata={"help": "Task type"})
     inference_mode: bool = field(default=False, metadata={"help": "Whether to use inference mode"})
+    runtime: PeftConfigRuntime = field(default_factory=PeftConfigRuntime, metadata={"help": "Runtime configurations"})
 
 
 @dataclass
--- a/src/peft/config.py
+++ b/src/peft/config.py
@@ -67,8 +67,11 @@ class PeftConfigMixin(PushToHubMixin):
     def to_dict(self) -> Dict:
         r"""
         Returns the configuration for your adapter model as a dictionary.
+        Does not include the runtime configuration.
         """
-        return asdict(self)
+        rv = asdict(self)
+        rv.pop('runtime')
+        return rv
 
     def save_pretrained(self, save_directory: str, **kwargs) -> None:
         r"""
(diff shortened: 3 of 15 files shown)


BenjaminBossan:
Thanks for creating this draft PR. 

> when the size of the data compared to the time to perform the operation on CPU memory is heavily time dominant, using ephemeral transfers has a fairly small VRAM overhead (depending on the size of the model/adapter) with orders of magnitude speed-up in certain operations.

I have to admit that I didn't fully understand this yet. Could you please explain this in more details or give a reference? 

From looking at the code, my understanding is that if `lora_A` or `lora_B` is on CPU, it is ensured that the DoRA weight norm is also kept placed on CPU. Where is this part then happening? Is it happening indirectly?

> we intentionally leave a (modest) chunk of VRAM for optimizations like these, and the end result is a much (MUCH) faster experience.

> Ephemeral transfers result in
            the data involved in intense operations being momentarily copied over to the GPU, and the results copied
            back to CPU.

Regarding performance:

> Using a patched Axolotl installation, I ran a merge of a Llama3-8B-DoRA adapter with r=64, where I capped the two cards to only use 7 GB of VRAM each for model loading (i.e. 14 GB available, where the model alone needs something like 16 GB).

Do you have a script to reproduce the experiment? What exactly is it measuring: loading, merging, saving? How would this affect inference and training?

> I am not happy with keeping the flag in LoraConfig, but I'm not sure how to propagate it to the right place, especially if we start using this in other places. It feels like this should be a runtime option, that does not persist, i.e. is not saved in configs and re-used inadvertently when people load a LoRA.

As is, this is very LoRA-specific, as it only affects DoRA (there should be a check that this should not be enabled if `use_dora is False`). I agree that putting it into the config might not seem like the right place, but I don't really see anywhere else to put it. Extra code could be added that this flag is not persisted as `True` when the config is saved.

kallewoof:
> > when the size of the data compared to the time to perform the operation on CPU memory is heavily time dominant, using ephemeral transfers has a fairly small VRAM overhead (depending on the size of the model/adapter) with orders of magnitude speed-up in certain operations.
> 
> I have to admit that I didn't fully understand this yet. Could you please explain this in more details or give a reference?

Sorry, what I meant was the following rule of thumb:

1. The smaller the data, the better
2. The bigger the time difference between performing the operation(s) on CPU vs GPU, the better.

Anywhere in the code where these two conditions apply, we should do what I refer to above as ephemeral transfers, which is simply to copy the values to the GPU, perform the computation, and pull the results back into CPU memory. I am beginning to regret even giving it a name, but I can't think of a succinct way to describe it.

> From looking at the code, my understanding is that if `lora_A` or `lora_B` is on CPU, it is ensured that the DoRA weight norm is also kept placed on CPU. Where is this part then happening? Is it happening indirectly?

When using ephemeral transfers we are putting the lora layers on the GPU. Because of this, the resulting DoRA weight norm ends up on the GPU. We don't want that, since it will take up GPU space, and the whole idea is to be ephemeral. As such, we pass in a `put it on the CPU` flag for the case when the incoming lora A/B's are on the CPU initially.

There are three cases:
1. Lora A/B are on the CPU, and we are **not** using ephemeral transfers: they remain on CPU throughout the operations, and the weight norm comes back on CPU.
2. Lora A/B are on the CPU, and we **are** using ephemeral transfers: copies of them are put on the GPU, the weight norm is calculated (ending up on the GPU), and is then explicitly moved to CPU, which is expected and desired.
3. Lora A/B are on the GPU, and the weight norm come back on the GPU as well. This is expected and desired. Use of ephemeral transfers is irrelevant here.

> Regarding performance:
> 
> > Using a patched Axolotl installation, I ran a merge of a Llama3-8B-DoRA adapter with r=64, where I capped the two cards to only use 7 GB of VRAM each for model loading (i.e. 14 GB available, where the model alone needs something like 16 GB).
> 
> Do you have a script to reproduce the experiment? What exactly is it measuring: loading, merging, saving? How would this affect inference and training?

This is the relevant code inside Axolotl (patched to add ephemeral_transfers):
```
        model_kwargs: Any = {}
        if cfg.lora_on_cpu:
            model_kwargs["max_memory"] = {"cpu": "256GiB"}
            model_kwargs["device_map"] = {"": "cpu"}
        if cfg.ephemeral_transfers:
            model_kwargs["ephemeral_transfers"] = cfg.ephemeral_transfers
        model = PeftModel.from_pretrained(
            model,
            cfg.lora_model_dir,
            is_trainable=(not inference),
            **model_kwargs,
        )
```
The lora_on_cpu part can be ignored, and was ununsed in my tests.

> > I am not happy with keeping the flag in LoraConfig, but I'm not sure how to propagate it to the right place, especially if we start using this in other places. It feels like this should be a runtime option, that does not persist, i.e. is not saved in configs and re-used inadvertently when people load a LoRA.
> 
> As is, this is very LoRA-specific, as it only affects DoRA (there should be a check that this should not be enabled if `use_dora is False`). I agree that putting it into the config might not seem like the right place, but I don't really see anywhere else to put it. Extra code could be added that this flag is not persisted as `True` when the config is saved.

If we had a good way to mark something as "do not store in final `adapter_config.json` file", that would be ideal and would make me happier about it being in the `LoraConfig` object.

kallewoof:
> Thanks a lot for explaining this in more detail, what you describe makes intuitive sense. Let's extend the help section of the new parameter to clearly indicate when users should set it and what they can expect.

Will do. I also would like to test inference/training and see if it might have a positive impact there, but I believe e.g. for the dora_init case, it is only called during initialization and then never again, so its impact should be rather small. I also don't think CPU offloading for stuff (especially lora layers) is broadly supported atm, but I could be mistaken.

I am also beginning to believe that what I am doing is similar to what the unsloth people are doing. I haven't looked at their code, but IIUC, they juggle things to/from CPU/GPU to speed things up, which is exactly what this method is about.

> > what I refer to above as ephemeral transfers
> 
> Okay, I was trying to google this and didn't find any description :)

Sorry, should've been more clear!

> > 2. Lora A/B are on the CPU, and we **are** using ephemeral transfers: copies of them are put on the GPU, the weight norm is calculated (ending up on the GPU)
> 
> Okay, I assume that this part is automatically taken care of already...

Yes. A torch operation such as `torch.linalg.norm()` will produce results on the same device as the parameters, which for case 2 above is the GPU.

> > and is then explicitly moved to CPU, which is expected and desired.
> 
> but it's not yet clear to me where this happens.

When case 2 above is true, the `place_on_cpu` flag is set in the call to `update_layer()` in `DoraLinearLayer`. Before return, the below code happens:

https://github.com/huggingface/peft/blob/dc2eae7878f0c51760df38993144fd34a7507411/src/peft/tuners/lora/dora.py#L61-L62

> I haven't worked with axolotl yet, so this unfortunately doesn't help me a lot.

Got it. I just meant to say that the speed-up happens inside the call to `PeftModel.from_pretrained()`, down the line, when it does the `dora_init` calls.

> I wonder if we could craft a PEFT example, or even better, a unit test, where we can check that this works correctly (or at least results in the expected speed up). Having a unit test would be really great to ensure that there are no regressions of this with future versions of PEFT.

That sounds like a good idea, but it relies on there being GPUs available for the unit tests. I haven't played with unit tests in PEFT yet so I'm clueless on how that works.

> > If we had a good way to mark something as "do not store in final `adapter_config.json` file", that would be ideal and would make me happier about it being in the `LoraConfig` object.
> 
> The first idea that comes to mind would be to have a private attribute on the config that stores all the attribute names that should not be persisted, something like `_attributes_not_to_persist: list[str]`. Then when the [`save_pretrained` method](https://github.com/huggingface/peft/blob/efda766f51398b3b40dcb35c0befbad8b9d5579f/src/peft/config.py#L49-L80) of the config is called, we pop all those attributes from the dict, then the attribute itself. WDYT?

That works for me! Sounds simple enough.

kallewoof:
> > That works for me! Sounds simple enough.

I thought about this a tiny bit more, and I think having a `runtime` dict inside the config that is dropped when present is even more straightforward.

BenjaminBossan:
> I thought about this a tiny bit more, and I think having a `runtime` dict inside the config that is dropped when present is even more straightforward.

I think this should also work. It's important that this is backwards compatible (new PEFT can load old configs) and ideally also forwards compatible (old PEFT can load new configs), but from the looks of it, it should work.

Let me know once this is ready for review. Please add a test for the new config option to ensure that it is indeed not stored. Testing the feature itself would also be nice as mentioned, if this isn't really feasible, an example would be nice.

src/peft/peft_model.py
```diff
@@ -374,6 +374,7 @@ def from_pretrained(
         is_trainable: bool = False,
         config: Optional[PeftConfig] = None,
         autocast_adapter_dtype: bool = True,
+        ephemeral_transfers: bool = False,
```
BenjaminBossan, line 377: How about adding this argument also to `load_adapter`? This would come into practice if users want to load multiple adapters at the same time.

tests/test_config.py
```diff
@@ -152,6 +152,19 @@ def test_from_pretrained_cache_dir_remote(self):
             PeftConfig.from_pretrained("ybelkada/test-st-lora", cache_dir=tmp_dirname)
             assert "models--ybelkada--test-st-lora" in os.listdir(tmp_dirname)
 
+    @parameterized.expand(ALL_CONFIG_CLASSES)
+    def test_save_pretrained_with_runtime_config(self, config_class):
```
BenjaminBossan, line 156: Let's also add a test that this works if users call `AutoModel.from_pretrained(..., ephemeral_transfer=True)`, and the same for `load_adapter` if the option is added there as well.

tests/test_common_gpu.py
```diff
@@ -1083,6 +1084,50 @@ def test_8bit_dora_merging(self):
...
+            use_dora=True,
+        )
+        peft_model = get_peft_model(model, config).eval()
+
+        # Save to disk
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            peft_model.save_pretrained(tmp_dir)
+
+            # Load from disk 100% on CPU without ephemeral transfers
+            start_time = time.time()
```
BenjaminBossan, line 1110:
```suggestion
            start_time = time.perf_counter()
```

Let's use `perf_counter` here and below.

tests/test_common_gpu.py
```diff
@@ -1083,6 +1084,50 @@ def test_8bit_dora_merging(self):
...
+
+            # Load again, with ephemeral transfers enabled
+            start_time = time.time()
+            peft_model = PeftModel.from_pretrained(
+                model,
+                tmp_dir,
+                device_map={"": "cpu"},
+                ephemeral_transfers=True,
+            ).eval()
+            elapsed_time2 = time.time() - start_time
```
BenjaminBossan, line 1126: Let's use a better name, e.g. `elapsed_time_ephemeral_transfer`.

tests/test_common_gpu.py
```diff
@@ -1083,6 +1084,50 @@ def test_8bit_dora_merging(self):
...
         assert torch.allclose(out_dora, out_unloaded, atol=atol, rtol=rtol)
 
+    @require_torch_gpu
+    @pytest.mark.single_gpu_tests
+    @require_bitsandbytes
+    def test_8bit_dora_cpu_offloading(self):
+        torch.manual_seed(0)
+        model = AutoModelForCausalLM.from_pretrained(
+            "facebook/opt-125m",
+            quantization_config=BitsAndBytesConfig(load_in_8bit=True),
```
BenjaminBossan, line 1094: Could you explain why we need to use bnb here?

src/peft/config.py
```diff
@@ -23,6 +23,30 @@
...
 
+@dataclass
+class PeftConfigRuntime:
+    """
+    This is the sub-configuration class to store the runtime configurations for the model.
+
+    Args:
+        ephemeral_transfers (`bool`): Whether to use ephemeral transfers for models partially kept in CPU memory.
+    """
+    ephemeral_transfers: bool = field(
```
BenjaminBossan, line 34: I'm still wondering about the name. Maybe we can find something that is a bit more specific. Just as an example, it could be `optimize_device_memory` or something like this. Of course, we may add more optimizations in the future, so it's hard to predict if that name will fit in the future. But I feel like users will be confused when reading "ephemeral transfers".

BenjaminBossan (changes requested):
Thanks a lot for the additions. The PR is on a good way, but let's harden it a little bit more. Please check my comments.

Apart from those, some more general points:

> It's a bit flaky, though, due to the tiny size of the model (facebook/opt-125m), but I think with a ramped up rank it should be OK.

How big is the variance in your tests? I currently can't test but I can do some testing next week. If it doesn't take too long, you could consider loading a few times and taking the average/median to compare. Or use a larger model if that helps.

Also, did you have the opportunity to check if/how this affects inference/training?

Finally, please run `make style` so that CI will pass successfully.

src/peft/config.py
kallewoof: I agree that the name is probably not good as is. `optimize_device_memory` sounds like it decreases VRAM usage, when in reality it gives a tiny momentary overhead. I'll think of "normal" succinct ways of saying it.

tests/test_common_gpu.py
kallewoof: We don't. I just used one of the existing tests as starting point. I can drop that part for sure.

tests/test_config.py
kallewoof: I initially did, but then I realized that it doesn't actually support adding this flag in the first place.

In `from_pretrained` (inside `PeftModel`), it does this little magic:
```python
        hf_hub_download_kwargs, class_kwargs, _ = cls._split_kwargs(kwargs)
```
and this code requires that the kwarg (`ephemeral_transfers` in this case) is a property of the class. But it isn't. Because it's stored inside the `runtime` property. As such it's dropped in `_` and ignored.

src/peft/peft_model.py
```diff
@@ -402,6 +403,12 @@ def from_pretrained(
                 loaded before calling `from_pretrained`.
             autocast_adapter_dtype (`bool`, *optional*):
                 Whether to autocast the adapter dtype. Defaults to `True`. Only relevant for specific adapter types.
+            ephemeral_transfers (`bool`, *optional*):
+                Whether to use ephemeral transfers for disk-offloaded modules. Defaults to `False`. This is useful
```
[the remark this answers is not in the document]
kallewoof: Fat-fingered. Should be CPU memory, of course.

src/peft/config.py
```diff
@@ -63,7 +90,7 @@ def save_pretrained(self, save_directory: str, **kwargs) -> None:
         os.makedirs(save_directory, exist_ok=True)
         auto_mapping_dict = kwargs.pop("auto_mapping_dict", None)
 
-        output_dict = asdict(self)
+        output_dict = self.to_dict()
```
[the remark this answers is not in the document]
kallewoof: I originally modified it because the other tests (the ones requiring that `load_from_json()` and `to_dict()` were identical) were now failing due to the extra `runtime` dictionary in the latter. I think moving the runtime stuff to the `LoraConfig` will address this partially, although I think those tests will now fail for `LoraConfig` only instead.

src/peft/config.py
BenjaminBossan: I see your point. "optmize" does not necessarily mean "decrease", but I agree that this is what many users might think. LMK if you come up with a better name.

src/peft/config.py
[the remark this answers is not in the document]
BenjaminBossan: It's okay if the LoRA unit test would need to be changed to account for this new key.

tests/test_common_gpu.py
BenjaminBossan: All right, please drop bnb, let's keep the test focused on what is needed.

src/peft/config.py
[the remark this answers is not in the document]
younesbelkada: Agreed with @BenjaminBossan with the same reasons he stated, if you could update the failing tests accordingly it would be really great

src/peft/config.py
younesbelkada: this sounds more or less (but I agree not exactly) as some sort of temporary CPU offloading. What about something like `ephemeral_cpu_offload` ?

src/peft/peft_model.py
younesbelkada: Yes that could work, however I am wondering if this can add some boilerplate since you would be able to activate that feature either from `load_adapter` or the lora config itself in that case no ?

younesbelkada: Thinking a bit more about it, I think this is probably fine, so I agree with @BenjaminBossan

src/peft/tuners/lora/layer.py
```diff
@@ -239,8 +240,16 @@ def dora_init(self, adapter_name: str) -> None:
         dora_layer = DoraLinearLayer(fan_in_fan_out=getattr(self, "fan_in_fan_out", False))
         lora_A = self.lora_A[adapter_name].weight
         lora_B = self.lora_B[adapter_name].weight
+        place_on_cpu = self.ephemeral_transfers and (lora_A.device.type == "cpu" or lora_B.device.type == "cpu")
+        if self.ephemeral_transfers:
+            if lora_A.device.type == "cuda":
+                lora_B = lora_B.to(lora_A.device)
+            else:
+                if lora_B.device.type != "cuda":
+                    lora_B = lora_B.to("cuda")
```
younesbelkada, line 249: what about the multi-GPU case here ? What would happen if the model is split across multiple GPUs, I suspect we might face a device mismatch ..

younesbelkada (commented):
Thanks a lot for adding this very nice feature, I left some minor comments. Could you also document this feature for end users in [`developper_guides/lora`](https://github.com/huggingface/peft/blob/d37dde61e1572f41cb43705f25f2015cad89b4c1/docs/source/developer_guides/lora.md) ?)

kallewoof:
Thanks a lot for the detailed feedback! Work pulled me away for a little bit, but I will address once that's out of the way.

src/peft/config.py
kallewoof: Well, the idea is to keep it in CPU memory all the time and, just as it is needed for some heavy calculation, copy it to GPU, and then remove it when finished. "Temporary CPU offloading" kind of sounds like the opposite of that, i.e. it's kept on GPU and then temporarily put in CPU memory for brief periods.

src/peft/config.py
```diff
@@ -23,6 +23,30 @@
...
+        default=False,
+        metadata={
+            "help": (
+                "Whether to use ephemeral transfers for models partially kept in CPU memory. Ephemeral transfers result in"
+                "the data involved in intense operations being momentarily copied over to the GPU, and the results copied"
+                "back to CPU. There is a momentary VRAM overhead, but operations are generally orders of magnitude faster"
+                "compared to performing them on the CPU. This is useful when parts of the model and/or components (such"
+                "as adapters) are kept in CPU memory until they are needed. Rather than perform expensive operations on"
+                "small data, the data is transferred to the GPU on-demand, the operation(s) performed, and the results"
+                "moved back to CPU memory. This brings a slight momentary VRAM overhead but gives orders of magnitude"
```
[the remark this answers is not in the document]
kallewoof: Removed last sentence and added "Currently only affects DoRA initialization."

src/peft/config.py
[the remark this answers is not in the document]
kallewoof: Reverted change and dropping runtime key from `output_dict` instead.

src/peft/tuners/lora/layer.py
kallewoof: This code is explicitly written to support multi GPU. Let's look at how the possible combinations will play out:

Let's denote chronological conditions as follows:
1. `A` is whether lora_A is on a "cuda" device; it infers that lora_B is placed on the same device (including device number) as lora_A. (L257-258)
2. `B` is whether lora_B is **not** on  a "cuda" device. It infers that lora_B is placed on an arbitrary cuda device (L260-261)
3. `not A` infers that lora_A is copied to the same device (including device number) as lora B

1. lora_A, lora_B on CPU: `not A → B` (cuda:A, cuda:A)
2. lora_A on CPU, lora_B on cuda:B: `not A → not B` (cuda:B, cuda:B)
3. lora_A on cuda:C, lora_B on cuda:D: `A` (cuda:C, cuda:C)

kallewoof: OK, I think I got tests in place that check all possible combinations on a multi GPU setup.

tests/test_config.py
kallewoof: So, the `test_8bit_dora_cpu_offloading` test does check `PeftModel.from_pretrained` with `ephemeral_transfers=True`. I'm not sure what a good test would be that checks this beyond what is done already. Is there some feature that is not being tested properly in this scenario?

kallewoof:
@BenjaminBossan @younesbelkada I believe I addressed all feedback. Let me know if I missed something and I will address.

src/peft/config.py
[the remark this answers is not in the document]
kallewoof: A tiny sidenote: I overrode this method in `LoraConfig` and dropped runtime there. Keeping `output_dict = self.to_dict()` here means no other changes necessary, so I kept that.

src/peft/tuners/lycoris_utils.py
```diff
@@ -79,7 +79,8 @@ def __init__(self, base_layer: nn.Module) -> None:
 
     @property
     @abstractmethod
-    def _available_adapters(self) -> set[str]: ...
+    def _available_adapters(self) -> set[str]:
+        ...
```
kallewoof, line 83: `make style` did this. Should I revert?

tests/test_config.py
```diff
@@ -152,6 +152,19 @@ def test_from_pretrained_cache_dir_remote(self):
...
+        r"""
+        Test if the config correctly removes runtime when saving
+        """
+        with tempfile.TemporaryDirectory() as tmp_dirname:
+            for model_name, revision in PEFT_MODELS_TO_TEST:
+                cfg = config_class.from_pretrained(model_name, revision=revision)
+                cfg.runtime.ephemeral_transfers = True
```
kallewoof, line 163: I'm confused why this is not erroring since it should only work for `LoraConfig`.

src/peft/tuners/lora/config.py
```diff
@@ -282,6 +309,15 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    runtime: LoraConfigRuntime = field(default_factory=LoraConfigRuntime, metadata={"help": "Runtime configurations"})
+
+    def to_dict(self):
```
BenjaminBossan, line 314: I have a small concern about having to use `to_dict`.

PEFT is used as a utility library by other packages. What I'm afraid could happen is that some of these packages may not use `peft_config.save_pretrained` directly but instead call `config = asdict(peft_config)` to do some stuff on the dict before persisting it. Then it would contain the runtime entry. If such a config is later loaded, it would be subtly different to what we would normally get with PEFT.

Maybe it's not a huge concern overall. I tried to think of a way to ensure that `asdict(peft_config)` would also remove the `runtime` key, but that would require some hacking (e.g. popping it from `__dataclass_fields__`) and I'm not sure if that would break more things down the line.

Anyway, if you have a smart idea to resolve this, LMK.

src/peft/tuners/lycoris_utils.py
BenjaminBossan: Not necessary. This change should already be on latest main, so you could merge with/rebase on main to remove this diff.

tests/test_common_gpu.py
```diff
@@ -1083,6 +1085,77 @@ def test_8bit_dora_merging(self):
...
+            start_time = time.perf_counter()
+            peft_model = PeftModel.from_pretrained(
+                model,
+                tmp_dir,
+                device_map={"": "cpu"},
+            ).eval()
+            elapsed_time = time.perf_counter() - start_time
+
+            # Load again, with ephemeral transfers enabled
+            start_time = time.time()
```
BenjaminBossan, line 1118:
```suggestion
            start_time = time.perf_counter()
```

tests/test_common_gpu.py
```diff
@@ -1083,6 +1085,77 @@ def test_8bit_dora_merging(self):
         assert torch.allclose(out_dora, out_unmerged, atol=atol, rtol=rtol)
         assert torch.allclose(out_dora, out_unloaded, atol=atol, rtol=rtol)
 
+    @require_torch_gpu
+    @pytest.mark.single_gpu_tests
+    def test_8bit_dora_cpu_offloading(self):
```
BenjaminBossan, line 1090: Let's rename the test, as "8bit" is not being used. Also, let's use the name of the runtime argument in the test name, depending on what we will settle on.

tests/test_common_gpu.py
```diff
@@ -1083,6 +1085,77 @@ def test_8bit_dora_merging(self):
...
+            ).eval()
+            elapsed_time_ephemeral_transfer = time.perf_counter() - start_time
+
+        # CPU only should be much slower
+        assert elapsed_time > 1.1 * elapsed_time_ephemeral_transfer
+
+    @require_torch_gpu
+    @require_torch_multi_gpu
+    @pytest.mark.multi_gpu_tests
+    def test_8bit_dora_cpu_offloading_multigpu(self):
```
BenjaminBossan, line 1133: Let's also rename this test.

tests/test_common_gpu.py
```diff
@@ -1083,6 +1085,77 @@ def test_8bit_dora_merging(self):
...
+            init_lora_weights=False,
+            use_dora=True,
+            runtime=LoraConfigRuntime(ephemeral_transfers=True),
+        )
+        peft_model = get_peft_model(model, config).eval()
+
+        layer = peft_model.base_model.model.model.decoder.layers[0].self_attn.v_proj
+        lora_A, lora_B = layer.lora_A, layer.lora_B
+
+        possible_combinations = ["cpu", "cuda", "cuda:0", "cuda:1"]
```
BenjaminBossan, line 1151: Nice work!

tests/test_config.py
BenjaminBossan: In that test, you could save the model that has `ephemeral_transfer=True` at the end, then load the config and check that it's `False`.

BenjaminBossan (changes requested):
Thanks for the updates. Not much is missing at this point. I have a few comments left, please take a look. Also, anything about what I wrote earler:

> How big is the variance in your tests? I currently can't test but I can do some testing next week. If it doesn't take too long, you could consider loading a few times and taking the average/median to compare. Or use a larger model if that helps.

> Also, did you have the opportunity to check if/how this affects inference/training?

I'll also run some checks once I'm back at my machine.

src/peft/tuners/lora/config.py
```diff
@@ -23,6 +23,31 @@
...
+
+    Args:
+        ephemeral_transfers (`bool`): Whether to use ephemeral transfers for models partially kept in CPU memory.
+    """
+
+    ephemeral_transfers: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Whether to use ephemeral transfers for models partially kept in CPU memory. Ephemeral transfers result in"
```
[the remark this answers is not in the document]
kallewoof: Oops. Fixed.

tests/test_common_gpu.py
kallewoof: Good catch. Using ephemeral_transfers until a name is decided.

src/peft/tuners/lycoris_utils.py
kallewoof: I don't think that's the case. I rebased on main before I started work on the changes earlier.

src/peft/tuners/lora/config.py
kallewoof: I definitely see the concern. I think the best way to address that is to ensure that we don't adhere to runtime configs found in loaded configurations. That way we can nearly guarantee that it won't actually matter.

And I think my tests so far are checking that this is indeed the case. Perhaps hand crafted config files that add runtime_config to them and then load those and checking that it's ignored should be added to the tests too.

src/peft/tuners/lycoris_utils.py
BenjaminBossan: Ah yes, then it's most likely because you have an older ruff version installed. Try v0.4.10.

tests/test_config.py
BenjaminBossan: > Would it be cheating to enable ephemeral transfers in the LoraConfig in the `test_common_gpu/test_dora_ephemeral_transfers` starting model and then check that it's not there on load?

Yes, sorry, that's what I meant, not here.

tests/test_common_gpu.py
kallewoof: D'oh. Thanks.

tests/test_config.py
```diff
@@ -152,6 +152,19 @@ def test_from_pretrained_cache_dir_remote(self):
...
             assert "models--ybelkada--test-st-lora" in os.listdir(tmp_dirname)
 
+    @parameterized.expand(ALL_CONFIG_CLASSES)
+    def test_save_pretrained_with_runtime_config(self, config_class):
+        r"""
+        Test if the config correctly removes runtime config when saving
+        """
+        with tempfile.TemporaryDirectory() as tmp_dirname:
+            for model_name, revision in PEFT_MODELS_TO_TEST:
+                cfg = config_class.from_pretrained(model_name, revision=revision)
```
BenjaminBossan, line 162: This test does not do what it appears to be doing. No matter the `config_class`, ` config_class.from_pretrained` will always return a `LoraConfig` instance because the stored config comes from `LoraConfig`. This is a bit unintuitive but hard to change now without breaking existing code.

kallewoof: Ahh, yeah, that's what I suspected. Thanks for verifying. I'll add a note about that for future explorers.

src/peft/tuners/lycoris_utils.py
kallewoof: That was it!

tests/test_config.py
kallewoof: Got it! I think I figured out a good way to do that now. It's in the test_dora_ephemeral_transfers code. Specifically, it:
```
        peft_model = get_peft_model(model, config).eval()
        # Check that ephemeral transfers are present
        assert peft_model.peft_config['default'].runtime_config.ephemeral_transfers
```
then
```
            peft_model = PeftModel.from_pretrained(
                model,
                tmp_dir,
                device_map={"": "cpu"},
            ).eval()
            elapsed_time = time.perf_counter() - start_time

            # Check that ephemeral transfers are absent
            assert not peft_model.peft_config['default'].runtime_config.ephemeral_transfers
```

kallewoof:
I bumped the model to opt-350b, after seeing an intermittent failure on my end. Feel free to at-bump me after merging if this is causing issues and I'll adjust it to be more stable.

tests/test_common_gpu.py
```diff
@@ -1083,6 +1085,85 @@ def test_8bit_dora_merging(self):
         assert torch.allclose(out_dora, out_unmerged, atol=atol, rtol=rtol)
         assert torch.allclose(out_dora, out_unloaded, atol=atol, rtol=rtol)
 
+    @require_torch_gpu
+    @pytest.mark.single_gpu_tests
+    def test_dora_ephemeral_transfers(self):
+        torch.manual_seed(0)
+        model = AutoModelForCausalLM.from_pretrained(
+            "facebook/opt-350m",
```
BenjaminBossan, line 1093: I ran the tests on a 4090 and unfortunately, it is very flaky for me, failing in 4 out of 5 tries with

> assert 2.1404892549999204 > (1.1 * 1.980610250000609)

and similar numbers. I also tried a bigger model, `bloomz-560m`, but that still failed. I wonder if we should scrap the performance test if the gain doesn't materialize on smaller models (and we don't want to load huge models during testing either).

Not sure what we can really test instead, maybe just that there is no error and the device, similar to the multi-GPU test?

As a "proof" that this works, I would suggest to instead add a short script to the `examples/` directory that uses a bigger model where there is a very noticeable gain. WDYT?

docs/source/developer_guides/lora.md
```diff
@@ -122,6 +122,22 @@ from peft import LoraConfig
 config = LoraConfig(use_dora=True, ...)
 ```
 
+If parts of the model or the DoRA adapter are offloaded to CPU you can get a significant speedup at the cost of some temporary (ephemeral) VRAM overhead by using `ephemeral_transfers=True` in `config.runtime_config`.
+
+```py
+from peft import LoraConfig, LoraRuntimeConfig
```
BenjaminBossan, line 128: This import fails as `LoraRuntimeConfig` is not loaded to the PEFT `__init__.py`.

src/peft/tuners/lora/config.py
BenjaminBossan: I agree that would be a good test. I think for that test to pass, we need an update to `from_pretrained` that removes the runtime config if present.

src/peft/tuners/lora/config.py
```diff
@@ -23,6 +23,31 @@
...
+@dataclass
+class LoraConfigRuntime:
+    """
+    This is the sub-configuration class to store the runtime configurations for the model.
+
+    Args:
+        ephemeral_transfers (`bool`): Whether to use ephemeral transfers for models partially kept in CPU memory.
+    """
+
+    ephemeral_transfers: bool = field(
```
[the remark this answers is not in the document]
BenjaminBossan: That name works for me.

BenjaminBossan (changes requested):
Thanks for all the updates. I think if we get the last few kinks ironed out, this PR should be good.

tests/test_common_gpu.py
kallewoof: Yes, that seems to be a better approach. My intention is to expand this in other areas post-merge. Hopefully one of those is more obvious.

I'll rewrite this to check that the resulting weights are identical instead, which ensures that this code does not screw anything up.

docs/source/developer_guides/lora.md
kallewoof: D'oh, thanks. Fixed.

src/peft/tuners/lora/config.py
kallewoof: You mean in `PeftModel.from_pretrained()`? I was hoping we could avoid that, but I think you're right, as w can't guarantee users won't throw runtime configs in odd places.

kallewoof: I dug around a bit and I'm not sure this is needed. I could be missing something, but:

1. The user provides a config which contains a runtime config: the ephemeral GPU offload flag is overridden by the `from_pretrained` code, resulting in a NOOP.
2. The user config does not contain a runtime config: expected.

tests/test_common_gpu.py
kallewoof: Added example that loads a DoRA onto a model, and optionally merges and saves it to disk, while also timing it.

The one annoying part is that it will obviously not get the time right if the model is not yet in the HF hub cache and has to be downloaded first. I might look into whether I can force a "download only if necessary" call before the actual model/DoRA load part.

kallewoof: I put in `snapshot_download` calls before the timers. It downloads some unnecessary stuff but those aren't very large, and it looks like it works otherwise.

kallewoof:
Thanks for the review! I believe I addressed all issues.

examples/ephemeral_gpu_offloading/load_with_dora.py
```diff
@@ -0,0 +1,90 @@
...
+    parser.add_argument("--model", type=str, default="NousResearch/Hermes-2-Pro-Mistral-7B", help="Model to load")
+    parser.add_argument(
+        "--dora",
+        type=str,
+        default="anamikac2708/Mistral-7B-DORA-finetuned-investopedia-Lora-Adapters",
+        help="DoRA to use",
+    )
+    parser.add_argument("--ephemeral_gpu_offload", action="store_true", help="Use ephemeral GPU offloading")
+    parser.add_argument(
+        "--merge_model", action="store_true", help="Merge the model with the DoRA model and save to disk"
```
BenjaminBossan, line 24: How about changing this argument from a bool to a path, so that users can directly indicate where to save the merged model. If nothing is passed, the model is not merged and saved.

examples/ephemeral_gpu_offloading/load_with_dora.py
```diff
@@ -0,0 +1,90 @@
...
+    if args.merge_model:
+        merged_model = peft_model.merge_and_unload(progressbar=True)
+        merged_model.save_pretrained("merged_model")
+
+
+if __name__ == "__main__":
+    main()
+
+"""
+Example outputs:
```
BenjaminBossan, line 68: I think it's better to put this docstring to the top of the file and integrate the comment into the docstring. Also, please add the license notice.

examples/ephemeral_gpu_offloading/load_with_dora.py
```diff
@@ -0,0 +1,90 @@
...
+
+(Note: if this was the first time you ran the script, or if your cache was cleared, the times shown above are invalid, due to the time taken to download the model and DoRA files. Just re-run the script in this case.)
+
+$ python load_with_dora.py --ephemeral_gpu_offload
+-------------------- Loading model --------------------
+Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:03<00:00,  1.11it/s]
+-------------------- Loading PeftModel --------------------
+-------------------- Done --------------------
+Model loading time: 4.28s
+PeftModel loading time: 16.59s
```
BenjaminBossan, line 86: As a data point, I got 20s vs 10s on my machine.

examples/ephemeral_gpu_offloading/load_with_dora.py
```diff
@@ -0,0 +1,90 @@
...
+        print(f"Failed to download model: {e}")
+        # We continue anyway as this might be e.g. a local directory or something
+    try:
+        snapshot_download(repo_id=args.dora)
+    except Exception as e:
+        print(f"Failed to download DoRA: {e}")
+        # We continue anyway as this might be e.g. a local directory or something
+
+    start = time.perf_counter()
+    print("-" * 20 + " Loading model " + "-" * 20)
```
BenjaminBossan, line 47: IMO, the `"-" * 20` parts are unnecessary.

examples/ephemeral_gpu_offloading/load_with_dora.py
```diff
@@ -0,0 +1,90 @@
...
+from peft import PeftModel
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Load a model with DoRA using ephemeral GPU offloading")
+    parser.add_argument("--model", type=str, default="NousResearch/Hermes-2-Pro-Mistral-7B", help="Model to load")
+    parser.add_argument(
+        "--dora",
+        type=str,
+        default="anamikac2708/Mistral-7B-DORA-finetuned-investopedia-Lora-Adapters",
```
BenjaminBossan, line 19: I'd feel a bit more comfortable pointing this to a PEFT-internal repo. I created a DoRA adapter for this with the same size:

`peft-internal-testing/DoRA-Hermes-2-Pro-Mistral-7B`

This adapter is not trained but for the purpose of this example, I don't think it matters.

src/peft/tuners/lora/config.py
BenjaminBossan: I meant the `PeftConfigMixin.from_pretrained` should be adapted. E.g. this piece of code would currently fail:

```python
import json
from dataclasses import asdict
from peft import LoraConfig, LoraRuntimeConfig

config = LoraConfig(runtime_config=LoraRuntimeConfig(ephemeral_gpu_offload=True))
dct = asdict(config)

assert dct["runtime_config"]["ephemeral_gpu_offload"] is True

with open("/tmp/peft/1859/adapter_config.json", "w") as f:
    json.dump(dct, f)

with open("/tmp/peft/1859/adapter_config.json", "r") as f:
    loaded = json.load(f)

# this fails
assert loaded["runtime_config"]["ephemeral_gpu_offload"] is False
```

BenjaminBossan (changes requested):
I think the tests are better now without the flakiness and the added example works well enough to illustrate the advantage, thanks a lot. I have a few suggestions to improve that example, and still one issue regarding the config file. Please check.

examples/ephemeral_gpu_offloading/load_with_dora.py
kallewoof: Thank you! I didn't think of just doing a non-trained DoRA! Couldn't find one readily available.

examples/ephemeral_gpu_offloading/load_with_dora.py
kallewoof: How about "--- text ---"?

src/peft/tuners/lora/config.py
kallewoof: That code isn't really doing anything inside the PEFT library though, it's just saving/loading JSON data. Since people might (incorrectly) put runtine config stuff in the JSON files themselves, the sane approach is to ignore it *inside PEFT* (and write helpful warnings when encountered), rather than trying to keep it out of the config files.

You could also flip it around: in a near future, we might actually *want* to pass runtime config content around using asdict, e.g. when serializing configurations for distributed training or something. In these cases, the above would be fully expected and desired behavior.

examples/ephemeral_gpu_offloading/load_with_dora.py
```diff
@@ -0,0 +1,103 @@
...
+        print(f"Failed to download model: {e}")
+        # We continue anyway as this might be e.g. a local directory or something
+    try:
+        snapshot_download(repo_id=args.dora)
+    except Exception as e:
+        print(f"Failed to download DoRA: {e}")
+        # We continue anyway as this might be e.g. a local directory or something
+
+    start = time.perf_counter()
+    print("--- Loading model ---")
```
BenjaminBossan, line 85: Not sure we need any special delimiter, but I guess this is also fine :)

The docstring above does, however, still show the old output.

src/peft/tuners/lora/config.py
BenjaminBossan: Hmm, but I don't see the harm in adding a check to `PeftModel.from_pretrained` and removing the runtime config there. The example code could be considered "outside PEFT", but using `asdict` on a PEFT config and getting a valid dict representation would be a reasonable assumption for a package that builds on top of PEFT.

> in a near future, we might actually _want_ to pass runtime config content around using asdict, e.g. when serializing configurations for distributed training or something

In that case, we can always make further changes. But as I said, I don't even want to change the output of `asdict(config)` but rather just ensure that when we read the config, we don't accidentally include the runtime config.

BenjaminBossan (commented):
> I may have messed this up, but I tried doing what you asked.

No, looks good.

> Try a 70B model/DoRA next. ;)

As soon as I have a sufficiently large GPU lying around :-p

examples/ephemeral_gpu_offloading/load_with_dora.py
kallewoof: Oops. Fixed.

src/peft/tuners/lora/config.py
kallewoof: OK, I think I got a good solution. How does 40ff66b461437d4ea171ab6d337fa2954a286afd look to you?

kallewoof: OK, I think I got a good solution. How does 40ff66b461437d4ea171ab6d337fa2954a286afd look to you?

kallewoof:
@BenjaminBossan  The thread is buggy as hell, so I'm commenting here instead: 

> Hmm, but I don't see the harm in adding a check to PeftModel.from_pretrained and removing the runtime config there. The example code could be considered "outside PEFT", but using asdict on a PEFT config and getting a valid dict representation would be a reasonable assumption for a package that builds on top of PEFT.

OK, I think I got a good solution. How does 40ff66b461437d4ea171ab6d337fa2954a286afd look to you?

kallewoof:
I don't think these are related to my changes?

* 3.10, windows-latest: FAILED tests/test_tuners_utils.py::TestModelAndLayerStatus::test_base_model_type_transformers_automodel - FileNotFoundError: No such file or directory: "C:\\Users\\runneradmin\\.cache\\huggingface\\hub\\models--google--flan-t5-small\\snapshots\\0fc9ddf78a1e988dac52e2dac162b0ede4fd74ab\\model.safetensors"
* 3.11, macos-12: FAILED tests/test_stablediffusion.py::StableDiffusionModelTester::test_add_weighted_adapter_base_unchanged_0_test_hf_internal_testing_tiny_stable_diffusion_torch_lora - huggingface_hub.utils._errors.LocalEntryNotFoundError: An error happened while trying to locate the file on the Hub and we cannot find the requested files in the local cache. Please check your connection and try again or make sure your Internet connection is on.

BenjaminBossan (approved):
Thanks for adding this nice runtime optimization for DoRA (and hopefully more in the future). Great work and thanks for your patience with the review.

> I don't think these are related to my changes?

Yes, don't worry, these are just issues with the CI getting timeouts from the Hub.

PR #1862, opened 2024-06-16
FIX multitask prompt tuning paper link

cep-ter:
This PR fixes multitask prompt tuning paper link.


younesbelkada (approved):
Thanks for fixing the prompt tuning paper link !

PR #1866, opened 2024-06-17
CI testing BNB: remove single GPU tests

BenjaminBossan:
In #1859, we checked removing the import checks, but the single-GPU BNB multi-backend branch is still
[stuck](https://github.com/huggingface/peft/actions/runs/9540932758/job/26293392525#logs). Therefore, check commenting the next step instead.

Note: I only uncommented one step and commented the next. For some reason, the GitHub diff view is confused about that.

**Edit**: Tests are failing but for unrelated reasons, most likely because of numpy 2.0.

### DIFF
--- a/.github/workflows/nightly-bnb.yml
+++ b/.github/workflows/nightly-bnb.yml
@@ -47,40 +47,40 @@ jobs:
             cd .. 
           fi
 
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Thanks ! 🤞

PR #1868, opened 2024-06-17
Workflow: Attempt to fix the current failures

younesbelkada:
Let's use `job_id.outcome` instead of `job_id.outputs.status` that always returns empty strings, e.g: https://github.com/huggingface/peft/actions/runs/9540932758/job/26293391870#step:7:14

cc @BenjaminBossan


younesbelkada:
Reference: https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context

BenjaminBossan (approved):
Let's give it a try :+1:

PR #1871, opened 2024-06-18
CI Downgrade numpy to <2.0 for Mac and Windows

BenjaminBossan:
As a temporary workaround to numpy 2.0 breaking CI, downgrade it for `macos-12` and `windows-latest`.


younesbelkada (approved):
Thanks so much !

BenjaminBossan:
Thanks for the review. I updated the comment, will merge without waiting for CI as the comment should not affect it.

Issue #1837, opened 2024-06-09
Tutorial notebook for applying PEFT with DNA Language models


rahulbshrestha:
### Feature request

A notebook tutorial that shows an example of how to use parameter-efficient fine-tuning techniques from the PEFT library (e.g LoRA, IA3, etc.) with a DNA Language Model (e.g https://huggingface.co/gagneurlab/SpeciesLM). 

The tutorial will contain an example of how to load and apply a DNA language model with and without PEFTs on an arbitrary task. There will be detailed descriptions for each part of the code, similar to a high-quality blog post like [this](https://huggingface.co/blog/AmelieSchreiber/protein-binding-partners-with-esm2).

### Motivation

This will show the possibility of using Hugging Face's PEFT library in a computational biological setting e.g with DNA Language Models to solve a task.

### Your contribution

I'll submit a PR if this is a good idea. 

Question: Should I place the notebook in the [examples folder](https://github.com/huggingface/peft/tree/main/examples)?

BenjaminBossan:
Thanks for suggesting to provide such a notebook. It would be indeed a good addition for our examples in the `examples/` directory.

Note that it looks like people had issues with ESM2 in the past (#1023). If you found a solution, that would be really helpful.

BenjaminBossan:
Resolved by #1873

PR #1873, opened 2024-06-18
Tutorial Notebook: Using the PEFT library with a DNA Language Model.

rahulbshrestha:
This PR adds a tutorial notebook that uses the PEFT library with a DNA Language Model to solve a downstream task.


BenjaminBossan:
Thanks, this looks good so far. Ping me once the notebook is ready for review.

For general context, check #1837.

rahulbshrestha:
@BenjaminBossan this tutorial is ready for review!

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "from transformers import Trainer, TrainingArguments\n",
+    "\n",
+    "# Define training arguments\n",
+    "training_args = TrainingArguments(\n",
+    "    output_dir='./results',\n",
+    "    evaluation_strategy=\"epoch\",\n",
```
BenjaminBossan, line 671: We get a warning that `evaluation_strategy` is deprecated and `eval_strategy` should be used instead. Let's switch.

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+    "# Step 3: Evaluate the Predictions (Optional)\n",
+    "from sklearn.metrics import accuracy_score, precision_recall_fscore_support\n",
+    "\n",
+    "accuracy = accuracy_score(test_labels, predicted_labels)\n",
+    "print(f\"Accuracy: {accuracy}\")\n",
+    "\n",
+    "precision, recall, f1, _ = precision_recall_fscore_support(test_labels, predicted_labels, average='weighted')\n",
+    "print(f\"Precision: {precision}\")\n",
+    "print(f\"Recall: {recall}\")\n",
+    "print(f\"F1-Score: {f1}\")"
```
BenjaminBossan, line 766: Precision/recall/F1 are all the same and identical to accuracy. I wonder if it's not enough to just report accuracy? As accuracy can be calculated easily, this would also allow us to remove the sklearn dependency, which is also not a PEFT dependency and would otherwise require users to install it.

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Accuracy: 0.795\n",
+      "Precision: 0.794957729468599\n",
+      "Recall: 0.795\n",
+      "F1-Score: 0.794799276945093\n"
```
BenjaminBossan, line 752: Could please add a sentence or two about what we should think about the results. Is this good, bad, are there maybe published values that this could be compared to? If the result is not so good, maybe add a sentence along the lines of: "The results are not very good but this is because we train with little data and haven't spent time tuning the model for better results...".

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "71b8a749-461e-4533-b1d0-cebc924d3dc0",
+   "metadata": {},
+   "source": [
+    "The LoraConfig object is instantiated to configure the PEFT parameters:\n",
+    "\n",
+    "- task_type: Specifies the type of task, in this case, sequence classification (SEQ_CLS).\n",
+    "- r: Number of re-parameterization steps.\n",
```
BenjaminBossan, line 865: Maybe better: The rank of the LoRA matrices.

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+   "cell_type": "code",
+   "execution_count": 116,
+   "id": "6c223937-86ea-42ef-991a-050f23b21ef9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from peft import LoraConfig, TaskType\n",
+    "\n",
+    "peft_config = LoraConfig(\n",
+    "    #task_type=\"SEQ_CLS\",\n",
```
BenjaminBossan, line 995: Why is this commented out?

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+    "    r=8,\n",
+    "    lora_alpha=32,\n",
```
BenjaminBossan, line 997: Typically, alpha is ~ 2x rank. Did you test that and found that 32 works better?

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+    "    target_modules=[\"query\", \"key\", \"value\"],\n",
+    "    lora_dropout=0.01,\n",
```
BenjaminBossan, line 999: This is a small value for dropout. Did you find that it works well?

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+       "TrainOutput(global_step=130, training_loss=0.6522058927095853, metrics={'train_runtime': 48.5074, 'train_samples_per_second': 41.231, 'train_steps_per_second': 2.68, 'total_flos': 0.0, 'train_loss': 0.6522058927095853, 'epoch': 10.0})"
+      ]
+     },
+     "execution_count": 119,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "from transformers import Trainer, TrainingArguments\n",
```
BenjaminBossan, line 1244: No need to import a second time.

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+    "\n",
+    "# Define training arguments\n",
+    "training_args = TrainingArguments(\n",
+    "    output_dir='./results',\n",
+    "    evaluation_strategy=\"epoch\",\n",
+    "    learning_rate=2e-5,\n",
```
BenjaminBossan, line 1250: For LoRA training, we can often get away with a ~10x higher learning rate. Did you test that? Maybe we can get better test scores this way.

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Accuracy: 0.65\n",
+      "Precision: 0.6567213280696427\n",
+      "Recall: 0.65\n",
+      "F1-Score: 0.6491592433189871\n"
```
BenjaminBossan, line 1330: Again, giving some context on these results and some ideas how to improve them would be nice.

BenjaminBossan (changes requested):
Thanks a lot for providing this detailed notebook on DNA-LM training with PEFT. I added a couple of comments, please check them out.

examples/dna_language_models/dna_lm.ipynb
rahulbshrestha: Is this necessary when specifying the LoraConfig? It seems to work fine without specifying the task type.

examples/dna_language_models/dna_lm.ipynb
rahulbshrestha: Yes, changed!

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+   "execution_count": 98,
+   "id": "efa9441d-f44c-4ca3-b24c-fa5c853896cd",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "test_sequences = []\n",
+    "train_sequences = []\n",
+    "val_sequences = []\n",
+    "\n",
+    "dataset_limit = 200\n",
```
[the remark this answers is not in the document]
rahulbshrestha: Yes, the entire datasets takes way longer.

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 97,
+   "id": "75f267a9-82d1-4343-982e-9b1ea542a330",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def get_kmers(seq, k=6, stride=1):\n",
```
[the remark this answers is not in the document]
rahulbshrestha: Added!

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,1369 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 91,
+   "id": "e73fae58-03e9-4acc-b0fc-9bc810c7d366",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "tokenizer = AutoTokenizer.from_pretrained(\"gagneurlab/SpeciesLM\", revision = \"downstream_species_lm\")\n",
```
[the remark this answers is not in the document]
rahulbshrestha: I added some information, is that enough?

examples/dna_language_models/dna_lm.ipynb
rahulbshrestha: I haven't experimented with the review comments regarding changing hyperparameters for obtaining a higher score, are these necessary for a tutorial notebook?

examples/dna_language_models/dna_lm.ipynb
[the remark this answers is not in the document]
BenjaminBossan: Yes, that helps. Still I think many users could be reluctant to try the model, so if you know another model that could be used instead, it would be nice to make the switch and just add a note here to suggest the SpeciesLM model.

examples/dna_language_models/dna_lm.ipynb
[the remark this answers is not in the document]
BenjaminBossan: Okay, so let's add a comment that this is an optimization for speed, but for best results users should set this variable to `None` and use the whole dataset.

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,642 @@
...
+    "import torch\n",
+    "from torch import nn\n",
+    "\n",
+    "class DNA_LM(nn.Module):\n",
+    "    def __init__(self, model, num_labels):\n",
+    "        super(DNA_LM, self).__init__()\n",
+    "        self.model = model\n",
+    "        self.classifier = nn.Linear(768, num_labels)\n",
+    "        self.in_features = 768\n",
+    "        self.out_features = 2\n",
```
BenjaminBossan, line 334: Those are not really needed, are they?

examples/dna_language_models/dna_lm.ipynb
BenjaminBossan: No, not necessary at all. It's just that when an example deviates from the defaults, I wonder if the author made these choices because they found them to work better or if there is another reason. Same with the dropout rate below.

BenjaminBossan (changes requested):
Thanks for the updates. There are only a few minor issues left.

Also, could you please run the whole notebook before committing it, so that the outputs are directly visible to users without having to run the notebook first.

examples/dna_language_models/dna_lm.ipynb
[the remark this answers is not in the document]
rahulbshrestha: I haven't used any other models, so will stick to SpeciesLM for now, since it's the one I'm most familiar with. It also comes from a lab from my university (TU Munich) so I know it is safe.

examples/dna_language_models/dna_lm.ipynb
rahulbshrestha: Apparently, it's needed else I get thrown an error when using Trainer that self.in_features and self.out_features is missing. I've made it a bit cleaner so the numbers aren't hard-coded.

rahulbshrestha:
@BenjaminBossan I just added the second review changes as you requested.

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,2941 @@
...
+      "metadata": {
+        "id": "efa9441d-f44c-4ca3-b24c-fa5c853896cd"
+      },
+      "outputs": [],
+      "source": [
+        "test_sequences = []\n",
+        "train_sequences = []\n",
+        "val_sequences = []\n",
+        "\n",
+        "dataset_limit = 200\n",
```
BenjaminBossan, line 482: Could you please add a comment that this value should be set to `None` if users want to use the whole dataset?

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,2941 @@
...
+        "        if labels is not None:\n",
+        "            loss_fct = nn.CrossEntropyLoss()\n",
+        "            loss = loss_fct(logits.view(-1, self.out_features), labels.view(-1))\n",
+        "\n",
+        "        return (loss, logits) if loss is not None else logits\n",
+        "\n",
+        "# Number of classes for your classification task\n",
+        "num_labels = 2\n",
+        "classification_model = DNA_LM(lm, num_labels)\n",
+        "classification_model.to('cuda')"
```
BenjaminBossan, line 713:
```suggestion
        "classification_model.to('cuda')";
```

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,2941 @@
...
+          "execution_count": 40,
+          "metadata": {},
+          "output_type": "execute_result"
+        }
+      ],
+      "source": [
+        "# Number of classes for your classification task\n",
+        "num_labels = 2\n",
+        "classification_model = DNA_LM(lm, num_labels)\n",
+        "classification_model.to('cuda')"
```
BenjaminBossan, line 1192:
```suggestion
        "classification_model.to('cuda')";
```

examples/dna_language_models/dna_lm.ipynb
```diff
@@ -0,0 +1,2941 @@
...
+              ")"
+            ]
+          },
+          "execution_count": 39,
+          "metadata": {},
+          "output_type": "execute_result"
+        }
+      ],
+      "source": [
+        "classification_model"
```
BenjaminBossan, line 1068: Can be removed

BenjaminBossan (changes requested):
Thanks for the updates, we're almost good. Just a few small things:

- Let's add a comment about the data size limit
- Let's not print the model so often, as it clutters the notebook and distracts from the important things
- It would be nice to have a short conclusion, e.g.: How much memory is saved with PEFT? If users want to improve the results, what could they test next?

rahulbshrestha:
@BenjaminBossan ok, I made changes as you requested!

BenjaminBossan (approved):
Thanks for adding this notebook, the last changes LGTM.

PR #1874, opened 2024-06-19
FIX Error when using VeRA with float16 or bfloat16

### LINKED ISSUE #1872: VeRA doesn't work with lower-precision models
### System Info

python 3.10.14
accelerate 0.31.0
peft 0.11.2.dev0 (current main)
transformers  4.41.2

### Who can help?

_No response_

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

```python
import torch
from peft import PeftModel, VeraConfig


device = "cuda"
dtype = torch.bfloat16 # or torch.float16

vera_config = VeraConfig(
    r=8,
    vera_dropout=0.05,
)


    "mistralai/Mistral-7B-v0.1", torch_dtype=dtype, device_map=device
)
model = PeftModel(model, vera_config)
model.set_adapter("default")

tokenizer = AutoTokenizer.from_pretrained("mistralai/Mistral-7B-v0.1")

model.eval()
tokenizer.padding_side = "left"

prompts = ["My mother", "My favourite", "I like to eat"]
model_inputs = tokenizer(prompts, return_tensors="pt", padding=True).to(device)

generated_ids = model.generate(**model_inputs, max_new_tokens=100, do_sample=True)
```

Error msg:
```
File "[REDACTED]/peft/tuners/vera/layer.py", line 267, in forward
    result = result + lambda_b * F.linear(lambda_d * F.linear(dropout(x), sliced_A), sliced_B)
RuntimeError: expected mat1 and mat2 to have the same dtype, but got: float != c10::Half
```
The error happens with either `torch.bfloat16` or `torch.float16`

Changing `device` to `torch.float32` doesn't have this problem.

A small patch in `peft/tuners/vera/layer.py`, casting `sliced_A` and `sliced_B` to the same dtype as `lambda_d` (float32) does workaround this problem:
```python
        previous_dtype = x.dtype

        if self.disable_adapters:
            if self.merged:
                self.unmerge()
        elif self.merged:
        else:
                if active_adapter not in self.vera_lambda_d.keys():
                    continue

                lambda_d = self.vera_lambda_d[active_adapter]
                lambda_b = self.vera_lambda_b[active_adapter]

                vera_A = self.vera_A[active_adapter]
                vera_B = self.vera_B[active_adapter]

                # As adapted layers may have different shapes and VeRA contains a single shared pair of A and B matrices,
                # we initialize these matrices with the largest required size for each dimension.
                # During the forward pass, required submatrices are sliced out from the shared vera_A and vera_B.
                sliced_A = vera_A[:, : self.in_features]
                sliced_B = vera_B[: self.out_features, :]
                
                dropout = self.vera_dropout[active_adapter]
                x = x.to(lambda_d.dtype)
                sliced_A = sliced_A.to(lambda_d.dtype)
                sliced_B = sliced_B.to(lambda_d.dtype)
                result = result + lambda_b * F.linear(lambda_d * F.linear(dropout(x), sliced_A), sliced_B)

        return result
```

### Expected behavior

It shouldn't break when using lower-precision models, though I only tested with `mistralai/Mistral-7B-v0.1`.

I'm not sure why `vera_A` and `vera_B` are casted to `bfloat16` to begin with, they are initialized as `float32`. I dug around but can't find where they are casted to `bfloat16`.

Also, the patch is probably suboptimal since if we have to cast them to `float32` in every forward pass instead of setting them as `float32` and keep it that way.

BenjaminBossan:
Resolves #1872

The issue was that we didn't consider `BufferDict` when auto-casting the adapter weights to float32 in PR #1706. This has now been addressed.

As #1706 was merged after the latest release, this bug should only affect users who install from main, so a patch release should not be needed.

As part of this PR, I also moved the `buffer_dict.py` up from `peft/tuners/vera` to `peft/tuners/` (with underscore to make it super clear that this is not for public usage). This is because we need to use it several times on a higher level than VeRA.


younesbelkada (approved):
Thanks a lot !

51616:
I confirm that running the script in #1872 using commit 5266ef3c4a4e2d5a47a80b02a5afecdd8514fb79 no longer have the error!

PR #1875, opened 2024-06-19
ENH: LoRA support for dynamically dispatching to custom layers

### LINKED ISSUE #1867: FEAT: LoRA: Possibility to dynamically register layers to dispatch
### Feature request

LoRA already implements a [dynamic dispatching approach](https://github.com/huggingface/peft/blob/076561bbd32ce8cd649ba4ec05cbc592cf2dd1c6/src/peft/tuners/lora/model.py#L290-L319) for different LoRA layer types, which for instance adopts based on which quantization library is available. Check if it would be possible to allow users to pass their own dispatching function to allow them to provide LoRA layers for types that are not supported by PEFT.

### Motivation

Right now, each time users want to target a new layer type with PEFT, we need to explicitly add support for it and release a new version of PEFT. This can be a long process and we generally don't want to support very niche layer types. However, it may not take much to allow users to easily register their own LoRA layers, so it's worth a try.

### Your contribution

I'll investigate this possibility once I have a bit of time on my hands.

BenjaminBossan:
Resolves #1867

## Description

This is an experimental feature with a private API for now. If this feature finds adoption, I will work on adding an official API.

With this PR, we allow users to register their own LoRA layer types. This way, they can add their own support for hitherto unsupported layer types, say `nn.Conv3d` or `nn.LSTM`. Without this PR, they can only do that by creating a PR on PEFT with support for this new type and getting it merged.

The custom dispatch mechanism also allows users to override existing layer type mapping. This way, they can, for instance, provide their own `lora.Linear` layer type, instead of using the one from PEFT, to adapt `nn.Linear` layers.

## Implementation

The implementation required only very few changes because we already have a mechanism for dynamic dispatching for LoRA. It is currently used, for instance, to dynamically add quantized target layers in case the right quantization library is installed.

This existing mechanism is now extended to include user provided LoRA layers if those were passed. These are checked first before checking the default PEFT supported layers.

## What's missing for this to become an official API?

Right now, the main reason why this cannot be an official API is the question of how to persist the config. In the current implementation, we add an attribute that is a mapping from target layer type to LoRA layer type:

`config._custom_modules == {CustomBaseLayer: CustomLoraLayer}`

The entries of this dict are Python classes. Therefore, they cannot be json-serialized. We could think of possible solutions how to serialize and deserialize custom Python objects, but this is not trivial and potentially a security risk. Thus I would only really start working on this if the demand is sufficiently high. At that point, I would also add a public API instead of requiring the use of a private API.

As is, users can still save and load PEFT models with custom LoRA layers, they only need to add two lines of code to their scripts, as documented.

We could also think about adding support for methods other than LoRA. However, this would require to implement the dynamic dispatch mechanism for those other methods, which right now only exists for LoRA.

### DIFF
--- a/docs/source/developer_guides/custom_models.md
+++ b/docs/source/developer_guides/custom_models.md
@@ -238,3 +238,64 @@ peft_model.print_trainable_parameters()
 ```python
 print(peft_model.targeted_module_names)
 ```
+
+## Unsupported module types
+
+In PEFT in general, methods like LoRA only work if the target modules are supported by PEFT. For example, it's possible to apply LoRA to `nn.Linear` and `nn.Conv2d` layers, but not, for instance, to `nn.LSTM`. If you find that the class of layer you want to apply PEFT to is not supported, feel free to open a [PEFT issue on GitHub](https://github.com/huggingface/peft/issues) and request the feature. Maintainers will implement this or guide you on how to implement this yourself if the demand for this module type is sufficiently high.
+
+### Experimental support for dynamic dispatch of custom modules in LoRA
+
+> [!WARNING]
+> This feature is experimental and subject to change, depending on its reception by the community. If we find that there is a big demand, we will introduce a public, stable API.
+
+For LoRA specifically, PEFT supports an experimental API for custom module types. Let's assume that you have a LoRA implementation for LSTMs. Normally, you would not be able to tell PEFT to use it, even if it would theoretically work with PEFT. With dynamic dispatch of custom layers, you can, however, make this work.
+
+The experimental API currently looks like this:
+
+```python
+class MyLoraLSTMLayer:
+    ...
+
+base_model = ...  # load the base model that uses LSTMs
+
+# add the LSTM layer names to target_modules
+config = LoraConfig(..., target_modules=["lstm"])
+# define a mapping from base layer type to LoRA layer type
+custom_module_mapping = {nn.LSTM: MyLoraLSTMLayer}
+# register the new mapping
+config._register_custom_module(custom_module_mapping)
+# after registration, create the PEFT model
+peft_model = get_peft_model(base_model, config)
+# do training
+```
+
+<Tip>
+When you call [`~get_peft_model`]`, you will see a warning because PEFT does not recognize the module type being targeted. In this case, you can ignore this warning.
+</Tip>
+
+By supplying a custom mapping, PEFT will take care of checking the layers of the base model against this custom mapping first and dispatching to the custom LoRA layer type if there is a match. Only after this will PEFT check the built-in LoRA layer types for a match. Therefore, this feature can also be used to override existing dispatch logic, e.g. if you want to use your own LoRA layer for `nn.Linear` instead of using the one provided by PEFT.
+
+When creating your custom LoRA module, please follow the same rules as the existing LoRA modules do. For this, check the [LoRA layer implementation](https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/layer.py). Notable constraints to consider:
+
+- The custom module should inherit from `nn.Module` and `peft.tuners.lora.layer.LoraLayer`
+- The learnable parameters should be stored in an `nn.ModuleDict` or `nn.ParameterDict`, where the key corresponds to the name of the specific adapter (remember that a model can have more than one adapter at a time).
+- Also, the name of these learnable parameter attributes should start with `"lora_"`, e.g. `self.lora_new_param = ...`.
+- Some methods are optional, e.g. you only need to implement `merge` and `unmerge` if you want to support weight merging.
+
+A caveat for this feature is that right now, the information about the custom module are not persisted when you save the model. Therefore, when loading the model, you have to register the custom modules again. Below we show how that works:
+
+```python
+# saving works as always and includes the parameters of the custom modules
+peft_model.save_pretrained(tmp_path / "lora-custom-module")
+
+# loading the model later:
+base_model = ...
+# create the config instance and register the custom module again, the same way as the first time
+config = LoraConfig(..., target_modules=["lstm"])
+custom_module_mapping = {nn.LSTM: MyLoraLSTMLayer}
+config._register_custom_module(custom_module_mapping)
+# pass the config instance to from_pretrained:
+peft_model = PeftModel.from_pretrained(model, tmp_path / "lora-custom-module", config=config)
+```
+
+If you use this feature and find it useful, or if you encountered some problems, let us know about it by creating on issue or a discussion on GitHub. This allows us to estimate the demand for this feature and add a public API if the demand is sufficiently high.
--- a/docs/source/developer_guides/custom_models.md
+++ b/docs/source/developer_guides/custom_models.md
@@ -277,7 +277,8 @@ By supplying a custom mapping, PEFT will take care of checking the layers of the
 
 When creating your custom LoRA module, please follow the same rules as the existing LoRA modules do. For this, check the [LoRA layer implementation](https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/layer.py). Notable constraints to consider:
 
-- The custom module should inherit from `nn.Module` and `peft.tuners.lora.layer.LoraLayer`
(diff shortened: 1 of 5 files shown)


docs/source/developer_guides/custom_models.md
```diff
@@ -238,3 +238,64 @@ peft_model.print_trainable_parameters()
...
+
+<Tip>
+When you call [`~get_peft_model`]`, you will see a warning because PEFT does not recognize the module type being targeted. In this case, you can ignore this warning.
+</Tip>
+
+By supplying a custom mapping, PEFT will take care of checking the layers of the base model against this custom mapping first and dispatching to the custom LoRA layer type if there is a match. Only after this will PEFT check the built-in LoRA layer types for a match. Therefore, this feature can also be used to override existing dispatch logic, e.g. if you want to use your own LoRA layer for `nn.Linear` instead of using the one provided by PEFT.
+
+When creating your custom LoRA module, please follow the same rules as the existing LoRA modules do. For this, check the [LoRA layer implementation](https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/layer.py). Notable constraints to consider:
+
+- The custom module should inherit from `nn.Module` and `peft.tuners.lora.layer.LoraLayer`
```
younesbelkada, line 280: here we should IMO state that the signature of the init method should have `base_layer` and  `adapter_name` in the correct order otherwise the API will fail

younesbelkada (approved):
Thanks a lot for this clean integration, docs and API ! Left only one nit but overall looks great !

docs/source/developer_guides/custom_models.md
BenjaminBossan: Good point, I added an entry for the `__init__`.

docs/source/developer_guides/custom_models.md
```diff
@@ -238,3 +238,66 @@ peft_model.print_trainable_parameters()
...
+## Unsupported module types
+
+In PEFT in general, methods like LoRA only work if the target modules are supported by PEFT. For example, it's possible to apply LoRA to `nn.Linear` and `nn.Conv2d` layers, but not, for instance, to `nn.LSTM`. If you find that the class of layer you want to apply PEFT to is not supported, feel free to open a [PEFT issue on GitHub](https://github.com/huggingface/peft/issues) and request the feature. Maintainers will implement this or guide you on how to implement this yourself if the demand for this module type is sufficiently high.
+
+### Experimental support for dynamic dispatch of custom modules in LoRA
+
+> [!WARNING]
+> This feature is experimental and subject to change, depending on its reception by the community. If we find that there is a big demand, we will introduce a public, stable API.
+
+For LoRA specifically, PEFT supports an experimental API for custom module types. Let's assume that you have a LoRA implementation for LSTMs. Normally, you would not be able to tell PEFT to use it, even if it would theoretically work with PEFT. With dynamic dispatch of custom layers, you can, however, make this work.
```
stevhliu, line 251:
```suggestion
PEFT supports an experimental API for custom module types for LoRA. Let's assume you have a LoRA implementation for LSTMs. Normally, you would not be able to tell PEFT to use it, even if it would theoretically work with PEFT. However, this is possible with dynamic dispatch of custom layers.
```

docs/source/developer_guides/custom_models.md
```diff
@@ -238,3 +238,66 @@ peft_model.print_trainable_parameters()
...
+# register the new mapping
+config._register_custom_module(custom_module_mapping)
+# after registration, create the PEFT model
+peft_model = get_peft_model(base_model, config)
+# do training
+```
+
+<Tip>
+When you call [`get_peft_model`], you will see a warning because PEFT does not recognize the module type being targeted. In this case, you can ignore this warning.
+</Tip>
```
stevhliu, line 274: Need blank lines after and before `<Tip>` to render the link inside properly :)

```suggestion
<Tip>

When you call [`get_peft_model`], you will see a warning because PEFT does not recognize the targeted module type. In this case, you can ignore this warning.

</Tip>
```

docs/source/developer_guides/custom_models.md
```diff
@@ -238,3 +238,66 @@ peft_model.print_trainable_parameters()
...
+
+By supplying a custom mapping, PEFT will take care of checking the layers of the base model against this custom mapping first and dispatching to the custom LoRA layer type if there is a match. Only after this will PEFT check the built-in LoRA layer types for a match. Therefore, this feature can also be used to override existing dispatch logic, e.g. if you want to use your own LoRA layer for `nn.Linear` instead of using the one provided by PEFT.
```
stevhliu, line 276:
```suggestion
By supplying a custom mapping, PEFT first checks the base model's layers against the custom mapping and dispatches to the custom LoRA layer type if there is a match. If there is no match, PEFT checks the built-in LoRA layer types for a match.

Therefore, this feature can also be used to override existing dispatch logic, e.g. if you want to use your own LoRA layer for `nn.Linear` instead of using the one provided by PEFT.
```

docs/source/developer_guides/custom_models.md
```diff
@@ -238,3 +238,66 @@ peft_model.print_trainable_parameters()
...
+
+When creating your custom LoRA module, please follow the same rules as the existing LoRA modules do. For this, check the [LoRA layer implementation](https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/layer.py). Notable constraints to consider:
```
stevhliu, line 278:
```suggestion
When creating your custom LoRA module, please follow the same rules as the [existing LoRA modules](https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/layer.py). Some important constraints to consider:
```

docs/source/developer_guides/custom_models.md
```diff
@@ -238,3 +238,66 @@ peft_model.print_trainable_parameters()
...
+
+- The custom module should inherit from `nn.Module` and `peft.tuners.lora.layer.LoraLayer`.
+- The `__init__` method of the custom module should have the positional arguments `base_layer` and `adapter_name`. After this, there are additional `**kwargs` that you are free to use or ignore.
+- The learnable parameters should be stored in an `nn.ModuleDict` or `nn.ParameterDict`, where the key corresponds to the name of the specific adapter (remember that a model can have more than one adapter at a time).
+- Also, the name of these learnable parameter attributes should start with `"lora_"`, e.g. `self.lora_new_param = ...`.
```
stevhliu, line 283:
```suggestion
- The name of these learnable parameter attributes should start with `"lora_"`, e.g. `self.lora_new_param = ...`.
```

docs/source/developer_guides/custom_models.md
```diff
@@ -238,3 +238,66 @@ peft_model.print_trainable_parameters()
...
+- Some methods are optional, e.g. you only need to implement `merge` and `unmerge` if you want to support weight merging.
+
+A caveat for this feature is that right now, the information about the custom module is not persisted when you save the model. Therefore, when loading the model, you have to register the custom modules again. Below we show how that works:
```
stevhliu, line 286:
```suggestion
Currently, the information about the custom module does not persist when you save the model. When loading the model, you have to register the custom modules again.
```

docs/source/developer_guides/custom_models.md
```diff
@@ -238,3 +238,66 @@ peft_model.print_trainable_parameters()
...
+# load the LoRA config that you saved earlier
+config = LoraConfig.from_pretrained(<model-path>)
+# register the custom module again, the same way as the first time
+custom_module_mapping = {nn.LSTM: MyLoraLSTMLayer}
+config._register_custom_module(custom_module_mapping)
+# pass the config instance to from_pretrained:
+peft_model = PeftModel.from_pretrained(model, tmp_path / "lora-custom-module", config=config)
+```
+
+If you use this feature and find it useful, or if you encountered some problems, let us know about it by creating on issue or a discussion on GitHub. This allows us to estimate the demand for this feature and add a public API if it is sufficiently high.
```
stevhliu, line 303:
```suggestion
If you use this feature and find it useful, or if you encounter problems, let us know by creating an issue or a discussion on GitHub. This allows us to estimate the demand for this feature and add a public API if it is sufficiently high.
```

stevhliu (approved):
Very cool! Left a comment about improving the introductory section a bit 😄

PR #1879, opened 2024-06-21
Fix early import of torch extension in BOFT

### LINKED ISSUE #1877: just import of peft leads to cuda context initialization
### System Info

PyTorch version: 2.2.1
peft version: 0.11.1

### Reproduction

The following snippet can trigger this bug:

```python
import multiprocessing
import peft
from torch import nn

def func():
    m = nn.Linear(2, 3).cuda(0)

if __name__ == "__main__":
    proc = multiprocessing.Process(target=func)
    proc.start(), proc.join()
```

Raise:
RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method

Issue: early import ``torch.utils.cpp_extension`` in ``peft/tuners/boft/layer.py``

How to fix: import ``cpp_extension`` module in the ``get_fbd_cuda`` function.

### Expected behavior

No RuntimeError. CUDA context is not initialized in the main process.

PhyscalX:
# Description

This PR resolves #1877.  

The proposed change simply moves the import  of ``torch.utils.cpp_extension`` into the function ``get_fbd_cuda``.

### DIFF
--- a/src/peft/tuners/boft/layer.py
+++ b/src/peft/tuners/boft/layer.py
@@ -77,6 +77,7 @@ def get_fbd_cuda():
     if _FBD_CUDA is not None:
         return _FBD_CUDA
 
+    # This import initializes cuda context and should thus be local, see issue 1877
     from torch.utils.cpp_extension import load
 
     curr_dir = os.path.dirname(__file__)
--- a/src/peft/tuners/boft/layer.py
+++ b/src/peft/tuners/boft/layer.py
@@ -27,7 +27,6 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 from torch.autograd import Function
-from torch.utils.cpp_extension import load
 
 from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
 
@@ -77,6 +76,8 @@ def get_fbd_cuda():
 
     if _FBD_CUDA is not None:
         return _FBD_CUDA
+
+    from torch.utils.cpp_extension import load
 
     curr_dir = os.path.dirname(__file__)
     # need ninja to build the extension


src/peft/tuners/boft/layer.py
```diff
@@ -78,6 +77,8 @@ def get_fbd_cuda():
     if _FBD_CUDA is not None:
         return _FBD_CUDA
 
+    from torch.utils.cpp_extension import load
```
BenjaminBossan, line 80: Let's add a comment: `# This import initializes cuda context and should thus be local, see issue 1877`

PhyscalX:
> Nice, thanks a lot for identifying the issue and fixing it.
> 
> I wonder: Would it make sense to add the code snippet you provided in the issue as a unit test to check for this in the future? Let me know what you think.

Yes, it makes sense. CUDA context initialization issue cloud be easily introduced by third-party libraries or JIT codes.

BenjaminBossan:
> Yes, it makes sense. CUDA context initialization issue could be easily reintroduced by third-party libraries or JIT codes.

I agree. Would you please add a new test to `tests/test_gpu_examples.py` that is based on the snippet you provided in the issue, with a short description of what this tests. That would be a great way to prevent regressions in the future.

PhyscalX:
> > Yes, it makes sense. CUDA context initialization issue could be easily reintroduced by third-party libraries or JIT codes.
> 
> I agree. Would you please add a new test to `tests/test_gpu_examples.py` that is based on the snippet you provided in the issue, with a short description of what this tests. That would be a great way to prevent regressions in the future.

This test can NOT be placed in ```tests/test_gpu_examples.py``, because ``require_torch_gpu`` will initialize CUDA context if available. As a result, my snippet will always fail.

BenjaminBossan:
> This test can NOT be placed in `tests/test_gpu_examples.py`, because `require_torch_gpu` will initialize CUDA context if available. As a result, my snippet will always fail.

Ah yes, that won't work. We would have to ensure in the tests that this specific test is started in a completely separate process. I'm not sure there is an easy solution for that, so for now we can just go ahead with this PR without a test.

PR #1884, opened 2024-06-24
FIX Init AdaLoRA to be identity transform

### LINKED ISSUE #1798: High loss when init with `AdaLora`
### System Info

torch==2.0.1
transformers==4.40.2
flash-attn==2.3.3
peft==0.11.1
accelerate==0.28.0
deepspeed==0.14.0
bitsandbytes==0.43.0
datasets==2.18.0
trl==0.8.6

python==3.10.10

### Who can help?

@BenjaminBossan 

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

run.py:
```python
import peft
from peft import AdaLoraConfig, get_peft_model

model = AutoModelForCausalLM.from_pretrained("meta-llama/Llama-2-7b-hf", device_map="cuda:0")
tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-2-7b-hf", verbose=False)
text = """Fine-tuning large pretrained models is often prohibitively costly due to their scale. Parameter-Efficient Fine-Tuning (PEFT) methods enable efficient adaptation of large pretrained models to various downstream applications by only fine-tuning a small number of (extra) model parameters instead of all the model's parameters. This significantly decreases the computational and storage costs. Recent state-of-the-art PEFT techniques achieve performance comparable to fully fine-tuned models.

PEFT is integrated with Transformers for easy model training and inference, Diffusers for conveniently managing different adapters, and Accelerate for distributed training and inference for really big models."""
inputs = tokenizer(text, return_tensors="pt").to("cuda:0")

# print loss for original model
outputs = model(**inputs, labels=inputs["input_ids"])
loss = outputs.loss
print(loss.item())

# print loss for adalora
    target_r=8,
    init_r=12,
    lora_alpha=8,
    task_type=peft.utils.peft_types.TaskType.CAUSAL_LM
)

model = get_peft_model(model, adalora_config, adapter_name="adalora")
outputs = model(**inputs, labels=inputs["input_ids"])
loss = outputs.loss
print(loss.item())
``` 
output:
```
2.1954474449157715
3.3121755123138428
``` 

### Expected behavior

The two losses should be the same. The diagonal elements of the AdaLoRA initialization should be all 0s, and therefore should not change the results. [paper](https://arxiv.org/html/2303.10512#:~:text=%CE%9B-,is%20initialized%20with%20zero,-while)

### LINKED ISSUE #1836: Different results when merging LORA weights into the base model vs. when not
### System Info

peft, transformers, pytorch, python (3.8)

### Who can help?

@BenjaminBossan 

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

I modified Nvidia's BERT fine-tuning example from [here](https://github.com/NVIDIA/DeepLearningExamples/blob/master/PyTorch/LanguageModeling/BERT/file_utils.py) by adding support for distributed data parallel during finetuning, using Pytorch DDP. 
My code is here:
https://github.com/ankur6ue/distributed_training/tree/master

I'm now trying to add LoRA, which I did as follows.

```python
# load base BERT model:
model = modeling.BertForQuestionAnswering(config)
# Create peft config
peft_config = LoraConfig(
  task_type=TaskType.QUESTION_ANS,
  inference_mode=False,
  r=16,
  lora_alpha=16,
  lora_dropout=0.1,
  bias="all", # will train the bias matrices.
  target_modules=["key", "query", "value"],
  modules_to_save=["qa_outputs"],
)
# Load model from checkpoint and apply LoRA
model.load_state_dict(checkpoint, strict=False)
lora_model = get_peft_model(model, peft_config)
lora_model.print_trainable_parameters()
lora_model.to(device)

# Run training loop. In the forward pass, I'm using the original model (which seems to have been modified in place), rather than the lora_model. If I call the below on the lora_model, I get an error. 
start_logits, end_logits = model(input_ids, segment_ids, input_mask)
```

The training seems to run ok, and I get nearly the same final value of loss as I do without using LoRA.
Now to save the LoRA weights, I can either save them separately like this:
```python
 lora_model.save_pretrained('results/bert_small_lora.bin')
```

or by merging into the base model:
```python
 merged_lora_model = lora_model.merge_and_unload()
output_model_file = os.path.join(args.output_dir, modeling.WEIGHTS_NAME)
torch.save({"model": merged_lora_model.state_dict()}, output_model_file)
```

Now if I run evaluation using the merged model, I get the expected F1 scores. However if I use the saved adaptor weights as follows:
```python
model = modeling.BertForQuestionAnswering(config)
peft_model = PeftModel.from_pretrained(model_, lora_model_location, is_trainable=False)
peft_model.cuda(0)
```
Run evaluation on the model
Then I get terrible results, indicating the LoRA weights are somehow not being applied.. 

What is the right way to do this?? There are so many questions about how to properly save and apply LoRA adapters, that it would be nice to have a proper example that explains how to do it comprehensively (without using the trainer, because that hides too much complexity and only seems to work on hugginsface models.. and when it doesn't work, it is next to impossible to debug..)

### Expected behavior

The evaluation output should be identical for when LoRA weights are merged with the main model vs. when they are not..

BenjaminBossan:
Resolves #1798 

There was an accidental change in a previous PR that initialized `lora_E` as normal, when it should be zeros. See the linked issue for context.

### DIFF
--- a/src/peft/tuners/adalora/layer.py
+++ b/src/peft/tuners/adalora/layer.py
@@ -78,7 +78,7 @@ class AdaLoraLayer(LoraLayer):
 
     def reset_lora_parameters(self, adapter_name):
         if adapter_name in self.lora_A.keys():
-            nn.init.normal_(self.lora_E[adapter_name], mean=0.0, std=0.02)
+            nn.init.zeros_(self.lora_E[adapter_name])
             nn.init.normal_(self.lora_A[adapter_name], mean=0.0, std=0.02)
             nn.init.normal_(self.lora_B[adapter_name], mean=0.0, std=0.02)
 
--- a/tests/test_decoder_models.py
+++ b/tests/test_decoder_models.py
@@ -169,6 +169,7 @@ class PeftDecoderModelTester(unittest.TestCase, PeftCommonTester):
             {
                 "model_ids": PEFT_DECODER_MODELS_TO_TEST,
                 "lora_kwargs": {"init_lora_weights": [False]},
+                "adalora_kwargs": {"init_lora_weights": [False]},
                 "ia3_kwargs": {"init_ia3_weights": [False]},
                 "boft_kwargs": {"init_weights": [False]},
                 "vera_kwargs": {"init_weights": [False]},
--- a/tests/test_encoder_decoder_models.py
+++ b/tests/test_encoder_decoder_models.py
@@ -91,6 +91,7 @@ class PeftEncoderDecoderModelTester(unittest.TestCase, PeftCommonTester):
             {
                 "model_ids": PEFT_ENCODER_DECODER_MODELS_TO_TEST,
                 "lora_kwargs": {"init_lora_weights": [False]},
+                "adalora_kwargs": {"init_lora_weights": [False]},
                 "ia3_kwargs": {"init_ia3_weights": [False]},
                 "vera_kwargs": {"init_weights": [False]},
                 "task_type": "SEQ_2_SEQ_LM",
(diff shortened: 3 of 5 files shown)


SunMarc (approved):
Make sense ! LGTM ! Thanks for fixing this and adding some tests !

Issue #896, opened 2023-09-01
Increase speed of adapter layer initialization


BenjaminBossan:
### Feature request

We are working on increasing the speed of adapter layer initialization. The first step towards this is #887, which tackles linear LoRA layers, which are arguably the most important ones. As discussed with @pacman100, after that PR, there are still some TODOs:

- [x] apply to Embedding and Conv2d layers (#915)
- [x] apply to compressed layer types, like Linear4bit, Linear8bit (#994)
- [ ] init LoRA modules (A, B) right on target device, not CPU (won't fix for the time being, seems pretty fast already with all the other changes)
- [x] same for AdaLora (#1106)
- [x] same for IA³ (#1106)
- [x] Init in `merge_and_unload` (#1106)

### Motivation

Initial testing showed ~10x increase in speed with the optimized initialization.

### Your contribution

#887 is the first step.

BenjaminBossan:
Not stale

osmalpkoras:
Hi guys, I am trying to load a `meta-llama/Meta-Llama-3-70B-Instruct` PEFT model with `LoraConfig(init_lora_weights="pissa", 16, 0.1, 64, none", CAUSAL_LM", target_modules="all-linear")` and I noticed that it is very slow in initializing the layers. To be precise, it's the `for key in key_list` loop in `peft.tuners.tuners_utils.BaseTuner.inject_adapter` that is very slow. 

Should this issue be actually fixed with these PR's? Because if it should be, than it does not work for `meta-llama/Meta-Llama-3-70B-Instruct` with the above LoraConfig.

BenjaminBossan:
Thanks for reporting this. Is it still slow when you remove `init_lora_weights="pissa"` from the config?

osmalpkoras:
Actually if I use the default method, it is very fast. Switching to `init_lora_weights="pissa_niter_32"` also made it very fast. I just notice that if I use this initialization with a multi-gpu deepspeed setup, it again takes a couple of hours, maybe due to some concurrency issues.

BenjaminBossan:
It is expected that using the default init or PiSSA with `n_iter` is faster than just `pissa`. However, a couple of hours still sounds excessive. Does it finish after a couple of hours or do you just cancel the process at that point?

PR #1887, opened 2024-06-25
FIX Make special LoRA inits DeepSpeed compatible

BenjaminBossan:
Resolves https://github.com/huggingface/accelerate/issues/2886

Possibly resolves
https://github.com/huggingface/peft/issues/896#issuecomment-2172455458

Some LoRA init methods need to access the base layer weight. Getting this access can fail or stall in distributed settings. For DeepSpeed, the weight is now gathered before trying to access it.

Note: Without DeepSpeed, this is a no-op and should thus not have any disadvantage. We don't have DS in our CI, so this is not tested.

I also made some small changes to OLoRA init to use `self.get_base_layer()` instead of `self.base_layer`.

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -24,7 +24,7 @@ from torch import svd_lowrank
 from transformers.pytorch_utils import Conv1D
 
 from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
-from peft.utils.integrations import dequantize_module_weight
+from peft.utils.integrations import dequantize_module_weight, gather_params_ctx
 from peft.utils.other import transpose
 
 from .config import LoraConfig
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
ping @tokenizer-decode as there are some slight changes to OLoRA.

SunMarc (approved):
LGTM ! Maybe we can leave a comment on why we added that for these inits since we don't do it for other inits

tokenizer-decode:
Could you explain to me why `self.get_base_layer()` rather than `self.base_layer`?  No mutation to base weights occured at that part so I'd assume both should return the same? How did this solve the problem? I am surprised.

BenjaminBossan:
> Could you explain to me why `self.get_base_layer()` rather than `self.base_layer`? No mutation to base weights occured at that part so I'd assume both should return the same?

The reason why we have `get_base_layer()` is that we support nesting LoRA adapters and other adapter types. Calling the method ensures we really get back the "original" base layer. When there is no nesting, both approaches are indeed the same.

> How did this solve the problem? I am surprised.

This change is not related to the DeepSpeed issue, that's what I wanted to convey when I wrote "I also made some small changes to OLoRA".

tokenizer-decode:
Oh now I see. Thanks for notifying. Looks good.

PR #1888, opened 2024-06-26
bypass print_trainable_parameter() if model is not peft model

delock:
The function call `print_trainable_parameter()` in `examples/sft/train.py` didn't take account of `--use_peft_lora False` and finetune will fail in such case.   This PR bypass this print if this function cannot be found in the model.


BenjaminBossan (approved):
Thanks for fixing this issue in the training script. Failing CI is unrelated, so I'll merge despite of it.

PR #1892, opened 2024-06-28
FIX Recursion while accessing attribute before initialization

### LINKED ISSUE #1349: 'PeftModelForCausalLM' object has no attribute 'base_model' for a deepspeed stage 3 model
### System Info

```
peft==0.7.1
transformers==4.36.2
deepspeed==0.12.4
```

### Who can help?

_No response_

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

Hello! I'm facing the issue with `deepspeed==0.12.4`, stage3, `transformers==4.36.2` and `peft==0.7.1`:
```
AttributeError: 'PeftModelForCausalLM' object has no attribute 
'_ds_child_entered'
....
....
  File ".../site-packages/peft/peft_model.py", line 528, in __getattr__
    return super().__getattr__(name)  # defer to nn.Module's logic
  File ".../torch/nn/modules/module.py", line 1695, in __getattr__
    raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
AttributeError: 'PeftModelForCausalLM' object has no attribute 'base_model'
``` 
and eventually
```
RecursionError: maximum recursion depth exceeded while calling a Python object
```

I'm using `pytorch lightning`:
```python
trainer = Trainer(
    ...
    strategy = DeepSpeedStrategy(stage=3)
)

class Module(LightningModule):
    def configure_model(self) -> None:
        deepspeed_config = self.trainer.strategy.config
        self.dschf = HfDeepSpeedConfig(deepspeed_config)
        model = get_peft_model(
            model,
            LoraConfig(
                task_type=TaskType.CAUSAL_LM,
                inference_mode=False,
                target_modules=target_modules,
                r=48,
                lora_alpha=16,
                lora_dropout=0.0,
            ),
        )

```

When deepspeed 3 is activated, transformers use some custom hooks to init the model differently. Apparently, that doesn't go well with `peft` assumptions.

### Expected behaviour

get_peft_model succeeds without an error

ret-1:
Resolves #1349
While using deepspeed, it will try to access attribute before the `__init__` method (which is wrapped by the following wrapper):

https://github.com/microsoft/DeepSpeed/blob/f0e3f01d7c7a3d8748212e61eaf487fab41168a7/deepspeed/runtime/zero/partition_parameters.py#L481-L522

However, at this time there is no `base_model` (or `model`) attribute, which would cause recursion in `__getattr__` method. So I make some modifications to the `__getattr__` method to fix this problem.

### DIFF
--- a/examples/feature_extraction/peft_lora_embedding_semantic_search.py
+++ b/examples/feature_extraction/peft_lora_embedding_semantic_search.py
@@ -194,7 +194,7 @@ class AutoModelForSentenceEmbedding(nn.Module):
         try:
             return super().__getattr__(name)  # defer to nn.Module's logic
         except AttributeError:
-            if name == 'model':  # see #1892: prevent infinite recursion if class is not initialized
+            if name == "model":  # see #1892: prevent infinite recursion if class is not initialized
                 raise
             return getattr(self.model, name)
 
--- a/examples/feature_extraction/peft_lora_embedding_semantic_search.py
+++ b/examples/feature_extraction/peft_lora_embedding_semantic_search.py
@@ -194,6 +194,8 @@ class AutoModelForSentenceEmbedding(nn.Module):
         try:
             return super().__getattr__(name)  # defer to nn.Module's logic
         except AttributeError:
+            if name == 'model':
+                raise
             return getattr(self.model, name)
 
 
--- a/examples/feature_extraction/peft_lora_embedding_semantic_search.py
+++ b/examples/feature_extraction/peft_lora_embedding_semantic_search.py
@@ -194,7 +194,7 @@ class AutoModelForSentenceEmbedding(nn.Module):
         try:
             return super().__getattr__(name)  # defer to nn.Module's logic
         except AttributeError:
-            if name == 'model':
+            if name == 'model':  # see #1892: prevent infinite recursion if class is not initialized
                 raise
             return getattr(self.model, name)
 
--- a/tests/test_initialization.py
+++ b/tests/test_initialization.py
@@ -21,6 +21,9 @@ from scipy import stats
 from torch import nn
 
 from peft import AdaLoraConfig, LoraConfig, PeftModel, PromptTuningConfig, VeraConfig, get_peft_model
+from peft import AdaLoraModel, AdaptionPromptModel, BOFTModel, IA3Model, LNTuningModel, LoraModel, PeftMixedModel, PolyModel, VeraModel
+from peft.tuners import MixedModel
+from peft.tuners.lycoris_utils import LycorisTuner
 from peft.utils import infer_device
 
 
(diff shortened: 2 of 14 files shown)


ret-1:
```python
# a simple code to show what happened
class PeftModel:
    def __init__(self, model):
        self.base_model = model
    
    def __getattr__(self, name):
        try:
            return super().__getattr__(name)
        except AttributeError:
            # recursion error
            return getattr(self.base_model, name)
        
import functools
def d(f):
    # this decorator is used in deepspeed
    @functools.wraps(f)
    def wrapper(module, *args, **kwargs):
        if not hasattr(module, 'abc'):
            print('abc')
        f(module, *args, **kwargs)
    return wrapper

class PeftModelForCausalLM(PeftModel):
    @d
    def __init__(self, model):
        super().__init__(model)

a = PeftModelForCausalLM(1)
```

examples/feature_extraction/peft_lora_embedding_semantic_search.py
```diff
@@ -194,6 +194,8 @@ def __getattr__(self, name: str):
         try:
             return super().__getattr__(name)  # defer to nn.Module's logic
         except AttributeError:
+            if name == 'model':
```
BenjaminBossan, line 197: Could you add a comment here, and for all the other changes as well, along the lines of:

`# see #1892: prevent infinite recursion if class is not initialized`

BenjaminBossan (changes requested):
Great investigation and thanks for providing the fix. In general, this looks good. What would be great if we could unit tests for this as well, which shouldn't be too hard, given the example you already provided.

Here is a suggestion for a test class that could be placed inside of `tests/test_initialization.py`:

```python
class TestNoInfiniteRecursionDeepspeed:
    # see #1892 for details
    classes = [
        PeftModel,
        # other classes
    ]

    @pytest.fixture
    def wrap_init(self):
        # emulates the wrapper from DeepSpeed
        import functools

        def decorator(f):
            @functools.wraps(f)
            def wrapper(self, *args, **kwargs):
                hasattr(self, "abc")  # any hasattr will do
                f(self, *args, **kwargs)
            return wrapper

        return decorator

    @pytest.fixture
    def model(self):
        class MyModule(nn.Module):
            def __init__(self):
                super().__init__()
                self.linear = nn.Linear(10, 10)

        return MyModule()

    @pytest.mark.parametrize("cls", classes)
    def test_no_infinite_recursion(self, cls, model, wrap_init):
        original_init = cls.__init__
        try:
            cls.__init__ = wrap_init(cls.__init__)
            # this would trigger an infinite loop before the fix in 1892
            cls(model, LoraConfig(target_modules=["linear"]))
        finally:
            # ensure there are no side effects of this test
            cls.__init__ = original_init
```

WDYT?

ret-1:
Hi @BenjaminBossan ,

Thank you for your feedback! I'm not very familiar with `pytest`, but after learning I think your suggestion is good, so I just added classes into the list (not sure whether to add `AutoModelForSentenceEmbedding` in `examples/feature_extraction/peft_lora_embedding_semantic_search.py`).

Please let me know if there's anything else I need to do : )

tests/test_initialization.py
```diff
@@ -601,3 +604,54 @@ def test_vera_mixing_save_projection_raises(self):
...
+        AdaptionPromptModel,
+        BOFTModel,
+        IA3Model,
+        LNTuningModel,
+        LoraModel,
+        PeftMixedModel,
+        PolyModel,
+        VeraModel,
+        MixedModel, 
+        LycorisTuner
```
BenjaminBossan, line 622: This won't quite work, we cannot initialize `LoraModel` etc. just like this. But I don't think we need to test this, as users typically don't need to do that by hand anyway. Instead, let's only test the classes that should be initialized directly be the user, `PeftModel` etc.

```python
    classes = [
        PeftModel,
        PeftMixedModel,
        PeftModelForSequenceClassification,
        PeftModelForQuestionAnswering,
        PeftModelForTokenClassification,
        PeftModelForCausalLM,
        PeftModelForSeq2SeqLM,
        PeftModelForFeatureExtraction,
    ]
```

(imports need to be adjusted accordingly)

I also noticed that `MyModule` needs to be changed slightly so that it "imitates" a language model:

```python
        class MyModule(nn.Module):
            def __init__(self):
                super().__init__()
                self.linear = nn.Linear(10, 10)
                # to emulate LMs:
                self.prepare_inputs_for_generation = None
                self._prepare_encoder_decoder_kwargs_for_generation = None
```

BenjaminBossan (changes requested):
Thanks for the updates. There is still a bit of an issue with the test, but it should be easy to fix, please check my comment. You can ensure that the test passes locally by running `pytest tests/test_initialization.py -k TestNoInfiniteRecursionDeepspeed`.

Furthermore, could you please merge with/rebase on the latest `main` branch, and then run `make style`?

ret-1:
Hi @BenjaminBossan ,

Thanks for your guidance! I made some updates and passed the unit tests. `make style` is also run. : )

BenjaminBossan:
Thanks for the updates

> `make style` is also run. : )

This does not seem to have worked correctly. Could you please try to run this manually:

```
ruff --version  # should be 0.4.10
ruff format src tests examples docs scripts docker
```

ret-1:
I think it's OK now!

PR #1897, opened 2024-07-01
FIX TEST: Higher tolerance for AdaLoRA in test

BenjaminBossan:
The test is flaky on CI, so this PR increases the tolerance to hopefully fix the flakines. I cannot reproduce the error locally (neither on GPU nor CPU), so I'm not 100% sure if this tolerance is enough to make the test reliable.


BenjaminBossan:
> If this persists, maybe we can think about adding the [is_flaky](https://github.com/huggingface/transformers/blob/e65502951593a76844e872fee9c56b805598538a/src/transformers/testing_utils.py#L2176) decorator, just like in transformers.

Let's hope this fixes it. Using this decorator could be a last resort if we don't find a better way.

PR #1898, opened 2024-07-01
FIX TEST Even higher tolerance for AdaLoRA in test

BenjaminBossan:
See #1897 for more context. The test is still flaky, increasing tolerance further.

Example of a failed run:

https://github.com/huggingface/peft/actions/runs/9745363462/job/26893196728

No matter how often I run this locally, I can't reproduce :-/


SunMarc (approved):
LGTM ! Hopefully, we don't need to increase the tolerance even more ! Is it possible to ssh in the runner to reproduce the error ?

BenjaminBossan:
> Is it possible to ssh in the runner to reproduce the error ?

Yes, I think Younes mentioned the possibility, but it requires updating the workflows. I'll take a look if the problem persists after this PR.

PR #1900, opened 2024-07-02
FIX: More VeRA tests, fix tests, more checks

BenjaminBossan:
- Fixes incorrect config for VeRA in a test
- Add VeRA to multi-adapter tests
- Add more checks on the VeRA A/B shapes

The latter becomes necessary when we add more than one VeRA adapter. The shapes for VeRA A and B are only determined once, when the first VeRA adapter is created. After that, they are fixed. However, users may add a second VeRA adapter. As long as that adapter targets the same layers and has the same rank, we're good. But if it targets other, bigger layers, or if it has increased rank, the shapes of VeRA A and/or VeRA B will be too small, resulting in an error during the forward pass. To prevent this, we already check the shapes during initialization of the new adapter and raise an error right away.


BenjaminBossan:
Could you please review @dkopi?

src/peft/tuners/vera/layer.py
```diff
@@ -100,6 +100,23 @@ def update_layer(
             # we can take any of the existing adapter's parameters, as they should all be identical
             vera_A_param = list(self.vera_A.values())[0]
             vera_B_param = list(self.vera_B.values())[0]
+
+            error_tmpl = (
+                "{} has a size of {} but {} is or greater required; this probably happened because an additional VeRA "
```
dkopi, line 105: a typo - should be "{} has a size of {} but {} or greater is required [...]", right?

tests/test_initialization.py
```diff
@@ -601,21 +601,60 @@ def test_use_prompt_tuning_init_text_raises(self):
...
         with pytest.raises(ValueError, match=msg):
             model.add_adapter("other", config1)
 
+    def test_vera_add_second_adapter_with_incompatible_input_shape(self):
+        config0 = VeraConfig(target_modules=["lin0"], r=8)
+        config1 = VeraConfig(target_modules=["lin1"])
+
+        model = get_peft_model(self.get_model(), config0)
+        # not full message but enough to identify the error
+        msg = "vera_A has a size of 10 but 20 is or greater required"
```
dkopi, line 643: same here

src/peft/tuners/vera/layer.py
dkopi: we could also have separate messages for incompatible r dimension or input/output dimensions;
can also give a hint that adding adapters in different order *may* resolve the problem

tests/test_initialization.py
```diff
@@ -601,21 +601,60 @@ def test_use_prompt_tuning_init_text_raises(self):
...
+            model.add_adapter("other", config1)
+
+    def test_vera_add_second_adapter_with_higher_rank(self):
+        config0 = VeraConfig(target_modules=["lin0"], r=123)
+        # second adapter has higher rank
+        config1 = VeraConfig(target_modules=["lin0"], r=456)
+
+        model = get_peft_model(self.get_model(), config0)
+        # not full message but enough to identify the error
+        msg = "vera_A has a size of 123 but 456 is or greater required"
```
dkopi, line 654: same here

dkopi (changes requested):
looks good, I've only added few comments regarding the raised message;

to easily fix this problem in vera I think the simplest solution would be to add optional config arguments that allow to specify the shared A/B shapes manually;
the other way would be to create separate "shared A/B" matrices per each adapter

BenjaminBossan:
@dkopi Thanks for the helpful comments, please check if I have addressed them sufficiently.

Also a thought that I had:

With your recent fix to the shape issue, we should be able to allow some models that formerly didn't work, right?

https://github.com/huggingface/peft/blob/09358aad308604f7a132cf94709bcc9399a2e1ab/src/peft/utils/constants.py#L200-L207

I cannot test at the moment, but I think we can uncomment them.

dkopi (approved):

BenjaminBossan:
@matthewdouglas LMK if you feel ready for reviewing this PR.

tests/test_custom_models.py
```diff
@@ -464,6 +464,16 @@
...
     ),
+    # Note: Currently, we cannot target lin0 and lin1 with different adapters when using VeRA. The reason is that the
+    # first adapter being created will result in a vera_A or vera_B shape that is too small for the next adapter
+    # (remember that VeRA shares these parameters across all layers), which results in an error.
+    (
+        "VeRA Same",
+        "vera",
+        VeraConfig,
+        {"target_modules": ["lin0"], "init_weights": False},
+        {"target_modules": ["lin0"], "init_weights": False},
```
sayakpaul, line 475: So essentially, these two adapters ("VeRA Same" and "vera") are the same. Right?

tests/test_initialization.py
```diff
@@ -601,21 +601,60 @@ def test_use_prompt_tuning_init_text_raises(self):
...
+            model.add_adapter("other", config1)
+
+    def test_vera_add_second_adapter_with_higher_rank(self):
+        config0 = VeraConfig(target_modules=["lin0"], r=123)
+        # second adapter has higher rank
+        config1 = VeraConfig(target_modules=["lin0"], r=456)
+
+        model = get_peft_model(self.get_model(), config0)
+        # not full message but enough to identify the error
+        msg = "vera_A has a size of 123 but 456 or greater is required"
```
sayakpaul, line 654: Can we programmatically determine 123 and 456?

tests/test_initialization.py
```diff
@@ -601,21 +601,60 @@ def test_use_prompt_tuning_init_text_raises(self):
...
         with pytest.raises(ValueError, match=msg):
             model.add_adapter("other", config1)
 
+    def test_vera_add_second_adapter_with_incompatible_input_shape(self):
+        config0 = VeraConfig(target_modules=["lin0"], r=8)
+        config1 = VeraConfig(target_modules=["lin1"])
+
+        model = get_peft_model(self.get_model(), config0)
+        # not full message but enough to identify the error
+        msg = "vera_A has a size of 10 but 20 or greater is required"
```
sayakpaul, line 643: Same. Can we programmatically determine the magic numbers here?

sayakpaul (approved):
Thanks. Very minor comments.

tests/test_custom_models.py
BenjaminBossan: "VeRA Same" is just the name of the test, "vera" is the name of the method being used. I added a comment at the top that describes the items of this tuple for clarification.

The two adapters here have the same config, but they are two different adapters.

tests/test_initialization.py
BenjaminBossan: Changed the code to make it obvious where the numbers come from.

tests/test_initialization.py
BenjaminBossan: Changed the code to make it obvious where the numbers come from.

BenjaminBossan:
@sayakpaul Your points should be addressed, LMK if you want to do another review or if the PR can be merged.

sayakpaul:
Thanks for clarifying the comments.

PR #1901, opened 2024-07-02
support Grouped-Query Attention

ttw1018:
To support Grouped-Query Attention, change `token_dim` to `token_dim // num_attention_heads * num_key_value_head` and `num_attention_heads` to `num_key_value_head`.

PR #1906, opened 2024-07-04
[WIP] ENH Add support for Qwen2

BenjaminBossan:
Add Qwen2 to default target modules, use tiny Qwen2 in tests.

~~Some tests are still be failing, waiting for #1901.~~ Done.


tests/testing_common.py
```diff
@@ -1419,6 +1419,9 @@ def _test_weighted_combination_of_adapters(self, model_id, config_cls, config_kw
         if issubclass(config_cls, AdaLoraConfig):
             # AdaLora does not support adding more than 1 adapter
             return pytest.skip(f"Test not applicable for {config_cls}")
+        if model_id.endswith("qwen2"):
```
sayakpaul, line 1422: Is `model_id` a sufficiently reliable way to detect if the model is `qwen2` type? Could we not look into its config and infer it from there?

sayakpaul (commented):
Overall looks harmless to me. Just a single comment.

tests/testing_common.py
sayakpaul: Make sense. I will respect your judgement then.

sayakpaul (approved):
Sorry for forgetting to approve.

PR #1908, opened 2024-07-05
FIX: Flaky multitask prompt tuning test fixed by setting the seed

BenjaminBossan:
Set the seed for test
`test_generate_text_with_other_init_2_ONLY_SOURCE_SHARED` because otherwise it is flaky and fails with ~5% probability.

I'm not 100% sure where this flakiness comes from but have documented my investigation in a comment in the test.

### DIFF
--- a/tests/test_multitask_prompt_tuning.py
+++ b/tests/test_multitask_prompt_tuning.py
@@ -273,6 +273,12 @@ class MultiTaskPromptTuningTester(TestCase, PeftCommonTester):
         ],
     )
     def test_generate_text_with_other_init(self, prompt_tuning_init) -> None:
+        # This test is flaky, hence fixing the seed. The reason is somehow related to:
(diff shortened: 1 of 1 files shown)


tests/test_multitask_prompt_tuning.py
```diff
@@ -246,6 +246,7 @@ def test_bf16_inference(self) -> None:
         _ = mpt.generate(input_ids=input_ids, task_ids=task_ids)
 
     def test_generate_text_with_random_init(self) -> None:
+        torch.manual_seed(0)
```
SunMarc, line 249: Does this also happen here ?

SunMarc (approved):
LGTM ! Thanks for fixing !

tests/test_multitask_prompt_tuning.py
BenjaminBossan: I had not encountered it before but after fixing the seed in the other test, this one failed on some CI runs. I expect it's the same error and it was just coincidence that I encountered this in the other test only.

PR #1509, opened 2024-02-26
Add lora+ implentation

moghadas76:
Implementing LoRA+ https://arxiv.org/abs/2402.12354

PR #1915, opened 2024-07-08
Add lora+ implementation

kallewoof:
[LoRA+](https://arxiv.org/abs/2402.12354)

Builds on #1509.


src/peft/helpers.py
```diff
@@ -146,3 +154,102 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
         is_peft_model = False
 
     return is_peft_model
+
+
+def get_module(name, opt_model):
```
BenjaminBossan, line 159: All this code here can be removed, right?

kallewoof: Yes, removed, thanks for pointing that out. I wonder if I screwed up a rebase somewhere. Anyway, gone now.

BenjaminBossan:
@kallewoof Thanks for pushing this PR forward. By now, we have waited long enough that I feel it's fair to continue with this PR.

I haven't done a full review yet, but it seems that tests are failing. Could you please take a look?

kallewoof:
@BenjaminBossan Thanks. I think I addressed the errors.

BenjaminBossan:
We get an error in Python 3.8 as we're missing a `from __future__ import annotations`.

kallewoof:
Thanks, fixed. Also added a license header.

src/peft/optimizers/loraplus.py
```diff
@@ -0,0 +1,98 @@
...
+from transformers.pytorch_utils import ALL_LAYERNORM_LAYERS
+from transformers.trainer_pt_utils import get_parameter_names
+
+from ..peft_model import PeftModel
+from ..tuners.lora.layer import Embedding
+
+
+def create_loraplus_optimizer(
+    model: PeftModel,
+    optimizer_cls: type[Optimizer],
```
BenjaminBossan, line 18: `type[Optimizer]` fails with Python 3.8, please add the `__future__` import to this file as well.

src/peft/optimizers/__init__.py
```diff
@@ -0,0 +1,20 @@
+# flake8: noqa
+# There's no way to ignore "F401 '...' imported but unused" warnings in this
```
BenjaminBossan, line 2: What if you add:

```python
__all__ = ["create_loraplus_optimizer"]

File: src/peft/optimizers/loraplus.py [hunk 4]
```diff
@@ -0,0 +1,98 @@
+"""
```
BenjaminBossan, line 1: License notice is missing.

src/peft/tuners/lora/__init__.py
```diff
@@ -14,7 +14,7 @@
 
 from peft.import_utils import is_bnb_4bit_available, is_bnb_available, is_eetq_available
 
-from .config import LoftQConfig, LoraConfig, LoraRuntimeConfig
+from .config import LoftQConfig, LoraConfig, LoraPlusConfig, LoraRuntimeConfig
```
BenjaminBossan, line 17: Hmm, do we really need it here? I think it's enough to have this in the global PEFT namespace.

src/peft/tuners/__init__.py
```diff
@@ -18,7 +18,7 @@
 # limitations under the License.
 
 from .adaption_prompt import AdaptionPromptConfig, AdaptionPromptModel
-from .lora import LoraConfig, LoraModel, LoftQConfig, LoraRuntimeConfig
+from .lora import LoraConfig, LoraModel, LoftQConfig, LoraPlusConfig, LoraRuntimeConfig
```
BenjaminBossan, line 21: Hmm, do we really need it here? I think it's enough to have this in the global PEFT namespace.

src/peft/utils/peft_types.py
```diff
@@ -57,6 +58,7 @@ class PeftType(str, enum.Enum):
     OFT = "OFT"
     POLY = "POLY"
     LN_TUNING = "LN_TUNING"
+    LORAPLUS = "LORAPLUS"
```
BenjaminBossan, line 61: I don't think it's needed here, we will never initialize a "LORAPLUS" PEFT model.

BenjaminBossan (changes requested):
Thanks for the updates. No in-depth review yet, but I found a couple of issues that should be quick to address.

src/peft/tuners/__init__.py
kallewoof: Will try -- that seems like it applies to the other ones in there too though. What is different?

src/peft/utils/peft_types.py
kallewoof: The tests rely on its existence in the "try all the peft types" test(s).

src/peft/tuners/__init__.py
BenjaminBossan: True, maybe this was added in a refactor for backwards compatibility, I don't quite remember. Anyway, since the others are "tuners" but LoRA+ is for optimizers, I still think it's better not to add `LoraPlusConfig` here.

kallewoof: Good point!

kallewoof:
Failing tests appear unrelated to the PR.

src/peft/optimizers/loraplus.py
```diff
@@ -0,0 +1,114 @@
...
+) -> Optimizer:
+    """
+    Creates a LoraPlus optimizer. Implementing LoRA+ https://arxiv.org/abs/2402.12354 Reference:
+    https://github.com/nikhil-ghosh-berkeley/loraplus/
+
+    Args:
+        model (`torch.nn.Module`): The model to be optimized.
+        optimizer_cls (`torch.optim.Optimizer`): The optimizer class to be used.
+        optimizer_kwargs (`dict`): Additional keyword arguments to be passed to the optimizer.
+        loraplus_lr_ratio (`float`): The ratio of the learning rate to be used for the embedding layer.
```
BenjaminBossan, line 47: The docstring is wrong. Also, could it be expanded a bit? For reference: https://github.com/nikhil-ghosh-berkeley/loraplus/tree/main?tab=readme-ov-file#lora-arguments

src/peft/optimizers/loraplus.py
```diff
@@ -0,0 +1,114 @@
...
+from transformers.trainer_pt_utils import get_parameter_names
+
+from ..peft_model import PeftModel
+from ..tuners.lora.layer import Embedding
+
+
+def create_loraplus_optimizer(
+    model: PeftModel,
+    optimizer_cls: type[Optimizer],
+    optimizer_kwargs: dict,
```
BenjaminBossan, line 35: A suggestion: Let's remove `optimizer_kwargs` and just add `**kwargs`. IMO, that makes calling this function easier, as we can use `create_loraplus_optimizer(..., weight_decay=1e-3)` instead of `create_loraplus_optimizer(..., optimizer_kwargs={..., "weight_decay": 1e-3})`. And since `lr` is not optional, let's make this a normal arg of `create_loraplus_optimizer`.

src/peft/optimizers/__init__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan, line 1:
```suggestion
# Copyright 2024-present the HuggingFace Inc. team.
```

src/peft/optimizers/loraplus.py
```diff
@@ -0,0 +1,114 @@
...
+    loraplus_lr_ratio: float,
+    loraplus_lr_embedding: float = 1e-6,
+) -> Optimizer:
+    """
+    Creates a LoraPlus optimizer. Implementing LoRA+ https://arxiv.org/abs/2402.12354 Reference:
```
BenjaminBossan, line 40: Let's add a line break before "Reference".

src/peft/optimizers/loraplus.py
```diff
@@ -0,0 +1,114 @@
...
+    https://github.com/nikhil-ghosh-berkeley/loraplus/
+
+    Args:
+        model (`torch.nn.Module`): The model to be optimized.
+        optimizer_cls (`torch.optim.Optimizer`): The optimizer class to be used.
+        optimizer_kwargs (`dict`): Additional keyword arguments to be passed to the optimizer.
+        loraplus_lr_ratio (`float`): The ratio of the learning rate to be used for the embedding layer.
+        loraplus_lr_embedding (`float`): The learning rate to be used for the embedding layer.
+    """
```
BenjaminBossan, line 49: Let's document the return parameter. Having an example section in the docstring would be nice to have.

src/peft/optimizers/loraplus.py
```diff
@@ -0,0 +1,114 @@
...
+            if name in decay_parameters:
+                param_groups["groupB"][name] = param
+            else:
+                param_groups["groupB_no_decay"][name] = param
+        else:
+            param_groups["groupA"][name] = param
+
+    assigned_param_groups = ""
+    for group in param_groups:
+        assigned_param_groups += f"{group}\n {list(param_groups[group].keys())}\n\n"
```
BenjaminBossan, line 77: Since we don't log this info, this can be removed.

tests/test_loraplus_helper.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
+
+# coding=utf-8
```
BenjaminBossan, line 3: Can be removed.

tests/test_loraplus_helper.py
```diff
@@ -0,0 +1,101 @@
...
+def test_lora_plus_helper_sucess():
+    model = SimpleNet()
+    optimizer_cls = bnb.optim.Adam8bit
+    optim_config = {
+        "lr": 5e-5,
+        "eps": 1e-6,
+        "betas": (0.9, 0.999),
+        "weight_decay": 0.0,
+    }
+    loraplus_lr_ratio = 0.2
```
BenjaminBossan, line 58: Not super important, but

> loraplus_lr_ratio should be >= 1

according to the original repo. I don't think we need to raise an error for that, but let's adjust the tests, in case users are checking for guidance.

BenjaminBossan (changes requested):
Thanks for continuing the work on LoRA+. I found a few areas for improvement, but overall it looks good already. Tested it on a small example and results were slightly improved.

src/peft/optimizers/loraplus.py
kallewoof: Updated. Maybe a bit wordy but seems like useful info.

src/peft/optimizers/loraplus.py
kallewoof: Good point.

src/peft/optimizers/__init__.py
kallewoof: Sure thing. But isn't the left part of copyright years normally kept as whatever year it began to exist?

tests/test_loraplus_helper.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
```
BenjaminBossan, line 1: Oh, just noticed the file name, let's remove the `_helper` part, as it's not necessary.

src/peft/optimizers/loraplus.py
kallewoof: Copied the original code comment over.

tests/test_loraplus_helper.py
kallewoof: Oops, yeah, good point. Set to >1.

BenjaminBossan:
@kallewoof LMK when this is ready for another review.

kallewoof:
@BenjaminBossan Should be ready! Sorry if I missed anything.

src/peft/optimizers/loraplus.py
```diff
@@ -0,0 +1,120 @@
...
+            param_groups["embedding"][name] = param
+        elif "lora_B" in name or param.ndim == 1:
+            if name in decay_parameters:
+                param_groups["groupB"][name] = param
+            else:
+                param_groups["groupB_no_decay"][name] = param
+        else:
+            param_groups["groupA"][name] = param
+
+    lr = kwargs["lr"]
```
BenjaminBossan, line 84: Since `lr` is a required parameter, let's add it to the function signature directly instead of hiding it inside the `kwargs`. We could make `lr` (and `loraplus_lr_ratio`) forced keyword arguments to avoid having too many positional arguments.

src/peft/optimizers/loraplus.py
```diff
@@ -0,0 +1,120 @@
...
+    weight_decay = kwargs.get("weight_decay", 0.0)
```
BenjaminBossan, line 85: Just wondering: Since the `weight_decay` is already set for each param group, is there any point in still passing it to `optimizer_cls`? If not, this line could be changed to `weight_decay = kwargs.pop("weight_decay", 0.0)`

BenjaminBossan (changes requested):
Fantastic, I think we're almost done. I only have two small comments left, the rest looks good.

src/peft/optimizers/loraplus.py
kallewoof: Good point, done.

src/peft/optimizers/loraplus.py
kallewoof: Isn't weight decay used in some optimizers? Or should that be separate from this weight decay?

BenjaminBossan:
@kallewoof Please ping me once this is ready for review. Otherwise, I don't know if you're still working on some changes or not :)

kallewoof:
@BenjaminBossan Ping! Sorry about all the force-pushes.

kallewoof:
@BenjaminBossan Sorry, I thought about this overnight and I think you're right that `weight_decay` should be popped. If someone provides a weight decay meant for something other than LoRA+ then it will be picked up by both, which is most likely undesired.

kallewoof:
Playing around with this, I noticed that `LoraPlusConfig` is actually not used anywhere. I think the code is missing a part where you create a `LoraConfig` with `lora+` feature and it initializes the optimizer and such.

BenjaminBossan:
> Playing around with this, I noticed that `LoraPlusConfig` is actually not used anywhere.

Hmm, yes, you're right. How about removing it completely then? I guess an argument could be made that something like this API could be useful:

```python
from peft import LoraPlusConfig

optimizer_config = LoraPlusConfig(...)
optimizer = create_loraplus_optimizer(model, optimizer_config)
```

to make it easier to share the config settings, but IMO the value is very marginal.

kallewoof:
> > Sorry about all the force-pushes.
> 
> Not a big deal with this PR, but especially on bigger ones they're better to avoid to make reviews easier. Note that there is no need to clean up the git history, if that was what you're going for, as we squash before merging.

Right. It's a very ingrained habit from other projects where they don't squash.

> > Sorry, I thought about this overnight and I think you're right that weight_decay should be popped. If someone provides a weight decay meant for something other than LoRA+ then it will be picked up by both, which is most likely undesired.
> 
> What do you mean by "both"?

Imagine `FutureOptimizerX` which has a fancy `weight_decay` parameter that does something, and now imagine someone attaching a LoRA+ optimizer to it:
```
def create_loraplus_optimizer(
    model: PeftModel, optimizer_cls: type[Optimizer], *, lr: float, loraplus_lr_ratio: float, **kwargs
) -> Optimizer:
```

The LoRA+ optimizer picks out and uses the passed-in weight decay value in its setup. Then, if we don't pop it,

```
    optimizer = optimizer_cls(optimizer_grouped_parameters, **kwargs)
```

the `FutureOptimizerX` optimizer will now *also* use our weight decay param. It is unclear which optimizer the user was referring to for the weight decay, but it's probably unlikely that they intended for both `FutureOptimizerX` and `LoRA+` to get the same weight decay with the same value.

Maybe I'm overcomplicating this?

> > Playing around with this, I noticed that `LoraPlusConfig` is actually not used anywhere.
> 
> Hmm, yes, you're right. How about removing it completely then? I guess an argument could be made that something like this API could be useful:
> 
> ```python
> from peft import LoraPlusConfig
> 
> optimizer_config = LoraPlusConfig(...)
> optimizer = create_loraplus_optimizer(model, optimizer_config)
> ```
> 
> to make it easier to share the config settings, but IMO the value is very marginal.

I think the cleanest approach is to remove it in this PR and then make a follow-up where we make it easier to use, if necessary.

Ultimately, without some tweaks to `transformers`, I don't think we can do this automatically e.g. from `get_peft_model` because we need to actually access the `Trainer`. I think.

BenjaminBossan:
> It is unclear which optimizer the user was referring to for the weight decay, but it's probably unlikely that they intended for both `FutureOptimizerX` and `LoRA+` to get the same weight decay with the same value.

Yes, I think that's the gist, it is unclear what the user meant in this case. However, if we pop the weight decay argument, the user can still do the following to pass `weight_decay` to the optimizer itself:

```python
optim = create_loraplus_optimizer(
    model, optimizer_cls=partial(FutureOptimizerX, weight_decay=weight_decay_X), weight_decay=weight_decay_lora, ...
)
```

It's a bit inconvenient but simple enough. If `weight_decay` is not popped, users cannot prevent it from being passed. Therefore, popping looks like the more flexible solution at the end of the day. The docstring could mention that just to make sure.

> I think the cleanest approach is to remove it in this PR and then make a follow-up where we make it easier to use, if necessary.

Agree.

> We probably should add an example to the docs on how to use it though, at least. Let me look into that.

That would be fantastic. I had already created an item on the backlog to extend the docs to include an optimizer section. If you'd already tackle this now, all the better :)

kallewoof:
> > It is unclear which optimizer the user was referring to for the weight decay, but it's probably unlikely that they intended for both `FutureOptimizerX` and `LoRA+` to get the same weight decay with the same value.
> 
> Yes, I think that's the gist, it is unclear what the user meant in this case. However, if we pop the weight decay argument, the user can still do the following to pass `weight_decay` to the optimizer itself:
> 
> ```python
> optim = create_loraplus_optimizer(
>     model, optimizer_cls=partial(FutureOptimizerX, weight_decay=weight_decay_X), weight_decay=weight_decay_lora, ...
> )
> ```
> 
> It's a bit inconvenient but simple enough. If `weight_decay` is not popped, users cannot prevent it from being passed. Therefore, popping looks like the more flexible solution at the end of the day. The docstring could mention that just to make sure.

Ah, right. Annoying, but at least there's a workaround. I think adding to docs is premature, unless there is an optimizer out there that takes a weight decay already?

> > I think the cleanest approach is to remove it in this PR and then make a follow-up where we make it easier to use, if necessary.
> 
> Agree.

Removed.

> > We probably should add an example to the docs on how to use it though, at least. Let me look into that.
> 
> That would be fantastic. I had already created an item on the backlog to extend the docs to include an optimizer section. If you'd already tackle this now, all the better :)

I took a stab at it. I wasn't sure where to put it, and I'm not sure this way of describing it is ideal, so feedback welcome!

kallewoof:
Actually I am renaming `weight_decay` to `loraplus_weight_decay` to circumvent this entire issue.

BenjaminBossan:
Failing CI looks unrelated. I won't restart it for now, since you want to make some changes anyway.

kallewoof:
Thanks. Pushed rename. I think I'm done unless the docs need work.

docs/source/developer_guides/lora.md
```diff
@@ -152,6 +152,42 @@ The default LoRA settings in PEFT add trainable weights to the query and value l
 config = LoraConfig(target_modules="all-linear", ...)
 ```
 
+### LoRA+ optimized LoRA
```
BenjaminBossan, line 155: As is, this is a subsection of `## Initialization`. Instead, let's create a new section `## Optimizers` below and make this subsection of that section.

docs/source/developer_guides/lora.md
```diff
@@ -152,6 +152,42 @@ The default LoRA settings in PEFT add trainable weights to the query and value l
...
+
+LoRA training can be optimized using [LoRA+](https://arxiv.org/abs/2402.12354), which uses different learning rates for the adapter matrices A and B, shown to increase finetuning speed by up to 2x and performance by 1-2%.
+
+Assuming you have a properly configured `PeftModel`, sub-class the `Trainer` replacing the `create_optimizer` method:
```
BenjaminBossan, line 159: I think we can remove the `Assuming ...` part, as this goes without saying.

docs/source/developer_guides/lora.md
```diff
@@ -152,6 +152,42 @@ The default LoRA settings in PEFT add trainable weights to the query and value l
...
+
+```py
+# Snippet based on Axolotl's trainer_builder.py: https://github.com/axolotl-ai-cloud/axolotl/blob/985819d89bec921e919e7e83042a869f04a25974/src/axolotl/core/trainer_builder.py#L305-L384
+from transformers import Trainer
+from peft.optimizers import create_loraplus_optimizer
+
+class MyTrainer(Trainer):
+    def create_optimizer(self):
```
BenjaminBossan, line 167: Is this really necessary? I think it works equally well if we initialize the optimizer directly and pass it to the `Trainer` as argument. Also, that way, it's easier to adopt the example in case that users don't use `Trainer` at all.

```python

optimizer = create_loraplus_optimizer(...)
scheduler = None

...
trainer = Trainer(
    ...,
    optimizers=(optimizer, scheduler),
)
```

BenjaminBossan (changes requested):
Thanks for creating the documentation. I have a couple of change requests for the docs, WDYT?

docs/source/developer_guides/lora.md
kallewoof: That's way better, yeah. Thanks.

kallewoof:
@BenjaminBossan Updated. Thanks for the guidance.

docs/source/developer_guides/lora.md
```diff
@@ -201,6 +165,28 @@ Assuming the original model had 5 layers `[0, 1, 2 ,3, 4]`, this would create a
...
+
+### LoRA+ optimized LoRA
+
+LoRA training can be optimized using [LoRA+](https://arxiv.org/abs/2402.12354), which uses different learning rates for the adapter matrices A and B, shown to increase finetuning speed by up to 2x and performance by 1-2%.
+
+```py
+from transformers import Trainer
+from peft.optimizers import create_loraplus_optimizer
+
+optimizer = create_loraplus_optimizer(...)
```
BenjaminBossan, line 180: Thanks for updating the docs.

In my example, the `...` were just intended as placeholders. At least for `create_loraplus_optimizer`, it would be good to show an actually valid call. Let's also include the part where the LoRA model is initialized, otherwise users may get the impression that this will work with non-LoRA models.

kallewoof: By `the part where the LoRA model is initialized` are you referring to `get_peft_model()` or something else?

I pushed a commit that adds params to the optimizer.

BenjaminBossan: > By `the part where the LoRA model is initialized` are you referring to `get_peft_model()`

Yes, to make it crystal clear that a LoRA model is expected here.

kallewoof: Right. I restored assumption and made it specifically say LoRA. If this isn't what you had in mind I'll think some more and rewrite it tomorrow. :)

docs/source/developer_guides/lora.md
```diff
@@ -165,6 +165,36 @@ Assuming the original model had 5 layers `[0, 1, 2 ,3, 4]`, this would create a
...
 
+## Optimizers
+
+LoRA training can optionally include special purpose optimizers. Currently the only such optimizer is LoRA+.
+
+### LoRA+ optimized LoRA
+
+LoRA training can be optimized using [LoRA+](https://arxiv.org/abs/2402.12354), which uses different learning rates for the adapter matrices A and B, shown to increase finetuning speed by up to 2x and performance by 1-2%.
+
+Assuming you have a properly set up `PeftConfig` for LoRA training:
```
BenjaminBossan, line 176: I was thinking more along the lines of just showing this in code. So e.g.:

```diff
from transformers import Trainer
from peft.optimizers import LoraConfig, get_peft_model, create_loraplus_optimizer
import bitsandbytes as bnb

+ base_model = ...
+ config = LoraConfig(...)
+ model = get_peft_model(base_model, config)

optimizer = create_loraplus_optimizer(
    model=model,
    optimizer_cls=bnb.optim.Adam8bit,
    lr=5e-5,
    loraplus_lr_ratio=16,
)
scheduler = None

...
trainer = Trainer(
    ...,
    optimizers=(optimizer, scheduler),
)
```

kallewoof: Got it. I put that in verbatim. I can fill out the ellipses but it feels unnecessary.

BenjaminBossan (approved):
Thanks for updating the docs. I think the usage of LoRA+ should now be pretty clear.

BenjaminBossan:
Thanks for all the work on this PR @kallewoof. Before merging, I would like to ensure that everyone involved is happy with the authorship.

The main author of this PR will be @kallewoof. On top, I would like to add @moghadas76 as a co-author. @shubhamjain0594 thanks for proposing to work on the PR, but as this did not manifest at the end, I would not add you as co-author. @stillmatic since you contributed to this PR, I'd also be happy to add you.

To the co-authors, please LMK if you would like to be added like so:

```
Co-authored-by: moghadas76 <s.m.moghadas2012@gmail.com>
Co-authored-by: Chris Hua <stillmatic@users.noreply.github.com>
```


If I don't hear back on this proposal in the next few days or if there is no consent, I will proceed to merge at my own discretion.

(note: failing Windows CI is unrelated)

BenjaminBossan:
Okay, I think this is enough time waited to get consent. As there was no disagreement, I plan to proceed as discussed above with the authorship.

Before merging, one last request @kallewoof: Could you please merge with/rebase on main, so that I can run the CI one last time? Thanks!

kallewoof:
@BenjaminBossan Sounds good. I rebased on main. If it is easier I will merge instead in the future.

BenjaminBossan:
Thanks everyone involved in this PR, good work.

PR #1919, opened 2024-07-10
Synchronize lora's merge, unmerge, etc. modifications to lora's tp_layer.

zhangsheng377:
There was a previous commit that moved the merge and other functions in LoraLayer to Linear, etc., but the LoraParallelLinear in tp_layer was missed.

### DIFF
--- a/src/peft/tuners/lora/tp_layer.py
+++ b/src/peft/tuners/lora/tp_layer.py
@@ -13,14 +13,17 @@
 # limitations under the License.
 
 import importlib
+import math
 import warnings
-from typing import Any, Optional
+from typing import Any, Optional, Union
 
 import torch
 import torch.nn as nn
 import torch.nn.init as init
 
-from peft.tuners.tuners_utils import BaseTunerLayer
+from peft.tuners.tuners_utils import check_adapters_to_merge, BaseTunerLayer
+from peft.utils import transpose
+from peft.utils.integrations import gather_params_ctx
 
 from .layer import LoraLayer
 
@@ -42,13 +45,14 @@ class LoraParallelLinear(nn.Module, LoraLayer):
         lora_alpha: int = 1,
         lora_dropout: float = 0.0,
         fan_in_fan_out: bool = False,
-        init_lora_weights: bool = True,
+        is_target_conv_1d_layer: bool = False,
+        init_lora_weights: Union[bool, str] = True,
         use_rslora: bool = False,
         use_dora: bool = False,
         **kwargs,
     ):
         super().__init__()
-        LoraLayer.__init__(self, base_layer=base_layer)
+        LoraLayer.__init__(self, base_layer=base_layer, **kwargs)
 
         if use_dora:
             raise ValueError(f"{self.__class__.__name__} does not support DoRA yet, please set it to False")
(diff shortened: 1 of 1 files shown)


zhangsheng377:
@BenjaminBossan

BenjaminBossan:
Thanks for this update. Could you please run `make style`?

zhangsheng377:
> Thanks for this update. Could you please run `make style`?

Ok. Sorry, I forgot it.

src/peft/tuners/lora/tp_layer.py
```diff
@@ -108,7 +116,7 @@ def update_layer(
         else:
             lora_dropout_layer = nn.Identity()
 
-        self.lora_dropout[adapter_name] = lora_dropout_layer
+        self.lora_dropout.update(nn.ModuleDict({adapter_name: lora_dropout_layer}))
```
BenjaminBossan, line 119: This should not be necessary, right?

BenjaminBossan (changes requested):
The PR is failing on Python 3.8 because of type annotation syntax. Adding `from __future__ import annotations` should fix that. I assume that you tried out that the changes you made to the megatron layer work?

zhangsheng377:
> The PR is failing on Python 3.8 because of type annotation syntax. Adding `from __future__ import annotations` should fix that. I assume that you tried out that the changes you made to the megatron layer work?

The `merge` func is that I copyed from [Linear.merge](https://github.com/zhangsheng377/peft/tree/main/src/peft/tuners/lora/layer.py)

But I forgot to add the `from __future__ import annotations`

BenjaminBossan:
I don't really have experience with megatron, so I'm not sure if these methods would just work when copied 1:1. Just to be sure, did you test with your setup that nothing breaks with these changes? If yes, I think we can merge and in the future try to be more careful to keep `tp_layer.py` in sync.

zhangsheng377:
> I don't really have experience with megatron, so I'm not sure if these methods would just work when copied 1:1. Just to be sure, did you test with your setup that nothing breaks with these changes? If yes, I think we can merge and in the future try to be more careful to keep `tp_layer.py` in sync.

Ha, yes, in our code we added the patch like [this](https://gitee.com/ascend/ModelLink/blob/master/modellink/training.py):
```python
setattr(peft.tuners.lora.LoraLayer, 'merge', peft.tuners.lora.Linear.merge)
setattr(peft.tuners.lora.LoraLayer, 'unmerge', peft.tuners.lora.Linear.unmerge)
setattr(peft.tuners.lora.LoraLayer, 'get_delta_weight', peft.tuners.lora.Linear.get_delta_weight)
```
It works good.

BenjaminBossan (approved):
> It works good.

Okay, great that you tested it. LGTM.

src/peft/tuners/lora/tp_layer.py
```diff
@@ -137,28 +146,51 @@ def update_layer(
...
+                self.loftq_init(adapter_name)
+        elif init_lora_weights:
             self.reset_lora_parameters(adapter_name, init_lora_weights)
 
+        # call this before dora_init
         self._move_adapter_to_device_of_base_layer(adapter_name)
+
+        if use_dora:
+            self.dora_init(adapter_name)
+            self.use_dora[adapter_name] = True
```
gapsong, line 171: We cannot reach here since we throw an error in line 58.
```
if use_dora:
    raise ValueError(f"{self.__class__.__name__} does not support DoRA yet, please set it to False")
```

BenjaminBossan: True, but this is here in case DoRA support will be added in the future.

PR #1926, opened 2024-07-14
fix参数encoder_reparameterization_type

sujeek:
This might be an easily overlooked issue of character loss. I'll fix it.


BenjaminBossan (approved):
Thanks for fixing the typo in the docs.

PR #1928, opened 2024-07-15
Fix attribute check for print_trainable_parameters method

anch0vy:
# Fix attribute check for print_trainable_parameters method

## Description
This PR fixes an incorrect attribute check in the main training script. The original code was checking for the existence of a method call instead of the method itself.

## Changes
- Changed `if hasattr(trainer.model, "print_trainable_parameters()")` to `if hasattr(trainer.model, "print_trainable_parameters")`


BenjaminBossan (approved):
Thanks for fixing this small error in the script.

Issue #1929, opened 2024-07-16
Pissa's [path_initial_model_for_weight_conversion] to build LoRA on top of base model


mtran14:
Hello,

I have a trained LoRA model that is initialized with the PiSSA method, and other than the trainable LoRA parameters (A,B), the method also alternate the original model weights from W to W', so `h=W'x + BAx`
I would like to reconvert the modified model W' back to the standard model W (please correct me if I'm wrong, but it seems possible using `path_initial_model_for_weight_conversion`) and  according to this [link](https://github.com/GraphPKU/PiSSA?tab=readme-ov-file#convert-pissa-to-lora). So the goal is to find A' and B' so that `h =Wx + B'A'x = W'x + BAx`

Here is my understanding of how to use `path_initial_model_for_weight_conversion`
```
model_name = 'openai/whisper-base'
whisper_model = WhisperForConditionalGeneration.from_pretrained('openai/whisper-base')
config = LoraConfig(
    r=r,
    lora_alpha=64,
    target_modules=['q_proj', 'v_proj'],
    lora_dropout=0.05,
    bias="none",
    use_rslora=True,
    init_lora_weights='pissa'
)
peft_model = get_peft_model(whisper_model, config)
peft_model.peft_config["default"].init_lora_weights = True 
peft_model.save_pretrained("pissa_init") # this should save A_0, B_0; base weights here is W'
...
train peft_model
...
peft_model.save_pretrained(
    'pissa-trained', path_initial_model_for_weight_conversion="pissa_init"
) # this should compute A' and B' from current A, B and the saved A_0, B_0; base weights here is W'
...
# new model
model = WhisperForConditionalGeneration.from_pretrained(model_name)
new_peft_model = PeftModel.from_pretrained(model, 'pissa-trained') # this should load A' and B', which is compatible with original weights W
```
However, the base model weights of `new_peft_model` is different from `whisper-base` official weights. Please correct me if I miss anything in my code, or I misunderstand any concepts.

Thank you for your help!

BenjaminBossan:
For me, this seems to work. I followed the [description here](https://github.com/huggingface/peft/tree/main/examples/pissa_finetuning). Below is a script to reproduce. Note how the weights behave exactly as expected. The line `peft_model.peft_config["default"].init_lora_weights = True` seems to be unnecessary.

```python
import torch
from transformers import WhisperForConditionalGeneration
from peft import get_peft_model, LoraConfig, PeftModel

torch.manual_seed(0)
[...]

PR #1930, opened 2024-07-16
FIX PiSSA & OLoRA with rank/alpha pattern, rslora

BenjaminBossan:
See https://github.com/huggingface/peft/issues/1929#issuecomment-2230780802

At the moment, when using PiSSA or OLoRA with weight conversion to restore the original base weights, there is an error when either of `rank_pattern`, `alpha_pattern`, or `rslora` is being used. This PR fixes this.

The issue is that we need to double the rank of the LoRA adapter. Right now, this is done by simply doubling r and alpha. But if `rank_pattern` and `alpha_pattern` are being used, those need to be doubled too.

Furthermore, when using `rslora`, the scaling is again different, namely `alpha / sqrt(r)`. This also needs to be adjusted.

Unfortunately, when using `rslora` together with `rank_pattern` and `alpha_pattern`, this gets way more complicated. Since we don't store the scaling in the `state_dict`, we would have to resolve all the patterns here to determine the correct scaling, i.e. reimplement the whole matching and init logic. This is a lot of work for a very edgy edge case.

Therefore, I opted to raise an error instead. This is not super nice, as the error is only raised when trying to save the model, i.e. a lot of time may already have been spent to train the model. But we cannot know this earlier, so not much can be done.

Overall, this fix is ugly because it further couples unrelated code. For instance, if we add new init methods that affect the scaling, we need to remember to change the saving logic accordingly. If anyone has a better idea, LMK.

## Update

I noticed that the test `test_lora_pissa_conversion_same_output_after_loading_with_quantization` was failing locally. This is because the argument `convert_mutated_to_lora` was renamed to `path_initial_model_for_weight_conversion` in #1828. This oversight has been addressed. However, this also means that the tests did not run in CI (nightly CI with GPU). The reason was that the PiSSA and OLoRA tests were missing the right marker, so I added them as well.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -379,7 +379,7 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
                         peft_config.lora_alpha *= 2
                     else:
                         # with rslora, we have scaling = alpha / sqrt(r), we thus adjust alpha to keep the same scaling
-                        peft_config.lora_alpha *= 2 ** 0.5
+                        peft_config.lora_alpha *= 2**0.5
 
                     if peft_config.rank_pattern:
                         peft_config.rank_pattern = {key: 2 * val for key, val in peft_config.rank_pattern.items()}
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -258,6 +258,13 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
             path_initial_model_for_weight_conversion = convert_pissa_to_lora
 
         def save_mutated_as_lora(peft_config, path_initial_model_for_weight_conversion, output_state_dict, kwargs):
+            if peft_config.use_rslora and (peft_config.rank_pattern or peft_config.alpha_pattern):
+                msg = (
+                    "Passing `path_initial_model_for_weight_conversion` to `save_pretrained` is not supported when "
+                    "using `rank_pattern` or `alpha_pattern` at the same time as `use_rslora=True`."
+                )
+                raise ValueError(msg)
+
             if not any(
                 str(peft_config.init_lora_weights).lower().startswith(prefix) for prefix in ["pissa", "olora", "true"]
             ):
@@ -368,7 +375,17 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
                 if path_initial_model_for_weight_conversion is not None:
                     peft_config.init_lora_weights = True
                     peft_config.r *= 2
-                    peft_config.lora_alpha *= 2
+                    if not peft_config.use_rslora:
+                        peft_config.lora_alpha *= 2
+                    else:
+                        # with rslora, we have scaling = alpha / sqrt(r), we thus adjust alpha to keep the same scaling
+                        peft_config.lora_alpha *= 2 ** 0.5
+
+                    if peft_config.rank_pattern:
+                        peft_config.rank_pattern = {key: 2 * val for key, val in peft_config.rank_pattern.items()}
+                    if peft_config.alpha_pattern:
+                        peft_config.alpha_pattern = {key: 2 * val for key, val in peft_config.alpha_pattern.items()}
+
                 peft_config.save_pretrained(output_dir, auto_mapping_dict=auto_mapping_dict)
             peft_config.inference_mode = inference_mode
 
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -231,9 +231,11 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
                 difference in adapter before and after fine-tuning is calculated. This difference can be represented as
                 the parameters of a standard LoRA adapter. Using this converted adapter does not require changes to the
                 base model, thus conveniently allowing the use of multiple PiSSA or OLoRA adapters with LoRA adapters,
-                and the activation or deactivation of any adapters.
+                and the activation or deactivation of any adapters. Note that this conversion is not supported if
+                `rslora` is used in combination with `rank_pattern` or `alpha_pattern`.
             kwargs (additional keyword arguments, *optional*):
                 Additional keyword arguments passed along to the `push_to_hub` method.
+
         """
         if os.path.isfile(save_directory):
             raise ValueError(f"Provided path ({save_directory}) should be a directory, not a file")
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import warnings
 from dataclasses import dataclass, field
 from typing import Literal, Optional, Union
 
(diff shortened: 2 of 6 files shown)


BenjaminBossan:
Tagging @fxmeng and @tokenizer-decode as this affects the code they contributed.

tokenizer-decode:
Oh, so we raise the error after all the training is done? Can you tell me why we are checking `peft_config.use_rslora and (peft_config.rank_pattern or peft_config.alpha_pattern):` after all the training is and before trying the save the model? Why are we not doing this when initializing PiSSA or OLoRA? Aren't `peft_config.rank_pattern ` and `peft_config.alpha_pattern` options determined before doing anything? Or maybe I did not understand you correctly.

tokenizer-decode:
Hmm. Maybe at the initialization if  `peft_config.use_rslora and (peft_config.rank_pattern or peft_config.alpha_pattern) ` is `True` we can warn the user saying you cannot use `path_initial_model_for_weight_conversion`? There is a chance he may stop there. I don't know. This is also not a good fix but at least gives user a chance. I would be pissed after spending bunch of many to an EC2 instance then getting this error.

BenjaminBossan:
@tokenizer-decode Good suggestion about adding a warning. I did that in the last commit, though unfortunately many users don't read warnings. Still, there is not much else that can be done here.

tokenizer-decode:
Yeah you are right not much else to do here. But I agree that this is a very edgy edge case so we should be ok.

BenjaminBossan:
@sayakpaul LMK if you don't want to review this, I'll look for someone else.

BenjaminBossan:
@tokenizer-decode While working on this, I also discovered a small error in a test caused by your PR, which I have fixed (see **update** section of PR description).

tokenizer-decode:
Oh I totally missed that. Thanks

src/peft/peft_model.py
```diff
@@ -258,6 +258,13 @@ def save_pretrained(
             path_initial_model_for_weight_conversion = convert_pissa_to_lora
 
         def save_mutated_as_lora(peft_config, path_initial_model_for_weight_conversion, output_state_dict, kwargs):
+            if peft_config.use_rslora and (peft_config.rank_pattern or peft_config.alpha_pattern):
```
sayakpaul, line 261: Can't we use this condition as soon as this permutation of config is requested and warn/error?

src/peft/tuners/lora/config.py
```diff
@@ -346,6 +347,26 @@ def __post_init__(self):
...
+                or (self.init_lora_weights == "olora")
+            )
+        ):
+            msg = (
+                "Using Rank-Stabilized LoRA with rank_pattern/alpha_pattern and post-training conversion of modified "
+                "base weights (PiSSA, OLoRA) means that you won't be able to pass "
+                "`path_initial_model_for_weight_conversion` to `save_pretrained` to restore the initial values of the "
+                "base weights; if you intend to do this, please ensure not to use rslora or rank_pattern/alpha_pattern."
+            )
+            warnings.warn(msg)
```
sayakpaul, line 368: Maybe we can document this limitation in the API spec itself, as well? 

I would also remember to include this in the release notes so that users are firmly aware of this.

tests/test_initialization.py
```diff
@@ -338,6 +338,194 @@ def test_lora_pissa_conversion_same_output_after_loading(self, data, tmp_path):
...
+        peft_model.peft_config["default"].init_lora_weights = True
+        peft_model.save_pretrained(tmp_path / "init-model")
+        peft_model.peft_config["default"].init_lora_weights = "pissa"
+
+        # modify the weights, or else the adapter performs an identity transformation
+        peft_model.base_model.linear.lora_B["default"].weight.data *= 2.0
+        output_pissa = peft_model(data)[0]
+
+        # sanity check
+        tol = 1e-06
```
sayakpaul, line 359: (nit) a stricter check for negations is to relax the tolerance a bit and see if that still holds IMO.

tests/test_initialization.py
```diff
@@ -338,6 +338,194 @@ def test_lora_pissa_conversion_same_output_after_loading(self, data, tmp_path):
...
+
+        # save the model normally
+        peft_model.save_pretrained(tmp_path / "pissa-model")
+        model_loaded = PeftModel.from_pretrained(deepcopy(model), tmp_path / "pissa-model")
+        output_loaded = model_loaded(data)[0]
+
+        assert torch.allclose(output_pissa, output_loaded, atol=tol, rtol=tol)
+        # sanity check: ranks should still be 8 as initially
+        assert model_loaded.peft_config["default"].r == 8
+        assert model_loaded.base_model.model.linear.lora_A["default"].weight.shape[0] == 32
```
sayakpaul, line 370: What is the difference between `r` being 8 and `rank_pattern` being 32 for `linear` here? 

`target_modules=["linear"], r=8, rank_pattern={"linear": 32}`

tests/test_initialization.py
```diff
@@ -338,6 +338,194 @@ def test_lora_pissa_conversion_same_output_after_loading(self, data, tmp_path):
...
+        model = self.get_model()
+        output_base = model(data)[0]
+
+        # use rank_pattern here
+        config = LoraConfig(init_lora_weights="pissa", target_modules=["linear"], r=8, rank_pattern={"linear": 32})
+        peft_model = get_peft_model(deepcopy(model), config)
+        # save the initial model
+        peft_model.peft_config["default"].init_lora_weights = True
+        peft_model.save_pretrained(tmp_path / "init-model")
+        peft_model.peft_config["default"].init_lora_weights = "pissa"
```
sayakpaul, line 352: Help me unpack this bit of code.

We are using `pissa` for `init_lora_weights` already in the `LoraConfig`. So, I don't understand what would differ between `peft_model.peft_config["default"].init_lora_weights = True` and `peft_model.peft_config["default"].init_lora_weights = "pissa"`.

tests/test_initialization.py
```diff
@@ -338,6 +338,194 @@ def test_lora_pissa_conversion_same_output_after_loading(self, data, tmp_path):
...
+        assert torch.allclose(output_pissa, output_converted, atol=tol, rtol=tol)
+        # rank should be double of what it was initially
+        assert model_converted.peft_config["default"].r == 16
+        assert model_converted.base_model.model.linear.lora_A["default"].weight.shape[0] == 64
+        # base model weights should be the same as the initial model
+        assert torch.allclose(
+            model.linear.weight, model_converted.base_model.model.linear.base_layer.weight, atol=tol, rtol=tol
+        )
+
+    def test_lora_pissa_conversion_same_output_after_loading_with_alpha_pattern(self, data, tmp_path):
```
sayakpaul, line 392: Same comments apply for this too.

sayakpaul (commented):
Thanks! Seems like a very deep edge case. I have left some comments to better understand some stuff. Once replied to, I will review one more time.

tests/test_initialization.py
tokenizer-decode: That was due to a bug in original PiSSA implementation. I fixed it when I was implementing olora but I didn’t change tests. What would differ is `init_lora_weights `should be `True` in the config file. See [here](https://github.com/huggingface/peft/pull/1828#issuecomment-2160483919)

src/peft/peft_model.py
BenjaminBossan: Yes, this is done below and we warn the user: https://github.com/huggingface/peft/pull/1930/files#diff-13380145401d203d5935c5189dd09879f990b81aa63e8e3aaff8ce9110333f0eR354

We don't raise an error since we cannot know at that point in time if they'll use the `path_initial_model_for_weight_conversion`. If they don't, there is no problem.

src/peft/tuners/lora/config.py
BenjaminBossan: I'll add a mention to the docstring of `save_pretrained`, as well as to the README files of the two methods.

I added an entry to the PEFT release notes to mention this change there as well. As OLoRA is as yet unreleased and PiSSA was released in the last version, and also given that this is an edge case, I think very few users, if any, will be affected.

tests/test_initialization.py
BenjaminBossan: True. At the moment, we use the same tolerance for `assert not torch.allclose(...)` and `assert torch.allclose(...)` throughout the whole PEFT code base. Therefore, I wouldn't change this now only for these specific tests. I think this could be a nice use case for a code-LLM to go through all tests and rewrite them to use higher tolerance for negative checks :)

tests/test_initialization.py
BenjaminBossan: Sorry, I missed this comment earlier, so addressing it now.

For this simple test, we only have a small model with a single linear layer. Therefore, if we set `rank_pattern={"linear": 32}`, we basically ignore `r=8` because `rank_pattern` takes precedence. If we had a real model, we could, however, set all ranks to 8 this way and pick individual layers with a different rank by using `rank_pattern`.

Does that answer the question?

src/peft/peft_model.py
sayakpaul: Okay that should cut it.

tests/test_initialization.py
sayakpaul: Yeah it does. Maybe just add a comment explaining this in the test? Because many developers refer to tests to understand different usage patterns and they may get confused here as well.

sayakpaul (approved):
Very minor changes requested. I think this is the best we can have right now.

BenjaminBossan:
@sayakpaul I added clarifying comments to the tests in my latest commit.

PR #1941, opened 2024-07-22
ENH Update VeRA preconfigured models

BenjaminBossan:
Some pre-configured models like mistral used not to work with VeRA because the weight shapes were not identical. However, since #1817, this is no longer a requirement. Therefore, this commented code can now be uncommented.

I have tested mistral and gemma and they worked. I haven't tested btlm and mixtral but with the update, I'm pretty sure they will work too.


BenjaminBossan:
ping @dkopi

BenjaminBossan:
Another gentle ping @dkopi

sayakpaul (approved):
Nice. More the coverage, better the uniformity.

Issue #1939, opened 2024-07-22
Memory usage spike when using `merge_and_unload`


snarayan21:
### System Info

peft version 0.23.4
transformers version 4.42.3

### Who can help?

@BenjaminBossan @sayakpaul

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

```
model = model.merge_and_unload()
```

### Expected behavior

Hi, I'm trying to merge lora adapters into the model with the `merge_and_unload` function. My model is on `cpu`. I'm near the cpu memory limit on my machine so I would like the `merge_and_unload` operation to happen in-place, but there seems to be a spike in memory usage when this function is called, causing cpu OOM. Is there an in-place version of this function / what makes `merge_and_unload` increase cpu memory usage significantly?

BenjaminBossan:
Do you have the error message? Could you also share the PEFT config and what base model you use?

snarayan21:
Hey @BenjaminBossan, there's no error message per se because it's a CPU oom -- the job just crashes. I'm using `meta-llama/Meta-Llama-3-70B` with the peft config below:
```
peft_config:
      r: 16
      peft_type: LORA
      task_type: CAUSAL_LM
      lora_alpha: 32
      lora_dropout: 0
      target_modules:
      - q_proj
      - k_proj
      - v_proj
      - o_proj
      - down_proj
      - up_proj
      - gate_proj
```

Looking at `_unload_and_optionally_merge`, is the LoRAModel holding on to additional unmerged parameters that it shouldn't be during the merge process?

snarayan21:
Ah I got to that same point in the code. That change _does_ seem to help at a very small scale (i'm tracking memory usage as LoRA adapters get merged on a tiny opt-125m model) -- will validate at larger scale as well.

Are there any other points where we might be allocating some extra tensor we don't need?

BenjaminBossan:
> will validate at larger scale as well.

Please report back once you have the results.

> Are there any other points where we might be allocating some extra tensor we don't need?

At first glance, I don't think so. Not saying there is no possibility at all, but at least no low hanging fruits.

snarayan21:
[...]

PR #1944, opened 2024-07-22
Decrease memory usage of `merge_and_unload`

snarayan21:
Addresses #1939

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -452,7 +452,7 @@ class Linear(nn.Module, LoraLayer):
                 else:
                     delta_weight = self.get_delta_weight(active_adapter)
                     if not self.use_dora[active_adapter]:
-                        base_layer.weight.data = base_layer.weight.data + delta_weight
+                        base_layer.weight.data += delta_weight
                     else:
                         # handle dora
                         # since delta_weight already includes scaling, set it to 1 here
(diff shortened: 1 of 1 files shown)


snarayan21:
@BenjaminBossan Would be great to get your approval here!

BenjaminBossan:
I checked the PEFT code on why we merge this way. Turns out we used to do it in-place but this PR changed it: #1372. IMO merging should not affect the error reported there, so I think we can undo the change, but I wanted to provide the necessary context.

Could you please also update the other merge functions?

https://github.com/huggingface/peft/blob/ba75bb14d115dfdff3294674d82d8097319ed74e/src/peft/tuners/lora/layer.py#L671
https://github.com/huggingface/peft/blob/ba75bb14d115dfdff3294674d82d8097319ed74e/src/peft/tuners/lora/layer.py#L927

Thanks.

I plan to make a release soon, aiming for tomorrow.

snarayan21:
@BenjaminBossan thanks for the pointers, found some additional places where this change is also needed.
 And yeah since #1372 was a training time error making this fix for merging adapters shouldn't be applicable.
 
 Hope this looks good to you.

snarayan21:
A release tomorrow with this fix would be ideal.

BenjaminBossan:
Strange that Windows CI is consistently failing. Hard to imagine that this is caused by the PR but I'm investigating further (but don't have a Windows machine myself). Hopefully this is just a temporary issue.

snarayan21:
Yeah quite weird:
```
The directory name is invalid: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpv612vppk\\model.safetensors'
```

BenjaminBossan:
Okay, so the same Windows errors as in #1947, which has no code changes.

BenjaminBossan (approved):
Thanks for restoring the more memory efficient merging behavior. A release tomorrow should be possible.

PR #1946, opened 2024-07-23
Release v0.12.0

BenjaminBossan:
Release planned for today.

Besides bumping the version, I also fixed an incorrect version indication in the docs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@
 from setuptools import find_packages, setup
 
 
-VERSION = "0.11.2.dev0"
+VERSION = "0.12.0"
 
 extras = {}
 extras["quality"] = [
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -15,7 +15,7 @@
 from setuptools import find_packages, setup
 
 
-VERSION = "0.11.2.dev0"
+VERSION = "0.12.0"
```
sayakpaul, line 18: This can be automated actually: https://github.com/huggingface/diffusers/blob/2c25b98c8ea74cfb5ec56ba49cc6edafef0b26af/Makefile#L86

sayakpaul (approved):

setup.py
BenjaminBossan: Yes, we can take a look into this at some point.

Issue #1940, opened 2024-07-22
Improving generalization of LoRA with wise-ft


SimJeg:
### Feature request

Include the [wise-ft](https://arxiv.org/abs/2109.01903) method in `LoraModel` models.

### Motivation

Wise-ft interpolates between the weights $W_{base}$ of a base model and the weights $W_{ft}$ of a finetuned model using the following formula:
$$W_{wise} = (1-\alpha) * W_{base} + \alpha * W_{ft} \tag{1}$$
where $\alpha \in [0,1]$. This approach effectively balances the trade-off between the zero-shot capabilities of the base model ($\alpha=0$) and the task-specific performance of the finetuned model ($\alpha=1$). It is particulary well suited when there is a distribution shift between the training data used for fine-tuning, and the test data used during inference.

For LoRA, as $W_{ft} = W_{base} + W_{delta}$, we can rewrite the formula (1) as:
 $$W_{wise} = W_{base} + \alpha * W_{delta} \tag{2}$$

Note that the formula above does not apply wise-ft correctly to DoRA, as for DoRA, $W_{ft} = m_{dora} * (W_{base} + W_{delta})$.

We successfully applied this method to win the 5 tracks of the [Amazon KDD Cup 2024](https://www.aicrowd.com/challenges/amazon-kdd-cup-2024-multi-task-online-shopping-challenge-for-llms/leaderboards), achieving performance improvements of up to 1.5%. In this competition, no training data was provided, hence there was a distribution shift between the training data built by the participants and the hidden test data used for evaluation.

### Your contribution

There are several ways to implement wise-ft for LoRA adapters: 
1. (no code) Update the configuration of the adapters and rescale the parameter `lora_alpha` by `α * lora_alpha` (which is not an int anymore)
2. (no code) Rescale the adapter weights by $\sqrt{\alpha}$ if no bias is used, as $W_{delta} = W_A \cdot W_B$. We did that in KDD cup.
3. Add a new attribute to the `LoraLayer` layers to update the scaling parameter of the adapters 
4. Add a new method in the `LoraModel` class that updates the `scaling` parameter of the LoRA adapters as follow:

```python
for module in self.model.modules():
    if isinstance(module, LoraLayer):
        module.scaling = dict((k, v * alpha) for k, v in module.scaling.items())
       # might be better to redefined scaling = lora_alpha / r OR sqrt(r) * alpha ? How to access use_rslora ?
```

[...]

PR #1951, opened 2024-07-24
[Add] scaling LoRA adapter weights with a context manager

ariG23498:
CC: @BenjaminBossan 
Related: #1940


ariG23498:
Hey @BenjaminBossan I tried with the following code to see whether we changed the scale and then reverted back to the original or not.

```py
from peft.utils.scaling_utils import set_adapter_scale
from peft import AutoPeftModelForCausalLM
from transformers import AutoTokenizer
from peft.tuners.lora.layer import LoraLayer


def check_scaling(model):
        if isinstance(module, LoraLayer):
            print(module.scaling)

model = AutoPeftModelForCausalLM.from_pretrained("ybelkada/opt-350m-lora")
tokenizer = AutoTokenizer.from_pretrained("facebook/opt-350m")

print("\n\nBEFORE")
check_scaling(model)

model = model.to("cuda")
model.eval()
inputs = tokenizer("Preheat the oven to 350 degrees and place the cookie dough", return_tensors="pt")


with set_adapter_scale(model=model, alpha=0.5):
    print("\n\nINSIDE")
    check_scaling(model)
    outputs = model.generate(input_ids=inputs["input_ids"].to("cuda"), max_new_tokens=50)


print(tokenizer.batch_decode(outputs.detach().cpu().numpy(), skip_special_tokens=True)[0])
print("\n\nAFTER")
check_scaling(model)
```

BenjaminBossan:
Thanks for reporting on the progress. Based on your example, it would be good to create the first unit test to ensure that this function works correctly. Could you please transform your example into a unit test that uses `assert`s instead of printing. I'd also do the testing a bit more precisely:

- Generate logits before scaling context (don't use tokens, as logits are more precise)
- Store scalings before scaling context
- Enter scaling context
- Generate logits and assert that they are different than before
- Check that the scalings are now the new value
- Exit scaling context
- Generate logits and assert that they are the same as before
- Check that the scalings are back to initial value

Also, I think you can move the new function to `helpers.py` for now.

ariG23498:
@BenjaminBossan I have added the unit tests as you have mentioned. How do they look?
Please let me know if there are any changes required from my side.

tests/test_helpers.py
```diff
@@ -66,3 +68,59 @@ def test_local_model_non_default_name(self, tmp_path):
...
+            for key in scales_before_scaling.keys():
+                assert scales_before_scaling[key] != scales_during_scaling[key]
+
+            # Generate logits during scaling
+            with torch.no_grad():
+                inputs = inputs.to(device)
+                model = model.to(device)
+                logits_during_scaling = model(**inputs).logits
+
+            assert torch.allclose(logits_before_scaling, logits_during_scaling)
```
BenjaminBossan, line 104: Shouldn't the logits be different when scaled?

tests/test_helpers.py
```diff
@@ -66,3 +68,59 @@ def test_local_model_non_default_name(self, tmp_path):
         # with adapter name
         result = check_if_peft_model(tmp_path / "peft-gpt2-other" / "other")
         assert result is True
+
```
BenjaminBossan, line 71: Please put this into a new test class with an appropriate name, as this is unrelated to `TestCheckIsPeftModel`. You can also define `scale_from_modules` inside of that class.

src/peft/helpers.py
```diff
@@ -146,3 +148,20 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
...
+def set_adapter_scale(model, alpha):
+    # 1. TODO: Check whether scaling is prohibited on model
+
+    # 2. Modify scaling values
+    original_scaling = {}
+    for module in model.modules():
+        if isinstance(module, LoraLayer):
+            original_scaling[module] = module.scaling.copy()
+            module.scaling = {k: v * alpha for k, v in module.scaling.items()}
+    yield
```
BenjaminBossan, line 163: Let's use
```python
try:
    yield
finally:
    # restore
```

That way it is ensured that the original scale is restored, even if there is an error in-between. There should be a test for this edge case too.

src/peft/helpers.py
```diff
@@ -146,3 +148,20 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
         is_peft_model = False
 
     return is_peft_model
+
+
+@contextmanager
+def set_adapter_scale(model, alpha):
+    # 1. TODO: Check whether scaling is prohibited on model
```
BenjaminBossan, line 155: This has yet to be implemented. E.g. you could check that the `original_scaling` is not empty after the loop exits. If it's empty, it probably means that the user passed an incompatible model and an error message with a good error message should be raised. There should be a test for that.

src/peft/helpers.py
```diff
@@ -146,3 +148,20 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
         is_peft_model = False
 
     return is_peft_model
+
+
+@contextmanager
+def set_adapter_scale(model, alpha):
```
BenjaminBossan, line 154: Let's add a nice docstring to this function. Also, let's update the docs.

tests/test_helpers.py
```diff
@@ -66,3 +68,59 @@ def test_local_model_non_default_name(self, tmp_path):
...
+        for key in scales_before_scaling.keys():
+            assert scales_before_scaling[key] == scales_after_scaling[key]
+
+        # Get logits after scaling
+        with torch.no_grad():
+            inputs = inputs.to(device)
+            model = model.to(device)
+            logits_after_scaling = model(**inputs).logits
+
+        assert torch.allclose(logits_before_scaling, logits_after_scaling)
```
BenjaminBossan, line 117: I think the function should already work with transformers and diffusers models as is. But let's add tests for those to ensure it's really working.

Also, we need tests for:

* merging inside the context
* using multiple adapters (as the scalings of all adapters are updated)
* a test that checks that restoration works even if we have different `r` and `alpha` on the layers (using `rank_pattern` and `alpha_pattern`).

BenjaminBossan (changes requested):
Thanks for the updates. The function itself already looks very promising. Let's do some more checks and extend the unit tests to be more comprehensive.

tests/test_helpers.py
ariG23498: Yes you are right. With a different scaling, the logits should be different.

In this colab notebook I have tried building a GPT2 model with pretrained weights and then build the LoRA version using the `get_peft_model()` API. Upon forward passing the same input, the logits of the two models are the same.

I think I am missing something. Is this intended behaviour?

[Colab Notebook](https://colab.research.google.com/gist/ariG23498/08ecf1672347ae6d031909866ad6266a/scratchpad.ipynb)

ariG23498:
@BenjaminBossan thanks for the tip! I was not aware of the `init_lora_weights=False` parameter. It works as is supposed to now.

Side note: Can you point me to a resource where I could understand how to make the model logits diterministic? Right now, with the same mode and same input the logits seems to be different (by a matter of small variance).

tests/test_helpers.py
ariG23498: Resolved with [this](https://github.com/huggingface/peft/pull/1951#issuecomment-2252956884) comment.

tests/test_helpers.py
ariG23498: Took care of this!

src/peft/helpers.py
ariG23498: This makes sense to me! Thanks for the suggestion.

src/peft/helpers.py
ariG23498: I have made changes to the code to test this. Could you give a quick skim and let me know you had a similar vision?

ariG23498: Added an initial doctring to the function. I think we need to iterate on the docstring and make it more verbose.

src/peft/helpers.py
ariG23498: Added an initial doctring to the function. I think we need to iterate on the docstring and make it more verbose.

tests/test_helpers.py
```diff
@@ -66,3 +69,99 @@ def test_local_model_non_default_name(self, tmp_path):
...
+
+    def test_not_lora_model(self):
+        model = AutoModelForCausalLM.from_pretrained("facebook/opt-125m")
+
+        # we expect a value error here because the model
+        # does not have lora layers
+        with pytest.raises(ValueError):
```
BenjaminBossan, line 165: Let's also match the error message.

src/peft/helpers.py
```diff
@@ -146,3 +148,66 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
...
+        for module in model.modules():
+            if isinstance(module, LoraLayer):
+                original_scaling[module] = module.scaling.copy()
+                module.scaling = {k: v * alpha for k, v in module.scaling.items()}
+
+        # check whether scaling is prohibited on model
+        # the original scaling dictionary should be empty
+        # if there were no lora layers
+        # checking empty dictionary: https://stackoverflow.com/a/23177452/10319735
+        if not bool(original_scaling):
```
BenjaminBossan, line 201:
```suggestion
        if not original_scaling:
```

`bool` is not necessary.

src/peft/helpers.py
```diff
@@ -146,3 +148,66 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
...
+            raise ValueError("model does not support scaling")
```
BenjaminBossan, line 202: Let's be more precise here: Right now, scaling is only supported for `LoraLayer`s. Let's add this info.

src/peft/helpers.py
```diff
@@ -146,3 +148,66 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
...
+
+    Example:
+        >>> model = ModelWithLoraLayer()
+        >>> alpha = 0.5
+        >>> with set_adapter_scale(model, alpha):
+        >>>     # Perform operations with the scaled model
+        >>> # The original scaling values are restored here
+    """
+    # initialize the flag with false
+    scale_changed = False
```
BenjaminBossan, line 182: We don't really need this, it should be enough to check if `original_scaling` is empty.

src/peft/helpers.py
```diff
@@ -146,3 +148,66 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
...
+            raise ValueError("model does not support scaling")
+
+        # change the flag to true since the scale has been changed
+        scale_changed = True
+        yield
+
+    finally:
+        # restore original scaling values after exiting the context
+        # only if the scale has been changed
+        if scale_changed and bool(original_scaling):
```
BenjaminBossan, line 211: This whole `if` can be removed: If `original_scaling` is empty, the loop below will not run anyway.

src/peft/helpers.py
```diff
@@ -146,3 +148,66 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
...
+        >>> with set_adapter_scale(model, alpha):
+        >>>     # Perform operations with the scaled model
+        >>> # The original scaling values are restored here
+    """
+    # initialize the flag with false
+    scale_changed = False
+
+    try:
+        # check if alpha has a valid data type
+        if not isinstance(alpha, float):
```
BenjaminBossan, line 186: I wonder if passing alpha as type `int` should not be fine, WDYT?

src/peft/helpers.py
```diff
@@ -146,3 +148,66 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
...
+        >>> model = ModelWithLoraLayer()
+        >>> alpha = 0.5
+        >>> with set_adapter_scale(model, alpha):
+        >>>     # Perform operations with the scaled model
+        >>> # The original scaling values are restored here
+    """
+    # initialize the flag with false
+    scale_changed = False
+
+    try:
```
BenjaminBossan, line 184: Let's move the `try` just before the `yield`, the rest should not be inside the `try` block. E.g. right now, the `try` block would catch the `raise ValueError("model does not support scaling")` error, which doesn't make sense.

src/peft/helpers.py
```diff
@@ -146,3 +148,66 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
...
+
+    Raises:
+        TypeError: If `alpha` is not of type float.
+        ValueError: If the model does not contain any `LoraLayer` instances,
+                    indicating that the model does not support scaling.
+
+    Yields:
+        None: This context manager does not yield any values.
+
+    Example:
```
BenjaminBossan, line 174: Syntax should be:

```
    Example:

    ```python
    >>> model = ModelWithLoraLayer()
```

etc.

src/peft/helpers.py
```diff
@@ -146,3 +148,66 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
...
+    Args:
+        model: The model containing `LoraLayer` modules whose scaling is to be adjusted.
+        alpha (float): The scaling factor to be applied. Must be of type float.
+
+    Raises:
+        TypeError: If `alpha` is not of type float.
+        ValueError: If the model does not contain any `LoraLayer` instances,
+                    indicating that the model does not support scaling.
+
+    Yields:
```
BenjaminBossan, line 171: Can be removed, as nothing is yielded.

BenjaminBossan (changes requested):
Thanks for all the updates. I have reviewed again and saw a few more areas to improve the PR further, please check out the comments. My [earlier comment](https://github.com/huggingface/peft/pull/1951#discussion_r1691604210) also still stands.

src/peft/helpers.py
ariG23498: Very nice catch!

src/peft/helpers.py
ariG23498: I am not sure with this one. Is there a range for alpha? If yes, we should definitely have a test for range and dtype.

tests/test_helpers.py
ariG23498: Addressed!

tests/test_helpers.py
ariG23498: Hey @BenjaminBossan I had some queries:
1. I see that the diffusion models with LoRAs in the Hub have a bigger base model. This in turn would be very slow for the testing pipeline. Do you have any suggestions to ease this?
2. Merging: As I understand, we merge the weights, then pass the model into the context manager, and get the merged lora outputs. After exiting from the context manager, the model should revive its merged weights. Am I understanding this correctly?
3. I am not sure I understand the usage of multiple adapters. If you could point me to some resource on this topic that would be very helpful.
4. I am unable to find an example where there is a use of `rank_pattern` and `alpha_pattern` -- Could you help me with this as well?

BenjaminBossan: 1. Indeed, SD models are quite big. Right now, we still use them in tests, see here: https://github.com/huggingface/peft/blob/7e7b55880e2e0f3c80309d4e578ee36f6ff24793/tests/test_stablediffusion.py#L28.
2. No, my idea was a bit different. The test should be like this: Create a LoRA model with an adapter, set the scale to value x and get the outputs. Then, set the scale again to x and merge the weights. After exiting the context, the weights are still merged. Now when we get the output, it should be the same as our initial output. This way, we can ensure that the temporarily set scalings are correctly accounted for when merging. Does that make sense?
3. For methods like LoRA, it's possible to add multiple adapters to the same model by calling `model.add_adapter(config2)` and activate it using `set_adapter`. There should be a test to ensure that within the scaling context, the scalings of all adapters is adjusted, and that after exiting the context, all adapter scalings are correctly reset.
4. Indeed, it looks like there is a gap in our test suite when it comes to `rank_pattern` and `alpha_pattern`. But the gist of it is quite simple: If I have a `LoraConfig(target_modules="all-linear", r=8, rank_pattern={"linear_xy": 16})`, then all linear layers will have rank 8, except for "linear_xy", which will have rank 16. Same idea for `alpha_pattern`. The test should ensure that the scaling context works for these models as well.

BenjaminBossan (commented):
> I am not sure with this one. Is there a range for alpha? If yes, we should definitely have a test for range and dtype.

I was thinking about that, but let's not put a range on alpha for now. Maybe it would make sense to demand positive values, but who knows, let the users decide. However, I would consider allowing ints, e.g. so that a user can use `set_adapter_scale(model, alpha=2)` -- it would be quite annoying if that fails.

tests/test_helpers.py
ariG23498: Thanks for the detailed reply @BenjaminBossan!

I was able to get all of the tests to work except the **merging** test. The logits after `merge_and_unload()` are not the same it seems.

Do you think I am making some mistake?

ariG23498:
@BenjaminBossan I have marked this PR for review. All the tests have been addressed except the merging test (as noted in this [reply](https://github.com/huggingface/peft/pull/1951#discussion_r1697218799)).

tests/test_helpers.py
BenjaminBossan: Merging will always result in a bit of divergence because of floating point imprecision. If the logits are identical up to a precision of `atol, rtol = 1e-4, 1e-4`, it's good enough. Could you please check that?

ariG23498: That was it! Made the changes in the current commit -- every test pass now :partying_face:

BenjaminBossan:
@ariG23498 Could you please run `make style`?

ariG23498:
Formatted the style and also rebased!

tests/test_helpers.py
```diff
@@ -66,3 +70,280 @@ def test_local_model_non_default_name(self, tmp_path):
...
+        scales_after_scaling = self.get_scale_from_modules(model)
+        for key in scales_before_scaling.keys():
+            assert scales_before_scaling[key] == scales_after_scaling[key]
+
+        with torch.no_grad():
+            logits_after_scaling = model(**inputs).logits
+
+        assert torch.allclose(logits_before_scaling, logits_after_scaling)
+
+    def test_wrong_scaling_datatype(self):
```
BenjaminBossan, line 126: Could this test be simplified to just check for the error? All the stuff with the logits doesn't really add much, right?

tests/test_helpers.py
```diff
@@ -66,3 +70,280 @@ def test_local_model_non_default_name(self, tmp_path):
...
+
+    def test_no_scaling(self):
+        base_model = AutoModelForCausalLM.from_pretrained("facebook/opt-125m")
+        tokenizer = AutoTokenizer.from_pretrained("facebook/opt-125m")
+        inputs = tokenizer("hello world", return_tensors="pt")
+
+        base_model.eval()
+
+        with torch.no_grad():
+            logits_base_model = base_model(
```
BenjaminBossan, line 179: Can be a single line.

tests/test_helpers.py
```diff
@@ -66,3 +70,280 @@ def test_local_model_non_default_name(self, tmp_path):
...
+    def test_not_lora_model(self):
+        model = AutoModelForCausalLM.from_pretrained("facebook/opt-125m")
+
+        # we expect a value error here because the model
+        # does not have lora layers
+        with pytest.raises(ValueError, match="scaling is only supported for models with `LoraLayer`s"):
+            with set_adapter_scale(model=model, alpha=0.5):
+                pass
+
+    def test_no_scaling(self):
```
BenjaminBossan, line 171: Let's rename: `test_scaling_set_to_zero`.

tests/test_helpers.py
```diff
@@ -66,3 +70,280 @@ def test_local_model_non_default_name(self, tmp_path):
...
+            lora_dropout=0.1,
+            bias="none",
+            init_lora_weights=False,
+        )
+        lora_model = get_peft_model(base_model, lora_config)
+        lora_model.eval()
+
+        with set_adapter_scale(model=lora_model, alpha=0.0):
+            with torch.no_grad():
+                logits_lora_model = lora_model(
```
BenjaminBossan, line 196: Can be a single line.

tests/test_helpers.py
```diff
@@ -66,3 +70,280 @@ def test_local_model_non_default_name(self, tmp_path):
...
+                    **inputs,
+                ).logits
+
+        assert torch.allclose(logits_base_model, logits_lora_model)
+
+    def test_diffusers_pipeline(self):
```
BenjaminBossan, line 202: Nice! Would you mind adding a similar one for transformers models loaded directly with PEFT? See: https://huggingface.co/docs/transformers/v4.43.3/en/peft#load-adapters-with--peft

tests/test_helpers.py
```diff
@@ -66,3 +70,280 @@ def test_local_model_non_default_name(self, tmp_path):
...
+            lora_dropout=0.1,
+            bias="none",
+            init_lora_weights=False,
+        )
+        model = get_peft_model(model, lora_config)
+        scales_before_scaling = self.get_scale_from_modules(model)
+
+        with set_adapter_scale(model=model, alpha=0.5):
+            model.add_adapter("other", lora_config)
+            model.set_adapter("other")
```
BenjaminBossan, line 264: Let's move these out of the context. Also, let's generate the logits for both "default" and "other" adapter with and without scaling and ensure that the scaling works for both.

tests/test_helpers.py
```diff
@@ -66,3 +70,280 @@ def test_local_model_non_default_name(self, tmp_path):
...
+        scales_after_scaling = self.get_scale_from_modules(model)
+        for key in scales_before_scaling.keys():
+            assert scales_before_scaling[key] == scales_after_scaling[key]
+
+        with torch.no_grad():
+            logits_after_scaling = model(**inputs).logits
+
+        assert torch.allclose(logits_before_scaling, logits_after_scaling)
+
+    # TODO (ariG23498): This currently fails
```
BenjaminBossan, line 320: Remove?

tests/test_helpers.py
```diff
@@ -66,3 +70,280 @@ def test_local_model_non_default_name(self, tmp_path):
...
+    def test_merging_adapter(self):
```
BenjaminBossan, line 321: Nice

BenjaminBossan (changes requested):
Thanks a lot for the updates, this looks very good already. I have a few comments but nothing big.

What would also be nice is if some documentation could be added, I think this would be the best place: https://github.com/huggingface/peft/blob/main/docs/source/package_reference/helpers.md

It's also okay to skip the docs for now and we can add it in a follow up PR, but it's important not to forget it, otherwise users will have a hard time to discover this great feature.

tests/test_helpers.py
ariG23498: Ah! That makes sense to me.

tests/test_helpers.py
ariG23498: Correct me if I am wrong, but with a directly loaded model from PEFT, there should not be any LoraLayer to check for. And that should mean, we will not be able to access the `scaling` attribute.

How would you want me to approach this?

ariG23498:
Updated the code, and also added an autodoc placeholder in the markdown file mentioned.

tests/test_helpers.py
BenjaminBossan: This should work. To give an example:

```python
# first create a peft checkpoint
config = LoraConfig(init_lora_weights=False)
model.save_pretrained("/tmp/peft/opt-lora")
del model

# load directly into transformers model
model.load_adapter("/tmp/peft/opt-lora")
type(model.model.decoder.layers[0].self_attn.v_proj)  # => peft.tuners.lora.layer.Linear
```

ariG23498:
@BenjaminBossan all the tests pass! Do you want me to rebase?

src/peft/helpers.py
```diff
@@ -146,3 +148,56 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
...
+
+    This context manager iterates over the modules of the given model and scales the `scaling` attribute of `LoraLayer`
+    instances by the specified `alpha` factor. The original scaling values are restored when the context manager exits.
+
+    Args:
+        model: The model containing `LoraLayer` modules whose scaling is to be adjusted.
+        alpha (float): The scaling factor to be applied. Must be of type float.
+
+    Raises:
+        TypeError: If `alpha` is not of type float. ValueError: If the model does not contain any `LoraLayer`
```
BenjaminBossan, line 166: Let's add a line break for the two errors. Also, the `TypeError` doesn't really need documenting IMO.

src/peft/helpers.py
```diff
@@ -146,3 +148,56 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
...
+        instances,
+                    indicating that the model does not support scaling.
+
+    Example:
+
+    ```python
+        >>> model = ModelWithLoraLayer()
```
BenjaminBossan, line 173: The example can be unindented by one level.

src/peft/helpers.py
```diff
@@ -146,3 +148,56 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
...
+
+
+@contextmanager
+def set_adapter_scale(model, alpha):
+    """
+    Context manager to temporarily set the scaling of LoraLayer modules in a model.
+
+    This context manager iterates over the modules of the given model and scales the `scaling` attribute of `LoraLayer`
+    instances by the specified `alpha` factor. The original scaling values are restored when the context manager exits.
+
```
BenjaminBossan, line 160: Let's also mention that this also works with diffusers and transformers models that have directly loaded LoRA adapters.

BenjaminBossan (changes requested):
Thanks, we're almost good now. Just a few small changes in the docstring and the tests and this PR should be good to go.

tests/test_helpers.py
```diff
@@ -223,6 +223,46 @@ def test_diffusers_pipeline(self):
...
+        model = get_peft_model(model, config)
+        model.save_pretrained(tmp_path / "opt-lora")
+        del model
+
+        # load directly into transformers model
+        model = AutoModelForCausalLM.from_pretrained(model_id)
+        model.load_adapter(tmp_path / "opt-lora")
+        # type(model.model.decoder.layers[0].self_attn.v_proj)  # => peft.tuners.lora.layer.Linear
+
+        tokenizer = AutoTokenizer.from_pretrained("facebook/opt-350m")
```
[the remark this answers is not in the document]
ariG23498: Thanks for the tip! I did not know about `pytest.fixtures`! TIL 😄

ariG23498:
Made the changes! Over to you @BenjaminBossan

tests/test_helpers.py
```diff
@@ -72,6 +72,11 @@ def test_local_model_non_default_name(self, tmp_path):
         assert result is True
 
 
+@pytest.fixture(scope="class")
```
BenjaminBossan, line 75: Please move this inside of the `TestScalingAdapters` class. Don't forget to add the `self` argument to the signature.

BenjaminBossan (changes requested):
Just a small thing, the rest looks good, thanks!

PR #1952, opened 2024-07-24
CI Fix Windows permission error on merge test

BenjaminBossan:
For some BS reason, Windows CI suddenly started throwing permission errors on test_merge_layers. These errors occur when using the `TempDirectory() `context manager, which raises a `PermissionError` on Windows when it tries to clean up after itself. Therefore, this context manager is now avoided in favor of manual clean up.

_More context:_

I investigated this issue first in #1947. My suspicion that this could be caused by a new pytest version was not confirmed. Maybe the reason is that GH rolled out a new Windows worker, not sure.

Also note that this is not the first time that this workaround is required, e.g. also here:

https://github.com/huggingface/peft/blob/e6cd24c907565040ee1766a5735afe3d13a71164/tests/test_custom_models.py#L1448-L1467

### DIFF
--- a/tests/testing_common.py
+++ b/tests/testing_common.py
@@ -16,6 +16,7 @@ import json
 import os
 import pickle
 import re
+import shutil
 import tempfile
 from collections import OrderedDict
 from dataclasses import replace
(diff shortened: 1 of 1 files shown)


sayakpaul (approved):
Windows :(

PR #1965, opened 2024-07-29
MAINT: Update ruff version to ~0.6.1

BenjaminBossan:
Changes:

- type comparisons now require is: `str is str`
- remove overridden class attribute `active_adapter`
- remove secondary import of `fbd_cuda`

_Update_: PR started with ruff v0.5.5, now upgraded to v0.6.1

Note. jupyter notebooks are excluded for now.


BenjaminBossan:
@yfeng95 it would be great if you could confirm that the `fbd_cuda` import is redundant and can be safely removed.

SunMarc (approved):
Nice ! Thanks for updating !

PR #1969, opened 2024-07-29
FIX: New bloom changes breaking prompt learning

BenjaminBossan:
Bloom had two dimensions of the attention layer transposed (compared to all other transformers models), which was fixed by:

https://github.com/huggingface/transformers/pull/31445

Therefore, for future transformers versions, skip the special handling in PEFT.

There is also an issue that prompt injection did not take place when `past_key_values` was a `Cache` object that is empty. This should now hopefully work as expected.

For the time being, `past_key_values` is still being passed as a tuple of tensors, not sure if eventually this must be a `Cache` object. If yes, this will need fixing in a future PR. Same goes for the usage of `get_seq_length()` on `Cache` objects.

Note: 
I tested this locally with transformers installed form main and bloom tests passed. PEFT CI only tests this nightly, not on PR.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -1708,9 +1708,6 @@ class PeftModelForCausalLM(PeftModel):
                 prompts = prompts.to(inputs_embeds.dtype)
                 model_kwargs["inputs_embeds"] = torch.cat((prompts, inputs_embeds), dim=1)
                 model_kwargs["input_ids"] = None
-                if peft_config.peft_type == PeftType.PREFIX_TUNING:
-                    new_past_key_values = self.get_prompt(batch_size=model_kwargs["input_ids"].shape[0])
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Would one of you @ArthurZucker or @gante be so kind to review this PR? Note that the code [here](https://github.com/huggingface/peft/pull/1969/files#diff-95787fcc5356568929e383449ade2c1bac5da43deccab17318652b62ed171ae7R1706-R1710) is just unindented, so the actual diff is smaller than it seems.

gante (approved):
LGTM 👍

gante:
@BenjaminBossan 

> not sure if eventually this must be a Cache object. 

I don't think it will, for the foreseeable future. We're out of bandwidth to convert all models :) A note: the cache class has a `to_legacy_cache()` method, PEFT can convert the `Cache` objects to the tuple format if needed

PR #1973, opened 2024-07-30
fix: lora+: include lr in optimizer kwargs

kallewoof:
Playing with the LoRA+ optimizer, I realized that the underlying optimizer will complain about the required positional `lr` argument if it is not included in the optimizer kwargs.

In [the original implementation](https://github.com/nikhil-ghosh-berkeley/loraplus/blob/a0c44bfec0eff97b9574469366bc6b0ccc28838d/lora_plus.py#L106), this was actually already the case, but somehow this fell out along the way.

### DIFF
--- a/src/peft/optimizers/loraplus.py
+++ b/src/peft/optimizers/loraplus.py
@@ -42,7 +42,6 @@ def create_loraplus_optimizer(
     Args:
         model (`torch.nn.Module`): The model to be optimized.
         optimizer_cls (`torch.optim.Optimizer`): The optimizer class to be used.
-        lr (`float`): The learning rate to be used for the optimizer.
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Thanks for fixing this issue. But isn't the better fix to add `lr` to the `kwargs`? I.e.

```diff
- lr = kwargs["lr"]
+ kwargs["lr"] = lr
```

Then we can leave `lr` in the function signature, which is more user friendly than raising a `KeyError` down the line.

kallewoof:
Ah, to make it a forced argument to `create_loraplus_optimizer`. Yes.

BenjaminBossan (approved):
Thanks for fixing this oversight.

PR #1987, opened 2024-08-02
fix: fix docker build gpus

XciD:


BenjaminBossan (approved):
Thanks a lot for fixing the docker build.

PR #1989, opened 2024-08-05
[Suggestions] Updates suggested for `helper.rescale_adapter_scale`

ariG23498:
Linked: #1940 

CC: @BenjaminBossan

### DIFF
--- a/src/peft/helpers.py
+++ b/src/peft/helpers.py
@@ -158,11 +158,10 @@ def rescale_adapter_scale(model, alpha):
     The original scaling values are restored when the context manager exits. This context manager works with the
     transformers and diffusers models that have directly loaded LoRA adapters.
 
-    For LoRA, applying this context manager with alpha in [0, 1] is strictly equivalent to
-    applying [wise-ft](https://arxiv.org/abs/2109.01903)
-    (see [#1940](https://github.com/huggingface/peft/issues/1940) for details).
-    It can improve the performances of the model if there is a distribution shiftbetween the training data used
-    for fine-tuning, and the test data used during inference.
+    For LoRA, applying this context manager with alpha in [0, 1] is strictly equivalent to applying
+    [wise-ft](https://arxiv.org/abs/2109.01903) (see [#1940](https://github.com/huggingface/peft/issues/1940) for
+    details). It can improve the performances of the model if there is a distribution shiftbetween the training data
+    used for fine-tuning, and the test data used during inference.
 
     Args:
         model: The model containing `LoraLayer` modules whose scaling is to be adjusted.
--- a/src/peft/helpers.py
+++ b/src/peft/helpers.py
@@ -151,12 +151,18 @@ def check_if_peft_model(model_name_or_path: str) -> bool:
 
 
 @contextmanager
-def set_adapter_scale(model, alpha):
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
Thanks for working on this PR so quickly.

Before going into details, I wanted to discuss something else that came to my mind, namely the argument name `alpha`. I wonder if we should rename it. Right now, the `scaling` factor is calculated as `config.lora_alpha / config.r`. I'm afraid that users could think that if they call, say, `rescale_adapter_scale(model, alpha=config.r)`, they would get a scaling of 1, as they might think that this `alpha` actually sets the `lora_alpha`. I'm not sure what a better name would be, as something like `scale` could also be confusing. WDYT? Also pinging @simjeg.

ariG23498:
Would something like `scale_multiplier` make sense? Essentially we do the exact operation.

SimJeg:
I agree using `alpha` is confusing. I proposed this based on the wise-ft paper but it will be confusing for users. I like `scale_multiplier`. @BenjaminBossan  is there any other similar variable in PEFT to have something harmonized ?

BenjaminBossan:
We have this in PEFT:

https://github.com/huggingface/peft/blob/b9260305e3e0129ba35214a40ca7ce60d7ea25b3/src/peft/tuners/lora/layer.py#L276

But as mentioned, I think "scale" is ambiguous, as it could be interpreted as the absolute value. However, I do like `scale_multiplier`, or even just `multiplier`, as I think that leaves less room for misinterpretation.

SimJeg:
@BenjaminBossan  should we add some warning if `device == "mps"` ?

BenjaminBossan:
> should we add some warning if `device == "mps"` ?

I'd prefer to add a sentence about this in the docstring until we know better what the issue really is, something along the lines of: "Warning: It has been reported that when using Apple's MPS backend for PyTorch, it is necessary to add a short sleep time after exiting the context before the scales are fully restored".

At the moment, we don't even know if this replicates for other ARM Macs and all PyTorch versions. If it's an extreme edge case, the warning could do more harm than good.

SimJeg:
Ok, once @ariG23498  add this to the docstring, I have no more comments :)

ariG23498:
Apologies for the delay! I have rebased just now.

src/peft/helpers.py
```diff
@@ -170,23 +178,23 @@ def set_adapter_scale(model, alpha):
...
     ...     outputs = model(**inputs)  # Perform operations with the scaled model
     >>> outputs = model(**inputs)  # The original scaling values are restored here
     ```
     """
-    # check if alpha has a valid data type
-    if not isinstance(alpha, (float, int)):
-        raise TypeError(f"{alpha} should be of type float, got {type(alpha)}")
+    # check if multiplier has a valid data type
+    if not isinstance(multiplier, (float, int)):
+        raise TypeError(f"{multiplier} should be of type float, got {type(multiplier)}")
```
BenjaminBossan, line 189: Oh, just a small issue I missed originally: I think it's better to make the message: `"Argument multiplier should be..."` and not use the value itself.

BenjaminBossan (approved):
LGTM. Thanks for the updates @ariG23498 and thanks @SimJeg for your valuable feedback.

Issue #1960, opened 2024-07-26
Inference with different LoRA adapters in the same batch does not use the correct module_to_save classifier


saeid93:
### System Info

Python 3.11.9
transformers==4.40.2
peft==0.11.2

### Who can help?

@BenjaminBossan 
I'm interested in using [Inference with different LoRA adapters in the same batch
](https://huggingface.co/docs/peft/en/developer_guides/lora#inference-with-different-lora-adapters-in-the-same-batch) feature, with a separate last layer classifier for each LoRA adapter. However, during the inference when having multiple requests destined to different adapter the peft library uses the active adapter for every request rather than the appropriate LoRA weight.
I should note that this problem only happens for ModuleToSave layers and the other layers (e.g. the base models) use the correct LoRA weights per each request.

This is because the per-request LoRA weights are not passed to ModulesToSaveWrapper class and the inference is always done using the active_adapter module to save:

https://github.com/huggingface/peft/blob/f2b6d13f1dbc971c7653aa65e82822ea2d84bb38/src/peft/utils/other.py#L264

The solution would be to pass the adapter_names all the way down to the forward function of ModuleToSaveWrapper `**kwargs` and write a similar logic as https://github.com/huggingface/peft/blob/f2b6d13f1dbc971c7653aa65e82822ea2d84bb38/src/peft/tuners/lora/layer.py#L327

for sending sub_batches of similar adapters together to each appropriate classifier.

However, I see that you are excluding the `special_peft_forward_args`:
https://github.com/huggingface/peft/blob/f2b6d13f1dbc971c7653aa65e82822ea2d84bb38/src/peft/peft_model.py#L761

possibly to avoid interfering with the base_models forward function e.g. 
https://github.com/huggingface/transformers/blob/5f841c74b62754f186a8c06a684d491524b7bc03/src/transformers/models/vit/modeling_vit.py#L813

I was able to solve this by modifying the mentioned functions but since it is a bug I think it can also be considered for being solved in the upstream or be mentioned in the [documentation as another caveat](https://huggingface.co/docs/peft/en/developer_guides/lora#inference-with-different-lora-adapters-in-the-same-batch). @stevhliu 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

[...]

PR #1990, opened 2024-08-05
fixing multiple LoRA in the same batch or vit

### LINKED ISSUE #1967: MobileViT does not work with Inference with different LoRA adapters in the same batch
### System Info

Python 3.11.9
transformers==4.40.2
peft==0.11.2

### Who can help?

@BenjaminBossan 

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

[MobileVit model](https://github.com/huggingface/transformers/blob/3fbaaaa64d1ef3d8327adb577994d3d11277c77a/src/transformers/models/mobilevit/modeling_mobilevit.py#L789) is not compatible with using multiple adapters in the same batch. Inferencing batches using multiple adapters using the adapter_names in the batch will trigger the following exception:

https://github.com/huggingface/peft/blob/273acf059e0f1f8bff1a3889f901475e9eb3b7ee/src/peft/tuners/lora/layer.py#L308

The root cause is that during the [unfolding operation](https://github.com/huggingface/transformers/blob/3fbaaaa64d1ef3d8327adb577994d3d11277c77a/src/transformers/models/mobilevit/modeling_mobilevit.py#L435) in the [transformers library MobileVit](https://github.com/huggingface/transformers/blob/3fbaaaa64d1ef3d8327adb577994d3d11277c77a/src/transformers/models/mobilevit/modeling_mobilevit.py#L372) the first dimension of the input is changed from `batch_size, ...` is changed to `batch_size * patch_size**2, ...` which makes it inconsistent with the `adapter_names` dimensions which is of length of `batch_size` and each entry refers to each of the batch items' adapter.

### Expected behavior

I solved this by a hack that modifies the `adapter_names` input size before sending it to the model and reverting it back to the original size for the classifier. It makes the entries proportional to the size made during the unfolding operation.

Also, we already discussed that there is a bug https://github.com/huggingface/peft/issues/1960 other than this MobileViT specific problem. Below script is the modifications needed both for https://github.com/huggingface/peft/issues/1960 and the mentioned problem together.

However, this is just a hack and I think this should work out of the box. I'm happy to investigate further when I get a chance to first solve https://github.com/huggingface/peft/issues/1960 . 

```python
# -------- changing the size of the adapter_names input ----------
        if model.base_model.model.base_model_prefix == "mobilevit":
            patch_size = model.config.patch_size
            multiply = patch_size ** 2
            resized_adapters_names = []
            for item in batch["adapter_names"]:
                multiplied = [item] * multiply
                resized_adapters_names += multiplied
            batch["adapter_names"] = resized_adapters_names
        outputs = model(**batch)

# -------- rest of the code ----------

"""
added this to solve https://github.com/huggingface/peft/issues/1960
"""

from typing import Any, Optional, Union
import torch
from torch.nn import BCEWithLogitsLoss, CrossEntropyLoss, MSELoss
from peft.peft_model import PeftModel
from transformers.modeling_outputs import ImageClassifierOutput, ImageClassifierOutputWithNoAttention
from transformers import ViTForImageClassification, MobileViTForImageClassification
from functools import partial

class PeftModelFixed(PeftModel):
    def forward(self, *args: Any, **kwargs: Any):
        """
        Forward pass of the model.
        """
        with self._enable_peft_forward_hooks(*args, **kwargs):
            # TODO removed this to avoid mixing
            # kwargs = {k: v for k, v in kwargs.items() if k not in self.special_peft_forward_args}
            return self.get_base_model()(*args, **kwargs)

class MobileViTForImageClassificationFixed(MobileViTForImageClassification):
    def forward(
        self,
        pixel_values: Optional[torch.Tensor] = None,
        output_hidden_states: Optional[bool] = None,
        labels: Optional[torch.Tensor] = None,
        return_dict: Optional[bool] = None,
        **kwargs # TODO added kwargs
    ) -> Union[tuple, ImageClassifierOutputWithNoAttention]:
        r"""
        labels (`torch.LongTensor` of shape `(batch_size,)`, *optional*):
            Labels for computing the image classification/regression loss. Indices should be in `[0, ...,
            config.num_labels - 1]`. If `config.num_labels == 1` a regression loss is computed (Mean-Square loss). If
            `config.num_labels > 1` a classification loss is computed (Cross-Entropy).
        """
        return_dict = return_dict if return_dict is not None else self.config.use_return_dict

        # TODO here
        outputs = self.mobilevit(pixel_values, output_hidden_states=output_hidden_states, return_dict=return_dict)

        pooled_output = outputs.pooler_output if return_dict else outputs[1]

        # TODO here
        adapter_names = kwargs["adapter_names"]
        patch_size = self.config.patch_size
        multiply = patch_size ** 2
        adapter_names_original = []
        for i in range(0, len(adapter_names), multiply):
            adapter_names_original.append(adapter_names[i])
        logits = self.classifier(self.dropout(pooled_output), adapter_names=adapter_names_original)

        loss = None
        if labels is not None:
            if self.config.problem_type is None:
                if self.num_labels == 1:
                    self.config.problem_type = "regression"
                elif self.num_labels > 1 and (labels.dtype == torch.long or labels.dtype == torch.int):
                    self.config.problem_type = "single_label_classification"
                else:
                    self.config.problem_type = "multi_label_classification"

            if self.config.problem_type == "regression":
                loss_fct = MSELoss()
                if self.num_labels == 1:
                    loss = loss_fct(logits.squeeze(), labels.squeeze())
                else:
                    loss = loss_fct(logits, labels)
            elif self.config.problem_type == "single_label_classification":
                loss_fct = CrossEntropyLoss()
                loss = loss_fct(logits.view(-1, self.num_labels), labels.view(-1))
            elif self.config.problem_type == "multi_label_classification":
                loss_fct = BCEWithLogitsLoss()
                loss = loss_fct(logits, labels)

        if not return_dict:
            output = (logits,) + outputs[2:]
            return ((loss,) + output) if loss is not None else output

        return ImageClassifierOutputWithNoAttention(
            loss=loss,
            logits=logits,
            hidden_states=outputs.hidden_states,
        )



def peftforward(self, *args, **kwargs):
    if self.disable_adapters or (self.active_adapter not in self.modules_to_save):
        return self.original_module(*args, **kwargs)

    # TODO changed to support LoRA
    adapter_names = kwargs["adapter_names"]
    kwargs = {}
    batch = args[0]
    unique_adapters = set(adapter_names)
    sub_batch_indices_list = []
    for adapter in unique_adapters:
        sub_batch_indices_list.append([index for index, item in enumerate(adapter_names) if item == adapter])

    results = [0 for i in range(len(batch))]
    for i, active_adapter in enumerate(unique_adapters):
        sub_batch = batch[sub_batch_indices_list[i]]
        output = self.modules_to_save[active_adapter](*(sub_batch,), **kwargs)
        for index, j in enumerate(sub_batch_indices_list[i]):
            results[j] = output[index]
    return torch.stack(results)

def change_forward_dynamically(model: PeftModel):
    model.classifier.forward = partial(peftforward, model.classifier)
    return model
        if not return_dict:
            output = (logits,) + outputs[1:]
            return ((loss,) + output) if loss is not None else output

        return ImageClassifierOutput(
            loss=loss,
            logits=logits,
            hidden_states=outputs.hidden_states,
            attentions=outputs.attentions,
        )


class MobileViTForImageClassificationFixed(MobileViTForImageClassification):
    def forward(
        self,
        pixel_values: Optional[torch.Tensor] = None,
        output_hidden_states: Optional[bool] = None,
        labels: Optional[torch.Tensor] = None,
        return_dict: Optional[bool] = None,
        **kwargs # TODO added kwargs
    ) -> Union[tuple, ImageClassifierOutputWithNoAttention]:
        r"""
        labels (`torch.LongTensor` of shape `(batch_size,)`, *optional*):
            Labels for computing the image classification/regression loss. Indices should be in `[0, ...,
            config.num_labels - 1]`. If `config.num_labels == 1` a regression loss is computed (Mean-Square loss). If
            `config.num_labels > 1` a classification loss is computed (Cross-Entropy).
        """
        return_dict = return_dict if return_dict is not None else self.config.use_return_dict

        # TODO here
        outputs = self.mobilevit(pixel_values, output_hidden_states=output_hidden_states, return_dict=return_dict)

        pooled_output = outputs.pooler_output if return_dict else outputs[1]

        # TODO here
        adapter_names = kwargs["adapter_names"]
        patch_size = self.config.patch_size
        multiply = patch_size ** 2
        adapter_names_original = []
        for i in range(0, len(adapter_names), multiply):
            adapter_names_original.append(adapter_names[i])
        logits = self.classifier(self.dropout(pooled_output), adapter_names=adapter_names_original)

        loss = None
        if labels is not None:
            if self.config.problem_type is None:
                if self.num_labels == 1:
                    self.config.problem_type = "regression"
                elif self.num_labels > 1 and (labels.dtype == torch.long or labels.dtype == torch.int):
                    self.config.problem_type = "single_label_classification"
                else:
                    self.config.problem_type = "multi_label_classification"

            if self.config.problem_type == "regression":
                loss_fct = MSELoss()
                if self.num_labels == 1:
                    loss = loss_fct(logits.squeeze(), labels.squeeze())
                else:
                    loss = loss_fct(logits, labels)
            elif self.config.problem_type == "single_label_classification":
                loss_fct = CrossEntropyLoss()
                loss = loss_fct(logits.view(-1, self.num_labels), labels.view(-1))
            elif self.config.problem_type == "multi_label_classification":
                loss_fct = BCEWithLogitsLoss()
                loss = loss_fct(logits, labels)

        if not return_dict:
            output = (logits,) + outputs[2:]
            return ((loss,) + output) if loss is not None else output

        return ImageClassifierOutputWithNoAttention(
            loss=loss,
            logits=logits,
            hidden_states=outputs.hidden_states,
        )

def peftforward(self, *args, **kwargs):
    if self.disable_adapters or (self.active_adapter not in self.modules_to_save):
        return self.original_module(*args, **kwargs)

    # TODO changed to support LoRA
    adapter_names = kwargs["adapter_names"]
    kwargs = {}
    batch = args[0]
    unique_adapters = set(adapter_names)
    sub_batch_indices_list = []
    for adapter in unique_adapters:
        sub_batch_indices_list.append([index for index, item in enumerate(adapter_names) if item == adapter])

    results = [0 for i in range(len(batch))]
    for i, active_adapter in enumerate(unique_adapters):
        sub_batch = batch[sub_batch_indices_list[i]]
        output = self.modules_to_save[active_adapter](*(sub_batch,), **kwargs)
        for index, j in enumerate(sub_batch_indices_list[i]):
            results[j] = output[index]
    return torch.stack(results)

def change_forward_dynamically(model: PeftModel):
    model.classifier.forward = partial(peftforward, model.classifier)
    return model
```

saeid93:
@BenjaminBossan 
This is the initial fix for fixing the batched LoRA inference problem explained #1960 .
For now, this only supports the vit model. This is an example and making a template for adding support for other models gradually. Generally there are two options to solve this according to the solution we discussed https://github.com/huggingface/peft/issues/1960#issuecomment-2252898911 since we need to change model specific details:

1. Change the signature of the model forward functions in the [transformer library](https://github.com/huggingface/transformers/blob/baf7e5c927744122c89ab1270c6c312541c7eb41/src/transformers/models/vit/modeling_vit.py#L813). The problem with this approach is that it needs Peft specific logic in transformers which I'm not sure is the best way for a general purpose library like transformers.
2. Change the forward function in Peft and patch it dynamically when the multiple LoRA request are in the inference batch.

I'm doing the second approach but each model needs different changes. Also, `generate` functions for generative models should be added. I'm happy to go through models one by one and also fix #1967, but it is better to review this first and then decide whether we want to go down this route of dynamically patching the forward functions or fixing it in the transformers library.

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -261,7 +261,27 @@ class ModulesToSaveWrapper(torch.nn.Module):
     def forward(self, *args, **kwargs):
         if self.disable_adapters or (self.active_adapter not in self.modules_to_save):
             return self.original_module(*args, **kwargs)
-        return self.modules_to_save[self.active_adapter](*args, **kwargs)
+        if "adapter_names" not in kwargs.keys():
+            return self.modules_to_save[self.active_adapter](*args, **kwargs)
+        # Batches requests with similar LoRAs into microbatches
+        adapter_names = kwargs["adapter_names"]
+        kwargs = {}
+        batch = args[0] # Get the batch dimension
+        unique_adapters = set(adapter_names)
+        sub_batch_indices_list = []
+        for adapter in unique_adapters:
+            sub_batch_indices_list.append(
+                [index for index, item in enumerate(adapter_names) if item == adapter]
+            )
+
+        results = [0 for i in range(len(batch))]
+        for i, active_adapter in enumerate(unique_adapters):
+            sub_batch = batch[sub_batch_indices_list[i]]
+            output = self.modules_to_save[active_adapter](*(sub_batch,), **kwargs)
+            for index, j in enumerate(sub_batch_indices_list[i]):
+                results[j] = output[index]
+        return torch.stack(results)
+
 
     def enable_adapters(self, enabled: bool):
         """Toggle the enabling and disabling of adapters
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -78,6 +78,7 @@ from .utils import (
     set_peft_model_state_dict,
     shift_tokens_right,
 )
+from .utils.patches import SUPPORTED_MULTILORA_MULTIHEAD_MAP, patch_multi_lora_forward
 
 
 PEFT_TYPE_TO_MODEL_MAPPING = {
@@ -767,6 +768,9 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
         Forward pass of the model.
         """
         with self._enable_peft_forward_hooks(*args, **kwargs):
+            # Patch the forward functiona dynamically to support LoRA weights
+            if "adapter_names" in kwargs.keys() and self.get_base_model().__class__ in SUPPORTED_MULTILORA_MULTIHEAD_MAP.keys():
+                return patch_multi_lora_forward(self.get_base_model())(*args, **kwargs)
             kwargs = {k: v for k, v in kwargs.items() if k not in self.special_peft_forward_args}
             return self.get_base_model()(*args, **kwargs)
 
--- a/src/peft/utils/patches.py
+++ b/src/peft/utils/patches.py
@@ -1,0 +1,108 @@
+# Copyright 2023-present the HuggingFace Inc. team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Optional, Union
+
+import torch
+from torch.nn import BCEWithLogitsLoss, CrossEntropyLoss, MSELoss
+from transformers import ViTForImageClassification
+from transformers.modeling_outputs import ImageClassifierOutput
+
+from . import PeftType
+
+
+class ViTForImageClassificationFixed(ViTForImageClassification):
+    def forward(
+        self,
+        pixel_values: Optional[torch.Tensor] = None,
+        head_mask: Optional[torch.Tensor] = None,
+        labels: Optional[torch.Tensor] = None,
+        output_attentions: Optional[bool] = None,
+        output_hidden_states: Optional[bool] = None,
+        interpolate_pos_encoding: Optional[bool] = None,
+        return_dict: Optional[bool] = None,
+        **kwargs
+    ) -> Union[tuple, ImageClassifierOutput]:
+        r"""
+        labels (`torch.LongTensor` of shape `(batch_size,)`, *optional*):
+            Labels for computing the image classification/regression loss. Indices should be in `[0, ...,
+            config.num_labels - 1]`. If `config.num_labels == 1` a regression loss is computed (Mean-Square loss), If
+            `config.num_labels > 1` a classification loss is computed (Cross-Entropy).
+        """
+        return_dict = (
+            return_dict if return_dict is not None else self.config.use_return_dict
+        )
+
+        outputs = self.vit(
+            pixel_values,
+            head_mask=head_mask,
+            output_attentions=output_attentions,
+            output_hidden_states=output_hidden_states,
+            interpolate_pos_encoding=interpolate_pos_encoding,
+            return_dict=return_dict,
+        )
+
+        sequence_output = outputs[0]
+
+        if "adapter_names" in kwargs.keys():
+            # This is changed to support adapter names
+            logits = self.classifier(
+                sequence_output[:, 0, :], adapter_names=kwargs["adapter_names"]
+            )
+        else:
+            logits = self.classifier(sequence_output[:, 0, :])
+
+        loss = None
+        if labels is not None:
+            # move labels to correct device to enable model parallelism
+            labels = labels.to(logits.device)
+            if self.config.problem_type is None:
+                if self.num_labels == 1:
+                    self.config.problem_type = "regression"
+                elif self.num_labels > 1 and (
+                    labels.dtype == torch.long or labels.dtype == torch.int
+                ):
+                    self.config.problem_type = "single_label_classification"
+                else:
+                    self.config.problem_type = "multi_label_classification"
+
+            if self.config.problem_type == "regression":
+                loss_fct = MSELoss()
+                if self.num_labels == 1:
+                    loss = loss_fct(logits.squeeze(), labels.squeeze())
+                else:
+                    loss = loss_fct(logits, labels)
+            elif self.config.problem_type == "single_label_classification":
+                loss_fct = CrossEntropyLoss()
+                loss = loss_fct(logits.view(-1, self.num_labels), labels.view(-1))
+            elif self.config.problem_type == "multi_label_classification":
+                loss_fct = BCEWithLogitsLoss()
+        if not return_dict:
+            output = (logits,) + outputs[1:]
+            return ((loss,) + output) if loss is not None else output
+
+        return ImageClassifierOutput(
+            loss=loss,
+            logits=logits,
+            hidden_states=outputs.hidden_states,
+            attentions=outputs.attentions,
+        )
+
+SUPPORTED_MULTILORA_MULTIHEAD_MAP = {
+    ViTForImageClassification: ViTForImageClassificationFixed
+}
+
+
+def patch_multi_lora_forward(model: PeftType.LORA) -> PeftType.LORA:
+    model.forward = SUPPORTED_MULTILORA_MULTIHEAD_MAP[model.__class__].forward.__get__(model)
+    return model


BenjaminBossan (commented):
Thanks a lot for following up on your issue. I can see that you put some thought into finding an elegant solution and I think it would be a viable solution. However, I wonder if there is another way, as sketched in my comment, please check. But I may well have missed something, LMK if my suggestion is not working.

saeid93:
No problem! Glad to be of any help.
Sorry, I couldn't find any comments on the pull request. Do you mean this comment on [MobileVit issue](https://github.com/huggingface/peft/issues/1967#issuecomment-2257868726)? If so, the problem is different from this one, this is to solve [this issue](https://github.com/huggingface/peft/issues/1960) for multiple LoRA adapters. The other [issue](https://github.com/huggingface/peft/issues/1960) is a MobileVit specific problem.

BenjaminBossan:
> Sorry, I couldn't find any comments on the pull request.

Wow, it's gone! No idea what happened, I did write it for sure...

Okay, so a second time. I was referring to these lines:

https://github.com/huggingface/peft/pull/1990/files#diff-b700510ad2034b549511a969d85f89f9094243a7f3c740e311dc1eb83ace9a79R57-R61

Are those the only real change to the `forward` function that are required? If yes, would it be possible to instead register a pre-forward hook for `classifier` to inject the argument? This could be easily achieved here:

https://github.com/huggingface/peft/blob/4611034ff85e26e1f9647ea1f505e9e50322ff0f/src/peft/tuners/lora/model.py#L434-L438

But maybe I'm missing something and more changes are necessary, or will be in the future for the issue. WDYT?

saeid93:
Yes, your suggestion worked perfectly!
Please check the new commit.

src/peft/utils/other.py
```diff
@@ -261,7 +261,27 @@ def _create_new_hook(self, old_hook):
     def forward(self, *args, **kwargs):
         if self.disable_adapters or (self.active_adapter not in self.modules_to_save):
             return self.original_module(*args, **kwargs)
-        return self.modules_to_save[self.active_adapter](*args, **kwargs)
+        if "adapter_names" not in kwargs.keys():
+            return self.modules_to_save[self.active_adapter](*args, **kwargs)
+        # Batches requests with similar LoRAs into microbatches
```
BenjaminBossan, line 266: Let's move this to a sub-method, similar to how we do this for LoRA:

https://github.com/huggingface/peft/blob/4611034ff85e26e1f9647ea1f505e9e50322ff0f/src/peft/tuners/lora/layer.py#L327

Also, with this added, I think it makes sense to have a similar method as in LoRA to check the arguments:

https://github.com/huggingface/peft/blob/4611034ff85e26e1f9647ea1f505e9e50322ff0f/src/peft/tuners/lora/layer.py#L302

Of course, we have to be careful not to be too restrictive here, given the other issue that you raised, and since the underlying module could be of any type.

src/peft/utils/other.py
```diff
@@ -261,7 +261,27 @@ def _create_new_hook(self, old_hook):
...
+        unique_adapters = set(adapter_names)
+        sub_batch_indices_list = []
+        for adapter in unique_adapters:
+            sub_batch_indices_list.append(
+                [index for index, item in enumerate(adapter_names) if item == adapter]
+            )
+
+        results = [0 for i in range(len(batch))]
+        for i, active_adapter in enumerate(unique_adapters):
+            sub_batch = batch[sub_batch_indices_list[i]]
```
BenjaminBossan, line 279: Hmm, here we assume that there is only 1 `args`, as any other `args` would be dropped, right? Also, what if other `args` or `kwargs` need to be sliced? We don't really know that so I think the best we can do is make a guess.

One suggestion that I have:

Check all `args` and `kwargs` if they're tensors and if they are a tensor, that they have the same length (i.e. batch size). In that case, slice those too. Otherwise, leave them as is. It's not perfect but I'm not sure what else could be done. WDYT?

BenjaminBossan (changes requested):
Thanks for the update, I think this version looks really nice.

I have added some comments where I think the code needs some further adjustments, please take a look. Apart from that, I have two more requests:

1. Let's update the [docs](https://github.com/huggingface/peft/blob/4611034ff85e26e1f9647ea1f505e9e50322ff0f/docs/source/developer_guides/lora.md#L323) to mention that `modules_to_save` is supported, but add the necessary caveats (depending on what we end up having in the code).
2. The unit tests should be updated to check for this use case. We don't need to test each type of possible model, but maybe a case similar to the original one with a classifier layer at the end and then perhaps one with two `modules_to_save`, say embedding and LM head. The tests could go [here](https://github.com/huggingface/peft/blob/4611034ff85e26e1f9647ea1f505e9e50322ff0f/tests/test_custom_models.py#L3126). LMK if you feel like giving this a try.

saeid93:
Sure, I'll be happy to add the tests. I'll add the updates when I get a chance.

BenjaminBossan:
@saeid93 Do you still plan on working on this?

saeid93:
> @saeid93 Do you still plan on working on this?

Yes, sorry I had some busy weeks, I'll work on it on this weekend if there isn't a strict deadline.

BenjaminBossan:
> Yes, sorry I had some busy weeks, I'll work on it on this weekend if there isn't a strict deadline.

Thanks. No worries about the time, I just wanted to ensure that you're still on it. If not, that's also okay, just let me know.

src/peft/utils/other.py
saeid93: Both of the functions are added in the new commit, please check that.

src/peft/utils/other.py
saeid93: I changed the input definition in the new version with x as the input to avoid the problems that you mentioned.

saeid93:
@BenjaminBossan I added the tests and also changed the code according to your comments, please let me know if further changes are required.

src/peft/utils/other.py
```diff
@@ -258,10 +258,50 @@ def _create_new_hook(self, old_hook):
...
+
+        if len(x) != len(adapter_names):
+            msg = (
+                "Length of `adapter_names` should be the same as the number of inputs, but got "
+                f"{len(adapter_names)} and {len(x)} respectively."
+            )
+            raise ValueError(msg)
+
+    def _mixed_batch_forward(
+        self, x: torch.Tensor, *args: Any, adapter_names: list[str], **kwargs: Any
```
BenjaminBossan, line 275: I think the issue of how to deal with the arguments is still unsolved. Right now, you're completely ignoring the `args`, which can lead to errors or even code that seems to pass but is actually incorrect. Say we have a module type with a forward function `forward(X, Y)`, where `X` and `Y` are tensors of the same size, as is, `Y` will just be dropped.

I think we have two options here:

1. We check all `args` and `kwargs` for tensors with the same length as `adapter_names` and those are being sliced. I.e. we make a best guess of what tensors are intended to be sliced and the rest is kept as is.
2. We restrict mixed batching to a handful of known modules whose signature we know and therefore can correctly handle the slicing. Most important would be `Linear`, then `Embedding`, `Conv2d` and transformers `Conv1D` would be nice to have.

Option 1 is a bit more flexible but option 2 more reliable and easier to implement. In fact, your current code is almost good for option 2, I'd just rename `x` to `input` and pass on `args`. Of course, we also need to add a type check to raise an error if the wrong module type is being used.

No matter what option is chosen, the restriction should be mentioned in the docs.

tests/test_custom_models.py
```diff
@@ -3178,12 +3178,19 @@ def test_mixed_adapter_batches_lora_mlp(self, mlp_lora):
...
         config0 = LoraConfig(target_modules=["lin0"], init_lora_weights=False)
         config1 = LoraConfig(target_modules=["lin1"], init_lora_weights=False)
         peft_model = get_peft_model(base_model, config0, "adapter0").eval()
         peft_model.add_adapter("adapter1", config1)
+        inputs = {"X": torch.arange(90).view(-1, 10).to(self.torch_device)}
+        self.run_checks(peft_model, inputs)
 
+    def test_mixed_adapter_batches_lora_different_classifiers(self, mlp_lora):
+        base_model = MLP().to(self.torch_device).eval()
+        config0 = LoraConfig(target_modules=["lin0"], modules_to_save=["sm"], init_lora_weights=False)
```
BenjaminBossan, line 3190: Here `modules_to_save` is applied to `"sm"`, which is the softmax layer. That doesn't really make sense. Let's apply it to the last linear layer, `"lin1"`.

tests/test_custom_models.py
```diff
@@ -3178,12 +3178,19 @@ def test_mixed_adapter_batches_lora_mlp(self, mlp_lora):
...
         base_model = MLP().to(self.torch_device).eval()
-        # target different lora layers
         config0 = LoraConfig(target_modules=["lin0"], init_lora_weights=False)
         config1 = LoraConfig(target_modules=["lin1"], init_lora_weights=False)
         peft_model = get_peft_model(base_model, config0, "adapter0").eval()
         peft_model.add_adapter("adapter1", config1)
+        inputs = {"X": torch.arange(90).view(-1, 10).to(self.torch_device)}
+        self.run_checks(peft_model, inputs)
 
+    def test_mixed_adapter_batches_lora_different_classifiers(self, mlp_lora):
```
BenjaminBossan, line 3188: Depending on what is decided implementation-wise, it would be good to have additional test(s):

- If embedding is supported, let's add a test with `ModelEmbConv1D`.
- If only limited module types are supported, let's add a test that check that an error with appropriate error message is raised if the wrong type is used.

Feel free to ask questions if you need help.

BenjaminBossan (changes requested):
Thanks for making the adjustments. I think there is still a bit of an issue to figure out when it comes to the signature of the mixed batch `forward` call. Please check my comments.

BenjaminBossan:
Gentle ping @saeid93

saeid93:
@BenjaminBossan sorry for the delay, I was on annual leave. I made the changes you asked for, please let me know if you need any more changes.

docs/source/developer_guides/lora.md
```diff
@@ -369,6 +369,22 @@ output = peft_model.generate(**inputs, adapter_names=adapter_names, max_new_toke
 
 Note that the order does not matter here, i.e. the samples in the batch don't need to be grouped by adapter as in the example above. We just need to ensure that the `adapter_names` argument is aligned correctly with the samples.
 
+In certain scenarios, it is necessary to replicate specific neural network layers for each set of LoRA weights. A common use case is in classification tasks, where the classification head may need to be customized for each LoRA weight. The `modules_to_save` feature allows for the creation of multiple replicas of a given layer, each corresponding to a different set of LoRA weights. For example:
```
BenjaminBossan, line 372: I think this paragraph is not super helpful as is, since it basically just explains what `modules_to_save` is for, the relation to what was described previously is not clear. I think what would be more helpful is to just mention that the feature described above also works with `modules_to_save`. A code example is not even necessary IMO.

docs/source/developer_guides/lora.md
```diff
@@ -382,3 +398,4 @@ Using this features has some drawbacks, namely:
   - Increase the batch size.
   - Try to avoid having a large number of different adapters in the same batch, prefer homogeneous batches. This can be achieved by buffering samples with the same adapter and only perform inference with a small handfull of different adapters.
   - Take a look at alternative implementations such as [LoRAX](https://github.com/predibase/lorax), [punica](https://github.com/punica-ai/punica), or [S-LoRA](https://github.com/S-LoRA/S-LoRA), which are specialized to work with a large number of different adapters.
+- The `modules_to_save` feature is currently only supported for the layers of types `Linear`, `Embedding`, `Conv2d` and `Conv1d`.
```
BenjaminBossan, line 401: Let's move this above the bullet point above, i.e. the one starting with `There is an expected overhead`.

src/peft/utils/other.py
```diff
@@ -258,10 +258,61 @@ def _create_new_hook(self, old_hook):
...
+        adapter_names = kwargs.pop("adapter_names", None)
+
         if self.disable_adapters or (self.active_adapter not in self.modules_to_save):
-            return self.original_module(*args, **kwargs)
-        return self.modules_to_save[self.active_adapter](*args, **kwargs)
+            return self.original_module(x, *args, **kwargs)
+        if adapter_names is None:
+            return self.modules_to_save[self.active_adapter](x, *args, **kwargs)
+        else:
+            return self._mixed_batch_forward(x, *args, adapter_names=adapter_names, **kwargs)
```
BenjaminBossan, line 315: More of a stylistic choice:

```suggestion
        return self._mixed_batch_forward(x, *args, adapter_names=adapter_names, **kwargs)
```

tests/test_custom_models.py
```diff
@@ -3178,15 +3201,34 @@ def test_mixed_adapter_batches_lora_mlp(self, mlp_lora):
...
```
BenjaminBossan, line 3211: Let's call this `test_mixed_adapter_batches_lora_multiple_modules_to_save`

tests/test_custom_models.py
```diff
@@ -3178,15 +3201,34 @@ def test_mixed_adapter_batches_lora_mlp(self, mlp_lora):
...
+        base_model = MLP().to(self.torch_device).eval()
+        config0 = LoraConfig(target_modules=["lin0"], modules_to_save=["lin1"], init_lora_weights=False)
+        config1 = LoraConfig(target_modules=["lin0"], modules_to_save=["lin1"], init_lora_weights=False)
+        peft_model = get_peft_model(base_model, config0, "adapter0").eval()
+        peft_model.add_adapter("adapter1", config1)
         inputs = {"X": torch.arange(90).view(-1, 10).to(self.torch_device)}
         self.run_checks(peft_model, inputs)
 
+    def test_mixed_adapter_batches_lora_unsupported_layer(self, mlp_lora):
```
BenjaminBossan, line 3220: Let's call this `test_mixed_adapter_batches_lora_unsupported_layer_raises`

tests/test_custom_models.py
```diff
@@ -3208,6 +3250,15 @@ def test_mixed_adapter_batches_lora_conv1d_emb(self):
         inputs = {"X": torch.arange(90).view(-1, 10).to(self.torch_device)}
         self.run_checks(peft_model, inputs)
 
+    def test_mixed_adapter_batches_lora_conv1d_emb_different_classifiers(self):
```
BenjaminBossan, line 3253: Let's call this `test_mixed_adapter_batches_lora_conv1d_emb_multiple_modules_to_save`

src/peft/utils/other.py
```diff
@@ -258,10 +258,61 @@ def _create_new_hook(self, old_hook):
...
+    def _mixed_batch_forward(
+        self, input: torch.Tensor, *args: Any, adapter_names: list[str], **kwargs: Any
+    ) -> torch.Tensor:
+        # This is a special method that handles the case when users pass the argument `adapter_names`. This is an
+        # extra argument that allows mixing different adapters in the same batch at inference time.
+
+        SUPPORTED_MODULES = (torch.nn.Linear, torch.nn.Embedding, torch.nn.Conv2d, torch.nn.Conv1d)
+
+        if not isinstance(self.original_module, SUPPORTED_MODULES):
+            raise TypeError("Mixed batching is only supported for Linear, Embedding, Conv2d, and Conv1D modules.")
```
BenjaminBossan, line 283: Just a nit: The list of supported layers in the error message could be constructed from `SUPPORTED_MODULES`.

BenjaminBossan (changes requested):
Thanks so much for resuming the work and making the requested adjustments. Overall this looks good, I only have a few small comments. Could you please check?

Also, could you please run `make style` to silence the linter? Note that by now, we've reached `ruff 0.6.5`, so you may have to upgrade its version in your environment.

saeid93:
@BenjaminBossan sure, glad to be of any help. I think all the comments have now been applied. I ran the style every time, probably it was the version mismatch, hopefully, it will go through this time. Let me know if further changes are needed.

BenjaminBossan:
Thanks for the latest changes @saeid93. The style check is still failing, did you successfully run `make style`? If that doesn't work for some reason, removing [this import](https://github.com/huggingface/peft/pull/1990/files#diff-4331e1b00e557c2af6e682bc221d1517f554d7ca24b9149520237b431d8d797fR49) should be sufficient.

saeid93:
@BenjaminBossan no problem! I did run the style, I'm not sure why it wasn't caught locally. I just removed the line. Hopefully, it will go through this time.

BenjaminBossan:
The linter is happy now :tada:. However, now we get an error with Python 3.8 because it does not support `list[str]` etc. Could you please add the `from __future__ import annotations` import to the top of `utils/other.py`, that should fix it.

saeid93:
Done 👍 I had to make a small change and use `tuple` instead of `Tuple` to avoid ruff complaints after adding `from __future__ import annotations`.

BenjaminBossan (approved):
Thanks a lot for extending the functionality of having different adapters in the same batch to `modules_to_save`. The changes look good, are well covered by tests, and documented. Nothing more to add!

PR #1992, opened 2024-08-06
FIX: Adjust transformers version check for bloom

BenjaminBossan:
See #1969

The fix to the bloom architecture was not actually released in transformers 4.43.3, which makes the version check invalid. Instead, now checking an attribute on the `BloomPreTrainedModel`.


gante (approved):
LGTM 👍

PR #1993, opened 2024-08-06
MAINT: Default to loading weights only for torch.load

BenjaminBossan:
The `torch.load` function allows to pass `weights_only=True`, which is more secure but may break some code that is more than just weights. For PEFT, this should not be the case, so the switch should just work.

By making the switch now, we can find out early if there are any problems, as `torch.load` will default to `True` in the future.


mikaylagawarecki (approved):
Thanks!

muellerzr (approved):
Nice, very clean :)

muellerzr:
We should do this exact thing over in accelerate I think for ease of use

PR #1994, opened 2024-08-07
BOFT bug fix when saving

Zeju1997:
fixing non-contiguous tensor when saving the model after merge_and_unload()

### DIFF
--- a/src/peft/tuners/boft/layer.py
+++ b/src/peft/tuners/boft/layer.py
@@ -510,9 +510,9 @@ class Linear(nn.Module, BOFTLayer):
                     # because of the copy operation.
                     orig_weight = base_layer.weight.data.clone()
                     butterfly_oft_mat, boft_s = self.get_delta_weight(active_adapter)
-                    orig_weight = torch.transpose(orig_weight, 0, 1)
-                    orig_weight = torch.mm(butterfly_oft_mat, orig_weight)
-                    orig_weight = torch.transpose(orig_weight, 0, 1)
+                    orig_weight = torch.transpose(orig_weight, 0, 1).contiguous()
+                    orig_weight = torch.mm(butterfly_oft_mat, orig_weight).contiguous()
+                    orig_weight = torch.transpose(orig_weight, 0, 1).contiguous()
                     orig_weight = orig_weight * boft_s
 
                     if not torch.isfinite(orig_weight).all():
@@ -524,9 +524,9 @@ class Linear(nn.Module, BOFTLayer):
                 else:
                     butterfly_oft_mat, boft_s = self.get_delta_weight(active_adapter)
                     orig_weight = base_layer.weight.data.clone()
-                    orig_weight = torch.transpose(orig_weight, 0, 1)
-                    orig_weight = torch.mm(butterfly_oft_mat, orig_weight)
-                    orig_weight = torch.transpose(orig_weight, 0, 1)
+                    orig_weight = torch.transpose(orig_weight, 0, 1).contiguous()
+                    orig_weight = torch.mm(butterfly_oft_mat, orig_weight).contiguous()
+                    orig_weight = torch.transpose(orig_weight, 0, 1).contiguous()
                     orig_weight = orig_weight * boft_s
 
                     self.base_layer.weight.data = orig_weight
--- a/src/peft/tuners/boft/layer.py
+++ b/src/peft/tuners/boft/layer.py
@@ -510,9 +510,9 @@ class Linear(nn.Module, BOFTLayer):
                     # because of the copy operation.
                     orig_weight = base_layer.weight.data.clone()
                     butterfly_oft_mat, boft_s = self.get_delta_weight(active_adapter)
-                    orig_weight = torch.transpose(orig_weight, 0, 1).contiguous()
-                    orig_weight = torch.mm(butterfly_oft_mat, orig_weight).contiguous()
-                    orig_weight = torch.transpose(orig_weight, 0, 1).contiguous()
+                    orig_weight = torch.transpose(orig_weight, 0, 1)
+                    orig_weight = torch.mm(butterfly_oft_mat, orig_weight)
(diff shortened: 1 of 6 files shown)


BenjaminBossan:
Thanks for adding this fix. Do you have an example where lack of contiguity leads to an error?

Zeju1997:
> Thanks for adding this fix. Do you have an example where lack of contiguity leads to an error?

sure, here is where I noticed the error, a small code snippet for training step:
```python
    oft_config = BOFTConfig(
        boft_block_size=args.block_size, # 32
        boft_n_butterfly_factor=args.n_butterfly_factor, # 1
        boft_dropout=0.05,
        bias="none",
        task_type="CAUSAL_LM",
    )

    train_data.start_iteration = 0

    print("Starting main loop")

    training_args = SFTConfig(
        output_dir=args.output_dir,
        dataloader_drop_last=True,
        eval_strategy="steps",
        num_train_epochs=args.num_train_epochs,
        eval_steps=args.eval_freq,
        save_strategy=args.save_strategy,
        logging_steps=args.log_freq,
        per_device_train_batch_size=args.batch_size,
        per_device_eval_batch_size=args.batch_size,
        learning_rate=args.learning_rate,
        lr_scheduler_type=args.lr_scheduler_type,
        warmup_steps=args.num_warmup_steps,
        gradient_accumulation_steps=args.gradient_accumulation_steps,
        gradient_checkpointing=args.gradient_checkpointing,
        fp16=args.fp16,
        bf16=args.bf16,
        weight_decay=args.weight_decay,
        run_name="llama-7b-finetuned",
        report_to="wandb",
        ddp_find_unused_parameters=False,
        disable_tqdm=False,
        max_seq_length=args.seq_length,
        dataset_text_field = "text",
    )

        args.model_path, 
    )

    trainer = SFTTrainer(
        model=model,
        args=training_args,
        train_dataset=train_data,
        peft_config=oft_config,
        packing=False,
    )

    print_trainable_parameters(trainer.model)

    trainer.train()
    
    print("Saving last checkpoint of the model")

    trainer.model = trainer.model.merge_and_unload()
    trainer.save_model(os.path.join(args.output_dir, str(args.num_train_epochs)))
    ```
    error occur at trainer.save_model, it states that some tensors are not  contiguous, I noticed it is because of the torch.transpose() and torch.mm() operation performed during the def merge() function

File: src/peft/tuners/boft/layer.py [hunk 1]
```diff
@@ -510,9 +510,9 @@ def merge(self, safe_merge: bool = False, adapter_names: Optional[list[str]] = N
                     # because of the copy operation.
                     orig_weight = base_layer.weight.data.clone()
                     butterfly_oft_mat, boft_s = self.get_delta_weight(active_adapter)
-                    orig_weight = torch.transpose(orig_weight, 0, 1)
-                    orig_weight = torch.mm(butterfly_oft_mat, orig_weight)
-                    orig_weight = torch.transpose(orig_weight, 0, 1)
+                    orig_weight = torch.transpose(orig_weight, 0, 1).contiguous()
+                    orig_weight = torch.mm(butterfly_oft_mat, orig_weight).contiguous()
+                    orig_weight = torch.transpose(orig_weight, 0, 1).contiguous()
```
BenjaminBossan, line 515: I wonder if we can simplify this by only calling `.contiguous()` at the very last step, i.e. line 532:

```diff
- self.base_layer.weight.data = orig_weight
+ self.base_layer.weight.data = orig_weight.contiguous()
```

Same for line 532. And since there is also `Conv2d`, should the call be added there as well? So that would be lines 820 and 834.

BenjaminBossan (changes requested):
> sure, here is where I noticed the error, a small code snippet for training step:

Thanks, I can confirm that this raises the error. Let's add a unit test to ensure that this works correctly. I think the best location would be after this line:

https://github.com/huggingface/peft/blob/c86966489144a4ec101ab5e53abc63631270e80e/tests/testing_common.py#L764

Here is the code that I used:

```python
        # serializing works without errors
            # serializing with torch.save works
            torch.save(model_unloaded.state_dict(), os.path.join(tmp_dirname, "model.bin"))

            # serializing with safetensors works
            save_file(model_unloaded.state_dict(), os.path.join(tmp_dirname, "model.safetensors"))
```

The `save_file` function must be imported: `from safetensors.torch import save_file`.

When I ran these extended tests, I noticed that OFT is also failing. However, this can be resolved in the same way in this line:

https://github.com/huggingface/peft/blob/c86966489144a4ec101ab5e53abc63631270e80e/src/peft/tuners/oft/layer.py#L174

```diff
-                base_layer.weight.data = new_weights
+                base_layer.weight.data = new_weights.contiguous()
```

Would you be so kind to fix that too?

tests/testing_common.py
```diff
@@ -763,6 +763,10 @@ def _test_safe_merge(self, model_id, config_cls, config_kwargs):
         # check that the logits are the same after unloading
         assert torch.allclose(logits_peft, logits_unloaded, atol=atol, rtol=rtol)
 
+        # serializing with safetensors works
+        from safetensors.torch import save_file
```
BenjaminBossan, line 767: Let's move this import to the top of the file.

BenjaminBossan (changes requested):
Thanks for the updates. Just two very small change requests, otherwise this looks good.

Zeju1997:
Can you check again? I noticed in the unit test I did not copy the whole test in the previous commit. Best.

BenjaminBossan:
I think your last commit was the missing piece :crossed_fingers:

Zeju1997:
@BenjaminBossan Hi, the tests are still failing, are they because of my added code?

BenjaminBossan:
> Hi, the tests are still failing, are they because of my added code?

No, I don't think so, this looks more like a HF Hub issue. This can sometimes happen, but not with the whole test matrix. I'll restart later, don't worry about it.

BenjaminBossan (approved):
Thanks for the fixes to merging BOFT and OFT.

PR #1995, opened 2024-08-08
FIX Import error in BOFT half precision test

BenjaminBossan:
There was a small last minute change in #1925 that broke the tests. It should be `boft.layer.Linear` and not `boft.Linear`. As these are GPU tests, CI didn't catch this and I did not run this locally to verify.

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -3093,7 +3093,7 @@ class TestBOFT:
     def test_boft_half_linear(self):
         # Check that we can use BoFT with model loaded in half precision
         layer = torch.nn.Linear(160, 160).cuda()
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
FYI @Edenzzzz the nightly tests on GPU failed, this should fix them.

sayakpaul (approved):
Thanks. Maybe in the future we could do:
https://github.com/huggingface/diffusers/blob/main/.github/workflows/run_tests_from_a_pr.yml

This will allow us to trigger desired tests on the GPU runner.

BenjaminBossan:
> Thanks. Maybe in the future we could do: https://github.com/huggingface/diffusers/blob/main/.github/workflows/run_tests_from_a_pr.yml
> 
> This will allow us to trigger desired tests on the GPU runner.

Normally I test GPU tests locally but as this was a late change, I forgot to re-run the tests, so it was just a personal oversight.

PR #1996, opened 2024-08-08
Update OFT to fix merge bugs

Zeju1997:
# changes for BOFT
- fixing small errors in comments of BOFT
- making Conv2D operation consistent with Linear

# changes for OFT
- [Important!] fixing the incorrect merging of OFT, it is not an implementation error, but more of an incorrect understanding of how orthogonal fine-tuning works, resulting in a wrong way of merging adapters
- update config.py/layer.py/model.py file to be consistent with other peft methods
- adding additional paramters oft_block_size, oft_dropout, bias
- fixing module dropout implementation error (in orthogonal finetuning, dropout is not the same as LoRA)
- fixing errors in merge and unmerge function in OFT 
- making Conv2D operation consistent with Linear


src/peft/tuners/boft/layer.py
```diff
@@ -315,11 +315,11 @@ def update_layer(
...
 
+        elif boft_block_size != 0 and boft_block_num != 0:
+            raise ValueError(f"You can only specify either boft_block_size ({boft_block_size}) or boft_block_num ({boft_block_num}), but not both simultaneously.")
+
         else:
-            raise ValueError(
-                f"You can only specify either boft_block_size ({boft_block_size}) or boft_block_num ({boft_block_num}), but not both simultaneously or setting both"
-                "to be 0, because boft_block_size x boft_block_num != in_features."
-            )
+            raise ValueError(f"Either `boft_block_size` or `boft_block_num` must be non-zero. Currently, boft_block_size = {boft_block_size} and boft_block_num = {boft_block_num}.")
```
BenjaminBossan, line 322: Do we need these checks? We already have checks here:

https://github.com/huggingface/peft/blob/41c274ecac6247a63e3b10f8536904f3ead82213/src/peft/tuners/boft/config.py#L127-L133

src/peft/tuners/boft/layer.py
```diff
@@ -752,8 +747,12 @@ def update_layer(
 
             boft_block_num = int(conv_filter_dim // boft_block_size)
 
+        elif boft_block_size != 0 and boft_block_num != 0:
+            raise ValueError(f"You can only specify either boft_block_size ({boft_block_size}) or boft_block_num ({boft_block_num}), but not both simultaneously.")
+
         else:
-            raise ValueError("Unknown error!")
+            raise ValueError(f"Either `boft_block_size` or `boft_block_num` must be non-zero. Currently, boft_block_size = {boft_block_size} and boft_block_num = {boft_block_num}.")
```
BenjaminBossan, line 754: Same comment.

src/peft/tuners/oft/layer.py
```diff
@@ -14,106 +14,293 @@
...
+import torch.nn.functional as F
 
-from peft.tuners.lycoris_utils import LycorisLayer, check_adapters_to_merge
+from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
 
 
-class OFTLayer(nn.Module, LycorisLayer):
+class MultiplicativeDropoutLayer(nn.Module):
+    """
+    Implements the multiplicative dropout layer for BOFT.
```
BenjaminBossan, line 28:
```suggestion
    Implements the multiplicative dropout layer for OFT.
```

src/peft/tuners/oft/layer.py
```diff
@@ -14,106 +14,293 @@
...
+            if x.shape[-1] != x.shape[-2]:
+                raise ValueError("The last two dimensions of input should be the same!")
+
+            D, H, _ = x.shape
+
+            # If block share, skip the multiplicative dropout
+            if D == 1:
+                return x
+
+            # Create a mask with 1s for matrices to be replaced with identity and 0s otherwise
```
BenjaminBossan, line 61: Sentence doesn't quite make sense, could you please check again?

src/peft/tuners/oft/layer.py
```diff
@@ -14,106 +14,293 @@
...
+
+            # Generate a flat tensor with desired number of 1s and 0s
+            mask = torch.cat([torch.ones(num_to_replace, device=x.device), torch.zeros(num_zeros, device=x.device)])
+
+            # Shuffle and reshape the mask
+            mask = mask[torch.randperm(D)].view(D, 1, 1)
+
+            # Use the mask to combine original matrices and identity matrices
+            eye_matrix = torch.eye(H, device=x.device).repeat(D, 1, 1)
+            x = (1 - mask) * x + mask * eye_matrix
```
BenjaminBossan, line 73: Just for my understanding, do we need some kind of scaling like 1/(1 - p) in normal dropout to ensure that there is no scale shift between training and inference?

src/peft/tuners/oft/layer.py
```diff
@@ -14,106 +14,293 @@
...
-        if r <= 0:
-            raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
+        # Initialize the MultiplicativeDropoutLayer for module_dropout > 0.0.
+        if module_dropout > 0.0:
+            oft_dropout_layer = MultiplicativeDropoutLayer(p=module_dropout)
+        else:
+            oft_dropout_layer = nn.Identity()
+        self.oft_dropout.update(nn.ModuleDict({adapter_name: oft_dropout_layer}))
+
+        if r == 0 and oft_block_size != 0:
```
BenjaminBossan, line 178: Can we move all these checks into the `OftConfig.__post_init__` instead? We want to find these errors as early as possible.

src/peft/tuners/oft/layer.py
```diff
@@ -282,100 +404,320 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
...
+            oft_dropout_layer = nn.Identity()
+        self.oft_dropout.update(nn.ModuleDict({adapter_name: oft_dropout_layer}))
 
+        # layer information from the base layer
         base_layer = self.get_base_layer()
-        base_weight = base_layer.weight.data
-        delta_weight = delta_weight[: base_weight.shape[0], : base_weight.shape[0]]
+        conv_filter_dim = self.in_features * base_layer.kernel_size[0] * base_layer.kernel_size[0]
+
+        if r == 0 and oft_block_size != 0:
```
BenjaminBossan, line 501: Same comment as above, could this be done in the `OFTConfig` init?

BenjaminBossan (changes requested):
Thanks a lot for fixing the issues you mentioned with OFT. I don't understand all the details but in general, this is to bring OFT closer to the BOFT implementation, is that right? Is that also why you decided to no longer use `LycorisTuner` as a base?

Did you check whether after these changes, the same code using OFT would still produce the exact same results? E.g. let's say a user has trained an OFT model checkpoint in PEFT 0.12. Now we merge this PR and release it in PEFT 0.13. If the user upgrades to 0.13 and loads the OFT checkpoint, would they still get the same results? This is really important because otherwise it would be a source of bugs.

If we cannot ensure the same results after this PR because, for instance, something was not implemented correctly previously, at the very least we have to add some mechanism to error or warn users when they try to load an old checkpoint with new PEFT versions.

I have added some smaller comments too, please check them out. On top of those, please run make style and ensure 120 character line limits (e.g. long strings).

Also, pinging @okotaku and @lukaskuhn-lku since they worked on the original OFT PR.

src/peft/tuners/boft/layer.py
Zeju1997: Yes, you are right, I forgot there is already a check in the config.py Updated.

src/peft/tuners/boft/layer.py
Zeju1997: Updated.

src/peft/tuners/oft/layer.py
Zeju1997: Yes. Updated.

src/peft/tuners/oft/layer.py
Zeju1997: Yes. Updated.

src/peft/tuners/oft/layer.py
Zeju1997: updated.

src/peft/tuners/oft/layer.py
Zeju1997: updated.

src/peft/tuners/oft/layer.py
Zeju1997: Thanks for the comment, actually this catched by self.training, meaning in inference there is never multiplicative dropout applied. (It will set certain blocks as identity and cannot be trained during inference.) Best.

Zeju1997:
@BenjaminBossan Hi, thank you so much. Because the main change (beside changing the format) is in the merge/unmerge, it should be possible to update it without affecting previous trained model checkpoints. I am currently testing it. But I am currently facing a problem:

```python
import transformers
import accelerate
import peft
from transformers import AutoImageProcessor
from peft import LoraConfig, BOFTConfig, OFTConfig, get_peft_model
import numpy as np
import evaluate
import torch
import os

from torchvision.transforms import (
    CenterCrop,
    Compose,
    Normalize,
    RandomHorizontalFlip,
    RandomResizedCrop,
    Resize,
    ToTensor,
)

from transformers import AutoModelForImageClassification, TrainingArguments, Trainer
from peft import PeftConfig, PeftModel, AutoPeftModel

os.environ["HF_API_TOKEN"] = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"

if __name__ == '__main__':
    dataset = load_dataset("food101", split="train[:5000]")

    output_dir = "output"

    model_checkpoint = "google/vit-base-patch16-224-in21k"

    labels = dataset.features["label"].names
    label2id, id2label = dict(), dict()
    for i, label in enumerate(labels):
        label2id[label] = i
        id2label[i] = label

    image_processor = AutoImageProcessor.from_pretrained(model_checkpoint)


    normalize = Normalize(mean=image_processor.image_mean, std=image_processor.image_std)
    train_transforms = Compose(
        [
            RandomResizedCrop(image_processor.size["height"]),
            RandomHorizontalFlip(),
            ToTensor(),
            normalize,
        ]
    )

    val_transforms = Compose(
        [
            Resize(image_processor.size["height"]),
            CenterCrop(image_processor.size["height"]),
            ToTensor(),
            normalize,
        ]
    )


    def preprocess_train(example_batch):
        """Apply train_transforms across a batch."""
        example_batch["pixel_values"] = [train_transforms(image.convert("RGB")) for image in example_batch["image"]]
        return example_batch


    def preprocess_val(example_batch):
        """Apply val_transforms across a batch."""
        example_batch["pixel_values"] = [val_transforms(image.convert("RGB")) for image in example_batch["image"]]
        return example_batch


    splits = dataset.train_test_split(test_size=0.1)
    train_ds = splits["train"]
    val_ds = splits["test"]

    train_ds.set_transform(preprocess_train)
    val_ds.set_transform(preprocess_val)


    def print_trainable_parameters(model):
        trainable_params = 0
        all_param = 0
        for _, param in model.named_parameters():
            all_param += param.numel()
            if param.requires_grad:
                trainable_params += param.numel()
        print(
            f"trainable params: {trainable_params} || all params: {all_param} || trainable%: {100 * trainable_params / all_param:.2f}"
        )

    model = AutoModelForImageClassification.from_pretrained(
        model_checkpoint,
        label2id=label2id,
        id2label=id2label,
        ignore_mismatched_sizes=True,  # provide this in case you're planning to fine-tune an already fine-tuned checkpoint
    )


    print_trainable_parameters(model)

    config = LoraConfig(
        r=32,
        lora_alpha=16,
        lora_dropout=0.1,
        bias="none",
        # modules_to_save=["classifier"],
    )

    lora_model = get_peft_model(model, config)
    print_trainable_parameters(lora_model)

    model_name = model_checkpoint.split("/")[-1]
    batch_size = 128

    args = TrainingArguments(
        f"{model_name}-finetuned-lora-food101",
        remove_unused_columns=False,
        evaluation_strategy="epoch",
        save_strategy="no",
        learning_rate=5e-3,
        per_device_train_batch_size=batch_size,
        gradient_accumulation_steps=1,
        per_device_eval_batch_size=batch_size,
        fp16=True,
        num_train_epochs=5,
        logging_steps=10,
        metric_for_best_model="accuracy",
        label_names=["labels"],
        eval_strategy="steps",
        eval_steps=100000000000,
    )

    metric = evaluate.load("accuracy")

    def compute_metrics(eval_pred):
        """Computes accuracy on a batch of predictions"""
        predictions = np.argmax(eval_pred.predictions, axis=1)
        return metric.compute(predictions=predictions, references=eval_pred.label_ids)

    def collate_fn(examples):
        pixel_values = torch.stack([example["pixel_values"] for example in examples])
        labels = torch.tensor([example["label"] for example in examples])
        return {"pixel_values": pixel_values, "labels": labels}

    trainer = Trainer(
        lora_model,
        args,
        train_dataset=train_ds,
        eval_dataset=val_ds,
        tokenizer=image_processor,
        data_collator=collate_fn,
    )
    eval_results = trainer.evaluate(val_ds)
    print('Before training', eval_results)

    repo_name = f"sgp-bench/peft-finetuned-lora-food101"

    train_results = trainer.train()
    
    token = os.getenv("HF_API_TOKEN")
    lora_model.push_to_hub(repo_name, token=token)

    eval_results = trainer.evaluate(val_ds)
    print('After training', eval_results)

    # config = PeftConfig.from_pretrained(repo_name)
    model = AutoModelForImageClassification.from_pretrained(
        config.base_model_name_or_path,
        label2id=label2id,
        id2label=id2label,
        ignore_mismatched_sizes=True,  # provide this in case you're planning to fine-tune an already fine-tuned checkpoint
    )
    # Load the LoRA model
    # inference_model = AutoPeftModel.from_pretrained(repo_name).to("cuda")
    trainer.model = PeftModel.from_pretrained(model, repo_name).to("cuda")
    eval_results = trainer.evaluate(val_ds)
    print('Load adapter weight and inference', eval_results)

```
I am currently debugging with LoRA and the official guide from doc.
I noticed that the adapter weight can be loaded, but the eval_accuracy is basically 0 (compared to the eval_accuracy after training of over 90%), indicating the weights are not loaded correctly?
If I do merge_and_unload(), the eval_accuracy is the same as after training.
Is there something wrong with the way I am loading adapter weights?

BenjaminBossan:
> I noticed that the adapter weight can be loaded, but the eval_accuracy is basically 0 (compared to the eval_accuracy after training of over 90%), indicating the weights are not loaded correctly?

I could reproduce your issue of low accuracy (I did not train, just loaded your checkpoint). I think the issue is this line, right?

> `# modules_to_save=["classifier"]`

For this reason, the classifier layer is randomly initialized, so the accuracy is low.

> If I do merge_and_unload(), the eval_accuracy is the same as after training.

Could you show me which steps you take to achieve this?

Zeju1997:
Ah, ok, you are right, in the old checkpoint with merge_and_unload I did have the classifier saved. 

It should be working now, I have tested it can load an old checkpoint using previous version of PEFT OFT.

src/peft/tuners/boft/layer.py
```diff
@@ -316,10 +316,7 @@ def update_layer(
             boft_block_num = int(self.in_features // boft_block_size)
 
         else:
-            raise ValueError(
-                f"You can only specify either boft_block_size ({boft_block_size}) or boft_block_num ({boft_block_num}), but not both simultaneously or setting both"
-                "to be 0, because boft_block_size x boft_block_num != in_features."
-            )
+            raise ValueError("Unknown error!")
```
BenjaminBossan, line 319: So this code should not be reachable. Still, let's use a more helpful error message. My proposal:

"Something went wrong, please report this error: https://github.com/huggingface/peft/issues".

src/peft/tuners/boft/layer.py
```diff
@@ -777,7 +770,7 @@ def update_layer(
         self.boft_R[adapter_name] = nn.Parameter(
             torch.zeros(boft_n_butterfly_factor + 1, boft_block_num, boft_block_size, boft_block_size)
         )
-        self.boft_s[adapter_name] = nn.Parameter(torch.ones(1, int(self.out_features)))
+        self.boft_s[adapter_name] = nn.Parameter(torch.ones(int(self.out_features), 1))
```
BenjaminBossan, line 773: So this transposes the `boft_s` vector. Could you explain why that is important? I was wondering if this would prevent the state dict from being loaded successfully but it seems it's not an issue (see my longer comment).

src/peft/tuners/oft/layer.py
```diff
@@ -14,106 +14,286 @@
...
+        Note, currently only support linear layer and convolutional layer, with further support for other layers to be
+        added soon.
 
+        Parameters:
+        base_layer: the pretrained model layer
+        """
+        self.base_layer = base_layer
         # OFT info
         self.oft_r = nn.ParameterDict({})
+        self.oft_s = nn.ParameterDict({})
```
BenjaminBossan, line 93: Here we introduce a new parameter `oft_s`. All existing checkpoints don't have this parameter, so that means when a user loads an old checkpoint, this parameter is randomly initialized, which means the checkpoint behaves different. See my longer comment on compatibility.

src/peft/tuners/oft/layer.py
```diff
@@ -14,106 +14,286 @@
...
+                raise ValueError(f"Input features ({self.in_features}) should be divisible by `oft_block_size` ({oft_block_size})")
+            r = int(self.in_features // oft_block_size)
+        elif r != 0 and oft_block_size == 0:
+            if self.in_features % r != 0:
+                raise ValueError(f"Input features ({self.in_features}) should be divisible by `r` ({r})!")
+            oft_block_size = int(self.in_features // r)
+        elif r != 0 and oft_block_size != 0:
+            raise ValueError(f"You can only specify either r ({r}) or oft_block_size ({oft_block_size}), but not both simultaneously.")
+        else:
+            raise ValueError(f"Either `r` or `oft_block_size` must be non-zero. Currently, r = {r} and oft_block_size = {oft_block_size}.")
```
BenjaminBossan, line 182: These lines should not be reachable because of the checks we have in the config class, but IMO it's fine to leave them here.

src/peft/tuners/oft/layer.py
```diff
@@ -282,100 +397,320 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
...
+        elif r != 0 and oft_block_size == 0:
+            if conv_filter_dim % r != 0:
+                raise ValueError(
+                    f"Convolutional kernel dimension ({conv_filter_dim}) must be divisible by r ({r})!"
+                )
+            oft_block_size = int(conv_filter_dim // r)
+        elif r != 0 and oft_block_size != 0:
+            raise ValueError(f"You can only specify either r ({r}) or oft_block_size ({oft_block_size}), but not both simultaneously.")
+        else:
+            raise ValueError(f"Either `r` or `oft_block_size` must be non-zero. Currently, r = {r} and oft_block_size = {oft_block_size}.")
```
BenjaminBossan, line 509: Same comment as earlier: Should be unreachable but it's fine if we leave the code.

tests/test_custom_models.py
```diff
@@ -228,25 +228,26 @@
     ########
     # OFT #
     ########
-    ("Vanilla MLP 1 OFT", "MLP", OFTConfig, {"target_modules": "lin0"}),
-    ("Vanilla MLP 2 OFT", "MLP", OFTConfig, {"target_modules": ["lin0"]}),
-    ("Vanilla MLP 5 OFT", "MLP", OFTConfig, {"target_modules": ["lin0"], "modules_to_save": ["lin1"]}),
+    ("Vanilla MLP 1 OFT", "MLP", OFTConfig, {"r": 2, "target_modules": "lin0"}),
+    ("Vanilla MLP 2 OFT", "MLP", OFTConfig, {"r": 2, "target_modules": ["lin0"], "r": 2}),
```
BenjaminBossan, line 232:
```suggestion
    ("Vanilla MLP 2 OFT", "MLP", OFTConfig, {"r": 2, "target_modules": ["lin0"]}),
```

BenjaminBossan (changes requested):
Thanks a lot for the updates. Could you please run `make style` so that CI can pass successfully?

I did some further tests regarding the backwards compatibility of the changes in this PR. If my testing is correct, this PR works without a problem for BOFT but it will cause an issue with OFT, namely that old OFT checkpoints will not work anymore after this PR. They will load fine but the result will be different, which is bad, as users will not expected this. We need to either ensure that the old checkpoint works as expected, or we have to add a check and if we detect an old checkpoint, we have to raise an error and tell the user that they either need to re-train their OFT checkpoint or use an older PEFT version.

Here is how I tested the compatibility:

```python
# try loading old oft and boft checkpoints
# First checkout and pull all branches to be tested. Then call like this:
# git checkout <branch 1>  # e.g. main
# python script.py oft     # either oft or boft
# git checkout <branch 2>  # e.g. PR branch
# python script.py oft     # same method as above
# call `rm -r /tmp/peft/` to try again with different branches

import os
import sys

import torch
from peft import BOFTConfig, OFTConfig, PeftModel, get_peft_model

inputs = torch.arange(5).view(-1, 1)

# oft
def main(method):
    if method.lower() == "oft":
        config_cls = OFTConfig
        path = "/tmp/peft/oft"
    else:
        config_cls = BOFTConfig
        path = "/tmp/peft/boft"

    if not os.path.exists(path + "/adapter_model.safetensors"):
        print(f"{method} adapter does not exist yet, creating it")
        peft_config = config_cls(target_modules=["q_proj", "v_proj"], init_weights=False)
        model.save_pretrained(path)

        model.eval()
        with torch.inference_mode():
            outputs = model(inputs).logits
        torch.save(outputs, path + "/output.pt")
        print(f"{method} adapter and output created saved in {path}")
    else:
        print(f"{method} adapter and outpput exist, loading it from {path}")
        model = PeftModel.from_pretrained(model, path)
        print(f"{method} adapter loaded successfully")
        model.eval()
        with torch.inference_mode():
            outputs_new = model(inputs).logits

        outputs_old = torch.load(path + "/output.pt", weights_only=True)
        assert torch.allclose(outputs_old, outputs_new, atol=1e-4, rtol=1e-4)
        print(f"{method} outputs match")


if __name__ == "__main__":
    method = sys.argv[1] if len(sys.argv) > 1 else "boft"
    assert method.lower() in ["boft", "oft"]
    main(method)
```

Does that make sense to you? I summarized the results from my tests in this table:

| branch 1 | branch 2 | method | status |
| --- | --- | --- | --- |
| main | PR | OFT | fail: outputs different |
| PR | main | OFT | `TypeError: OFTConfig.__init__() got an unexpected keyword argument 'bias'` |
| main | PR | BOFT | success |
| PR | main | BOFT | success |


In general, we really want to retain _backwards compatibility_ but in my book it's acceptable if the previous implementation was not correct and is now fixed. Also note that a new OFT checkpoint created after this PR cannot be loaded with older PEFT versions -- there is no _forward compatibility_ -- but this is okay.

Zeju1997:
> Thanks a lot for the updates. Could you please run `make style` so that CI can pass successfully?
> 
> I did some further tests regarding the backwards compatibility of the changes in this PR. If my testing is correct, this PR works without a problem for BOFT but it will cause an issue with OFT, namely that old OFT checkpoints will not work anymore after this PR. They will load fine but the result will be different, which is bad, as users will not expected this. We need to either ensure that the old checkpoint works as expected, or we have to add a check and if we detect an old checkpoint, we have to raise an error and tell the user that they either need to re-train their OFT checkpoint or use an older PEFT version.
> 
> Here is how I tested the compatibility:
> 
> ```python
> # try loading old oft and boft checkpoints
> # First checkout and pull all branches to be tested. Then call like this:
> # git checkout <branch 1>  # e.g. main
> # python script.py oft     # either oft or boft
> # git checkout <branch 2>  # e.g. PR branch
> # python script.py oft     # same method as above
> # call `rm -r /tmp/peft/` to try again with different branches
> 
> import os
> import sys
> 
> import torch
> from transformers import AutoModelForCausalLM
> from peft import BOFTConfig, OFTConfig, PeftModel, get_peft_model
> 
> model_id = "facebook/opt-125m"
> inputs = torch.arange(5).view(-1, 1)
> 
> # oft
> def main(method):
>     if method.lower() == "oft":
>         config_cls = OFTConfig
>         path = "/tmp/peft/oft"
>     else:
>         config_cls = BOFTConfig
>         path = "/tmp/peft/boft"
> 
>     model = AutoModelForCausalLM.from_pretrained(model_id)
>     if not os.path.exists(path + "/adapter_model.safetensors"):
>         print(f"{method} adapter does not exist yet, creating it")
>         peft_config = config_cls(target_modules=["q_proj", "v_proj"], init_weights=False)
>         model = get_peft_model(model, peft_config)
>         model.save_pretrained(path)
> 
>         model.eval()
>         with torch.inference_mode():
>             outputs = model(inputs).logits
>         torch.save(outputs, path + "/output.pt")
>         print(f"{method} adapter and output created saved in {path}")
>     else:
>         print(f"{method} adapter and outpput exist, loading it from {path}")
>         model = PeftModel.from_pretrained(model, path)
>         print(f"{method} adapter loaded successfully")
>         model.eval()
>         with torch.inference_mode():
>             outputs_new = model(inputs).logits
> 
>         outputs_old = torch.load(path + "/output.pt", weights_only=True)
>         assert torch.allclose(outputs_old, outputs_new, atol=1e-4, rtol=1e-4)
>         print(f"{method} outputs match")
> 
> 
> if __name__ == "__main__":
>     method = sys.argv[1] if len(sys.argv) > 1 else "boft"
>     assert method.lower() in ["boft", "oft"]
>     main(method)
> ```
> 
> Does that make sense to you? I summarized the results from my tests in this table:
> 
> branch 1	branch 2	method	status
> main	PR	OFT	fail: outputs different
> PR	main	OFT	`TypeError: OFTConfig.__init__() got an unexpected keyword argument 'bias'`
> main	PR	BOFT	success
> PR	main	BOFT	success
> In general, we really want to retain _backwards compatibility_ but in my book it's acceptable if the previous implementation was not correct and is now fixed. Also note that a new OFT checkpoint created after this PR cannot be loaded with older PEFT versions -- there is no _forward compatibility_ -- but this is okay.

Yes, this should be expected, the reason is that the load is not correct in the previous version. I think the best way is to do "we have to add a check and if we detect an old checkpoint, we have to raise an error and tell the user that they either need to re-train their OFT checkpoint or use an older PEFT version." this. Where do you suggest to add the check?

src/peft/tuners/oft/layer.py
Zeju1997: Updated.

src/peft/tuners/oft/layer.py
Zeju1997: Updated.

tests/test_custom_models.py
Zeju1997: Updated.

src/peft/tuners/oft/layer.py
Zeju1997: Actually, it depends on the init_weights, if init_weights=True (which should be the default behavior), then oft_s is basically like an identity and should not make changes. If init_weights=False, then it is randomly initialized.

BenjaminBossan: True, but when users load checkpoints, they have trained the model, so `oft_s` won't be an identity transform.

Zeju1997:
@BenjaminBossan Sorry for the late reply, I added the check_kwargs, do you think it is ok like that? As for the oft_s, if the user used an older version, it would be not existent, therefore, even if it is trained, it will still be an identity transform, therefore should make no difference.

src/peft/config.py
```diff
@@ -213,6 +214,21 @@ def _get_peft_type(
         loaded_attributes = cls.from_json_file(config_file)
         return loaded_attributes["peft_type"]
 
+    @classmethod
+    def check_kwargs(cls, **kwargs):
```
BenjaminBossan, line 218: Let's move this to `OFTConfig`. Here, just return `kwargs` without any further change.

src/peft/config.py
```diff
@@ -213,6 +214,21 @@ def _get_peft_type(
...
+        r"""
+        Check if the kwargs are valid for the configuration.
+
+        Args:
+            kwargs (additional keyword arguments, *optional*):
+                Additional keyword arguments passed along to the child class initialization.
+        """
+        if "oft_block_size" in kwargs:
+            warnings.warn(
+                'OFT has been updated since 0.12.1.dev0. Your trained adapter weights may not be compatible with the latest version of OFT. Please retrain your adapter weights.')
```
BenjaminBossan, line 228: 1. Note that `dev0` is not a real version, so let's not refer to that. The next release version will be 0.13.0, so let' use that version.
2. "may not be compatible": We are pretty sure it is incompatible when trained, right? Let's phrase it as "is incompatible".
3. Let's also mention that users can downgrade PEFT to version 0.12.0 and then the adapter will still work.

src/peft/config.py
```diff
@@ -213,6 +214,21 @@ def _get_peft_type(
...
+    @classmethod
+    def check_kwargs(cls, **kwargs):
+        r"""
+        Check if the kwargs are valid for the configuration.
+
+        Args:
+            kwargs (additional keyword arguments, *optional*):
+                Additional keyword arguments passed along to the child class initialization.
+        """
+        if "oft_block_size" in kwargs:
```
BenjaminBossan, line 226: Hmm, so `oft_block_size` is the new parameter, right? So if we're loading an old OFT model, it should be missing. Therefore, should the check not be `if "oft_block_size" not in kwargs`?

BenjaminBossan (changes requested):
Thanks for making these latest changes.

Concerning the compatibility check, I have a few comments, please check.

Also, I think it would be good to have a unit test for this check. I created a checkpoint with the old OFT implementation here: https://huggingface.co/peft-internal-testing/OFT-tiny-random-OPTForCausalLM. You can use it to load and then check that the warning has been raised.

src/peft/config.py
Zeju1997: @BenjaminBossan What do you mean exactly? Where should I put the check_kwargs?

BenjaminBossan: Sorry, let me be more precise.

Inside of `PeftConfigMixin`, let's define:

```python
    @classmethod
    def check_kwargs(cls, **kwargs):
        """Check kwargs before initializing the config instance.

        Subclasses can override this method to add specific checks.

        """
        return kwargs
```

This way, this doesn't do anything for all methods that are not OFT. The `PeftConfigMixin` should not have to worry about OFT implementation details.

Next, inside of `OFTConfig`, we can override this method with the code here:

```python
    @classmethod
    def check_kwargs(cls, **kwargs):
        ... # code from here
        return super().check_kwargs(**kwargs)  # let's also call super just in case
```

Does that make sense?

BenjaminBossan (changes requested):
Thanks for making further corrections to OFT. I have a few smaller comments and also a question about the parameter finding algorithm, please take a look.

Also, please ensure to run `make style` on your code changes.

Zeju1997:
@BenjaminBossan Can you review again? Best.

BenjaminBossan:
Thanks for the updates @Zeju1997 could you please run `make style`?

BenjaminBossan:
@Zeju1997 gentle reminder about this PR.

BenjaminBossan:
This still did not resolve the issue. Could you please check your ruff version? CI is using `ruff-0.6.7`.

Zeju1997:
@BenjaminBossan Yes, I have ruff==0.6.7 installed, so locally I only have two errors:
make quality
ruff check src tests examples docs scripts docker
examples/boft_controlnet/utils/dataset.py:1:1: I001 [*] Import block is un-sorted or un-formatted
   |
 1 | / import random
 2 | | 
 3 | | import numpy as np
 4 | | import torch
 5 | | import wandb
 6 | | from datasets import load_dataset
 7 | | from diffusers import DDIMScheduler
 8 | | from PIL import Image
 9 | | from torchvision import transforms
10 | | from utils.pipeline_controlnet import LightControlNetPipeline
11 | | 
12 | | 
13 | | def image_grid(imgs, rows, cols):
   | |_^ I001
14 |       assert len(imgs) == rows * cols
   |
   = help: Organize imports

examples/int8_training/peft_adalora_whisper_large_training.py:1:1: I001 [*] Import block is un-sorted or un-formatted
   |
 1 | / import argparse
 2 | | import gc
 3 | | import json
 4 | | import logging
 5 | | import math
 6 | | import os
 7 | | from dataclasses import dataclass
 8 | | from datetime import datetime
 9 | | from pathlib import Path
10 | | from random import randint
11 | | from typing import Any, Dict, List, Union
12 | | 
13 | | # datasets imports
14 | | import datasets
15 | | 
16 | | # metric imports
17 | | import evaluate
18 | | import numpy as np
19 | | import torch
20 | | import transformers
21 | | import wandb
22 | | 
23 | | # accelerate imports
24 | | from accelerate import Accelerator, dispatch_model
25 | | from accelerate.logging import get_logger
26 | | from datasets import Audio, DatasetDict, IterableDatasetDict, interleave_datasets, load_dataset
27 | | 
28 | | # hf imports
29 | | from huggingface_hub import HfApi
30 | | from torch.utils.data import DataLoader
31 | | from tqdm import tqdm
32 | | from transformers import (
33 | |     BitsAndBytesConfig,
34 | |     SchedulerType,
35 | |     WhisperForConditionalGeneration,
36 | |     WhisperProcessor,
37 | |     get_scheduler,
38 | |     set_seed,
39 | | )
40 | | from transformers.models.whisper.english_normalizer import BasicTextNormalizer
41 | | 
42 | | # peft imports
43 | | from peft import AdaLoraConfig, LoraConfig, PeftModel, get_peft_model
44 | | 
45 | | 
46 | | logger = get_logger(__name__, log_level="INFO")
   | |_^ I001
   |
   = help: Organize imports

Found 2 errors.
[*] 2 fixable with the `--fix` option.
make: *** [Makefile:9: quality] Error 1
These two are kept the same as in the github main.

BenjaminBossan:
Hmm, that's strange, it could be that somehow your ruff does not pick up on the settings inside of the `pyproject.toml`. I ran it locally on your branch and this is the diff that I get:

```diff
modified   src/peft/tuners/boft/config.py
@@ -32,7 +32,9 @@ class BOFTConfig(PeftConfig):
         boft_block_num (`int`): Number of BOFT blocks per injected layer.
         boft_n_butterfly_factor (`int`): Number of butterfly factors across different layers.
         target_modules (`Union[List[str],str]`): The names of the modules to apply the adapter to.
-        boft_dropout (`float`): The multiplicative dropout probability, by setting OFT blocks to identity during training, similar to the dropout layer in LoRA.
+        boft_dropout (`float`):
+            The multiplicative dropout probability, by setting OFT blocks to identity during training, similar to the
+            dropout layer in LoRA.
         fan_in_fan_out (`bool`): Set this to True if the layer to replace stores weight like (fan_in, fan_out).
             For example, gpt-2 uses `Conv1D` which stores weights like (fan_in, fan_out) and hence this should be set
             to `True`.
modified   src/peft/tuners/oft/config.py
@@ -27,7 +27,9 @@ class OFTConfig(PeftConfig):
     Args:
         r (`int`): OFT rank, number of OFT blocks per injected layer.
         oft_block_size (`int`): OFT block size across different layers.
-        module_dropout (`float`): The multiplicative dropout probability, by setting OFT blocks to identity during training, similar to the dropout layer in LoRA.
+        module_dropout (`float`):
+            The multiplicative dropout probability, by setting OFT blocks to identity during training, similar to the
+            dropout layer in LoRA.
         target_modules (`Optional[Union[List[str], str]]`):
             The names of the modules to apply the adapter to. If this is specified, only the modules with the specified
             names will be replaced. When passing a string, a regex match will be performed. When passing a list of
modified   src/peft/tuners/oft/layer.py
@@ -44,8 +44,8 @@ class MultiplicativeDropoutLayer(nn.Module):
 
         Parameters:
         x (Tensor): The input tensor of shape (D, H, H), where `D` represents
-                    the number of OFT blocks, and `H` is the size of the square
-                    blocks along the last two dimensions, the block size in OFT.
+                    the number of OFT blocks, and `H` is the size of the square blocks along the last two dimensions,
+                    the block size in OFT.
         """
         if self.training:
             # Ensure the last two dimensions are the same
@@ -152,7 +152,8 @@ class OFTLayer(BaseTunerLayer):
             adapter_name (`str`): Name for the adapter to add.
             r (`int`): Rank for the added adapter.
             oft_block_size (`int`): The block size for added adapter.
-            module_dropout (`float`): The multiplicative dropout probability for disabling adapter blocks during training.
+            module_dropout (`float`):
+                The multiplicative dropout probability for disabling adapter blocks during training.
             coft (`bool`): Whether to use the constrained variant of OFT or not.
             eps (`float`):
                 The control strength of COFT. The freedom of rotation. Only has an effect if `coft` is set to True.
modified   src/peft/tuners/oft/model.py
@@ -331,8 +331,8 @@ class OFTModel(BaseTuner):
         self, progressbar: bool = False, safe_merge: bool = False, adapter_names: Optional[List[str]] = None
     ) -> torch.nn.Module:
         r"""
-        This method merges the OFT layers into the base model. This is needed if someone wants to use the base model
-        as a standalone model.
+        This method merges the OFT layers into the base model. This is needed if someone wants to use the base model as
+        a standalone model.
 
         Args:
             progressbar (`bool`):
```

Zeju1997:
@BenjaminBossan Thanks! I have manually updated the diff you posted. Best.

BenjaminBossan (changes requested):
One small issue: The `oft/layer.py` is missing the `from __future__ import annotations` import while using the more modern type annotation syntax like `tuple[...]`, leading to failures in older Python versions. Could you please add the import there?

BenjaminBossan:
> I have manually updated the diff you posted.

Just in case you didn't know, you can save the diff in a patch file and apply it with `git`.

Zeju1997:
@BenjaminBossan Updated. Best.

BenjaminBossan:
Your latest change angered the linter again because you removed some type imports that were still being used:

```
ruff check src tests examples docs scripts docker
src/peft/tuners/oft/layer.py:119:38: F821 Undefined name `Set`
    |
118 |     @property
119 |     def _available_adapters(self) -> Set[str]:
    |                                      ^^^ F821
[12](https://github.com/huggingface/peft/actions/runs/11034315376/job/30648552707?pr=1996#step:5:13)0 |         return {*self.oft_r}
    |

src/peft/tuners/oft/layer.py:326:71: F821 Undefined name `List`
    |
324 |         self.is_target_conv_1d_layer = is_target_conv_1d_layer
325 | 
326 |     def merge(self, safe_merge: bool = False, adapter_names: Optional[List[str]] = None) -> None:
    |                                                                       ^^^^ F8[21](https://github.com/huggingface/peft/actions/runs/11034315376/job/30648552707?pr=1996#step:5:22)
327 |         """
328 |         Merge the active adapter weights into the base weights
    |
```

You can either add the imports back in or you change `List` to `list` and `Set` to `set` in the file.

Zeju1997:
@BenjaminBossan Updated to avoid using List and Set. Best.

src/peft/tuners/oft/layer.py
```diff
@@ -282,100 +425,320 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
...
+                    )
+                    orig_weights = torch.transpose(orig_weights, 0, 1)
+                    orig_weights = torch.mm(oft_mat, orig_weights)
+                    orig_weights = torch.transpose(orig_weights, 0, 1)
+                    orig_weights = orig_weights * oft_s
+                    orig_weights = orig_weights.view(
+                        self.out_features, self.in_features, base_layer.kernel_size[0], base_layer.kernel_size[0]
+                    )
+
+                    self.base_layer.weight.data = orig_weights.contiguous()
```
BenjaminBossan, line 601:
```suggestion
                    base_layer.weight.data = orig_weights.contiguous()
```

src/peft/tuners/oft/layer.py
```diff
@@ -282,100 +425,320 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
...
```
BenjaminBossan, line 617:
```suggestion
                    base_layer.weight.data = orig_weights.contiguous()
```

src/peft/tuners/oft/layer.py
```diff
@@ -282,100 +425,320 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
...
+                orth_rotate = self._cayley_batch(oft_r)
+                orth_rotate = dropout(orth_rotate)
+                oft_mat = self._block_diagonal(orth_rotate, rank)
+
+                oft_rotation = oft_mat @ oft_rotation
+                oft_scale = oft_s * oft_scale
+
+            x = x.to(self.get_base_layer().weight.data.dtype)
+
+            orig_weights = self.base_layer.weight.data
```
BenjaminBossan, line 716:
```suggestion
            orig_weights = self.get_base_layer().weight.data
```

src/peft/tuners/oft/layer.py
```diff
@@ -282,100 +425,320 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
...
+            orig_weights = orig_weights.view(
+                self.out_features,
+                self.in_features * self.base_layer.kernel_size[0] * self.base_layer.kernel_size[0],
```
BenjaminBossan, line 719:
```suggestion
                self.in_features * self.get_base_layer().kernel_size[0] * self.get_base_layer().kernel_size[0],
```

src/peft/tuners/oft/layer.py
```diff
@@ -282,100 +425,320 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
...
+            orig_weights = torch.transpose(orig_weights, 0, 1)
+            oft_rotation = oft_rotation.to(previous_dtype)
+            orig_weights = orig_weights.to(previous_dtype)
+            rotated_weight = torch.mm(oft_rotation, orig_weights)
+            rotated_weight = torch.transpose(rotated_weight, 0, 1)
+
+            scaled_rotated_weight = rotated_weight * oft_scale
+
+            scaled_rotated_weight = scaled_rotated_weight.view(
+                self.out_features, self.in_features, self.base_layer.kernel_size[0], self.base_layer.kernel_size[0]
```
BenjaminBossan, line 730:
```suggestion
                self.out_features, self.in_features, self.get_base_layer().kernel_size[0], self.get_base_layer().kernel_size[0]
```

src/peft/tuners/oft/layer.py
```diff
@@ -282,100 +425,320 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
...
+            )
+            result = F.conv2d(
+                input=x,
+                weight=scaled_rotated_weight,
+                bias=self.base_layer.bias,
+                padding=self.base_layer.padding[0],
+                stride=self.base_layer.stride[0],
```
BenjaminBossan, line 737:
```suggestion
                bias=self.get_base_layer().bias,
                padding=self.get_base_layer().padding[0],
                stride=self.get_base_layer().stride[0],
```

Zeju1997:
I think the test files have been updated, I will fetch the upstream and fix make test again.

src/peft/tuners/oft/model.py
```diff
@@ -76,33 +90,283 @@ class OFTModel(LycorisTuner):
...
+            if hasattr(new_module, "base_layer"):
+                new_module.base_layer.state = child.state
+            else:
+                new_module.state = child.state
+            new_module.to(child.weight.device)
+
+        # dispatch to correct device
+        for name, module in new_module.named_modules():
+            if self.prefix in name:
+                module.to(child.weight.device)
```
BenjaminBossan, line 185:
```suggestion
        meta = torch.device("meta")
        # dispatch to correct device
        for name, module in new_module.named_modules():
            if self.prefix in name:
                if not any(p.device == meta for p in module.parameters()):
                    module.to(child.weight.device)
```

This will fix the `low_cpu_mem_usage` errors.

src/peft/tuners/oft/layer.py
```diff
@@ -282,100 +425,323 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
...
-        delta_weight = self.get_delta_weight(adapter_name)
+            scaled_rotated_weight = rotated_weight * oft_scale
 
-        base_layer = self.get_base_layer()
-        base_weight = base_layer.weight.data
-        delta_weight = delta_weight[: base_weight.shape[0], : base_weight.shape[0]]
+            scaled_rotated_weight = scaled_rotated_weight.to(previous_dtype)
+            if self.base_layer.bias is not None:
+                self.base_layer.bias = self.base_layer.bias.to(previous_dtype)
+            result = F.linear(input=x, weight=scaled_rotated_weight, bias=self.base_layer.bias)
```
BenjaminBossan, line 467: Let's not mutate the bias of the base layer.

```suggestion
            bias = self.get_base_layer().bias.to(previous_dtype) if self.get_base_layer().bias is not None else None
            result = F.linear(input=x, weight=scaled_rotated_weight, bias=bias)
```

BenjaminBossan (changes requested):
Thanks for the adjustments. I checked the tests that are still failing and suggested some fixes that should help with those.

However, I still get some failing tests with `PeftMixedModel`. I think the problem comes down to the following:

With the new OFT changes, the forward code no longer is: `result = self.base_layer(x)` and then do something with the result. Instead, we retrieve the weight of the base layer, do some operations on it, then call `F.linear(x, weight)`.

This is generally fine but it breaks if we have nested adapters. So e.g. an OFT layer that wraps a LoRA layer that wraps an `nn.Linear` layer. In that case, since we use the `weight` of the `nn.Linear` layer (the `self.get_base_layer().weight`), we basically bypass the LoRA layer's contribution.

I'm not sure if there is a way to rewrite the forward pass to fix that, i.e. to use `self.base_layer(x)` and do the OFT modifications on top. It would be great if there was a way. But if it doesn't work, what we need to do is remove OFT from the methods supported by `PeftMixedModel`. I think that's also fine, as I don't believe many people use it and this change is backwards breaking anyway. LMK how you want to proceed.

BenjaminBossan:
Alright, then let's remove OFT support for mixed models.

First change is fairly simple, let's remove it from the mapping here:

https://github.com/huggingface/peft/blob/ccc350151f95a9ff95da046bae5671da75eab52f/src/peft/mixed_model.py#L49

Next is a little bit more work, we need to adjust the tests here:

https://github.com/huggingface/peft/blob/main/tests/test_mixed.py. There, we mostly need to remove it from the parameterized tests, which should be straightforward. Then we have a these instances:

https://github.com/huggingface/peft/blob/ccc350151f95a9ff95da046bae5671da75eab52f/tests/test_mixed.py#L563-L566

https://github.com/huggingface/peft/blob/ccc350151f95a9ff95da046bae5671da75eab52f/tests/test_mixed.py#L766-L768

Here we should not instantiate `config4`, which is the `OFTConfig`, and also remove `config4` from the list of adapters when calling `set_adapter`.

Zeju1997:
@BenjaminBossan Thanks, I have updated the mixed peft test.

tests/test_mixed.py
```diff
@@ -759,14 +748,6 @@ def test_decoder_model(self):
         assert torch.isfinite(output3).all()
         assert not torch.allclose(output2, output3)
 
-        torch.manual_seed(4)
-        config4 = OFTConfig(r=2, task_type="CAUSAL_LM", target_modules=["q_proj", "v_proj"], init_weights=False)
-        peft_model.add_adapter("adapter4", config4)
-        peft_model.set_adapter(["adapter0", "adapter1", "adapter2", "adapter3", "adapter4"])
```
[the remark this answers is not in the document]
Zeju1997: Sure, updated.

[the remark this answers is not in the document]
Zeju1997: @BenjaminBossan Could you check again, please? Best.

tests/test_mixed.py
```diff
@@ -747,6 +747,11 @@ def test_decoder_model(self):
         output3 = peft_model.generate(**input_dict)
         assert torch.isfinite(output3).all()
         assert not torch.allclose(output2, output3)
+        
+        torch.manual_seed(4)
+        config4 = OFTConfig(r=2, task_type="CAUSAL_LM", target_modules=["q_proj", "v_proj"], init_weights=False)
+        peft_model.add_adapter("adapter4", config4)
```
BenjaminBossan, line 753: These lines should be removed. We only want to keep the `peft_model.set_adapter(["adapter0", "adapter1", "adapter2", "adapter3"])` part. Also don't forget to remove the `OFTConfig` import, as it should not be needed anymore.

Zeju1997:
@BenjaminBossan Sure, updated.

src/peft/tuners/oft/config.py
```diff
@@ -111,9 +137,53 @@ class OFTConfig(LycorisConfig):
...
+        r"""
+        Check if the kwargs are valid for the configuration.
+
+        Args:
+            kwargs (additional keyword arguments, *optional*):
+                Additional keyword arguments passed along to the child class initialization.
+        """
+        if "oft_block_size" not in kwargs:
+            raise ValueError(
+                "OFT has been updated since PEFT 0.13.0. Your trained adapter weights are incompatible with the latest version of OFT. Please retrain your adapter weights with newer PEFT versions."
```
BenjaminBossan, line 186: Let's break this long line to be 120 chars at most. Also, there is a space missing at the end of the sentence.

BenjaminBossan (changes requested):
I took a look at the failing tests. We need to change a few things to make them pass:


1. The config tests actually don't make a lot of sense as they are. For now, simply remove OFT [here](https://github.com/huggingface/peft/blob/ccc350151f95a9ff95da046bae5671da75eab52f/tests/test_config.py#L37).
2. We need to ensure for a few tests that OFT weights are not initialized as identity transform, same as we already do with BOFT. So each time in `test_decoder_models.py`, `test_encoder_decoder_models.py`, and `test_feature_extraction.py`, when there is the line `"boft_kwargs": {"init_weights": [False]}`, also add `"oft_kwargs": {"init_weights": [False]}`.
3. OFT no longer supports `Conv1D` layers. Either this support needs to be added back in or GPT2 tests need to be skipped, same as for BOFT (see [here](https://github.com/huggingface/peft/blob/ccc350151f95a9ff95da046bae5671da75eab52f/tests/test_decoder_models.py#L66)).
4. [This line](https://github.com/huggingface/peft/blob/ccc350151f95a9ff95da046bae5671da75eab52f/tests/testing_common.py#L1287) needs to add OFT.
5. Also add [OFT](https://github.com/huggingface/peft/blob/ccc350151f95a9ff95da046bae5671da75eab52f/tests/testing_common.py#L649-L650) here, same as BOFT.

I think that covers all, but maybe I missed something. The complete diff I got locally is:

```diff
modified   tests/test_config.py
@@ -34,7 +34,6 @@ from peft import (
     LoKrConfig,
     LoraConfig,
     MultitaskPromptTuningConfig,
-    OFTConfig,
     PeftConfig,
     PeftType,
     PolyConfig,
@@ -61,7 +60,6 @@ ALL_CONFIG_CLASSES = (
     LoKrConfig,
     LoraConfig,
     MultitaskPromptTuningConfig,
-    OFTConfig,
     PolyConfig,
     PrefixTuningConfig,
     PromptEncoderConfig,
modified   tests/test_decoder_models.py
@@ -22,6 +22,7 @@ from transformers import AutoModelForCausalLM, AutoTokenizer
 from peft import (
     AdaLoraConfig,
     BOFTConfig,
+    OFTConfig,
     HRAConfig,
     LoraConfig,
     PrefixTuningConfig,
@@ -59,7 +60,7 @@ def skip_boft_or_hra_and_gpt2(test_list):
     return [
         test
         for test in test_list
-        if not (("GPT2LMHeadModel" in test[1]) and ((test[2] == BOFTConfig) or (test[2] == HRAConfig)))
+        if not (("GPT2LMHeadModel" in test[1]) and ((test[2] == BOFTConfig) or (test[2] == OFTConfig) or (test[2] == HRAConfig)))
     ]
 
 
@@ -205,6 +206,7 @@ class PeftDecoderModelTester(unittest.TestCase, PeftCommonTester):
                 "adalora_kwargs": {"init_lora_weights": [False]},
                 "ia3_kwargs": {"init_ia3_weights": [False]},
                 "boft_kwargs": {"init_weights": [False]},
+                "oft_kwargs": {"init_weights": [False]},
                 "vera_kwargs": {"init_weights": [False]},
                 "fourierft_kwargs": {"init_weights": [False]},
                 "hra_kwargs": {"init_weights": [False]},
@@ -222,6 +224,7 @@ class PeftDecoderModelTester(unittest.TestCase, PeftCommonTester):
                 "lora_kwargs": {"init_lora_weights": [False]},
                 "ia3_kwargs": {"init_ia3_weights": [False]},
                 "boft_kwargs": {"init_weights": [False]},
+                "oft_kwargs": {"init_weights": [False]},
                 "vera_kwargs": {"init_weights": [False]},
                 "fourierft_kwargs": {"init_weights": [False]},
                 "hra_kwargs": {"init_weights": [False]},
@@ -336,6 +339,7 @@ class PeftDecoderModelTester(unittest.TestCase, PeftCommonTester):
                 "adalora_kwargs": {"init_lora_weights": [False]},
                 "ia3_kwargs": {"init_ia3_weights": [False]},
                 "boft_kwargs": {"init_weights": [False]},
+                "oft_kwargs": {"init_weights": [False]},
                 "vera_kwargs": {"init_weights": [False]},
                 "fourierft_kwargs": {"init_weights": [False]},
                 "hra_kwargs": {"init_weights": [False]},
@@ -373,6 +377,7 @@ class PeftDecoderModelTester(unittest.TestCase, PeftCommonTester):
                 "ia3_kwargs": {"init_ia3_weights": [False]},
                 "adalora_kwargs": {"init_lora_weights": [False]},
                 "boft_kwargs": {"init_weights": [False]},
+                "oft_kwargs": {"init_weights": [False]},
                 "vera_kwargs": {"init_weights": [False]},
                 "fourierft_kwargs": {"init_weights": [False]},
                 "hra_kwargs": {"init_weights": [False]},
modified   tests/test_encoder_decoder_models.py
@@ -173,6 +173,7 @@ class PeftEncoderDecoderModelTester(unittest.TestCase, PeftCommonTester):
                 "adalora_kwargs": {"init_lora_weights": [False]},
                 "ia3_kwargs": {"init_ia3_weights": [False]},
                 "boft_kwargs": {"init_weights": [False]},
+                "oft_kwargs": {"init_weights": [False]},
                 "vera_kwargs": {"init_weights": [False]},
                 "hra_kwargs": {"init_weights": [False]},
                 "task_type": "SEQ_2_SEQ_LM",
@@ -207,6 +208,7 @@ class PeftEncoderDecoderModelTester(unittest.TestCase, PeftCommonTester):
                 "adalora_kwargs": {"init_lora_weights": [False]},
                 "ia3_kwargs": {"init_ia3_weights": [False]},
                 "boft_kwargs": {"init_weights": [False]},
+                "oft_kwargs": {"init_weights": [False]},
                 "vera_kwargs": {"init_weights": [False]},
                 "hra_kwargs": {"init_weights": [False]},
                 "task_type": "SEQ_2_SEQ_LM",
modified   tests/test_feature_extraction_models.py
@@ -111,6 +111,7 @@ class PeftFeatureExtractionModelTester(unittest.TestCase, PeftCommonTester):
                 "adalora_kwargs": {"init_lora_weights": [False]},
                 "ia3_kwargs": {"init_ia3_weights": [False]},
                 "boft_kwargs": {"init_weights": [False]},
+                "oft_kwargs": {"init_weights": [False]},
                 "vera_kwargs": {"init_weights": [False]},
                 "hra_kwargs": {"init_weights": [False]},
                 "task_type": "FEATURE_EXTRACTION",
@@ -164,6 +165,7 @@ class PeftFeatureExtractionModelTester(unittest.TestCase, PeftCommonTester):
                 "adalora_kwargs": {"init_lora_weights": [False]},
                 "ia3_kwargs": {"init_ia3_weights": [False]},
                 "boft_kwargs": {"init_weights": [False]},
+                "oft_kwargs": {"init_weights": [False]},
                 "vera_kwargs": {"init_weights": [False]},
                 "hra_kwargs": {"init_weights": [False]},
                 "task_type": "FEATURE_EXTRACTION",
@@ -180,6 +182,7 @@ class PeftFeatureExtractionModelTester(unittest.TestCase, PeftCommonTester):
                 "lora_kwargs": {"init_lora_weights": [False]},
                 "ia3_kwargs": {"init_ia3_weights": [False]},
                 "boft_kwargs": {"init_weights": [False]},
+                "oft_kwargs": {"init_weights": [False]},
                 "hra_kwargs": {"init_weights": [False]},
                 "task_type": "FEATURE_EXTRACTION",
             },
modified   tests/testing_common.py
@@ -652,7 +652,7 @@ class PeftCommonTester:
         if issubclass(config_cls, PromptLearningConfig):
             return pytest.skip(f"Test not applicable for {config_cls}")
 
-        if issubclass(config_cls, BOFTConfig):
+        if issubclass(config_cls, (OFTConfig, BOFTConfig)):
             return pytest.skip(f"Test not applicable for {config_cls}")
 
         if ("gpt2" in model_id.lower()) and (config_cls != LoraConfig):
@@ -1290,7 +1290,7 @@ class PeftCommonTester:
         model = get_peft_model(model, config)
         model = model.to(self.torch_device)
 
-        if config.peft_type not in ("LORA", "ADALORA", "IA3", "BOFT", "VERA", "FOURIERFT", "HRA", "VBLORA"):
+        if config.peft_type not in ("LORA", "ADALORA", "IA3", "BOFT", "OFT", "VERA", "FOURIERFT", "HRA", "VBLORA"):
             with pytest.raises(AttributeError):
                 model = model.unload()
         else:
```

There are a few more failing that are unrelated to this PR. We need #2106 merged first before these tests can pass.

Zeju1997:
@BenjaminBossan Thank you so much for the detailed suggestions. I have updated the test files. Best.

BenjaminBossan:
Thanks for the quick action. You missed a few changes still in `test_decoder_models.py`, `test_encoder_decoder_models.py`, and [this line](https://github.com/huggingface/peft/pull/1996/commits/65ce38b833e08893043814725e870e1d2b3ff2ce#diff-d904228fa17421a9b7d8436ef706bd44bcda957e3978fa34624efbdcc8eb191eR187) is missing a space at the end. After this, we have yet to wait for #2106 to be merged first.

Zeju1997:
> Thanks for the quick action. You missed a few changes still in `test_decoder_models.py`, `test_encoder_decoder_models.py`, and [this line](https://github.com/huggingface/peft/pull/1996/commits/65ce38b833e08893043814725e870e1d2b3ff2ce#diff-d904228fa17421a9b7d8436ef706bd44bcda957e3978fa34624efbdcc8eb191eR187) is missing a space at the end. After this, we have yet to wait for #2106 to be merged first.

@BenjaminBossan Sorry, I don't understand, what do you mean with a white space at the end?

Zeju1997:
@BenjaminBossan So locally, I ran the `pytest tests/ -k test_encoder_decoder_models` and `pytest tests/ -k test_decoder_models` and I did not get any error.

I do get some error with `pytest tests/ -k test_feature_extraction_models`:
FAILED tests/test_feature_extraction_models.py::PeftFeatureExtractionModelTester::test_training_gradient_checkpointing_27_test_hf_internal_testing_tiny_random_DebertaModel_oft - RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.
FAILED tests/test_feature_extraction_models.py::PeftFeatureExtractionModelTester::test_training_gradient_checkpointing_31_test_hf_internal_testing_tiny_random_DebertaV2Model_oft - RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.

BenjaminBossan:
> I do get some error with `pytest tests/ -k test_feature_extraction_models`

Yeah, I can reproduce that, strange. Maybe it's related to the special layer types that Deberta uses.. I think we can skip this specific Deberta test, similar to how we skip AdaLoRA + Roberta:

https://github.com/huggingface/peft/blob/c29810bad23f24c637d8e0bce42866d6748d94e2/tests/testing_common.py#L1105-L1107

Please also merge with/rebase on the latest main from PEFT.

Zeju1997:
@BenjaminBossan I fixed the error in test_feature_extraction_models and fetched the latest main. Best.

BenjaminBossan:
Hmm, it seems like `test_deeply_nested` is flaky now. Let me investigate a bit next week.

Zeju1997:
@BenjaminBossan Thank you so much! I tested the test_deeply_nested, so locally there was no error.

BenjaminBossan:
@Zeju1997 I can confirm that `test_deeply_nested` works locally, both on GPU and CPU. Therefore, I'm not sure what's causing the error on CI. Checking the error message, the tensors look identical, so it's most likely a precision problem. Could you please increase the [tolerance in this test](https://github.com/huggingface/peft/blob/aa3bd8fbf6820bb000e148a289b6b0c48c339774/tests/test_mixed.py#L543-L544) to 1e-4?

Zeju1997:
@BenjaminBossan Thanks! I have updated.

BenjaminBossan:
Thanks, but unfortunately, this has not solved the issue. I tested on another Linux PC with Intel CPU and still can't reproduce the issue. At this point, I don't know what else can be done, it's most likely some problem with the hardware of the GitHub CI runners. I would propose to skip this test on Linux for now:

```python
...
import platform
...

# in test
        if platform.system() == "Linux":
            self.skipTest("This test fails but only on GitHub CI with Linux systems.")
```

If you add this, you can also do revert the tolerance changes, since they didn't help.

Zeju1997:
@BenjaminBossan Updated.

Zeju1997:
@BenjaminBossan I think all the tests have been passed. Thank you again for the great help during the process!

BenjaminBossan (approved):
Thanks a lot for this PR and for your patience. Finally CI is green, I'll merge right away.

src/peft/tuners/oft/config.py
```diff
@@ -111,9 +137,54 @@ class OFTConfig(LycorisConfig):
...
+
+        Args:
+            kwargs (additional keyword arguments, *optional*):
+                Additional keyword arguments passed along to the child class initialization.
+        """
+        if "oft_block_size" not in kwargs:
+            raise ValueError(
+                "OFT has been updated since PEFT 0.13.0. Your trained adapter weights are incompatible "
+                "with the latest version of OFT. Please retrain your adapter weights with newer PEFT versions."
+                "Alternatively, downgrade PEFT to version 0.12.0 to use the old adapter weights."
```
BenjaminBossan, line 188: Ah, damn, a minor issue I noticed just now: Since PEFT 0.13.0 was released already, this message is no longer correct.
```suggestion
                "OFT has been updated since PEFT 0.14.0. Your trained adapter weights are incompatible "
                "with the latest version of OFT. Please retrain your adapter weights with newer PEFT versions."
                "Alternatively, downgrade PEFT to version 0.13.0 to use the old adapter weights."
```

BenjaminBossan: Ping @Zeju1997 just this small change and we're good.

Zeju1997: @BenjaminBossan Sure, updated. Best.

src/peft/tuners/oft/config.py
```diff
@@ -111,9 +137,54 @@ class OFTConfig(LycorisConfig):
...
+        Check if the kwargs are valid for the configuration.
+
+        Args:
+            kwargs (additional keyword arguments, *optional*):
+                Additional keyword arguments passed along to the child class initialization.
+        """
+        if "oft_block_size" not in kwargs:
+            raise ValueError(
+                "OFT has been updated since PEFT 0.14.0. Your trained adapter weights are incompatible"
+                "with the latest version of OFT. Please retrain your adapter weights with newer PEFT versions."
```
BenjaminBossan, line 187: Please add back the spaces:

```suggestion
                "OFT has been updated since PEFT 0.14.0. Your trained adapter weights are incompatible "
                "with the latest version of OFT. Please retrain your adapter weights with newer PEFT versions. "
```

Zeju1997:
@BenjaminBossan Updated. Best,

BenjaminBossan:
We've got it, finally. Thanks for your patience.

BenjaminBossan:
@Zeju1997 Unfortunately, the regression tests are failing now for BOFT with the `Conv2d` layer:

https://github.com/huggingface/peft/actions/runs/11136746324/job/30949028303#step:8:182

The issue is most likely that the `boft_s` parameter has been transposed:

https://github.com/huggingface/peft/pull/1996/files#diff-76cd9e2dc2dd53a73bdd1a2ca6f475b44978690fea8e417bbd3952ae1c7929adR775

This means that users wouldn't be able to load BOFT checkpoints including `Conv2d`. Can we transpose this vector back to its initial shape?

Zeju1997:
@BenjaminBossan OK, I will look into this.

Zeju1997:
@BenjaminBossan Is this related to old checkpoints from previous versions or also the current version?

BenjaminBossan:
> Is this related to old checkpoints from previous versions or also the current version?

It's an [old BOFT `Conv2d` checkpoint](https://huggingface.co/peft-internal-testing/regression-tests/tree/main/boft_conv2d/0.11.1) that can't be loaded with the current code.

PR #2003, opened 2024-08-13
Update lora.md (typos)

nir-sh-automat-it:
Fixed some typos (examle -> example, and some more)


BenjaminBossan (approved):
Thanks for catching and fixing these typos.

PR #2004, opened 2024-08-14
ENH: Warn when a user provided model name in the config renamed

### LINKED ISSUE #2001: Issue with PEFT model save_pretrained
### System Info

Model: Custom Phi-3 implementation, but also causes issue with normal Phi-3
OS: Linux(ubuntu)
Transformers version: 4.43.3
PEFT version: 0.11.1

### Who can help?

_No response_

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

```python
from peft import LoraConfig, get_peft_model, prepare_model_for_kbit_training, AutoPeftModelForCausalLM, PeftModel, PeftConfig, PeftModelForCausalLM

model_path = "/content/gdrive/My Drive/Colab Notebooks/DOE/src/models/weights/Phi-3-mini-128k-instruct"

model = AutoModelForCausalLM.from_pretrained(model_path)

peft_config = LoraConfig(**{
    "r": 16,
    "lora_alpha": 32,
    "lora_dropout": 0.05,
    "bias": "none",
    "task_type": "CAUSAL_LM",
    "target_modules": "all-linear",
    "modules_to_save": None,
    "base_model_name_or_path": "/content/gdrive/My Drive/Colab Notebooks/DOE/src/models/weights/Phi-3ex"
})

kbit_model = prepare_model_for_kbit_training(model)
peft_model = get_peft_model(kbit_model, peft_config)
peft_model.save_pretrained("/content/gdrive/MyDrive/Colab Notebooks/DOE/test_peft_model/phi-peft-v3")

``` 

### Expected behavior

The resultant adapter_config.json is following JSON:
```
{
  "alpha_pattern": {},
  "auto_mapping": null,
  "base_model_name_or_path": "/content/gdrive/My Drive/Colab Notebooks/DOE/src/models/weights/Phi-3-mini-128k-instruct",
  "bias": "none",
  "fan_in_fan_out": false,
  "inference_mode": true,
  "init_lora_weights": true,
  "layer_replication": null,
  "layers_pattern": null,
  "layers_to_transform": null,
  "loftq_config": {},
  "lora_alpha": 32,
  "lora_dropout": 0.05,
  "megatron_config": null,
  "megatron_core": "megatron.core",
  "modules_to_save": null,
  "peft_type": "LORA",
  "r": 16,
  "rank_pattern": {},
  "revision": null,
  "target_modules": [
    "gate_up_proj",
    "o_proj",
    "qkv_proj",
    "dAnd these can be still setup in private environmentown_proj"
  ],
  "task_type": "CAUSAL_LM",
  "use_dora": false,
  "use_rslora": false
}
``` 

But I would ideally want  _**"/content/gdrive/My Drive/Colab Notebooks/DOE/src/models/weights/Phi-3ex"**_ to be the value for _**base_model_name_or_path**_ in resultant adapter_config.

In my use case I am using a custom version of Phi-3 model(I have subclassed Phi-3 and overridden some methods and changed the architecture a bit), so when I create the peft_model by passing my modified model, the adapter_config sets the value of **_base_model_name_or_path_** to  _**Phi-3-mini-128k-instruct**_. Obviously when loading this peft model with `from_pretrained`, it can't find the base model and results in an error. Thus I would appreciate if I can know where I might be going wrong or if this an actual bug.

BenjaminBossan:
Resolves #2001

In PEFT, users can provide a custom `base_model_name_or_path` argument to the PEFT config. However, this value is overridden by the model's `name_or_path` attribute. This can be surprising for users. Therefore, there is now a warning about this.

To see why that can be relevant, check the original issue.

It eludes me why this argument can be passed by users only to be overridden by PEFT. This was implemented by Younes before I joined the project.

### DIFF
--- a/tests/test_initialization.py
+++ b/tests/test_initialization.py
@@ -1248,3 +1248,47 @@ class TestLoadAdapterOfflineMode:
         with self.hub_offline_ctx():
             # does not raise
             PeftModel.from_pretrained(base_model, peft_model_id)
+
+
+class TestCustomModelConfigWarning:
+    # Check potential warnings when the user provided base_model_name_or_path is overridden by PEFT. See #2001 for
+    # context. We use LoRA for this test but the same applies to other methods
+    @pytest.fixture
+    def custom_module(self):
+        class MyModule(nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.lin = nn.Linear(10, 10)
+
+        return MyModule()
+
+    def test_no_warning_by_default_transformers_model(self, recwarn):
+        # first a sanity test that there is no warning by default when using a model from transformers
+        model = AutoModelForCausalLM.from_pretrained("facebook/opt-125m")
+        get_peft_model(model, LoraConfig())
+        for warning in recwarn.list:
+            assert "renamed" not in str(warning.message)
+
+    def test_no_warning_by_default_custom_model(self, custom_module, recwarn):
+        # same as above but with a custom model
+        get_peft_model(custom_module, LoraConfig(target_modules=["lin"]))
+        for warning in recwarn.list:
+            assert "renamed" not in str(warning.message)
+
+    def test_warning_name_transformers_model(self, recwarn):
+        # The base_model_name_or_path provided by the user is overridden.
+        model = AutoModelForCausalLM.from_pretrained("facebook/opt-125m")
+        get_peft_model(model, LoraConfig(base_model_name_or_path="custom_name"))
+        msg = "was renamed from 'custom_name' to 'facebook/opt-125m'"
+        assert any(msg in str(warning.message) for warning in recwarn.list)
+
+    def test_warning_name_custom_model(self, custom_module, recwarn):
+        get_peft_model(custom_module, LoraConfig(target_modules=["lin"], base_model_name_or_path="custom_name"))
+        msg = "was renamed from 'custom_name' to 'None'"
+        assert any(msg in str(warning.message) for warning in recwarn.list)
+
+    def test_warning_name_custom_model_with_custom_name(self, custom_module, recwarn):
+        custom_module.name_or_path = "foobar"
+        get_peft_model(custom_module, LoraConfig(target_modules=["lin"], base_model_name_or_path="custom_name"))
+        msg = "was renamed from 'custom_name' to 'foobar'"
+        assert any(msg in str(warning.message) for warning in recwarn.list)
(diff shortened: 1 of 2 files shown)


tests/test_initialization.py
```diff
@@ -1248,3 +1248,47 @@ def test_load_from_hub_then_offline_model(self):
...
+
+    def test_no_warning_by_default_custom_model(self, custom_module, recwarn):
+        # same as above but with a custom model
+        get_peft_model(custom_module, LoraConfig(target_modules=["lin"]))
+        for warning in recwarn.list:
+            assert "renamed" not in str(warning.message)
+
+    def test_warning_name_transformers_model(self, recwarn):
+        # The base_model_name_or_path provided by the user is overridden.
+        model = AutoModelForCausalLM.from_pretrained("facebook/opt-125m")
```
sayakpaul, line 1280: Let's test with a smaller model? Plenty available here: https://huggingface.co/hf-internal-testing

tests/test_initialization.py
```diff
@@ -1248,3 +1248,47 @@ def test_load_from_hub_then_offline_model(self):
...
+        get_peft_model(model, LoraConfig(base_model_name_or_path="custom_name"))
+        msg = "was renamed from 'custom_name' to 'facebook/opt-125m'"
+        assert any(msg in str(warning.message) for warning in recwarn.list)
+
+    def test_warning_name_custom_model(self, custom_module, recwarn):
+        get_peft_model(custom_module, LoraConfig(target_modules=["lin"], base_model_name_or_path="custom_name"))
```
sayakpaul, line 1286: "custom_name" could be assigned to a variable and reused for easier understanding.

sayakpaul (commented):
Thanks!

I slightly prefer catching warnings with a with context more, but that is just preferential.

BenjaminBossan:
Thanks @sayakpaul. The [last commit](https://github.com/huggingface/peft/pull/2004/commits/c1aa816d9b6076935b57fc2ff96c8a775ae758c0) should address your comments.

> I slightly prefer catching warnings with a with context more

You mean something like `pytest.warns`? The reason I didn't go with that is that it should not be used to check for _absence_ of warnings, so the tests would have different methods depending on whether there is a warning or not. I wanted to keep the code very similar between the tests, which is why I chose `recwarn`. If I only test for presence of warnings, I also prefer using a context manager.

sayakpaul (approved):
LGTM!

PR #2005, opened 2024-08-14
TST Add LNTuningConfig and LoKrConfig to tests

BenjaminBossan:
These two configs were missing in `test_config.py`. Also, reordered the list of all config classes to be sorted, which makes it easier to spot missing configs.


SunMarc (approved):
Nice ! Thanks !

PR #2002, opened 2024-08-12
Update docs for X-LoRA and some bugfixes

EricLBuehler:
Refs #1966. Adds some stuff to the docs, but the main focus is on the loading bugfix.

I ran:
```
ruff format
make quality
make style
pytest tests/test_xlora.py -v
```
With:
```
> ruff --version
ruff 0.5.7
> pytest --version
pytest 8.3.2
```

Looks like some tests are still failing, I'll take a look!

PR #2010, opened 2024-08-16
Fix usage of deprecated parameters/functions in X-LoRA

EricLBuehler:
As pointed out by @benjamin-marie in #2002, we were calling `prepare_model_for_int8_training` and passing `use_flash_attention_2`. These are deprecated and this PR fixes any usages.

### DIFF
--- a/src/peft/tuners/xlora/model.py
+++ b/src/peft/tuners/xlora/model.py
@@ -188,7 +188,7 @@ class XLoraModel(BaseTuner):
         >>> model = AutoModelForCausalLM.from_pretrained(
         ...     "mistralai/Mistral-7B-Instruct-v0.1",
         ...     trust_remote_code=True,
-        ...     _attn_implementation="flash_attention_2",
+        ...     attn_implementation="flash_attention_2",
         ...     device_map="cuda:0",
         ...     torch_dtype=torch.bfloat16,
         ...     quantization_config=int8_config,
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
@EricLBuehler Is this ready for review? Please always ping me when that's the case.

EricLBuehler:
Sounds good @BenjaminBossan, yes this is ready for review!

BenjaminBossan (commented):
Thanks for fixing these small issues. Just a question: Why use `_attn_implementation` and not `attn_implementation`? I did a test and the latter seems to work.

EricLBuehler:
Thanks for the recommendation, there was no reason, actually. I guess I've been using the wrong parameter then.

src/peft/tuners/xlora/model.py
```diff
@@ -170,8 +170,8 @@ class XLoraModel(BaseTuner):
 
     Example:
         ```py
-        >>> from transformers import AutoModelForCausalLM, AutoConfig
-        >>> from peft import LoraConfig, PeftModel, get_peft_model, prepare_model_for_int8_training
```
BenjaminBossan, line 174: Sorry I didn't notice earlier: `prepare_model_for_int8_training` is deprecated, but `prepare_model_for_kbit_training` replaces it and can be used here.

EricLBuehler: Ah ok thanks! I updated it to call `prepare_model_for_kbit_training`.

BenjaminBossan (approved):
Thanks for the fixes.

EricLBuehler:
Thanks!

PR #1905, opened 2024-07-03
Add MoSLoRA

wutaiqiang:
Link: https://arxiv.org/pdf/2406.11909v1

Hello, I am one of the authors of this paper.

In this PR, I try to add MoSLoRA, a subspace-inspired Low-Rank Adaptation (LoRA) method. MoSLoRA employs a learnable mixer to fuse more subspaces and more flexibly, while vanilla LoRA can be considered a special case with a fixed identity matrix being the mixer. Please read the paper for more details. 

Related Issue: https://github.com/huggingface/peft/issues/1878 

Also, the Implement is similar to DoRA, you can just set "use_moslora=True" to enable MoSLoRA.

For the test, I follow the test items of DoRA and added test functions in the following files:

> tests/regression/test_regression.py
tests/regression/test_regression.py
tests/test_common_gpu.py 
tests/test_custom_models.py
tests/test_gpu_examples.py
tests/test_initialization.py
tests/test_torch_compile.py 

And I try the "pytest xxx.py -k moslora" and all the tests pass.

Thanks for reviewing! @BenjaminBossan

PR #2013, opened 2024-08-17
Add MoSLoRA

wutaiqiang:
The previous PR:

https://github.com/huggingface/peft/pull/1905 

Updates:
- sync the latest changes in PEFT
- add testing examples


BenjaminBossan:
@wutaiqiang did you see my [reply](https://github.com/huggingface/peft/pull/1905#issuecomment-2298499321) on the previous discussion?

wutaiqiang:
@BenjaminBossan Thanks for your reminder :)

Indeed, the experiments on finetuning Roberta tend to overfit easily, and the performances for these three methods (LoRA, MoSLoRA, MoSLORA_init) are almost similar.

Following your suggestion, I am running experiments reported in my paper (https://arxiv.org/pdf/2406.11909), i.e., fintuning LLaMA 3 on the QA tasks. As the LLaMA 3 contains 8B parameters, please wait for a day or two. I would update the training code, training logs, weights, and benchmark scores in this pr after finishing the test.

Thanks again for your time and great help.

BenjaminBossan:
Thanks a lot @wutaiqiang for investigating this further.

wutaiqiang:
@BenjaminBossan I would like to share further experiments for comparing LoRA, MoSLoRA, and MoSLoRA_Init (cf. https://github.com/huggingface/peft/pull/1905).

 For testing, I ran the experiments by finetuning LLMs on the QA tasks. The details are as follows:

- Base model: [Phi-3.5-mini-instruct(3.82B)](https://huggingface.co/microsoft/Phi-3.5-mini-instruct) 
- Training set: [commonsense_170k.json](https://github.com/AGI-Edgerunners/LLM-Adapters/blob/main/ft-training_set/commonsense_170k.json) containing 170K QA pairs
- Test set: ARC-Easy, openbookqa,social_i_qa,ARC-Challenge,winogrande,piqa,boolq

To reproduce the results, you can use the file provided in [my repo](https://github.com/wutaiqiang/peft/tree/main/examples/moslora_llm_qa_finetune). The hyperparameters are:

- rank=8
- alpha=16
- batch_size=16 
- num_epochs=1
- learning_rate=1e-4
- seed=1/2/3/4/5 (Repeat for 5 times)

Then, you can start the training for LoRA by:
```
bash train_test_31.sbatch
```
MoSLoRA by:
```
bash train_test_32.sbatch
```
MoSLoRA_init by:
```
bash train_test_33.sbatch
```

After that, the trained models would be saved in "output_models" and the inference results would be saved in "output_results". Also, you can check the logs for the training process and evaluate the results. The experiments are conducted on an 80G A100 GPU (48G used) and 28h for 5 epochs.

> My training logs, weights, and inference results can be found in [Google Drive](https://drive.google.com/drive/folders/1ChCilPhvZYw-vKCbr9qVf4-Z84O9_RnF?usp=sharing).

The results of seed 1/2/3/4/5 are:

| Benchmark    |   | ARC-E | OBQA | SIQA  | ARC-C | WinoG | PIQA  | BoolQ |
|--------------|---|-------|------|-------|-------|-------|-------|-------|
| Size         |   | 2376  | 500  | 1954  | 1172  | 1267  | 1838  | 3270  |
| LoRA         | 1 | 94.57 | 83.4 | 74.71 |  83.6 | 77.19 | 85.69 | 68.89 |
|              | 2 | 90.82 | 74.8 | 64.79 | 79.52 | 71.58 | 84.92 | 66.97 |
|              | 3 | 94.61 | 84.8 | 73.08 | 83.44 | 46.56 | 85.31 | 68.99 |
|              | 4 | 94.48 |   82 | 51.45 | 84.04 | 76.79 | 85.31 |  68.1 |
|              | 5 | 94.69 | 83.6 | 75.33 | 83.87 | 76.64 | 85.25 | 68.19 |
| MoSLoRA      | 1 | 89.64 | 77.4 | 66.53 | 77.13 | 74.98 | 85.36 | 69.11 |
|              | 2 | 94.36 | 84.4 | 75.58 | 84.21 | 75.69 | 85.31 | 66.97 |
|              | 3 | 94.19 | 82.4 | 72.93 | 83.78 | 75.29 | 84.11 | 67.83 |
|              | 4 | 94.44 | 83.4 | 75.12 | 83.79 | 76.24 | 84.98 | 68.77 |
|              | 5 | 94.44 | 83.8 | 75.13 | 84.12 | 76.71 | 85.85 | 68.25 |
| MoSLoRA Init | 1 | 94.36 | 83.8 | 73.38 |  84.2 | 76.71 | 85.26 | 68.44 |
|              | 2 | 90.53 | 76.2 | 71.95 | 77.13 |  74.9 |  84.6 | 68.37 |
|              | 3 | 79.84 | 73.4 | 58.49 | 69.11 | 77.19 | 84.05 | 65.77 |
|              | 4 | 94.69 | 83.2 | 72.26 | 83.53 | 77.34 | 84.54 | 68.96 |
|              | 5 | 94.61 |   84 | 74.97 | 83.53 | 70.24 | 85.03 | 68.89 |

The average scores are:

|   Method  |  Average | Weighted Average |
|-----|------| -------|
| LoRA | 78.51 | 77.71 |
|MoSLoRA | **80.06** | **79.02** |
|MoSLoRA Init | 78.38 | 77.55 |

where:
Average = Sum(Average of each benchmark)/7
Weighted Average = Sum(Benchmark size * Average of each benchmark)/Sum(Benchmark size)

We can observe that:

a) **The MoSLoRA are more robust than LoRA and MoSLoRA_Init regarding different seeds**.
b) **For both Average and Weighted Average, MoSLoRA outperforms LoRA and MoSLoRA_Init (about 1.5 higher)**.

-----


Also, as mentioned in https://github.com/huggingface/peft/pull/1905#issuecomment-2211677146, we show that:

i) Though mathematically equivalent initialized, the optimization process would be different if W is learnable. The proof can be found at [here](https://github.com/wutaiqiang/MoSLoRA/blob/main/MoSLoRA_proof.pdf).

ii) We can consider vanilla LoRA as $A I B$, where $I$ is a **fixed identity** matrix. In MoSLoRA, we upgrade **fixed $I$** to a **learnable** $W$.

We consider these as why MoSLoRA outperforms LoRA.

BenjaminBossan:
Thank you for running these additional experiments. I agree that the average difference in score looks more substantial here. Still, from looking at the numbers, this appeared to be driven by outliers. For that reason, I plotted the scores for each method to get a better feel. We have the same tasks from left to right as in your table, just that the scores are sorted within each task from lowest to highest. Here is the figure:

image

As you can see, the scores are mostly identical except for a handful of outliers. Those are more frequent for LoRA and init-MosLoRA, which basically explains the entire difference in scores.

To illustrate this, I plotted only the top 3 scores for each task:

image

I know this is a bit of cherry-picking and maybe the advantage of MosLoRA is that it results in fewer outliers. But I'm not convinced that this isn't just random. For instance, as I [reported](https://github.com/huggingface/peft/pull/1905#issuecomment-2298499321) in my earlier experiments, it was MosLoRA with the only negative outlier out of 20 tested seeds.

I tried to run the experiments you shared with more seeds but unfortunately my current setup runs into OOM. IMO, ideally, we could have a sufficiently large run of experiments to show a significant difference between the methods. For the values you provided, the difference is far from significant (using rank-sum test) but that's expected given the sample size.

Again, I really appreciate the discussion and don't want to discourage your contribution. To give more context, as a maintainer I need to ensure that adding new methods provides enough value to compensate for the increase in complexity of the code base. Since MosLoRA directly modifies the LoRA code, the complexity cost is higher than for other methods that provide separate implementations. And even though you showed that from the gradient perspective, MosLoRA is not identical to LoRA, I think you can see that just from the forward path, it is pretty much identical. And I don't see how the change in gradient is actually beneficial overall. I hope this makes sense.

wutaiqiang:
Thanks for your responses. I agree that the MoSLoRA brings higher complexity since MosLoRA directly modifies the LoRA code.

> Still, from looking at the numbers, this appeared to be driven by outliers. 

When I sorted out the experimental results, I also found that there are some outliers for all the three methods. Therefore, I  checked the generated results and found that the fine-tuned LLM just answer wrong. The training on such QA task is quite sensitive and hard to fit well. Obviously, MoSLoRA is more robust than LoRA and MoSLoRA_init  with less outliers.

> I tried to run the experiments you shared with more seeds but unfortunately my current setup runs into OOM. 

Sorry for that as it requires around 50G GPU Memory. I also try the 1.1B TinyLLaMA, but the results are quite low. That is why I use the 3.8B Phi-3. (I think the reason is that such reasoning ability and instruction-following ability require more parameters than 1.1B).

>  I think you can see that just from the forward path, it is pretty much identical.

Indeed, the forward is the same **only for the first batch**,  and the gradients are not the same. Therefore, for the second batch, things would be different and the forward process are not the same. 

>  run the experiments you shared with more seeds.

Let me run for more seeds. I would like to run seeds another 10 seeds.

> Again, I really appreciate the discussion and don't want to discourage your contribution. 

Thanks for your kindness. I agree that new methods should be checked carefully. Please wait for more testing results :)

BenjaminBossan:
Thanks for your kind response and your patience @wutaiqiang. I'm looking forward to seeing the outcome of the further experiments.

wutaiqiang:
@BenjaminBossan Thanks for your patience. I regret to say that the A100 GPUs in my lab are employed for other tasks. I can not finish the additional experiments. Moreover, I would like to introduce two pieces of evidence to prove the effectiveness of MoSLoRA.

i) In the paper TeamLoRA (https://arxiv.org/pdf/2408.09856v1), they compare the LoRA and MoSLoRA, and the results are:

image

Compared to LoRA, our proposed MoSLoRA improves by 0.53 regarding the average score. I do not know any of the authors of TeamLoRA, the comparison is independent.

ii) The paper MoSLoRA is accepted by the EMNLP 2024 with a score of 3.5/4/4.5 (full: 5). 
image

I do understand that ```Since MosLoRA directly modifies the LoRA code, the complexity cost is higher than for other methods that provide separate implementations.```  If you still worry about this, I think maybe I can implement MoSLoRA in separate files and pull another request for PEFT.

BenjaminBossan:
Thanks for the latest updates and I fully understand that it can be hard to get the resources to run all these experiments.

> If you still worry about this, I think maybe I can implement MoSLoRA in separate files and pull another request for PEFT.

If this is not too much work for you, this would be a good compromise.

wutaiqiang:
Okay. I will pull another request after I finish the implementations. Thanks for your help.

PR #2021, opened 2024-08-20
[tests] enable `test_mixed_adapter_batches_lora_opt_timing` on XPU

faaany:
After Fix:
```bash
====================================== short test summary info ======================================
PASSED tests/test_custom_models.py::TestMixedAdapterBatches::test_mixed_adapter_batches_lora_opt_timing
=================================== 1 passed, 1 warning in 7.59s ====================================
```

Just like other tests in this file, this function should not only apply to NV GPU. We can actually remove this test marker.

### DIFF
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -3276,7 +3276,6 @@ class TestMixedAdapterBatches:
         with pytest.raises(ValueError, match=msg):
             peft_model.forward(**inputs)
 
-    @require_torch_gpu
     def test_mixed_adapter_batches_lora_opt_timing(self):
         # Use a more realistic model (opt-125m) and do a simple runtime check to ensure that mixed adapter batches
         # don't add too much overhead. These types of tests are inherently flaky, so we try to add in some robustness.


faaany:
@BenjaminBossan

BenjaminBossan (commented):
The issue with this change is that this would mean the test also runs on CPU. As the comment further below indicates, we want to avoid this to prevent flakiness:

https://github.com/huggingface/peft/blob/eb5eb6efb527c8ef8df26798178b81e3ac8f3662/tests/test_custom_models.py#L3340-L3341

I tried the test again just now on CPU and the time differences are indeed much smaller (~25%) compared to GPU (~150%), so this is still true. One solution would be to check if either GPU or XPU is being used.

faaany:
> The issue with this change is that this would mean the test also runs on CPU. As the comment further below indicates, we want to avoid this to prevent flakiness:
> 
> https://github.com/huggingface/peft/blob/eb5eb6efb527c8ef8df26798178b81e3ac8f3662/tests/test_custom_models.py#L3340-L3341
> 
> I tried the test again just now on CPU and the time differences are indeed much smaller (~25%) compared to GPU (~150%), so this is still true. One solution would be to check if either GPU or XPU is being used.

Sure, should I add a marker called "require_torch_accelerator" just like in accelerate?

faaany:
> > Sure, should I add a marker called "require_torch_accelerator" just like in accelerate?
> 
> Hmm, I think let's not go that far yet, a simple check at the start of the test function for the device is more explicit, we can add an extra decorator if we need the same check in many places.


Sure, let me update!

faaany:
@BenjaminBossan  how about the `require_non_cpu` solution I introduced in this PR https://github.com/huggingface/peft/pull/2026 ?

BenjaminBossan:
> how about the `require_non_cpu` solution I introduced in this PR #2026 ?

Sounds good, I merged that PR.

faaany:
> Thanks for extending this test, LGTM.
> 
> Btw. is there some public place where I can check the XPU tests?

do you mean the test summary on XPU?

BenjaminBossan:
> do you mean the test summary on XPU?

Yes, so that me and others can check the current state.

faaany:
> > do you mean the test summary on XPU?
> 
> Yes, so that me and others can check the current state.

Sure, currently we don't upload the test results to a public repo. But let me come out with a solution. Talk to you next Monday.

BenjaminBossan:
> Sure, currently we don't upload the test results to a public repo. But let me come out with a solution. Talk to you next Monday.

Thanks. It's not super high priority, but right now if we ever break something in PEFT for XPU, we won't know until someone comes to us to report it.

yao-matrix:
> > Sure, currently we don't upload the test results to a public repo. But let me come out with a solution. Talk to you next Monday.
> 
> Thanks. It's not super high priority, but right now if we ever break something in PEFT for XPU, we won't know until someone comes to us to report it.

We have a 2-step plan:
1. screen HF libraries(transformers, accelerate, peft, diffusers, trl) UTs, run in XPU, and fix or extend if we target them run on XPU
2. work w/ Hugging Face to provide CI machines and integrate into CI flow
We are in 1st now, and progress is transformers & accelerate & peft are done, diffusers & trl are undergoing. Once this step is done, we are glad to work w/ you guys to enable CI, :).

BenjaminBossan:
Thanks for the update on the plan.

Issue #2018, opened 2024-08-19
With tied embeddings adapter merged to tied layers


ltoniazzi:
### System Info

`peft=0.12.0`
`transformers =4.44.0`

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [x] My own task or dataset (give details below)

### Reproduction

With Gemma2, a model where `tie_word_embeddings = True`, using `target_modules=["lm_head"]` and merging the adapter  leads to merging the adapter to the tied/embedding layer, which is incorrect.

```python
from transformers import AutoModelForCausalLM
import torch
model = AutoModelForCausalLM.from_pretrained("google/gemma-2-2b-it")
# model.config.tie_word_embeddings = False # doing this does not change the outcome

config = LoraConfig(
    r=8,
    lora_alpha=16,
    target_modules=["lm_head"],
    bias="none",
    task_type="CAUSAL_LM",
    init_lora_weights=False,
)
clone_lm_head_orig = model.lm_head.weight.data.clone()
model = get_peft_model(model, config)

# Check embed_tokens and lm_head point to the same data
assert model.model.model.embed_tokens.weight.data.data_ptr() == model.model.lm_head.weight.data.data_ptr()

# Merge adapter in the base model
model.merge_and_unload(safe_merge=True, adapter_names=["default"])

# Check adapter is merged
assert not torch.equal(clone_lm_head_orig, model.model.lm_head.weight.data)

# Check embedding layer is unchanged by the lm_head adapter merging
assert model.model.model.embed_tokens.weight.data.data_ptr() != model.model.lm_head.weight.data.data_ptr(), "Embedding layer should have not changed"

```

### Expected behavior

I think that merging should not succeed silently, but either a:
- succeed and a warning is raised, that you are merging adapters to layers that are tied to other layers,
- error raised,
- merging should be performed correctly (though it might complicate saving the model as a new layer will be present).

### Related issues

- https://github.com/huggingface/peft/issues/1035
- https://github.com/huggingface/peft/issues/1750

BenjaminBossan:
Thanks for opening this issue. Yes, I agree that this is an easy source of errors, and having a warning would help.

The main reason why this is not implemented yet is that merging is a layer-level operation in PEFT. The individual layer can, however, not know if its weights are tied are not. Therefore, we cannot easily check for this. It could be possible to refactor this to work differently but I don't see an easy way.

[...]

PR #2025, opened 2024-08-20
Warn if using tied target module with `tie_word_embeddings`

ltoniazzi:
### Context

Solving issue #2018.

- Raise a warning if the user is requesting an output `target_module` when the embeddings are tied, because this could lead to errors, for example when merging the adapter.
- Also refactored the code to get the model config

#### Todo

- [x] ~~Try if load with `tie_word_embeddings=False` is an actual option. Load Gemma2 with finetuned different `lm_weights` and check that the lm_head is not replaced with the embedding (even if cloned). If it works, try to merge an adapter to lm_weight and then load it to check if embed and lm_head are kept separate. (the main concern is that the loading model's architecture might ignore any `lm_head` weight present in safetensors, as it happens in llama.cpp for example).~~
- [x] In the end I only checked how to save the base model as untied, then the user can work from there, although I DID NOT DO  A THOROUGH CHECK OF THIS APPROACH, only checked this:
```python
import torch

# Load original tied model
model = AutoModelForCausalLM.from_pretrained("google/gemma-2-2b-it", tie_word_embeddings=False)
# Set the randomly initialized lm_head to the previously tied embeddings
model.lm_head.weight.data = model.model.embed_tokens.weight.data.clone()
assert torch.equal(model.lm_head.weight.data, model.model.embed_tokens.weight.data)

# Save the untied model
untied_model_dir = "tmp_model"
model.save_pretrained(untied_model_dir)
model.config.save_pretrained(untied_model_dir)
# Now use the original model but in untied format
model = AutoModelForCausalLM.from_pretrained(untied_model_dir)

assert torch.equal(model.lm_head.weight.data, model.model.embed_tokens.weight.data)
assert model.model.embed_tokens.weight.data.data_ptr() != model.lm_head.weight.data.data_ptr()
```

- [x] Add in warning about porting to other formats.

### DIFF
--- a/src/peft/mapping.py
+++ b/src/peft/mapping.py
@@ -131,7 +131,6 @@ def get_peft_config(config_dict: dict[str, Any]) -> PeftConfig:
     return PEFT_TYPE_TO_CONFIG_MAPPING[config_dict["peft_type"]](**config_dict)
 
 def warn_if_tied_embeddings_in_target_modules(model_config, peft_config):
-    
     if model_config.get("tie_word_embeddings"):
         for target_module in peft_config.target_modules:
             if target_module in EMBEDDING_LAYER_NAMES:
--- a/src/peft/mapping.py
+++ b/src/peft/mapping.py
@@ -130,6 +130,17 @@ def get_peft_config(config_dict: dict[str, Any]) -> PeftConfig:
 
     return PEFT_TYPE_TO_CONFIG_MAPPING[config_dict["peft_type"]](**config_dict)
 
+def warn_if_tied_embeddings_in_target_modules(model_config, peft_config):
+    common_output_target_module = [
+        "lm_head",
+    ]
+    
+    if model_config.get("tie_word_embeddings"):
+        for target_module in peft_config.target_modules:
+            if target_module in common_output_target_module:
+                warnings.warn(
+                    f"{model_config['tie_word_embeddings']=} and a tied {target_module=} is passed to peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to infect and adapter to {target_module}?"
+                )
 
 def get_peft_model(
     model: PreTrainedModel,
@@ -166,6 +177,9 @@ def get_peft_model(
     old_name = peft_config.base_model_name_or_path
     new_name = model.__dict__.get("name_or_path", None)
     peft_config.base_model_name_or_path = new_name
+
+    warn_if_tied_embeddings_in_target_modules(model_config, peft_config)
+
     if (old_name is not None) and (old_name != new_name):
         warnings.warn(
             f"The PEFT config's `base_model_name_or_path` was renamed from '{old_name}' to '{new_name}'. "
--- a/src/peft/mapping.py
+++ b/src/peft/mapping.py
@@ -65,7 +65,7 @@ from .tuners import (
     XLoraConfig,
 )
 from .tuners.tuners_utils import BaseTuner as _BaseTuner
-from .utils import _prepare_prompt_learning_config, EMBEDDING_LAYER_NAMES
+from .utils import _prepare_prompt_learning_config
 
 
 if TYPE_CHECKING:
@@ -130,13 +130,6 @@ def get_peft_config(config_dict: dict[str, Any]) -> PeftConfig:
 
     return PEFT_TYPE_TO_CONFIG_MAPPING[config_dict["peft_type"]](**config_dict)
 
-def warn_if_tied_embeddings_in_target_modules(model_config, peft_config):
-    if model_config.get("tie_word_embeddings"):
-        for target_module in peft_config.target_modules:
-            if target_module in EMBEDDING_LAYER_NAMES:
-                warnings.warn(
-                    f"{model_config['tie_word_embeddings']=} and a tied {target_module=} is passed to peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to use the target module {target_module}?"
-                )
 
 def get_peft_model(
     model: PreTrainedModel,
@@ -173,8 +166,6 @@ def get_peft_model(
     old_name = peft_config.base_model_name_or_path
     new_name = model.__dict__.get("name_or_path", None)
     peft_config.base_model_name_or_path = new_name
-
-    warn_if_tied_embeddings_in_target_modules(model_config, peft_config)
 
     if (old_name is not None) and (old_name != new_name):
         warnings.warn(
--- a/src/peft/mapping.py
+++ b/src/peft/mapping.py
@@ -132,7 +132,7 @@ def get_peft_config(config_dict: dict[str, Any]) -> PeftConfig:
 
 def warn_if_tied_embeddings_in_target_modules(model_config, peft_config):
     common_output_target_module = [
-        "lm_head",
+        "lm_head", "embed_tokens",
     ]
     
     if model_config.get("tie_word_embeddings"):
--- a/src/peft/mapping.py
+++ b/src/peft/mapping.py
@@ -136,7 +136,7 @@ def warn_if_tied_embeddings_in_target_modules(model_config, peft_config):
         for target_module in peft_config.target_modules:
             if target_module in EMBEDDING_LAYER_NAMES:
                 warnings.warn(
-                    f"{model_config['tie_word_embeddings']=} and a tied {target_module=} is passed to peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to infect and adapter to {target_module}?"
+                    f"{model_config['tie_word_embeddings']=} and a tied {target_module=} is passed to peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to use the  target module {target_module}?"
                 )
 
 def get_peft_model(
--- a/src/peft/mapping.py
+++ b/src/peft/mapping.py
@@ -135,7 +135,7 @@ def warn_if_tied_embeddings_in_target_modules(model_config, peft_config):
         for target_module in peft_config.target_modules:
             if target_module in EMBEDDING_LAYER_NAMES:
                 warnings.warn(
-                    f"{model_config['tie_word_embeddings']=} and a tied {target_module=} is passed to peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to use the  target module {target_module}?"
+                    f"{model_config['tie_word_embeddings']=} and a tied {target_module=} is passed to peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to use the target module {target_module}?"
                 )
 
 def get_peft_model(
--- a/src/peft/mapping.py
+++ b/src/peft/mapping.py
@@ -64,7 +64,7 @@ from .tuners import (
     VeraModel,
     XLoraConfig,
 )
-from .tuners.tuners_utils import BaseTuner as _BaseTuner
+from .tuners.tuners_utils import BaseTuner
 from .utils import _prepare_prompt_learning_config
 
 
@@ -103,7 +103,7 @@ PEFT_TYPE_TO_CONFIG_MAPPING: dict[str, type[PeftConfig]] = {
     "HRA": HRAConfig,
 }
 
-PEFT_TYPE_TO_TUNER_MAPPING: dict[str, type[_BaseTuner]] = {
+PEFT_TYPE_TO_TUNER_MAPPING: dict[str, type[BaseTuner]] = {
     "LORA": LoraModel,
     "LOHA": LoHaModel,
     "LOKR": LoKrModel,
@@ -159,10 +159,7 @@ def get_peft_model(
             The revision of the base model. If this isn't set, the saved peft model will load the `main` revision for
             the base model
     """
-    model_config = getattr(model, "config", {"model_type": "custom"})
-    if hasattr(model_config, "to_dict"):
-        model_config = model_config.to_dict()
-
+    model_config = BaseTuner.get_model_config(model)
     old_name = peft_config.base_model_name_or_path
     new_name = model.__dict__.get("name_or_path", None)
     peft_config.base_model_name_or_path = new_name
--- a/src/peft/mapping.py
+++ b/src/peft/mapping.py
@@ -65,7 +65,7 @@ from .tuners import (
     XLoraConfig,
 )
 from .tuners.tuners_utils import BaseTuner as _BaseTuner
-from .utils import _prepare_prompt_learning_config
+from .utils import _prepare_prompt_learning_config, EMBEDDING_LAYER_NAMES
 
 
 if TYPE_CHECKING:
@@ -131,13 +131,10 @@ def get_peft_config(config_dict: dict[str, Any]) -> PeftConfig:
     return PEFT_TYPE_TO_CONFIG_MAPPING[config_dict["peft_type"]](**config_dict)
 
 def warn_if_tied_embeddings_in_target_modules(model_config, peft_config):
-    common_output_target_module = [
-        "lm_head", "embed_tokens",
-    ]
     
     if model_config.get("tie_word_embeddings"):
         for target_module in peft_config.target_modules:
-            if target_module in common_output_target_module:
+            if target_module in EMBEDDING_LAYER_NAMES:
                 warnings.warn(
                     f"{model_config['tie_word_embeddings']=} and a tied {target_module=} is passed to peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to infect and adapter to {target_module}?"
                 )
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -447,6 +447,7 @@ class LoraModel(BaseTuner):
 
         Currently gptq quantization and replicated layers do not support merging.
         """
+        self._warn_if_tied_embeddings_in_target_modules(self.model)
         if getattr(self.model, "quantization_method", None) == "gptq":
             raise ValueError("Cannot merge LORA layers when the model is gptq quantized")
         if self.peft_config.get("layer_replication"):
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -447,7 +447,15 @@ class LoraModel(BaseTuner):
 
         Currently gptq quantization and replicated layers do not support merging.
         """
-        self._warn_if_tied_embeddings_in_target_modules(self.model)
+        tied_target_modeules = self._get_tied_target_modeules(self.model)
+        if tied_target_modeules:
+            warnings.warn(
+                f"Model with `tie_word_embeddings=True` and the {tied_target_modeules=} are part of the adapter. "
+                "This can lead to complications when merging the adapter. "
+                "You can opt to merge the adapter after cloning the weights (to untie the embeddings), "
+                "and then load the merged model `tie_word_embeddings=False`: "
+                "\n```python\nAutoModelForCausalLM.from_pretrained('path/to/merged/model', tie_word_embeddings=False)\n```\n"
+            )
         if getattr(self.model, "quantization_method", None) == "gptq":
             raise ValueError("Cannot merge LORA layers when the model is gptq quantized")
         if self.peft_config.get("layer_replication"):
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -447,15 +447,7 @@ class LoraModel(BaseTuner):
 
         Currently gptq quantization and replicated layers do not support merging.
         """
-        tied_target_modules = self._get_tied_target_modules(self.model)
-        if tied_target_modules:
-            warnings.warn(
-                f"Model with `tie_word_embeddings=True` and the {tied_target_modules=} are part of the adapter. "
-                "This can lead to complications when merging the adapter. "
-                "You can opt to merge the adapter after cloning the weights (to untie the embeddings), "
-                "and then load the merged model with `tie_word_embeddings=False`: "
-                "\n```python\nAutoModelForCausalLM.from_pretrained('path/to/merged/model', tie_word_embeddings=False)\n```\n"
-            )
+        super()._check_merge_allowed()
         if getattr(self.model, "quantization_method", None) == "gptq":
             raise ValueError("Cannot merge LORA layers when the model is gptq quantized")
         if self.peft_config.get("layer_replication"):
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -453,7 +453,7 @@ class LoraModel(BaseTuner):
                 f"Model with `tie_word_embeddings=True` and the {tied_target_modeules=} are part of the adapter. "
                 "This can lead to complications when merging the adapter. "
                 "You can opt to merge the adapter after cloning the weights (to untie the embeddings), "
-                "and then load the merged model `tie_word_embeddings=False`: "
+                "and then load the merged model with `tie_word_embeddings=False`: "
                 "\n```python\nAutoModelForCausalLM.from_pretrained('path/to/merged/model', tie_word_embeddings=False)\n```\n"
             )
         if getattr(self.model, "quantization_method", None) == "gptq":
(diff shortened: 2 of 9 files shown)


src/peft/mapping.py
```diff
@@ -130,6 +130,14 @@ def get_peft_config(config_dict: dict[str, Any]) -> PeftConfig:
 
     return PEFT_TYPE_TO_CONFIG_MAPPING[config_dict["peft_type"]](**config_dict)
 
+def warn_if_tied_embeddings_in_target_modules(model_config, peft_config):
+    
+    if model_config.get("tie_word_embeddings"):
+        for target_module in peft_config.target_modules:
+            if target_module in EMBEDDING_LAYER_NAMES:
+                warnings.warn(
+                    f"{model_config['tie_word_embeddings']=} and a tied {target_module=} is passed to peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to infect and adapter to {target_module}?"
```
ltoniazzi, line 139:
```suggestion
                    f"{model_config['tie_word_embeddings']=} and a tied {target_module=} is passed to peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to use the  target module {target_module}?"
```

src/peft/mapping.py
```diff
@@ -130,6 +130,13 @@ def get_peft_config(config_dict: dict[str, Any]) -> PeftConfig:
 
     return PEFT_TYPE_TO_CONFIG_MAPPING[config_dict["peft_type"]](**config_dict)
 
+def warn_if_tied_embeddings_in_target_modules(model_config, peft_config):
+    if model_config.get("tie_word_embeddings"):
+        for target_module in peft_config.target_modules:
+            if target_module in EMBEDDING_LAYER_NAMES:
+                warnings.warn(
+                    f"{model_config['tie_word_embeddings']=} and a tied {target_module=} is passed to peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to use the  target module {target_module}?"
```
ltoniazzi, line 138:
```suggestion
                    f"{model_config['tie_word_embeddings']=} and a tied {target_module=} is passed to peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to use the target module {target_module}?"
```

src/peft/mapping.py
```diff
@@ -130,6 +130,13 @@ def get_peft_config(config_dict: dict[str, Any]) -> PeftConfig:
 
     return PEFT_TYPE_TO_CONFIG_MAPPING[config_dict["peft_type"]](**config_dict)
 
+def warn_if_tied_embeddings_in_target_modules(model_config, peft_config):
+    if model_config.get("tie_word_embeddings"):
```
ltoniazzi, line 134: @BenjaminBossan This works if `model_config` is always a `dict`. Is this true?

BenjaminBossan: Yes, we can assume that, since earlier we have:

```python
    model_config = getattr(model, "config", {"model_type": "custom"})
    if hasattr(model_config, "to_dict"):
        model_config = model_config.to_dict()
```

As long as your code takes into account that `model_config` could just be `{"model_type": "custom"}` (in case this is a custom model), it's fine.

src/peft/mapping.py
```diff
@@ -130,6 +130,13 @@ def get_peft_config(config_dict: dict[str, Any]) -> PeftConfig:
...
+        for target_module in peft_config.target_modules:
+            if target_module in EMBEDDING_LAYER_NAMES:
+                warnings.warn(
+                    f"{model_config['tie_word_embeddings']=} and a tied {target_module=} is passed to peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to use the target module {target_module}?"
```
ltoniazzi, line 138:
```suggestion
                    f"{model_config['tie_word_embeddings']=} and the tied {target_module=} is in the peft config. This can lead to complications, for example when merging the adapter. Are you sure you want to use the target module {target_module}?"
```

src/peft/mapping.py
ltoniazzi: The code will work but it will not be able to find out if embeddings are tied in the case `model_config={"model_type": "custom"}`.

BenjaminBossan (changes requested):
Thanks for creating this PR. I think we need to rethink the approach here, as the current one will not work in all situations.

1. `get_peft_model` is a very generic function and is also used for prompt tuning methods, for instance. Therefore, we cannot assume that `peft_config.target_modules` exist.
2. Not all methods allow to merge the weights, thus we should not warn in those cases (false warnings should be avoided as much as possible).
3. Even if `peft_config.target_modules` does exist, it could be a string, so looping over it will not always be correct.
4. As we already observed, it will not work for custom models with tied weights, but let's consider this out of scope for now.

So how can we correctly identify when a warning is needed? My proposal is that this needs to be solved on a different level:

The check if there is a tied target layer needs to live on the corresponding method's model level (e.g. `LoraModel`), as only there can we really know which layers are targeted. Thankfully, the models that support merging all inherit from `BaseTuner`. There, we have the [`inject_adapter`](https://github.com/huggingface/peft/blob/95821e5ce4bf47366bbcb533b384f093f147b5ed/src/peft/tuners/tuners_utils.py#L366) method. If you look at [this line](https://github.com/huggingface/peft/blob/95821e5ce4bf47366bbcb533b384f093f147b5ed/src/peft/tuners/tuners_utils.py#L428), you can see that all modules that are actually targeted are stored in `self.targeted_module_names`. Therefore, after exiting the loop, we can add a new method that takes this list and checks if any of the keys are tied weights using the logic you proposed.

This new check should be implemented as a new method on the `BaseTuner` class, so that subclasses such as `LoraModel` may choose to override the method if there ever is a need.

Additionally, I wonder if there should be a warning when the user attempts to merge. One could argue that this is too late, but even at this point, there are workarounds: If the user clones the tied weights, they can merge without affecting the other weight (at the cost of extra memory).

This additional warning could be added to the [`_check_merge_allowed`](https://github.com/huggingface/peft/blob/95821e5ce4bf47366bbcb533b384f093f147b5ed/src/peft/tuners/tuners_utils.py#L359) method and it could re-use the same method as mentioned above to perform the check. However, the warning message should be a bit different.

I know this is all a bit more complicated that initially thought and not necessarily what you "signed up for". So let me know if you still want to work on this or not, in which case I'll put this on my backlog.

ltoniazzi:
> I know this is all a bit more complicated that initially thought and not necessarily what you "signed up for".

Not at all thanks, sounds really good, I'll have a go!

BenjaminBossan:
> Not at all thanks, sounds really good, I'll have a go!

Thanks a lot.

ltoniazzi:
@BenjaminBossan I made a version addressing your suggestions. (I can't run the whole test suite as I do not have a cuda-compatible gpu). Also, I refactored getting the model config in the code base.

> However, the warning message should be a bit different.

I feel like the new message can be the same. Let me know.

src/peft/tuners/tuners_utils.py
```diff
@@ -494,6 +494,24 @@ def _unloading_checks(self, adapter_names: Optional[list[str]]):
         if is_modules_to_save_available and len(adapters_to_consider) > 1:
             raise ValueError("Cannot unload multiple adapters that specify `modules_to_save`.")
 
+    @staticmethod
+    def get_model_config(model, default={"model_type": "custom"}):
```
BenjaminBossan, line 498: Good idea. Let's add a docstring that explains what this returns, mentioning that the config will be converted to a dict and that custom models will just get a dummy config.

BenjaminBossan (changes requested):
Thanks for updating the PR.

> I feel like the new message can be the same. Let me know.

I think the error message is good as is for when the model is being initialized. When merging, I think we could show a different warning, where we mention that if the weight is cloned beforehand, merging should work, at the cost of higher memory usage.

To implement this, I would change the `_warn_if_tied_embeddings_in_target_modules` method from warning to just performing the check and returning a bool (renaming the method accordingly). Then during injection, if the check returns `True`, the current warning is given, and during merging, if the check returns True, the adapted warning is given. WDYT?

> (I can't run the whole test suite as I do not have a cuda-compatible gpu.)

This is fine.

src/peft/tuners/tuners_utils.py
ltoniazzi: Addressed the comments 👍

src/peft/tuners/tuners_utils.py
```diff
@@ -494,6 +500,31 @@ def _unloading_checks(self, adapter_names: Optional[list[str]]):
         if is_modules_to_save_available and len(adapters_to_consider) > 1:
             raise ValueError("Cannot unload multiple adapters that specify `modules_to_save`.")
 
+    @staticmethod
+    def get_model_config(model: nn.Module, default: dict | None = {"model_type": "custom"}) -> dict:
```
BenjaminBossan, line 504: Let's avoid having a dictionary as a default argument. Generally in Python, we should not use mutable default arguments. I think the best alternative is to simply remove `default` as an argument entirely, just hard-code `{"model_type": "custom"}`.

I guess we could also create a constant `{"model_type": "custom"}` inside of `utils/constants.py` and use it consistently in the code base, up to you if you want to go that far.

src/peft/tuners/tuners_utils.py
```diff
@@ -494,6 +500,31 @@ def _unloading_checks(self, adapter_names: Optional[list[str]]):
...
+        """
+        This method gets the config from a model in dictionary form.
```
BenjaminBossan, line 506: Let's add that `{"model_type": "custom"}` is returned if no config is found.

src/peft/tuners/lora/model.py
```diff
@@ -447,6 +447,15 @@ def _check_merge_allowed(self):
 
         Currently gptq quantization and replicated layers do not support merging.
         """
+        tied_target_modules = self._get_tied_target_modules(self.model)
```
BenjaminBossan, line 450: I think it's better to move this check to `BaseTuner` here:

https://github.com/huggingface/peft/blob/f3c7c6e5c1e1b543e1f006ef38b3fdab3a4fefaf/src/peft/tuners/tuners_utils.py#L359-L364

That way, all models that inherit from it, not just LoRA, benefit from the warning. For LoRA, we need of course to call `super()._check_merge_allowed()` in that case.

src/peft/tuners/tuners_utils.py
```diff
@@ -494,6 +500,31 @@ def _unloading_checks(self, adapter_names: Optional[list[str]]):
...
+                Model to get the config from.
+            default (`dict|None`, *optional*)::
+                What to return if model does not have a config attribute.
+        """
+        model_config = getattr(model, "config", default)
+        if hasattr(model_config, "to_dict"):
+            model_config = model_config.to_dict()
+        return model_config
+
+    def _get_tied_target_modules(self, model):
```
BenjaminBossan, line 519:
```suggestion
    def _get_tied_target_modules(self, model) -> list[str]:
```

BenjaminBossan (changes requested):
Thanks for the changes. I have a few small suggestions for improvements, please check them out?

It would also be great to add unit tests for this, but probably this will be a bit more complicated. I leave it up to you if you want to give this a try, otherwise I'll work on it in a subsequent PR.

src/peft/tuners/tuners_utils.py
ltoniazzi: I did this as in one case [here](https://github.com/huggingface/peft/blob/900f96c40ddebae9d76bed374c8baed60e8b34e9/src/peft/peft_model.py#L1234), `default` is `None`. 

I'll try handling that `None` separately.

ltoniazzi: I just did [this](https://github.com/huggingface/peft/pull/2025/files#diff-95787fcc5356568929e383449ade2c1bac5da43deccab17318652b62ed171ae7R1237) as it's a bit unclear if in this case the `model_config` needs to default to `None` or if it can be the DUMMY one, let me know!

ltoniazzi:
Sure very happy to write tests! I'll put them in `tests/test_tuners_utils.py`. 

Just one question: to mock models with tied embeddings, should I use the test model `"HuggingFaceH4/tiny-random-LlamaForCausalLM"` but loaded with:
```python
model = AutoModelForCausalLM.from_pretrained("microsoft/Phi-3-mini-4k-instruct", tie_word_embeddings=True)
```

ltoniazzi:
@BenjaminBossan I added the test here `tests/test_tuners_utils.py` 👍

tests/test_tuners_utils.py
```diff
@@ -1065,3 +1068,119 @@ def forward(self, X):
...
+            triggered=False,
+        )
+        assert_warning_triggered(
+            records_merge_not_tied,
+            warn_start=warn_start,
+            warn_end=warn_end_merge,
+            triggered=False,
+        )
+
+        # No warning when loading model with tie_word_embeddings although but not relevant target module
```
ltoniazzi, line 1166:
```suggestion
        # No warning when loading model with tie_word_embeddings but not relevant target module
```

src/peft/tuners/tuners_utils.py
```diff
@@ -361,7 +361,15 @@ def _check_merge_allowed(self):
 
         Raise a ValueError if it is not possible to merge the adapter with the given configuration.
         """
-        pass
+        tied_target_modules = self._get_tied_target_modules(self.model)
+        if tied_target_modules:
+            warnings.warn(
+                f"Model with `tie_word_embeddings=True` and the {tied_target_modules=} are part of the adapter. "
+                "This can lead to complications when merging the adapter. "
+                "You can opt to merge the adapter after cloning the weights (to untie the embeddings), "
```
BenjaminBossan, line 369: Honestly, I didn't know about the option to pass `tie_word_embeddings=False`. Is there even a need to clone the weights in that case?

tests/test_tuners_utils.py
```diff
@@ -1065,3 +1068,119 @@ def forward(self, X):
...
+
+    def test_get_model_config_as_dict(self):
+        config = BaseTuner.get_model_config(ModelWithDictConfig())
+        assert config == MockModelConfig.config
+
+    def test_get_model_config_with_no_config(self):
+        config = BaseTuner.get_model_config(ModelWithNoConfig())
+        assert config == DUMMY_MODEL_CONFIG
+
+    def test_warn_for_tied_embeddings_inject_and_merge(self):
```
BenjaminBossan, line 1108: Thanks a lot for adding these tests. They are already looking quite good. I think, however, that this last test can be simplified a bit.

As you correctly observed, there are 6 scenarios to test:

- Warning for `get_peft_model` and warning for merging.
- Valid warning vs no tied embeddings vs tied embeddings but not targeted.

Instead of cramming those into a single test, let's make this 6 separate tests. It should also be fine to make it 3 tests, where `get_peft_model` and merging are checked together. Hopefully, this should make the `assert_warning_triggered` function unnecessary.

You probably also had a bit of an issue that unrelated warnings could be recorded. Maybe this can be made simpler by using the `recwarn` fixture. Then you can just check that any warning has been recorded with the corresponding message, something like:

`assert any(str(warning.message).startswith(msg) for warning in recwarn.list)`

tests/test_tuners_utils.py
```diff
@@ -1065,3 +1068,119 @@ def forward(self, X):
...
+class ModelWithDictConfig(nn.Module):
+    def __init__(self):
+        self.config = MockModelConfig.config
+
+
+class ModelWithNoConfig(nn.Module):
+    pass
+
+
+class TestBaseTunerMethods(unittest.TestCase):
```
BenjaminBossan, line 1095: Let's split this test class into 2: One for `get_model_config` and one for the tied embeddings.

BenjaminBossan (changes requested):
Thanks so much for making the updates, using `DUMMY_MODEL_CONFIG` consistently and extending the tests. This looks quite good already, but I have some suggestions for improvements, please chekc.

> I just did [this](https://github.com/huggingface/peft/pull/2025/files#diff-95787fcc5356568929e383449ade2c1bac5da43deccab17318652b62ed171ae7R1237) as it's a bit unclear if in this case the model_config needs to default to None or if it can be the DUMMY one, let me know!

The change you made looks good as is.

> Just one question: to mock models with tied embeddings, should I use the test model "HuggingFaceH4/tiny-random-LlamaForCausalLM" but loaded with:

I didn't know that this was an option. Yes, looks like the right choice.

ltoniazzi:
@BenjaminBossan I think I've address the comments 👍

BenjaminBossan:
Thanks for the latest updates. I only have one more question, namely when it comes to how to untie the weights. In the script you provide, you clone the weights but is that even necessary if `tie_word_embeddings=False` is passed? To give an example, I tried this (I changed the model but only because I had already downloaded it):

```python
>>> model = AutoModelForCausalLM.from_pretrained("google/gemma-2b", tie_word_embeddings=False)
>>> model.model.embed_tokens.weight.data_ptr()
126062054867008
>>> model.lm_head.weight.data_ptr()  # <= different data ptr
126051845931072
>>> model.model.embed_tokens.weight.sum()
tensor(952564.6250, grad_fn=<SumBackward0>)
>>> model.lm_head.weight.sum()
tensor(255.3427, grad_fn=<SumBackward0>)

>>> from peft import LoraConfig, get_peft_model
>>> config = LoraConfig(init_lora_weights=False, target_modules=["embed_tokens"])
>>> model = get_peft_model(model, config)
>>> unloaded = model.merge_and_unload()
>>> unloaded.model.embed_tokens.weight.sum()  # <= embed weights changed
tensor(985655.8125)
>>> unloaded.lm_head.weight.sum()  # <= lm head stayed the same
tensor(255.3427)
```

ltoniazzi:
> In the script you provide, you clone the weights but is that even necessary if `tie_word_embeddings=False` is passed? To give an example, I tried this (I changed the model but only because I had already downloaded it):

Yes I agree with your script but the user wants to fix` lm_head` after loading. Because when loading with `tie_word_embeddings=False` the `lm_head` is randomly initialized, so the user wants to set it to equal the embed layer, which means they want a clone of it. 

This cloning also seems to allow to save it correctly. If you do not clone (beside actaully re-tieing the embeddings), then when you load the saved-untied model the last assertion below will fail, otherwise, if you clone, it will pass:
```python
model = AutoModelForCausalLM.from_pretrained("google/gemma-2-2b-it", tie_word_embeddings=False)
# Set the randomly initialized lm_head to the previously tied embeddings
model.lm_head.weight.data = model.model.embed_tokens.weight.data
assert torch.equal(model.lm_head.weight.data, model.model.embed_tokens.weight.data)

# Save the untied model
untied_model_dir = "tmp_model"
model.save_pretrained(untied_model_dir)
model.config.save_pretrained(untied_model_dir)
# Now use the original model but in untied format
model = AutoModelForCausalLM.from_pretrained(untied_model_dir)

assert model.model.embed_tokens.weight.data.data_ptr() != model.lm_head.weight.data.data_ptr()
assert torch.equal(model.lm_head.weight.data, model.model.embed_tokens.weight.data)
```

BenjaminBossan:
> Because when loading with `tie_word_embeddings=False` the `lm_head` is randomly initialized, so the user wants to set it to equal the embed layer, which means they want a clone of it.

Oh wow, I did not know that the LM head will be randomly initialized, that's quite surprising IMO. I would have expected to get the same parameter values, just not tied. Thanks for making me aware of that.

ltoniazzi:
Not sure how to reproduce [the error in the git actions](https://github.com/huggingface/peft/actions/runs/10578337704/job/29356715757?pr=2025)
```bash
ruff check src tests examples docs scripts docker
All checks passed!
ruff format --check src tests examples docs scripts docker
189 files already formatted
doc-builder style src/peft tests docs/source --max_len 119 --check_only
  File "/opt/hostedtoolcache/Python/3.8.18/x64/bin/doc-builder", line 8, in <module>
    sys.exit(main())
  File "/opt/hostedtoolcache/Python/3.8.[18](https://github.com/huggingface/peft/actions/runs/10578337704/job/29356715757?pr=2025#step:5:19)/x64/lib/python3.8/site-packages/doc_builder/commands/doc_builder_cli.py", line 47, in main
    args.func(args)
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/doc_builder/commands/style.py", line 28, in style_command
    raise ValueError(f"{len(changed)} files should be restyled!")
ValueError: 1 files should be restyled!
make: *** [Makefile:11: quality] Error 1
Error: Process completed with exit code 2.
```

BenjaminBossan:
@ltoniazzi could you please run `make style` to make the linter happy? The ruff version being used is 0.6.2.

BenjaminBossan:
I ran `make style` locally and this is the diff I get:

```diff
@@ -530,8 +530,8 @@ model = AutoModelForCausalLM.from_pretrained(untied_model_dir)
     @staticmethod
     def get_model_config(model: nn.Module) -> dict:
         """
-        This method gets the config from a model in dictionary form.
-        If model has not attribute config, then this method returns a default config.
+        This method gets the config from a model in dictionary form. If model has not attribute config, then this
+        method returns a default config.
```

ltoniazzi:
> could you please run make style

Done!

tests/test_tuners_utils.py
```diff
@@ -1065,3 +1067,85 @@ def forward(self, X):
...
+    )
+
+    def _get_peft_model(self, tie_word_embeddings, target_module):
+        model = get_peft_model(
+            AutoModelForCausalLM.from_pretrained(self.model_id, tie_word_embeddings=tie_word_embeddings),
+            LoraConfig(target_modules=[target_module]),
+        )
+        return model
+
+    def _is_warn_triggered(self, rrecwarn, endswith):
```
BenjaminBossan, line 1123: Did you call it `rrecwarn` to avoid naming conflicts? If yes, how about just passing the `recwarn.list`, which is all we need, and call it `warning_list` or so.

src/peft/tuners/tuners_utils.py
```diff
@@ -361,7 +361,33 @@ def _check_merge_allowed(self):
...
+
+# Save the untied model
+untied_model_dir = "dir/for/untied/model"
+model.save_pretrained(untied_model_dir)
+model.config.save_pretrained(untied_model_dir)
+
+# Now use the original model but in untied format
+model = AutoModelForCausalLM.from_pretrained(untied_model_dir)
+```
+"""
```
BenjaminBossan, line 389: I see why you left-aligned the code snippet so that it is nicely printed. But this is really an eye-sore to read in code. Here is a trick to that let's us use the correct indentation but still get a nice warning message by using `textwrap.dedent`:

```python
            example_code = textwrap.dedent(
                """
                ```python

                # Load original tied model
                model = AutoModelForCausalLM.from_pretrained("google/gemma-2-2b-it", tie_word_embeddings=False)

                # Set the randomly initialized lm_head to the previously tied embeddings
                model.lm_head.weight.data = model.model.embed_tokens.weight.data.clone()

                # Save the untied model
                untied_model_dir = "dir/for/untied/model"
                model.save_pretrained(untied_model_dir)
                model.config.save_pretrained(untied_model_dir)

                # Now use the original model but in untied format
                model = AutoModelForCausalLM.from_pretrained(untied_model_dir)
                ```
                """
            )
            warnings.warn(
                f"Model with `tie_word_embeddings=True` and the {tied_target_modules=} are part of the adapter. "
                "This can lead to complications. "
                "You can opt to merge the adapter after cloning the weights (to untie the embeddings). "
                "You can untie the embeddings by loading the model with `tie_word_embeddings=False`. For example:"
                + example_code
            )
```

The `textwrap` module is from the standardlib and needs to be imported.

BenjaminBossan (changes requested):
Thanks, very nicely done PR. I just have two tiny comments for cosmetic reasons, otherwise this can be merged.

src/peft/tuners/tuners_utils.py
ltoniazzi: Addressed both thanks!

BenjaminBossan (approved):
Thanks so much, great work, hopefully this will help users in the future to avoid this potential pitfall.

ltoniazzi:
@BenjaminBossan Thanks so much for your help! ❤️ 

Btw, [a test on main failed](https://github.com/huggingface/peft/actions/runs/10611843107/job/29412246020), do you think it's related to this PR?

BenjaminBossan:
> Btw, [a test on main failed](https://github.com/huggingface/peft/actions/runs/10611843107/job/29412246020), do you think it's related to this PR?

Don't worry, this is a known issue with X-LoRA that came about with a recent change in transformers.

PR #2028, opened 2024-08-21
ENH Raise error when applying modules_to_save on tuner layer

### LINKED ISSUE #2027: Unable to load peft checkpoint
### System Info

 **- transformers version: 4.44.0**
**-  peft 0.12.0**
- Platform: Linux-4.18.0-477.27.1.el8_8.x86_64-x86_64-with-glibc2.28
**- Python version: 3.9.4**
- Huggingface_hub version: 0.23.2
- Safetensors version: 0.4.3
**- Accelerate version: 0.33.0**
**- PyTorch version (GPU?): 2.4.0+cu121 (True)**
- Jax version: not installed
- Using GPU in script?: <fill in>
- GPU type: Tesla V100-FHHL-16GB

### Who can help?

@BenjaminBossan @saya

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

Hello, I am trying to load peft checkpoint and continue training, but I have this error:

```
site-packages/transformers/trainer.py:3108: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.
  torch.load(os.path.join(checkpoint, OPTIMIZER_NAME), map_location=map_location)
```

`**ValueError: loaded state dict contains a parameter group that doesn't match the size of optimizer's group**`

I tried google this error, but everyone's problem was, that they changed model's parameters somehow, but I didn't change anything. I use the same code and continue training with same Trainer. I use this:

`model = PeftModel.from_pretrained(model, checkpoint_path, is_trainable=True)`

```
trainer = Trainer(
        model=model,
        args=training_args,
        train_dataset=Train_tokenized,
        eval_dataset=Eval_tokenized,
        callbacks=[MetricsLoggingCallback(log_file),EarlyStoppingCallback(early_stopping_patience=2)]
    )

train_result = trainer.train(resume_from_checkpoint="checkpoint_path")
```

Thanks for any help.

### Expected behavior

----

BenjaminBossan:
Relates to #2027

Normally, when selecting the layers for fine-tuning, PEFT already ensures that the same layer is not targeted for both parameter-efficient fine-tuning (e.g. LoRA layer) and full fine-tuning (via `modules_to_save`), as that makes no sense.

However, there is a loophole when the `modules_to_save` is applied ex post. This happens for instance when having a task type like sequence classification, where PEFT will automatically add the classification head to `modules_to_save` for the user. This loophole is now closed by adding a check to `ModulesToSaveWrapper` that validates that the targeted layer is not a tuner layer.

This does not fully resolve #2027 but will raise an early error in the future to avoid confusion. **Edit**: #2028 should fully resolve it, but having this PR would still be useful.

On top of this, the error message inside of
`ModulesToSaveWrapper.check_module` has been slightly adjusted. Previously, the class name would be used, which can be confusing. E.g. for LoRA, the class name of the linear LoRA layer is just `"Linear"`, which looks the same as `nn.Linear`. Therefore, the full name is now shown.

Moreover, a now obsolete test was removed.

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -202,7 +202,15 @@ class ModulesToSaveWrapper(torch.nn.Module):
         # ModuleList, even though their forward methods cannot be called
         forbidden_classes = (torch.nn.ModuleDict, torch.nn.ModuleList, torch.nn.ParameterDict, torch.nn.ParameterList)
         if isinstance(self.original_module, forbidden_classes):
-            cls_name = self.original_module.__class__.__name__
+            cls_name = self.original_module.__class__
+            raise TypeError(f"modules_to_save cannot be applied to modules of type {cls_name}")
+
+        # local import to avoid circular import
+        from peft.tuners.tuners_utils import BaseTunerLayer
+
+        if isinstance(self.original_module, BaseTunerLayer):
+            # e.g. applying modules_to_save to a lora layer makes no sense
+            cls_name = self.original_module.__class__
             raise TypeError(f"modules_to_save cannot be applied to modules of type {cls_name}")
 
     @property
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -49,7 +49,7 @@ from peft import (
     get_peft_model,
 )
 from peft.tuners.tuners_utils import BaseTunerLayer
-from peft.utils import ModulesToSaveWrapper, infer_device
+from peft.utils import infer_device
 
 from .testing_common import PeftCommonTester
 from .testing_utils import get_state_dict, require_torch_gpu
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -1530,29 +1530,6 @@ class PeftCustomModelTester(unittest.TestCase, PeftCommonTester):
         assert torch.allclose(output_custom1, output_custom2)
         assert torch.allclose(output_default, output_custom1)
 
-    @parameterized.expand(["merge_and_unload", "unload"])
-    def test_double_wrapping_merge_and_unload(self, method):
-        # see issue #1485
-        from transformers import AutoModelForTokenClassification
-
-        model = AutoModelForTokenClassification.from_pretrained("hf-internal-testing/tiny-random-RobertaModel")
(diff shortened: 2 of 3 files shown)


src/peft/utils/other.py
```diff
@@ -202,7 +202,15 @@ def check_module(self):
         # ModuleList, even though their forward methods cannot be called
         forbidden_classes = (torch.nn.ModuleDict, torch.nn.ModuleList, torch.nn.ParameterDict, torch.nn.ParameterList)
         if isinstance(self.original_module, forbidden_classes):
-            cls_name = self.original_module.__class__.__name__
+            cls_name = self.original_module.__class__
```
sayakpaul, line 205: Why discard `__name__`?

sayakpaul (approved):
This makes sense. Thank you!

src/peft/utils/other.py
BenjaminBossan: This is what I referred to above:

> the error message inside of
> `ModulesToSaveWrapper.check_module` has been slightly adjusted. Previously, the class name would be used, which can be confusing. E.g. for LoRA, the class name of the linear LoRA layer is just `"Linear"`, which looks the same as `nn.Linear`. Therefore, the full name is now shown.

So the error message used to be:

> 'modules_to_save cannot be applied to modules of type Linear'

which is very confusing, as the user has no idea that "Linear" refers to PEFT's LoRA Linear layer. Now the message would state:

> modules_to_save cannot be applied to modules of type <class 'peft.tuners.lora.layer.Linear'>"

PR #2029, opened 2024-08-21
Update layer.py

DogitoErgoSum:
Improves CPU latency by nearly 33% on my computer when the redundant multiplication by one is not perfomed. The same trick might be applied elsewhere.

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -553,7 +553,7 @@ class Linear(nn.Module, LoraLayer):
                 x = x.to(lora_A.weight.dtype)
 
                 if not self.use_dora[active_adapter]:
-                    result = result + lora_B(lora_A(dropout(x))) * scaling
+                    result += lora_b(lora_a(x)) * scaling if scaling != 1.0 else lora_b(lora_a(x))
                 else:
                     x = dropout(x)
                     result = result + self.lora_magnitude_vector[active_adapter](


DogitoErgoSum:
Also, you could consider saving the merged result of lora_A and lora_B as attributes for each adapter. For example:
`adapter_combined_weight = torch.matmul(lora_b.weight, lora_a.weight)`

During the forward pass, use this combined weight instead of performing the two matrix operations (lora_A(x) -> lora_B(x)). While this approach was less effective on the CPU, I suspect it might perform better on the GPU.

If needed, i can repeat the experiments using the GPU.

BenjaminBossan:
Thanks for this proposal. Could you please share how you benchmarked the time saving from dropping the scalar multiplication?

I tried something very simple and for me the difference was much smaller:

```python
>>> a = nn.Linear(16, 4096)
>>> b = nn.Linear(4096, 16)
>>> x = torch.randn(4, 4096)
>>> %timeit a(b(x))
38.2 µs ± 465 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
>>> %timeit a(b(x)) * 1.0
41.4 µs ± 429 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

So that's only 8% difference. And this operation is only a small part of the whole computation, so I'd think the difference would be hardly noticeable. Finally, most of the time `alpha != rank`, so the optimization would not apply.

> Also, you could consider saving the merged result of lora_A and lora_B as attributes for each adapter. For example:
> `adapter_combined_weight = torch.matmul(lora_b.weight, lora_a.weight)`

This has some disadvantages. First, if we save this, we require extra memory and the matrix is relatively big. Second, this operation bypasses the `forward` method, which could lead to hooks not being called. That can be important, e.g. for distributed training.

DogitoErgoSum:
Initially, I did something very similar to your test. But even when performing the same test, I achieved results similar to my previous ones:
`
a = nn.Linear(16, 4096)
b = nn.Linear(4096, 16)
x = torch.randn(4, 4096)
%timeit a(b(x))
# 51.8 µs ± 551 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
%timeit a(b(x)) * 1.0
# 68.4 µs ± 711 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
`
However, when I ran the same code on a different machine, I got the following results:
`
%timeit a(b(x))
# 221 µs ± 6.33 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
%timeit a(b(x)) * 1.0
# 264 µs ± 20.9 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
`

DogitoErgoSum:
> This has some disadvantages. First, if we save this, we require extra memory and the matrix is relatively big. Second, this operation bypasses the `forward` method, which could lead to hooks not being called. That can be important, e.g. for distributed training.

Understandable, my intention was to speed up inference, not training. During training, merging the matrices A and B defeats the purpose of LoRA. I should have phrased it better.

BenjaminBossan:
Thanks a lot for your further explanations. Even if I'm being skeptical, I want to emphasize that I appreciate the discussion. I would just like to avoid making the already complex code more complex if the benefit is not noticeable on a realistic task. Note that a lot more code would need to be changed to make this consistent everywhere.

> The perfomance difference seems to be machine dependent.

Yes, that was to be expected, but I did not foresee the difference being that big. Could you please also test on your machine how the change would affect a normal training run? My guess would be that in the grand scheme of things it's not noticeable.

> Understandable, my intention was to speed up inference, not training.

Note that for inference, it's already possible to merge the adapter into the base weight, entirely removing the overhead caused by LoRA.

DogitoErgoSum:
> Yes, that was to be expected, but I did not foresee the difference being that big. Could you please also test on your machine how the change would affect a normal training run? My guess would be that in the grand scheme of things it's not noticeable.

Yes, I will compare the code during a training run and inference, then link the code to it.

PR #2033, opened 2024-08-22
FIX: Don't target the classification head when using target_modules="all-linear"

### LINKED ISSUE #2027: Unable to load peft checkpoint

BenjaminBossan:
Fixes #2027

When using a transformers sequence classification model, `target_modules="all-linear"` should not wrap the classification head with LoRA. This is because it is already wrapped with `ModulesToSave`, i.e. it will be fully fine-tuned, which is the generally desired behavior.

Before this bug fix, the classification head would be double-wrapped. With #2028, this now raises an error. With this PR, it is avoided completely. Still, keeping #2028 is good because it helps prevent other situations where double-wrapping might occur due to mis-configuration.

Note that there is no fool-proof method to detect the classification head, we have to rely on transformers convention.

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -31,7 +31,7 @@ from transformers.pytorch_utils import Conv1D
 
 from peft.utils import INCLUDE_LINEAR_LAYERS_SHORTHAND
 from peft.utils.constants import DUMMY_TARGET_MODULES
-from peft.utils.peft_types import PeftType
+from peft.utils.peft_types import PeftType, TaskType
 
 from ..config import PeftConfig
 from ..utils import ModulesToSaveWrapper, _get_submodules
@@ -814,11 +814,23 @@ def _maybe_include_all_linear_layers(peft_config: PeftConfig, model: nn.Module) 
             names = name.rsplit(".", 1)[-1]  # get the base name
             linear_module_names.add(names)
 
-    # ignore the last classification head for text generation models
+    # Try to remove linear layers that should not be targeted as best as possible. We have to rely on convention as
+    # there are no hard rules to detect these modules.
+    module_names_to_exclude = set()
     output_emb = model.get_output_embeddings()
     if output_emb is not None:
+        # ignore the last classification head for text generation models
         last_module_name = [name for name, module in model.named_modules() if module is output_emb][0]
-        linear_module_names -= {last_module_name}
+        module_names_to_exclude.add(last_module_name)
+    elif peft_config.task_type == TaskType.SEQ_CLS:
+        # ignore classifier head for classification models (issue 2027)
+        # there is no fix name for the classifier head, "score" and "classifier" are most common
+        cls_head = getattr(model, "score", None) or getattr(model, "classifier", None)
+        if cls_head is not None:
+            last_module_name = [name for name, module in model.named_modules() if module is cls_head][0]
+            module_names_to_exclude.add(last_module_name)
+
+    linear_module_names -= module_names_to_exclude
     peft_config.target_modules = linear_module_names
     return peft_config
 
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -30,7 +30,7 @@ from transformers import PreTrainedModel
 from transformers.pytorch_utils import Conv1D
 
 from peft.utils import INCLUDE_LINEAR_LAYERS_SHORTHAND
-from peft.utils.constants import DUMMY_TARGET_MODULES
+from peft.utils.constants import DUMMY_TARGET_MODULES, SEQ_CLS_HEAD_NAMES
 from peft.utils.peft_types import PeftType, TaskType
 
 from ..config import PeftConfig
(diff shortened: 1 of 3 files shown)


src/peft/tuners/tuners_utils.py
```diff
@@ -814,11 +814,23 @@ def _maybe_include_all_linear_layers(peft_config: PeftConfig, model: nn.Module)
...
         last_module_name = [name for name, module in model.named_modules() if module is output_emb][0]
-        linear_module_names -= {last_module_name}
+        module_names_to_exclude.add(last_module_name)
+    elif peft_config.task_type == TaskType.SEQ_CLS:
+        # ignore classifier head for classification models (issue 2027)
+        # there is no fix name for the classifier head, "score" and "classifier" are most common
+        cls_head = getattr(model, "score", None) or getattr(model, "classifier", None)
+        if cls_head is not None:
+            last_module_name = [name for name, module in model.named_modules() if module is cls_head][0]
+            module_names_to_exclude.add(last_module_name)
```
sayakpaul, line 831: Perhaps we could define a MAP between the task types and the attributes we know we should exclude and use that? 

```python
EXCLUSION_MAP = {TaskType.SEQ_CLS: ["score", "classifier"], ...}

...

cls_head = None
for exclude_candidate in EXLCUSION_MAP[TaskType.SEQ_CLS]:
    cls_head = getattr(model, exclude_candidate, None)

if cls_head is not None:
    ...
```

The advantage of that is we just have to update the MAP in case we discover more attrbiutes and it should work out nicely. 

WDYT?

tests/test_tuners_utils.py
```diff
@@ -330,6 +336,23 @@ def test_maybe_include_all_linear_layers_diffusion(self):
...
+        # sanity check
+        assert isinstance(model.score, nn.Linear)
+
+        config = LoraConfig(task_type="SEQ_CLS", target_modules="all-linear")
+        model = get_peft_model(model, config)
+        assert isinstance(model.base_model.score, ModulesToSaveWrapper)
+
+        # the bug was that these were lora.Linear instances
+        assert isinstance(model.base_model.score.original_module, nn.Linear)
+        assert isinstance(model.base_model.score.modules_to_save["default"], nn.Linear)
```
sayakpaul, line 354: Sleek!

sayakpaul (commented):
Just a single comment regarding the design. But not a blocker.

src/peft/tuners/tuners_utils.py
BenjaminBossan: Good point about making this easier to extend. I'm not sure if a map is the right approach, because for causal LM (the `if` condition above), we use a different approach based on `get_output_embeddings()`, so the map could not be used consistently for that task. But I will move `["score", "classifier"]` into a constant and use that.

sayakpaul: I still think a map or any adjacent approach is better in the long term but you of course know better here.

sayakpaul (approved):

PR #2034, opened 2024-08-25
Bug: set correct pre-commit-hooks version

ltoniazzi:
### Context

Since the upgrade of pre-commit-hooks version #1965 , my pre-commit fails, and it looks like the version does not exists. 

### Fix

This PR fixes the version to [latest](https://pypi.org/project/pre-commit-hooks/).


BenjaminBossan (approved):
Thanks for fixing this, my bad for introducing this error.

PR #2039, opened 2024-08-26
Add VB-LoRA

leo-yangli:
Hi,

I am one of the authors of VB-LoRA: Extreme Parameter Efficient Fine-Tuning with Vector Banks. You can find our paper [here](https://arxiv.org/abs/2405.15179).

This PR integrates our method, VB-LoRA. The implementation and tests follow VeRA and LoRA. I have run pytest locally, and all the tests have passed.


tests/test_custom_models.py
```diff
@@ -1005,7 +1055,7 @@ def test_disable_adapters(self, test_name, model_id, config_cls, config_kwargs):
         else:
             rtol, atol = 1e-5, 1e-8
         assert not torch.allclose(outputs_before, outputs_after, rtol=rtol, atol=atol)
-        assert torch.allclose(outputs_before, outputs_disabled)
+        assert torch.allclose(outputs_base, outputs_disabled)
```
BenjaminBossan, line 1058: If `init_weights` is implemented for VB-LoRA, this can be reverted, right?

tests/test_custom_models.py
```diff
@@ -1072,8 +1124,8 @@ def test_disable_adapters_with_merging(self, test_name, model_id, config_cls, co
         # unmerged or merged should make no difference
         assert torch.allclose(outputs_after, outputs_unmerged, atol=atol, rtol=rtol)
 
-        # check that disabling adapters gives the same results as before training
-        assert torch.allclose(outputs_before, outputs_disabled, atol=atol, rtol=rtol)
+        # check that disabling adapters gives the same results as base model
+        assert torch.allclose(outputs_base, outputs_disabled, atol=atol, rtol=rtol)
```
BenjaminBossan, line 1128: If `init_weights` is implemented for VB-LoRA, this can be reverted, right?

tests/test_vblora.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class TestVBLoRA:
+    @pytest.fixture
+    def mlp(self):
+        model = MLP()
+        return model
+
+    def test_vblora_parameters(self, mlp):
+        config = VBLoRAConfig(target_modules=["lin0", "lin1", "lin3"], vector_length=2, num_vectors=10)
```
BenjaminBossan, line 54: Let's assign `vector_length` and `num_vectors` to variables and use those below in the assertions, so that it's clear where those numbers come from.

tests/test_vblora.py
```diff
@@ -0,0 +1,123 @@
...
+        assert torch.allclose(output, output_loaded, atol=1e-3, rtol=1e-3)
+
+    def test_save_load_save_topk_only(self, mlp, tmp_path):
+        config = VBLoRAConfig(
+            target_modules=["lin0", "lin1", "lin3"], topk=2, vector_length=2, num_vectors=10, save_topk_weights=True
+        )
+        mlp_vblora = get_peft_model(mlp, config)
+        save_path = tmp_path / "vblora"
+        mlp_vblora.save_pretrained(save_path)
+        print("save_path", save_path)
```
BenjaminBossan, line 102: delete

tests/test_vblora.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def test_save_load_save(self, mlp, tmp_path):
+        config = VBLoRAConfig(target_modules=["lin0", "lin1", "lin3"], vector_length=2, num_vectors=10)
+        mlp_vblora = get_peft_model(mlp, config)
+        save_path = tmp_path / "vblora"
+        mlp_vblora.save_pretrained(save_path)
+
+        assert os.path.exists(save_path / "adapter_config.json")
+
+        mlp_vblora_loaded = PeftModel.from_pretrained(mlp, save_path)
```
BenjaminBossan, line 88: This is not quite clean: the call to `get_peft_model` above will mutate `mlp`, so this `mlp` here is not really the base model. Let's create a fresh instance of `mlp`.

As you cannot reuse the fixture, you could change the `mlp` fixture to a function `get_mlp` and call this. Or even just initialize `MLP()` directly, as the `mlp` fixture does only that.

tests/test_vblora.py
```diff
@@ -0,0 +1,123 @@
...
+        config = VBLoRAConfig(
+            target_modules=["lin0", "lin1", "lin3"], topk=2, vector_length=2, num_vectors=10, save_topk_weights=True
+        )
+        mlp_vblora = get_peft_model(mlp, config)
+        save_path = tmp_path / "vblora"
+        mlp_vblora.save_pretrained(save_path)
+        print("save_path", save_path)
+        assert os.path.exists(save_path / "adapter_config.json")
+
+        mlp_vblora_loaded = PeftModel.from_pretrained(mlp, save_path)
```
BenjaminBossan, line 105: Same comment as above regarding re-using `mlp`.

tests/test_vblora.py
```diff
@@ -0,0 +1,123 @@
...
+        mlp_vblora.save_pretrained(save_path)
+
+        assert os.path.exists(save_path / "adapter_config.json")
+
+        mlp_vblora_loaded = PeftModel.from_pretrained(mlp, save_path)
+
+        input = torch.randn(5, 10)
+        output = mlp_vblora(input)
+        output_loaded = mlp_vblora_loaded(input)
+        assert torch.allclose(output, output_loaded, atol=1e-3, rtol=1e-3)
```
BenjaminBossan, line 93: These tolerances are quite generous, can we use smaller ones?

tests/test_vblora.py
```diff
@@ -0,0 +1,123 @@
...
+        mlp_vblora.save_pretrained(save_path)
+        print("save_path", save_path)
+        assert os.path.exists(save_path / "adapter_config.json")
+
+        mlp_vblora_loaded = PeftModel.from_pretrained(mlp, save_path)
+
+        input = torch.randn(5, 10)
+        output = mlp_vblora(input)
+        output_loaded = mlp_vblora_loaded(input)
+        assert torch.allclose(output, output_loaded, atol=1e-3, rtol=1e-3)
```
BenjaminBossan, line 110: These tolerances are quite generous, can we use smaller ones?

src/peft/tuners/vblora/layer.py
```diff
@@ -0,0 +1,263 @@
...
+            raise ValueError(f"`r` {r} should be a positive integer value")
+        if topk <= 0:
+            raise ValueError(f"`topk` {topk} should be a positive integer value")
+
+        if self.in_features % vector_length != 0:
+            raise ValueError(f"`in_features` {self.in_features} must be divisible by `vector_length` {vector_length}")
+        if self.out_features % vector_length != 0:
+            raise ValueError(
+                f"`out_features` {self.out_features} must be divisible by `vector_length` {vector_length}"
+            )
```
BenjaminBossan, line 83: It would be great to have tests for this. They can go to `test_initialization.py`.

tests/test_vblora.py
```diff
@@ -0,0 +1,123 @@
...
+        X = self.lin1(X)
+        X = self.relu(X)
+        X = self.lin2(X)
+        X = self.relu(X)
+        X = self.lin3(X)
+        X = self.sm(X)
+        return X
+
+
+class TestVBLoRA:
```
BenjaminBossan, line 47: So the vector bank is shared among all VB-LoRA layers, right? Could we add a test here that checks that this is happening correctly?

src/peft/tuners/vblora/model.py
```diff
@@ -0,0 +1,403 @@
...
+                    "fan_in_fan_out is set to False but the target module is `Conv1D`. "
+                    "Setting fan_in_fan_out to True."
+                )
+                kwargs["fan_in_fan_out"] = vblora_config.fan_in_fan_out = True
+        else:
+            raise ValueError(
+                f"Target module {target} is not supported. Currently, only the following modules are supported: "
+                "`torch.nn.Linear`, `transformers.pytorch_utils.Conv1D`."
+            )
+        r = kwargs.pop("r")
```
BenjaminBossan, line 232: Why is this necessary?

src/peft/utils/save_and_load.py
```diff
@@ -184,6 +184,27 @@ def renamed_dora_weights(k):
...
+    elif config.peft_type == PeftType.VBLORA:
+        to_return = {}
+        if config.num_vectors < 2**8:
+            indices_dtype = torch.uint8
+        elif config.num_vectors < 2**15:
+            indices_dtype = torch.int16
+        elif config.num_vectors < 2**31:
+            indices_dtype = torch.int32
+        else:
+            indices_dtype = torch.int64
```
BenjaminBossan, line 196: How much of a difference does this really make? Could this be omitted to make the code simpler?

src/peft/tuners/vblora/layer.py
```diff
@@ -0,0 +1,263 @@
...
+            Linear._get_low_rank_matrix(vblora_logits_B, vblora_vector_bank, self.topk[adapter])
+            .transpose(1, 2)
+            .reshape(-1, vblora_logits_B.shape[1])
+        )
+
+        output_tensor = transpose(B @ A, self.fan_in_fan_out)
+        return output_tensor
+
+    def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
+        previous_dtype = x.dtype
```
BenjaminBossan, line 231: I wonder if we can make a sanity check here that users are not trying to train a model that was loaded with `save_topk_weights`. E.g. we could check if `self.training` and if there is -inf in the logits, WDYT? If this check is expensive, it could be done only during the first iteration.

BenjaminBossan (changes requested):
Thanks a lot for this PR that adds VB-LoRA to PEFT. Your implementation already looks quite advanced, is following the existing examples, and includes the required testing, nice :tada: 

In my first review, I focused on the actual VB-LoRA implementation, so I haven't checked the examples and docs yet. As you can see, I added a bunch of comments but those should not be big issues. Please check if they make sense.

Regarding the paper, I also found two typos, in case you want to fix them: "virtrual" and "backpropgatation".

src/peft/tuners/vblora/layer.py
```diff
@@ -0,0 +1,263 @@
...
+            vblora_dropout_layer = nn.Identity()
+
+        self.vblora_dropout.update(nn.ModuleDict({adapter_name: vblora_dropout_layer}))
+        self.vblora_logits_A[adapter_name] = nn.Parameter(
+            torch.zeros(r, self.in_features // vector_length, num_vectors), requires_grad=True
+        )
+        self.vblora_logits_B[adapter_name] = nn.Parameter(
+            torch.zeros(self.out_features // vector_length, r, num_vectors), requires_grad=True
+        )
+        self.vblora_vector_bank = vblora_vector_bank
```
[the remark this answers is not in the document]
leo-yangli: The `vblora_vector_bank` is initialized and stored in the `VBLoRAModel` class because only one instance of the vector bank is needed. Line 99 gets a reference to this vector bank, so it's important to keep. In the `__init__` function, Line 40 acts like a placeholder and can be removed. WDYT?

src/peft/utils/save_and_load.py
```diff
@@ -338,7 +360,30 @@ def set_peft_model_state_dict(
             PeftType.VERA: "vera_lambda_",
             PeftType.FOURIERFT: "fourierft_",
             PeftType.HRA: "hra_",
+            PeftType.VBLORA: "vblora_",
         }[config.peft_type]
+        if config.peft_type == PeftType.VBLORA and config.save_topk_weights:
+            b, _ = model.vblora_vector_bank[adapter_name].shape
```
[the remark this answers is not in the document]
leo-yangli: That's right! I'll add a comment.

tests/test_custom_models.py
```diff
@@ -789,6 +829,8 @@ def test_merge_layers(self, test_name, model_id, config_cls, config_kwargs):
             config_kwargs["init_ia3_weights"] = False
         elif issubclass(config_cls, LNTuningConfig):
             pass
+        elif issubclass(config_cls, VBLoRAConfig):
```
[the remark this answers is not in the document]
leo-yangli: We just want to skip the last else statement `config_kwargs["init_weights"] = False` here but still proceed with the test.

tests/test_custom_models.py
```diff
@@ -888,7 +933,11 @@ def test_only_params_are_updated(self, test_name, model_id, config_cls, config_k
 
         model.train()
         # this high learning rate was found through testing to be necessary to avoid flakiness
-        lr = 100.0 if config_kwargs.get("use_dora") and model_id == "EmbConv1D" else 0.5
+        lr = (
+            100.0
+            if config_kwargs.get("use_dora") and model_id == "EmbConv1D" or issubclass(config_cls, VBLoRAConfig)
```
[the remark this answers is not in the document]
leo-yangli: Agree. I will modify the code.

tests/test_vblora.py
leo-yangli: will do.

tests/test_vblora.py
leo-yangli: Agree. will use smaller values.

tests/test_vblora.py
leo-yangli: Agree. will use smaller values.

src/peft/tuners/vblora/layer.py
leo-yangli: Agree. will add tests for this.

tests/test_vblora.py
leo-yangli: Right! This feature is already tested in line 71. Do you think I need to add a separate test for it?

src/peft/tuners/vblora/layer.py
```diff
@@ -0,0 +1,263 @@
...
+            warnings.warn("Already unmerged. Nothing to do.")
+            return
+
+        while len(self.merged_adapters) > 0:
+            active_adapter = self.merged_adapters.pop()
+            if active_adapter in self.vblora_logits_A.keys():
+                self.get_base_layer().weight.data -= self.get_delta_weight(active_adapter)
+
+    @staticmethod
+    def _get_low_rank_matrix(logits, vblora_vector_bank, topk):
```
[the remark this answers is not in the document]
leo-yangli: It doesn't have to be a `staticmethod`. I will modify it.

src/peft/tuners/vblora/config.py
```diff
@@ -0,0 +1,139 @@
...
+            "This only works when target_modules is a list of str."
+        },
+    )
+
+    def __post_init__(self):
+        self.peft_type = PeftType.VBLORA
+        self.target_modules = (
+            set(self.target_modules) if isinstance(self.target_modules, list) else self.target_modules
+        )
+        if self.save_topk_weights:
```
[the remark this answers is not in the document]
leo-yangli: Agree. I will delete it.

src/peft/tuners/vblora/config.py
```diff
@@ -0,0 +1,139 @@
...
+        },
+    )
+    init_logits_std: float = field(
+        default=0.1,
+        metadata={
+            "help": (
+                "The logits are initialized with a normal distribution with a standard deviation of init_logits_std."
+            ),
+        },
+    )
```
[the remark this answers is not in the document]
leo-yangli: Agree. I will add some comments here.

src/peft/tuners/vblora/config.py
```diff
@@ -0,0 +1,139 @@
...
+        default=None,
+        metadata={
+            "help": (
+                "List of module names or regex expression of the module names to replace with VBLoRA."
+                "For example, ['q', 'v'] or '.*decoder.*(SelfAttention|EncDecAttention).*(q|v)$'. "
+                "Only linear layers are supported."
+            )
+        },
+    )
+    save_topk_weights: bool = field(
```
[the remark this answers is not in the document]
leo-yangli: Right. I will modify it.

src/peft/tuners/vblora/config.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+@dataclass
+class VBLoRAConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`VBLoRAConfig`].
+
+    Paper: https://arxiv.org/abs/.
+
+    Args:
```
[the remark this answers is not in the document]
leo-yangli: Agree. I will add it.

src/peft/tuners/vblora/model.py
leo-yangli: I will delete this line.

src/peft/tuners/vblora/config.py
```diff
@@ -0,0 +1,139 @@
...
+        fan_in_fan_out (`bool`):
+            Set this to True if the layer to replace stores weight like (fan_in, fan_out). For example, gpt-2 uses
+            `Conv1D` which stores weights like (fan_in, fan_out) and hence this should be set to `True`.
+        bias (`str`):
+            Bias type for Vera. Can be 'none', 'all' or 'vera_only'. If 'all' or 'vera_only', the corresponding biases
+            will be updated during training. Be aware that this means that, even when disabling the adapters, the model
+            will not produce the same output as the base model would have without adaptation.
+        modules_to_save (`List[str]`):
+            List of modules apart from Vera layers to be set as trainable and saved in the final checkpoint.
+        init_weights (`bool`):
```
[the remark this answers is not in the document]
leo-yangli: In our approach, we shouldn't initialize the additive adaptor to zero because if the vector bank is initialized to zero, both matrices A and B will be zero, resulting in a zero gradient. As a result, we can't align with other methods regarding `init_weights=True`. Do you have any suggestions in this case?

src/peft/tuners/vblora/layer.py
leo-yangli: I added some lines of code to check if both `self.training` and `save_topk_weights` are true.

src/peft/utils/save_and_load.py
leo-yangli: This would be useful when aiming for "extreme" parameter efficiency and aligns with the claims made in the paper.

tests/test_custom_models.py
leo-yangli: Yes, but we are having difficulty implementing `init_weights` in the same way as LoRA.

tests/test_custom_models.py
leo-yangli: Yes, but we are having difficulty implementing `init_weights` in the same way as LoRA.

tests/test_custom_models.py
```diff
@@ -823,6 +865,9 @@ def test_safe_merge(self, test_name, model_id, config_cls, config_kwargs):
         elif issubclass(config_cls, LNTuningConfig):
             # LNTuning do not take init_weights
             pass
+        elif issubclass(config_cls, VBLoRAConfig):
+            # VBLoRA do not take init_weights
```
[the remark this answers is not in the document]
leo-yangli: The test proceeds. VB-LoRA doesn't need to set `config_kwargs["init_weights"] = False`.

src/peft/tuners/vblora/layer.py
```diff
@@ -0,0 +1,263 @@
...
+            result = self.base_layer(x, *args, **kwargs)
+        elif self.merged:
+            result = self.base_layer(x, *args, **kwargs)
+        else:
+            result = self.base_layer(x, *args, **kwargs)
+            for active_adapter in self.active_adapters:
+                if active_adapter not in self.vblora_logits_A.keys():
+                    continue
+
+                vblora_logits_A = self.vblora_logits_A[active_adapter]
```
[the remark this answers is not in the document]
leo-yangli: will do.

src/peft/tuners/vblora/layer.py
```diff
@@ -0,0 +1,263 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan, line 1: Let's update the dates in all new files to 2024.

src/peft/tuners/vblora/layer.py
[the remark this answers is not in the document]
BenjaminBossan: Sounds good.

src/peft/utils/save_and_load.py
BenjaminBossan: Okay, is there some measurement how this impacts size? Maybe it's in the paper and I missed it.

tests/test_vblora.py
BenjaminBossan: I see, thanks.

BenjaminBossan (commented):
Thanks for your replies.

Regarding the discussion about `init_weights`, as it came up multiple times, I'll reply here:

From my understanding, the problem is that A and B matrices share the same vector bank. Therefore, if the vector bank is initialized to zeros, both A and B are zeros and therefore there is no gradient. This is unlike LoRA, where only B is zeros.
Also, it's not possible to set the logits for B to zero, as a softmax operation is performed, thus the weights will always sum up to 1.

I have three suggestions:

1. For those cases where we want VB-LoRA to be a no-op, we just manually call `nn.init.ones_(model.base_model.vblora_vector_bank["default"])`. The `init_weights` argument should be removed.
2. We add the option to pass `init_weights=False` (with `True` being default) and in that case initialize the vector bank to zeros. This model is untrainable, so we give a warning that this is option only exists for testing purposes.
3. Add a new parameter that scales the contribution from B, which is a learnable parameter that starts at 0. This way, the vector bank is randomly initialized but VB-LoRA still starts as a no-op. Of course, this would be a deviation from the paper.

I also have a general comment about the docs: Let's make sure to mention that there is an option for VB-LoRA to reduce the size of the file when saving, but that the loaded adapter can only be used for inference after that. Ideally, you could give a numerical example (e.g. for Llama3 8b with parameter ..., the size is reduced from X to Y).

src/peft/utils/save_and_load.py
```diff
@@ -338,7 +360,30 @@ def set_peft_model_state_dict(
...
+            for k in keys:
+                if "_topk_indices" in k:
+                    v = state_dict[k].to(torch.long)
+                    new_key = k.replace("_topk_indices", "")
+                    if k.replace("_topk_indices", "_topk_weights") in state_dict:
+                        topk_weights = state_dict[k.replace("_topk_indices", "_topk_weights")]
+                        topk_weights = torch.cat([topk_weights, 1 - topk_weights.sum(-1, keepdim=True)], dim=-1)
+                        topk_logits = torch.log(topk_weights)
+                    else:
+                        raise ValueError("TopK weights were missing!")
```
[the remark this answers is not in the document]
BenjaminBossan: Yeah, let's not assume that the users manually edit the save files. If they do, a lot of things can break, we cannot check for all of them.

src/peft/utils/save_and_load.py
leo-yangli: Sure, let's take the Llama2 7B model as an example.

We use a vector bank of 2,048 vectors, each with a length of 256. The rank is set to 4, and K is set to 2, applied to all linear layers. 2,048 vectors can be indexed with an int16.

1. Using int16: The float32-equivalent number of parameters is 680,448, i.e., 2.59 MB.
2. Using int32: The float32-equivalent number of parameters is 758,258, i.e., 2.89 MB. (+11.6% compare to int16)
3. Using int64 (default dtype for index): The float32-equivalent number of parameters is 914,688, i.e., 3.49 MB. (+34.7% compare to int16).

In fact, I believe int16 is sufficient for most cases since it can support up to 32,768 vectors (assuming we don't use negative numbers for indexing). However, I considered all possible cases to ensure robustness. I'd like to know your opinion.

src/peft/tuners/vblora/layer.py
leo-yangli: I added a sanity check based on your feedback.

src/peft/tuners/vblora/config.py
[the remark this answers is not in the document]
leo-yangli: Updated.

src/peft/tuners/vblora/config.py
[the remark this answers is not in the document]
leo-yangli: Updated.

src/peft/tuners/vblora/config.py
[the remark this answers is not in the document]
leo-yangli: Updated.

src/peft/tuners/vblora/config.py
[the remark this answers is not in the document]
leo-yangli: Deleted.

src/peft/tuners/vblora/config.py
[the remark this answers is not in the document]
leo-yangli: I followed your first suggestion in the comment: 
1. I removed `init_weights`.
2. In tests, I set the vector bank to zero to make VB-LoRA as an identity operation. And initialized the vector bank if training is involved in the test.

src/peft/tuners/vblora/layer.py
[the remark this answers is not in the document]
leo-yangli: Updated.

src/peft/tuners/vblora/layer.py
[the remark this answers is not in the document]
leo-yangli: Updated.

src/peft/tuners/vblora/layer.py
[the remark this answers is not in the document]
leo-yangli: Updated.

src/peft/utils/save_and_load.py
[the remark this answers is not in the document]
leo-yangli: Updated.

src/peft/utils/save_and_load.py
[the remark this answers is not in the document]
leo-yangli: I removed the ValueError line.

tests/test_custom_models.py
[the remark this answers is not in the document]
leo-yangli: Updated.

tests/test_custom_models.py
leo-yangli: I followed your first suggestion in the comment: 
1. I removed `init_weights`.
2. In tests, I set the vector bank to zero to make VB-LoRA as an identity operation. And initialized the vector bank if training is involved in the test.

I recovered this line back.

tests/test_custom_models.py
leo-yangli: Updated.

tests/test_vblora.py
leo-yangli: Updated.

tests/test_vblora.py
leo-yangli: Deleted.

tests/test_vblora.py
leo-yangli: Updated.

tests/test_vblora.py
leo-yangli: Updated.

tests/test_vblora.py
leo-yangli: Updated.

tests/test_vblora.py
leo-yangli: Updated.

src/peft/tuners/vblora/layer.py
leo-yangli: Added a new test.

src/peft/tuners/vblora/model.py
leo-yangli: Updated.

leo-yangli:
Thank you for your valuable feedback! I have updated the code accordingly. 

I also updated the documentation in `docs/source/package_reference/vblora.md`. Please let me know if there is anything else I can improve.

BenjaminBossan:
Thanks for the updates @leo-yangli. I haven't reviewed yet but saw that the CI fails because of the following test:

`test_parameters_after_loading_model_110_Vanilla_MLP_4_VBLoRA`

Could you please check?

leo-yangli:
Thank you for your feedback. I just fixed that failed test.

leo-yangli:
It's ready to be reviewed.

examples/sequence_classification/VBLoRA.ipynb
```diff
@@ -0,0 +1,775 @@
...
+   "source": [
+    "# Using VB-LoRA for sequence classification"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ddfc0610-55f6-4343-a950-125ccf0f45ac",
+   "metadata": {},
+   "source": [
+    "In this example, we fine-tune Roberta on a sequence classification task using VB-LoRA."
```
BenjaminBossan, line 16: Let's add a sentence that this notebook is based on `examples/sequence_classficiation/VeRA.ipynb` and users can compare the results.

examples/sequence_classification/VBLoRA.ipynb
```diff
@@ -0,0 +1,775 @@
...
+   "outputs": [],
+   "source": [
+    "peft_config = VBLoRAConfig(\n",
+    "    task_type=\"SEQ_CLS\", \n",
+    "    r=rank,\n",
+    "    topk=2,\n",
+    "    target_modules=['key', 'value', 'query', 'output.dense', 'intermediate.dense'],\n",
+    "    num_vectors=num_vectors,\n",
+    "    vector_length=vector_length,\n",
+    "    save_only_topk_weights=True,\n",
```
BenjaminBossan, line 102: Let's add a comment here that to explain this argument. Using this also means that we cannot resume training from checkpoints if training is interrupted for some reason, right? Let's also mention that.

tests/test_initialization.py
```diff
@@ -1170,6 +1171,37 @@ def test_vera_add_second_adapter_with_higher_rank(self):
...
+
+    def test_vblora_with_incompatible_vector_length(self):
+        vector_length = 3
+        config1 = VBLoRAConfig(target_modules=["lin0"], vector_length=vector_length)
+        model = self.get_model()
+        msg = f"`in_features` {model.lin0.in_features} must be divisible by `vector_length` {vector_length}"
+        with pytest.raises(ValueError, match=msg):
+            get_peft_model(model, config1)
+
+        config2 = VBLoRAConfig(target_modules=["lin1"], vector_length=vector_length)
```
BenjaminBossan, line 1199: Let's split this into 2 tests.

src/peft/tuners/vblora/config.py
```diff
@@ -0,0 +1,166 @@
...
+    Args:
+        r (`int`):
+            The rank of incremental matrices.
+        num_vectors (`int`):
+            Number of vectors in the vector bank. Use higher values when the model size increases.
+        vector_length (`int`):
+            The length of the vectors in the vector bank. The length of the vectors should be divisible by the hidden
+            dimension of the model.
+        topk (`int`):
+            K value for topk selection.
```
BenjaminBossan, line 38: It would be great if you could explain a bit more: What is the default, when should users decrease or increase this number? We can make the assumption that users are familiar with LoRA and would like to try VB-LoRA, they should not have to read the paper to understand the hyper-parameters.

src/peft/tuners/vblora/config.py
```diff
@@ -0,0 +1,166 @@
...
+            The names of the modules to apply the adapter to. If this is specified, only the modules with the specified
+            names will be replaced. When passing a string, a regex match will be performed. When passing a list of
+            strings, either an exact match will be performed or it is checked if the name of the module ends with any
+            of the passed strings. If this is specified as 'all-linear', then all linear/Conv1D modules are chosen,
+            excluding the output layer. If this is not specified, modules will be chosen according to the model
+            architecture. If the architecture is not known, an error will be raised -- in this case, you should specify
+            the target modules manually.
+        save_only_topk_weights (`bool`):
+            Whether to only save the topk weights. Models saved in this mode can be used for merging or inference only,
+            not for resuming training.
```
BenjaminBossan, line 49: Let's mention here that the point is to reduce the file size.

src/peft/tuners/vblora/layer.py
```diff
@@ -0,0 +1,249 @@
...
+        top_k_logits, indices = logits.topk(topk, dim=-1)
+        topk_weights = F.softmax(top_k_logits, dim=-1)
+        return (topk_weights.unsqueeze(-1) * vblora_vector_bank[indices]).sum(-2)
+
+    def _get_lora_matrices(self, adapter, cast_to_fp32=False) -> Tuple[torch.Tensor, torch.Tensor]:
+        vblora_logits_A = self.vblora_logits_A[adapter]
+        vblora_logits_B = self.vblora_logits_B[adapter]
+
+        # Check for infinity values when training. If found, training was likely resumed from a `save_only_topk_weights` model.
+        if self.training and vblora_logits_A[0, 0].isinf().any():
```
BenjaminBossan, line 191: Nice idea to check only a single row.

src/peft/tuners/vblora/layer.py
```diff
@@ -0,0 +1,249 @@
...
+            raise RuntimeError(
+                "Found infinity values in logits. Ensure training was not resumed from a `save_only_topk_weights` model."
```
BenjaminBossan, line 193: Let's add: "**VB-LoRA** logits", otherwise this could be confusing.

tests/test_common_gpu.py
```diff
@@ -1572,6 +1573,28 @@ def test_vera_add_new_adapter_does_not_change_device(self, mlp):
         assert model.lin0.vera_A.other.device.type == "cuda"
         assert model.lin0.vera_lambda_d.other.device.type == "cuda"
 
+    def test_vblora_add_new_adapter_does_not_change_device(self, mlp):
+        # same as first test, but using VBLoRA
+        config = VBLoRAConfig(target_modules=["lin0"], vector_length=2)
+        model = get_peft_model(mlp, config)
+        model = model.cuda()
+        model.lin0.vblora_logits_A.cpu()
```
BenjaminBossan, line 1581: Let's also check `vblora_logits_B`.

BenjaminBossan (changes requested):
Thanks a lot for the updates. This is almost good to be merged, I only saw a few tiny issues left. Please take a look.

BenjaminBossan:
Just something that came to my mind out of curiosity: Did you try experiments with multiple VB-LoRA adapters (for different tasks) that share the same vector bank? So e.g. train the logits and vector bank on task 1, then freeze the vector bank. Next, create a new VB-LoRA for task 2 that uses the frozen vector bank from task 1 and only trains logits? Probably this won't work well enough for task 2 but it would be really cool if it did to save even more parameters.

tests/test_common_gpu.py
leo-yangli: Sure. Added some lines to check `vblora_logits_B`.

src/peft/tuners/vblora/layer.py
leo-yangli: Added.

src/peft/tuners/vblora/layer.py
leo-yangli: Thank you!

src/peft/tuners/vblora/config.py
leo-yangli: Added.

src/peft/tuners/vblora/config.py
leo-yangli: Added some explanations.

tests/test_initialization.py
leo-yangli: Updated.

examples/sequence_classification/VBLoRA.ipynb
leo-yangli: Added a comment.

examples/sequence_classification/VBLoRA.ipynb
leo-yangli: Added.

leo-yangli:
Hi @BenjaminBossan , 

That's a great question! Training in an MTL setting is currently an area of ongoing research for us, and we have tried the exact same idea. Preliminary results indicate that when the two tasks are similar (for example, RTE and MRPC from the GLUE benchmark), the second task can still achieve decent performance with a frozen vector bank, although there is a noticeable drop in performance (e.g., 2%). This performance gap tends to decrease when we alternate training between tasks. However, a more robust training strategy for an MTL setting still needs to be explored. I hope this answered your question.

I updated the code based on your review. Moreover, I added a new feature (count VB-LoRA savable parameters) with tests to the code, and it is ready for review.

Thanks!

leo-yangli:
Thanks for pointing out the mistake. I have fixed it.

I've also added a new test, `test_save_with_topk_weights`, to test the weights saved in `save_only_topk_weights=True` mode are as expected.

leo-yangli:
Thank you @BenjaminBossan! The code quality has improved a lot thanks to your insightful review.

PR #2042, opened 2024-08-27
Fixing scalings logging test

EricLBuehler:


tests/test_xlora.py
```diff
@@ -144,7 +144,7 @@ def test_scalings_logging_methods(self, tokenizer, model):
 
         model.disable_scalings_logging()
 
-        inputs = tokenizer.encode("Python is a", add_special_tokens=False, return_tensors="pt")
+        inputs = tokenizer.encode("Python is a", add_special_tokens=False, return_tensors="pt", clean_up_tokenization_spaces=True)
```
EricLBuehler, line 147: In transformers 4.45, the default is False. Perhaps making it `True` fixes it, as the FutureWarning in [the log](https://github.com/huggingface/peft/actions/runs/10557574065/job/29245435767) indicates?

EricLBuehler:
@BenjaminBossan this should be ready for review!

tests/test_xlora.py
```diff
@@ -128,6 +129,8 @@ def test_functional(self, tokenizer, model):
         )
         assert torch.isfinite(outputs[: inputs.shape[1] :]).all()
 
+    # TODO: remove the skip when 4.45 is released!
+    @pytest.mark.skipif(int(transformers.__version__.split(".")[1]) < 45, reason="Requires transformers >= 4.45")
```
BenjaminBossan, line 133: Let's do the version check similar to how we do it here:

https://github.com/huggingface/peft/blob/c9f7240afc616d72ea62fc9fba7bb53e858fbde1/src/peft/peft_model.py#L1670

This should be more robust. It requires the import: `import packaging.version`.

BenjaminBossan (approved):
Thanks for fixing the test. I checked locally that it's currently skipped and that it would pass with the latest transformers version.

PR #2045, opened 2024-08-30
ENH: Faster adapter loading if there are a lot of target modules

BenjaminBossan:
This is an optimization to reduce the number of entries in the `target_modules` set. The reason is that in some circumstances, `target_modules` can contain hundreds of entries. Since each target module is checked against each module of the net (which can be thousands), this can become quite expensive when many adapters are being added. Often, the `target_modules` can be condensed in such a case, which speeds up the process.

A context in which this can happen is when diffusers loads non-PEFT LoRAs. As there is no meta info on target_modules in that case, they are just inferred by listing all keys from the state_dict, which can be quite a lot. See: https://github.com/huggingface/diffusers/issues/9297

As there is a small chance for undiscovered bugs, we apply this optimization only if the list of `target_modules` is sufficiently big. Therefore, for normal PEFT users, this should not have any effect.

Example:

```python
>>> from peft.tuners.tuners_utils import _find_minimal_target_modules
>>> target_modules = [f"model.decoder.layers.{i}.self_attn.q_proj" for i in range(100)]
>>> target_modules += [f"model.decoder.layers.{i}.self_attn.v_proj" for i in range(100)]
>>> other_module_names = [f"model.encoder.layers.{i}.self_attn.k_proj" for i in range(100)]
>>> _find_minimal_target_modules(target_modules, other_module_names)
{"q_proj", "v_proj"}
```

As shown in https://github.com/huggingface/diffusers/issues/9297, the speed improvements for loading many diffusers LoRAs can be substantial. When loading 30 adapters, the time would go up from 0.6 sec per adapter to 3 sec per adapter. With this fix, the time goes up from 0.6 sec per adapter to 1 sec per adapter.

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -433,6 +433,21 @@ class BaseTuner(nn.Module, ABC):
         # update peft_config.target_modules if required
         peft_config = _maybe_include_all_linear_layers(peft_config, model)
 
+        # This is an optimization to reduce the number of entries in the target_modules list. The reason is that in some
+        # circumstances, target_modules can contain hundreds of entries. Since each target module is checked against
+        # each module of the net (which can be thousands), this can become quite expensive when many adapters are being
+        # added. Often, the target_modules can be condensed in such a case, which speeds up the process.
+        # A context in which this can happen is when diffusers loads non-PEFT LoRAs. As there is no meta info on
+        # target_modules in that case, they are just inferred by listing all keys from the state_dict, which can be
+        # quite a lot. See: https://github.com/huggingface/diffusers/issues/9297
+        # As there is a small chance for undiscovered bugs, we apply this optimization only if the list of
+        # target_modules is sufficiently big.
+        if isinstance(peft_config.target_modules, (list, set)) and len(peft_config.target_modules) >= 20:
+            names_not_match = [n for n in key_list if n not in peft_config.target_modules]
+            new_target_modules = find_minimal_target_modules(peft_config.target_modules, names_not_match)
+            if len(new_target_modules) < len(peft_config.target_modules):
+                peft_config.target_modules = new_target_modules
+
         for key in key_list:
             # Check for modules_to_save in case
             if _check_for_modules_to_save and any(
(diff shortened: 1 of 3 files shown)


src/peft/tuners/tuners_utils.py
```diff
@@ -433,6 +433,21 @@ def inject_adapter(self, model: nn.Module, adapter_name: str, autocast_adapter_d
...
+        # This is an optimization to reduce the number of entries in the target_modules list. The reason is that in some
+        # circumstances, target_modules can contain hundreds of entries. Since each target module is checked against
+        # each module of the net (which can be thousands), this can become quite expensive when many adapters are being
+        # added. Often, the target_modules can be condensed in such a case, which speeds up the process.
+        # A context in which this can happen is when diffusers loads non-PEFT LoRAs. As there is no meta info on
+        # target_modules in that case, they are just inferred by listing all keys from the state_dict, which can be
+        # quite a lot. See: https://github.com/huggingface/diffusers/issues/9297
+        # As there is a small chance for undiscovered bugs, we apply this optimization only if the list of
+        # target_modules is sufficiently big.
+        if isinstance(peft_config.target_modules, (list, set)) and len(peft_config.target_modules) >= 20:
```
sayakpaul, line 445: `20` could be assigned to a variable.

src/peft/tuners/tuners_utils.py
```diff
@@ -433,6 +433,21 @@ def inject_adapter(self, model: nn.Module, adapter_name: str, autocast_adapter_d
...
+            names_not_match = [n for n in key_list if n not in peft_config.target_modules]
+            new_target_modules = find_minimal_target_modules(peft_config.target_modules, names_not_match)
```
sayakpaul, line 447: Maybe we keep `find_minimal_target_modules()` as a pseudo-private method to denote the experimental nature of this feature?

src/peft/tuners/tuners_utils.py
```diff
@@ -781,6 +796,86 @@ def _move_adapter_to_device_of_base_layer(self, adapter_name: str, device: Optio
                 adapter_layer[adapter_name] = adapter_layer[adapter_name].to(device)
 
 
+def find_minimal_target_modules(
+    target_modules: list[str] | set[str], other_module_names: list[str] | set[str]
```
sayakpaul, line 800: I would prefer this (`list[str] | set[str]`) to `Union[List[str], Set[str]]` more. Nice.

sayakpaul (approved):
Excellent stuff! Would be nice to also update the original description of this PR to hint at the speedup achievable with this change.

src/peft/tuners/tuners_utils.py
sayakpaul: Yeah same here.

PR #2050, opened 2024-09-05
avoid saving boft_P in adapter model

sywangyi:


sywangyi:
you could use https://github.com/huggingface/peft/blob/main/examples/boft_dreambooth/train_dreambooth.sh to reproduce, wo the PR. the unet adapter size is 485995200B, with the PR, the adapter size drop to 54387152 

training params from the output is
trainable params: 13,611,904 || all params: 879,497,668 || trainable%: 1.5477

training precision is float32

sywangyi:
@BenjaminBossan please help review

BenjaminBossan (changes requested):
Great catch, thanks for providing a fix. Indeed, I think that the `boft_P` buffer needs not be saved. Pinging @Zeju1997 to kindly ask to check this as well.

Let's also add a unit test for this. I tested locally using this:

```python
import torch
from peft import BOFTConfig, PeftModel, get_peft_model


class TestBoft:
    def test_boft_state_dict(self, tmp_path):
        # see #2050
        # ensure that the boft_P buffer is not stored in the checkpoint file and is not necessary to load the model
        # correctly
        torch.manual_seed(0)

        inputs = torch.arange(10).view(-1, 1).to(0)
        model = AutoModelForCausalLM.from_pretrained(model_id).to(0)
        model.eval();
        output_base = model(inputs).logits

        config = BOFTConfig(init_weights=False)
        model.eval();
        output_peft = model(inputs).logits

        atol, rtol = 1e-9, 1e-9
        # sanity check: loading boft changed the output
        assert not torch.allclose(output_base, output_peft, atol=atol, rtol=rtol)

        del model

        # check that the boft_P buffer is not present
        state_dict = load_file(tmp_path / "adapter_model.safetensors")
        assert not any("boft_P" in key for key in state_dict)

        # sanity check: the model still produces the same output after loading
        model = AutoModelForCausalLM.from_pretrained(model_id).to(0)
        model = PeftModel.from_pretrained(model, tmp_path)
        output_loaded = model(inputs).logits
        assert torch.allclose(output_peft, output_loaded, atol=atol, rtol=rtol)
```

If you agree, just create a new file `tests/test_boft.py` and add it there (with the copyright notice on top).

One thing I was concerned about was whether this would cause trouble loading old checkpoints that still contain the `boft_P` buffer. It should not matter, but just to be sure, I created such a checkpoint and wrote a test to check that this checkpoint can be successfully loaded. This test can be added below the previous one.

```python
    def test_boft_old_checkpoint_including_boft_P(self, tmp_path):
        # see #2050
        # This test exists to ensure that after the boft_P buffer was made non-persistent, old checkpoints can still be
        # loaded successfully.
        torch.manual_seed(0)

        inputs = torch.arange(10).view(-1, 1).to(0)
        model = AutoModelForCausalLM.from_pretrained(model_id).to(0)

        # first create the expected output
        config = BOFTConfig(init_weights=False)
        model.eval();
        output_peft = model(inputs).logits
        del model

        model = AutoModelForCausalLM.from_pretrained(model_id).to(0)
        # checkpoint from before the PR whose state_dict still contains boft_P
        hub_id = "peft-internal-testing/boft-tiny-opt-peft-v0.12"
        model = PeftModel.from_pretrained(model, hub_id)
        output_old = model(inputs).logits

        atol, rtol = 1e-9, 1e-9
        assert torch.allclose(output_peft, output_old, atol=atol, rtol=rtol)
```

tests/test_boft.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
```
BenjaminBossan, line 1: No need for this.

tests/test_boft.py
```diff
@@ -0,0 +1,86 @@
...
+
+# coding=utf-8
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan, line 4:
```suggestion
# Copyright 2024-present the HuggingFace Inc. team.
```

BenjaminBossan (changes requested):
Thanks for adding the tests, I only have some minor comments. Could you please also run `make style` with a recent ruff version? Apart from that, let's wait a bit if Zeju wants to add anything, otherwise I'll merge as is.

BenjaminBossan:
@sywangyi For some reason, the tests fail on MacOS. I don't have a machine to test this on, but looking at the logs, the tensors look identical but still the `torch.allclose` call returns `False`:

```
FAILED tests/test_boft.py::TestBoft::test_boft_old_checkpoint_including_boft_P - assert False
 +  where False = <built-in method allclose of type object at 0x10e63f150>(
tensor([
[[ 0.2096,  0.0000, -0.0209,  ...,  0.1637, -0.1071,  0.0516]],\n\n        
[[ 0.0901,  0.0000,  0.0214,  ...,  0...0301]],\n\n        
[[ 0.0866,  0.0000,  0.0419,  ...,  0.1571, -0.1145, -0.0144]]
],\n       grad_fn=<UnsafeViewBackward0>), tensor([
[[ 0.2096,  0.0000, -0.0209,  ...,  0.1637, -0.1071,  0.0516]],\n\n        
[[ 0.0901,  0.0000,  0.0214,  ...,  0..., -0.0659,  ...,  0.0609, -0.0228,  0.0301]],\n\n        
[[ 0.0866,  0.0000,  0.0419,  ...,  0.1571, -0.1145, -0.0144]]]), atol=1e-09, rtol=1e-09)
 +    where <built-in method allclose of type object at 0x10e63f150> = torch.allclose
```

(I added some line breaks for readability)

Probably it's enough to loosen the tolerance, 1e-9 really isn't necessary.

Other than that, I'd wait until end of week if Zeju replies, otherwise I'll merge even without their approval.

PR #2051, opened 2024-09-05
TST: Fewer inference steps for stable diffusion tests

BenjaminBossan:
Reduce the number of inference steps for stable diffusion tests. These tests are the slowest ones on CI, this should hopefully help.


sayakpaul (approved):
I wonder what would the corgi look like in 3 steps 🤭

PR #2058, opened 2024-09-10
TST Speed up vision model tests

BenjaminBossan:
The HRA vision model test is extremely slow on CI (> 600 sec, ~50% of total time). This change speeds up the test by using a smaller ResNet model to run the tests.

It's still not clear why HRA was so slow specifically -- LoRA is 40x faster -- but that can be fixed separately.

HRA author has been [notified](https://github.com/huggingface/peft/pull/1864#issuecomment-2332035181) 5 days ago but no response so far.

### DIFF
--- a/tests/test_vision_models.py
+++ b/tests/test_vision_models.py
@@ -76,7 +76,7 @@ class TestPastKV:
 
 
 class TestResnet:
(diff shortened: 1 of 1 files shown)


sayakpaul (approved):
Good one!

PR #2059, opened 2024-09-10
TST: Make X-LoRA tests faster

BenjaminBossan:
After some recent optimizations, the X-LoRA tests are now among the slowest ones. Part of that is that the lora adapters are re-created for each test. By changing the fixture scope, they're now only created once. I think this should be safe, as these files are not modified in the tests.

I also enabled `test_scalings_logging_methods` with the latest transformers to ensure that this test also passes.

### DIFF
--- a/tests/test_xlora.py
+++ b/tests/test_xlora.py
@@ -36,15 +36,15 @@ class TestXlora:
     model_id = "facebook/opt-125m"
     num_loras = 4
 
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Locally, X-LoRA tests duration went down from 1m11s to 0m37s.

EricLBuehler (approved):
Looks good!

PR #2061, opened 2024-09-11
Update permissions for githubtoken stale.yml

glegendre01:


BenjaminBossan (approved):
Thanks for the quick fix, :crossed_fingers: that it works.

PR #2068, opened 2024-09-13
FIX: Bug that prevents BOFT from loading multiple adapters

BenjaminBossan:
There was a bug in BOFT that made it impossible in some circumstances to load more than one adapter (creating more than one adapter was possible though). This was because a code path that adjusts `boft_n_butterfly_factor` was only visited when creating a fresh adapter, but not when updating with the 2nd adapter. This was fixed by moving this code path from the BOFT layer's `__init__` method to `update_layer`.

A test for loading multiple adapters was added. Since this was a gap in our test suite, this test will be applied to all appropriate PEFT methods, not only BOFT, but the others methods are all passing without needing further changes.

For good measure, I also added BOFT to the test suite that checks multiple active adapters. These tests would have also passed without the fix in this PR, since these tests do not load multiple adapters but instead create them, which always worked. Still it's better to have these tests as well.

### DIFF
--- a/src/peft/tuners/boft/layer.py
+++ b/src/peft/tuners/boft/layer.py
@@ -264,6 +264,14 @@ class BOFTLayer(BaseTunerLayer):
         """
         Update the linear layer with trainable BOFT weights. Override for other layer types.
         """
+        # Attempt to load the CUDA extension during model initialization
+        if not get_fbd_cuda():
+            self.fbd_cuda_available = False
+            # If the CUDA extension is not available, set the butterfly factor to 1 to speed up the finetuning process
+            boft_n_butterfly_factor = 1
+        else:
+            self.fbd_cuda_available = True
+
         # to be consistent with the paper notation
         boft_n_butterfly_factor = boft_n_butterfly_factor - 1
         if boft_n_butterfly_factor < 0:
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
@Zeju1997 it would be nice if you could give this a review, thanks.

Zeju1997:
@BenjaminBossan Thanks for the effort put into the fix! :D

src/peft/tuners/boft/layer.py
```diff
@@ -264,6 +264,14 @@ def update_layer(
         """
         Update the linear layer with trainable BOFT weights. Override for other layer types.
         """
+        # Attempt to load the CUDA extension during model initialization
+        if not get_fbd_cuda():
+            self.fbd_cuda_available = False
+            # If the CUDA extension is not available, set the butterfly factor to 1 to speed up the finetuning process
+            boft_n_butterfly_factor = 1
```
sayakpaul, line 271: Should we log this as an info as well?

sayakpaul (approved):
Thanks for fixing this!

src/peft/tuners/boft/layer.py
BenjaminBossan: This could be logged, yes. Since this part of the code was not changed in this PR, only moved around, I'll leave that open for a subsequent PR.

Issue #1982, opened 2024-07-31
LoftQ init without scaling


cobbswordsman:
https://github.com/huggingface/peft/blob/f13d860e9f58e908ac85e78ac37528ec8c84eb99/src/peft/tuners/lora/layer.py#L226

I find scaling with pissa and dora，but no scaling with loftq init Lora_A and lora_B.
It's a bug?or no need to do that

BenjaminBossan:
I agree, it looks like it could make sense to consider the scaling factor when applying LoftQ. @fxmeng could you comment on that?

BenjaminBossan:
Another ping @fxmeng.

PR #2073, opened 2024-09-17
Add scaling option to loftq

sparsh2:
Add missing scaling option to loftq like dora


BenjaminBossan:
Thanks for the PR. For others, the context is issue #1982.

Honestly, I'm unsure if adding the scale here is correct or not. @fxmeng it would be great if you could take a look.

sparsh2:
Let me close this PR till we have a clarity on the scaling in #1982

PR #2074, opened 2024-09-18
[tests] skip some tests for XPU devices

faaany:
eetq/hqq/aqlm doesn't support XPU yet. So we want to skip them for now.


BenjaminBossan (approved):
Thanks for the update to the tests. We'll (hopefully) also add torchao and optimum-quanto support in the near future. Should those also be skipped on XPU?

yao-matrix:
> Thanks for the update to the tests. We'll (hopefully) also add torchao and optimum-quanto support in the near future. Should those also be skipped on XPU?

Yes, we can skip as of now. And torchao and optimum-quanto is in our priority enabling list, once they are enabled in xpu, we will submit PR to remove the skipping.

faaany:
If these tests live in a separate file or folder, we could also ignore them using  `pytest --ignore` to save code in codebase.

BenjaminBossan:
I agree that the organization of tests is a bit messy in PEFT, but let's keep them as is for now, I don't think these few extra lines of code are hurting.

Issue #2070, opened 2024-09-16
[Feature] Add Quantization Support for VeRA Method


ZiadHelal:
### Feature request

This pull request aims to introduce support for quantization in the VeRA PEFT method. This enhancement is in response to the issue #1921, which highlighted the need for compatibility between VeRA and quantized models.

### Motivation

The motivation behind this feature request is to extend the functionality of the VeRA method to support models that have been quantized using techniques such as bitsandbytes. Currently, users face issues with dimension mismatches and incorrect tensor shapes when applying VeRA to quantized models. By adding support for quantization, we can improve the usability and flexibility of VeRA, making it a more powerful tool for a broader range of use cases.

### Your contribution

I am willing to submit a pull request with the implementation of this feature. The PR will include:
- Code changes to support quantized weights in the VeRA method.
- (Possible) Tests to validate the correct handling of quantized models.
- Updated documentation and examples demonstrating the new functionality.

Please let me know if there are any additional requirements or considerations for this feature. I look forward to contributing this enhancement to the project and am open to any feedback or suggestions. @BenjaminBossan

BenjaminBossan:
Thanks a lot @ZiadHelal for offering to work on this feature. Please go ahead and submit a PR. You can also create a draft PR that does not yet cover all your points, and I can give early feedback, which is often better than trying to finish everything in one step.

PR #2076, opened 2024-09-18
FEAT: Support quantization for VeRA using bitsandbytes (#2070)

ZiadHelal:
This PR introduces support for 4-bit and 8-bit quantization in the VeRA method, leveraging `bitsandbytes`.

Addresses #2070 

Changes made:
- Created bnb.py for both 8-bit & 4-bit linear layers
- Updated model.py for quantization handling and refactored _find_dim
- Modified init.py for new imports
- Ensured VeRA-specific logic in quantized setting
- Added new documentation


ZiadHelal:
Hey @BenjaminBossan could you take a look at this draft PR? I will remove the separate (quantization) config from config.py in vera folder later on, I just want to know if I'm going in the right direction or not.

BenjaminBossan (commented):
Thanks a lot for working on adding 8-bit quantization to VeRA. Super nice!

I haven't done an in-depth review, only a high level check. Here are some observations:

1. The most important next step IMHO is to add one or a few tests, similar to what we have for LoRA + bnb (see `tests/test_common_gpu.py`). Let's validate quickly that the approach works.
2. I see you copied the `dispatch_bnb_8bit` similar to what we have in LoRA. Note, however, that as of now, LoRA is the only PEFT method that uses that dispatch approach. This is because there are now so many LoRA "backends" that doing otherwise became unwieldy. For VeRA, we're not at that stage yet, so I would remove it. Here is how the code looked like [before the dispatch refactor](https://github.com/huggingface/peft/blob/cbd783b4dfa3a1ce645e419dd1cd03e82cf6eb6c/src/peft/tuners/lora/model.py#L261-L281). I would also be fine with adding the dispatch refactor in this PR too, but then we'd need the full refactor, i.e. including `dispatch_default`.
3. Do you plan on adding 4 bit as well?
4. Regarding what you mentioned, yes, I agree:
>I will remove the separate (quantization) config from config.py in vera folder later on

ZiadHelal:
Hey @BenjaminBossan,

I've finished the 8-bit quantization and now it works with all tests passed, however, with 4-bit it's a bit tricky due to bnb packing of weights implementation. I've added a work-around (which is not correct) in the forward method and it can now train any model but again I think it's not correct, also it fails for the test_vera_bnb_quantization_from_pretrained_safetensors test in 4-bit.

I would very much appreciate it if you could direct me in which way to proceed for the 4-bit implementation.

BenjaminBossan:
Thanks for the updates.

> I've added a work-around (which is not correct) in the forward method and it can now train any model but again I think it's not correct

Could you give me a pointer what lines of code exactly you mean here?

> also it fails for the test_vera_bnb_quantization_from_pretrained_safetensors test in 4-bit.

For me the test fails too, albeit already during initialization, not in the forward pass.

Also pinging @vvvm23 and @dkopi for awareness.

src/peft/tuners/vera/bnb.py
```diff
@@ -0,0 +1,408 @@
...
+                    dropout = self.vera_dropout[active_adapter]
+
+                    requires_conversion = not torch.is_autocast_enabled()
+                    if requires_conversion:
+                        expected_dtype = result.dtype
+                        compute_dtype = lambda_d.dtype
+                        if x.dtype != compute_dtype:
+                            x = x.to(compute_dtype)
+
+                    sliced_A = vera_A.repeat(1, self.in_features)
```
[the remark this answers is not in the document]
BenjaminBossan: The way that the original layer's shape is determined no longer works when bnb layers are used. Specifically, these lines:

https://github.com/huggingface/peft/blob/f4cf170a9c51d822f950cde0a0e1c87dc013403a/src/peft/tuners/vera/model.py#L122-L125

Could you try replacing those lines with these instead:

```python
            if isinstance(module, nn.Linear):
                module_shape = module.in_features, module.out_features
            elif isinstance(module, Conv1D):
                module_shape = module.weight.ds_shape if hasattr(module.weight, "ds_shape") else module.weight.shape
```

This should allow you to use the normal `forward` code for 4 bit, since the `repeat` call, as you mentioned, is incorrect.

ZiadHelal:
Hey @BenjaminBossan, it now works on my side! thanks for your help.

Could you check if it works now with you?

BenjaminBossan:
Thanks @ZiadHelal I can confirm that the tests are now passing on my machine. I think what would be great is if we could take one of the VeRA examples and verify that it works with 4bit and 8bit bnb. Of course, results won't be exactly the same, but we should expect roughly similar outcomes, probably slightly worse. This would be a nice confirmation that the implementation is correct. Is that something you would be willing to tackle?

Apart from that, please run `make style` on your PR, so that our linter is happy and tests can be run.

ZiadHelal:
Hi @BenjaminBossan, sorry for my late reply!

I've run `make style` and the code should now be good for the linter. Regarding the tests, I've added several tests primarily for CausalLM but I would be happy to add further for the audio and seq2seq models or if you have any other tests in mind, please let me know.

BenjaminBossan:
Hmm, somehow the linter is still not happy, do you have the latest ruff version?

Anyway, I checked this out locally and here is the diff that I get:

```diff
modified   src/peft/tuners/vera/bnb.py
@@ -134,9 +134,9 @@ if is_bnb_available():
                 torch.Tensor: The computed delta weight for the VeRA adapter.
 
             Note:
-                This method implements the VeRA-specific weight update. Unlike LoRA, VeRA uses shared
-                projection matrices (vera_A and vera_B) across all layers, along with per-layer
-                trainable parameters (lambda_d and lambda_b).
+                This method implements the VeRA-specific weight update. Unlike LoRA, VeRA uses shared projection
+                matrices (vera_A and vera_B) across all layers, along with per-layer trainable parameters (lambda_d and
+                lambda_b).
             """
             # Retrieve shared projection matrices
             vera_A = self.vera_A[adapter]
@@ -187,9 +187,9 @@ if is_bnb_available():
                 torch.Tensor: Output tensor after applying the VeRA adaptation.
 
             Note:
-                This method implements the VeRA-specific forward pass. It applies the shared
-                projections (vera_A and vera_B) along with the per-layer trainable parameters
-                (lambda_d and lambda_b) to compute the adapter output.
+                This method implements the VeRA-specific forward pass. It applies the shared projections (vera_A and
+                vera_B) along with the per-layer trainable parameters (lambda_d and lambda_b) to compute the adapter
+                output.
             """
             if self.disable_adapters:
                 if self.merged:

```

ZiadHelal:
I've run again the `make style` command now. Hope it works!

BenjaminBossan:
Ouch, a bunch of tests are failing. Could you please investigate? Please LMK if you need help.

ZiadHelal:
I've run the tests that are failing on my machine specifically `test_initialization.py`, `test_feature_extraction_models.py`, `test_decoder_models.py`, and `test_custom_models.py`. It appears that the refactored lines that you suggested for the function `_find_dim` in `model.py` in VeRA are causing these errors. Idk how to proceed next whether are these lines are correct and we need to adjust the tests accordingly or implementing a work around for `_find_dim` to get the dimensions of vera_A & vera_B based on whether quantization is used or not.

ZiadHelal:
This approach is what I have in mind (I haven't implemented it yet but should work ig).

```python
    def _find_dim(self, config) -> tuple[int, int]:
        """
        Finds the largest input and output dimensions across linear layers that have been wrapped with VeRA.

        This will be used for determining the size of the shared vera_A and vera_B matrices.
        """
        model_config = self.get_model_config(self.model)

        peft_config = self._prepare_adapter_config(config, model_config)
        peft_config = _maybe_include_all_linear_layers(peft_config, self.model)

        loaded_in_4bit = getattr(self.model, "is_loaded_in_4bit", False)

        largest_shape = None
        for key, module in self.model.named_modules():
            if not self._check_target_module_exists(peft_config, key):
                continue

            if loaded_in_4bit:
                if isinstance(module, nn.Linear):
                    module_shape = module.in_features, module.out_features
                elif isinstance(module, Conv1D):
                    module_shape = module.weight.ds_shape if hasattr(module.weight, "ds_shape") else module.weight.shape
                else:
                    continue
            else:
                if isinstance(module, (nn.Linear, Conv1D)):
                    module_shape = tuple(module.weight.shape)
                    if isinstance(module, Conv1D):
                        module_shape = module_shape[::-1]
                else:
                    continue

            if largest_shape is None:
                largest_shape = module_shape
                continue

            if module_shape != largest_shape:
                largest_shape = tuple(max(a, b) for a, b in zip(largest_shape, module_shape))

        if largest_shape is None:
            msg = "No layers types compatible with VeRA were found. Please check `peft_config.target_modules`."
            raise ValueError(msg)

        return largest_shape
```

BenjaminBossan:
I took a closer look and the code that I suggested had a simple error, I was returning the shapes in the wrong order. So the correct code should be:

```python
            if isinstance(module, nn.Linear):
                module_shape = module.out_features, module.in_features
            elif isinstance(module, Conv1D):
                module_shape = module.weight.ds_shape if hasattr(module.weight, "ds_shape") else module.weight.shape
                module_shape = module_shape[::-1]
```

As to your suggestion: Yes, possibly there needs to be some special handling for quantized weights. I haven't checked that yet.

ZiadHelal:
I added your line suggestions and it now works for all the tests except (I think) one test `test_decoder_models.py` which is complaining about something not related to VeRA. Maybe it won't fail in the pipeline. Can you run the tests again and see if it relates to VeRA, if so then I will apply my approach of adding the special handling for quantized weights.

BenjaminBossan:
> one test `test_decoder_models.py` which is complaining about something not related to VeRA.

Which one is it?

ZiadHelal:
`test_generate_half_prec` this is the one but with several precisions

ZiadHelal:
@BenjaminBossan, my bad. it should pass this test, sorry for the confusion.

BenjaminBossan:
That would be strange, as VeRA tests are not run:

https://github.com/huggingface/peft/blob/ccc350151f95a9ff95da046bae5671da75eab52f/tests/testing_common.py#L946-L947

Anyway, I'll start the CI, let's see.

BenjaminBossan:
No, I don't think that they're related. This could possibly be caused by the latest transformers release, not sure. I'll investigate tomorrow.

ZiadHelal:
Okay, thanks!

BenjaminBossan:
Small update, it is indeed unrelated, the tests started breaking due to a recent change in transformers. I'm trying to get to the bottom of it.

ZiadHelal:
Ok got it, thanks for the update!

BenjaminBossan:
The fix is now merged. Once you merge with/rebase on main, the tests should pass.

src/peft/tuners/vera/bnb.py
```diff
@@ -0,0 +1,408 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan, line 1: Let's update the date to 2024.

BenjaminBossan (changes requested):
Thanks for the latest updates. I somehow forgot about this PR, sorry for my silence, don't hesitate to ping me if you don't hear back for a few days.

Overall, this looks really good and is ready to be merged except for two small issues. One I have commented on. The other is that the docs need updates to reflect that VeRA now supports bnb quantization:

https://github.com/huggingface/peft/blob/aa3bd8fbf6820bb000e148a289b6b0c48c339774/docs/source/package_reference/vera.md#L28

Besides that, I think we should also extend `quantization.md`. I'm not 100% sure what the best way is to update the document, but maybe the easiest is to add a new section [before the "Next steps" section](https://github.com/huggingface/peft/blob/aa3bd8fbf6820bb000e148a289b6b0c48c339774/docs/source/developer_guides/quantization.md#L189) called "Other supported PEFT methods". There, you could mention that besides, LoRA, bnb also works with VeRA.

BenjaminBossan:
Ah, forgot to mention, I ran a test modified from the `VeRA.ipynb` notebook comparing no quantization vs 8bit vs 4bit. Overall, the results were quite similar, with a bit of degradation for 4bit, but that's expected, so I think the implementation works as expected.

ZiadHelal:
Sorry for late reply. I did the requests you mentioned and added also for the docs the other peft methods that support quantization which weren't added at the first place.

ZiadHelal:
Hey @BenjaminBossan, any update regarding the new documentation changes?

docs/source/package_reference/vera.md
```diff
@@ -22,13 +22,6 @@ When saving the adapter parameters, it's possible to eschew storing the low rank
 
 To handle different shapes of adapted layers, VeRA initializes shared A and B matrices with the largest required size for each dimension. During the forward pass, submatrices A and B for a given layer are sliced out from these shared matrices and used as described in the paper. For example, adapting two linear layers of shapes (100, 20) and (80, 50) will create A and B matrices of shapes (rank, 50) and (100, rank) respectively. Then, to adapt a layer of shape (100, 20), submatrices A and B of shapes (rank, 20) and (100, rank) will be extracted.
 
-VeRA currently has the following constraints:
-
-- Only `nn.Linear` layers are supported.
```
BenjaminBossan, line 27: The constraint about linear layers still holds true, so let's not delete this line.

BenjaminBossan (changes requested):
Thanks for the ping. I was out of office for a few days. This PR is almost good to go, you were just a bit overeager when removing the VeRA constraints in the docs.

ZiadHelal:
Done @BenjaminBossan

BenjaminBossan (approved):
Thanks a lot for adding bitsandbytes quantization to VeRA, very nice addition!

ZiadHelal:
Thanks for your help and effort during this PR!

PR #2078, opened 2024-09-19
ENH: Add default target layers for gemma2 architecture

BenjaminBossan:
Google's gemma 2 models have a slightly different architecture than gemma 1 and thus a different `model_type` attribute. This PR adds default `target_layer`s for gemma 2 that correspond to the default `target_layer`s of gemma 1.

### DIFF
--- a/src/peft/utils/constants.py
+++ b/src/peft/utils/constants.py
@@ -72,7 +72,13 @@ TRANSFORMERS_MODELS_TO_LNTUNING_TARGET_MODULES_MAPPING = {
     "mistral": ["input_layernorm", "post_attention_layernorm", "norm"],
     "phi": ["input_layernorm", "final_layernorm"],
     "gemma": ["input_layernorm", "post_attention_layernorm", "norm"],
-    "gemma2": ["input_layernorm", "norm"],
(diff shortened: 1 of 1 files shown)


src/peft/utils/constants.py
```diff
@@ -72,6 +72,7 @@ def starcoder_model_postprocess_past_key_value(past_key_values):
     "mistral": ["input_layernorm", "post_attention_layernorm", "norm"],
     "phi": ["input_layernorm", "final_layernorm"],
     "gemma": ["input_layernorm", "post_attention_layernorm", "norm"],
+    "gemma2": ["input_layernorm", "norm"],
```
SunMarc, line 75: Is it normal that they are not the same since you said that it was the same layers as gemma1 in the description of the PR.

SunMarc (approved):
LGTM ! Left a comment

src/peft/utils/constants.py
BenjaminBossan: Not sure what I did there, I checked again and gemma2 has in fact more layer norm layers than gemma1, not fewer. I updated the targets accordingly.

Issue #2079, opened 2024-09-19
Support Conv3d layer


jsilter:
### Feature request

Currently, many adapters support Conv1d and Conv2d layers, but not Conv3d. I would like to be able to use these methods on Conv3d layers.

### Motivation

I would like to use PEFT on models which process 3D images (eg video or volumetric images). As it stands, PEFT cannot be applied.

### Your contribution

I have an implementation for IA3 which I am happy to contribute. I believe there are 8 total adapter methods which currently support convolutional layers, and each one would need to be modified individually. I can probably find the time for a few other methods as well if there is interest.

BenjaminBossan:
Adding support for more layers would indeed be a nice addition to PEFT. If you already have an implementation for IA³, please feel free to open a PR. When it comes to other methods, we can add support step by step. Depending on the method, it could be easy or quite hard. The biggest leverage would be if LoRA had support, since that's the most used one and its implementation is quite straightforward.

jsilter:
Great, I just created a PR at #2082. It includes support for IA^3 and LoRA.

BenjaminBossan:
Resolved via #2082.

PR #2082, opened 2024-09-20
Support Conv3d layer in LoRA and IA3

jsilter:
Support Conv3d layer in LoRA and IA3

* Add Conv3d layers for LoRA and IA3
* Add test cases in test_custom_models

Issue GH-2079


tests/test_custom_models.py
```diff
@@ -784,6 +808,51 @@ def forward(self, X):
...
+            X = torch.stack([X] * 3, dim=-1)
+        X = X.float().reshape(-1, 5, 3, 3, 3)
+        X = self.conv3d(X)
+        X = self.relu(X)
+        X = self.flat(X)
+        X = self.lin0(X)
+        X = self.sm(X)
+        return X
+
+class ModelConv3D2(nn.Module):
```
BenjaminBossan, line 832: I think this class is not used anywhere?

BenjaminBossan (commented):
Thanks a lot for this PR. I didn't have time yet to do a thorough review, will do that next week. In the meantime, could you please check my comment and also run `make style`?

tests/test_custom_models.py
jsilter: I added some test cases with that in the latest commit

src/peft/tuners/lora/layer.py
```diff
@@ -1113,6 +1125,16 @@ def __repr__(self) -> str:
         return "lora." + rep
 
 
+class Conv2d(ConvNd):
```
BenjaminBossan, line 1128: Similar argument to what I mentioned above for IA³ and for `self.conv_fn`.

BenjaminBossan (changes requested):
Thanks a lot for this PR to add `Conv3d` support to IA³ and LoRA. Overall, this already looks pretty good. I made some small suggestions regarding the implementations, please check.

Thanks also for adding the tests. I saw that you already made some adjustments with regard to tolerances. However, for me, they don't appear to be sufficient. When I run the tests locally on GPU, I still get these failing tests:

```
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_disable_adapters_with_merging_019_Conv3d_1_LoRA - AssertionError: assert False
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_disable_adapters_with_merging_020_Conv3d_2_LoRA - AssertionError: assert False
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_disable_adapters_with_merging_021_Conv3d2_1_LoRA - AssertionError: assert False
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_disable_adapters_with_merging_023_Conv3d_1_LoRA_with_DoRA - AssertionError: assert False
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_disable_adapters_with_merging_024_Conv3d_2_LoRA_with_DoRA - AssertionError: assert False
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_disable_adapters_with_merging_025_Conv3d2_1_LoRA_with_DoRA - AssertionError: assert False
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_disable_adapters_with_merging_026_Conv3d2_2_LoRA_with_DoRA - AssertionError: assert False
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_merge_layers_020_Conv3d_2_LoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_merge_layers_021_Conv3d2_1_LoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_merge_layers_022_Conv3d2_2_LoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_merge_layers_024_Conv3d_2_LoRA_with_DoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_merge_layers_025_Conv3d2_1_LoRA_with_DoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_merge_layers_026_Conv3d2_2_LoRA_with_DoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_safe_merge_019_Conv3d_1_LoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_safe_merge_020_Conv3d_2_LoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_safe_merge_021_Conv3d2_1_LoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_safe_merge_022_Conv3d2_2_LoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_safe_merge_023_Conv3d_1_LoRA_with_DoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_safe_merge_024_Conv3d_2_LoRA_with_DoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_safe_merge_025_Conv3d2_1_LoRA_with_DoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_safe_merge_026_Conv3d2_2_LoRA_with_DoRA - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_safe_merge_049_Conv3d2_1_IA3 - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_safe_merge_050_Conv3d2_2_IA3 - AssertionError
FAILED tests/test_custom_models.py::PeftCustomModelTester::test_safe_merge_051_Conv3d2_3_IA3 - AssertionError
```

I didn't check all in details, but did a spot check and it was always a matter of tolerances being too strict. Do you observe the same behavior?

src/peft/tuners/ia3/layer.py
```diff
@@ -198,15 +198,15 @@ def __init__(
         IA3Layer.__init__(self, base_layer, is_feedforward=is_feedforward)
         self.fan_in_fan_out = fan_in_fan_out
         self._active_adapter = adapter_name
+        self._kernel_rank = base_layer.weight.dim()
```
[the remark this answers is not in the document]
jsilter: I was thinking that "dim" might be confusing, a "2D" convolution will in general have a 4D weight tensor (batch, channel, height, width) so I was trying to differentiate spatial dimension from tensor dimension. Although since I'm checking the "dim" attribute maybe it is better to be consistent.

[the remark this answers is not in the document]
BenjaminBossan: Personally, I find it less confusing than rank, but I'm open to other suggestions.

src/peft/tuners/lora/layer.py
jsilter: I don't follow; do you mean a sanity check on the self._kernel_dim? I added those in the most recent commits.

src/peft/tuners/lora/model.py
```diff
@@ -345,7 +345,7 @@ def dynamic_dispatch_func(target, adapter_name, lora_config, **kwargs):
             # no module could be matched
             raise ValueError(
                 f"Target module {target} is not supported. Currently, only the following modules are supported: "
-                "`torch.nn.Linear`, `torch.nn.Embedding`, `torch.nn.Conv2d`, `transformers.pytorch_utils.Conv1D`."
+                "`torch.nn.Linear`, `torch.nn.Embedding`, `torch.nn.Conv2d`, `torch.nn.Conv3d`, `transformers.pytorch_utils.Conv1D`."
```
BenjaminBossan, line 348: Let's add a line break to stick with 120 chars.

src/peft/tuners/lora/dora.py
```diff
@@ -133,12 +133,13 @@ def __repr__(self) -> str:
         return "lora.dora." + rep
 
 
-class DoraConv2dLayer(DoraLinearLayer):
+class DoraConvNdLayer(DoraLinearLayer):
```
BenjaminBossan, line 136: I wonder if we should not keep `DoraConv2dLayer` around for backwards compatibility, and do the same changes as with IA³ and LoRA with a base class and 2 subclasses.

tests/test_custom_models.py
BenjaminBossan: Could you explain why we need 2 separate models with `Conv3d`? What does this one cover that the first one does not?

BenjaminBossan (changes requested):
Thanks a lot for the updates. I have a few more comments:

> I don't follow; do you mean a sanity check on the self._kernel_dim? I added those in the most recent commits.

What I mean is that in the LoRA `Conv2d.__init__`, we could add a line: `self.conv_fn = F.conv2d` and analogous for `Conv3d`. Then in the `forward`, we can get rid of `F_conv = F.conv3d if self._convd >= 3 else F.conv2d` and just use `self.conv_fn` instead. Same for DoRA. WDYT?

> I just pushed an update with increased tolerances (https://github.com/huggingface/peft/commit/57e7db9bc34079cdaeacea9d13b90a19ef5961c3), in which all tests pass.

I can confirm that all tests now pass on my machines too.

Could you please also update this line:

https://github.com/huggingface/peft/blob/f4cf170a9c51d822f950cde0a0e1c87dc013403a/src/peft/utils/other.py#L298

This makes it so that `Conv3d` is working also with `modules_to_save`. I think this will also require updating:

https://github.com/huggingface/peft/blob/f4cf170a9c51d822f950cde0a0e1c87dc013403a/tests/test_custom_models.py#L3397

tests/test_custom_models.py
jsilter: I suppose we don't, I was just blindly copying the existing Conv2d. But on closer inspection that one was made specifically for BOFT.

jsilter:
> Thanks a lot for the updates. I have a few more comments:
> ...

I implemented all of these changes. Although in the Dora case I wonder if having those 2 sub-classes just to avoid a single trinary statement is a bit over-engineered, if we're going to do that for LoRA (and IA^3 and maybe more) it probably makes sense to be consistent.

src/peft/tuners/lora/layer.py
```diff
@@ -917,18 +925,26 @@ def update_layer(self, adapter_name, r, lora_alpha, lora_dropout, init_lora_weig
...
-        dora_layer = DoraConv2dLayer(fan_in_fan_out=False)
+        dora_layer_class = self._get_dora_layer_class()
+        dora_layer = dora_layer_class(fan_in_fan_out=False)
         lora_A = self.lora_A[adapter_name].weight
         lora_B = self.lora_B[adapter_name].weight
         scaling = self.scaling[adapter_name]
         dora_layer.update_layer(base_layer=self.get_base_layer(), lora_A=lora_A, lora_B=lora_B, scaling=scaling)
         self.lora_magnitude_vector[adapter_name] = dora_layer
 
+    def _get_dora_layer_class(self) -> Type[_DoraConvNdLayer]:
```
BenjaminBossan, line 944:
```suggestion
    def _get_dora_layer_class(self) -> type[_DoraConvNdLayer]:
```

This should also work.

src/peft/tuners/lora/layer.py
```diff
@@ -867,6 +870,9 @@ def __init__(
         LoraLayer.__init__(self, base_layer)
 
         self._active_adapter = adapter_name
+        self._kernel_dim = base_layer.weight.dim()
+        self._convd = self._kernel_dim - 2
```
BenjaminBossan, line 874: Do we really need a separate attribute for this or can we not just use `_kernel_dim`?

BenjaminBossan (commented):
Thanks for the new changes, not much is missing now. Could you please check my comments and also run `make style` again?

> Although in the Dora case I wonder if having those 2 sub-classes just to avoid a single trinary statement is a bit over-engineered, if we're going to do that for LoRA (and IA^3 and maybe more) it probably makes sense to be consistent.

It's also so that the `DoraConv2dLayer` class is being preserved, just in case users are using it directly for some reason.

src/peft/tuners/lora/layer.py
jsilter: Certainly using _kernel_dim would work; I thought this would make it more obvious what was going on. I just updated the branch.

tests/test_custom_models.py
BenjaminBossan: Then I would suggest to remove that model, as I think it could be confusing to readers of the code and also makes testing ever so slightly slower.

BenjaminBossan (changes requested):
Things look great now, I only have a small request left to simplify the testing.

A more general question: Do you have an example of using `Conv3d`? I'm asking as it would be nice to have a more or less realistic example of using `Conv3d` (but not necessary for this PR).

jsilter:
@BenjaminBossan I don't have a specific realistic example at the moment. I'm working on analyzing CT images (in the vein of https://github.com/reginabarzilaygroup/Sybil) but nothing ready to be published. If you think a toy example would be useful I can add that to my todo list; video classification would be the obvious choice.

BenjaminBossan:
Thanks for the info, I agree something like video classification would be a good example. Would that be something you could take a look at soon or rather do it separately from this PR?

jsilter:
I'm not sure when I'll be able to get to it so I'd prefer to keep it separate if that's alright.

BenjaminBossan (approved):
The PR looks great, thanks so much for adding support for `Conv3d` in LoRA and IA³ with minimal code changes being required.

Issue #1932, opened 2024-07-17
`load_adapter` seems to require the base model to be identical


zyzhang1130:
### System Info

peft:  0.11.1, accelerate: 0.30.1, transformers: 4.41.2, trl: 0.9.4, python: 3.10.14

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

```
model, tokenizer = setup_chat_format(model, tokenizer) #this line adds addition token embedding
model.load_adapter(peft_model_id, adapter_name="default") #peft_model_id points to the same model before its embedding layer was resized
```

### Expected behavior

https://github.com/huggingface/peft/blob/52684952136b3dcc1120834f8af2d8f5aaa1c16a/src/peft/peft_model.py#L1028 as the title suggests. I tried to use `model.load_adapter(peft_model_id, adapter_name="default")` to load adapter from a peft model that has different dimensions for the token embedding, and got the following message:
```
<_IncompatibleKeys, len() = 2>

[...]

PR #2084, opened 2024-09-21
Fix Inconsistent Missing Keys Warning for Adapter Weights in PEFT

yaswanth19:
Refer #1932 for more context.

The main logic is we would have a specific prefix of adapter related keys and if such prefix is present in any of the missing keys then raise a warning else the adapter is loaded succesfully.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -1203,9 +1203,13 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
         missing_keys, unexpected_keys = load_result.missing_keys, load_result.unexpected_keys
         tuner = self.peft_config[adapter_name].peft_type
 
-        # Check for missing keys related to the adapter tuner
         tuner_prefix = PEFT_TYPE_TO_PREFIX_MAPPING.get(tuner, "")
-        adapter_missing_keys = [key for key in missing_keys if tuner_prefix in key]
+        adapter_missing_keys = []
+
+        # Filter missing keys specific to the current adapter and tuner prefix.
+        for key in missing_keys:
+            if tuner_prefix in key and adapter_name in key:
+                adapter_missing_keys.append(key)
 
         load_result = _IncompatibleKeys(missing_keys=adapter_missing_keys, unexpected_keys=unexpected_keys)
 
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -106,7 +106,21 @@ PEFT_TYPE_TO_MODEL_MAPPING = {
     PeftType.VBLORA: VBLoRAModel,
 }
 
-
+PEFT_TYPE_TO_PREFIX_MAPPING = {
+    PeftType.ADALORA: "lora_",
+    PeftType.BOFT: "boft_",
+    PeftType.FOURIERFT: "fourierft_",
+    PeftType.HRA: "hra_",
+    PeftType.IA3: "ia3_",
+    PeftType.LN_TUNING: "ln_tuning_",
+    PeftType.LOHA: "hada_",
+    PeftType.LOKR: "lokr",
+    PeftType.LORA: "lora_",
+    PeftType.OFT: "oft_",
+    PeftType.POLY: "poly_",
+    PeftType.VBLORA: "vblora_",
+    PeftType.VERA: "vera_lambda",
+}
 class PeftModel(PushToHubMixin, torch.nn.Module):
     """
     Base model encompassing various Peft methods.
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -34,7 +34,6 @@ from huggingface_hub import HfFileSystem, ModelCard, ModelCardData, hf_hub_downl
 from safetensors import safe_open
 from safetensors.torch import save_file as safe_save_file
 from torch.nn import BCEWithLogitsLoss, CrossEntropyLoss, MSELoss
-from torch.nn.modules.module import _IncompatibleKeys
 from transformers import PreTrainedModel
 from transformers.modeling_outputs import QuestionAnsweringModelOutput, SequenceClassifierOutput, TokenClassifierOutput
 from transformers.utils import PushToHubMixin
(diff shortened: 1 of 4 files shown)


yaswanth19:
Hi @BenjaminBossan 

I think this logic should cover most of the cases. I haven't added prompt learning techniques here as they don't have a concrete prefix. Please let me know if I am missing any edge cases.

src/peft/peft_model.py
```diff
@@ -105,7 +105,21 @@
     PeftType.VBLORA: VBLoRAModel,
 }
 
-
+PEFT_TYPE_TO_PREFIX_MAPPING = {
```
BenjaminBossan, line 108: We have this exact same mapping in [`save_and_load`](https://github.com/huggingface/peft/blob/f4cf170a9c51d822f950cde0a0e1c87dc013403a/src/peft/utils/save_and_load.py#L360-L374). I would suggest that we factor this out into a constant in `utils/constants.py` and re-use it in both locations.

src/peft/peft_model.py
```diff
@@ -105,7 +105,21 @@
...
+    PeftType.HRA: "hra_",
+    PeftType.IA3: "ia3_",
+    PeftType.LN_TUNING: "ln_tuning_",
+    PeftType.LOHA: "hada_",
+    PeftType.LOKR: "lokr",
+    PeftType.LORA: "lora_",
+    PeftType.OFT: "oft_",
+    PeftType.POLY: "poly_",
+    PeftType.VBLORA: "vblora_",
+    PeftType.VERA: "vera_lambda",
```
BenjaminBossan, line 121:
```suggestion
    PeftType.VERA: "vera_lambda_",
```

src/peft/peft_model.py
```diff
@@ -1137,6 +1151,20 @@ def load_adapter(
         load_result = set_peft_model_state_dict(
             self, adapters_weights, adapter_name=adapter_name, ignore_mismatched_sizes=ignore_mismatched_sizes
         )
+        missing_keys, unexpected_keys = load_result.missing_keys, load_result.unexpected_keys
+        tuner = self.peft_config[adapter_name].peft_type
+
+        if unexpected_keys:
```
BenjaminBossan, line 1157: In a first step, I would suggest not making any warnings or print statements here, since this reflects the current state of things. Instead, I would suggest that we only determine the `adapter_missing_keys` and then override the `load_results.missing_keys` with the new value. WDYT?

BenjaminBossan (changes requested):
Thanks a lot for this PR. I made a few suggestion, please check.

On top of those, I would also like to see a test being added. LMK if you feel up to that task and if you need support with it.

src/peft/peft_model.py
yaswanth19: Makes sense- overwriting the missing keys should work. Incase adapter is not loaded perfectly then the user can inspect  the `load_result`.

yaswanth19:
Yes @BenjaminBossan, Let me know which test cases should be handled. I will add those.

BenjaminBossan:
> Let me know which test cases should be handled. I will add those.

I checked our existing tests and I think it would actually be fine if we extend one of our existing tests and check that the `missing_keys` are empty. Take a look at this one:

https://github.com/huggingface/peft/blob/f4cf170a9c51d822f950cde0a0e1c87dc013403a/tests/testing_common.py#L520-L536

There we're just loading two adapters and do no further checks (the test is just that there is no error). We could assign the `load_result` and assert that there are no missing keys for both of them. WDYT?

Apart from that, I just noticed that after your last pushes, when I go to https://github.com/huggingface/peft/pull/2084/files, for some reason I see 27 changed files (even though it only says 1 file on this page). Not sure what's going on. Maybe it's a GitHub bug, but it makes it super hard to review the PR. I would suggest to wait a day and see if it goes away by itself. If not, maybe you need to rebase on the latest main or if that doesn't help, create a fresh PR. Again, let's wait a day and only try those options if there is still the same error then.

BenjaminBossan:
Hmm, I still see 27 changed files :-/ Could you please try re-basing? Otherwise, we may need a new PR with just the relevant changes.

yaswanth19:
Rebased the branch correctly and added test cases. Please review the PR.

src/peft/peft_model.py
```diff
@@ -1185,6 +1200,19 @@ def load_adapter(
...
+
+        tuner_prefix = PEFT_TYPE_TO_PREFIX_MAPPING.get(tuner, "")
+        adapter_missing_keys = []
+
+        # Filter missing keys specific to the current adapter and tuner prefix.
+        for key in missing_keys:
+            if tuner_prefix in key and adapter_name in key:
+                adapter_missing_keys.append(key)
+
+        load_result = _IncompatibleKeys(missing_keys=adapter_missing_keys, unexpected_keys=unexpected_keys)
```
BenjaminBossan, line 1214: I guess you probably first tried to override `load_result.missing_keys` and as that doesn't work, you decided to create a new instance of `_IncompatibleKeys`. However, this is not necessary. You can actually mutate `load_result.missing_keys`, i.e. you can call `clear`, `append`, and `extend` on it.

BenjaminBossan (changes requested):
Thanks for the updates. The rebase appears to have fixed the issue with the GH diff, nice. I have a suggestion to simplify the code a bit. Moreover, could you please run `make style` on the PR?

yaswanth19:
Done with changes.

BenjaminBossan:
As you can see, the CI is failing and the reason is that the test does not pass when checking VB-LoRA. The reason for that is that the vector bank in VB-LoRA is shared among all layers, so there isn't an individual one for each layer. I don't think there is an easy way to account for that in the code changes you made. My suggestion is therefore to simply skip the test for VB-LoRA, adding a comment explaining why. Could you please adjust the test accordingly?

To check locally that the error is fixed, just run `pytest tests/ -k test_load_multiple_adapters -v`.

yaswanth19:
Done, Skipping the VBLORA test case and added appropriate comment.

BenjaminBossan (approved):
Great, thanks for this PR, this should hopefully reduce confusion about missing keys in the future. Nice work.

yaswanth19:
Yup Thanks for guiding me, and I would be happy to help with any other issues or feature requests :hugs:

BenjaminBossan:
Sounds good, thanks! You can watch the PEFT issues for potential contributions, especially when there is the `contributions-welcome` tag.

PR #2086, opened 2024-09-23
Update setup.py to update contact info

sayakpaul:


BenjaminBossan (approved):
Thanks. Not sure what this email is used for but still good to update.

PR #2087, opened 2024-09-23
Fix func docstring

kwonmha:
Fix docstring of function `load_adapter`.

1. `model_id` type hint
add Pathlike to keep consistency with other functions that call `load_adapter`.

2. add missing `model_id` docstring
3. remove `peft_config` docstring as `peft_config` argument doesn't exist.


BenjaminBossan (approved):
Thanks for adjusting the type annotation and fixing the docstring.

PR #2090, opened 2024-09-23
FIX Raise an error when performing mixed adapter inference and passing non-existing adapter names

BenjaminBossan:
PEFT allows [mixed batch adapter inference](https://huggingface.co/docs/peft/developer_guides/lora#inference-with-different-lora-adapters-in-the-same-batch), i.e. when predicting, the same batch can use different adapters by passing the `adapter_names` argument. However, when users pass an adapter name that does not correspond to any of the existing adapters, these samples are currently being ignored (i.e. just the base model output is used). This is unexpected and can easily lead to errors, e.g. when users mistype the name of an adapter.

This PR fixes this issue by checking all the existing adapter names first and comparing them to the `adapter_names` that the user passed. If there are unexpected entries, an error is raised.

Due to this fix, an error in the test
`test_mixed_adapter_batches_lora_merged_raises` was discovered and promptly fixed.

### DIFF
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -434,6 +434,19 @@ class LoraModel(BaseTuner):
         if self.training:
             raise ValueError("Cannot pass `adapter_names` when the model is in training mode.")
 
+        # Check that users only passed actually existing adapters.
+        # Note: We cannot do this on the layer level, as each individual layer may not have each adapter. Still, we want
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
ping @SunMarc Could I get a review please, it should hopefully not take long.

BenjaminBossan:
No worries, that happens sometimes. Thanks a lot for the review.

Issue #1962, opened 2024-07-26
Deprecation: Transformers will no longer support `past_key_values` to be tuples


BenjaminBossan:
As reported by @ArthurZucker:

> Quick question, I am seeing this in peft: https://github.com/huggingface/peft/blob/f2b6d13f1dbc971c7653aa65e82822ea2d84bb38/src/peft/peft_model.py#L1665 where there is a reliance on get_seq_length() which we are deprecating + we will no longer convert the cache to tuple object automatically in 2 releases

This will presumably affect all prompt learning methods and thus needs to be fixed soon.

For Llama, I identified the following tests, which would result in `past_key_values` being tuples and can serve as a starting point to work on this issue:

```
tests/test_decoder_models.py::PeftDecoderModelTester::test_inference_safetensors_70_test_trl_internal_testing_tiny_random_LlamaForCausalLM_prefix_tuning
tests/test_decoder_models.py::PeftDecoderModelTester::test_passing_input_embeds_works_70_test_trl_internal_testing_tiny_random_LlamaForCausalLM_prefix_tuning
tests/test_decoder_models.py::PeftDecoderModelTester::test_training_prompt_learning_tasks_72_test_trl_internal_testing_tiny_random_LlamaForCausalLM_prefix_tuning
```

(note that there are more tests that would fail, this is just a selection)

I haven't really worked on the prompt learning methods in PEFT and know little about the inner workings of transformers cache, so any support would be welcome.

piyush-llm:
@BenjaminBossan @ArthurZucker  Can you provide more information regarding deprecation of get_seq_length(). I have some idea about prompt learning methods and transformer's KV Cache.

gante:
Hi @piyush-llm 👋 

Alongside a Cache, the `prepare_inputs_for_generation` method also returns `cache_positions` on models that have received the update. This new tensor contains the indexes of the cache that will be updated, so its last value can be used as the sequence length :)

BenjaminBossan:
Thanks for the details. Also check out #869 for discussion and code snippets.

YOUSEFNANIS:
> As reported by @ArthurZucker:
> 
> > Quick question, I am seeing this in peft: https://github.com/huggingface/peft/blob/f2b6d13f1dbc971c7653aa65e82822ea2d84bb38/src/peft/peft_model.py#L1665
> >  where there is a reliance on get_seq_length() which we are deprecating + we will no longer convert the cache to tuple object automatically in 2 releases
> 
[...]

Issue #869, opened 2023-08-27
How to correctly use Prefixing Tuning?


Vincent-Li-9701:
### System Info

peft 0.5.0
transformers 4.32.0

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

```
model = AutoModelForSeq2SeqLM.from_pretrained('bigscience/T0pp', load_in_8bit=True)
model = prepare_model_for_int8_training(model)
config = PrefixTuningConfig(
    task_type=TaskType.SEQ_2_SEQ_LM,
    num_virtual_tokens=100,
    token_dim=model.config.hidden_size,
    num_transformer_submodules=1,
    num_attention_heads=model.config.num_heads,
    num_layers=model.config.num_layers,
    encoder_hidden_size=1792,
)
model = get_peft_model(model, config)
```

### Expected behavior

I'm assuming `num_layers`, `num_attention_heads`, and `token_dim` need to match the base model. In the sample `num_transformer_submodules` is 1. But encoder-decoder has two transformers right? Should this be 2? 

When I run the code above I got

```
File "/python3.10/site-packages/transformers/models/t5/modeling_t5.py", line 551, in forward
position_bias = position_bias + mask  # (batch_size, n_heads, seq_length, key_length)
RuntimeError: The size of tensor a (3) must match the size of tensor b (103) at non-singleton dimension 3
```
When I print out the shape of `position_bias` and `mask`. `mask` has 100 more tokens than `position_bias` seems like on the decoder side. It's also taking in the prefix embeddings

Vincent-Li-9701:
@pacman100 Would you mind taking a look?

WhoopeeHg:
have you been solved this problem?

Vincent-Li-9701:
> have you been solved this problem?
@WhoopeeHg no unfortunately I decided to skip the prefix tuning part since I found that to be less effective than P-Tuning or LoRA on my dataset.

Vincent-Li-9701:
Based on my discussion with others, the problem seems to surface when we load the model with 8 bit quantization.

vikram71198:
@pacman100 @BenjaminBossan this issue occurs even without 8 bit quantization and basically renders HF's integration of Prefix Tuning useless, unless this bug is fixed.

BenjaminBossan:
[...]

PR #2096, opened 2024-09-25
Fix to prefix tuning to fit transformers

BenjaminBossan:
See #869, #1962

Edit: Also https://github.com/huggingface/transformers/issues/33852.

Edit2: The issue might be fixed with latest PEFT and transformers, so let's not merge this PR unless we can find a test case that still fails.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ extras["test"] = extras["dev"] + [
     "diffusers",
     "scipy",
     "protobuf",
+    "sentencepiece",
 ]
 
 setup(


src/peft/peft_model.py
```diff
@@ -719,6 +719,12 @@ def get_prompt(self, batch_size: int, task_ids: Optional[torch.Tensor] = None) -
             if TRANSFORMERS_MODELS_TO_PREFIX_TUNING_POSTPROCESS_MAPPING.get(self.config.model_type, None) is not None:
                 post_process_fn = TRANSFORMERS_MODELS_TO_PREFIX_TUNING_POSTPROCESS_MAPPING[self.config.model_type]
                 past_key_values = post_process_fn(past_key_values)
+            elif peft_config.num_transformer_submodules == 1:
+                # Dont' apply this to encoder-decoder models and not to models requiring special processing.
+                # local import in case users use a very old transformers version
+                from transformers import DynamicCache
+
+                past_key_values = DynamicCache.from_legacy_cache(past_key_values)
             return past_key_values
```
zucchini-nlp, line 728:
```suggestion
            elif peft_config.num_transformer_submodules == 1:
                # Dont' apply this to encoder-decoder models and not to models requiring special processing.
                # local import in case users use a very old transformers version
                from transformers import DynamicCache

                past_key_values = DynamicCache.from_legacy_cache(past_key_values)
            elif peft_config.num_transformer_submodules == 2 and self.base_model._supports_cache_class:
                # Dont' apply this to encoder-decoder models that don't support new Cachc format yet
                # If we don't apply this, prefix-tuning fails to update cross-attn cache
                from transformers import EncoderDecoderCache

                past_key_values = EncoderDecoderCache.from_legacy_cache(past_key_values)
                past_key_values.is_updated = {layer_idx: False for layer_idx in range(len(past_key_values.cross_attention_cache.key_cache))}
```

zucchini-nlp (commented):
I added a suggestion on how to make T5 happe with new cache format. This will also require another PR on transformers side, will open now and link here :)

zucchini-nlp (approved):
LGTM, thanks for fixing!

BenjaminBossan:
Thanks so much for your help.

PR #2098, opened 2024-09-25
Add new features: Safe LoRA

chiayi-hsu:
Hello,

We have published a paper called Safe LoRA (https://arxiv.org/abs/2405.16833).
This work focuses on improving the safety of well-trained LoRA models.
Additionally, I have provided an example implementation in the model.py file to illustrate how to apply the Safe LoRA approach effectively.

### DIFF
--- a/src/peft/tuners/lokr/config.py
+++ b/src/peft/tuners/lokr/config.py
@@ -14,7 +14,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import Optional, Union
+from typing import Literal, Optional, Union
 
 from peft.tuners.lycoris_utils import LycorisConfig
 from peft.utils import PeftType
@@ -40,6 +40,8 @@ class LoKrConfig(LycorisConfig):
             Perform rank decomposition of left kronecker product matrix.
         decompose_factor (`int`):
             Kronecker product decomposition factor.
+        rank_dropout_scale ('bool)
+            Whether to scale the rank dropout while training, defaults to `False`.
         target_modules (`Optional[Union[List[str], str]]`):
             The names of the modules to apply the adapter to. If this is specified, only the modules with the specified
             names will be replaced. When passing a string, a regex match will be performed. When passing a list of
@@ -53,8 +55,8 @@ class LoKrConfig(LycorisConfig):
             When passing a list of strings, either an exact match will be performed or it is checked if the name of the
             module ends with any of the passed strings.
         init_weights (`bool`):
-            Whether to perform initialization of adapter weights. This defaults to `True`, passing `False` is
-            discouraged.
+            Whether to perform initialization of adapter weights. This defaults to `True`. Use "lycoris" to initialize
+            weights in the style of the LYCORIS repository. Passing `False` is discouraged.
         layers_to_transform (`Union[List[int], int]`):
             The layer indices to transform. If a list of ints is passed, it will apply the adapter to the layer indices
             that are specified in this list. If a single integer is passed, it will apply the transformations on the
@@ -91,6 +93,7 @@ class LoKrConfig(LycorisConfig):
         metadata={"help": "Perform rank decomposition of left kronecker product matrix."},
     )
     decompose_factor: int = field(default=-1, metadata={"help": "Kronecker product decomposition factor."})
+    rank_dropout_scale: bool = field(default=False, metadata={"help": "Rank dropout scale"})
     target_modules: Optional[Union[list[str], str]] = field(
         default=None,
         metadata={
@@ -103,12 +106,12 @@ class LoKrConfig(LycorisConfig):
         default=None,
         metadata={"help": "List of module names or regex expression of the module names to exclude from LoKr."},
     )
-    init_weights: bool = field(
+    init_weights: Union[bool, Literal["lycoris"]] = field(
         default=True,
         metadata={
             "help": (
-                "Whether to initialize the weights of the LoKr layers with their default initialization. Don't change "
-                "this setting, except if you know exactly what you're doing."
+                "Whether to initialize the weights of the LoKr layers with their default initialization. Can be True, False or 'lycoris'."
+                "Default is True. Don't change this setting to False, except if you know exactly what you're doing."
             ),
         },
     )
--- a/src/peft/tuners/lokr/layer.py
+++ b/src/peft/tuners/lokr/layer.py
@@ -126,6 +126,23 @@ class LoKrLayer(nn.Module, LycorisLayer):
         if adapter_name in self.lokr_t2:
             nn.init.kaiming_uniform_(self.lokr_t2[adapter_name], a=math.sqrt(5))
 
+    # Initializes weight matrices similar to the way initialized in the LyCORIS repository.
+    def reset_adapter_parameters_lycoris_way(self, adapter_name):
+        if adapter_name in self.lokr_w1:
+            nn.init.kaiming_uniform_(self.lokr_w1[adapter_name], a=math.sqrt(5))
+        else:
+            nn.init.kaiming_uniform_(self.lokr_w1_a[adapter_name], a=math.sqrt(5))
+            nn.init.kaiming_uniform_(self.lokr_w1_b[adapter_name], a=math.sqrt(5))
+
+        if adapter_name in self.lokr_w2:
+            nn.init.zeros_(self.lokr_w2[adapter_name])
+        else:
+            nn.init.zeros_(self.lokr_w2_b[adapter_name])
+            nn.init.kaiming_uniform_(self.lokr_w2_a[adapter_name], a=math.sqrt(5))
+
+        if adapter_name in self.lokr_t2:
+            nn.init.kaiming_uniform_(self.lokr_t2[adapter_name], a=math.sqrt(5))
+
     def update_layer(
         self,
         adapter_name: str,
@@ -160,6 +177,7 @@ class LoKrLayer(nn.Module, LycorisLayer):
         self.scaling[adapter_name] = alpha / r
         self.rank_dropout[adapter_name] = rank_dropout
         self.module_dropout[adapter_name] = module_dropout
+        self.rank_dropout_scale[adapter_name] = kwargs["rank_dropout_scale"]
         base_layer = self.get_base_layer()
 
         # Determine shape of LoKr weights
@@ -192,7 +210,10 @@ class LoKrLayer(nn.Module, LycorisLayer):
 
         # Initialize weights
         if init_weights:
-            self.reset_adapter_parameters(adapter_name)
+            if init_weights == "lycoris":
+                self.reset_adapter_parameters_lycoris_way(adapter_name)
+            else:
+                self.reset_adapter_parameters(adapter_name)
         else:
             self.reset_adapter_parameters_random(adapter_name)
 
@@ -215,7 +236,7 @@ class LoKrLayer(nn.Module, LycorisLayer):
             w2 = self.lokr_w2_a[adapter_name] @ self.lokr_w2_b[adapter_name]
 
         # Make weights with Kronecker product
-        weight = make_kron(w1, w2)
+        weight = make_kron(w1, w2, self.scaling[adapter_name])
         weight = weight.reshape(self.get_base_layer().weight.shape)
 
         # Perform rank dropout during training - drop rows of addition weights
@@ -223,7 +244,8 @@ class LoKrLayer(nn.Module, LycorisLayer):
         if self.training and rank_dropout:
             drop = (torch.rand(weight.size(0)) > rank_dropout).float()
             drop = drop.view(-1, *[1] * len(weight.shape[1:])).to(weight.device)
-            drop /= drop.mean()
+            if self.rank_dropout_scale[adapter_name]:
+                drop /= drop.mean()
             weight *= drop
 
         return weight
--- a/src/peft/tuners/lokr/model.py
+++ b/src/peft/tuners/lokr/model.py
@@ -109,6 +109,7 @@ class LoKrModel(LycorisTuner):
         kwargs = config.to_dict()
         kwargs["r"] = config.rank_pattern.get(target_name_key, config.r)
         kwargs["alpha"] = config.alpha_pattern.get(target_name_key, config.alpha)
+        kwargs["rank_dropout_scale"] = config.rank_dropout_scale
 
         if isinstance(target, LoKrLayer):
             target.update_layer(adapter_name, **kwargs)
--- a/src/peft/tuners/lycoris_utils.py
+++ b/src/peft/tuners/lycoris_utils.py
@@ -71,6 +71,7 @@ class LycorisLayer(BaseTunerLayer):
         self.alpha = {}
         self.scaling = {}
         self.rank_dropout = {}
+        self.rank_dropout_scale = {}
         self.module_dropout = {}
 
         # Tuner info
--- a/tests/test_initialization.py
+++ b/tests/test_initialization.py
@@ -32,6 +32,7 @@ from transformers import AutoModelForCausalLM
 from peft import (
     AdaLoraConfig,
     IA3Config,
+    LoKrConfig,
     LoraConfig,
     PeftMixedModel,
     PeftModel,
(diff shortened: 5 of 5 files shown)


BenjaminBossan:
Thanks for proposing this PR to add the Safe LoRA method to PEFT. I have only skimmed the code and paper, so this is not an in-depth review yet.

Based on what I saw, this is my high level understanding. As a user, I start out by training a LoRA adapter on my dataset using an aligned model as the base model. Next, I want to restore safety which may be reduced after training. For this, I take my trained LoRA model, the base model, and the aligned model, Then Safe LoRA will create a new adapter that "injects" the alignment back into my LoRA adapter. Please correct me if my understanding is incorrect.

Implementation-wise, I wonder if we really need a `SafeLoRA` class. Here is a proposal for a different user API, LMK what you think:

```python
peft_model = ...  # my trained LoRA model

# apply_safe_lora is a new function that does what SafeLoRA currently does
# additional arguments like select_layers_type are option args for apply_safe_lora
aligned_model = apply_safe_lora(peft_model, base_model_id, aligned_model_id)

# persist the new safe LoRA adapter
aligned_model.save_pretrained(...)
```

IMO, this would be a simpler API and it would achieve the same thing at the end.

In this example, the `aligned_model` could actually be the same `peft_model` as initially, just with a new LoRA adapter loaded which is the Safe LoRA adapter.

Another concern I saw when I read the code is that it looks like it will require a lot of memory. IIUC, we need to have the PEFT model, a copy of the PEFT model, the base model, and the aligned model in memory all at once. Even on CPU RAM, this could be difficult to achieve for many users. I wonder if it would be possible to load the weights from the base model and the aligned model one at a time, at least as an option.

I have already implemented something like this for another use case. Here is the code:

https://github.com/huggingface/peft/blob/ccc350151f95a9ff95da046bae5671da75eab52f/src/peft/utils/loftq_utils.py#L263-L411

It only works for safetensors, because pickle does not allow to load the weights lazily, but I still think it could be a nice addition.

Also, do we really need a full copy of the PEFT model? Would it be sufficient to only copy the LoRA weights? Ideally, I would like to see a solution that works even if the user has less than twice the memory required for the base model. If this doesn't work, it's also okay, but it would greatly reduce the number of users who can use the method.

chiayi-hsu:
Thank you for taking the time to read the code and the paper.

Yes, your understanding is correct. 

Since SafeLoRA is not a new training method for LoRA, but rather a process that enhances the safety of a **well-trained** LoRA model through certain operations, it may not be necessary to have a separate SafeLoRA class.

The most CPU memory-intensive operation here is likely when executing ```get_aligned_matrix()```, as it requires loading both the base model and the aligned model. If it were possible to implement lazy loading for both models simultaneously, allowing for the subtraction of weights in the same position to obtain the aligned matrix, it could potentially be a solution.

Regarding your last question, my current code in ```projected_weighted()``` does load the complete PEFT model (aligned model + LoRA). However, it is indeed possible to operate only on the LoRA weights without needing to load the weights of the aligned model as well. Once the user has obtained the new LoRA weights, they can then add them to the original aligned model.

BenjaminBossan:
Thanks for confirming and further explaining. Would you be willing to make the suggested changes? I think it would help a lot with user adoption of the new method.

chiayi-hsu:
Yes, I will make the changes you suggested.

I would like to ask if SafeLoRA is modified into the form of a function like ```apply_safelora``` instead of being a ```safeLoRA``` class, should it still be placed under ```tuners/```, or somewhere else?

BenjaminBossan:
> Yes, I will make the changes you suggested.

Great, thanks.

> I would like to ask if SafeLoRA is modified into the form of a function like `apply_safelora` instead of being a `safeLoRA` class, should it still be placed under `tuners/`, or somewhere else?

Good question, I think it should be placed elsewhere. I don't have a strong opinion, how about `utils/safelora.py`?

BenjaminBossan:
@chiayi-hsu do you still plan on working on this?

chiayi-hsu:
Yes, I’ll update the new version this week.

Thanks!

Benjamin Bossan ***@***.***>於 2024年10月28日 週一，18:39寫道：

chiayi-hsu:
Hello, 

I have changed the ```SafeLoRA```` class into the ```apply_safelora```` function in the ````peft/utils/safelora.py````. For loading the base and aligned models, based on your suggestion, I now use ```safetensors``` to load the weights.

Additionally, regarding the projected weights, I originally loaded the weights of both the complete base model and the PEFT model, but I have now modified it to only load the weights of the PEFT model. I think this should be more memory-efficient for the user's GPU/CPU.

If you have any suggestions, please let me know. 

Thank you!

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
...
+from safetensors.torch import save_file
+from transformers.utils import cached_file
+from transformers.utils.hub import get_checkpoint_shard_files
+
+
+@dataclass
+class SafeLoRAConfig:
+    """
+    This is the configuration class to store the configuration of a safeLoRA.
+    """
```
BenjaminBossan, line 35: Could you please also document the parameters in the docstring? You can just use the same description as below in the help.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan, line 1:
```suggestion
# Copyright 2024-present the HuggingFace Inc. team.
```

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+@dataclass
+class SafeLoRAConfig:
+    """
+    This is the configuration class to store the configuration of a safeLoRA.
+    """
+
+    base_model_path: str = field(
+        default=None,
```
BenjaminBossan, line 38: Let's not provide any defaults if `None` is not a valid argument.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
...
+        metadata={"help": "The path of the base model for obtaining the aligned matrix"},
+    )
+
+    aligned_model_path: str = field(
+        default=None,
```
BenjaminBossan, line 43: Let's not provide any defaults if `None` is not a valid argument.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
...
+        metadata={"help": "The path of the aligned model for obtaining the aligned matrix"},
+    )
+
+    peft_model_path: str = field(
+        default=None,
```
BenjaminBossan, line 48: Let's not provide any defaults if `None` is not a valid argument.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
...
+    def __post_init__(self):
+        if self.base_model_path is None:
+            raise ValueError("base_model_path cannot be None.")
+        if self.aligned_model_path is None:
+            raise ValueError("aligned_model_path cannot be None.")
+        if self.peft_model_path is None:
+            raise ValueError("peft_model_path cannot be None.")
+
+
+class SafetensorLoader:
```
BenjaminBossan, line 86: AFAICT, this class is very similar to `loftq_utils._SafetensorLoader`. Instead of copying it with small modifications, would it be possible to re-use the same class? We can move it elsewhere so that it can be more easily imported.

If this proves difficult, please let me know which parts you needed to change and why, and I can help with modifying the class to work in both cases.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
...
+    sl_align = SafetensorLoader(aligned_model_path)
+    sl_base = SafetensorLoader(base_model_path)
+
+    base = [name for name in sl_base.weight_map.keys() if any(v in name for v in list(peft_config["target_modules"]))]
+    align = [
+        name for name in sl_align.weight_map.keys() if any(v in name for v in list(peft_config["target_modules"]))
+    ]
+    v = []
+    for name_base, name_align in zip(base, align):
+        assert (
```
BenjaminBossan, line 165: Let's raise a proper `ValueError`.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
...
+    The dimensions between the base model's weights and the aligned model's weights should be the same.
+    """
+    sl_align = SafetensorLoader(aligned_model_path)
+    sl_base = SafetensorLoader(base_model_path)
+
+    base = [name for name in sl_base.weight_map.keys() if any(v in name for v in list(peft_config["target_modules"]))]
+    align = [
+        name for name in sl_align.weight_map.keys() if any(v in name for v in list(peft_config["target_modules"]))
+    ]
+    v = []
```
BenjaminBossan, line 163: Let's use a more descriptive variable name.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
...
+
+def project_weights(configs, peft_weights, v):
+    ori_peft_weights = copy.deepcopy(peft_weights)
+    names_A = [name for name in peft_weights.keys() if "lora_A" in name]
+    names_B = [name for name in peft_weights.keys() if "lora_B" in name]
+    idx = 0
+    i = 0
+    dis = []
+    cos_total = []
+    for name_A, name_B in zip(names_A, names_B):
```
BenjaminBossan, line 185: Can we use `enumerate` and then we don't need the separate `idx` variable?

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
...
+        v.append((vec).detach().cpu())
+    return v
+
+
+def project_weights(configs, peft_weights, v):
+    ori_peft_weights = copy.deepcopy(peft_weights)
+    names_A = [name for name in peft_weights.keys() if "lora_A" in name]
+    names_B = [name for name in peft_weights.keys() if "lora_B" in name]
+    idx = 0
+    i = 0
```
BenjaminBossan, line 182: Let's give this variable a proper name.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
...
+        A = ori_peft_weights[name_A]
+        if configs.devices != "cpu":
+            P = v[idx].to(torch.bfloat16).to(configs.devices)
+        else:
+            P = v[idx].to("cpu")
+        W = torch.mm(P, ori_peft_weights[name_B])
+        fW = torch.mm(W, A)
+        ori = torch.mm(ori_peft_weights[name_B], A)
+
+        cos = numpy.round(torch.nn.functional.cosine_similarity(fW.reshape(1, -1), ori.reshape(1, -1)).item(), 5)
```
BenjaminBossan, line 195: Can we use torch instead of numpy?

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
...
+        dist = 1 / (1 + torch.norm(peft_weights[name_B].reshape(1, -1) - W.reshape(1, -1)))
+
+        dis.append(dist.item())
+        idx += 1
+    return peft_weights, cos_total
+
+
+def apply_safelora(configs):
+    """
+    ===================================================
```
BenjaminBossan, line 212: Let's remove the `=`. Also, the first line should be a description of what this function does. You can also include a link to the paper.

Regarding the syntax for examples, check for instance here: https://github.com/huggingface/peft/blob/8eeae0a63f28a80ef42b6bfc977c34ffc0b650ff/src/peft/peft_model.py#L1404-L1429

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,249 @@
...
+            peft_weights[name_B] = ori_peft_weights[name_B]
+
+        dist = 1 / (1 + torch.norm(peft_weights[name_B].reshape(1, -1) - W.reshape(1, -1)))
+
+        dis.append(dist.item())
+        idx += 1
+    return peft_weights, cos_total
+
+
+def apply_safelora(configs):
```
BenjaminBossan, line 210: Let's add a type annotation for `SafeLoRAConfig`. Also, let's rename the variable `safelora_config`.

BenjaminBossan (changes requested):
Thanks for the updates. I reviewed the code and left a couple of comments. Before pushing your changes, always make sure to run `make style`.

When this is addressed, there will be a few more steps before we can merge the PR:

1. We should add an example to the `examples/` folder. Ideally, this can be something similar to what's in the paper, so that we can replicate the results there.
2. This feature should be documented.

Without these steps, users will have a hard time finding this feature.

src/peft/utils/safelora.py
chiayi-hsu: I want this function to behave like the ```.from_pretrained``` method, where if the user provides a repository name, it will automatically download the files if they are not available locally, and if the user provides a local path, it will load the model directly. However, in the current function, the input needs to be a PEFT model, and it retrieves the base model from the PEFT model's config. Since my input is not a PEFT model, this causes an error.

https://github.com/huggingface/peft/blob/8eeae0a63f28a80ef42b6bfc977c34ffc0b650ff/src/peft/utils/loftq_utils.py#L262C1-L282C27

BenjaminBossan: I see. I think this can be fixed. Could you try if this change works:

```diff
modified   src/peft/utils/loftq_utils.py
@@ -269,8 +269,15 @@ class _SafetensorLoader:
 
     def __init__(self, peft_model, model_path):
         if model_path is None:
+            if isinstance(peft_model, str):
+                name_or_path = peft_model
+                base_model_prefix = None
+            else:
+                name_or_path = peft_model
+                base_model_prefix = getattr(peft_model.get_base_model(), "base_model_prefix", None)
+            base_model_prefix = getattr(peft_model.get_base_model(), "base_model_prefix", None)
             try:
-                model_path = snapshot_download(peft_model.base_model.config._name_or_path, local_files_only=True)
+                model_path = snapshot_download(name_or_path, local_files_only=True)
             except (AttributeError, HFValidationError) as exc:
                 raise ValueError(
                     "The provided model does not appear to be a transformers model or is a local model. In this case, "
@@ -286,7 +293,7 @@ class _SafetensorLoader:
             model_path = os.path.join(model_path, suffix)
 
         self.model_path = model_path
-        self.base_model_prefix = getattr(peft_model.get_base_model(), "base_model_prefix", None)
+        self.base_model_prefix = base_model_prefix
         self.prefix = "base_model.model."
         self.is_sharded = False
         self.weight_map = None
```

This way, if the first argument is just a HF model id, it should also work. If this works, we should also rename the arguments, but that can be decided later.

chiayi-hsu: I tried the modified version and changed the last line shown below, but it still has issues.

```diff

modified   src/peft/utils/loftq_utils.py
@@ -269,8 +269,15 @@ class _SafetensorLoader:
 
     def __init__(self, peft_model, model_path):
         if model_path is None:
+           if isinstance(peft_model, str):
+                name_or_path = peft_model
+                base_model_prefix = None
+            else:
+                name_or_path = peft_model
+                base_model_prefix = getattr(peft_model.get_base_model(), "base_model_prefix", None)
-            base_model_prefix = getattr(peft_model.get_base_model(), "base_model_prefix", None)
```

1. When the first argument is repo_id (HF model id), the error message is ```LocalEntryNotFoundError: Cannot find an appropriate cached snapshot folder for the specified revision on the local disk and outgoing traffic has been disabled. To enable repo look-ups and downloads online, pass 'local_files_only=False' as input```. To enable automatic model downloading, change ```local_files_only=True``` to ```False``` in ```snapshot_download(name_or_path, local_files_only=True)```.
2. When the first argument is a local path instead of a repo id (HF model id), it cannot pass the section shown below and goes into an exception.  I think it may be necessary to check if the path exists on disk, and only if it doesn’t, proceed with ```snapshot_download```.
```python
    try:
        model_path = snapshot_download(name_or_path,local_files_only=True)
    except (AttributeError, HFValidationError) as exc:
        raise ValueError(
            "The provided model does not appear to be a transformers model or is a local model. In this case, "
            "you must pass the model_path argument that points to the safetensors file."
        ) ```

Comment (BenjaminBossan) [reply src/peft/utils/safelora.py:86]: I see, thanks for testing. How about this: We make `local_files_only` an argument to the class. By default, it is `True`, so same as now, but the caller can pass `False`, which would allow the model to be downloaded. Then `apply_safelora` can also take a `local_files_only` argument, which defaults to `False`, and it is passed down to `_SafetensorLoader`. WDYT?

Comment (chiayi-hsu) [reply src/peft/utils/safelora.py:86]: I think that sounds feasible and there's no issue with it.

Regarding the two points I mentioned above, the user is inputting a local path instead of a repo_id, so the necessary modifications may be as follows:
```diff

class _SafetensorLoader:
+    def __init__(self, peft_model, model_path=None, local_files_only=True):
        if model_path is None:
            if isinstance(peft_model, str):
                name_or_path = peft_model
                base_model_prefix = None
            else:
                name_or_path = peft_model
                base_model_prefix = getattr(peft_model.get_base_model(), "base_model_prefix", None)
+           if os.path.exists(name_or_path):
+                model_path = name_or_path
+           else:
                try:
+                    model_path = snapshot_download(name_or_path,local_files_only=local_files_only)
                except (AttributeError, HFValidationError) as exc:
                    raise ValueError(
                        "The provided model does not appear to be a transformers model or is a local model. In this case, "
                        "you must pass the model_path argument that points to the safetensors file."
                    ) 

        suffix = "model.safetensors"
        if not model_path.endswith(suffix):
            model_path = os.path.join(model_path, suffix)

```

This version works for both the local path and repo_id. 
If you have any suggestions, please let me know.

BenjaminBossan: This sounds good. Let's just rename the `peft_model` argument to `peft_model_or_model_id`. It's a bit ugly but otherwise the name would be confusing.

chiayi-hsu:
Sorry for the inconvenience. The pull request was closed due to syncing with the latest version of PEFT. I have resubmitted the PR.

Issue #2044, opened 2024-08-29
exclude_modules to keep specific layers or other quirky components out of a target_modules selection


bghira:
### Feature request

The LoraConfig class should accept an optional `exclude_modules` list of regular expressions in an analogue to `target_modules` that then should be referenced when matching against entries from `target_modules`.

### Motivation

Targeting `proj_out` and `proj_mlp` on some models greatly improves the results, but we find that the effect is more reliable and robust if the final proj_out and proj_mlp layers are excluded.

### Your contribution

I can quickly integrate, test, and make use of this feature in [bghira/simpletuner](https://github.com/bghira/simpletuner)

I'm not sure I'll have time to submit a PR in the next two weeks, but after that perhaps I will be able to. Someone who grabs it before then will be much appreciated 💌

BenjaminBossan:
Thanks for opening this feature request. You're not the first one who would like to have this feature. If you or someone else wants to work on the PR, contributions are welcome.

That said, I just wanted to ask if you know that there are already ways to achieve this. One way is to use `layers_pattern`, which is used to match the string, and `layers_to_transform`, which is a list of indices of the layers you would like to match. More in the [docs](https://huggingface.co/docs/peft/v0.12.0/en/package_reference/lora#peft.LoraConfig). It's a bit of an obscure feature but sounds like it is exactly what you require.

The most flexible way to tackle this type of problem is, however, to pass a string to `target_modules`. In this case a regex match is applied. This pretty much allows all kinds of fancy patterns, such as matching all `proj_out` except for `layers.31.proj_out`, just as an example.

bghira:
oh, i see. so you are suggesting i programmatically search the keys and build a list of targets that match my exclude and target patterns! that is pretty good. maybe it's enough

BenjaminBossan:
What I meant is something like this: Say I have `facebook/opt-125m`. This model has 12 attention layers. Say I want to match `fc1` but only layers 0-10, not 11. I can achieve this by passing:

```python
config = LoraConfig(target_modules=r".*\.(?!11)\d+\.fc1$")
```

JINO-ROHIT:
Hi @BenjaminBossan just trying to understand and if possible would like to take this up and contribute.

The issue is to add an extra exclude params in the lora config to not use specific layers right? Is this needed since passing a 'fc' layers in target modules matches across all the layers?

JINO-ROHIT:
alright @BenjaminBossan thanks, id like to take this up

[...]

PR #2102, opened 2024-09-26
FEAT: Adding exclude modules param(#2044)

JINO-ROHIT:
This draft PR addresses the exclude modules parameter to keep specific layers out of target modules

Addresses #2044 

Changes made:

1. Changed LoraConfig class to take in exclude_modules param.
2. Added the exclusion logic within the check_target_module_exists method.

@BenjaminBossan Can you have a look if this is the right direction? This is my first time contributing at peft , would love to have your guidance here :)


BenjaminBossan (changes requested):
Thanks for taking on this feature. I haven't done an in-depth review yet, but from a quick look, I saw this:

1. Right now, you only added the option to the `LoraConfig`. However, since the logic is implemented on the `check_target_module_exists` function, actually all methods can benefit from this feature (not the prompt learning ones, as they don't have `target_modules`, but all the rest like LoKr, BOFT, IA³, etc.). Let's add the argument to all relevant configs, i.e. the config of each method that has `target_modules`.
2. We need to have tests for this new feature. Please check https://github.com/huggingface/peft/blob/main/tests/test_tuners_utils.py, which already contains a bunch of tests to check that the correct modules are targeted. Similar tests would need to be added for `exclude_modules`. Could you please take a look? Don't hesitate to ask if you have questions.

JINO-ROHIT:
@BenjaminBossan thanks for the headsup, ill work on the points

JINO-ROHIT:
@BenjaminBossan ive added the test cases too, lmk what you think

tests/test_tuners_utils.py
```diff
@@ -415,6 +415,35 @@ def test_realistic_example(self):
         ]
         assert model.targeted_module_names == expected
 
+class TestExcludedModuleNames(unittest.TestCase):
```
BenjaminBossan, line 418: Can we get additional tests:

1. Let's add tests for `exclude_modules` being a list of str.
2. What happens if all targeted modules are excluded? I think a nice error message would be the right call in this situation.
3. What happens if we have excluded modules that don't exist? I think it's fine if some of the entries from `exclude_modules` don't match, as we also allow that for `target_modules`. But what if none of the entries in `exclude_modules` matches? I think this should also result in an error. WDYT?

BenjaminBossan (changes requested):
Thanks a lot for the latest changes. The tests are already shaping up quite nicely, but let's improve the test coverage a bit more (see my suggestion). 

Moreover, you missed a few configs still. These are all I could find, but please double-check:

- AdaLoraConfig
- BoftConfig
- VBLoRAConfig
- FourierFTConfig
- LNTuningConfig
- PolyConfig

tests/test_tuners_utils.py
JINO-ROHIT: yeap makes sense, let me work through the tests and edge cases, also the missed out modules

JINO-ROHIT: Hey @BenjaminBossan i managed to add 1 and 2 cases. I wanted some help on how 3 should be done. As in, I see currently _check_target_module_exists is a private abstract method that every peft methods override and check for target modules existence. Now for exclude_modules, should we do the same? Or do i open a for loop within inject_adapter and check for keys there?

src/peft/tuners/tuners_utils.py
```diff
@@ -914,6 +914,20 @@ def check_target_module_exists(config, key: str) -> bool | re.Match[str] | None:
         `bool` | `re.Match[str]` | `None`: True of match object if key matches any target modules from config, False or
         None if no match found
     """
+    if config.target_modules and config.exclude_modules:
+        if config.exclude_modules == config.target_modules:
```
BenjaminBossan, line 918: I don't think that this is a sufficient check. Just to give some examples:

- `target_modules` is a list of str and `exclude_modules` is a str, but both essentially target the same modules
- `target_modules` is `["lin0", "lin1", "lin2"]` and `exclude_modules` is `["lin0", "lin1"]`. They look different, but if `"lin2"` doesn't exist, they're actually the same

We thus need to check what modules were actually targeted, we can't rely on the passed arguments.

tests/test_tuners_utils.py
BenjaminBossan: I don't think we need to go that far. If we check closer, we can see e.g. for LoRA:

https://github.com/huggingface/peft/blob/aa3bd8fbf6820bb000e148a289b6b0c48c339774/src/peft/tuners/lora/model.py#L158-L160

And if we check where that's used, it all comes back to this line:

https://github.com/huggingface/peft/blob/aa3bd8fbf6820bb000e148a289b6b0c48c339774/src/peft/tuners/tuners_utils.py#L486

So what would be an elegant way to implement some sanity checks here? I have one proposal, but maybe you have a better idea: Right now, `check_target_module_exists` simply returns a bool, i.e. `True` or `False` if the module matched. We could extend that and, say, return a special object if a module was excluded due to `exclude_modules`.

This special value could be a very simple object. Its `__bool__` should return `False` (so that users who rely on checking `if not self._check_target_module_exists(peft_config, key)` don't suddenly get different results).

When we call `_check_target_module_exists`, we can collect keys that did not match because of `exclude_modules` in one list and keys that did not match for other reasons in another list. When we exit the loop, if no key was matched because they were all matching `exclude_modules`, we can raise a nice error message to say so. Otherwise, we raise the existing error:

https://github.com/huggingface/peft/blob/aa3bd8fbf6820bb000e148a289b6b0c48c339774/src/peft/tuners/tuners_utils.py#L506-L510

On the other hand, if not a single key was excluded due to `exclude_modules`, even though `exclude_modules` was passed by the user, we can give a nice warning saying something like "You passed exclude_modules=[...] but no key was matching this".

BenjaminBossan (changes requested):
Thanks for your further work on this PR. You raised a good question about how to implement the checks I mentioned. I made a proposal, but please LMK if you have a better idea.

JINO-ROHIT:
Hey, thanks for the suggestions. I think I fixed both the issues. Happy to work if more fixed needed, lmk

JINO-ROHIT:
gotcha, ive made the changes and i think it looks okayish now

BenjaminBossan (changes requested):
Great refactor, thanks a lot, this looks very clean now. I only found a small issue, otherwise this looks good.

JINO-ROHIT:
yeap, done now :)

BenjaminBossan:
Please run `make style` :)

JINO-ROHIT:
hmm weird, i actually ran make style already. is this because of windows? should i switch to a ubuntu and try again?

BenjaminBossan:
I don't think it's a Windows issue. Do you perhaps have another ruff version installed? CI uses v0.6.8.

JINO-ROHIT:
yeah i think its the version, hopefully it works now

JINO-ROHIT:
my bad , i didnt notice the doc-builder wasnt installed, it should be alright now

JINO-ROHIT:
i see the tests are failed due to a typeerror, this seems like an older python version run right?

JINO-ROHIT:
sure lemme get on it

JINO-ROHIT:
done!

BenjaminBossan:
You really seem to be on the bad side of the CI, now there is an issue with ruff again. It should be easy enough to address, please check.

JINO-ROHIT:
yeah seems like after adding the import, I had to lowercase the List Union[list[int], int]

BenjaminBossan:
Of course that's not the end of it, now there is a merge conflict. It should be easy enough to fix but don't hesitate to ask if you have questions.

Thanks A TON for your patience.

JINO-ROHIT:
hey, any clue why these tests are failing?

BenjaminBossan:
Looks like it's unrelated, I'll investigate tomorrow.

JINO-ROHIT:
alrighty

JINO-ROHIT:
ahh okay, like this?

BenjaminBossan (changes requested):
The logic still doesn't work quite right but the good news is that we can simplify it.

Right now, we check `self.targeted_module_names` and also `is_target_modules_in_base_model`, which is a bit redundant. So let's get rid of the latter.

Additionally, to ensure that X-LoRA does not raise an error (since there, it's okay to not have target modules), we need to check that we're not using dummy target modules (a special value used by X-LoRA). I sketched this out and this diff makes the tests pass for me:

```diff
modified   src/peft/tuners/tuners_utils.py
@@ -437,13 +437,12 @@ class BaseTuner(nn.Module, ABC):
         peft_config = self._prepare_adapter_config(peft_config, model_config)
 
         self._prepare_model(peft_config, model)
-        is_target_modules_in_base_model = False
         key_list = [key for key, _ in model.named_modules()]
 
-        if getattr(peft_config, "target_modules", None) == DUMMY_TARGET_MODULES:
+        uses_dummy_target_modules = getattr(peft_config, "target_modules", None) == DUMMY_TARGET_MODULES
+        if uses_dummy_target_modules:
             # dummy adapter, we allow not matching any module
             key_list = []
-            is_target_modules_in_base_model = True
 
         # update peft_config.target_modules if required
         peft_config = _maybe_include_all_linear_layers(peft_config, model)
@@ -494,19 +493,12 @@ class BaseTuner(nn.Module, ABC):
                 unmatched_modules.append(key)
             else:
                 self.targeted_module_names.append(key)
-                is_target_modules_in_base_model = True
                 parent, target, target_name = _get_submodules(model, key)
                 ctx = init_empty_weights if low_cpu_mem_usage else nullcontext
                 with ctx():
                     self._create_and_replace(peft_config, adapter_name, target, target_name, parent, current_key=key)
 
-        # Handle X-LoRA case.
-        if not is_target_modules_in_base_model and hasattr(peft_config, "target_modules"):
-            raise ValueError(
-                f"Target modules {peft_config.target_modules} not found in the base model. "
-                f"Please check the target modules and try again."
-            )
-        elif not self.targeted_module_names:
+        if not self.targeted_module_names and not uses_dummy_target_modules:
             if excluded_modules and not unmatched_modules:
                 # All targeted modules were excluded
                 raise ValueError(
```

JINO-ROHIT:
thanks so much for the detailed suggestion!

JINO-ROHIT:
some unrelated tests failing right? i think with windows some tensor equality is not fulfilled?

BenjaminBossan (changes requested):
Yes, CI is failing for unrelated reasons. I'll just re-run until the flaky test passes :)

So this PR is basically finished, but when I did a final review, I saw a minor oversight, please check my comment. But then we should really be good to go!

JINO-ROHIT:
yeap done! thanks for guiding all the way through!

BenjaminBossan (approved):
Thanks for this nice PR, good work.

Issue #2103, opened 2024-09-26
Lora PISSA init: not support gpt2


suyang160:
### System Info

peft  0.13.0
transformers 4.44.2
torch  2.4.0
Python 3.12.4

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [X] My own task or dataset (give details below)

### Reproduction

import os
os.environ["WANDB_DISABLED"] = "true"
from transformers import GPT2LMHeadModel, GPT2Tokenizer, Trainer, TrainingArguments
from peft import LoraConfig, get_peft_model
from torch.utils.data import Dataset
from torchsummary import summary
import torch
from datasets import load_dataset, config
from trl import SFTTrainer

model_name = "gpt2"  
model = GPT2LMHeadModel.from_pretrained(model_name)
tokenizer = GPT2Tokenizer.from_pretrained(model_name)

tokenizer.pad_token = tokenizer.eos_token

lora_config = LoraConfig(
    r=8,  
    lora_alpha=8,  
    lora_dropout=0, 
    target_modules=["attn.c_attn"],  
    init_lora_weights="pissa",
    fan_in_fan_out=True,
    bias="none"
)

model = get_peft_model(model, lora_config)

dataset = load_dataset("imdb", split="train[:1%]")

trainer = SFTTrainer(
    model=model,
    train_dataset=dataset,
    dataset_text_field="text",
    max_seq_length=128,
    tokenizer=tokenizer,
)

### Expected behavior

Hello, I found that current pissa init code forget to consider the fin_in_fin_out parameter to transpose the matrix weight, which makes gpt2 training failed because of dimension mismatch. I have fixed the bug with the following code:

    def pissa_init(self, adapter_name, init_lora_weights):
        weight = self.get_base_layer().weight
        dtype = weight.dtype
        if dtype not in [torch.float32, torch.float16, torch.bfloat16]:
            raise TypeError(
                "Please initialize PiSSA under float32, float16, or bfloat16. "
                "Subsequently, re-quantize the residual model to help minimize quantization errors."
            )
        weight = transpose(weight.to(torch.float32),self.fan_in_fan_out)
        if init_lora_weights == "pissa":
            # USV^T = W <-> VSU^T = W^T, where W^T = weight.data in R^{out_channel, in_channel},
            V, S, Uh = torch.linalg.svd(weight.data, full_matrices=False)
[...]

PR #2104, opened 2024-09-26
FIX: Transpose weight matrix based on fan_in_fan_out condition in PiSSA initialization (#2103)

suyang160:
Previously, the weight matrix was converted to float32 without considering the need for transposition. This update ensures that the weight matrix is transposed when the fan_in_fan_out condition is met, resolving dimension mismatch issues during GPT-2 training.

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -221,7 +221,7 @@ class LoraLayer(BaseTunerLayer):
                 "Please initialize PiSSA under float32, float16, or bfloat16. "
                 "Subsequently, re-quantize the residual model to help minimize quantization errors."
             )
-        weight = weight.to(torch.float32)
+        weight = transpose(weight.to(torch.float32), self.fan_in_fan_out)
         if init_lora_weights == "pissa":
             # USV^T = W <-> VSU^T = W^T, where W^T = weight.data in R^{out_channel, in_channel},
             V, S, Uh = torch.linalg.svd(weight.data, full_matrices=False)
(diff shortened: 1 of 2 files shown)


BenjaminBossan (changes requested):
Thanks for this PR. First of all, my apologies for not responding earlier. The notification somehow slipped my attention and I just wasn't aware of this PR. In the future, feel free to ping my after a couple of days when there is no response.

The fix looks good, thanks for that. Let's add some tests to ensure that this bug doesn't happen again. For this, could you please add the following tests to the [existing PiSSA tests](https://github.com/huggingface/peft/blob/5e91b546354159a143b1c71dc6bfc15569e067ad/tests/test_gpu_examples.py#L1823):

```python
    @pytest.mark.parametrize("device", ["cuda", "cpu"])
    def test_gpt2_pissa_4bit(self, device, tmp_path):
        # see 2104
        self.get_errors(bits=4, device=device, model_id="gpt2", tmp_path=tmp_path)

    @pytest.mark.parametrize("device", ["cuda", "cpu"])
    def test_gpt2_pissa_8bit(self, device, tmp_path):
        # see 2104
        self.get_errors(bits=8, device=device, model_id="gpt2", tmp_path=tmp_path)
```

For this to work, we need to make some changes to these lines though:

https://github.com/huggingface/peft/blob/5e91b546354159a143b1c71dc6bfc15569e067ad/tests/test_gpu_examples.py#L1721

https://github.com/huggingface/peft/blob/5e91b546354159a143b1c71dc6bfc15569e067ad/tests/test_gpu_examples.py#L1730

There, we need to change `isinstance(module, torch.nn.Linear)` to `isinstance(module, (torch.nn.Linear, Conv1D))`, where `Conv1D` is imported from `transformers.pytorch_utils`.

suyang160:
@BenjaminBossan Thank you for your feedback and suggestions, I've updated this PR.

PR #2106, opened 2024-09-27
FIX: Change check if past_key_values is empty

BenjaminBossan:
After transformers merged this PR:

https://github.com/huggingface/transformers/pull/33703

The bool of `past_key_values` (a `Cache` instance) would change from `False` to `True` in one of our checks:

https://github.com/huggingface/peft/blob/ccc350151f95a9ff95da046bae5671da75eab52f/src/peft/peft_model.py#L1779

Use `get_seq_length()` method instead, which is consistent before and after that commit.

I checked the tests with the new change for both transformers before and after that commit and they passed, so this change should be backwards compatible.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -1776,7 +1776,7 @@ class PeftModelForCausalLM(PeftModel):
 
             # no past_key_values or past_key_values empty cache
             requires_prompt_injection = (model_kwargs["past_key_values"] is None) or (
-                isinstance(model_kwargs["past_key_values"], transformers.Cache) and not model_kwargs["past_key_values"]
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
@zucchini-nlp Could you please review? You can ignore the test that's being skipped, it's unrelated and will be treated in a separate PR.

zucchini-nlp (approved):
LGTM! We don't usually check for bool existence of cache object, but I will check why it was `False` before. Unless it was `None` before, in which case it is expected if there is not cache at all

BenjaminBossan:
Thanks for the review @zucchini-nlp. I gave you write access to the repo as otherwise, your review was not deemed sufficient by GitHub to allow me to merge :D

PR #2110, opened 2024-09-28
Update install.md

Salehbigdeli:
When installing from source, user will run `make test`, `make style` or ...


BenjaminBossan (approved):
Thanks for this PR. I agree it makes sense to include all dependencies for this particular install instruction.

PR #2113, opened 2024-09-30
FIX low_cpu_mem_usage consolidates devices

BenjaminBossan:
See: https://github.com/huggingface/diffusers/pull/9510#issuecomment-2378316687

Right now, the `low_cpu_mem_usage=True` option does not consolidate the devices. E.g. when the model is on GPU and the `state_dict` on CPU, the adapter weight will be on CPU after loading, when it should be GPU. This fix ensures that the devices are consolidated.


sayakpaul (approved):
Thanks a lot!

PR #2117, opened 2024-09-30
ENH: Improved attribute access for modules_to_save

### LINKED ISSUE #2099: Using module_to_save to save parameters inited by nn.parameters dose't work!
### System Info

- `transformers` version: 4.44.0                   
- Platform: Windows-10-10.0.19041-SP0              
- Python version: 3.10.14                          
- Huggingface_hub version: 0.24.6                  
- Safetensors version: 0.4.4                       
- Accelerate version: 0.33.0                       
- PyTorch version (GPU?): 2.1.2+cpu (False)        
- Jax version: not installed                       

### Who can help?

@BenjaminBossan @sayakpaul

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

```python
import evaluate
import numpy as np
from transformers import TrainingArguments, Trainer, ViTForImageClassification
from transformers import ViTImageProcessor, DefaultDataCollator
from torchvision.transforms import RandomResizedCrop, Compose, Normalize, ToTensor

food = load_dataset("food101", split="train[:5000]")
labels = food["train"].features["label"].names
label2id, id2label = dict(), dict()
for i, label in enumerate(labels):
    label2id[label] = str(i)
    id2label[str(i)] = label

SIZE = 512
checkpoint = "google/vit-base-patch16-224-in21k"
image_processor = ViTImageProcessor.from_pretrained(checkpoint, size=SIZE)
model = ViTForImageClassification.from_pretrained(
    checkpoint,
    num_labels=len(labels),
    id2label=id2label,
    label2id=label2id,
    image_size=SIZE
)

lora_config = LoraConfig(
    lora_alpha=16,
    lora_dropout=0.3,
    bias='none',
    target_modules=['query', 'key', 'value'],
    modules_to_save=[
        'classifier',
        "model.vit.embeddings.position_embeddings"
    ]
)


normalize = Normalize(mean=image_processor.image_mean, std=image_processor.image_std)
size = (
    image_processor.size["shortest_edge"]
    if "shortest_edge" in image_processor.size
    else (image_processor.size["height"], image_processor.size["width"])
)
_transforms = Compose([RandomResizedCrop(size), ToTensor(), normalize])


accuracy = evaluate.load("accuracy")

def compute_metrics(eval_pred):
    predictions, labels = eval_pred
    predictions = np.argmax(predictions, axis=1)
    return accuracy.compute(predictions=predictions, references=labels)


data_collator = DefaultDataCollator()

    output_dir="my_awesome_food_model",
    remove_unused_columns=False,
    eval_strategy="epoch",
    save_strategy="epoch",
    learning_rate=5e-5,
    per_device_train_batch_size=16,
    gradient_accumulation_steps=4,
    per_device_eval_batch_size=16,
    num_train_epochs=3,
    warmup_ratio=0.1,
    logging_steps=10,
    load_best_model_at_end=True,
    metric_for_best_model="accuracy",
    push_to_hub=True,
)

trainer = Trainer(
    model=model,
    args=training_args,
    data_collator=data_collator,
    train_dataset=food["train"],
    eval_dataset=food["test"],
    tokenizer=image_processor,
)

trainer.train()
```

### Expected behavior

I have increased the resolution of the image to 512, and I hope to retrain the position emb initialized by nn.parameters (as I have increased the image resolution) while fine-tuning the model using Lora.However,  I found that the final saved Lora model did not include the position emb. How should I solve this problem?

BenjaminBossan:
Resolves #2099

So far, if a module was wrapped due to `modules_to_save`, we handled access to the `weight` and `bias` attribute (albeit incorrectly in case of disabled adapters!). However, there could be more attributes than those that could be accessed, in which case we got an error so far.

Instead of special properties, we now implement a generic `__getattr__` method that can deal with any attribute. The implementation is a bit complex to take into account the way that `torch.nn.Module` handles `__getattr__`.

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -227,17 +227,25 @@ class ModulesToSaveWrapper(torch.nn.Module):
         # use a property to ensure that active_adapter is not set directly, instead use the set_adapter method
         return self._active_adapter
 
-    @property
-    def weight(self):
-        if self.active_adapter not in self.modules_to_save:
-            return self.original_module.weight
-        return self.modules_to_save[self.active_adapter].weight
+    def __getattr__(self, name: str):
+        # Note: This whole method may seem overly complex at first but PyTorch messes with __getattr__ in a way that
+        # requires very careful handling to avoid infinite recursion.
+        try:
+            return super().__getattr__(name)
+        except AttributeError:
+            pass
 
-    @property
-    def bias(self):
-        if self.active_adapter not in self.modules_to_save:
-            return self.original_module.bias
-        return self.modules_to_save[self.active_adapter].bias
+        if "_modules" not in self.__dict__:
+            raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
+
+        # Could not find the attribute the PyTorch way. So let's check if it's an attribute on the
+        # original_module/modules_to_save.
+        modules = self.__dict__["_modules"]
+        if self.disable_adapters:
+            module = modules["original_module"]
+        else:
+            module = modules["modules_to_save"][self.active_adapter]
+        return getattr(module, name)
 
     def update(self, adapter_name):
         context_manager = nullcontext()
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -243,8 +243,12 @@ class ModulesToSaveWrapper(torch.nn.Module):
         modules = self.__dict__["_modules"]
         if self.disable_adapters:
             module = modules["original_module"]
-        else:
+        elif self.active_adapter in modules["modules_to_save"]:
             module = modules["modules_to_save"][self.active_adapter]
+        else:
+            # For some reason, there is no module corresponding to the active adapter; this should normally not be
+            # reached and exists as a failsafe (otherwise, a KeyError would be raised)
+            raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
         return getattr(module, name)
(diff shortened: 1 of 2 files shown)


src/peft/utils/other.py
```diff
@@ -227,17 +227,25 @@ def active_adapter(self) -> str:
...
+        if "_modules" not in self.__dict__:
+            raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
+
+        # Could not find the attribute the PyTorch way. So let's check if it's an attribute on the
+        # original_module/modules_to_save.
+        modules = self.__dict__["_modules"]
+        if self.disable_adapters:
+            module = modules["original_module"]
+        else:
+            module = modules["modules_to_save"][self.active_adapter]
```
SunMarc, line 247: That's probably the case but are we sure that `self.active_adapter` is in `modules["modules_to_save"]` since we changed a bit the logic.

SunMarc (approved):
LGTM ! Thanks for adding this ! Just a nit

src/peft/utils/other.py
BenjaminBossan: Yes, normally, there should always be an entry corresponding to `self.active_adapter`. If not, something must have seriously gone wrong (like users overriding private attributes). Still, I have extended the check and added a test just to be on the safe side.

The previous behavior was incorrect, as it didn't check for disabled adapters, this new behavior should be the correct one.

Issue #2115, opened 2024-09-30
Ineffective Fine-Tuning Bug: Using `get_peft_model()` Before Loading LoRA Produces Outputs Identical to the Base Model


Hoper-J:
### System Info

transformers==4.41.2
peft==0.11.1

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

**Solution**: To prevent ineffective fine-tuning when using LoRA, **do not use** `get_peft_model()` **before loading LoRA weights** with `PeftModel.from_pretrained()`. Instead, simply load the LoRA weights using `PeftModel.from_pretrained()` on the original model.

```python
# Correct way
model = PeftModel.from_pretrained(model, PATH)

# Incorrect way
model = get_peft_model(model, lora_config)
model = PeftModel.from_pretrained(model, PATH)
```

**If you’re interested, here’s the background story on how I discovered the issue:**

It took me three hours of eliminating all other potential issues before I found the root cause. The problem arose while I was writing instructional code for instructional purposes, converting an example from using `load_state_dict` to PEFT for learning purposes.

```python
# Original project code (correct):
# Apply LoRA configuration to text_encoder and unet
text_encoder = get_peft_model(text_encoder, lora_config)
unet = get_peft_model(unet, lora_config)

# If set to resume training, load the weights from the last model checkpoint
# You can modify the model_path to specify another path
if resume:
    # Load only the weights from the last training,
    # not the whole model as in model = torch.load(...)
    text_encoder = torch.load(os.path.join(model_path, "text_encoder.pt"))
    unet = torch.load(os.path.join(model_path, "unet.pt"))
```

Converted to PEFT format:

```python
# Apply LoRA configuration to text_encoder and unet
text_encoder = get_peft_model(text_encoder, lora_config)
unet = get_peft_model(unet, lora_config)

# If set to resume training, load the weights from the last model checkpoint
if resume:
    # Load the LoRA model using PEFT's from_pretrained method
    text_encoder = PeftModel.from_pretrained(text_encoder, os.path.join(model_path, "text_encoder"))
    unet = PeftModel.from_pretrained(unet, os.path.join(model_path, "unet"))
```

[...]

PR #2118, opened 2024-09-30
ENH: Warn when from_pretrained misses PEFT keys

BenjaminBossan:
After merging #2084, we now clean up the `missing_keys` when loading a PEFT adapter to remove all but the relevant keys (the fact that base model keys are missing is expected when loading a PEFT adapter).

Since the presence of `missing_keys` now really means that something might have gone wrong during loading, we can now warn the user if they call `PeftModel.from_pretrained`.

Note that load_adapter still does not warn, as here we return the `load_result` and users can already check, but for `from_pretrained`, they don't have that possibility.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -592,11 +592,16 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
             **kwargs,
         )
 
-        if load_result.missing_keys:
+        # 1. Remove VB-LoRA vector bank, since it's a shared parameter set via the VBLoRAModel
+        # 2. Remove the prompt encoder, as it does not need to be part of the checkpoint
+        missing_keys = [
+            k for k in load_result.missing_keys if "vblora_vector_bank" not in k and "prompt_encoder" not in k
+        ]
+        if missing_keys:
             # Let's warn here since (in contrast to load_adapter) we don't return the load result, so it could be quite
             # difficult for users to even notice that something might have gone wrong here. As we filter out non PEFT
             # keys from the missing keys, this gives no false positives.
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
See #2115 for more context.

Ping @yaswanth19 for awareness.

src/peft/peft_model.py
```diff
@@ -592,6 +592,17 @@ def from_pretrained(
...
+        # 2. Remove the prompt encoder, as it does not need to be part of the checkpoint
+        missing_keys = [
+            k for k in load_result.missing_keys if "vblora_vector_bank" not in k and "prompt_encoder" not in k
+        ]
+        if missing_keys:
+            # Let's warn here since (in contrast to load_adapter) we don't return the load result, so it could be quite
+            # difficult for users to even notice that something might have gone wrong here. As we filter out non PEFT
+            # keys from the missing keys, this gives no false positives.
+            warnings.warn(f"Found missing adapter keys while loading the checkpoint: {missing_keys}")
+
```
SunMarc, line 605: Since we do that for `missing_keys`, can this also be done for `unexpected_keys` ?

SunMarc (approved):
Thanks for adding this @BenjaminBossan ! This will be really helpful for users to confirm that everything has been loaded correctly. Left a question

src/peft/peft_model.py
BenjaminBossan: Yes, I plan on investigating this too. I wanted to go one step at a time to be sure not to give "false positive" warnings when in fact everything is okay, so I preferred to be cautious.

Issue #2119, opened 2024-10-01
Request to Include Named Entity Recognition and Relation Extraction Model Finetuning Examples and Guidance Request 


HarikrishnanK9:
### Feature request

As an NLP enthusiast working on Named Entity Recognition (NER) and Relation Extraction (RE) tasks, I would like to request the inclusion of NER and RE-related  examples, best practices, and **guidance for fine-tuning** models specifically for these tasks on your GitHub page.
The documentation on fine-tuning models for NER and RE would help guide researchers in developing state-of-the-art models without having to reinvent the wheel.Currently Finetuning examples for NER and RE tasks are not available there.These techniques have relevance in identifying technical terms,chemical names,etc from texts and recognize the relationship between identified domain specified technical terms.

### Motivation

ner models are critical in various technologies to identify particular domain related words from an input text,Especially if we are dealing with business terms,chemical names,particle names,and other domain specific technical terms.2nd importance is identify the relation between identified entities.Eg: sentence:Apple acquired Beats for $3 billion in 2014;Entities:Apple, Beats, $3 billion,2014;Relation:acquired

### Pre-requisite

First there must be an appropriate dataset for finetuning, including 1)sentences,2)entities corresponding to that sentence in next column and 3)finally relation between identified entities.Dataset must include diverse domain specific datas like finance,medical,chemical,news,politics,etc then only a general kind of model can identify entities generally.If the purpose is domain specific then particulr domain specific dataset is a must.

BenjaminBossan:
I agree that having such examples could be useful for many users. If you have an existing example that uses full fine-tuning that we can take as a starting point, that would be very helpful. I also added a call for contribution tag so that hopefully someone from the community with expertise on this topic can step in.

JINO-ROHIT:
Hi guys, Im happy to take this up, but do we have any example guidelines? Like should we open a seperate folder and should it be a jupyter notebook or py scripts?

BenjaminBossan:
Thanks a lot for offering to work on this @JINO-ROHIT.

There is no specific template we expect from examples. If you check the `examples` directory, you'll see it's a bit chaotic. I would suggest to add either a script or notebook (whatever you like most) to the `examples/token_classification` directory (as NER would be considered "token classification").

As a dataset, I know that [conll](https://huggingface.co/datasets/eriktks/conll2003) is used for NER but maybe there is a better dataset by now or @HarikrishnanK9 has something else in mind.

[...]

PR #2126, opened 2024-10-04
adding peft lora example notebook for ner

JINO-ROHIT:
Hi @BenjaminBossan , 

Ive added a lora example for NER, which is mainly a clone of the huggingface NER notebook by Sylvain here https://huggingface.co/learn/nlp-course/chapter7/2#data-collation with couple changes here and there.

lmk what you think

### DIFF
--- a/examples/token_classification/peft_lora_ner.ipynb
+++ b/examples/token_classification/peft_lora_ner.ipynb
@@ -481,7 +481,7 @@
     {
      "data": {
       "application/vnd.jupyter.widget-view+json": {
-       "model_id": "b026a52831444da0b73341eaff1c1ba3",
+       "model_id": "f3d4b44fe9b34e9bb60de391325f2e3c",
        "version_major": 2,
        "version_minor": 0
       },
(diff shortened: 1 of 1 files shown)


BenjaminBossan (changes requested):
Context: #2119

Thanks @JINO-ROHIT for taking this task. Good find of a base notebook to adapt. @HarikrishnanK9 is this what you were looking for?

I haven't done a fine-grained analysis yet, but it looks like the `compute_metrics` are note being used. Therefore, we don't get the interesting metrics defined in there. Could you please check that?

JINO-ROHIT:
will do

JINO-ROHIT:
i added the eval metric part, the recall and precision seem quite low, seems to be the classes are imbalanced, would you like me to handle that?

examples/token_classification/peft_lora_ner.ipynb
```diff
@@ -0,0 +1,1152 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "trainable params: 294,912 || all params: 108,021,513 || trainable%: 0.2730\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Configure LoRA (Low-Rank Adaptation) for fine-tuning\n",
+    "peft_config = LoraConfig(target_modules = [\"query\", \"key\"])\n",
```
BenjaminBossan, line 445:
```suggestion
    "peft_config = LoraConfig(target_modules=[\"query\", \"key\"], task_type=TaskType.TOKEN_CLS))\n",
```

`TaskType` needs to be imported from PEFT.

examples/token_classification/peft_lora_ner.ipynb
```diff
@@ -0,0 +1,1152 @@
...
+      "  warnings.warn(\n"
+     ]
+    }
+   ],
+   "source": [
+    "args = TrainingArguments(\n",
+    "    \"bert-finetuned-ner-lora\",\n",
+    "    evaluation_strategy=\"epoch\",\n",
+    "    save_strategy=\"epoch\",\n",
+    "    learning_rate=2e-5,\n",
```
BenjaminBossan, line 519: The learning rate should be higher for LoRA. This is probably why you find worse results than expected. I checked  2e-4, 2e-3, and 2e-2, and got best results for 2e-3. However, they were still slightly worse than full fine-tuning and I only ran for 3 epochs, so some further experiments could yield better results still.

examples/token_classification/peft_lora_ner.ipynb
```diff
@@ -0,0 +1,1152 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Import required libraries\n",
```
BenjaminBossan, line 9: Let's add a few sentences to the beginning before starting with the code. Something along the lines of: "In this notebook, we learn how to ...". You can also add the link to the course here.

BenjaminBossan (changes requested):
Thanks a lot for the updates. I checked the notebook and found the source for the low performance, namely the learning rate. Check my comment.

Also, when I first ran this notebook, I got an error:

> ImportError: To be able to use evaluate-metric/seqeval, you need to install the following dependencies['seqeval'] using 'pip install seqeval' for instance'

So let's mention at the beginning which packages the user should install. We can assume that torch and PEFT (and all their dependencies) are installed, but besides `seqeval`, let's also mention `datasets` and `evaluate`.

Moreover, it would be nice to set the batch size to a higher value. The base model is quite small, so even with batch sizes of 32 or more, most users should be able to run the notebook. This will speed up training considerably. You can add a comment that users have to reduce the value if they run OOM.

JINO-ROHIT:
oh yeah, i totally missed the lr part. The metrics look waaay better now, thanks!

BenjaminBossan (approved):
Thanks for adding this notebook to show NER with PEFT, LGTM.

From my point of view, it's ready to merge. I'll give @HarikrishnanK9 some time to take a look to possibly give more feedback. If we don't hear back by Friday, I'll merge.

JINO-ROHIT:
gotcha

HarikrishnanK9:
@JINO-ROHIT Thank You @BenjaminBossan Just go ahead with the merge.

PR #2127, opened 2024-10-04
fix merging bug / update boft conv2d scaling variable

Zeju1997:
Fixing the bug that boft_s cannot be loaded with older version checkpoints.


Zeju1997:
@BenjaminBossan I think this should fix the problem. Best.

BenjaminBossan (approved):
Thanks for the quick fix.

I ran `pytest tests/regression/test_regression.py -s --regression -k boft` on this branch and it passed.

PR #2135, opened 2024-10-07
MNT Remove Python 3.8 since it's end of life

BenjaminBossan:
The end of life of Python 3.8 has arrived:

https://devguide.python.org/versions/

Therefore, Python 3.8 is removed from CI. Many tests are already failing as a transformers import raises an error that looks like it's caused by missing Python 3.8 support:

> E   RuntimeError: Failed to import transformers.models.auto because of the following error (look up to see its traceback):
E   'type' object is not subscriptable

([example 1](https://github.com/huggingface/peft/actions/runs/11207561120/job/31149919883), [example 2](https://github.com/huggingface/peft/actions/runs/11207542441/job/31149871178))

By default, Python 3.11 is now used.

Python 3.12 should be added to the CI matrix now, but that's for a separate PR.

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -2722,10 +2722,8 @@ class PeftHqqGPUTests(unittest.TestCase):
         assert cc_matrix.min() > 0.97
 
 
-# TODO: unskip the tests once https://github.com/casper-hansen/AutoAWQ/issues/466 is fixed
 @require_torch_gpu
 @require_auto_awq
-@pytest.mark.skip(reason="Needs https://github.com/casper-hansen/AutoAWQ/issues/466 to be fixed first")
 class PeftAwqGPUTests(unittest.TestCase):
     r"""
     Awq + peft tests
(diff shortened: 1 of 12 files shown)


tests/test_gpu_examples.py
```diff
@@ -2722,10 +2722,8 @@ def test_hqq_lora_model_outputs(self):
         assert cc_matrix.min() > 0.97
 
 
-# TODO: unskip the tests once https://github.com/casper-hansen/AutoAWQ/issues/466 is fixed
```
sayakpaul, line 2725: Unrelated?

sayakpaul (approved):
Thanks. Just a single comment.

tests/test_gpu_examples.py
BenjaminBossan: Not totally unrelated, I had to update the AutoAWQ wheels and the new wheels should contain the fix mentioned here.

BenjaminBossan:
Update: The transformers issue might be unintended and not caused by intentionally dropping support for Python 3.8, which is only planned for the end of the month (see internal discussion `#transformers-tests`).

sayakpaul:
Will propagate this into `diffusers` once this PR is merged. Cc: @yiyixuxu

PR #2139, opened 2024-10-09
FIX Type annoations in vera/bnb.py

BenjaminBossan:
The file was missing the `from __future__ import annotations` part. As this code is only running nightly with GPU, the normal CI missed this omission. With this PR, nightly CI should be green again.

When I tested it locally with GPU, I didn't check with Python 3.8, so this error did not occur. Once we stop supporting Python 3.8 very soon, this will also no longer be necessary, but we still need to fix it until then.

### DIFF
--- a/src/peft/tuners/vera/bnb.py
+++ b/src/peft/tuners/vera/bnb.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import annotations
 
 import warnings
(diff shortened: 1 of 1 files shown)


sayakpaul (approved):
TIL! How does this affect?

BenjaminBossan:
> TIL! How does this affect?

When you have code like:

```python
def foo(x: list[str]):
    pass
```

This normally requires Python 3.9+ or you need to use `from typing import List` and use `List[str]`. However, with the `__future__` import, you can use the type annotation syntax from Python 3.9+ inside of Python 3.8.

sayakpaul:
Thanks for clarifying!

PR #2140, opened 2024-10-09
FIX Prompt learning with latest transformers error

BenjaminBossan:
The error in PEFT is occurring after this transformers change:

https://github.com/huggingface/transformers/pull/33870

Now, in our tests, some model_kwargs no longer necessarily contain past_key_values, resulting in a KeyError. We now account for this possibility. Affected models were opt and gpt2.

Example of failing CI: https://github.com/huggingface/peft/actions/runs/11256453704/job/31298371856

Note that CI won't detect this issue as it requires the latest transformers install. I ran this locally using `pytest tests/test_decoder_models.py -k "(prefix or prompt) and (opt or gpt2)" -v`.


gante:
Makes sense 👍 (and apologies for breaking things here)

gante (approved):

PR #2148, opened 2024-10-14
Adding a brand new fine-tuning method, Bone.

JL-er:
In the paper "https://arxiv.org/pdf/2409.15371", we introduce a brand new PEFT (Parameter-Efficient Fine-Tuning) method: BONE (BLOCK AFFINE). This is a completely new structure that differs from the LoRA series. In terms of performance, it has already surpassed PISSA.

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -509,17 +509,27 @@ class BaseTuner(nn.Module, ABC):
                 )
             elif not excluded_modules and unmatched_modules:
                 # None of the targeted modules matched
-                raise ValueError(
+                error_msg = (
                     f"Target modules {peft_config.target_modules} not found in the base model. "
                     f"Please check the target modules and try again."
                 )
+                if peft_config.layers_to_transform is not None:
+                    error_msg += f" Note: You specified 'layers_to_transform': {peft_config.layers_to_transform}."
+                if peft_config.layers_pattern is not None:
+                    error_msg += f" You also specified 'layers_pattern': {peft_config.layers_pattern}."
+                raise ValueError(error_msg)
             else:
                 # Some modules did not match and some matched but were excluded
-                raise ValueError(
+                error_msg = (
                     "No modules were targeted for adaptation. "
                     "This might be caused by a combination of mismatched target modules and excluded modules. "
                     "Please check your `target_modules` and `exclude_modules` configuration."
                 )
+                if peft_config.layers_to_transform is not None:
+                    error_msg += f" Note: You specified 'layers_to_transform': {peft_config.layers_to_transform}."
+                if peft_config.layers_pattern is not None:
+                    error_msg += f" You also specified 'layers_pattern': {peft_config.layers_pattern}."
+                raise ValueError(error_msg)
 
         elif hasattr(peft_config, "exclude_modules") and peft_config.exclude_modules and not excluded_modules:
             # exclude_modules was passed but was not used


src/peft/tuners/bone/config.py
```diff
@@ -0,0 +1,126 @@
...
+@dataclass
+class BoneConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`BoneModel`].
+
+    Args:
+        r (`int`):
+            The rank of Bone across different layers. It is best to set 'r' to an even number; otherwise, the default
+            initialization method will not work.
+        apply_GS (`bool`):
```
BenjaminBossan, line 33: Remove

src/peft/tuners/bone/layer.py
```diff
@@ -0,0 +1,260 @@
...
+        init_weights: bool,
+        **kwargs,
+    ) -> None:
+        """Internal function to create bone adapter
+
+        Args:
+            adapter_name (`str`): Name for the adapter to add.
+            r (`int`): Rank for the added adapter.
+            init_weights (`bool`): Whether to initialize weights.
+            apply_GS (`bool`): Whether to apply Gram-Schmidt orthogonalization or not.
```
BenjaminBossan, line 64: Remove

src/peft/tuners/bone/layer.py
```diff
@@ -0,0 +1,260 @@
...
+from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
+
+from einops import rearrange
+
+
+class BoneLayer(BaseTunerLayer):
+    # All names of layers that may contain (trainable) adapter weights
+    adapter_layer_names = ("bone_block",)
+    # All names of other parameters that may contain adapter-related parameters
+    other_param_names = ("bone_r")
```
BenjaminBossan, line 32:
```suggestion
    other_param_names = ("bone_r",)
```

src/peft/tuners/bone/layer.py
```diff
@@ -0,0 +1,260 @@
...
+        if r <= 0:
+            raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
+
+        self.bone_r[adapter_name] = r
+
+        # Determine shape of Bone weights
+        base_layer = self.get_base_layer()
+        if isinstance(base_layer, nn.Linear):
+            self.bone_block[adapter_name] = nn.Parameter(torch.zeros(self.out_features//r, r, r), requires_grad=True)
+        # elif isinstance(base_layer, nn.Conv2d):
```
BenjaminBossan, line 75: Remove

src/peft/tuners/bone/layer.py
```diff
@@ -0,0 +1,260 @@
...
+        """
+        if r <= 0:
+            raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
+
+        self.bone_r[adapter_name] = r
+
+        # Determine shape of Bone weights
+        base_layer = self.get_base_layer()
+        if isinstance(base_layer, nn.Linear):
+            self.bone_block[adapter_name] = nn.Parameter(torch.zeros(self.out_features//r, r, r), requires_grad=True)
```
BenjaminBossan, line 74: Do we need to check if `self.out_features % r == 0`?

src/peft/tuners/bone/layer.py
```diff
@@ -0,0 +1,260 @@
...
+        # elif isinstance(base_layer, nn.Conv2d):
+        #     self.bone_block[adapter_name] = nn.Parameter(
+        #         torch.empty(self.in_features * base_layer.kernel_size[0] * base_layer.kernel_size[0], r),
+        #         requires_grad=True,
+        #     )
+        else:
+            raise TypeError(f"Bone is not implemented for base layers of type {type(base_layer).__name__}")
+
+        # Initialize weights
+    #     if init_weights:
```
BenjaminBossan, line 84: Should this code be uncommented? As is, `init_weights` has no effect.

src/peft/tuners/bone/layer.py
```diff
@@ -0,0 +1,260 @@
...
+import warnings
+from typing import Any, List, Optional, Union
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
+
+from einops import rearrange
```
BenjaminBossan, line 25: We don't have einops as a PEFT dependency. Since this is only used in one place, could you re-write it without using `rearrange`?

src/peft/tuners/bone/layer.py
```diff
@@ -0,0 +1,260 @@
...
+        # (b)float16 while being on CPU, we need to cast the weights to float32, perform the merge and then cast back to
+        # (b)float16 because some CPUs have slow bf16/fp16 matmuls.
+        cast_to_fp32 = device.type == "cpu" and (dtype == torch.float16 or dtype == torch.bfloat16)
+
+        weight_bone = self.bone_block[adapter]
+        
+        if cast_to_fp32:
+            weight_bone = weight_bone.float()
+        
+        #weight_bone = weight_bone.to(orig_weight.dtype)
```
BenjaminBossan, line 209: Remove?

src/peft/tuners/bone/layer.py
```diff
@@ -0,0 +1,260 @@
...
+        else:
+            
+            for active_adapter in self.active_adapters:
+                if active_adapter not in self.bone_block.keys():
+                    continue
+                orig_weight = self.base_layer.weight.data.clone()
+                delta_weight = self.get_delta_weight(active_adapter, orig_weight)
+                orig_weight += delta_weight
+
+            result = F.linear(input=x, weight=orig_weight, bias=self.base_layer.bias)
```
BenjaminBossan, line 239: Just an idea from my side:

Right now, you're calculating `W2 = W + dW`, then `result = W2@x + bias`. Instead, I wonder if we could re-arrange this somehow to use `result = self.base_layer(x, *args, **kwargs)`, then `result = result + bone_result`. The `bone_result` would ideally be calculated in a parameter-efficient way, which should hopefully alleviate the memory overhead that Bone has compared to LoRA. For this, I wonder if the Bone contribution can be calculated group-wise and added incrementally instead of instantiating a big `dW`, so basically trading more compute for less memory. Do you think that would be possible?

src/peft/tuners/bone/layer.py
```diff
@@ -0,0 +1,260 @@
...
+
+            # def fn_bone(weight, b, x, bias):
```
BenjaminBossan, line 241: Remove?

BenjaminBossan (changes requested):
Thanks for this PR to add the Bone method to PEFT. It's already looking fine but I found a few issues, please check out my comments.

Apart from those, to get to the next stage, could you please:

1. Run `make style`
2. Clarify the following: Is this implementation using row or column blocks? Would it make sense to add both as an option?
3. Related to that, depending on the option and the choice of `r`, we probably need to add some sanity checks on the shape of the targeted weight matrix to make sure that we can evenly divide it into blocks, right?
4. Let's add a few tests cases to get started. For this, check [these lines](https://github.com/huggingface/peft/blob/93ddb1015a637e72c6e61a82852c7bb127b13d66/tests/test_custom_models.py#L289-L296) and implement something similar for Bone. Then you can run the tests using `pytest tests/test_custom_models.py -k bone -v` to help identify potential issues.

src/peft/tuners/bone/layer.py
JL-er: I think Bone does not need to init_weight

src/peft/tuners/bone/layer.py
JL-er: I think 'result = result + bone_result' directly is not feasible. The Bone repository has flash-bone that integrates linear and bone calculations, but I'm not very skilled at writing operators, so the current optimization speed is insufficient.

src/peft/tuners/bone/layer.py
JL-er: I need test

src/peft/tuners/bone/layer.py
BenjaminBossan: Just to explain, we usually have the `init_weight` parameter so that we can switch between initializing the weights as identity transform (default, as it normally helps training) and non-identity transform (non-default). Having a non-identity initialization is nice to have for unit testing, where we want to measure the effect of the adapter.

src/peft/tuners/bone/layer.py
BenjaminBossan: If we have `Y = (W + W@dW + dW)@X + b`, we could rewrite it as `Y = W@X + (W@dW + dW)@X + b`, right? And since `W@X + b` is the base result, i.e. `self.base_layer(x, *args, **kwargs)`, this should work?

Another question would be if we can re-define `delta_weight` in a way that would allow to reduce the memory overhead.

https://github.com/huggingface/peft/pull/2148/files#diff-34b7ebce9fc88273adb34c46598bee8a92b72f5c1e7c420585e03938396a1448R210-R214

My thought was if we can somehow calculate the activations per block and add them incrementally to the base result. Maybe it's not possible or it would not help to reduce memory, but if we can trade off compute for less memory, it is usually a worthwhile trade. If it's simple not possible, feel free to ignore my comment :)

src/peft/tuners/bone/layer.py
JL-er: image
Looks good.

src/peft/tuners/bone/layer.py
JL-er: Is it possible to add checkpointing specifically for bone_result? The optimization here indeed requires careful consideration.

JL-er:
> Thanks for this PR to add the Bone method to PEFT. It's already looking fine but I found a few issues, please check out my comments.
> 
> Apart from those, to get to the next stage, could you please:
> 
> 1. Run `make style`
> 2. Clarify the following: Is this implementation using row or column blocks? Would it make sense to add both as an option?
> 3. Related to that, depending on the option and the choice of `r`, we probably need to add some sanity checks on the shape of the targeted weight matrix to make sure that we can evenly divide it into blocks, right?
> 4. Let's add a few tests cases to get started. For this, check [these lines](https://github.com/huggingface/peft/blob/93ddb1015a637e72c6e61a82852c7bb127b13d66/tests/test_custom_models.py#L289-L296) and implement something similar for Bone. Then you can run the tests using `pytest tests/test_custom_models.py -k bone -v` to help identify potential issues.

1.  Run `make style` √
2. Currently, column grouping is supported. This will be explained in the code (as there are still many issues to address, such as quantization training and memory saving). Other grouping methods will be added later.
3. I added a check [here] (https://github.com/JL-er/PEFT-Bone/blob/main/src/peft/tuners/bone/layer.py#L64)
4. Currently, it is possible to fine-tune models like LLaMA normally, but due to the checks mentioned in issue 3, some structures do not pass.
image

I believe it can be merged now.

src/peft/tuners/bone/layer.py
BenjaminBossan: I would not know how. I think users should just use the existing checkpointing utilities if they want to make use of this.

src/peft/tuners/bone/layer.py
BenjaminBossan: Did you consider this?

BenjaminBossan (changes requested):
Thanks for the updates to the PR. We still have a few boxes to tick before we can merge though.

> Currently, it is possible to fine-tune models like LLaMA normally, but due to the checks mentioned in issue 3, some structures do not pass.

This needs to be addressed. Can we change the values so that the check passes? In the worst case, we can define a new module class where the shapes fit. The goal is that all tests are green.

Also:

- Great additions to the test but we should also add tests for the other model types. For instance, check [how HRA did it](https://github.com/huggingface/peft/pull/1864/files#diff-f691a80c5fbe336138af718016fd86299b7c0265b1f51c73577c709dcdfe3545).
- There is still some commented code here and there.
- We need an entry to the docs, again, check HRA. It's important to mention to pros and cons of the method.
- Let's also add an example that users can use to get started. This will greatly help with adoption. Ideally, this would be a script similar to the experiments in your paper/original repo. That way, we can verify that the PEFT implementation is correct.
- And after you finished, don't forget to call `make style`.

src/peft/tuners/bone/layer.py
JL-er: The init_weight is now provided, but it is still initialized from 0.

JL-er:
> Thanks for the updates to the PR. We still have a few boxes to tick before we can merge though.
> 
> > Currently, it is possible to fine-tune models like LLaMA normally, but due to the checks mentioned in issue 3, some structures do not pass.
> 
> This needs to be addressed. Can we change the values so that the check passes? In the worst case, we can define a new module class where the shapes fit. The goal is that all tests are green.
> 
> Also:
> 
> * Great additions to the test but we should also add tests for the other model types. For instance, check [how HRA did it](https://github.com/huggingface/peft/pull/1864/files#diff-f691a80c5fbe336138af718016fd86299b7c0265b1f51c73577c709dcdfe3545).
> * There is still some commented code here and there.
> * We need an entry to the docs, again, check HRA. It's important to mention to pros and cons of the method.
> * Let's also add an example that users can use to get started. This will greatly help with adoption. Ideally, this would be a script similar to the experiments in your paper/original repo. That way, we can verify that the PEFT implementation is correct.
> * And after you finished, don't forget to call `make style`.

Most of the tests have now passed, but due to Bone being initialized to 0, an error occurs when encountering            
 # sanity check that the logits are different
assert not torch.allclose(logits_base, logits_peft, atol=atol, rtol=rtol)

BenjaminBossan:
> Most of the tests have now passed, but due to Bone being initialized to 0, an error occurs when encountering
> #sanity check that the logits are different

Thanks for working on those. Regarding that, I have a suggestion. Right now, we have:

```python
        if init_weights:
            self.reset_bone_parameters(adapter_name, r)

...

    def reset_bone_parameters(self, adapter_name: str, r):
        self.bone_block[adapter_name] = nn.Parameter(torch.zeros(self.out_features // r, r, r), requires_grad=True)
```

I would suggest that for `init_weights=False`, we create the `bone_block` with non-zero entries, e.g. with random uniform weights. That way, for those tests, we can just pass `init_weights=False` and they should pass. A lot of other methods already do that.

JL-er:
> > Most of the tests have now passed, but due to Bone being initialized to 0, an error occurs when encountering
> > #sanity check that the logits are different
> 
> Thanks for working on those. Regarding that, I have a suggestion. Right now, we have:
> 
> ```python
>         if init_weights:
>             self.reset_bone_parameters(adapter_name, r)
> 
> ...
> 
>     def reset_bone_parameters(self, adapter_name: str, r):
>         self.bone_block[adapter_name] = nn.Parameter(torch.zeros(self.out_features // r, r, r), requires_grad=True)
> ```
> 
> I would suggest that for `init_weights=False`, we create the `bone_block` with non-zero entries, e.g. with random uniform weights. That way, for those tests, we can just pass `init_weights=False` and they should pass. A lot of other methods already do that.

The last error, I don't understand why 'different' can pass but 'same' cannot.
image

JL-er:
> > > Most of the tests have now passed, but due to Bone being initialized to 0, an error occurs when encountering
> > > #sanity check that the logits are different
> > 
> > 
> > Thanks for working on those. Regarding that, I have a suggestion. Right now, we have:
> > ```python
> >         if init_weights:
> >             self.reset_bone_parameters(adapter_name, r)
> > 
> > ...
> > 
> >     def reset_bone_parameters(self, adapter_name: str, r):
> >         self.bone_block[adapter_name] = nn.Parameter(torch.zeros(self.out_features // r, r, r), requires_grad=True)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I would suggest that for `init_weights=False`, we create the `bone_block` with non-zero entries, e.g. with random uniform weights. That way, for those tests, we can just pass `init_weights=False` and they should pass. A lot of other methods already do that.
> 
> The last error, I don't understand why 'different' can pass but 'same' cannot. image

This issue has been resolved because 'Bone same' reuses 'base_model.weight'. However, since the Bone structure requires computation through 'base_model.weight', I modified the forward part of the calculation. The calculation method in the commented section might not be compatible with the Bone structure.
image

BenjaminBossan (changes requested):
Thanks a lot for the updates. We're making good progress. I was checking the tests and still found quite a few failures. Btw you can run them locally using `pytest tests/ -k "bone"`.

Currently, tests involving GPT2 are failing because Bone does not support `Conv1D` layers. You can skip these tests, check how other methods achieve that:

https://github.com/huggingface/peft/blob/e8259ff7bc72365ec730efb1f4005a95071d5dfd/tests/test_decoder_models.py#L70-L83

We also have multiple tests failing that require Bone to be initialized as non-identity transform. For those tests to pass, we only need to ensure that `init_weights=False` is passed to the config. Please check how this is achieved for other methods:

https://github.com/huggingface/peft/blob/e8259ff7bc72365ec730efb1f4005a95071d5dfd/tests/test_decoder_models.py#L212-L263

https://github.com/huggingface/peft/blob/e8259ff7bc72365ec730efb1f4005a95071d5dfd/tests/test_encoder_decoder_models.py#L92-L104

https://github.com/huggingface/peft/blob/e8259ff7bc72365ec730efb1f4005a95071d5dfd/tests/test_feature_extraction_models.py#L109-L123

(there are more instances in the same files, please ensure that all are covered)

Let's fix those issues and then hopefully all the tests should pass.

tests/testing_common.py
```diff
@@ -119,6 +120,10 @@
     {
         "target_modules": None,
     },
+    # Bone
+    {
+        "target_modules": None,
```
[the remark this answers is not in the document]
JL-er: Now all the tests have passed : )

JL-er:
image

Is there anything else I need to do about this error?

JL-er:
> image
> 
> Is there anything else I need to do about this error?

I reran make style, and it seems to be fine now. Thank you！！！
image

JL-er:
image
fix

tests/test_custom_models.py
```diff
@@ -1420,7 +1443,7 @@ def test_multiple_adapters_automatic_modules_to_save(self):
         assert "default" in model.base_model.classifier.modules_to_save
         assert "other" in model.base_model.classifier.modules_to_save
 
-    @parameterized.expand([IA3Config, LoHaConfig, LoKrConfig, LoraConfig, HRAConfig])
+    @parameterized.expand([IA3Config, LoHaConfig, LoKrConfig, LoraConfig, HRAConfig, BoneConfig(r=2)])
```
BenjaminBossan, line 1446: This does not quite work, we need the class here, not an instance:
```suggestion
    @parameterized.expand([IA3Config, LoHaConfig, LoKrConfig, LoraConfig, HRAConfig, BoneConfig])
```

Instead, below, add these lines:

```python
        if config_cls == BoneConfig:
            config_cls = partial(config_cls, r=2)
```

There is already similar code for IA3, just put this code blow that.

tests/test_custom_models.py
```diff
@@ -1445,7 +1468,7 @@ def test_multiple_adapters_mixed_modules_to_save(self, config_cls):
         model.set_adapter("other")
         model(**inputs)
 
-    @parameterized.expand([IA3Config, LoHaConfig, LoKrConfig, LoraConfig, HRAConfig])
+    @parameterized.expand([IA3Config, LoHaConfig, LoKrConfig, LoraConfig, HRAConfig, BoneConfig(r=2)])
```
BenjaminBossan, line 1471: Same issue as above.

BenjaminBossan (changes requested):
Thanks for further working on fixing the tests. One part is not working, but I added a suggestion that will fix the problem.

Also, since this PR is now getting close to the finish line, did you have time to work on these point I mentioned earlier?

> * We need an entry to the docs, again, check HRA. It's important to mention to pros and cons of the method.
> * Let's also add an example that users can use to get started. This will greatly help with adoption. Ideally, this would be a script similar to the experiments in your paper/original repo. That way, we can verify that the PEFT implementation is correct.

JL-er:
> * We need an entry to the docs, again, check HRA. It's important to mention to pros and cons of the method.

- Yes, I need to explain the pros and cons of Bone, but where should I write it?
- I plan to move the finetune files from the Bone repository into the example directory for reproducibility. Regarding the correctness of PEFT adaptation, I have tested it locally 3 times and all results were normal. After merging with PEFT, I will test it once more.

BenjaminBossan:
> * Yes, I need to explain the pros and cons of Bone, but where should I write it?

You could check [this PR](https://github.com/huggingface/peft/pull/2160/files) and how the docs are updated there (ignore the rest).

> * I plan to move the finetune files from the Bone repository into the example directory for reproducibility.

That would be a good addition. Let's make it easy for users to pick this method up and adapt it to their own needs.

JL-er:
> > * Yes, I need to explain the pros and cons of Bone, but where should I write it?
> 
> You could check [this PR](https://github.com/huggingface/peft/pull/2160/files) and how the docs are updated there (ignore the rest).
> 
> > * I plan to move the finetune files from the Bone repository into the example directory for reproducibility.
> 
> That would be a good addition. Let's make it easy for users to pick this method up and adapt it to their own needs.

The aforementioned issues have all been resolved.

JL-er:
I'm sorry, due to a conflict with the official PEFT branch, I mistakenly merged it, which caused my previous repository to be deleted and the PR to be forcibly closed. I have now fixed the issue. Should I resubmit or can you reopen my old PR?

BenjaminBossan:
If you can somehow salvage this branch and re-open this PR, it would be better, since that allows us to have the whole discussion in one place. But don't worry if this doesn't work, we can also continue the discussion on the new PR. Just let me know your decision.

JL-er:
> If you can somehow salvage this branch and re-open this PR, it would be better, since that allows us to have the whole discussion in one place. But don't worry if this doesn't work, we can also continue the discussion on the new PR. Just let me know your decision. 
I am really sorry that I could not recover the old branch. Let’s discuss further question under the new PR, which is ready to be tested.

PR #2149, opened 2024-10-14
FIX Don't assume past_key_valus for encoder models

BenjaminBossan:
Don't assume that `past_key_values` is part of the `model_kwargs`.

This fix is similar to #2140 but for encoder-decoder models. It became necessary after https://github.com/huggingface/transformers/pull/34048 was merged into transformers.

Fixes the [currently failing CI](https://github.com/huggingface/peft/actions/runs/11319951259/job/31476746284).


gante (approved):
Makes sense, LGTM 👍

PR #2150, opened 2024-10-14
Use `SFTConfig` instead of `SFTTrainer` keyword args

qgallouedec:
`SFTTrainer`'s keyword args like  `packing`, `dataset_kwargs`, `dataset_text_field`, `max_seq_length` has been deprecated and will be soon removed. Instead, we use `SFTConfig` (subclass of `TrainingArguments`)

This PR updates the code related to `SFTTrainer`.

### DIFF
--- a/docs/source/accelerate/fsdp.md
+++ b/docs/source/accelerate/fsdp.md
@@ -108,7 +108,7 @@ Notice that we are using LoRA with  rank=8, alpha=16 and targeting all linear la
 
 Let's dive a little deeper into the script so you can see what's going on, and understand how it works.
 
-The first thing to know is that the script uses FSDP for distributed training as the FSDP config has been passed. The `SFTTrainer` class handles all the heavy lifting of creating PEFT model using the peft config that is passed. After that when you call `trainer.train()`, Trainer internally uses 🤗 Accelerate to prepare model, optimizer and trainer using the FSDP config to create FSDP wrapped model which is then trained. The main code snippet is below:
+The first thing to know is that the script uses FSDP for distributed training as the FSDP config has been passed. The [`~trl.SFTTrainer`] class handles all the heavy lifting of creating PEFT model using the peft config that is passed. After that when you call `trainer.train()`, Trainer internally uses 🤗 Accelerate to prepare model, optimizer and trainer using the FSDP config to create FSDP wrapped model which is then trained. The main code snippet is below:
 
 ```python
 # trainer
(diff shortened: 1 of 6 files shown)


BenjaminBossan:
Thanks for addressing these trl deprecations. For my understanding, some arguments have been dropped without adding them to `SFTConfig`. Is that because for those, the default values were used?

qgallouedec:
Which ones?

docs/source/accelerate/fsdp.md
```diff
@@ -119,13 +119,6 @@ trainer = SFTTrainer(
     train_dataset=train_dataset,
     eval_dataset=eval_dataset,
     peft_config=peft_config,
-    packing=data_args.packing,
-    dataset_kwargs={
-        "append_concat_token": data_args.append_concat_token,
-        "add_special_tokens": data_args.add_special_tokens,
-    },
-    dataset_text_field=data_args.dataset_text_field,
-    max_seq_length=data_args.max_seq_length,
```
BenjaminBossan, line 128: Here would be an example where arguments are removed from `SFTTrainer` but no equivalent arguments were added to `training_args`.

qgallouedec: Ok I see what you mean.

When you run the script with, for example, `--max_seq_length 123`, the value will no longer feed `data_args` but `training_args`. Everything happens behind the scenes when arguments are parsed.


```
python example.py --output_dir tmp --max_seq_length 123
```

Before:

```python
from dataclasses import dataclass
from typing import Optional
from transformers import HfArgumentParser, TrainingArguments

@dataclass
class DataTrainingArguments:
    dataset_name: Optional[str] = None
    max_seq_length: int = 512

if __name__ == "__main__":
    parser = HfArgumentParser((DataTrainingArguments, TrainingArguments))
    data_args, training_args = parser.parse_args_into_dataclasses()
    print(data_args.max_seq_length)  # 123
```

After:

```python
from dataclasses import dataclass
from typing import Optional
from transformers import HfArgumentParser
from trl import SFTConfig

@dataclass
class DataTrainingArguments:
    dataset_name: Optional[str] = None

if __name__ == "__main__":
    parser = HfArgumentParser((DataTrainingArguments, SFTConfig))
    data_args, training_args = parser.parse_args_into_dataclasses()
    print(training_args.max_seq_length)  # 123
```

BenjaminBossan (approved):
Thanks for the PR and explaining the change. I tried the updated script and it worked. LGTM.

PR #2151, opened 2024-10-15
FIX: Sft train script FSDP QLoRA embedding mean resizing error

BenjaminBossan:
Resizing the embedding layer with `mean_resizing=True,` which has been introduced in transformers > 4.45, will result in an error. This is because for FSDP + QLoRA the embedding matrix can be on meta device, in which case mean resizing fails. Therefore, if these conditions are detected, the script will set `mean_resizing=False`.

Also updated the recommended package versions to newer versions that I have checked to be working.


gante (approved):
LGTM, thank you for this downstream fix 💛

PR #2156, opened 2024-10-17
FIX Missing low_cpu_mem_usage argument

BenjaminBossan:
The newly introduced `low_cpu_mem_usage` argument was not propagated to the `add_adapter` method of all `PeftModel` task types. This is now fixed and tests were added.

Tests were only added for LoRA to avoid running many relatively slow and redundant tests and because `low_cpu_mem_usage` is not implemented for prompt learning models.


sayakpaul (approved):
Thanks! I like the modular design of the tests!

Issue #2155, opened 2024-10-17
LoraConfig conflict when using `layers_to_transform` in `LlamaModel`


Evan02580:
### System Info

peft: 0.13.2
transformers: 4.43.1

### Who can help?

@BenjaminBossan @sayakpaul 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [X] My own task or dataset (give details below)

### Reproduction

When I tried to use LoraConfig and aimed to apply lora in first and last layers like:
```python
lora_config = LoraConfig(
    r = 8,
    lora_alpha=16,
    target_modules=["q_proj", "k_proj", "v_proj", "o_proj"],
    layers_to_transform=[0,31],
    lora_dropout=0,
    bias = "none",
)
model = LlamaModel.from_pretrained("meta-llama/Meta-Llama-3-8B", torch_dtype=torch.bfloat16)
llama_model = get_peft_model(model, lora_config)
```
It came the problem that:
```
*** ValueError: Target modules ['q_proj', 'k_proj', 'v_proj', 'o_proj'] not found in the base model. Please check the target modules and try again.
```
The similar thing happen if I use `layers_pattern` instead of `target_modules` (but it should be my misunderstanding of `layers_pattern`):
```python
lora_config = LoraConfig(
    ...
    layers_to_transform = 1, 
    layers_pattern = ["q_proj", "k_proj", "v_proj", "o_proj"], 
    ...
)
get_peft_model(model, lora_config)
```
```
*** ValueError: Target modules {'v_proj', 'q_proj'} not found in the base model. Please check the target modules and try again.
```
But this time the problem shoud be the problem of default value of `target_modules`.

However, when I use `model = AutoModelForCausalLM.from_pretrained("meta-llama/Meta-Llama-3-8B", torch_dtype=torch.bfloat16, trust_remote_code=True)` instead, it made it.

### Expected behavior

I'm not sure if it was the problem of `LlamaModel`. And I do also confuse about the use of `layers_patten`, since of doc of LoRA mentioned: 

- `layers_to_transform`: List of layers to be transformed by LoRA. If not specified, all layers in `target_modules` are transformed.
- `layers_pattern`: Pattern to match layer names in `target_modules`, if `layers_to_transform` is specified. By default `PeftModel` will look at common layer pattern (`layers`, `h`, `blocks`, etc.), use it for exotic and custom models.

[...]

PR #2157, opened 2024-10-17
DOC: Improve docs for layers_pattern argument

BenjaminBossan:
Addresses part of #2155.

Also fix type annotations where appropriate.


SunMarc (approved):
Thanks for improving the docs !

PR #2158, opened 2024-10-17
Change layers_pattern logic

BenjaminBossan:
Addresses part of #2155. When merging, add **co-author**: `Singaraj B <singaraj.yali@gmail.com>`.

## Description

So far, the `layers_pattern` argument would only work if there was a prefix to the pattern. As an example, if the module name is:

`decoder.layer.0.attn.to_q`

and we pass `layers_pattern="layer"`, this would match. However, if the module name was:

`layer.0.attn.to_q`

i.e. without prefix before `"layer"`, it would not work.

Usually, when we create a model with `AutoModelForFoo.from_pretrained`, the `"layer"` part would never be first. However, if we load a model directly, e.g. through `LlamaModel.from_pretrained`, there is actually no prefix. As a consequence, we get no match there.

With this PR, the prefix is made optional, so that the second pattern also matches.

## Status

I'm not sure yet if this should be merged, as it is technically backwards incompatible. Users can still target the desired modules by carefully crafting a regex for `target_modules` so that it only matches the desired layer indices. However, this is tedious and `layers_pattern` was introduced to avoid having to do this.


BenjaminBossan:
@githubnemo If you have time to look at this, your opinion would be appreciated (but not high prio).

tests/test_tuners_utils.py
```diff
@@ -121,20 +125,26 @@
...
+    # some realistic examples: module using nn.Sequential, without and with prefix before 'blocks'
+    ("blocks.1.weight", ["weight"], [1], ["blocks"], True),
     ("blocks.1.bias", ["weight"], [1], ["blocks"], False),
     ("mlp.blocks.1.weight", ["weight"], [1], ["blocks"], True),
     ("mlp.blocks.1.bias", ["weight"], [1], ["blocks"], False),
+    # multiple indices could potential match, we want the first one to count
+    ("model.layers.1.layers.0.up_proj", ["up_proj"], [1], ["layers"], True),
+    ("model.layers.1.layers.0.up_proj", ["up_proj"], [0], ["layers"], False),
+    ("layers.1.layers.0.up_proj", ["up_proj"], [1], ["layers"], True),
+    ("layers.1.layers.0.up_proj", ["up_proj"], [0], ["layers"], False),
```
githubnemo, line 147: We should make sure that we can target the latter portion as well with a pattern, in this case maybe test a layer pattern like `["\d+\.layers"]`

githubnemo (commented):
LGTM otherwise :)

githubnemo (approved):

PR #2159, opened 2024-10-17
added checks for layers to transforms and layer pattern in lora

JINO-ROHIT:
This PR is aimed to address point 3 in #2155 to adjust the error message when layers transforms and layer patterns are specified.

Tbh, I dont quite understand the functionality of layern_patterns but assuming that we need layers_pattern when layers_transform is applied, ive added this simple check.

### DIFF
--- a/src/peft/tuners/adalora/config.py
+++ b/src/peft/tuners/adalora/config.py
@@ -64,17 +64,17 @@ class AdaLoraConfig(LoraConfig):
         self.exclude_modules = (
             set(self.exclude_modules) if isinstance(self.exclude_modules, list) else self.exclude_modules
         )
-        # check for layers_to_transform and layers_pattern
-        if self.layers_to_transform is not None and self.layers_pattern is None:
-            raise ValueError("When `layers_to_transform` is specified, `layers_pattern` must also be specified. ")
         # if target_modules is a regex expression, then layers_to_transform should be None
         if isinstance(self.target_modules, str) and self.layers_to_transform is not None:
             raise ValueError("`layers_to_transform` cannot be used when `target_modules` is a str.")
 
-        # if target_modules is a regex expression, then layers_pattern should be None
-        if isinstance(self.target_modules, str) and self.layers_pattern is not None:
-            raise ValueError("`layers_pattern` cannot be used when `target_modules` is a str.")
-
+        # check for layers_to_transform and layers_pattern
+        if (self.layers_pattern is not None) and (self.layers_to_transform is None):
+            raise ValueError("When `layers_pattern` is specified, `layers_to_transform` must also be specified. ")
+        
+        # check for layers_to_transform and layers_pattern
+        if (self.layers_to_transform is not None) and (self.layers_pattern is None):
+            raise ValueError("When `layers_to_transform` is specified, `layers_pattern` must also be specified.")
         # Check if 'r' has been set to a non-default value
         if self.r != 8:  # 8 is the default value for 'r' in LoraConfig
             warnings.warn(
--- a/src/peft/tuners/adalora/config.py
+++ b/src/peft/tuners/adalora/config.py
@@ -64,6 +64,11 @@ class AdaLoraConfig(LoraConfig):
         self.exclude_modules = (
             set(self.exclude_modules) if isinstance(self.exclude_modules, list) else self.exclude_modules
         )
+        # check for layers_to_transform and layers_pattern
+        if self.layers_to_transform is not None and self.layers_pattern is None:
+            raise ValueError(
+                "When `layers_to_transform` is specified, `layers_pattern` must also be specified. "
+            )
         # if target_modules is a regex expression, then layers_to_transform should be None
         if isinstance(self.target_modules, str) and self.layers_to_transform is not None:
             raise ValueError("`layers_to_transform` cannot be used when `target_modules` is a str.")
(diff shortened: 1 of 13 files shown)


BenjaminBossan:
Thanks for working on this PR so quickly. Could you please propagate this check to the other config classes that support this too? Also, it would be great to add a test to `tests/test_config.py` for this. LMK if you have questions about this.

Apart from this fix, I think we can also improve the error reporting when users pass `layers_pattern` and `layers_to_transform`. Right now, we only check `target_modules`. In particular, I'm thinking about this message:

https://github.com/huggingface/peft/blob/57a452ac1122da23fadd95aa9e6ef04136346f02/src/peft/tuners/tuners_utils.py#L512-L515

As the user reported, they got the error

> *** ValueError: Target modules {'v_proj', 'q_proj'} not found in the base model. Please check the target modules and try again.

But the real reason was how they defined `layers_pattern` and `layers_to_transform`. Therefore, I think it would be nice if we check if those arguments were set and if yes, extend the error message accordingly. This could be in a separate PR though.

JINO-ROHIT:
ive added a test case and extended to other classes.

I noticed that one unrelated testcase was failing:

```
FAILED tests/test_config.py::TestPeftConfig::test_regex_with_layer_indexing_lora - AssertionError: Regex pattern did not match.
```

Should we try and fix this?

JINO-ROHIT:
ouu also ill raise another PR to handle the second part

JINO-ROHIT:
> As the user reported, they got the error
> 
> > *** ValueError: Target modules {'v_proj', 'q_proj'} not found in the base model. Please check the target modules and try again.
> 
> But the real reason was how they defined `layers_pattern` and `layers_to_transform`. Therefore, I think it would be nice if we check if those arguments were set and if yes, extend the error message accordingly. This could be in a separate PR thoug

ouuu, il raise another PR to handle this second part

BenjaminBossan:
Could you please run `make style`?

JINO-ROHIT:
Hi @BenjaminBossan done!

BenjaminBossan:
@JINO-ROHIT Oh I see there is an error there. It is actually possible to have `layers_pattern` being `None`, as we use a different regex in that case:

https://github.com/huggingface/peft/blob/e8259ff7bc72365ec730efb1f4005a95071d5dfd/src/peft/tuners/tuners_utils.py#L981-L982

This is why a bunch of tests are failing. I didn't know that (or forgot), so this is my bad. I think we can still do the reverse though, i.e. if `layers_to_transform` is `None`, we should not have `layers_pattern`. WDYT?

JINO-ROHIT:
yeah missed this too, did you mean to do something like this?

BenjaminBossan:
Ah no, I don't think that's quite right. You added `if layer_indexes is not None` but above that, we already check `if is_using_layer_indexes`, so that's redundant.

What I mean is that the check

```python
        if self.layers_to_transform is not None and self.layers_pattern is None:
            raise ValueError("When `layers_to_transform` is specified, `layers_pattern` must also be specified. ")
```

should be changed to

```python
        if self.layers_pattern is not None and self.layers_to_transform is None:
            raise ValueError("When `layers_pattern` is specified, `layers_to_transform` must also be specified. ")
```

I also like to add some parentheses in these cases for readability:

`if (self.layers_pattern is not None) and (self.layers_to_transform is None):`

JINO-ROHIT:
actually this check also runs into the same test case failing

```
FAILED tests/test_config.py::TestPeftConfig::test_regex_with_layer_indexing_lora - AssertionError: Regex pattern did not match.
```

ill have a deeper look why, atm i cant wrap my head around this, ill keep you posted

JINO-ROHIT:
yeah, now i get why the testcase werent passing after seeing the other test config files, ive made the reverse check and moved the check down so now i think all the testcases should pass

src/peft/tuners/adalora/config.py
```diff
@@ -68,10 +68,13 @@ def __post_init__(self):
         if isinstance(self.target_modules, str) and self.layers_to_transform is not None:
             raise ValueError("`layers_to_transform` cannot be used when `target_modules` is a str.")
 
-        # if target_modules is a regex expression, then layers_pattern should be None
-        if isinstance(self.target_modules, str) and self.layers_pattern is not None:
-            raise ValueError("`layers_pattern` cannot be used when `target_modules` is a str.")
+        # check for layers_to_transform and layers_pattern
+        if (self.layers_pattern is not None) and (self.layers_to_transform is None):
```
BenjaminBossan, line 72: I think these checks should be adjusted like so:
```suggestion
        if self.layers_pattern and not self.layers_to_transform:
```

Why? 

1. If users pass `layers_pattern=[], layers_to_transform=None`, we should not raise.
2. If users pass `layers_pattern=["foo"], layers_to_transform=[]`, we should raise

Right now, we don't do that because of the strict `None` check.

BenjaminBossan (changes requested):
As you can see, some tests in the CI are failing. This is actually expected, as some of them just don't make sense anymore, like this one:

https://github.com/huggingface/peft/blob/0d5894271bf5848793caa4223eac949827ef6a0d/tests/test_tuners_utils.py#L97

While checking this, I also noticed some room for improvement in the checking logic, please take a look at my comment.

I think this needs to be changed and the tests adjusted a bit, like I mentioned. You can call `pytest tests/test_tuners_utils.py` locally to ensure that all relevant tests pass.

JINO-ROHIT:
Sure, im away for a couple days on a conference, ill be back and fix this :)

BenjaminBossan:
Thanks @JINO-ROHIT. It's not urgent, so take your time, enjoy the conference.

JINO-ROHIT:
Thank you very much @BenjaminBossan , im back . the checks now are making a lot more sense, ive fixed them and run tests.

JINO-ROHIT:
Hi @BenjaminBossan any clue why this is failing?

BenjaminBossan (approved):
Thanks for adding this check to the configs.

PR #2162, opened 2024-10-18
MNT: Remove version pin of diffusers

BenjaminBossan:
The `diffusers<0.21.0` pin was added way back in #936 and then we forget to ever remove it.

This is now causing trouble, as the old diffusers version still uses `cached_download`, which was removed from `huggingface_hub`:

> ImportError: cannot import name 'cached_download' from 'huggingface_hub'


sayakpaul (approved):
Thanks

PR #2163, opened 2024-10-18
FIX VeRA failure on multiple GPUs

BenjaminBossan:
The shared buffers `vera_A` and `vera_B` could be on the wrong device when using multiple GPUs, resulting in an error. This PR moves the them to the correct device to fix the error.

Example of a failing run: https://github.com/huggingface/peft/actions/runs/11396317278/job/31709933958

Since these buffers are shared, I chose *not* to move the whole buffer to the device. Instead, when we create the slices from those buffers during `forward`, I move the devices only there. This could be inefficient in terms of runtime, but IIUC, the alternative would be to create new copies of these buffers per device, using more memory.

The failing tests were introduced in #2076 but the error was already there beforehand.

I did not discover these failing tests earlier because we had a concurrent error caused by a transformers issue which looked very similar and I wrongly assumed that the VeRA error was caused by the same issue. But now that the issue has been fixed, the error still persists, prompting me to investigate.

### DIFF
--- a/src/peft/tuners/vera/bnb.py
+++ b/src/peft/tuners/vera/bnb.py
@@ -161,8 +161,8 @@ if is_bnb_available():
                 lambda_d = lambda_d.float()
                 lambda_b = lambda_b.float()
 
-            sliced_A = vera_A[:, : self.in_features]
-            sliced_B = vera_B[: self.out_features, :]
+            sliced_A = vera_A[:, : self.in_features].to(lambda_d.device)
+            sliced_B = vera_B[: self.out_features, :].to(lambda_d.device)
             lambda_b = lambda_b.unsqueeze(-1)
             lambda_d = lambda_d.unsqueeze(-1)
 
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
@dkopi @vvvm23 It would be great if you could double check if this is the best solution or if there is a better way.

SunMarc (approved):
LGTM ! Thanks for fixing !

dkopi:
Sorry for the (too) late reply, I was really busy when received the notification, and forgot about the PR later.
If it's about the model parallelism (weights of first N layers on GPU 1, next layers on GPU 2 etc) then probably a _slightly_ better solution would be to move whole buffers between GPUs, as with the most optimal split there would be only as many transfers as there are GPUs minus one. Just not sure how transformers library splits the weights by default.
And the difference in runtime should be negligible imo, so the current solution is also good.

In case of other parallelisms, I'm not sure what would be the best solution, but the current one should be ok too.

PR #2167, opened 2024-10-19
fix fsdp_auto_wrap_policy

### LINKED ISSUE #2166: fsdp_auto_wrap_policy is not working when FSDP_TRANSFORMER_CLS_TO_WRAP and the model's _no_split_modules are None.
### System Info

image

### Who can help?

@BenjaminBossan @sayakpaul

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

image
train any model that do not has _no_split_modules in FSDP mode where there is no FSDP_TRANSFORMER_CLS_TO_WRAP

### Expected behavior

in fact in src/peft/utils/other.py: 
```
default_transformer_cls_names_to_wrap = (
        ",".join(model._no_split_modules) if getattr(model, "_no_split_modules", None) is not None else ""
    )
    transformer_cls_names_to_wrap = os.environ.get(
        "FSDP_TRANSFORMER_CLS_TO_WRAP", default_transformer_cls_names_to_wrap
    ).split(",")
```

**produces [""] and then causes error.**

eljandoubi:
fix the issue that fsdp_auto_wrap_policy is not working when FSDP_TRANSFORMER_CLS_TO_WRAP and the model's _no_split_modules are None
@BenjaminBossan @sayakpaul 

Fixes #2166

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -525,6 +525,8 @@ def fsdp_auto_wrap_policy(model):
     ).split(",")
     transformer_cls_to_wrap = {PrefixEncoder, PromptEncoder, PromptEmbedding}
     for layer_class in transformer_cls_names_to_wrap:
+        if len(layer_class)==0:
+            continue
         transformer_cls = get_module_class_from_name(model, layer_class)
         if transformer_cls is None:
             raise Exception("Could not find the transformer layer class to wrap in the model.")
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -525,7 +525,7 @@ def fsdp_auto_wrap_policy(model):
     ).split(",")
     transformer_cls_to_wrap = {PrefixEncoder, PromptEncoder, PromptEmbedding}
     for layer_class in transformer_cls_names_to_wrap:
-        if len(layer_class)==0:
+        if len(layer_class) == 0:
             continue
         transformer_cls = get_module_class_from_name(model, layer_class)
         if transformer_cls is None:
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -3774,6 +3774,11 @@ class TestFSDPWrap:
         # check that this does not raise:
         FSDP(model, auto_wrap_policy=fsdp_auto_wrap_policy(model), use_orig_params=False, sync_module_states=True)
 
+    def test_fsdp_auto_wrap_policy_does_not_raise_on_custom_model(self):
+        # See #2167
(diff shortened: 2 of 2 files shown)


BenjaminBossan:
Thanks for working on this fix. Could you give an example of a model that would currently fail but work with this fix? Ideally, we can build a unit test based on this.

eljandoubi:
1. Donut
2. Pix2Struct

BenjaminBossan:
I could not replicate for Donut:

```python
>>> from peft.utils.other import fsdp_auto_wrap_policy
>>> model = DonutSwinPreTrainedModel.from_pretrained('naver-clova-ix/donut-base')
>>> model._no_split_modules
['DonutSwinStage']
>>> fsdp_auto_wrap_policy(model)  # works
```

For Pix2struct, I do get:

```python
>>> model = Pix2StructForConditionalGeneration.from_pretrained("google/pix2struct-base")
>>> model._no_split_modules
None
>>> fsdp_auto_wrap_policy(model)
Exception: Could not find the transformer layer class to wrap in the model.
```

but of course I don't need to use PEFT's `fsdp_auto_wrap_policy` for FSDP training, it's just there to help users. Do you have a use case where you can't switch to another auto wrap policy?

eljandoubi:
In the Transformers `Trainer` class, it uses `fsdp_auto_wrap_policy` in `_fsdp_qlora_plugin_updates`, which is automatically applied when training a `PeftModel` in FSDP mode.

BenjaminBossan:
Thanks a lot for the pointer, it makes sense that `Trainer` should still work in that case.

Let's add a small test to ensure that the function does not fail in such cases. We already have a test class here:

https://github.com/huggingface/peft/blob/e8259ff7bc72365ec730efb1f4005a95071d5dfd/tests/test_gpu_examples.py#L3737

We can just add the new test in there. As to the model, I think it's sufficient to just create a custom model and check that calling `fsdp_auto_wrap_policy` does not raise an error. So something like:

```python
    def test_fsdp_auto_wrap_policy_does_not_raise_on_custom_model(self):
        # See #2167
        # Avoid raising on custom models since Trainer uses fsdp_auto_wrap_policy automatically for PEFT + FSDP
        class MyModule(nn.Module):
            def __init__(self):
                super().__init__()
                self.lin = nn.Linear(2, 3)

        fsdp_auto_wrap_policy(MyModule())  # does not raise
```

eljandoubi:
You already have a toy model called `SimpleModel`. I am using it for testing.

BenjaminBossan (approved):
Thanks for investigating the issue and providing a fix. LGTM.

> You already have a toy model called SimpleModel. I am using it for testing.

Good catch.

PR #2168, opened 2024-10-22
CPT Tuner

tsachiblau:
Hey,

This pull request introduces Context-Aware Prompt Tuning (CPT), a new and effective technique that builds on In-Context Learning (ICL) and Prompt Tuning (PT) with enhancements through adversarial optimization. CPT allows for better generalization and stability on various classification tasks.

The approach is based on a research paper, which will soon be available. The core idea of CPT is demonstrated and implemented in the following repository:
https://github.com/tsachiblau/CPT.

We are submitting this pull request to integrate the CPT method into the PEFT library, allowing users to experiment with this novel method. Thank you for reviewing this contribution!

The paper is attached
[Context_aware_Prompt_Tuning__Advancing_In_Context_Learning_with_Adversarial_Methods_PEFT.pdf](https://github.com/user-attachments/files/17474228/Context_aware_Prompt_Tuning__Advancing_In_Context_Learning_with_Adversarial_Methods_PEFT.pdf)

Thanks,
Tsachi


BenjaminBossan:
Hi, thanks for creating this PR to add this new method to PEFT.

I did not have time yet to do a review, but I wanted to alert you that with the provided information, you could be de-anonymized as paper author. Not sure if that's a big deal for the submission process, but just wanted to let you know.

tsachiblau:
I plan to upload it to arXiv anyway, so that works for me. Thanks for letting me know.

src/peft/tuners/cpt/model.py
```diff
@@ -0,0 +1,207 @@
...
+        else:
+            CPT_tokens_type_mask = [2] * self.config.num_virtual_tokens
+
+        MIN_VALUE = 1e-10
+
+        # Calculate normalized epsilon values for input, output, and format tokens
+        normalized_format_eps = (
+            self.config.opt_projection_format_epsilon
+            * torch.sqrt(
+                torch.Tensor([self.config.token_dim / 2048])
```
BenjaminBossan, line 100: Why 2048?

src/peft/tuners/cpt/model.py
```diff
@@ -0,0 +1,207 @@
...
+            )
+        )
+        normalized_input_eps = self.config.opt_projection_epsilon * torch.sqrt(
+            torch.Tensor([self.config.token_dim / 2048])
+        )
+
+        epsilon = torch.ones_like(torch.Tensor(CPT_tokens_type_mask)).to(torch.float32) * MIN_VALUE
+        CPT_tokens_type_mask = torch.Tensor(CPT_tokens_type_mask).long()
+
+        epsilon[(CPT_tokens_type_mask > 0) & (torch.remainder(CPT_tokens_type_mask, 4) == 1)] = normalized_format_eps
```
BenjaminBossan, line 110: Same comment about where the numbers come from.

src/peft/tuners/cpt/model.py
```diff
@@ -0,0 +1,207 @@
...
+        epsilon[(CPT_tokens_type_mask > 0) & (torch.remainder(CPT_tokens_type_mask, 4) == 3)] = normalized_format_eps
+        epsilon[(CPT_tokens_type_mask > 0) & (torch.remainder(CPT_tokens_type_mask, 4) == 2)] = normalized_input_eps
+
+        return epsilon
+
+    def projection(self):
```
BenjaminBossan, line 116: I wonder if it would promote readability if we made this method `get_projection` and then it returns `new_embeddings_weights` instead of assigning it directly. Then we can assign on the caller site. WDYT?

src/peft/tuners/cpt/model.py
```diff
@@ -0,0 +1,207 @@
...
+    @staticmethod
+    def calculate_loss(base_model_output, labels, CPT_type_mask, config):
+        """
+        Computes the loss for CPT models with optional exponential decay.
+
+        Args:
+            base_model_output (ModelOutput): Output from the base model containing logits.
+            labels (torch.Tensor): Ground-truth labels for the input tokens.
+            CPT_type_mask (torch.Tensor): Token type mask used for filtering valid loss terms.
+            config (Namespace): Configuration object containing loss-related hyperparameters.
```
BenjaminBossan, line 143: For config, would the type not better be `CPTConfig`?

src/peft/tuners/cpt/model.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Returns:
+            ModelOutput: The base model output with computed loss.
+        """
+
+        if config.opt_weighted_loss_type in ["decay"]:
```
BenjaminBossan, line 149: Why not `if config.opt_weighted_loss_type == "decay":`?

src/peft/tuners/cpt/model.py
```diff
@@ -0,0 +1,207 @@
...
+                decay_value = 1
+                for label_mask_idx in torch.flip(labels_ids, [0]):
+                    exponential_decay[shift_CPT_type_mask[i] == label_mask_idx] = decay_value
+                    decay_value *= config.opt_loss_decay_factor
+                shift_labels_weights[i] *= exponential_decay
+
+            # Compute the weighted mean loss
+            loss = (loss[shift_labels_bool] * shift_labels_weights[shift_labels_bool]).mean()
+            base_model_output.loss = loss
+        elif config.opt_weighted_loss_type not in ["none"]:
```
BenjaminBossan, line 185: Same comment as above?

tests/CPT_test.py
```diff
@@ -0,0 +1,332 @@
...
+    Trainer,
+    TrainingArguments,
+)
+
+from peft import CPTConfig, get_peft_model
+
+
+TEMPLATE = {"input": "input: {}", "intra_seperator": " ", "output": "output: {}", "inter_seperator": "\n"}
+
+MODEL_NAME = "bigscience/bloom-1b7"
```
BenjaminBossan, line 21: We should use a much smaller model for the unit tests. E.g. would  `"hf-internal-testing/tiny-random-OPTForCausalLM"` be enough here? Ideally, we can get the time per test to < 20 sec on CPU.

tests/CPT_test.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def add_string_labels(example):
+        if example["label"] == 0:
+            example["label_text"] = "negative"
+        elif example["label"] == 1:
+            example["label_text"] = "positive"
+        return example
+
+    train_dataset = data["train"].select(range(4)).map(add_string_labels)
+    test_dataset = data["validation"].select(range(10)).map(add_string_labels)
```
BenjaminBossan, line 78: We only use a small sample of the whole dataset, right? Instead of loading the whole dataset and filtering it down so much, how about just hard-coding these few samples? Alternatively, we could create a mini dataset and load that instead.

tests/CPT_test.py
```diff
@@ -0,0 +1,332 @@
...
+
+    return {"train": train_dataset, "test": test_dataset}
+
+
+@pytest.fixture(scope="module")
+def collator(global_tokenizer):
+    class CPTDataCollatorForLanguageModeling(DataCollatorForLanguageModeling):
```
BenjaminBossan, line 85: For my understanding, is this special collator required because of the dataset or because of CPT? If it's CPT-specific, should we add it to PEFT so that users can easily import it?

tests/CPT_test.py
```diff
@@ -0,0 +1,332 @@
...
+                batch["sample_mask"] = tensor_sample_mask
+
+            return batch
+
+    collator = CPTDataCollatorForLanguageModeling(global_tokenizer, training=True, mlm=False)
+    return collator
+
+
+def dataset(data, tokenizer):
+    class CPTDataset(Dataset):
```
BenjaminBossan, line 125: Same argument: If this is CPT-specific, should we properly implement in PEFT?

tests/CPT_test.py
```diff
@@ -0,0 +1,332 @@
...
+                "attention_mask": self.attention_mask[idx],
+                "input_type_mask": self.input_type_mask[idx],
+            }
+
+    dataset = CPTDataset(data, tokenizer, TEMPLATE)
+
+    return dataset
+
+
+def test_model_initialization_text(global_tokenizer, config_text):
```
BenjaminBossan, line 204: IMO this test is not necessary.

tests/CPT_test.py
```diff
@@ -0,0 +1,332 @@
...
+    """Test model loading and PEFT model initialization."""
+    base_model = AutoModelForCausalLM.from_pretrained(MODEL_NAME, cache_dir=".", trust_remote_code=True)
+
+    model = get_peft_model(base_model, config_text)
+    assert model is not None, "PEFT model initialization failed"
+
+
+def test_model_initialization_random(global_tokenizer, config_random):
```
BenjaminBossan, line 212: IMO this test is not necessary.

tests/CPT_test.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+def test_model_training_random(sst_data, global_tokenizer, collator, config_random):
+    """Perform a short training run to verify the model and data integration."""
+
+    base_model = AutoModelForCausalLM.from_pretrained(MODEL_NAME, cache_dir=".", trust_remote_code=True)
+    model = get_peft_model(base_model, config_random)
+    emb = model.prompt_encoder.default.embedding.weight.data.clone().detach()
+    training_args = TrainingArguments(
+        output_dir="./results",
```
BenjaminBossan, line 227: Let's use a temporary directory here. You can use the `tmp_path` fixture from pytest.

tests/CPT_test.py
```diff
@@ -0,0 +1,332 @@
...
+    base_model = AutoModelForCausalLM.from_pretrained(MODEL_NAME, cache_dir=".", trust_remote_code=True)
+
+    model = get_peft_model(base_model, config_random)
+    assert model is not None, "PEFT model initialization failed"
+
+
+def test_model_training_random(sst_data, global_tokenizer, collator, config_random):
+    """Perform a short training run to verify the model and data integration."""
+
+    base_model = AutoModelForCausalLM.from_pretrained(MODEL_NAME, cache_dir=".", trust_remote_code=True)
```
BenjaminBossan, line 223: Are `cache_dir` and `trust_remote_code` required?

tests/CPT_test.py
```diff
@@ -0,0 +1,332 @@
...
+        remove_unused_columns=False,
+        save_strategy="no",
+        logging_steps=1,
+    )
+
+    train_dataset = dataset(sst_data["train"], global_tokenizer)
+
+    trainer = Trainer(model=model, args=training_args, train_dataset=train_dataset, data_collator=collator)
+
+    try:
```
BenjaminBossan, line 239: No need to catch the exception, just let it fail.

tests/CPT_test.py
```diff
@@ -0,0 +1,332 @@
...
+        trainer.train()
+    except Exception as e:
+        pytest.fail(f"Training failed with error: {e}")
+
+    assert torch.all(model.prompt_encoder.default.embedding.weight.data.clone().detach().cpu() == emb.cpu())
```
BenjaminBossan, line 244: For my understanding, this is to test that these embeddings are frozen? Let's add a comment.

tests/CPT_test.py
```diff
@@ -0,0 +1,332 @@
...
+
+    model.prompt_encoder.default.projection()
+    delta_emb = model.prompt_encoder.default.delta_embedding.weight.data.clone().detach()
+    norm_delta = delta_emb.norm(dim=1).cpu()
+    epsilon = model.prompt_encoder.default.get_epsilon().cpu()
+    assert torch.all(norm_delta <= epsilon)
+
+
+def test_model_training_text(sst_data, global_tokenizer, collator, config_text):
```
BenjaminBossan, line 253: Similar comments to the test above.

tests/CPT_test.py
```diff
@@ -0,0 +1,332 @@
...
+
+    model.prompt_encoder.default.projection()
+    delta_emb = model.prompt_encoder.default.delta_embedding.weight.data.clone().detach()
+    norm_delta = delta_emb.norm(dim=1).cpu()
+    epsilon = model.prompt_encoder.default.get_epsilon().cpu()
+    assert torch.all(norm_delta <= epsilon)
+    assert torch.all((norm_delta == 0) == (~non_label_idx))
+
+
+def test_model_batch_training_text(sst_data, global_tokenizer, collator, config_text):
```
BenjaminBossan, line 294: Same argument as for the tests above. Also, what exactly is different in this test, just the batch size? Why is it important to test batch size 1 and 2?

tests/CPT_test.py
BenjaminBossan, line 1: Let's rename this to `test_cpt.py`

BenjaminBossan (changes requested):
Thanks for this PR to add the Context-aware Prompt Tuning method to PEFT.

> I plan to upload it to arXiv anyway, so that works for me. Thanks for letting me know.

It's now there, right: https://arxiv.org/abs/2410.17222? Let's add a link to the paper into the docstring of the config class for reference.

Something small I noticed:

> VERA (Kopiczko et al., 2023) builds on LoRA by incorporating adaptive learning rates

I think this is not an accurate characterization of VeRA. Did you maybe mean to reference LoRA+ instead?

I reviewed your method and added a couple of comments, please check them out. On top of these, I have a more general question for my understanding: According to the paper, part of the reason why this method works will relates to the changes in the loss calculation and how the parameters are updated. Is this fully covered by the CPT implementation here or would users need to consider something in addition when defining their training loop?

Regarding the testing, thanks for including a few functional tests. Let's also add CPT to the general testing framework, similar to how we do it for prompt tuning:

https://github.com/huggingface/peft/blob/fb6108a78e6fb0469084127c61c68aa7c92145f3/tests/testing_common.py#L129

However, IIUC, it only works for decoder models (causal LM), right? That means we need to create a separate `PeftTestConfigManager = ClassInstantier(CLASSES_MAPPING)` instance that uses `CLASSES_MAPPING` with CPT added on top. This instance should then be used in `test_decoder_models.py`.

Finally, before merging this, we also need to add some docs and at least one example. However, we can work on those in a later iteration and iron out the implementation first.

tsachiblau:
Thank you for the constructive feedback! 🙂

> Something small I noticed:

Sorry, I can’t find where I mentioned it. Could you please point me to it?

> Is this fully covered by the CPT implementation here or would users need to consider something in addition when defining their training loop?

Yes, it is fully covered by this implementation, including both loss and projection. Full details are available in the demo (https://github.com/tsachiblau/CPT/tree/main/notebooks) and will also be added to this repository to make it easier to use.

> However, IIUC, it only works for decoder models (causal LM), right? That means we need to create a separate PeftTestConfigManager = ClassInstantier(CLASSES_MAPPING) instance that uses CLASSES_MAPPING with CPT added on top. This instance should then be used in test_decoder_models.py

I’m not sure how to address the next comment. Could you clarify?

> we can work on those in a later iteration and iron out the implementation first.

Let’s handle it later.

As for the remaining comments, I’ve addressed them all and pushed the changes to the branch.

src/peft/tuners/cpt/config.py
```diff
@@ -0,0 +1,102 @@
...
+    cpt_token_ids: Optional[list[int]] = field(
+        default=None, metadata={"help": "Tensor of token IDs used for CPT prompts."}
+    )
+    cpt_mask: Optional[list[int]] = field(default=None, metadata={"help": "Tensor mask applied to CPT tokens."})
+    cpt_tokens_type_mask: Optional[list[int]] = field(
+        default=None, metadata={"help": "Mask indicating the type of each CPT token."}
+    )
+
+    # Prompt tuning initialization method
+    cpt_prompt_tuning_init: Optional[str] = field(
```
BenjaminBossan, line 57: I think this can be type annotated as `CPTPromptInit`.

src/peft/tuners/cpt/config.py
```diff
@@ -0,0 +1,102 @@
...
+    task_type = "CAUSAL_LM"  # Specifies that CPT is used for causal language modeling.
+    num_transformer_submodules = 1  # Number of transformer submodules used.
+
+    def __post_init__(self):
+        """
+        Post-initialization hook to set additional attributes after the config is initialized.
+        """
+        self.peft_type = PeftType.CPT  # Specifies that the PEFT type is CPT.
+        self.target_modules = None  # Placeholder for target modules in CPT.
+        self.task_type = "CAUSAL_LM"  # Ensures task type is causal language modeling.
```
BenjaminBossan, line 102: > > Also, since this and the next arguments should not be None, I think it makes more sense to remove the default=None and to remove the Optional type annotation. Then we don't need to check that in check_config.

> If the user choose RANDOM option then this values should be None.

Okay, let's add a check here to ensure that the argument is correctly set.

BenjaminBossan (changes requested):
Thanks for all the updates to the PR.

> > Something small I noticed:
> 
> Sorry, I can’t find where I mentioned it. Could you please point me to it?

This is in section 2 of the paper, 2nd paragraph starting with "Efficient Fine-Tuning".

> Yes, it is fully covered by this implementation, including both loss and projection.

Okay, nice.

> Yes, we only support causal LMs. Adding test_CPT to testing_common.py causes errors because the configuration is not initialized correctly, and I’m unsure how to address this. Could you clarify how to correctly include my tests?

So I think the following approach should work. See this line:

https://github.com/huggingface/peft/blob/b3176eff49574057d91445e23a0d2f50706f05de/tests/testing_common.py#L202

Let's create a new instance below called `PeftTestConfigManagerForDecoderModels`. You instantiate it the same, but with the class mapping extended to add CPT:

```python
PeftTestConfigManagerForDecoderModels = ClassInstantier({**CLASSES_MAPPING, **DECODER_MODELS_EXTRA})
```

Of course, we need to define `DECODER_MODELS_EXTRA`, which should be:

```python
DECODER_MODELS_EXTRA = {
    "cpt": (CPTConfig, CONFIG_TESTING_KWARGS[12])
}
```

You can add this after the definition of `CLASSES_MAPPING`:

https://github.com/huggingface/peft/blob/b3176eff49574057d91445e23a0d2f50706f05de/tests/testing_common.py#L124

Next, inside of `test_decoder_models.py`, we have to use this new class:

```diff
- from .testing_common import PeftCommonTester, PeftTestConfigManager
+ from .testing_common import PeftCommonTester, PeftTestConfigManagerForDecoderModels as PeftTestConfigManager
```

Let me know if you have further questions.

> As for the remaining comments, I’ve addressed them all and pushed the changes to the branch.

After all the changes have been made, make sure to also call `make style` to make the linter happy.

There are still a couple of unaddressed comments, please check again.

tsachiblau:
> This is in section 2 of the paper, 2nd paragraph starting with "Efficient Fine-Tuning".
Thanks, I will check it out.

I've made the other requested changes. :)

tests/testing_common.py
```diff
@@ -1189,10 +1202,24 @@ def _test_training_prompt_learning_tasks(self, model_id, config_cls, config_kwar
...
+        if issubclass(config_cls, CPTConfig):
+            parameters = []
+            list_names = []
+            for name, param in model.prompt_encoder.named_parameters():
+                if name not in ['default.embedding.weight']:
+                    parameters.append(param)
+                    list_names.append(name)
+                else:
+                    assert param.grad is None
+            ''
```
BenjaminBossan, line 1214: Remove

tests/testing_common.py
```diff
@@ -1189,10 +1202,24 @@ def _test_training_prompt_learning_tasks(self, model_id, config_cls, config_kwar
         loss = output.sum()
         loss.backward()
 
+        if issubclass(config_cls, CPTConfig):
+            parameters = []
+            list_names = []
```
BenjaminBossan, line 1207: Why do we need `list_names`?

tests/testing_common.py
```diff
@@ -1189,10 +1202,24 @@ def _test_training_prompt_learning_tasks(self, model_id, config_cls, config_kwar
...
+            for name, param in model.prompt_encoder.named_parameters():
+                if name not in ['default.embedding.weight']:
```
BenjaminBossan, line 1209:
```suggestion
                if name != "default.embedding.weight":
```

tests/test_cpt.py
```diff
@@ -0,0 +1,299 @@
...
+    TrainingArguments,
+)
+
+from peft import CPTConfig, get_peft_model
+
+
+TEMPLATE = {"input": "input: {}", "intra_seperator": " ", "output": "output: {}", "inter_seperator": "\n"}
+
+MODEL_NAME = "bigscience/bloom-1b7"
+MAX_INPUT_LENGTH = 1024
```
BenjaminBossan, line 36: For unit testing, we should use much smaller models. Check what the other tests are using. One possibility would be `"hf-internal-testing/tiny-random-OPTForCausalLM"`.

tests/testing_common.py
```diff
@@ -1189,10 +1202,24 @@ def _test_training_prompt_learning_tasks(self, model_id, config_cls, config_kwar
...
+            ''
+        else:
+            parameters = model.prompt_encoder.parameters()
+
         # check that prompt encoder has grads
-        for param in model.prompt_encoder.parameters():
+        for param in parameters:
             assert param.grad is not None
 
+
```
BenjaminBossan, line 1222: Remove

tests/test_cpt.py
```diff
@@ -0,0 +1,299 @@
...
+            }
+
+    dataset = CPTDataset(data, tokenizer, TEMPLATE)
+
+    return dataset
+
+
+def test_model_initialization_text(global_tokenizer, config_text):
+    """Test model loading and PEFT model initialization."""
+    base_model = AutoModelForCausalLM.from_pretrained(MODEL_NAME, cache_dir=".", trust_remote_code=True)
```
BenjaminBossan, line 220:
```suggestion
    base_model = AutoModelForCausalLM.from_pretrained(MODEL_NAME)
```

Same below

src/peft/peft_model.py
```diff
@@ -648,6 +650,8 @@ def _setup_prompt_encoder(self, adapter_name: str):
             prompt_encoder = PromptEncoder(config)
         elif config.peft_type == PeftType.PREFIX_TUNING:
             prompt_encoder = PrefixEncoder(config)
+        elif config.peft_type == PeftType.CPT:
+            prompt_encoder = CPTEmbedding(config, self.word_embeddings)
```
BenjaminBossan, line 654: Right now, this would just work even if the model is not a causal LM. Only later would users get an error. Let's add a check here. I think that `model.config.is_decoder` should work, though I'm not 100% sure. Let's raise an error if the model type is not correct and let's also add a test for that error.

BenjaminBossan (changes requested):
Thanks for the updates, this PR is approaching the finish line. Please ensure to always run `make style` though, or else the linter will complain and tests won't run (I checked them locally and they passed).

Now, let's get back to what I mentioned earlier, namely adding some docs and an example. To give a recent example of a PR with docs and examples, check this one:

https://github.com/huggingface/peft/pull/2172/files

The example doesn't have to be this elaborate, but it should be something that users can easily adopt to their own uses cases. Maybe you can add something that resembles one of the experiments from the paper. That way, we can use the example to ensure that the experiment can be replicated with the PEFT implementation.

When writing the docs, put yourself in the shoes of a user who may not have read the paper and might be curious why they should consider this method.

tests/testing_common.py
tsachiblau: It is redundant

tests/testing_common.py
tsachiblau: Done

tests/testing_common.py
tsachiblau: Done

tests/test_cpt.py
tsachiblau: Done

tests/testing_common.py
tsachiblau: Done

tests/test_cpt.py
tsachiblau: Done

src/peft/peft_model.py
tsachiblau: Could you please clarify where I should verify if the model type is correct? model.config.is_decoder is a property of the base model, right?

tsachiblau:
I added the documentation, except for an example, as I'm unsure where to place it. I noticed that some models have examples in the /examples/ directory, but I can't find a way to access these examples from https://huggingface.co/docs/peft/.

src/peft/peft_model.py
BenjaminBossan: Yes, the check would be `self.base_model.config.is_decoder`. You can check it at the start of the `_setup_prompt_encoder` method.

tsachiblau: Done

docs/source/package_reference/cpt.md
```diff
@@ -0,0 +1,31 @@
...
+
+The abstract from the paper is:
+
+*Traditional fine-tuning is effective but computationally intensive, as it requires updating billions of parameters. CPT, inspired by ICL, PT, and adversarial attacks, refines context embeddings in a parameter-efficient manner. By optimizing context tokens and applying a controlled gradient descent, CPT achieves superior accuracy across various few-shot classification tasks, showing significant improvement over existing methods such as LoRA, PT, and ICL.*
+
+## CPTConfig
+
+[[autodoc]] tuners.cpt.config.CPTConfig
+
+## CPTModel
```
BenjaminBossan, line 28: There is no `CPTModel`, only `CPTEmbedding`.

BenjaminBossan (commented):
> except for an example, as I'm unsure where to place it. I noticed that some models have examples in the /examples/ directory, but I can't find a way to access these examples from https://huggingface.co/docs/peft/.

Yes, examples should go into the `examples/` directory. For this method, `examples/causal_language_modeling/` could be a good option.

I'm not sure why you want to access the examples from the docs. The docs can link to the example, but I don't understand what else you would like to achieve there.

docs/source/package_reference/cpt.md
tsachiblau: Done

tsachiblau: Can I add a link to the exmple in the cpt.md file?

BenjaminBossan: Yes, I see no reason not to add a link there. Just be aware that the link would be invalid until the PR is merged.

tsachiblau:
Merge is done

tsachiblau:
Anything else that needs to be done?

BenjaminBossan:
@tsachiblau Could you please run `make style`?

tsachiblau:
Done

BenjaminBossan:
@tsachiblau The linter is still complaining. Do you have the right ruff version installed? It should be v0.6.9.

BenjaminBossan:
@tsachiblau ruff has now passed successfully but doc-builder is complaining, can you try running it?

`doc-builder style src/peft tests docs/source --max_len 119`

tsachiblau:
Done

BenjaminBossan:
The doc builder is still complaining, can you successfully run `make style` locally?

tsachiblau:
Yes, I get this message

tsachiblau:
Yes, I get this message: 

> ruff check --fix src tests examples docs scripts docker
> All checks passed!
> ruff format src tests examples docs scripts docker
> 209 files left unchanged
> doc-builder style src/peft tests docs/source --max_len 119

BenjaminBossan:
Ah yes, somehow I looked at an old log, now the linting indeed passes but the tests are failing. Currently I can't investigate why they're failing, but if the `is_decoder` attribute is not reliable, we unfortunately need to remove the check.

tsachiblau:
Done again :)

BenjaminBossan:
@tsachiblau A couple of tests are failing. Mostly that concerns a test with gradient chechkpointing that checks the existence of a gradient on the prompt encoder. Could you check if this is a false alarm and the test needs adapting or if something else is going on? There is also another failing test during initialization. You can check the logs of the CI for more details. Thanks.

tsachiblau:
Lets try again

src/peft/peft_model.py
```diff
@@ -1779,7 +1779,7 @@ def _cpt_forward(
             else:
                 N_tokens = input_ids.shape[1]
             input_type_mask = torch.zeros((batch_size, N_tokens)).to(device)
-            input_type_mask[:, -1] = 4
+            input_type_mask[:, :] = 4
```
BenjaminBossan, line 1782: `input_type_mask.fill_(4)` would also work. Could you add a short comment on what "4" means here?

tests/testing_common.py
```diff
@@ -1164,7 +1164,10 @@ def _test_training_gradient_checkpointing(self, model_id, config_cls, config_kwa
 
         for n, param in model.named_parameters():
             if "prompt_encoder." in n:  # prompt tuning methods
-                assert param.grad is not None
+                if not issubclass(config_cls, CPTConfig):
```
BenjaminBossan, line 1167: As a reader might not know that `delta_embedding` is associated with CPT, could you please add a comment to explain the presence/absence of grads here?

src/peft/tuners/cpt/config.py
```diff
@@ -0,0 +1,134 @@
...
+    cpt_token_ids: Optional[list[int]] = field(
+        default=None, metadata={"help": "Tensor of token IDs used for CPT prompts."}
+    )
+    cpt_mask: Optional[list[int]] = field(default=None, metadata={"help": "Tensor mask applied to CPT tokens."})
+    cpt_tokens_type_mask: Optional[list[int]] = field(
+        default=None, metadata={"help": "Mask indicating the type of each CPT token."}
+    )
+
+    # Prompt tuning initialization method
+    cpt_prompt_init: Optional[str] = field(
```
BenjaminBossan, line 54: Using `Literal["TEXT", "RANDOM"]` as type annotation would be a bit more precise.

BenjaminBossan (changes requested):
Thanks for the updates. The tests are now passing (some flaky tests are failing, but we can ignore those for how). I did another check of the PR and found some smaller areas for improvement. Moreover, I saw that some of my previous comments are still unaddressed. If you disagree with a suggestion I made, just let me know, not everything needs to be changed, but if there is no reply I can't tell if you read it or not.

src/peft/peft_model.py
tsachiblau: 4 is the id for the tokens used for the loss calculation. I changed the code to 
input_type_mask = torch.ones((batch_size, N_tokens)).to(device) * 4

tests/testing_common.py
tsachiblau: We have two embeddings in the model. delta_embedding is the only one that should be updated with grads.
I added a comment in the code.

src/peft/tuners/cpt/config.py
tsachiblau: > I think this comment is still relevant.
It already exists in the code.

> Using Literal["TEXT", "RANDOM"] as type annotation would be a bit more precise.
I changed it.

tsachiblau:
> Still relevant

what do you mean?

> WDYT about this suggestion?

I can not see what you suggest.


> WDYT?

I can not see what you suggest.

src/peft/tuners/cpt/config.py
```diff
@@ -0,0 +1,82 @@
...
+    # Tokenizer configuration
+    tokenizer_name_or_path: Optional[str] = field(
+        default=None,
+        metadata={
+            "help": "The tokenizer to use for prompt tuning initialization. Only used if prompt_tuning_init is `TEXT`"
+        },
+    )
+
+    # Virtual token configurations
+    num_virtual_tokens: int = field(default=0, metadata={"help": "Number of virtual tokens used in the prompt."})
```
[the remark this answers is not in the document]
BenjaminBossan: I think having 0 as the default here makes little sense. WDYT about using a good default here, say, 10?

src/peft/tuners/cpt/config.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # CPT-specific static attributes
```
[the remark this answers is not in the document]
BenjaminBossan: Does it ever make sense to let users pass these arguments? If not, I would remove them here and place them inside the `__post_init__` method.

src/peft/tuners/cpt/config.py
BenjaminBossan: What I mean is that a check should be performed here which checks what you mentioned in the quote above.

src/peft/tuners/cpt/model.py
```diff
@@ -0,0 +1,207 @@
...
+
+        return prompt_embeddings + delta_prompt_embeddings
+
+    def set_updated_tokens(self):
+        """
+        Sets up a backward hook to selectively update token gradients based on the CPT token type mask.
+        """
+        if self.config.CPT_prompt_tuning_init == PromptTuningInit.TEXT:
+            tensor_ICL_mask = torch.Tensor(self.config.CPT_tokens_type_mask).long()
+            mask_input_template = torch.remainder(tensor_ICL_mask, 4) == 1
```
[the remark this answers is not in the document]
BenjaminBossan: Bumping this comment.

src/peft/tuners/cpt/model.py
BenjaminBossan: My suggestion is to rename this method to `get_projection`. Then, in the last line, instead of `self.delta_embedding.weight.data = new_embeddings_weights`, just return `new_embeddings_weights`. It is then on the caller side that the `delta_embeddings` are updated.

src/peft/tuners/cpt/model.py
BenjaminBossan: Bump.

src/peft/tuners/cpt/model.py
BenjaminBossan: Bump.

src/peft/tuners/cpt/model.py
BenjaminBossan: Why not `if config.opt_weighted_loss_type == "none":`?

tsachiblau:
> My suggestion is to change the type annotation to Literal["none", "decay"].


Done 

> I think having 0 as the default here makes little sense. WDYT about using a good default here, say, 10?


Done

> Does it ever make sense to let users pass these arguments? If not, I would remove them here and place them inside the __post_init__ method.


Done

tsachiblau:
> My suggestion is to change the type annotation to Literal["none", "decay"].


Done.

> I think having 0 as the default here makes little sense. WDYT about using a good default here, say, 10?


Done.

> Does it ever make sense to let users pass these arguments? If not, I would remove them here and place them inside the post_init method.


Done.

> What I mean is that a check should be performed here which checks what you mentioned in the quote above.


Some tests will fail if the user is given the option to initialize it incorrectly, as we previously encountered with causal LM. I think that this initialization if fine.


> My suggestion is to rename this method to get_projection. Then, in the last line, instead of self.delta_embedding.weight.data = new_embeddings_weights, just return new_embeddings_weights. It is then on the caller side that the delta_embeddings are updated.


Done.


> Why not if config.opt_weighted_loss_type == "none":

Done.



All bumping are done as well.

tsachiblau:
I changed it.

BenjaminBossan (approved):
Thanks a lot for addressing my last few comments. The PR LGTM now, thanks for this great addition.

> Yes, examples should go into the examples/ directory. For this method, examples/causal_language_modeling/ could be a good option.

It would be great to eventually provide a standalone example of CPT. This will really help users getting started and thus improve adoption. It's not a prerequisite for merging, just something to mention. Let me know if you want to merge as is or if you would like to add an example first (or if you plan to add it in a separate PR).

tsachiblau:
I will add it later in a separate PR. Thank you very much for all your help and guidance!

BenjaminBossan:
> I will add it later in a separate PR. Thank you very much for all your help and guidance!

Great, looking forward to it. Thanks for your good work and your patience with my review :)

tsachiblau:
Should it apear now on the left side of your website (https://huggingface.co/docs/peft/index) under Adapters?

BenjaminBossan:
> Should it apear now on the left side of your website (https://huggingface.co/docs/peft/index) under Adapters?

Yes, but only if you switch to the "main" docs, as the normal docs show the state of the latest release: https://huggingface.co/docs/peft/main/en/package_reference/cpt.

Issue #2171, opened 2024-10-22
LoRA + DeBERTa: loading model gives erratic, non-deterministic results


jchook:
### System Info
A100 Colab

### Issue

Here is a Colab repro to demonstrate my issue: https://colab.research.google.com/drive/1Z5FL1QDePY0j-8XL0o1V9o27_MbU-02E?usp=sharing

It's very possible that I am doing something fundamentally wrong or using the `peft` framework incorrectly. However, I tried to copy most of this code from [the official LoRA sequence classification example](https://github.com/huggingface/peft/blob/main/examples/sequence_classification/LoRA.ipynb)

On this toy **IMDB sentiment classification example using LoRA + DeBERTA base**, I get consistent ~100% test accuracy when I do any of the following:
- Measure accuracy on the train set during training
- Measure accuracy on the test set immediately after training an epoch
- FFT the base model without the PEFT framework

However, **I get erratic results when saving (per-epoch) and loading a trained LoRA adapter from disk**.

This occurs without any modifications to the code, in the same python session, or across fresh python sessions. So, I have ruled-out residual python state in RAM, etc.

The results appear to shift randomly between the following outcomes: 
- very high accuracy, close to 100% 
- unreasonably low accuracy, close to 0% (i.e. inverted labels)
- something in between

```
{'accuracy': 0.3}
{'accuracy': 0.0}
{'accuracy': 0.979}
{'accuracy': 1.0}
{'accuracy': 0.303}
{'accuracy': 0.812}
{'accuracy': 0.591}
{'accuracy': 1.0}
```

### Who can help?

@BenjaminBossan

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

Here is a Colab repro to demonstrate my issue: https://colab.research.google.com/drive/1Z5FL1QDePY0j-8XL0o1V9o27_MbU-02E?usp=sharing

In the Colab, after running the other cells to install, load the `imdb` dataset, and train a LoRA classifier, I am running this test cell repeatedly and seeing erratic results:

```python
import torch
from peft import PeftModel, PeftConfig
from transformers import AutoModelForCausalLM, AutoTokenizer

peft_model_id = "./output/lora_epoch_1"
config = PeftConfig.from_pretrained(peft_model_id)
inference_model = AutoModelForSequenceClassification.from_pretrained(config.base_model_name_or_path)
[...]

PR #2175, opened 2024-10-24
DOC: Extend modules_to_save doc with pooler example

BenjaminBossan:
See #2171

Adding this example may make it easier for users to find the issue if they encounter it.

### DIFF
--- a/docs/source/developer_guides/troubleshooting.md
+++ b/docs/source/developer_guides/troubleshooting.md
@@ -118,6 +118,12 @@ You should probably TRAIN this model on a down-stream task to be able to use it 
 
 The mentioned layers should be added to `modules_to_save` in the config to avoid the described problem.
 
+<Tip>
+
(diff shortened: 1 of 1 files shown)


docs/source/developer_guides/troubleshooting.md
```diff
@@ -118,6 +118,12 @@ You should probably TRAIN this model on a down-stream task to be able to use it
 
 The mentioned layers should be added to `modules_to_save` in the config to avoid the described problem.
 
+<Tip>
+
+As an example, when loading a model that is using the DeBERTa architecture for sequence classification, you'll see a warning that the following weights are newly initialized: `['classifier.bias', 'classifier.weight', 'pooler.dense.bias', 'pooler.dense.weight']`. From this, it follows that the `classifier` and `pooler` layers should be added to added: `modules_to_save=["classifier", "pooler"]`.
```
stevhliu, line 123:
```suggestion
As an example, when loading a model that is using the DeBERTa architecture for sequence classification, you'll see a warning that the following weights are newly initialized: `['classifier.bias', 'classifier.weight', 'pooler.dense.bias', 'pooler.dense.weight']`. From this, it follows that the `classifier` and `pooler` layers should be added to: `modules_to_save=["classifier", "pooler"]`.
```

stevhliu (approved):
LGTM, thanks for updating with an example!

PR #2177, opened 2024-10-24
Hotswap allow different alpha scalings and ranks

BenjaminBossan:
See https://github.com/huggingface/diffusers/pull/9453 for context.

Hotswapping of LoRA adapters is already implemented, but when alpha scalings or ranks differ, this triggers recompilation of the model is compiled, which is inefficient. Users can now call `prepare_model_for_compiled_hotswap` to prevent recompilation in many cases (see the doc update for caveats).


BenjaminBossan:
_not stale_

BenjaminBossan:
@sayakpaul I had some time to brush up this PR. How do we proceed? As you can see, I have added diffusers tests, but since https://github.com/huggingface/diffusers/pull/9453 is not merged, it won't work yet (i.e. there are recompiles). That PR is awaiting this PR, so there is a mutual dependency.

Also, for now `Conv2d` is [not fully implemented](https://github.com/huggingface/peft/pull/2177/files#diff-42a695c99871b8f3f55a7a95f22c3d9772b8a4bcf0941e6e87fb8eb01597781aR341-R342), but to confirm that it works, we could test with only `Linear`, right?

sayakpaul:
> Also, for now Conv2d is [not fully implemented](https://github.com/huggingface/peft/pull/2177/files#diff-42a695c99871b8f3f55a7a95f22c3d9772b8a4bcf0941e6e87fb8eb01597781aR341-R342), but to confirm that it works, we could test with only Linear, right?

That should work for now. I would add a note about this somewhere. 

> @sayakpaul I had some time to brush up this PR. How do we proceed? As you can see, I have added diffusers tests, but since https://github.com/huggingface/diffusers/pull/9453 is not merged, it won't work yet (i.e. there are recompiles).

To make progress and help unblock this PR, how about we merge this PR with sufficient testing (which I believe we already have) and mark the `diffusers` test with an xfail?

src/peft/utils/hotswap.py
```diff
@@ -87,14 +311,47 @@ def hotswap_adapter_from_state_dict(model, state_dict, adapter_name, parameter_p
...
+            continue
+
+        # Compiled models don't work with swap_tensors because there are weakrefs for the tensor. It is unclear if
+        # this workaround could not cause trouble but the tests indicate that it works.
+        if old_val.shape == new_val.shape:
             old_val.data = new_val.data
         else:
-            torch.utils.swap_tensors(old_val, new_val)
+            if old_val.dim() != 2:
+                # TODO conv2d
```
githubnemo, line 341:
```suggestion
                # TODO conv1d & conv2d
```

I don't think that there's a difference in implementation but to be sure...

githubnemo (commented):
LGTM with some minor comments

src/peft/utils/hotswap.py
```diff
@@ -13,23 +13,245 @@
...
+        # LoRA B
+        for adapter_name, lora_module in module.lora_B.items():
+            weight = lora_module.weight
+            original_rank = weight.size(1)
+
+            if original_rank == target_rank:
+                continue
+
+            if original_rank > target_rank:
+                # TODO: is this necessary or can we just continue???
```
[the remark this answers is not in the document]
BenjaminBossan: Okay, I removed the comment.

src/peft/utils/hotswap.py
```diff
@@ -13,23 +13,245 @@
...
+                    kernel_size=lora_module.kernel_size,
+                    stride=lora_module.stride,
+                    padding=lora_module.padding,
+                    bias=lora_module.bias,
+                )
+                new_layer.weight.data = padded
+            else:
+                padded = torch.zeros(weight.size(0), target_rank, device=weight.device, dtype=weight.dtype)
+                padded[:, :original_rank] = weight
+                new_layer = torch.nn.Linear(target_rank, weight.size(0), bias=lora_module.bias)
```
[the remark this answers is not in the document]
BenjaminBossan: I refactored this to split it into sub-functions for Linear and for Conv2d. That should facilitate adding new types in the future.

src/peft/utils/hotswap.py
```diff
@@ -13,23 +13,245 @@
...
+    if target_rank is not None:
+        _pad_lora_weights(model, target_rank=target_rank)
+
+    if not config:
+        return
+    if target_rank is None:
+        return
+
+    if not isinstance(config, dict):
+        config = {"dummy": config}
```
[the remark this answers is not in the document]
BenjaminBossan: done

src/peft/utils/hotswap.py
```diff
@@ -87,14 +311,47 @@ def hotswap_adapter_from_state_dict(model, state_dict, adapter_name, parameter_p
 
     # actual swapping
     for key, new_val in state_dict.items():
+        module_name = ".".join(key.split(".")[:-3])
```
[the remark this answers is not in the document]
BenjaminBossan: Added

src/peft/utils/hotswap.py
BenjaminBossan: Since I'd say we leave out conv1d for now, we don't need a TODO comment for it. I'll probably complete the conv2d part soon if we can confirm with diffusers that it works.

src/peft/utils/hotswap.py
```diff
@@ -71,122 +71,166 @@ def _convert_scalings_to_tensor(model):
...
+            f"({original_rank}). This is not possible."
+        )
+
+    out_features, in_features = weight.shape
+
+    if is_lora_A:
+        # LoRA A affects out_features
+        padded = torch.zeros(target_rank, in_features, device=weight.device, dtype=weight.dtype)
+        padded[:original_rank, :] = weight
+        new_layer = torch.nn.Linear(in_features, target_rank, bias=lora_module.bias)
```
githubnemo, line 111: Doesn't the bias need padding in this case as well?

src/peft/utils/hotswap.py
```diff
@@ -71,122 +71,166 @@ def _convert_scalings_to_tensor(model):
...
+        padded[:, :in_channels, :, :] = weight
+        new_layer = torch.nn.Conv2d(
+            target_rank,
+            out_channels,
+            kernel_size=lora_module.kernel_size,
+            stride=lora_module.stride,
+            padding=lora_module.padding,
+            bias=lora_module.bias,
+            groups=lora_module.groups,
+        )
```
githubnemo, line 189: Possibly missing bias padding. See above.

githubnemo (commented):
The changes look good. While inspecting `_get_padded_*` I found potentially missing bias handling during padding. Can you take a look?

sayakpaul:
I will take a look at the PR too. Thanks, Ben!

Sayak Paul | sayak.dev

BenjaminBossan (commented):
Thanks @githubnemo for catching the potential issue with the bias. I have addressed this, as well as an earlier comment that i missed last time.

src/peft/utils/hotswap.py
```diff
@@ -13,23 +13,302 @@
...
+    if is_lora_A:
+        # LoRA A affects out_features
+        padded = torch.zeros(target_rank, in_features, device=weight.device, dtype=weight.dtype)
+        padded[:original_rank, :] = weight
+        new_layer = torch.nn.Linear(in_features, target_rank, bias=lora_module.bias is not None)
+    else:
+        # LoRA B affects in_features
+        padded = torch.zeros(out_features, target_rank, device=weight.device, dtype=weight.dtype)
+        padded[:, :original_rank] = weight
+        new_layer = torch.nn.Linear(target_rank, out_features, bias=lora_module.bias is not None)
```
sayakpaul, line 116: I would also add a comment to note that we might revisit if we have to consider if the corresponding layers come from a quantization backend.

src/peft/utils/hotswap.py
```diff
@@ -13,23 +13,302 @@
...
+        raise ValueError(
+            "Something went wrong when trying to pad the LoRA Linear bias, the new shape should be "
+            f"{lora_module.bias.shape} but {new_layer.bias.shape} was found. Please open an issue on PEFT "
+            "(https://github.com/huggingface/peft/issues) and report this error."
+        )
+
+    new_layer.weight.data = padded
+    # Copy bias if present
+    if lora_module.bias is not None:
+        new_layer.bias.data = lora_module.bias.data
```
sayakpaul, line 135: Could potentially also leverage the `copy_()` method?

src/peft/utils/hotswap.py
```diff
@@ -13,23 +13,302 @@
...
+    """
+    Helper function that prepares the model so that it can later be compiled and then used with hot-swapping.
+
+    It is necessary to call this function on the model for hot-swapping to work if both of these are true:
+
+    - the different LoRA adapters have different ranks and/or different alpha values (i.e. scalings)
+    - you plan to torch.compile the model and want to avoid re-compilation
+
+    It is important to call this function *after* the first LoRA adapter has been loaded (i.e. the one that will be
+    swapped out) but *before* the model is compiled.
```
sayakpaul, line 264: Possibly a small code-snippet after this?

src/peft/utils/hotswap.py
```diff
@@ -13,23 +13,302 @@
...
+        target_rank (`int`, *optional*):
+            The target rank to pad the LoRA weights to. Should be the maximum rank among all LoRA adapters that will be
+            hot-swapped. If not specified, the target ranks will not be changed.
+        config (`LoraConfig` or `dict[str, LoraConfig]`, *optional*):
+            Optionally pass the `LoraConfig`s of the LoRA adapters. If passed, the rank in the configs will be updated
+            to `target_rank`.
+    """
+    is_compiled = hasattr(model, "_orig_mod")
+    if is_compiled:
+        raise ValueError("Call prepare_model_for_compiled_hotswap *before* compiling the model")
```
sayakpaul, line 283: Well, `_orig_mod` is present when we do `torch.compile()` but compilation can be also triggerd with `model.compile()`.

src/peft/utils/hotswap.py
```diff
@@ -13,23 +13,302 @@
...
+        _pad_lora_weights(model, target_rank=target_rank)
+
+    if not config:
+        return
+    if target_rank is None:
+        return
+
+    if not isinstance(config, dict):
+        # config can be either a PeftConfig, or a dict of PeftConfigs like PeftModel.peft_config
+        config = {"dummy": config}
```
sayakpaul, line 296: This config won't have any repercussions because it's never assigned to a `PeftModel`, right?

src/peft/utils/hotswap.py
```diff
@@ -87,14 +368,48 @@ def hotswap_adapter_from_state_dict(model, state_dict, adapter_name, parameter_p
...
+        # this workaround could not cause trouble but the tests indicate that it works.
+        if old_val.shape == new_val.shape:
             old_val.data = new_val.data
         else:
-            torch.utils.swap_tensors(old_val, new_val)
+            if old_val.dim() != 2:
+                # TODO conv2d
+                raise NotImplementedError
+            if old_val.shape[0] > new_val.shape[0]:
+                old_val.data.fill_(0)
```
sayakpaul, line 402: (nit): Would it make sense to have `0` assigned as a `PAD_VALUE` var and use it?

tests/test_gpu_examples.py
```diff
@@ -4172,18 +4173,19 @@ def test_hotswapping_compiled_model_does_not_trigger_recompilation(self):
         assert "__recompiles" in stderr.decode()
 
     @pytest.mark.xfail(strict=True, reason="Requires hotswap to be implemented in diffusers")
-    def test_hotswapping_compiled_diffusion_model_does_not_trigger_recompilation(self):
+    @pytest.mark.parametrize("ranks", ["7,13", "13,7"])  # the ranks of the 2 LoRAs as str
```
sayakpaul, line 4176: Should we also check for different alphas?

tests/test_initialization.py
```diff
@@ -3036,6 +3025,262 @@ def test_hotswap_extra_key_raises(self, tmp_path):
...
+
+        for name, param in model.named_parameters():
+            if "lora_A" in name and name.endswith(".weight"):
+                assert param.shape[0] == new_rank
+            elif "lora_B" in name and name.endswith(".weight"):
+                assert param.shape[1] == new_rank
+            elif "lora_A" in name and name.endswith(".bias"):
+                assert False, "LoRA A should not have a bias term"
+            elif "lora_B" in name and name.endswith(".bias"):
+                assert param.shape[0] == 10  # output shape of conv layer
```
sayakpaul, line 3282: Should we also add a test for checking if the correct config was updated?

sayakpaul (commented):
Thanks a lot, Benjamin! I have left some comments. LMK if they make sense.

tests/test_gpu_examples.py
BenjaminBossan: In `run_compiled_model_hotswap.py`, we set:

```py
    config0 = LoraConfig(init_lora_weights=False, r=rank0, lora_alpha=alpha0, target_modules=["q_proj", "v_proj"])
    config1 = LoraConfig(init_lora_weights=False, r=rank1, lora_alpha=alpha1, target_modules=["q_proj"])
```

so the alphas are indeed different.

tests/test_initialization.py
BenjaminBossan: We have `test_prepare_model_for_compiled_hotswap_scalings_update_config` above, is there something more that should be tested with regard to configs?

BenjaminBossan (commented):
Thanks for the feedback, Sayak. LMK if I addressed all your points.

docs/source/package_reference/hotswap.md
```diff
@@ -31,12 +32,42 @@ with torch.inference_mode():
...
+inputs = ...
+device = ...
+max_rank = ...  # maximum rank among all LoRA adapters that will be used
+model = AutoModelForCausalLM.from_pretrained(model_id).to(device)
+
+# load lora 0
+model = PeftModel.from_pretrained(model, <path-adapter-0>)
+# Prepare the model to allow hotswapping even if ranks/scalings of 2nd adapter differ.
+# You can skip this step if all ranks and scalings are identical.
+prepare_model_for_compiled_hotswap(model, target_rank=max_rank)
```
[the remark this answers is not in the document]
sayakpaul: Fair enough.

src/peft/utils/hotswap.py
```diff
@@ -13,23 +13,302 @@
...
+    Convert the LoRA scaling values into torch.tensors to prevent recompilation if they change.
+    """
+    for module in model.modules():
+        if not isinstance(module, LoraLayer):
+            continue
+
+        scaling = module.scaling
+        for key, val in scaling.items():
+            if isinstance(val, float):
+                scaling[key] = torch.tensor(val, device=module.weight.device)
```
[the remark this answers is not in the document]
sayakpaul: Think having that explicitly mentioned might just be more readable for maintainers. But no strong opinions.

src/peft/utils/hotswap.py
sayakpaul: I am thinking about functions like this:
https://github.com/huggingface/peft/blob/2825774d2de1c8bd0604ac685867edd79d608a9e/src/peft/tuners/lora/bnb.py#L546

src/peft/utils/hotswap.py
sayakpaul: We could check with [`torch.compiler.is_compiling()`](https://pytorch.org/docs/stable/generated/torch.compiler.is_compiling.html).

tests/test_initialization.py
sayakpaul: Perhaps alse for `alpha`s? IIUC it only checks for ranks.

sayakpaul (approved):
Thanks, Benjamin! Just last few comments. But nothing merge-blocking.

src/peft/utils/hotswap.py
sayakpaul: Fair enough. Thanks for looking into it.

src/peft/utils/hotswap.py
sayakpaul: Okay. Then maybe having a comment to mention the `.compile()` use case and that we don't know how to deal with that yet would be nice?

sayakpaul (approved):
Just a single comment. Thanks for the all hard work and iterations!

src/peft/utils/hotswap.py
BenjaminBossan: Okay, I did find a way to check for `model.compile()`, namely the `_compiled_call_impl` attribute: https://github.com/pytorch/pytorch/blob/1eba9b3aa3c43f86f4a2c807ac8e12c4a7767340/torch/nn/modules/module.py#L2991

It is private, so I used `getattr` to ensure this doesn't break easily. I also tested this and it seems to work, although I'm not sure when users should use `model.compile` instead of `torch.compile(model)`.

githubnemo (approved):

tests/test_gpu_examples.py
```diff
@@ -4136,75 +4139,218 @@ def test_prefix_tuning_multiple_devices_encoder_decoder_model(self):
...
+        """Raise an error when torch recompiles in the context.
+
+        Raises a torch._dynamo.exc.RecompileError error.
+        """
+        prev_value = torch._dynamo.config.error_on_recompile
+        torch._dynamo.config.error_on_recompile = True
+        try:
+            yield
+        finally:
+            torch._dynamo.config.error_on_recompile = prev_value
```
githubnemo, line 4173: I think you can replace this with `torch._dynamo.config.patch(error_on_recompile=True)` if you want. See https://github.com/pytorch/pytorch/blob/83c0763dda1f93c6cf552ba88260a0dc7a3ecb70/torch/utils/_config_module.py#L226

tests/test_gpu_examples.py
```diff
@@ -4136,75 +4139,218 @@ def test_prefix_tuning_multiple_devices_encoder_decoder_model(self):
...
+
+    #######
+    # LLM #
+    #######
+
+    def check_hotswap(self, do_hotswap, model_id, ranks, alpha_scalings):
```
githubnemo, line 4179: Maybe introduce a context manager that calls [`torch._dynamo.reset()`](https://github.com/pytorch/pytorch/blob/fd0cd6a08f706b7bb1dedb296217b6441e4fb9ff/torch/_dynamo/__init__.py#L85) to reset the caches will work around the need for different models?

githubnemo (commented):
I'm OK with having different models for different scenarios to work around the dynamo cache but there seems to be infrastructure in place that could be used to work around this.

I think that the recompilation detection itself utilizes the caching so disabling caching will not work either way.

BenjaminBossan:
@githubnemo Very nice finds, both of your suggestions work and are much better than the workarounds I came up with.

PR #2180, opened 2024-10-26
BUG🐛: Fixed scale related bugs in LoKr | Added rank_dropout_scale parameter

yaswanth19:
This PR adds a new parameter `rank_dropout_scale` and fixes scale related bugs in the LoKr. Please refer to the following function in Lycoris:
https://github.com/KohakuBlueleaf/LyCORIS/blob/258387f586beabfca71646a9671027f75ed34597/lycoris/modules/lokr.py#L347

### DIFF
--- a/src/peft/tuners/lokr/config.py
+++ b/src/peft/tuners/lokr/config.py
@@ -41,7 +41,7 @@ class LoKrConfig(LycorisConfig):
         decompose_factor (`int`):
             Kronecker product decomposition factor.
         rank_dropout_scale ('bool)
-            Scale the rank dropout while training.
+            Whether to scale the rank dropout while training, defaults to `False`.
         target_modules (`Optional[Union[List[str], str]]`):
             The names of the modules to apply the adapter to. If this is specified, only the modules with the specified
             names will be replaced. When passing a string, a regex match will be performed. When passing a list of
--- a/src/peft/tuners/lokr/config.py
+++ b/src/peft/tuners/lokr/config.py
@@ -55,8 +55,8 @@ class LoKrConfig(LycorisConfig):
             When passing a list of strings, either an exact match will be performed or it is checked if the name of the
             module ends with any of the passed strings.
         init_weights (`bool`):
-            Whether to perform initialization of adapter weights. This defaults to `True`, passing `False` is
-            discouraged.
+            Whether to perform initialization of adapter weights. This defaults to `True`,Use "lycoris" to initialize
+            weights in the style of the LYCORIS repository. Passing `False` is discouraged.".
         layers_to_transform (`Union[List[int], int]`):
             The layer indices to transform. If a list of ints is passed, it will apply the adapter to the layer indices
             that are specified in this list. If a single integer is passed, it will apply the transformations on the
@@ -106,12 +106,12 @@ class LoKrConfig(LycorisConfig):
         default=None,
         metadata={"help": "List of module names or regex expression of the module names to exclude from LoKr."},
     )
-    init_weights: bool = field(
+    init_weights: Union[bool, str] = field(
         default=True,
         metadata={
             "help": (
-                "Whether to initialize the weights of the LoKr layers with their default initialization. Don't change "
-                "this setting, except if you know exactly what you're doing."
+                "Whether to initialize the weights of the LoKr layers with their default initialization. Can be True, False or 'lycoris'. Default is True"
+                "Don't change this setting, except if you know exactly what you're doing."
             ),
         },
     )
--- a/src/peft/tuners/lokr/config.py
+++ b/src/peft/tuners/lokr/config.py
@@ -55,7 +55,7 @@ class LoKrConfig(LycorisConfig):
             When passing a list of strings, either an exact match will be performed or it is checked if the name of the
             module ends with any of the passed strings.
         init_weights (`bool`):
-            Whether to perform initialization of adapter weights. This defaults to `True`,Use "lycoris" to initialize
+            Whether to perform initialization of adapter weights. This defaults to `True`.Use "lycoris" to initialize
             weights in the style of the LYCORIS repository. Passing `False` is discouraged.".
         layers_to_transform (`Union[List[int], int]`):
             The layer indices to transform. If a list of ints is passed, it will apply the adapter to the layer indices
--- a/src/peft/tuners/lokr/config.py
+++ b/src/peft/tuners/lokr/config.py
@@ -40,6 +40,8 @@ class LoKrConfig(LycorisConfig):
             Perform rank decomposition of left kronecker product matrix.
         decompose_factor (`int`):
             Kronecker product decomposition factor.
+        rank_dropout_scale ('bool)
+            Scale the rank dropout while training.
         target_modules (`Optional[Union[List[str], str]]`):
             The names of the modules to apply the adapter to. If this is specified, only the modules with the specified
             names will be replaced. When passing a string, a regex match will be performed. When passing a list of
@@ -91,6 +93,7 @@ class LoKrConfig(LycorisConfig):
         metadata={"help": "Perform rank decomposition of left kronecker product matrix."},
     )
     decompose_factor: int = field(default=-1, metadata={"help": "Kronecker product decomposition factor."})
+    rank_dropout_scale: bool = field(default=False, metadata={"help": "Rank dropout scale"})
     target_modules: Optional[Union[list[str], str]] = field(
         default=None,
         metadata={
--- a/src/peft/tuners/lokr/config.py
+++ b/src/peft/tuners/lokr/config.py
@@ -14,7 +14,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import Optional, Union
+from typing import Literal, Optional, Union
 
 from peft.tuners.lycoris_utils import LycorisConfig
 from peft.utils import PeftType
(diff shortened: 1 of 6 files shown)


yaswanth19:
@BenjaminBossan  Please review the minor corrections to the LoKr. The default initialization of W1 matrix is zeros and I am not sure to change it or not. Also shall I rewrite this LoKr implementation to remove lycoris dependency as I can use most of the LycorisLokr implementation code.

src/peft/tuners/lokr/config.py
```diff
@@ -40,6 +40,8 @@ class LoKrConfig(LycorisConfig):
             Perform rank decomposition of left kronecker product matrix.
         decompose_factor (`int`):
             Kronecker product decomposition factor.
+        rank_dropout_scale ('bool)
+            Scale the rank dropout while training.
```
BenjaminBossan, line 44:
```suggestion
            Whether to scale the rank dropout while training, defaults to `False`.
```

src/peft/tuners/lokr/layer.py
```diff
@@ -156,10 +156,11 @@ def update_layer(
             raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
 
         self.r[adapter_name] = r
-        self.alpha[adapter_name] = alpha
-        self.scaling[adapter_name] = alpha / r
+        self.alpha[adapter_name] = alpha or r
```
BenjaminBossan, line 159: Could you explain why this is needed?

src/peft/tuners/lokr/model.py
```diff
@@ -91,7 +92,7 @@ class LoKrModel(LycorisTuner):
 
     def _create_and_replace(
         self,
-        config: LycorisConfig,
+        config: LycorisConfig | LoKrConfig,
```
BenjaminBossan, line 95: `LoKrConfig` is a subclass of `LycorisConfig`, so this should not be necessary. Do you see issues caused by this type annotation?

BenjaminBossan (changes requested):
Thanks a lot for working on these bug fixes for LoKr, this is really appreciated.

I have some small comments, but overall this looks good.

> The default initialization of W1 matrix is zeros and I am not sure to change it or not.

I'd say leave it as is. What I think could be useful is to add a new option to the config that allows users to use the lycoris way of layer initialization, then they have a choice of which one they want. That could be done in a separate PR.

> Also shall I rewrite this LoKr implementation to remove lycoris dependency as I can use most of the LycorisLokr implementation code.

I'm not sure what you mean by this, could you please elaborate.

yaswanth19:
>I'm not sure what you mean by this, could you please elaborate.

I meant to say, we still have lycoris_utils dependency. Also the codebase is kind of old for LoKr. Since I have updated the codebase of LoKr w.r.p latest PEFT standards in LycorisLoKr #2133, It would be simple copy paste and we don't need a seperate implementation

src/peft/tuners/lokr/layer.py
yaswanth19: It is used in case when alpha is 0. So we set to `r` hence no scaling would be performed

src/peft/tuners/lokr/model.py
yaswanth19: Added just for better type annotation; but yeah since `LoKrConfig` is a subclass of `LycorisConfig`, this change would not be needed.

yaswanth19:
Hmm makes sense given LoKr/LoHa is used less frequently. Then I will make the suggested changes and also will once look at LoHa if the same bugs are present in that too or not. The big question is what to do with LycorisLoKr #2133 :sweat_smile: . Shall I close it since we are using very little of `lycoris` package to have the feature what we thought and rewriting the LoKr also reaps very little benefit.

yaswanth19:
@BenjaminBossan Done with suggested changes :white_check_mark: 

>I'd say leave it as is. What I think could be useful is to add a new option to the config that allows users to use the lycoris way of layer initialization, then they have a choice of which one they want. That could be done in a separate PR.

A separate flag is not needed, they can pass `init_weight=False` to initialize with random weights instead of zero weights.

BenjaminBossan:
> A separate flag is not needed, they can pass `init_weight=False` to initialize with random weights instead of zero weights.

What I meant is that we found that lycoris initializes layers a bit differently than PEFT, as I mentioned on the other PR:

> In PEFT, at the start, we initialize w1 to zeros and w2_a and w2_b randomly. LyCORIS, however, initializes w2_b to zeros and w1 and w2_a randomly.

We could have an init option like `init_weights = {True, False, "lycoris"}` and if the latter is chosen, we initialize the same way as lycoris does.

src/peft/tuners/lokr/layer.py
BenjaminBossan: Hmm, not sure about this one. Could the user not set `alpha = r` directly? Or is there some precedence that 0 implies that `scaling=1`?

src/peft/tuners/lokr/config.py
```diff
@@ -53,8 +55,8 @@ class LoKrConfig(LycorisConfig):
             When passing a list of strings, either an exact match will be performed or it is checked if the name of the
             module ends with any of the passed strings.
         init_weights (`bool`):
-            Whether to perform initialization of adapter weights. This defaults to `True`, passing `False` is
-            discouraged.
+            Whether to perform initialization of adapter weights. This defaults to `True`,Use "lycoris" to initialize
```
BenjaminBossan, line 58:
```suggestion
            Whether to perform initialization of adapter weights. This defaults to `True`. Use "lycoris" to initialize
```

src/peft/tuners/lokr/layer.py
```diff
@@ -126,6 +126,22 @@ def reset_adapter_parameters_random(self, adapter_name: str):
         if adapter_name in self.lokr_t2:
             nn.init.kaiming_uniform_(self.lokr_t2[adapter_name], a=math.sqrt(5))
 
+    def reset_adapter_parameters_lycoris_way(self, adapter_name):
```
BenjaminBossan, line 129: Could you please add a short comment about the "lycoris way", as it may not be obvious to everyone.

BenjaminBossan (changes requested):
Thanks so much for the updates. This PR is almost good to go. I found some nits that I commented on. Apart from that, since a new initialization scheme was added, let's add a test for this.

I think the best place would be in `test_initialization.py`. Let's create a test class similar to what we have for [LoRA](https://github.com/huggingface/peft/blob/ff6dd9ed7f42f2c7e312ea4e7c13539d2c312944/tests/test_initialization.py#L55). The tests can be very simple: One test for `init_weight=True` that checks that we get the same output as from the base model. Same for `init_weight="lycoris"`. Finally, a test for `init_weight=False` that checks that the output is different from the base model's output. No need for any stats on the outputs. LMK if you have questions.

yaswanth19:
@BenjaminBossan Added the testcases, please review it.

src/peft/tuners/lokr/config.py
```diff
@@ -53,8 +55,8 @@ class LoKrConfig(LycorisConfig):
             When passing a list of strings, either an exact match will be performed or it is checked if the name of the
             module ends with any of the passed strings.
         init_weights (`bool`):
-            Whether to perform initialization of adapter weights. This defaults to `True`, passing `False` is
-            discouraged.
+            Whether to perform initialization of adapter weights. This defaults to `True`.Use "lycoris" to initialize
```
BenjaminBossan, line 58:
```suggestion
            Whether to perform initialization of adapter weights. This defaults to `True`. Use "lycoris" to initialize
```

src/peft/tuners/lokr/config.py
```diff
@@ -53,8 +55,8 @@ class LoKrConfig(LycorisConfig):
...
+            weights in the style of the LYCORIS repository. Passing `False` is discouraged.".
```
BenjaminBossan, line 59:
```suggestion
            weights in the style of the LYCORIS repository. Passing `False` is discouraged.
```

src/peft/tuners/lokr/config.py
```diff
@@ -103,12 +106,12 @@ class LoKrConfig(LycorisConfig):
         default=None,
         metadata={"help": "List of module names or regex expression of the module names to exclude from LoKr."},
     )
-    init_weights: bool = field(
+    init_weights: Union[bool, str] = field(
```
BenjaminBossan, line 109:
```suggestion
    init_weights: Union[bool, Literal["lycoris"]] = field(
```

a bit more precise type annotation.

src/peft/tuners/lokr/config.py
```diff
@@ -103,12 +106,12 @@ class LoKrConfig(LycorisConfig):
...
         default=True,
         metadata={
             "help": (
-                "Whether to initialize the weights of the LoKr layers with their default initialization. Don't change "
-                "this setting, except if you know exactly what you're doing."
+                "Whether to initialize the weights of the LoKr layers with their default initialization. Can be True, False or 'lycoris'. Default is True"
+                "Don't change this setting, except if you know exactly what you're doing."
```
BenjaminBossan, line 114:
```suggestion
                "Don't change this setting to False, except if you know exactly what you're doing."
```

src/peft/tuners/lokr/config.py
```diff
@@ -103,12 +106,12 @@ class LoKrConfig(LycorisConfig):
...
         metadata={"help": "List of module names or regex expression of the module names to exclude from LoKr."},
     )
-    init_weights: bool = field(
+    init_weights: Union[bool, str] = field(
         default=True,
         metadata={
             "help": (
-                "Whether to initialize the weights of the LoKr layers with their default initialization. Don't change "
-                "this setting, except if you know exactly what you're doing."
+                "Whether to initialize the weights of the LoKr layers with their default initialization. Can be True, False or 'lycoris'. Default is True"
```
BenjaminBossan, line 113: Line is too long

BenjaminBossan (changes requested):
Fantastic, thanks for the updates. Tests look good to. There are only a handful of small issues left, please take a look.

yaswanth19:
@BenjaminBossan  The doc string changes are Addressed.

BenjaminBossan (approved):
Thanks for addressing my last points. This PR LGTM, thanks a lot for your amazing work on LoKr.

I'd like to leave this PR open for a little bit to see if there is more feedback from other folks. If not, I'll probably merge it at the start of next week.

BenjaminBossan:
Again, thanks so much @yaswanth19 for implementing these fixes and enhancements for LoKr. Since there hasn't been any further feedback, I decided to merge now.

LMK if you plan on working on further fixes for LoKr or LoHa. Ideally, we could have them all in the same PEFT release, as users may have to retrain their models, depending on the type of fix.

yaswanth19:
@BenjaminBossan I don't have any plans to work this immediately, probably it would be best if somebody else can pick this so to have the changes in the same release

BenjaminBossan:
All right, thanks for letting me know.

PR #2183, opened 2024-10-28
DOC: fix broken link in the README of loftq

dennis2030:
# Why This PR?

The README file in loftq contains broken link, which might confuse the reader.

# How to fix?

Simply update the link name to the correct section.


BenjaminBossan (approved):
Thanks for fixing the link.

Issue #2134, opened 2024-10-06
PEFT doesn't inject virtual tokens into generate forward pass


Kami-chanw:
### System Info

- `transformers` version: 4.46.0.dev0
- Platform: Linux-5.4.0-148-generic-x86_64-with-glibc2.31
- Python version: 3.9.19
- Huggingface_hub version: 0.24.0
- Safetensors version: 0.4.3
- Accelerate version: 0.33.0
- Accelerate config:    not found
- PyTorch version (GPU?): 2.4.1+cu121 (True)
- Tensorflow version (GPU?): not installed (NA)
- Flax version (CPU?/GPU?/TPU?): not installed (NA)
- Jax version: not installed
- JaxLib version: not installed
- Using distributed or parallel set-up in script?: <fill in>
- Using GPU in script?: <fill in>
- GPU type: NVIDIA RTX A6000

### Who can help?

@BenjaminBossan @sayakpaul 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [X] My own task or dataset (give details below)

### Reproduction

I met the problem by the following code:
```python
import peft
from transformers import IdeficsForVisionText2Text, AutoProcessor
import sys
import torch

sys.path.insert(0, "..")
import config  # my own file

device = torch.device("cuda:3")
model = IdeficsForVisionText2Text.from_pretrained(
    config.idefics_9b_path, torch_dtype=torch.float16
).to(device)
processor = AutoProcessor.from_pretrained(
    config.idefics_9b_path, torch_dtype=torch.float16
)
model = peft.get_peft_model(
    model,
    peft.PrefixTuningConfig(
        peft_type="PREFIX_TUNING",
        task_type="CAUSAL_LM",
        num_virtual_tokens=2,
        token_dim=4096,
        num_transformer_submodules=1,
        num_attention_heads=32,
        num_layers=32,
        encoder_hidden_size=768,
    ),
    mixed=False,
)
inputs = processor(["hello"]).to(device)
model.eval()
model.generate(**inputs)
```
When I add `print(past_key_values)` in transformers side, I got `DynamicCache()`, which means the virtual tokens weren't injected to forward pass.

### Expected behavior

It should get a cache with length of `num_virtual_tokens`.

BenjaminBossan:
Could you please clarify where you check `past_key_values` and what you would expect there?

[...]

PR #2189, opened 2024-10-30
FIX: Prefix tuning with model on multiple devices

BenjaminBossan:
See #2134

After introducing the usage of `DynamicCache` for prefix tuning, a bug could now occur if the model is dispatched to different devices. This is because we need to move the key and value cache for each layer to that layer's respective device.

The new code mostly consists of code copied from transformers to be consistent with how transformers solves this.

Note that this only works if the `hf_device_map` attribute is set on the model.

### DIFF
--- a/src/peft/utils/integrations.py
+++ b/src/peft/utils/integrations.py
@@ -120,3 +120,49 @@ def get_bnb_param_type(param: torch.nn.Parameter) -> Literal[False, "4bit", "8bi
     if param.__class__.__name__ == "Int8Params":
         return "8bit"
     return False
+
+
+# adapted from:
+# https://github.com/huggingface/transformers/blob/eab6c491d439e83d5e31c660df6f7e36592eb0a2/src/transformers/generation/utils.py#L1617-L1643
+def get_layer_device_map(model):
+    """
+    Derive the device map for the layers of the model.
+    """
+    main_device = [d for d in model.hf_device_map.values() if d not in ["cpu", "disk"]][0]
+
+    execution_device_map = {
+        name: main_device if device in ["cpu", "disk"] else device for name, device in model.hf_device_map.items()
+    }
+
+    if execution_device_map is None:
+        return None
+
+    if len(execution_device_map) == 1 and "" in execution_device_map:
+        return {idx: execution_device_map[""] for idx in range(model.config.num_hidden_layers)}
+
+    layer_device_map = {}
+    for layer in execution_device_map:
+        for idx in range(model.config.num_hidden_layers):
+            if f".{idx}." in f"{layer}.":
+                layer_device_map[idx] = execution_device_map[layer]
+                break
+    for idx in range(model.config.num_hidden_layers):
+        if idx not in layer_device_map:
+            raise RuntimeError(f"layer {idx} has not been mapped to a device.")
+    return layer_device_map
+
+
+# adapted from:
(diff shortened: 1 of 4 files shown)


src/peft/utils/integrations.py
```diff
@@ -120,3 +120,49 @@ def get_bnb_param_type(param: torch.nn.Parameter) -> Literal[False, "4bit", "8bi
...
+    return layer_device_map
+
+
+# adapted from:
+# https://github.com/huggingface/transformers/blob/eab6c491d439e83d5e31c660df6f7e36592eb0a2/src/transformers/cache_utils.py#L1159-L1179
+def map_cache_to_layer_device_map(model, cache) -> None:
+    """
+    Ensure that the key and value cache of the model are on the same device as their corresponding layers.
+    """
+    if not (isinstance(cache, transformers.Cache) and hasattr(model, "hf_device_map")):
```
zucchini-nlp, line 161: prob `or not hastatt(model, "hf_device_map")` otherwise we don't skip cases when the cache is tuple and model has no device map

Also I'm wondering if this didn't fail for tuple because accelerate handles device allocations for tensors, but can't handle for object? Since in tuple format we also don't do any device mapping usually

src/peft/utils/integrations.py
```diff
@@ -120,3 +120,49 @@ def get_bnb_param_type(param: torch.nn.Parameter) -> Literal[False, "4bit", "8bi
...
+        return
+
+    layer_device_map = get_layer_device_map(model)
+    for idx in range(model.config.num_hidden_layers):
+        layer_device = layer_device_map[idx]
+        cache.key_cache[idx] = cache.key_cache[idx].to(layer_device)
+        cache.value_cache[idx] = cache.value_cache[idx].to(layer_device)
```
zucchini-nlp, line 168: This might fail for encoder-decoder cache, can we also try with T5 model in multi-gpu environment to see if it works?

src/peft/utils/integrations.py
BenjaminBossan: Hmm, maybe I misunderstand, but right now we have "not (A and B)", which is the same as "not A or not B".

src/peft/utils/integrations.py
BenjaminBossan: I tried and it didn't work because the `get_layer_device_map` function would fail. I tried to check why it doesn't fail when the function is called in transformers. But for T5, it is never reached. Honestly, I couldn't figure out why it's different for encoder-decoder. My solution for now is to call `map_cache_to_layer_device_map` only when `peft_config.num_transformer_submodules == 1` and leave encoder-decoder untouched for now.

src/peft/utils/integrations.py
zucchini-nlp: oh right, I didn't see the brackets around I guess XD

src/peft/utils/integrations.py
zucchini-nlp: yes, sounds good to me, and we can add multi-gpu support for encoder-decoder as more models get converted to new cache format

zucchini-nlp (approved):
LGTM! Thanks for adding this

BenjaminBossan:
@zucchini-nlp I did manage to make this work with T5, the fix was actually quite simple. Could you please review again?

zucchini-nlp (approved):
Awesome that it worked for T5! Thanks!

PR #2191, opened 2024-10-31
FIX: Check for prefix tuning + gradient checkpointing fails

BenjaminBossan:
See #869

Since transformers is moving to the new cache implementation, we had to change prefix tuning to use this too. However, caching does not work with gradient checkpointing. Therefore, this currently runs into an error about size mismatches.

Now, PEFT checks for gradient checkpointing and raises a helpful error.

Related to this change, I relaxed the `_test_training_gradient_checkpointing` test, which would skip all prompt learning methods, when in fact only prefix tuning needs to be skipped.


zucchini-nlp (approved):
Thanks!

PR #2197, opened 2024-11-03
Dora_datacollector_updated

shirinyamani:
This PR just updates the datacollector for fine-tuning using Dora


BenjaminBossan (approved):
Thanks for fixing this.

I think when creating the PR, something went wrong, as it shows 39 commits, even though there is only one change. But it's no big deal, I can still merge.

shirinyamani:
Hi Benjamin @BenjaminBossan 
Thanks for all the good you do. I just added the implementation [SnapKV Cache paper](https://arxiv.org/abs/2404.14469) to [cache_utils.py](https://github.com/huggingface/transformers/blob/main/src/transformers/cache_utils.py) file.  To reflect the SnapKV approach, I added the implementation on [llama_modeling](https://github.com/huggingface/transformers/tree/main/src/transformers/models/llama) under [llama_snapkv.py ](src/transformers/models/llama/llama_snapkv.py) for the user to see what changes has to be applied to flash_attention2 to reflect the snap_kv, however, I'm not entirely sure tht there is the best location for this implementation. I’m confident that the main SnapKV logic belongs in cache_utils.py, but I would really appreciate your insights on whether the llama_snapkv.py example is well-placed or if it should be integrated differently.
Much appreciated!

BenjaminBossan:
@shirinyamani Thanks for working on this. I'm not a transformers maintainer, so I can't really tell you where best to implement this feature. It's best if you reach out to the transformers folks for this. What you could also do is create a draft PR with your implementation the way that you suggested, I'm sure they'll let you know if you have to change something.

PR #2201, opened 2024-11-06
Add new feature of SafeLoRA

chiayi-hsu:
The pull request was closed due to syncing with the latest version of PEFT, so I have requested the pull request again.
I have made all the necessary changes based on our previous conversations in this version.

If there are any issues, please let me know. 

Thank you.

### DIFF
--- a/examples/safelora/README.md
+++ b/examples/safelora/README.md
@@ -1,0 +1,24 @@
+#Safe LoRA 
+
+The official code of Safe LoRA:the Silver Lining of Reducing Safety Risks when Fine-tuning Large Language Models
+
+
+## Quick Start
+
+Please import the SafeLoraConfig and apply_safelora first.
+Then, fill in the paths for the base, aligned, and PEFT models according to your needs. There are two types of `select_layers_type`: `threshold` and `number`. The `threshold` type will determine how many layers will be projected based on the value you set. The `number` type directly specifies the number of projected layers.
+
+```python
+
+from peft.utils.safelora import SafeLoraConfig, apply_safelora
+
+config = SafeLoraConfig(base_model_path='../LLM_Models/llama-2-7b-hf/',\
+                            aligned_model_path='../LLM_Models/llama-2-7b-chat-fp16/',
+                            peft_model_path = '../finetuneLLM/finetuned_models/samsumBad-7b-fp16-peft-seed-42',
+                            devices='cuda',
+                            select_layers_type='threshold',
+                            save_weights=True)
+
+final_lora_weight = apply_safelora(config)
+
+```
--- a/examples/safelora/README.md
+++ b/examples/safelora/README.md
@@ -1,4 +1,4 @@
-#Safe LoRA 
+# Safe LoRA 
 
 The official code of Safe LoRA:the Silver Lining of Reducing Safety Risks when Fine-tuning Large Language Models
 
--- a/examples/safelora/safelora_inference.py
+++ b/examples/safelora/safelora_inference.py
@@ -1,0 +1,11 @@
+from peft import PeftConfig
+from peft.utils.safelora import SafeLoraConfig, apply_safelora
+
+config = SafeLoraConfig(base_model_path='../LLM_Models/llama-2-7b-hf/',\
+                            aligned_model_path='../LLM_Models/llama-2-7b-chat-fp16/',
+                            peft_model_path = '../finetuneLLM/finetuned_models/samsumBad-7b-fp16-peft-seed-42',
+                            devices='cuda',
+                            select_layers_type='threshold',
+                            save_weights=True)
+
+final_lora_weight = apply_safelora(config)
--- a/src/peft/utils/safelora.py
+++ b/src/peft/utils/safelora.py
@@ -176,7 +176,7 @@ def project_weights(configs, peft_weights, v):
     return peft_weights, cos_total
 
 
-def apply_safelora(configs):
+def apply_safelora(safelora_config: SafeLoraConfig):
     """
 
     The official code of Safe LoRA: The Silver Lining of Reducing Safety Risks when Finetuning Large Language Models: https://arxiv.org/abs/2405.16833
@@ -203,28 +203,28 @@ def apply_safelora(configs):
 
     """
 
-    peft_config = PeftConfig.from_pretrained(configs.peft_model_path)
+    peft_config = PeftConfig.from_pretrained(safelora_config.peft_model_path)
 
     projected_matrix = get_aligned_matrix(
-        configs.base_model_path, configs.aligned_model_path, configs.devices, peft_config, configs
+        safelora_config.base_model_path, safelora_config.aligned_model_path, safelora_config.devices, peft_config, safelora_config
     )
 
     with safe_open(
-        f"{os.path.join(configs.peft_model_path, 'adapter_model.safetensors')}", framework="pt", device=configs.devices
+        f"{os.path.join(safelora_config.peft_model_path, 'adapter_model.safetensors')}", framework="pt", device=safelora_config.devices
     ) as f:
-        if configs.devices == "cpu":
-            peft_weights = {name: f.get_tensor(name).to(configs.dtype) for name in f.keys()}
+        if (safelora_config.devices).lower() == "cpu":
+            peft_weights = {name: f.get_tensor(name).to(safelora_config.dtype) for name in f.keys()}
         else:
-            peft_weights = {name: f.get_tensor(name).to(configs.dtype) for name in f.keys()}
-    if configs.select_layers_type == "threshold":
-        final_weights, _ = project_weights(configs, peft_weights, projected_matrix)
-    elif configs.select_layers_type == "number":
-        _, cos = project_weights(configs, peft_weights, projected_matrix)
-        thrs = torch.sort(torch.Tensor(cos))[0][: configs.num_proj_layers][-1]
-        configs.threshold = thrs
-        final_weights, _ = project_weights(configs, peft_weights, projected_matrix)
+            peft_weights = {name: f.get_tensor(name).to(safelora_config.dtype) for name in f.keys()}
+    if safelora_config.select_layers_type == "threshold":
+        final_weights, _ = project_weights(safelora_config, peft_weights, projected_matrix)
+    elif safelora_config.select_layers_type == "number":
+        _, cos = project_weights(safelora_config, peft_weights, projected_matrix)
+        thrs = torch.sort(torch.Tensor(cos))[0][: safelora_config.num_proj_layers][-1]
+        safelora_config.threshold = thrs
+        final_weights, _ = project_weights(safelora_config, peft_weights, projected_matrix)
 
-    if configs.save_weights:
-        save_file(final_weights, f"{os.path.join(configs.peft_model_path, 'adapter_model.safetensors')}")
+    if safelora_config.save_weights:
+        save_file(final_weights, f"{os.path.join(safelora_config.peft_model_path, 'adapter_model.safetensors')}")
 
     return final_weights
--- a/src/peft/utils/safelora.py
+++ b/src/peft/utils/safelora.py
@@ -1,0 +1,187 @@
+# Copyright 2024-present the HuggingFace Inc. team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# Reference paper: https://arxiv.org/abs/2405.16833
+
+
+import copy
+import json
+import os
+from dataclasses import dataclass, field
+
+import torch
+from huggingface_hub import snapshot_download
+from loftq_utils import _SafetensorLoader as SafetensorLoader
+from safetensors import SafetensorError, safe_open
+from safetensors.torch import save_file
+from transformers.utils import cached_file
+from transformers.utils.hub import get_checkpoint_shard_files
+
+
+@dataclass
+class SafeLoraConfig:
+    """
+    This is the configuration class to store the configuration of a safeLoRA.
+    """
+
+    base_model_path: str = field(
+        default=None,
+        metadata={"help": "The path of the base model for obtaining the aligned matrix"},
+    )
+
+    aligned_model_path: str = field(
+        default=None,
+        metadata={"help": "The path of the aligned model for obtaining the aligned matrix"},
+    )
+
+    peft_model_path: str = field(
+        default=None,
+        metadata={"help": "The path of the LoRA wieghts and configs."},
+    )
+
+    select_layers_type: str = field(
+        default="number",
+        metadata={"help": "How to select projection layers? options: [threshold, number]"},
+    )
+
+    threshold: float = field(
+        default=0.5,
+        metadata={"help": "The threshold of cosine similarity."},
+    )
+
+    num_proj_layers: int = field(
+        default=10,
+        metadata={"help": "The number of projected layers."},
+    )
+
+    devices: str = field(
+        default="cuda",
+        metadata={"help": "Devices are used in SafeLoRA. (gpu or cpu)"},
+    )
+
+    save_weights: bool = field(
+        default=True,
+        metadata={"help": "Replacing and saving SafeLoRA weights to the original LoRA file."},
+    )
+    local_files_only: bool = field(
+        default = False,
+        metadata={'help': "Using for snapshot_download"},
+        )
+
+    def __post_init__(self):
+        if self.base_model_path is None:
+            raise ValueError("base_model_path cannot be None.")
+        if self.aligned_model_path is None:
+            raise ValueError("aligned_model_path cannot be None.")
+        if self.peft_model_path is None:
+            raise ValueError("peft_model_path cannot be None.")
+
+
+def get_aligned_matrix(base_model_path, aligned_model_path, devices, peft_config):
+    """
+    Get projected matrix by following the config (target_modules) from the peft model.
+    The dimensions between the base model's weights and the aligned model's weights should be the same.
+    """
+    sl_align = SafetensorLoader(aligned_model_path)
+    sl_base = SafetensorLoader(base_model_path)
+
+    base_model_parameters = [name for name in sl_base.weight_map.keys() if any(v in name for v in list(peft_config["target_modules"]))]
+    align_model_parameters = [
+        name for name in sl_align.weight_map.keys() if any(v in name for v in list(peft_config["target_modules"]))
+    ]
+    safety_vector = []
+    for name_base, name_align in zip(base_model_parameters, align_model_parameters):
+        assert (
+            sl_base.get_tensor(name_base).shape == sl_align.get_tensor(name_align).shape
+        ), "The dimensions of the base model's weight should be the same with the aligned model's weight."
+        vec = sl_base.get_tensor(name_base) - sl_align.get_tensor(name_align)
+        vec = vec.to(devices)
+        if devices == "cpu":
+            vec = vec.to(torch.float32)
+        vec = torch.mm(vec, vec.t()) / torch.norm(vec)
+        safety_vector.append((vec).detach().cpu())
+    return safety_vector
+
+
+def project_weights(configs, peft_weights, v):
+    ori_peft_weights = copy.deepcopy(peft_weights)
+    vars_names_LoRA_A = [name for name in peft_weights.keys() if "lora_A" in name]
+    vars_names_LoRA_B = [name for name in peft_weights.keys() if "lora_B" in name]
+    num_projected_layers = 0
+    dis = []
+    cos_total = []
+    for idx, (name_A, name_B) in enumerate(zip(vars_names_LoRA_A, vars_names_LoRA_B)):
+        A = ori_peft_weights[name_A]
+        if configs.devices != "cpu":
+            P = v[idx].to(torch.bfloat16).to(configs.devices)
+        else:
+            P = v[idx].to("cpu")
+        W = torch.mm(P, ori_peft_weights[name_B])
+        fW = torch.mm(W, A)
+        ori = torch.mm(ori_peft_weights[name_B], A)
+        cos = torch.round(torch.nn.functional.cosine_similarity(fW.reshape(1, -1), ori.reshape(1, -1))*10**5) / 10**5
+        cos_total.append(cos.item())
+        if cos <= configs.threshold:
+            num_projected_layers += 1
+            peft_weights[name_B] = W
+        else:
+            peft_weights[name_B] = ori_peft_weights[name_B]
+
+        dist = 1 / (1 + torch.norm(peft_weights[name_B].reshape(1, -1) - W.reshape(1, -1)))
+
+        dis.append(dist.item())
+    return peft_weights, cos_total
+
+
+def apply_safelora(configs):
+    """
+    ===================================================
+    An example of how to use apply_safelora() function
+    ===================================================
+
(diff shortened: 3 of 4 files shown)


examples/safelora/README.md
```diff
@@ -0,0 +1,24 @@
+# Safe LoRA 
+
+The official code of Safe LoRA:the Silver Lining of Reducing Safety Risks when Fine-tuning Large Language Models
+
```
BenjaminBossan, line 4:
```suggestion
The official code of Safe LoRA: The Silver Lining of Reducing Safety Risks when Fine-tuning Large Language Models
```

examples/safelora/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Quick Start
+
+Please import the SafeLoraConfig and apply_safelora first.
```
BenjaminBossan, line 8:
```suggestion
Please import the `SafeLoraConfig` and `apply_safelora` first.
```

examples/safelora/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+config = SafeLoraConfig(base_model_path='../LLM_Models/llama-2-7b-hf/',\
+                            aligned_model_path='../LLM_Models/llama-2-7b-chat-fp16/',
+                            peft_model_path = '../finetuneLLM/finetuned_models/samsumBad-7b-fp16-peft-seed-42',
+                            devices='cuda',
+                            select_layers_type='threshold',
+                            save_weights=True)
+
+final_lora_weight = apply_safelora(config)
+
```
BenjaminBossan, line 23: Can we add a bit more to the example. For instance, how to save and load these weights?

examples/safelora/README.md
```diff
@@ -0,0 +1,24 @@
...
+Please import the SafeLoraConfig and apply_safelora first.
+Then, fill in the paths for the base, aligned, and PEFT models according to your needs. There are two types of `select_layers_type`: `threshold` and `number`. The `threshold` type will determine how many layers will be projected based on the value you set. The `number` type directly specifies the number of projected layers.
+
+```python
+
+from peft.utils.safelora import SafeLoraConfig, apply_safelora
+
+config = SafeLoraConfig(base_model_path='../LLM_Models/llama-2-7b-hf/',\
+                            aligned_model_path='../LLM_Models/llama-2-7b-chat-fp16/',
```
BenjaminBossan, line 16: Let's use the HF model ids for these two.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,229 @@
...
+        safelora_config.base_model_path, safelora_config.aligned_model_path, safelora_config.devices, peft_config, safelora_config
+    )
+
+    with safe_open(
+        f"{os.path.join(safelora_config.peft_model_path, 'adapter_model.safetensors')}", framework="pt", device=safelora_config.devices
+    ) as f:
+        if (safelora_config.devices).lower() == "cpu":
+            peft_weights = {name: f.get_tensor(name).to(safelora_config.dtype) for name in f.keys()}
+        else:
+            peft_weights = {name: f.get_tensor(name).to(safelora_config.dtype) for name in f.keys()}
```
BenjaminBossan, line 217: These 2 lines are identical

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,229 @@
...
+    ori_peft_weights = copy.deepcopy(peft_weights)
+    vars_names_LoRA_A = [name for name in peft_weights.keys() if "lora_A" in name]
+    vars_names_LoRA_B = [name for name in peft_weights.keys() if "lora_B" in name]
+    num_projected_layers = 0
+    dis = []
+    cos_total = []
+    for idx, (name_A, name_B) in enumerate(zip(vars_names_LoRA_A, vars_names_LoRA_B)):
+        A = ori_peft_weights[name_A]
+        if configs.devices != "cpu":
+            P = v[idx].to(torch.bfloat16).to(configs.devices)
```
BenjaminBossan, line 159: Should we really hard-code bf16 here? Let's make it configurable. Also, I think with recent pytorch versions, bf16 works on CPU.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,229 @@
...
+        else:
+            P = v[idx].to("cpu")
+        W = torch.mm(P, ori_peft_weights[name_B])
+        fW = torch.mm(W, A)
+        ori = torch.mm(ori_peft_weights[name_B], A)
+        cos = torch.round(torch.nn.functional.cosine_similarity(fW.reshape(1, -1), ori.reshape(1, -1)) * 10**5) / 10**5
```
BenjaminBossan, line 165: `torch.round` has a `decimals` argument, maybe use that instead of *10**5 (do we need the rounding at all?).

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,229 @@
...
+def get_aligned_matrix(base_model_path, aligned_model_path, devices, peft_config, configs):
+    """
+    Get projected matrix by following the config (target_modules) from the peft model.
+    The dimensions between the base model's weights and the aligned model's weights should be the same.
+    """
+    sl_align = SafetensorLoader(aligned_model_path, local_files_only=configs.local_files_only)
+    sl_base = SafetensorLoader(base_model_path, local_files_only=configs.local_files_only)
+
+    base_model_parameters = [
+        name for name in sl_base.weight_map.keys() if any(v in name for v in list(peft_config.target_modules))
```
BenjaminBossan, line 129: This check for `v in list(peft_config.target_modules)` is not always valid. E.g., `target_modules` can be a str in which case a regex match is executed. There can be other edge cases too. I think what should work is to use `LoraModel._check_target_module_exists(peft_config, name)` ([here](https://github.com/huggingface/peft/blob/b1fd97dc3e71d91abf9a8605e8be74ea6bb751c6/src/peft/tuners/lora/model.py#L159)).

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,229 @@
...
+    ]
+    align_model_parameters = [
+        name for name in sl_align.weight_map.keys() if any(v in name for v in list(peft_config.target_modules))
+    ]
```
BenjaminBossan, line 133: Should we also check that `base_model_parameters` and `align_model_parameters` are the same?

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,229 @@
...
+        vec = sl_base.get_tensor(name_base) - sl_align.get_tensor(name_align)
+        vec = vec.to(devices)
+        if devices == "cpu":
+            vec = vec.to(torch.float32)
+        vec = torch.mm(vec, vec.t()) / torch.norm(vec)
+        safety_vector.append((vec).detach().cpu())
+    return safety_vector
+
+
+def project_weights(configs, peft_weights, v):
```
BenjaminBossan, line 149: Let's rename `configs` to `config` or `safelora_config`.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,229 @@
...
+        default="TheBloke/Llama-2-7B-Chat-fp16",
+        metadata={"help": "The path of the aligned model for obtaining the aligned matrix."},
+    )
+
+    peft_model_path: str = field(
+        default="LisaSchunke/llama-2-7b-peft-finetuned-20000-dataset",
+        metadata={"help": "The path of the LoRA wieghts and configs."},
+    )
+
+    select_layers_type: str = field(
```
BenjaminBossan, line 75: Instead of `str`, we can annotate this as `Literal["threshold", "number"]`.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,229 @@
...
+        default=0.5,
+        metadata={"help": "The threshold of cosine similarity for selecting projected layers."},
+    )
+
+    num_proj_layers: int = field(
+        default=10,
+        metadata={"help": "The number of projected layers."},
+    )
+
+    devices: str = field(
```
BenjaminBossan, line 90:
```suggestion
    device: str = field(
```

examples/safelora/safelora_inference.py
```diff
@@ -0,0 +1,11 @@
...
+from peft.utils.safelora import SafeLoraConfig, apply_safelora
+
+config = SafeLoraConfig(base_model_path='../LLM_Models/llama-2-7b-hf/',\
+                            aligned_model_path='../LLM_Models/llama-2-7b-chat-fp16/',
+                            peft_model_path = '../finetuneLLM/finetuned_models/samsumBad-7b-fp16-peft-seed-42',
+                            devices='cuda',
+                            select_layers_type='threshold',
+                            save_weights=True)
+
+final_lora_weight = apply_safelora(config)
```
BenjaminBossan, line 11: The example should show inference, here we only create the weights. What are the next steps?

BenjaminBossan (changes requested):
Thanks for the update to the SafeLoRA PR. I did another review and found a few areas to improve. Please take a look. Also, please run `make style` once you're finished with your changed.

src/peft/utils/safelora.py
chiayi-hsu: Yes, you're right; bf16 does work on CPUs. I'll make it configurable.

examples/safelora/README.md
chiayi-hsu: I have added more descriptions to the example.
If you feel there are still any missing parts, please let me know.

examples/safelora/README.md
chiayi-hsu: Has been modified.

src/peft/utils/safelora.py
chiayi-hsu: Has been modified.

```diff
- if (safelora_config.devices).lower() == "cpu":
-        peft_weights = {name: f.get_tensor(name).to(safelora_config.dtype) for name in f.keys()}
- else:
-        peft_weights = {name: f.get_tensor(name).to(safelora_config.dtype) for name in f.keys()}
+ peft_weights = {name: f.get_tensor(name).to(safelora_config.dtype) for name in f.keys()}
```

src/peft/utils/safelora.py
chiayi-hsu: Has been modified.

```diff
- if configs.devices != "cpu":
-         P = v[idx].to(torch.bfloat16).to(configs.devices)
- else:
-         P = v[idx].to("cpu")
+ P = v[idx].to(safelora_config.dtype).to(safelora_config.device)

File: src/peft/utils/safelora.py (hunk 7 above)
Comment (chiayi-hsu) [reply src/peft/utils/safelora.py]: Has been modified.

```diff
- cos = torch.round(torch.nn.functional.cosine_similarity(fW.reshape(1, -1), ori.reshape(1, -1)) * 10**5) / 10**5
+ cos = torch.nn.functional.cosine_similarity(fW.reshape(1, -1), ori.reshape(1, -1))
```

src/peft/utils/safelora.py
chiayi-hsu: I have added a check to verify if the model weights are the same.

```diff
+ if (sl_base.get_tensor(name_base) == sl_align.get_tensor(name_align)).all():
+        raise ValueError("The weights of the base Model and the aligned Model should be different.")
```

src/peft/utils/safelora.py
chiayi-hsu: Has been modified.

src/peft/utils/safelora.py
chiayi-hsu: Has been modified.

examples/safelora/safelora_inference.py
chiayi-hsu: I have added more explanations in the README.md and also included code on how to use the SafeLoRA model.

src/peft/utils/safelora.py
chiayi-hsu: I have tried to use ```LoraModel._check_target_module_exists(peft_config, name)```.
However, since the model here is not a PEFT model but a complete model, using its name to find the target module will result in not finding it

BenjaminBossan:
@chiayi-hsu Once you're finished with your changes and want me to give another review, please ping me.

chiayi-hsu:
@BenjaminBossan I have completed the modifications. Please help review them. Thanks!

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,227 @@
...
+from .loftq_utils import _SafetensorLoader as SafetensorLoader
+
+
+@dataclass
+class SafeLoraConfig:
+    """
+    This is the configuration class to store the configuration of a safeLora.
+
+
+    Args:
```
BenjaminBossan, line 37: Could you please format the docstring to be in line with the other docstrings used in PEFT. As an example, check here: https://github.com/huggingface/peft/blob/8874ab5ed8886f988855c93cedd0f323724d35f7/src/peft/tuners/lora/config.py#L128

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,227 @@
...
+        metadata={"help": "The path of the base model for obtaining the aligned matrix."},
+    )
+
+    aligned_model_path: str = field(
+        default="TheBloke/Llama-2-7B-Chat-fp16",
+        metadata={"help": "The path of the aligned model for obtaining the aligned matrix."},
+    )
+
+    peft_model_path: str = field(
+        default="LisaSchunke/llama-2-7b-peft-finetuned-20000-dataset",
```
BenjaminBossan, line 72: IMO, it doesn't make sense to set default values here, I would remove them. WDYT?

src/peft/utils/safelora.py
BenjaminBossan: I meant something else. Would we expect that `base_model_parameters == align_model_parameters`? If not, under what circumstances would they differ?

BenjaminBossan (changes requested):
Thanks a lot for the updates. I did another review. Most of what I found are just smaller things like docs, please take a look.

Now as a next step, it is important that we also add some unit tests. This not going to be very straightforward, because we cannot easily test model alignment and we also don't want to use any big models during unit testing.

One proposal for this would be to use a small model like `hf-internal-testing/tiny-random-OPTForCausalLM` as the base model. Then let's modify some weights (setting them to 0?) and save this as the "aligned" model. Then call `apply_safelora` with these 2 models and various options to see if those tests pass. This would not really check the alignment though.

In addition, we could think about adding a true alignment test for the nightly run with GPU. For this test, it would be okay to use a bigger model (but ideally still not too big).

LMK what you think about this testing strategy and if you have further questions.

Apart from this, please call `make style` on your PR, as this is a prerequisite for the CI to pass.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,227 @@
...
+        name for name in sl_align.weight_map.keys() if any(v in name for v in list(peft_config.target_modules))
+    ]
+    safety_vector = []
+    for name_base, name_align in zip(base_model_parameters, align_model_parameters):
+        if sl_base.get_tensor(name_base).shape != sl_align.get_tensor(name_align).shape:
+            raise ValueError(
+                "The dimensions of the base model's weight should be the same with the aligned model's weight."
+            )
+        if (sl_base.get_tensor(name_base) == sl_align.get_tensor(name_align)).all():
+            raise ValueError("The weights of the base Model and the aligned Model should be different.")
```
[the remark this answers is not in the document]
chiayi-hsu: In our setting, the base model refers to a model that has not undergone training with techniques such as RLHF or SFT, while the aligned model refers to one that has been trained using these techniques. Currently, open-source models such as LLaMA 2 or 3, Mistral, and Gemma have released both base models and chat/instruct (aligned) models. Therefore, we expect the weights of the base model and the aligned model to differ.

chiayi-hsu:
Yes, it still needs to be addressed.

github-actions[bot] ***@***.***>於 2024年12月27日 週五，16:04寫道：

chiayi-hsu:
@BenjaminBossan 
I have made the modifications based on your review and ensured that the style passes by running make style. Regarding the unit test, we can follow your suggestion to use a small model like hf-internal-testing/tiny-random-OPTForCausalLM and modify part of the weights in each layer to be 0, treating it as the so-called aligned model. In our setting, we require that the weights of the base and aligned models be entirely different because we want the aligned model to be a fully fine-tuned model obtained through RLHF techniques. This ensures that the aligned matrix obtained is more flexibly applicable to various attention layers where LoRA is applied.

I would like to ask about the unit test part. Do I need to write test scripts for this? Are there any specific rules or things I should pay attention to?

examples/safelora/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+```
+### Save SafeLoRA's Weights
+If you set `save_weights=False`, but still want to save the weights, you can use the following code.
+
+```python
+from safetensors.torch import save_file
+
+path = ...  # your PEFT model path
+save_file(final_lora_weight, os.path.join(path, "adapter_model.safetensors"))
```
BenjaminBossan, line 34: Here, `path` would be the same as `peft_path` above, right? If so, let's use the same name.

examples/safelora/safelora_inference.py
```diff
@@ -0,0 +1,28 @@
+import os
```
BenjaminBossan, line 1: Let's add the copyright notice.

examples/safelora/README.md
```diff
@@ -0,0 +1,46 @@
...
+### Use SafeLoRA Model
+Next, you can load the base model of the Peft Model along with the Peft model itself to use a model that has both downstream task utility and alignment.
+
+```python
+from transformers import AutoModelForCausalLM
+from peft import PeftConfig, PeftModel
+
+model = AutoModelForCausalLM.from_pretrained(<base-model-id>)
+model = PeftModel.from_pretrained(model, <SafeLoRA-path>)
+```
```
BenjaminBossan, line 46: There is no mention of the `safelora_inference.py` script in this README. Let's mention this script with a short explanation of what it does.

examples/safelora/safelora_inference.py
```diff
@@ -0,0 +1,28 @@
...
+    select_layers_type="threshold",
+    save_weights=True,
+)
+
+final_lora_weight = apply_safelora(config)
+
+save_file(
+    final_lora_weight,
+    f"{os.path.join('../finetuneLLM/finetuned_models/samsumBad-7b-fp16-peft-seed-42', 'adapter_model.safetensors')}",
+)
```
BenjaminBossan, line 25: Can this be removed since we have already set `save_weights=True` above?

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,257 @@
...
+
+from .loftq_utils import _SafetensorLoader as SafetensorLoader
+
+
+@dataclass
+class SafeLoraConfig:
+    """
+    This is the configuration class to store the configuration of a [`safeLora`].
+
+
```
BenjaminBossan, line 37:
```suggestion
```

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,257 @@
...
+import peft
+from peft import PeftConfig
+
+from .loftq_utils import _SafetensorLoader as SafetensorLoader
+
+
+@dataclass
+class SafeLoraConfig:
+    """
+    This is the configuration class to store the configuration of a [`safeLora`].
```
BenjaminBossan, line 35: Regarding [`safeLora`]: There is no object of this name, right? Therefore, this syntax will not work. We could rephrase this to:

> This is the configuration class to set the parameters for SafeLoRA.

Then add a link to the paper.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,227 @@
...
+    )
+
+    aligned_model_path: str = field(
+        default="TheBloke/Llama-2-7B-Chat-fp16",
+        metadata={"help": "The path of the aligned model for obtaining the aligned matrix."},
+    )
+
+    peft_model_path: str = field(
+        default="LisaSchunke/llama-2-7b-peft-finetuned-20000-dataset",
+        metadata={"help": "The path of the LoRA wieghts and configs."},
```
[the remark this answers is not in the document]
BenjaminBossan: Typo is still there.

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,257 @@
...
+        },
+    )
+
+    def __post_init__(self):
+        if self.base_model_path is None:
+            raise ValueError("base_model_path cannot be None.")
+        if self.aligned_model_path is None:
+            raise ValueError("aligned_model_path cannot be None.")
+        if self.peft_model_path is None:
+            raise ValueError("peft_model_path cannot be None.")
```
BenjaminBossan, line 131: Since there are no longer any default values for these fields, I don't believe we need to perform these checks anymore.

src/peft/utils/safelora.py
BenjaminBossan: Still open.

src/peft/utils/safelora.py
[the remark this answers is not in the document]
BenjaminBossan: I understand the difference between aligned model and base model. However, it should be possible to align a model without changing each and every parameter, right? E.g., in the future we could a new model where the aligned model only changes half of the weights, I don't see why that couldn't be possible. Does the SafeLoRA algorithm really require each parameter to be different? Can we not skip the layers if the parameters are identical?

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,257 @@
...
+
+        dist = 1 / (1 + torch.norm(peft_weights[name_B].reshape(1, -1) - W.reshape(1, -1)))
+
+        dis.append(dist.item())
+    return peft_weights, cos_total
+
+
+def apply_safelora(safelora_config: SafeLoraConfig):
+    """
+
```
BenjaminBossan, line 193:
```suggestion
```

src/peft/utils/safelora.py
```diff
@@ -0,0 +1,257 @@
...
+    Therefore, applying `apply_safelora()` is intended to help preserve the alignment of the final models.
+
+    It is important to note that the model weights of the aligned model and the base model must be of the same size.
+    Additionally, for SafeLoRA, only the safetensors format is supported.
+
+    Args:
+        safelora_config: The config of SafeLora.
+
+    Returns:
+        `torch.nn.Module`: The Lora model is applied SafeLoRA.
```
BenjaminBossan, line 207: This is incorrect, the return value is a `state_dict` containing the PEFT weights.

BenjaminBossan (changes requested):
Thank you for the updates.

I reviewed the PR again and found a few more things that can be improved, it's mostly about clarity and formatting.

Regarding the unit tests:

I think it would be easiest to proceed as follows. Let's create a new file, `tests/test_safelora.py`. Next, I created a small template for you to get started:

```python
class TestSafeLora:

    @pytest.fixture(scope="class")
    def aligned_model_path(self, tmp_path):
        # we create a fake aligned model where the weights differ from the base model
        model = AutoModelForCausalLM.from_pretrained(self.model_id)
        for param in model.parameters():
            # modify the parameters to be different
            ...

        model.save_pretrained(tmp_path / "aligned_model")

    @pytest.fixture
    def lora_path(self, tmp_path):
        # create a LoRA adapter
        model = AutoModelForCausalLM.from_pretrained(self.model_id)
        lora_config = LoraConfig(init_lora_weights=False)  # initialize LoRA so that it's not a no-op
        model.save_pretrained(tmp_path / "lora")

    def test_safelora_with_threshold(self, aligned_model_path, lora_path):
        ...  # code to test

    def test_safelora_with_num_proj_layers(self, aligned_model_path, lora_path):
        ...  # code to test

    def test_safelora_with_save_weights_false(self, aligned_model_path, lora_path):
        ...  # code to test

    # etc. more tests for the different options here
```

I hope this makes sense, if not, feel free to ask questions.

As discussed earlier, since this does not use a real aligned model but just creates a fake one, we cannot really test if the final model is better aligned or not. For this, we rely on the paper results and assume they're correct. If you have a good idea for a real alignment test, we can also add that test though. Just ensure that we're only using very small models to not slow down the CI.

chiayi-hsu:
I'm still working on it, but due to an upcoming paper submission
deadline, there will be some delays.

Benjamin Bossan ***@***.***> 於 2025年2月10日 週一 上午11:54寫道：

BenjaminBossan:
Hey @chiayi-hsu I was recently thinking about this PR and wondered if you're interested in reviving it.

PR #2202, opened 2024-11-07
FIX: LoRA & DoRA for depthwise-convolutional layers

gslama12:
I don't know if this is something that you want to fix, but i am currently doing research on PEFT methods for edge device models and made this small change to allow LoRA & DoRA to be applied to CNNs that use depthwise-separable convolution.

**Example:**
For a depthwise conv layer like `Conv2d(32,  32, kernel=(3,3), groups=32)`, the weight would be `(32,1,3,3)`. When we apply LoRA we cascade a `Conv2d(32,  r, kernel=(3,3))` and a` Conv2d(r,  32, kernel=(1,1))`. This means that we get 2 weights with dimensions `(r, 32, 3, 3)` and `(32, r, 1, 1)`. This means that even if we choose r=1, the LoRA adapter alone has more parameters than the original depthwise-convolution, which not parameter efficient. There is also no way around that, as the dw-convolution will always have less parameters than the lora adapter.


gslama12:
Closed, because this can be handled using "target_modules=" in LoraConfig.

BenjaminBossan:
It could make sense to emit a warning in case that the LoRA adapter has more parameters than the original layer, but as you say, users can control which layers to target and it's their job to choose the layers and hyper-parameters in a way that actually leads to reduced trainable parameters.

Issue #2170, opened 2024-10-22
Invalid `none` check for loftq_config attribute in `LoraConfig`


sirluk:
### System Info

- `transformers` version: 4.45.2
- Platform: Linux-5.14.0-427.37.1.el9_4.x86_64-x86_64-with-glibc2.34
- Python version: 3.12.7
- Huggingface_hub version: 0.26.0
- Safetensors version: 0.4.5
- Accelerate version: 1.0.1
- Accelerate config:    not found
- PyTorch version (GPU?): 2.5.0+cu124 (True)
- Tensorflow version (GPU?): not installed (NA)
- Flax version (CPU?/GPU?/TPU?): not installed (NA)
- Jax version: not installed
- JaxLib version: not installed
- Using distributed or parallel set-up in script?: <fill in>
- Using GPU in script?: <fill in>
- GPU type: NVIDIA A100-SXM4-80GB

### Who can help?

@BenjaminBossan 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

```python
from peft import LoraConfig

LoraConfig(init_lora_weights="loftq")
```
From the code in peft.tuners.lora.config.py this should raise a `ValueError`. 
```python
if self.loftq_config is None:
    raise ValueError("`loftq_config` must be specified when `init_lora_weights` is 'loftq'.")
```
However the loftq_config field of `LoraConfig` is defined  to be a dict by default, so it is never None.

### Expected behavior

`ValueError` should be raised when `init_lora_weights="loftq"` but `loftq_config` is not specified.

BenjaminBossan:
Good catch, indeed this check doesn't make sense as is. I'll fix it once I have a bit of time (it's not high priority). If someone else wants to take this, feel free to do so.

sparsh2:
@BenjaminBossan I would like to take this up if its okay

sirluk:
@BenjaminBossan @sparsh2 If its fine I would do it since I already started working on it as part of a different pull request that ideally should have the same structure as the fix for this

BenjaminBossan:
Thanks, please go ahead @sirluk since you already started to work on it. @sparsh2 I hope you did not put any time into this yet.

sparsh2:
No issues @sirluk @BenjaminBossan. @sirluk please go aheahd

PR #2215, opened 2024-11-14
[FIX] Invalid `None` check for `loftq_config` attribute in `LoraConfig`

sirluk:
Fix for #2170 

I improved the handling of loftq attributes in the `__init__` method of `LoraConfig`.

Aside from fixing the issue of the None check I added a warning when `init_lora_weights != "loftq"` but `loftq_config` is not specified. In that case I think it also makes sense to reset `loftq_config` to an empty dict.

I added two testcases for the expected Error/Warning is raised.

fyi: on my local machine a lot of gpu tests are failing but I dont think it is related to the changes as the same tests are failing when using the repo state from the main branch.


BenjaminBossan:
_Just a note: For some reason, the MacOS CI tests are not running. Last time this happened, it was (probably) some GH issue that resolved itself after waiting a day. I'll check this again tomorrow._

BenjaminBossan (approved):
Thanks a lot for the fixes to the LoftQ config checking and associated tests. LGTM.

PR #2222, opened 2024-11-19
CI Update AutoAWQ version to fix CI

BenjaminBossan:
Currently, the [CI fails](https://github.com/huggingface/peft/actions/runs/11905036144/job/33174835723#step:5:62) because awq tries to import a non-existing function from transformers (presumably it was there at one point but no longer is):

>     from transformers.modeling_utils import shard_checkpoint
> E   ImportError: cannot import name 'shard_checkpoint' from 'transformers.modeling_utils' (/opt/conda/envs/peft/lib/python3.11/site-packages/transformers/modeling_utils.py)

This has been [fixed](https://github.com/casper-hansen/AutoAWQ/commit/a28c747903bd917aaf43700e9ca8db65e74de19b) in awq v0.2.7. Therefore, this newer version is now used in CI.

### DIFF
--- a/docker/peft-gpu/Dockerfile
+++ b/docker/peft-gpu/Dockerfile
@@ -41,8 +41,8 @@ SHELL ["/bin/bash", "-c"]
 RUN source activate peft && \ 
     python3 -m pip install --no-cache-dir bitsandbytes optimum auto-gptq && \
     # Add autoawq for quantization testing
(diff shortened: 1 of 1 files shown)


SunMarc (approved):
Thanks for the update @BenjaminBossan

LysandreJik:
no problem for me!

PR #2223, opened 2024-11-19
TST: Move slow compile tests to nightly CI

BenjaminBossan:
These tests use `torch.compile` on semi-realistic models and are thus slow to execute. In sum, they take ~3 min to finish, with an overall CI runtime of ~15 min, so it's significant.

As these tests are very unlikely to be affected by most code changes, it should be fine to move them to nightly CI instead of running them on each PR. Also, presence of GPUs might speed the tests up.

Note that apart from moving the tests around, they are completely unchanged. They pass locally.

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -15,6 +15,8 @@ import importlib
 import itertools
 import os
 import re
+import subprocess
+import sys
 import tempfile
 import unittest
 from collections import Counter, defaultdict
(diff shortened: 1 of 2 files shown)


tests/test_gpu_examples.py
```diff
@@ -4069,3 +4071,78 @@ def test_prefix_tuning_multiple_devices_encoder_decoder_model(self):
         peft_config = PrefixTuningConfig(num_virtual_tokens=10, task_type="SEQ_2_SEQ_LM")
         model = get_peft_model(model, peft_config)
         model.generate(**inputs)  # does not raise
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="test requires a GPU")
+@pytest.mark.single_gpu_tests
```
sayakpaul, line 4077: Is this how we detect if a test is run nightly? If so, fine!

sayakpaul (approved):
Makes absolute sense!

tests/test_gpu_examples.py
BenjaminBossan: GPU tests are skipped for normal CI, and for nightly, they're included. So it's more of an indirect filter.

PR #2225, opened 2024-11-20
[FEAT] EVA: ensure deterministic behavior of SVD on multi gpu setups

sirluk:
Deterministic behavior of SVD computations on layer inputs is important on multi gpu setups to ensure resulting EVA state dicts are identical. Currently deterministic behaviour of SVD is ensured by just setting torch.manual_seed() inside `_get_eva_state_dict` which is not ideal because it can interfere with the seed set by the user. I therefore moved this functionality inside the `IncrementalPCA` class.

I also added a new argument for `get_eva_state_dict`: `gather_distributed_inputs` which is `False` by default and defines if inputs should be concatenated for the SVD computation for distributed runs.

### DIFF
--- a/examples/eva_finetuning/eva_finetuning_multi_gpu.py
+++ b/examples/eva_finetuning/eva_finetuning_multi_gpu.py
@@ -87,7 +87,7 @@ sampler.set_epoch(0)
 
 # Wrap model in DDP
 model = model.to(local_rank)
-model = DDP(model, device_ids=[local_rank], output_device=local_rank, find_unused_parameters=False)
+model = DDP(model, device_ids=[local_rank], output_device=local_rank)
 
 # setup peft config
 eva_config = EvaConfig(rho=rho)
@@ -96,7 +96,8 @@ peft_config = LoraConfig(
 )
 
 # EVA initialization
-eva_state_dict = get_eva_state_dict(model, dataloader, peft_config)
+# It is important to set `gather_distributed_inputs=True` here if you use a distributed data sampler.
+eva_state_dict = get_eva_state_dict(model, dataloader, peft_config, gather_distributed_inputs=True)
 eva_state_dict = {".".join(["base_model.model"] + k.split(".")[1:]): v for k, v in eva_state_dict.items()}
 
 # cleanup ddp
--- a/examples/eva_finetuning/eva_finetuning_multi_gpu.py
+++ b/examples/eva_finetuning/eva_finetuning_multi_gpu.py
@@ -96,8 +96,7 @@ peft_config = LoraConfig(
 )
 
 # EVA initialization
-# It is important to set `gather_distributed_inputs=True` here if you use a distributed data sampler.
-eva_state_dict = get_eva_state_dict(model, dataloader, peft_config, gather_distributed_inputs=True)
+eva_state_dict = get_eva_state_dict(model, dataloader, peft_config)
 eva_state_dict = {".".join(["base_model.model"] + k.split(".")[1:]): v for k, v in eva_state_dict.items()}
 
 # cleanup ddp
--- a/src/peft/tuners/lora/eva.py
+++ b/src/peft/tuners/lora/eva.py
@@ -469,7 +469,7 @@ def _load_eva_state_dict(
             elif new_rank != r:
                 if peft_config.eva_config.adjust_scaling_factors:
                     alpha *= new_rank / r
-            if new_rank != r or module.lora_A[adapter_name].weight.device == "meta":
+            if new_rank != r or module.lora_A[adapter_name].weight.device.type == "meta":
                 module.update_layer(r=new_rank, lora_alpha=alpha, init_lora_weights="eva", **update_layer_kwargs)
             module.lora_A[adapter_name].weight.copy_(w)
             new_target_modules.append(name_in_base_model)
--- a/src/peft/tuners/lora/eva.py
+++ b/src/peft/tuners/lora/eva.py
@@ -43,8 +43,14 @@ class _Hook:
     A base class for hooks that prepares layer inputs for EVA.
     """
 
-    def __init__(self, name: str, prepare_layer_inputs_fn: Optional[callable] = None):
+    def __init__(
+        self,
+        name: str,
+        prepare_layer_inputs_fn: Optional[callable] = None,
+        gather_distributed_inputs: bool = True,
+    ):
         self.name = name
+        self.gather_distributed_inputs = gather_distributed_inputs
         if prepare_layer_inputs_fn is None:
             self._prepare_layer_inputs_fn = self._prepare_layer_inputs_fn_default
         else:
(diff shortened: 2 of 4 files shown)


src/peft/tuners/lora/eva.py
```diff
@@ -43,8 +43,14 @@ class _Hook:
     A base class for hooks that prepares layer inputs for EVA.
     """
 
-    def __init__(self, name: str, prepare_layer_inputs_fn: Optional[callable] = None):
+    def __init__(
+        self,
+        name: str,
+        prepare_layer_inputs_fn: Optional[callable] = None,
+        gather_distributed_inputs: bool = False,
```
BenjaminBossan, line 50: For my understanding, setting `gather_distributed_inputs=True` in a non-distributed setting makes no sense. In what circumstances would we `gather_distributed_inputs=False` in a distributed setting?

src/peft/tuners/lora/eva.py
```diff
@@ -314,6 +317,14 @@ def _get_rank_distribution(hooks, layer_hook_map, equal_inputs_map, rank_budget,
     if len(dataloader) == 0:
         raise ValueError("dataloader is empty")
 
+    # check if dist is initialized
+    if dist.is_initialized() and gather_distributed_inputs:
+        warnings.warn(
+            "torch.distributed is initialized and `gather_distributed_inputs` is True, "
+            "therefore EVA initialization will gather tensors from all ranks. "
+            "Ensure the model does not recieve the same inputs on different ranks."
```
BenjaminBossan, line 325:
```suggestion
            "Ensure the model does not receive the same inputs on different ranks."
```

src/peft/tuners/lora/eva.py
```diff
@@ -579,6 +597,7 @@ def get_eva_state_dict(
             case model_inputs is the mask used to determine which indices should be used for SVD (created by
             `prepare_model_inputs_fn_language_modeling`).
         adapter_name (str): The name of the adapter to compute the SVD for.
+        gather_distributed_inputs (bool): Whether to gather the layer inputs from all ranks. Default is False.
```
BenjaminBossan, line 600: Let's add: "Set this to True if ...".

src/peft/tuners/lora/eva.py
```diff
@@ -672,6 +693,7 @@ def initialize_lora_eva_weights(
             case model_inputs is the mask used to determine which indices should be used for SVD (created by
             `prepare_model_inputs_fn_language_modeling`).
         adapter_name (str): The name of the adapter to initialize the weights for.
+        gather_distributed_inputs (bool): Whether to gather the layer inputs from all ranks. Default is False.
```
BenjaminBossan, line 696: Let's add: "Set this to True if ...".

src/peft/utils/incremental_pca.py
```diff
@@ -66,11 +68,18 @@ def __init__(
             if lowrank_q < n_components:
                 raise ValueError("lowrank_q must be greater than or equal to n_components.")
 
-            def svd_fn(X):
-                U, S, V = torch.svd_lowrank(X, q=lowrank_q, niter=lowrank_niter)
+            def svd_fn(X, seed):
+                if seed is not None:
```
BenjaminBossan, line 72: I haven't used it, but could this be a good situation to use [`fork_rng`](https://pytorch.org/docs/stable/random.html#torch.random.fork_rng)?

src/peft/utils/incremental_pca.py
```diff
@@ -66,11 +68,18 @@ def __init__(
             if lowrank_q < n_components:
                 raise ValueError("lowrank_q must be greater than or equal to n_components.")
 
-            def svd_fn(X):
-                U, S, V = torch.svd_lowrank(X, q=lowrank_q, niter=lowrank_niter)
+            def svd_fn(X, seed):
```
BenjaminBossan, line 71: Would it make sense to turn `svd_fn` into a proper method, with all arguments stored as `self` attributes (so `self.lowrank = lowrank` etc.). Then we don't need to `partial` over the arguments. Furthermore, this is more in line with sklearn: Save all the class arguments as `self` attributes and don't put any further logic inside of `__init__` -- this is because this logic is not executed when running `set_params` on the object.

Of course, this `IncrementalPCA` class is not a proper sklearn `BaseEstimator` with `get_params` and `set_params`, so it doesn't really matter, but it could still be nice for possible future integration to keep with the sklearn standards.

BenjaminBossan (commented):
Thanks for these small fixes for distributed EVA. Overall, this looks good as is but I have added some comments for potential improvement, please check.

src/peft/tuners/lora/eva.py
sirluk: My thinking when adding this parameter was that users might use a distributed environment but for some reason do not create a distributed dataloader for EVA. For those cases we would have the same examples as layer inputs and thus when gathering we would have duplicates in the input to the SVD computation. Not sure if this is a valid scenario of if we can disregard this.

src/peft/utils/incremental_pca.py
sirluk: I didnt know about this function but it looks useful. I will check if I can implement it.

src/peft/utils/incremental_pca.py
sirluk: done, I cleaned up the __init__ method and defined the svd methods as proper class methods

src/peft/tuners/lora/eva.py
BenjaminBossan: Ah yes, that makes sense. Could you please add this info to the docstring below, that way users know when they should set this in a distributed setting.

One thing I wonder: Could we make this `Optional[bool]` and if the value is `None` and the setting is distributed, automatically switch to `True`, otherwise to `False`? This is probably the most common case. Users still can pass `False` explicitly in a distributed setting if they want to.

sirluk: I think we can leave it as `bool` and set the default to `True` as the Hooks anyway only gather if `dist.is_initialized()`. So if the setting is not distributed this argument has no effect.

BenjaminBossan, line 50: Since `gather_distributed_inputs` defaults to `True` now in `initialize_lora_eva_weights`, should it also be changed here?

examples/eva_finetuning/eva_finetuning_multi_gpu.py
```diff
@@ -96,7 +96,8 @@
 )
 
 # EVA initialization
-eva_state_dict = get_eva_state_dict(model, dataloader, peft_config)
+# It is important to set `gather_distributed_inputs=True` here if you use a distributed data sampler.
+eva_state_dict = get_eva_state_dict(model, dataloader, peft_config, gather_distributed_inputs=True)
```
BenjaminBossan, line 100:
```suggestion
eva_state_dict = get_eva_state_dict(model, dataloader, peft_config)
```

As `True` is the default.

BenjaminBossan (commented):
Thanks for the updates. I have some nits, WDYT?

sirluk:
@BenjaminBossan thanks for the suggestions, I integrated them!

BenjaminBossan (approved):
Thanks for these updates to EVA, LGTM.

PR #2226, opened 2024-11-20
CI: Skip EETQ tests while broken

BenjaminBossan:
EETQ tries to import the `shard_checkpoint` function from transformers but it has been removed in the latest version. Therefore, trying to use EETQ currently results in an import error that crashes all the tests. This fix results in EETQ tests being skipped if there is an import error.

The issue has been reported to EETQ: https://github.com/NetEase-FuXi/EETQ/issues/34


SunMarc (approved):
Thanks for fixing this ! LGTM ! If you have some time, could you also do it for transformers ?

PR #2142, opened 2024-10-10
[FEAT] New LoRA Initialization Method: Explained Variance Adaptation

sirluk:
### Description

In our work "One Initialization to Rule them All: Fine-tuning via Explained Variance Adaptation" ([paper](https://arxiv.org/abs/2410.07170)) we introduce a new data-driven initialization method for LoRA adapters. We will present this work at [ENLSP](https://neurips2024-enlsp.github.io/#)  (oral) and [AFM](https://adaptive-foundation-models.org/call-for-papers.html) at NeurIPS'24.

Eva initializes LoRA in a data-driven manner based on information of the downstream data using SVD. It also adaptively allocates ranks throughout the model.

### Implementation Details

1. As the initialization is data-driven we created a new config class `EvaConfig` which at minimum needs to receive a value for the `dataloader` argument. Other parameters have default values which worked well in our experiments. The instantiated EvaConfig is passed as a parameter to LoraConfig. (we took inspiration from the LoftQConfig implementation).
In LoraConfig "init_lora_weights" needs to be set to "eva"
2. We modify the `__init__` method of `LoraModel` to create a state dict for eva initialzation before initializing the parent class to be able to populate the `rank_pattern`argument in `LoraConfig`. This way we directly initialize with a modified rank distribution allowing us to directly load the eva_state_dict.
3. All other code necessary for the initialzation is self-contained in peft/tuners/lora/eva.py
4. Under examples/eva_finetuning/eva_finetuning.py we add an example script for eva initialization

### Open Questions

- Not sure if `self.model.load_state_dict(eva_state_dict, strict=False)` is always a valid way to initialize. E.g. when the model is wrapped with torch.compile
- We currently only tested a single GPU setting. We are not sure what the best way is to have multi-gpu support (only run init on rank 0 and copy to other ranks when done?)

### Example Usage

```python
eva_config = EvaConfig(
    dataloader = dataloader
)
peft_config = LoraConfig(
    r = 16,
    lora_alpha = 1,
    target_modules = ['q_proj', 'k_proj', 'v_proj', 'o_proj'],
    init_lora_weights = "eva",
    eva_config = eva_config,
)
peft_model = get_peft_model(model, peft_config)
```

PR #2231, opened 2024-11-21
Adding CorDA as an optional initialization method of LoRA

iboing:
Existing PEFT methods are mostly agnostic of the context of a task of concern, e.g., a downstraem task to learn or some pre-trained world knowledge to maintain. In our paper (https://openreview.net/pdf?id=Gi00NVru6n) accepted by NeurIPS 2024, we propose a PEFT method named Corda, context-oriented decomposition adaptation, which builds task-aware LoRA adapters from weight decomposition oriented by the context of the task concerned for **task-aware PEFT**. 

Concretely, CorDA randomly collects a few (by default 256) data samples from a target task, e.g. questions from a QA dataset or instructions to write a code or solve a math problem, and feeds these samples into a pre-trained LLM. We can obtain the covariance matrix of the input activation of each linear layer, i.e., $C=XX^T\in\mathcal{R}^{d_{in}\times d_{in}}$, where $X$ is the input of this linear layer. We then perform singular value decomposition (SVD) for the weight $W\in \mathcal{R}^{d_{out}\times d_{in}}$ multiplied by the covariance matrix, i.e., $\verb|SVD|(WC) = U\Sigma V^T$, where $U$ and $V$ are singular vectors and $\Sigma$ is the diagonal matrix with the singular values arranged in descending order. In this way, the context expressed by these representative covariance matrices is able to orientate the decomposition, such that the principal components are most associated with the task of concern. To ensure the same inference result with the pre-trained model at the start of adaptation, we multiply the inverse of these covariance matrices with the decomposed components, $\hat{W}=U\Sigma V^T C^{-1}$, where $\hat{W}$ is the weight after decomposition and reconstruction. 

Thanks to the task-awareness, CorDA enables two optional modes, **knowledge-preserving adaptation mode (KPM)** and **instruction-previewed adaptation mode (IPM**). In KPM, we use questions from question-answering dataset whose knowledge needs to be preserved, such as TriviaQA and NQopen, to obtain the covariance matrices. After our context-oriented decomposition, we use the components with the smallest $r$ singular values, $U_{[:,-r:]}$, $\Sigma_{[-r:]}$, and $(V^TC^{-1})\_{\left[-r:,:\right]}$ to initialize the learnable LoRA adapters $A=\sqrt{\Sigma}\_{[-r:]}(V^TC^{-1})\_{[-r:,:]}$ and $B=U\_{[:,-r:]}\sqrt{\Sigma}\_{[-r:]}$. The other components that compact the question-answering ability are frozen during adaptation. KPM enables to learn new tasks effectively while keeping the world knowledge associated with $C$ as sound as possible. Alternatively, when one only aims to achieve performance as high as possible on the finetuning task without concern for world knowledge maintenance, our IPM will be favored. In this mode, CorDA uses the instruction and response from the fine-tuning task (e.g., Math or Code) to produce the covariance matrices. The principal components with large singular values capturing the characteristics of the finetuning task in advance can better accommodate the new ability. So we initialize adapters as $A= \sqrt{\Sigma}\_{[:r]} (V^T C^{-1})\_{[:r,:]}$ and $B =U_{[:,:r]} \sqrt{\Sigma}_{[:r]}$, and freeze the remaining components. The implementations of KPM and IPM are compared as follows:

| Mode | Collect covariance from | LoRA $A$ | LoRA $B$ |
|---|---|---|---
|KPM | questions from a knowledge benchmark to maintain | $A=\sqrt{\Sigma}\_{[-r:]}(V^T C^{-1})\_{[-r:,:]}$ | $B=U_{[:,-r:]}\sqrt{\Sigma}_{[-r:]}$ |
IPM | instructions and responses from a downstream task to learn | $A= \sqrt{\Sigma}\_{[:r]} (V^T C^{-1})\_{[:r,:]}$ | $B =U_{[:,:r]} \sqrt{\Sigma}_{[:r]}$ |

fig1

Similar to PiSSA, our method is also an initialization method for LoRA adapter with the same LoRA structure, and our IPM also uses the principal components to initialize $A$ and $B$. But PiSSA adopts the naive SVD and does not consider task awareness. Our method builds task-aware adapters, enabling two optional modes to satisfy customized needs: 
- KPM not only achieves better fine-tuning performance than LoRA but also effectively preserves pre-trained knowledge.
- IPM can further accelerate convergence and enhance the fine-tuning performance on downstream tasks, surpassing PiSSA in our experiments. 

Some results are shown below. 

### 1. The superiority of our decomposition method.

fig2

Plain SVD is the normal SVD on pre-trained weights, as adopted by PiSSA. This experiment discards the smallest $r$ components after different decomposition methods and tests the perplexity on Wiki and PTB. It indicates that our decomposition better compacts the associated task abilities into the principal components, and thus can better maintain the corresponding knowledge by freezing them in KPM, or better learn new abilities by adapting them in IPM. 

### 2. Adaptation performance with knowledge-preserving mode (sample from NQopen, fine-tune on Math)

| Method | Model | NQ open | GSM8k | Math | Avg. |
|---|---|---|---|---|---|
|Pre-trained|Llama-2-7b| 14.99 | -| - | - |
|LoRA|Llama-2-7b|1.27| 42.68 | 5.88 | 16.61 |
|**CorDA (KPM)** |Llama-2-7b| **8.20** | **46.32**	| **7.00** | **20.51** |
|Pre-trained|Llama-2-13b|23.63|-|-|-|
|LoRA|Llama-2-13b| 16.26 | 57.24 | 8.92 | 27.47 |
|**CorDA (KPM)** |Llama-2-13b| **19.86** | **59.29** | **9.62** | **29.59** |
|Pre-trained|Llama-3-8b|13.41|-|-|-|
|LoRA|Llama-3-8b| 8.75 | 72.33 | 24.04| 35.04 |
|**CorDA (KPM)** |Llama-3-8b| **9.61** | **74.68** | **25.34** | **36.54** |
|Pre-trained|Gemma-2-9b|12.85|-|-|-|
|LoRA|Gemma-2-9b| 9.28 | 83.47 | 42.30| 45.02 |
|**CorDA (KPM)** |Gemma-2-9b|**10.17** | **84.08** | **42.64** | **45.63** |

### 3. Adaptation performance with instruction-previewed mode (sample from Math, fine-tune on Math)

| Method | Model | GSM8k | Math |
| --- | --- | --- | ---|
|LoRA| Llama-2-7b | 42.68 | 5.88 |
|PiSSA | Llama-2-7b | 51.63 | 7.32 |
| **CorDA (IPM)** | Llama-2-7b | **53.45** | **8.64** |
|LoRA| Llama-2-13b | 57.24 | 8.92 |
|PiSSA | Llama-2-13b |60.88	| 11.08|
| **CorDA (IPM)** | Llama-2-13b | **62.47** |**11.54** |
|LoRA| Gemma-2-9b | 83.47 |	42.30 |
|PiSSA | Gemma-2-9b | 84.23	| 43.52|
| **CorDA (IPM)** | Gemma-2-9b | **84.45** | **43.88** |

fig3


BenjaminBossan (commented):
Thanks for this PR to add CorDA to PEFT.

I don't have time to do a proper review this week, so I just did a very quick check and added some comments. I'll do a proper review next week.

5eqn:
@BenjaminBossan Thanks for your review! As @iboing 's collaborator, I've revised the code according to your comments.

iboing:
@BenjaminBossan  Thank you for your review!

BenjaminBossan:
Thanks a lot for the updates. The week has been quite busy so far, but I'm hopeful I can give this a proper review tomorrow. Just from reading the description of the method and skimming the code, I have some questions/comments:

1. The description of the method mentions that it depends on the input data. I could not find in the code where this happens, for instance `preprocess_corda` does not take any data as input and in the `corda_finetuning.py` script I also don't see how CorDA incorporates the dataset. Could you please point me to where this happens?
2. You're using logs throughout the code. In general, we don't do this in PEFT. Could you please remove those calls.
3. Purely from the description (but I haven't studied the paper yet), this method reminded me of EVA, which recently added to PEFT (#2142, https://arxiv.org/abs/2410.07170). Do you know if the two methods are similar? AFAICT, the respective papers don't mention another.

5eqn:
@BenjaminBossan Thanks for your comments! Let me clarify:

> I also don't see how CorDA incorporates the dataset. Could you please point me to where this happens?

In [corda_finetuning/preprocess.py#L38-L59
](https://github.com/huggingface/peft/blob/aed135d8930bb0b1d958e1cb0f6a57199f60f980/examples/corda_finetuning/preprocess.py#L38-L59) `calib_loader` is the dataset iterator, and we're passing `config.corda_config.run_model = lambda: run_model(model, calib_loader)`, this way CorDA gets a closure that *captures a reference* to the dataset.

This does not happen in `corda_finetuning/corda_finetuning.py` because it's the file to do the *actual finetuning*, but as we've preprocessed CorDA initial weights and residual model, we can train it like *standard LoRA models*, thus there's no need for dataset.

> You're using logs throughout the code. In general, we don't do this in PEFT. Could you please remove those calls.

Sure, I've updated the PR so that logs are removed.

examples/corda_finetuning/README.md
```diff
@@ -0,0 +1,128 @@
...
+trainer = SFTTrainer(
+    model=peft_model,
+    args=training_args,
+    train_dataset=dataset,
+    tokenizer=tokenizer,
+)
+trainer.train()
+peft_model.save_pretrained("corda-llama-2-7b")
+```
+
```
BenjaminBossan, line 119: Let's mention also the `preprocess.py` and `corda_finetuning.py` scripts here, what they do, and how users should use them.

examples/corda_finetuning/preprocess.py
```diff
@@ -0,0 +1,151 @@
...
+        "--model_id",
+        type=str,
+        default="meta-llama/Llama-2-7b-hf",
+        help="Pretrained model ID",
+    )
+    parser.add_argument(
+        "--alpha",
+        type=float,
+        default=0.5,
+        help="hyper-parameter alpha for ASVD",
```
BenjaminBossan, line 95: Could you add more details here, what are good value ranges, what does this parameter change? Also, I don't see it being used in the script.

src/peft/tuners/lora/config.py
```diff
@@ -120,6 +120,100 @@ def __post_init__(self):
...
+            you should configure the `run_model` callback to run the model exactly `sample_count` times. Typically you
+            should run model inference on your dataset in this callback.
+        hooked_model (`Optional[nn.Module]`):
+            Model to hook when building covariance. If none, original model will be hooked. This is only useful when
+            you want to hook a different model than the one you are training, typically you should leave this `None`.
+        sample_count (`int`):
+            Divisor for each hook call. You should make sure `run_model` calls the model exactly `sample_count` times
+            for consistent behaviour.
+        corda_method (`Literal["ipm", "kpm"]`):
+            Method to build adapter. The KPM not only achieves better performance than LoRA on fine-tuning tasks, but
```
BenjaminBossan, line 149: Let's clarify here what "KMP" and "IPM" mean.

src/peft/tuners/lora/config.py
```diff
@@ -199,6 +293,9 @@ class LoraConfig(PeftConfig):
         eva_config (`Optional[EvaConfig]`):
             The configuration of EVA. At a minimum the dataset argument needs to be set (use the same dataset as for
             finetuning).
+        corda_config (`Optional[CordaConfig]`):
```
BenjaminBossan, line 296: Could you please also update the docstring and the help of `init_lora_weights` to include `"corda"`?

src/peft/tuners/lora/config.py
```diff
@@ -466,6 +572,10 @@ def __post_init__(self):
             )
             warnings.warn(msg)
 
+        # convert corda_config to dict
+        if self.corda_config and not isinstance(self.corda_config, dict):
```
BenjaminBossan, line 576: What happens if users pass `init_lora_weights="corda"` but don't pass `corda_config`? Can we use the default `CordaConfig`?

And what if users pass the `corda_config` but don't set `init_lora_weights="corda"`? Probably an error should be raised on that case.

src/peft/tuners/lora/layer.py
```diff
@@ -250,6 +253,56 @@ def pissa_init(self, adapter_name, init_lora_weights):
...
+                "Please initialize CorDA under float32, float16, or bfloat16. "
+                "Subsequently, re-quantize the residual model to help minimize quantization errors."
+            )
+        weight = weight.to(torch.float32)
+        out_dim = weight.data.size(0)
+        in_dim = weight.data.size(1)
+
+        # Calculate WC from covariance matrix
+        if not hasattr(linear, "eigens"):
+            raise ValueError("eigens attribute not found, it's expected to be set by the pre-injection hook.")
```
BenjaminBossan, line 271: What should a user do with this info? Probably if they encounter it, it's not a user error but an issue with PEFT, right? If so, let's instruct the user to open an issue on PEFT.

src/peft/tuners/lora/layer.py
```diff
@@ -250,6 +253,56 @@ def pissa_init(self, adapter_name, init_lora_weights):
...
+        if torch.isnan(V).any() or torch.isinf(V).any():
+            raise ValueError("nan or inf in V")
+
+        # Sanity check
+        if U.size(0) != out_dim or U.size(1) != r:
+            raise ValueError(f"U size mismatch: {U.size()} vs. ({out_dim}, {r})")
+        if S.size(0) != r:
+            raise ValueError(f"S size mismatch: {S.size()} vs. ({r},)")
+        if V.size(0) != in_dim or V.size(1) != r:
+            raise ValueError(f"V size mismatch: {V.size()} vs. ({in_dim}, {r})")
```
BenjaminBossan, line 292: Similar comment as above: What should the user do if they encounter this error? Is it something they can mitigate by setting other hyper-parameters for CorDA or using a different dataset sample?

src/peft/tuners/lora/layer.py
```diff
@@ -250,6 +253,56 @@ def pissa_init(self, adapter_name, init_lora_weights):
...
+        S /= self.scaling[adapter_name]
+
+        # Init lora_A and lora_B weights
+        lora_A = V.t().mul(S.sqrt().view(-1, 1)).contiguous()
+        lora_B = U.mul(S.sqrt()).contiguous()
+        self.lora_A[adapter_name].weight.data = lora_A
+        self.lora_B[adapter_name].weight.data = lora_B
+        weight = weight.data - self.scaling[adapter_name] * lora_B @ lora_A
+        weight = weight.to(dtype)
+        self.get_base_layer().weight.data = weight
```
BenjaminBossan, line 304: Here we modify the base weights. This means that if a user uses this CorDA adapter and then loads another LoRA adapter trained on the same base model, that LoRA adapter will not work correctly. This is the same issue we have with PiSSA and OLoRA. To address this, these methods implement a way to convert the adapter into an ordinary LoRA adapter that works with the unmodified base weights, see e.g. here: https://github.com/huggingface/peft/tree/main/examples/pissa_finetuning#convert-pissa-to-lora

Do you think a similar feature could be implemented for CorDA? If this is not planned, we need to document that it's not advisable to use CorDA with other LoRA adapters, and we should probably add a check and emit a warning if we detect this.

src/peft/utils/corda_utils.py
```diff
@@ -0,0 +1,314 @@
+# Copyright 2024-present the HuggingFace Inc. team.
```
BenjaminBossan, line 1: Since `corda_utils.py`  is specific to LoRA, let's move it to `tuners/lora/`. It can also be renamed to just `corda.py`.

BenjaminBossan (changes requested):
I finally had time to read the paper and do an in-depth review. Thanks a lot for this PR, the method looks very intriguing and the implementation is already very mature.

While going through the PR, I found a couple of bigger and smaller issues and commented on them, please check those comments out.

Apart from those, I see 3 bigger areas that will probably need some work before we can merge.

1. Use of a function argument in the config

The `CordaConfig` has the `run_model` argument. This is problematic because it not only is a function and thus not JSON-serializable, but also it closes around the whole dataset so would be quite large if serialized. You already recognized this and remove the argument from the config before serializing it. But I think it's not a sufficient solution, as the serialized config would no longer represent how a user can reproduce the CorDA initialization.

We had a similar discussion when implementing EVA, as it also needs a dataset for initialization. The solution we came up with was to add one [additional step](https://github.com/huggingface/peft/tree/60978d759b2ed91f52e5c155845316a1a3e54986/examples/eva_finetuning#quick-start), which is calling `initialize_lora_eva_weights(peft_model, dataloader)` after the `get_peft_model` call. I would suggest to take the same approach here with CorDA and remove the `run_model` argument from the config. Perhaps you can take a look at the EVA PR (#2142) and take inspiration from the implementation there.

2. Compatibility with other LoRA adapters

As CorDA requires to adjust the base weights of the model, it means that other LoRA adapters that were trained on those base weights will not work properly anymore if they're used in conjunction with CorDA.

I see 2 possible solutions for this:

- Check if multiple adapters are being loaded with at least one CorDA adapter, then give a warning about the issue. Also document this well.
- Add a possibility to convert the CorDA adapter into a normal LoRA adapter that works with the base weights.

The latter is something we offer for PiSSA and OLoRA (requiring to double the LoRA rank), which would otherwise suffer from the same shortcoming. Maybe you can check these implementations and see if the same approach can be taken for CorDA. I think this will add a lot of user convenience.

Some pointers for this:
- https://github.com/huggingface/peft/blob/60978d759b2ed91f52e5c155845316a1a3e54986/src/peft/peft_model.py#L285
- https://github.com/huggingface/peft/blob/60978d759b2ed91f52e5c155845316a1a3e54986/src/peft/peft_model.py#L367
- https://github.com/huggingface/peft/blob/60978d759b2ed91f52e5c155845316a1a3e54986/src/peft/peft_model.py#L380
- https://github.com/huggingface/peft/blob/60978d759b2ed91f52e5c155845316a1a3e54986/src/peft/peft_model.py#L1142
- https://github.com/huggingface/peft/blob/60978d759b2ed91f52e5c155845316a1a3e54986/src/peft/tuners/lora/config.py#L453

3. Testing

Right now, the unit tests are very basic. Ideally, we can extend those further. However, I think it's fine to wait with this until we have figured out the other details about the PR.

> The three methods are compared as follows:

Thanks for summarizing the distinction between the 3 methods. I think this would be a great addition to have in the docs. Could you please add this?

src/peft/tuners/lora/layer.py
iboing: Yes, you are right! It also works for CorDA in the same way as PiSSA and OLoRA (double the lora rank). We are integrating this feature.

src/peft/utils/corda_utils.py
```diff
@@ -0,0 +1,314 @@
...
+    out_dim = w.size(0)
+    in_dim = w.size(1)
+    min_dim = min(in_dim, out_dim)
+
+    if not hasattr(linear, "covariance_matrix"):
+        raise ValueError("build covariance matrix with calib_cov_distribution first")
+    covariance_matrix = linear.covariance_matrix.float()
+
+    damp = 0.01
+    while True:
```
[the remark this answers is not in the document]
iboing: Yes, this is used when covariance matrix is not invertible to add compensation on its diagonal until it is invertible to ensure the same inference as the original model before fine-tuning. 

When a matrix is square and invertible, using `pinv` returns nearly the same inverse result as `inv`, but may have larger numerical error than `inv`. When a square  matrix is not invertible, `pinv` and `inv` return different results. Currently `inv` works well in our experiments.  Can we leave it as it is for now, because we need more investigation to see the impact of using `pinv` on the final adaptation performance?

examples/corda_finetuning/README.md
```diff
@@ -0,0 +1,128 @@
+# CorDA: Context-Oriented Decomposition Adaptation of Large Language Models for Task-Aware Parameter-Efficient Fine-tuning
+
+## Introduction
```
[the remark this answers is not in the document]
iboing: I have updated this part, offering guidance based on users' point of view and removing details. Please let me know if you have further suggestions. 

>  this would be a great addition to have in the docs.

Of course, I have added the summarized comparison of the three methods in the Introduction. Please let me know if you actually want me to add this in docs/source/developer_guides/lora.md

examples/corda_finetuning/preprocess.py
iboing: sorry, this is useless and can be removed. @5eqn will remove this argument.

src/peft/tuners/lora/layer.py
BenjaminBossan: Nice, this will be very useful to have.

src/peft/utils/corda_utils.py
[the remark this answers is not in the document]
BenjaminBossan: Yes, of course you can leave it as is. I was just curious if you had checked if it works.

examples/corda_finetuning/README.md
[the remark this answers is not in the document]
BenjaminBossan: Thanks for the updates. I think this looks good as it is.

examples/corda_finetuning/corda_finetuning.py
```diff
@@ -0,0 +1,265 @@
+import copy
```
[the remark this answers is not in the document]
5eqn: Sure, I've added copyright to all 3 code files in corda_finetuning.

5eqn:
@BenjaminBossan Thanks for your detailed review!

I've removed `run_model` and `hooked_model` from `CordaConfig`, user should pass the model runner to `preprocess_corda` *before* calling `get_peft_model` as described in updated documents. This is different from EVA where the model weights are initialized *after* calling `get_peft_model`, but these two methods should all address the config serialization problem you've mentioned. Example usages are updated synchronously.

Also, I've gone through all your comments, please check my code edits and replies.

5eqn:
> @5eqn Thanks so much for the updates. LMK if you want to get a review now or if I should wait until you finish the remaining changes. Also, could you please fix the small merge conflict?

@BenjaminBossan Remaining changes are pushed, I've added the ability to convert CorDA weights to LoRA, and added supplementary tests for this feature. The code is ready for review.

src/peft/tuners/lora/config.py
```diff
@@ -120,6 +120,113 @@ def __post_init__(self):
             raise ValueError("`tau` must be between 0.0 and 1.0.")
 
 
+@dataclass
+class CordaInitConfig:
```
BenjaminBossan, line 124: I have a suggestion: Can we completely remove the `CordaInitConfig`? It should not be serialized anyway when we save the model, so having a config is not really useful. And the `run_model` and `hooked_model` arguments can be directly passed to `preprocess_corda`, right?

docs/source/developer_guides/lora.md
```diff
@@ -54,6 +54,33 @@ lora_config = LoraConfig(init_lora_weights="pissa_niter_[number of iters]", ...)
...
+[CorDA](https://arxiv.org/pdf/2406.05223) builds task-aware LoRA adapters from weight decomposition oriented by the context of downstream task to learn (instruction-previewed mode, IPM) or world knowledge to maintain (knowledge-preserved mode, KPM).
+The KPM not only achieves better performance than LoRA on fine-tuning tasks, but also mitigates the catastrophic forgetting of pre-trained world knowledge.
+When preserving pre-trained knowledge is not a concern, 
+the IPM is favored because it can further accelerate convergence and enhance the fine-tuning performance. 
+
+You need to configure the initialization method to "corda", and specify the mode of IPM or KPM and the dataset to collect covariance matrices. 
+
+```py
+init_config = CordaInitConfig(
+    run_model=run_model,  # The callback to run your model with sample dataset
```
BenjaminBossan, line 68: I suggested below to completely remove the need for `CordaInitConfig` and to pass the `run_model` function directly to `preprocess_corda`. Still, for this small example, it could be illuminating for the reader to show how a typical `run_model` is defined.

examples/corda_finetuning/README.md
```diff
@@ -0,0 +1,188 @@
...
+|PiSSA | Llama-2-13b |60.88	| 11.08|
+| **CorDA (IPM)** | Llama-2-13b | **62.47** |**11.54** |
+|LoRA| Gemma-2-9b | 83.47 |	42.30 |
+|PiSSA | Gemma-2-9b | 84.23	| 43.52|
+| **CorDA (IPM)** | Gemma-2-9b | **84.45** | **43.88** |
+
+
+## Quick Start
+
+- Knowledge-preserved adaptation mode
```
BenjaminBossan, line 68:
```suggestion
#### Knowledge-preserved adaptation mode
```

Let's use subsections so that we later link this if necessary.

examples/corda_finetuning/README.md
```diff
@@ -0,0 +1,188 @@
...
+    model=peft_model,
+    args=training_args,
+    train_dataset=dataset,
+    tokenizer=tokenizer,
+)
+trainer.train()
+peft_model.save_pretrained("corda-llama-2-7b")
+```
+
+- Instruction-previewed adaptation mode
```
BenjaminBossan, line 120:
```suggestion
#### Instruction-previewed adaptation mode
```

examples/corda_finetuning/README.md
```diff
@@ -0,0 +1,188 @@
...
+    --first_eigen --calib_dataset "MetaMATH"
+```
+
+`corda_finetuning.py`: This script fine-tunes the preprocessed model built above on a downstream task.
+
+Example usage:
+
+```bash
+python corda_finetuning.py \
+    --model_name_or_path {path_to_residual_model} \
```
BenjaminBossan, line 158: Just to be clear, this should be the same as `your_save_path` above, right? Let's use the same placeholder name to make it clear.

examples/corda_finetuning/corda_finetuning.py
```diff
@@ -0,0 +1,265 @@
...
+
+    data_collator = DataCollatorForSupervisedDataset(tokenizer=tokenizer)
+    data_module = {
+        "train_dataset": train_dataset,
+        "data_collator": data_collator,
+    }
+    trainer = Trainer(model=model, tokenizer=tokenizer, args=script_args, **data_module)
+    model.config.use_cache = False
+    trainer.train()
+    trainer.save_state()
```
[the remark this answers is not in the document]
BenjaminBossan: Reminder of this comment.

examples/corda_finetuning/datautils.py
```diff
@@ -0,0 +1,282 @@
...
+    elif name == "nqopen":
+        traindata = load_dataset("nq_open", split="train")
+        tot_text = "\n\n".join(traindata["question"])
+    elif name == "alpaca":
+        # this is for chat models
+        data_path = "data/alpaca_data.json"
+        list_data_dict = jload(data_path)
+        traindataset = []
+        selected_data_dict = random.sample(list_data_dict, nsamples)
+        # random_indices = np.random.choice(len(list_data_dict), nsamples, replace=False)
```
BenjaminBossan, line 164: Here and below, there are a few commented lines. Can those be removed?

examples/corda_finetuning/datautils.py
```diff
@@ -0,0 +1,282 @@
...
+        tot_text = "\n\n".join(traindata["sentence"])
+    elif name == "traivia_qa":
+        traindata = load_dataset("trivia_qa", "rc", split="train")
+        tot_text = "\n\n".join(traindata["question"])
+    elif name == "nqopen":
+        traindata = load_dataset("nq_open", split="train")
+        tot_text = "\n\n".join(traindata["question"])
+    elif name == "alpaca":
+        # this is for chat models
+        data_path = "data/alpaca_data.json"
```
BenjaminBossan, line 160: Here and below, it appears that we assume that `data/alpaca_data.json` etc. already exist. Can we instead load them from HF Hub with the `datasets` package?

src/peft/tuners/lora/config.py
```diff
@@ -360,6 +474,15 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    corda_config: Union[CordaConfig, dict] = field(
```
BenjaminBossan, line 477: I think it's clearer if we make this argument `Optional`  and initialize it as `None` by default (below we already check for `None`).

BenjaminBossan (changes requested):
Thanks a lot for these latest changes. This is a nice improvement to the CorDA integration in PEFT.

I have still found a couple of minor issues. Also, 2 bigger points:

1. I added a suggestion to completely get rid of `CordaInitConfig`, please check if that makes sense.
2. I could not run the included example as is, because it assumes that the data is already downloaded. Ideally, the scripts should run for the users when they follow the instructions. One solution would be to load the data from HF Hub, another to add an instruction on how and where to download the data.

5eqn:
@BenjaminBossan Thank you so much for your review!

> I added a suggestion to completely get rid of CordaInitConfig, please check if that makes sense.

It absolutely makes sense, I've changed the code so that `preprocess_corda` accepts `run_model` and `hooked_model` separately as optional arguments.

> I could not run the included example as is, because it assumes that the data is already downloaded. Ideally, the scripts should run for the users when they follow the instructions. One solution would be to load the data from HF Hub, another to add an instruction on how and where to download the data.

Sure, I've changed the code so that it directly loads dataset from @iboing's [CorDA collection](https://huggingface.co/collections/iboing/corda-66918a95105d80e9e131012f).

iboing:
I have added some argument explanation for the advanced usage in Readme, and will complement the command for `convert corda to lora` in readme soon. It is ready for re-review.

5eqn:
The command for converting CorDA to LoRA is added in the example README.

BenjaminBossan (changes requested):
Thanks for the updates, this PR is almost ready to be merged.

I only have a couple of comments about some documentation and test coverage, the rest LGTM.

Thanks also for updating the example to now use the uploaded datasets, this makes it much more usable. I could successfully go through the steps. A minor roadblock I had was that I could not run the preprocessing with `meta-llama/Llama-2-7b-hf` because it required more than 24 GB of memory. This was a bit surprising to me, given that it's loaded in fp16. Do you think there could be a way to improve memory usage? If not, it's also fine, just wanted to bring this up in case you have some ideas.

Regarding the failing CI, it appears like the failures are unrelated to this PR. Locally, the CorDA tests passed. I'll restart the CI at a later point.

examples/corda_finetuning/README.md
```diff
@@ -0,0 +1,243 @@
...
+|Pre-trained|Gemma-2-9b|12.85|-|-|-|
+|LoRA|Gemma-2-9b| 9.28 | 83.47 | 42.30| 45.02 |
+|**CorDA (KPM)** |Gemma-2-9b|**10.17** | **84.08** | **42.64** | **45.63** |
+
+- Performance with instruction-previewed mode (sample from Math, fine-tune on Math)
+
+| Method | Model | GSM8k | Math |
+| --- | --- | --- | ---|
+|LoRA| Llama-2-7b | 42.68 | 5.88 |
+|CorDA | Llama-2-7b | 51.63 | 7.32 |
```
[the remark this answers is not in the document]
iboing: it is also the result of PiSSA, it is a mistake in the [this update](https://github.com/huggingface/peft/pull/2231/commits/f05256584dbd91ef047a2d36d59ccd66d1f25572)

examples/corda_finetuning/README.md
```diff
@@ -0,0 +1,243 @@
...
+|KPM | questions from the knowledge benchmark to maintain | $A=\sqrt{\Sigma}\_{[-r:]}(V^T C^{-1})\_{[-r:,:]}$ | $B=U_{[:,-r:]}\sqrt{\Sigma}_{[-r:]}$ |
+IPM | instructions and responses from the downstream task to learn | $A= \sqrt{\Sigma}\_{[:r]} (V^T C^{-1})\_{[:r,:]}$ | $B =U_{[:,:r]} \sqrt{\Sigma}_{[:r]}$ |
+
+### Comparison with alternative methods
+
+The distinction between CorDA with other similar LoRA initialization methods is summarized as follows:
+
+| Method | Initialization for | SVD on | Data-driven | Support knowledge maintenance |
+| - | - | - | - | - |
+| CorDA | $A$ and $B$ | weights | no | no |
```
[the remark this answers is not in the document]
5eqn: Sorry, this change is unintended, I've changed it back.

examples/corda_finetuning/README.md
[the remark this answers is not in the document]
5eqn: Sorry, this change is unintended, I've changed it back.

examples/corda_finetuning/README.md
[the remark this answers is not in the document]
iboing: sorry it is a mistake in [this update](https://github.com/huggingface/peft/pull/2231/commits/f05256584dbd91ef047a2d36d59ccd66d1f25572) replacing ''pissa'' with ''corda''. We will fix this.

src/peft/tuners/lora/config.py
```diff
@@ -120,6 +120,89 @@ def __post_init__(self):
...
+        corda_method (`Literal["ipm", "kpm"]`):
+            Method to build adapter. The KPM (Knowledge-Preserved Mode) not only achieves better performance than LoRA
+            on fine-tuning tasks, but also mitigates the catastrophic forgetting of pre-trained world knowledge. When
+            preserving pre-trained knowledge is not a concern, the IPM (Instruction-Previewed Mode) is favored because
+            it can further accelerate convergence and enhance the fine-tuning performance.
+        verbose (`bool`):
+            If true, prints the progress of CorDA initialization.
+        use_float16_for_covariance (`bool`):
+            If true, uses float16 for the covariance matrix. This can reduce the memory usage of the covariance matrix
+            by half, but may lead to numerical instability.
```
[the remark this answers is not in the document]
5eqn: Sure, I've added default values for all arguments.

tests/test_initialization.py
```diff
@@ -329,6 +331,374 @@ def test_lora_pissa_linear_init_default(self, data):
         peft_model = get_peft_model(deepcopy(model), config)
         assert torch.allclose(output, peft_model(data)[0], atol=1e-06)
 
+    def test_lora_corda_linear_init_default(self, data, tmp_path):
```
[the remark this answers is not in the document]
5eqn: Sure, I've extracted all CorDA tests to `class TestCordaInitialization`, and parametrized all tests over two CorDA methods. Note that when data is not enough, KPM fails to build an invertible covariance, causing KPM not to change base weights at all, so I've changed the `data` fixture so that it contains `1000` samples instead of `10` in `class TestCordaInitialization`.

src/peft/tuners/lora/corda.py
```diff
@@ -0,0 +1,345 @@
...
+
+    for name, module in target_modules(hooked_model, config):
+        module.covariance_matrix = 0
+        module.mean = 0
+        module.std = 0
+        module.register_forward_hook(hook)
+
+    run_model()
+
+    if hooked_model is not model:
```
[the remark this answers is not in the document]
5eqn: Sure, I've also added a unit test for this.

iboing:
Hi @BenjaminBossan  , regarding the memory consumption, the covariance is collected in `torch.float32` by default. To improve memory usage, users can specify `use_float16_for_covariance=True` in `CordaConfig` to collect covariance matrices in `torch.float16`. But this may cause numerical instability only in a few cases, such that the initialized model does not ensure the exact same inference result as the original model. Users are suggested to check, e.g., comparing the inference result of Wiki/PTB perplexity before and after preprocessing, if they choose to perform in `torch.float16`. 

I have added this note about memory in the doc in [this update ](https://github.com/huggingface/peft/pull/2231/commits/9f7a698d0f89d8ebf3dcbb031de05f3aaed0d3ce).

src/peft/tuners/lora/corda.py
```diff
@@ -0,0 +1,345 @@
...
+
+    Args:
+        model (`nn.Module`):
+            Model to preprocess.
+        lora_config (`LoraConfig`):
+            Lora configuration of the model. `lora_config.corda_config` should be set.
+        run_model (`Callable[[], None]`):
+            Callback to run the model when building covariance. This will be run once regardless of `sample_count`, so
+            you should configure the `run_model` callback to run the model exactly `sample_count` times. Typically you
+            should run model inference on your dataset in this callback.
```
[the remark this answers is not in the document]
5eqn: Sure, the comment is updated.

5eqn:
I've pushed a commit that calculates `sample_size` automatically in `preprocess_corda`, so that the user no longer needs to specify it in `preprocess_corda`. The document of `preprocess_corda` has updated accordingly.

BenjaminBossan:
Sorry for the delay in reviewing, we had a release past Friday and are fighting the CI today, hopefully I'll have time tomorrow.

src/peft/tuners/lora/model.py
```diff
@@ -269,11 +269,13 @@ def _replace_module(self, parent, child_name, new_module, child):
                 weight = (
                     child.qweight
                     if hasattr(child, "qweight")
-                    else child.W_q
-                    if hasattr(child, "W_q")
-                    else child.weight
-                    if hasattr(child, "weight")
-                    else next(child.parameters())
+                    else (
```
BenjaminBossan, line 272: Why this change? This can be reverted, right?

tests/test_initialization.py
```diff
@@ -1763,6 +1765,471 @@ def new_state_dict():
...
+
+    def get_uniform(self, amin, amax, size=(10000,)):
+        unif = torch.distributions.uniform.Uniform(amin, amax)
+        samples = unif.sample(size)
+        return samples
+
+    def get_normal(self, mean, std, size=(10000,)):
+        normal = torch.distributions.normal.Normal(mean, std)
+        samples = normal.sample(size)
+        return samples
```
BenjaminBossan, line 1781: These are not used and can be removed.

tests/test_initialization.py
```diff
@@ -1763,6 +1765,471 @@ def new_state_dict():
...
+                self.linear = nn.Linear(1000, 1000)
+
+            def forward(self, x):
+                return self.linear(x)
+
+        return MyModule().eval().to(self.torch_device)
+
+    @pytest.fixture
+    def data(self):
+        # larger data is required to pass KPM test
```
BenjaminBossan, line 1797: Let's set a seed to ensure that the tests are not flaky.

tests/test_initialization.py
```diff
@@ -1763,6 +1765,471 @@ def new_state_dict():
...
+            init_lora_weights="corda",
+            target_modules=["linear"],
+            corda_config=CordaConfig(covariance_file=tmp_path / "covariance_cache.pt", corda_method=corda_method),
+        )
+        preprocess_corda(model, config)
+        peft_model = get_peft_model(model, config)
+        assert torch.allclose(output, peft_model(data)[0], atol=1e-06)
+
+    @pytest.mark.parametrize("corda_method", ("ipm", "kpm"))
+    def test_lora_corda_hooked_model_linear_init_default(self, data, tmp_path, corda_method):
```
BenjaminBossan, line 1848: Could you please add a comment on how this test differs from the previous one.

tests/test_initialization.py
```diff
@@ -1763,6 +1765,471 @@ def new_state_dict():
...
+
+        return MyModule().eval().to(self.torch_device)
+
+    @pytest.fixture
+    def data(self):
+        # larger data is required to pass KPM test
+        return torch.rand(1000, 1000).to(self.torch_device)
+
+    @pytest.mark.parametrize("corda_method", ("ipm", "kpm"))
+    def test_lora_corda_linear_init_default(self, data, tmp_path, corda_method):
```
BenjaminBossan, line 1801: Checking this test, we're always asserting that the CorDA outputs are identical to the output of the base model. I know that this is expected, but it makes the test a bit weak, since if CorDA just did nothing, the test would still pass. Is there an additional check that could be performed to ensure that `preprocess_corda` is not just a no-op?

examples/corda_finetuning/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+
+#### Note about memory consumption 
+
+The process of collecting covariance matrices is performed in `torch.float32` by default. If you would like to reduce the memory consumption of preprocessing, you can specify `use_float16_for_covariance=True` in `CordaConfig` to collect covariance matrices in `torch.float16`. But this may cause numerical instability only in a few cases, such that the initialized model does not ensure the exact same inference result as the original model. So it is suggested to check, e.g., comparing the inference result of Wiki/PTB perplexity before and after preprocessing, if you choose to perform in `torch.float16`. 
+
+
+
+
+
```
BenjaminBossan, line 182: Let's remove those empty lines.

examples/corda_finetuning/README.md
```diff
@@ -0,0 +1,252 @@
...
+Arguments:
+
+- `--first_eigen` uses the largest $r$ singular values and vectors to initialize the learnable adapter for the instruction-previewed adaptation mode. 
+- `--calib_dataset` specifies the dataset to sample data to obtain covariance matrices. Instruction-previewed mode uses the downstream task dataset you are learning, such as  `"MetaMATH"`, `"codefeedback"`, `"WizLMinstruct"`, `"alpaca"`, or other choices.
+
+
+#### Note about memory consumption 
+
+The process of collecting covariance matrices is performed in `torch.float32` by default. If you would like to reduce the memory consumption of preprocessing, you can specify `use_float16_for_covariance=True` in `CordaConfig` to collect covariance matrices in `torch.float16`. But this may cause numerical instability only in a few cases, such that the initialized model does not ensure the exact same inference result as the original model. So it is suggested to check, e.g., comparing the inference result of Wiki/PTB perplexity before and after preprocessing, if you choose to perform in `torch.float16`.
```
BenjaminBossan, line 177: How about bfloat16?

src/peft/tuners/lora/corda.py
```diff
@@ -0,0 +1,347 @@
...
+    Build necessary CorDA fields for a model.
+
+    Args:
+        model (`nn.Module`):
+            Model to preprocess.
+        lora_config (`LoraConfig`):
+            Lora configuration of the model. `lora_config.corda_config` should be set.
+        run_model (`Optional[Callable[[], None]]`):
+            Callback to run the model when building covariance. Typically you should run model inference on your sample
+            dataset in this callback. The size of sample dataset is typically 256. `run_model` can be `None` only if
```
BenjaminBossan, line 71: > The size of sample dataset is typically 256.

Is this so?

examples/corda_finetuning/README.md
```diff
@@ -0,0 +1,252 @@
+# CorDA: Context-Oriented Decomposition Adaptation of Large Language Models for Task-Aware Parameter-Efficient Fine-tuning
+
+## Introduction
+
+
+Existing PEFT methods are mostly agnostic of the context of a task of concern, e.g., a downstream task to learn or some pre-trained world knowledge to maintain.
+[CorDA](https://openreview.net/pdf?id=Gi00NVru6n) builds task-aware LoRA adapters from weight decomposition oriented by the context of the task concerned. 
+
+Concretely, CorDA randomly collects a few (by default 256) data samples from a target task, e.g. questions from a QA dataset or instructions to write a code or solve a math problem, and feeds these samples into a pre-trained LLM. We can obtain the covariance matrix of the input activation of each linear layer, i.e., $C=XX^T\in\mathcal{R}^{d_{in}\times d_{in}}$.
```
BenjaminBossan, line 9: > (by default 256)

This is no longer true, is it?

BenjaminBossan (changes requested):
Thanks for the recent updates. This looks very good and the test coverage is now also excellent. I've reviewed the changes and basically the only comments left are a bit of code documentation and clean ups.

> I've pushed a commit that calculates sample_size automatically in preprocess_corda, so that the user no longer needs to specify it in preprocess_corda.

Thanks for handling this automatically. Honestly, I haven't fully understood how/why this new mechanism for sample count determination works. Perhaps a short comment on the code on how/why this works could be helpful.

5eqn:
@BenjaminBossan Thanks for your review! I've read your comments, the code will be updated later.

examples/corda_finetuning/README.md
iboing: We did not test bfloat16. Actually the numerical instability when using float16 only occurs in a few cases.

examples/corda_finetuning/README.md
iboing: It is still 256 in our provided example, see [this argument](https://github.com/huggingface/peft/blob/5a06739573281bd2e4586cbed1ebe35f6ca12d61/examples/corda_finetuning/preprocess.py#L111). For user's own preprocessing, it will be the size of the dataloader provided by users. I have changed "by default 256" into "by default 256 in our `preprocess.py`".

iboing: `calib_loader_size` determines the size of the calibration dataset (sample count) in our `preprocess.py`. 

For users, they provide the calibration dataloader themselves, and pass it to  `run_model` in [this line ](https://github.com/huggingface/peft/blob/d5f0a44fecfeb1df6cec425b34877dfab39a1e66/examples/corda_finetuning/preprocess.py#L73). In our previous version, the users need to specify `sample_count` as an argument in `CordaConfig` and need to ensure it equals to the size of the dataloader they provide. So we remove this argument in [this update](https://github.com/huggingface/peft/pull/2231/commits/5a06739573281bd2e4586cbed1ebe35f6ca12d61), and calculate `sample_count` automatically in [this line](https://github.com/huggingface/peft/blob/d5f0a44fecfeb1df6cec425b34877dfab39a1e66/src/peft/tuners/lora/corda.py#L182C1-L182C9). As a result, uses only need to pass a dataloader with a determined `calib_loader_size `, and do not need to specify the same sample count again in CordaConfig.

src/peft/tuners/lora/corda.py
5eqn: We've experimented with sample size of 32 and 256, the latter (256) performs better. Larger sample size does not evidently improve performance, and comes with more computational cost.

src/peft/tuners/lora/model.py
5eqn: You're right, this change is due to my formatter settings. I've reverted the change.

tests/test_initialization.py
5eqn: You're right, I've removed these lines.

tests/test_initialization.py
5eqn: Sure, I've added `torch.manual_seed(233)` and made sure all CorDA tests pass locally.

tests/test_initialization.py
5eqn: You're right, I've modified the tests so that it doubles LoRA weight, and check if output is different from original output. This way if CorDA is no-op, the weights are initialized zero and the tests would not pass.

tests/test_initialization.py
5eqn: Sure, I've added a comment in line 1868.

5eqn:
Hi @BenjaminBossan , we've updated the code according your comments. The code is now ready for review.

paischer101:
Hi @iboing @5eqn, this is very interesting work!
I am a first author of the EVA paper and we noticed that in the table in your [README](https://github.com/huggingface/peft/blob/9d4a5200dd340fecd89b9be4178825095b18c508/examples/corda_finetuning/README.md#comparison-with-alternative-methods) you specified EVA as not being able to maintain knowledge of the base model.

We (tagging @sirluk here) think that this might be a mistake, as with EVA we can simply recover the base model performance by unloading the adapters. This is not the case for PiSSA or CorDA as they alter the base model's weights during initialization so they are subject to change during fine-tuning, i.e. you cannot restore the base model weights as they change during fine-tuning. By initializing with the components that contain the least amount of information this can be mitigated as shown in your work. However, we intentionally followed the design of LoRA (B = 0), which allows us to fully recover the base model by unloading adapter weights, i.e. full knowledge maintenance of the base model. Therefore, we believe the table should be altered to say "yes" for EVA in the knowledge maintenance column. What are your thoughts @iboing @5eqn?

We also updated our manuscript on arXiv and added a reference to CorDA :)
Also many thanks to @BenjaminBossan as he made us aware of this pull request.

iboing:
Hi @paischer101 , many thanks for your attention. I think I have made some explanations when you just came by my poster at NeurIPS'24. 

Actually PiSSA, CorDA, OLoRA have the ability of unloading the adapters and keeping the base weights unchanged. Concretely for PiSSA and CorDA, the weight change is $\Delta W=AB-A_0 B_0$, and we can keep the base weight unchanged and set the lora adapter as [ $A$, $A_0$] and [ $B$, $-B_0$] , which doubles the rank. So, all these methods can unload the adapters and keep base weights unchanged. 

Here "maintain knowledge" does not refer to the ability to keep base weights unchanged. LoRA and PiSSA are task-agnostic. Your method EVA is data-dependent considering the task for finetuning, but seems to not include the task awareness of pre-trained knowledge to maintain. I do not find any results in your paper about preserving knowledge. Our method explicitly associate a knowledge benchmark with the pricipal components after our decomposition and keep them frozen in KPM. This is what "maintain knowledge" means in our context. So, we mark "no" in that column. 

We can remove this table but it is up to the suggestion of @BenjaminBossan

BenjaminBossan:
Thanks for the constructive discussion @paischer101 and @iboing. From my point of view, this looks like a question of terminology. If we can keep the table, but add a sentence to clarify what knowledge preservation means, that should be enough. I think the idea is already explained well elsewhere, but for readers who directly jump to the table, it could make sense to add a short description below that explains what is meant be "Support knowledge maintenance" (which btw. should be "Supports").

iboing:
Hi @BenjaminBossan  , thanks for your suggestion. I have added a sentence of explanation below the table. Please let us know if there is further necessary change for this pr.

src/peft/tuners/lora/corda.py
githubnemo: Maybe a better way of saying this would be

> Experiments have shown 256 samples to be a good default value.

paischer101:
Sorry for the delayed response on my end.
@iboing thanks for clarofication on unloading the weights, you are correct, this is possible for the other inits as well.

Also, I agree with @BenjaminBossan, it is a question of terminology. KPM in CorDA is simply a different use case we did not include in our paper. The terminology of "knowledge preservation" led to this confusion. 
In any case, thanks @iboing for adding the clarifying sentence to clear up such confusions.

src/peft/tuners/lora/corda.py
iboing: @githubnemo  I have updated the saying following your suggestion

src/peft/tuners/lora/corda.py
```diff
@@ -0,0 +1,347 @@
...
+        module.sample_count = 0
+        module.covariance_matrix = 0
+        module.mean = 0
+        module.std = 0
+        module.register_forward_hook(hook)
+
+    run_model()
+
+    # In some edge cases you might need to hook a model different from the model to add adapters,
+    # this case you would specify `hooked_model` and set it to a different model from `model`.
```
githubnemo, line 200: Do you have a specific example for this case? If so, it would be worth a ton to mention it in the comment :)

src/peft/tuners/lora/corda.py
```diff
@@ -0,0 +1,347 @@
...
+
+        # free memory
+        del covariance, input
+
+    for name, module in target_modules(hooked_model, config):
+        module.sample_count = 0
+        module.covariance_matrix = 0
+        module.mean = 0
+        module.std = 0
+        module.register_forward_hook(hook)
```
githubnemo, line 195: Correct me if I'm wrong but these hooks are never unregistered, right? So the forward hook would be needlessly run during training time as well. In that case, it is probably worthwhile to save the handles returned by `register_forward_hook` and run `handle.remove()` afterwards.

BenjaminBossan:
@iboing Could you please run `make style`? The doc builder is complaining, probably because of the long line.

githubnemo (commented):
I have two minor comments still but otherwise LGTM :)

src/peft/tuners/lora/corda.py
5eqn: This code exists for a very subtle case in our future work related to CorDA, providing the example would introduce many new concepts, and we believe specifying `hooked_model` is very unlikely needed for users, so we won't provide a specific example. If the user actually needs to obtain covariance on a model different from the model to add adapters, just specify `hooked_model` as the model to obtain covariance, we believe this is intuitive and does not need furthur comments. Does this clear up your confusion?

src/peft/tuners/lora/corda.py
5eqn: Great catch! I've added the logic to remove hooks, and corresponding test in `test_initialization.py` function `test_lora_corda_hook_unregister`.

5eqn:
@githubnemo @BenjaminBossan Thank you for your review! I've added handle removal as proposed by @githubnemo , and also fixed the logic of automatically dividing `sample_count` after collecting covariance. Corresponding tests are appended.

BenjaminBossan (approved):
The latest changes LGTM. Thanks a lot for this PR and for your patience. From my point of view, this is ready to be merged. I'll wait if there are any comments left by @githubnemo before merging.

githubnemo:
I think this is good to go :)
Thank you for the great work everyone!

PR #2238, opened 2024-11-27
FIX: Small regression in BNB LoRA output

BenjaminBossan:
Our regression tests reveal that the 8bit LoRA BNB regression test is [failing](https://github.com/huggingface/peft/actions/runs/12042604161/job/33576548074#step:8:254). To reproduce, run:

`pytest tests/regression/test_regression.py -s --regression -k test_lora_8bit`

The regression was introduced in #2122. We didn't notice this earlier because of other failing tests in the nightly CI.

The cause of the error is subtle. In the original code, we would calculate the LoRA output, convert the `dtype` if necessary, then add it to the base output. After the mentioned PR, we calculate the LoRA output, add it to the base output, then convert the `dtype` if necessary ([code](https://github.com/huggingface/peft/pull/2122/files#diff-d6321d6b9970ada644811f3992899f3ef7df4581fced2a7584e0a51b03cc3355)). The difference is very small on a per layer basis, but it can accumulate over the layers, leading to a significant difference in outputs, as witnessed by the regression test.

This PR rolls back this specific part of the PR (both for 8bit and 4bit) while leaving the main change of that PR intact.

### DIFF
--- a/src/peft/tuners/lora/bnb.py
+++ b/src/peft/tuners/lora/bnb.py
@@ -235,7 +235,7 @@ if is_bnb_available():
                             x = x.to(compute_dtype)
 
                     if not self.use_dora[active_adapter]:
-                        result = result + lora_B(lora_A(dropout(x))) * scaling
+                        output = lora_B(lora_A(dropout(x))) * scaling
                     else:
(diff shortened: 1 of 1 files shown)


ariG23498 (approved):
Very good catch.

Issue #2065, opened 2024-09-13
Merge LoRA into 405B


junzhang-zj:
Env: 8 A100-80G GPUs, Transformer: 4.43.3, torch: 2.4.0+cu121

Goal: Get a merged INT8 405B

Path 1: 
1) Load IN8 405B & BF16 LoRA --> Merge --> Save

Path 2:
1) Load BF16 405B & BF16 LoRA --> Merge --> Save
2) Load INT8 merged 405B --> Save

When I tried to merge LoRA into LLaMA-3.1 405B with load_in_8bit=True, I got an error in BNB; 
so I tried to use CPU and merge 405B with bf16, and it was still killed suddenly.

BenjaminBossan:
Could you please provide more information: What code did you use to merge? What is the full error message that you got for both tests you did?

junzhang-zj:
Thanks, for your prompt reply!

For case 2, no error was received and it was killed.

My test code:

`
import torch
from peft import PeftModel
import transformers
from transformers import AutoTokenizer, AutoModelForCausalLM

model_name_or_path = "llama3/Meta-Llama-3.1-405B"
lora_path = ""
output_path = ""

base_model = AutoModelForCausalLM.from_pretrained(
    model_name_or_path,
    load_in_8bit=False,
    torch_dtype=torch.bfloat16, 
    device_map={"": "cpu"} 
)

lora_model = PeftModel.from_pretrained(
    base_model,
    lora_path,
    torch_dtype=torch.bfloat16,
    device_map={"": "cpu"}
)


model = lora_model.merge_and_unload(progressbar=True)
model.save_pretrained(output_path) 
base_tokenizer = AutoTokenizer.from_pretrained(model_name_or_path)
base_tokenizer.save_pretrained(output_path)
`

junzhang-zj:
For case 1, I made a device_map allocation in advance.

Return: Error an illegal memory access was encountered at line 524 in file /mmfs1/gscratch/zlab/timdettmers/git/bitsandbytes/csrc/ops.cu

num_gpus = 8
num_layers = 126

# Specify how many layers GPU 0 should handle
layers_for_gpu0 = num_layers // num_gpus // 2
remaining_layers = num_layers - layers_for_gpu0
layers_per_remaining_gpu = remaining_layers // (num_gpus - 1)
extra_layers = remaining_layers % (num_gpus - 1)

# Assign layers to GPU 0
for layer in range(layers_for_gpu0):
[...]

PR #2245, opened 2024-12-02
ENH: Updates for upcoming BNB Int8 release

matthewdouglas:
This PR contains changes to PEFT that relate to an upcoming 0.45.0 release that we are planning for bitsandbytes. Primarily the changes are related to [bitsandbytes#1401](https://github.com/bitsandbytes-foundation/bitsandbytes/pull/1401).

1. Bitsandbytes will introduce a public API function for LLM.int8() weight dequantization. We use this API when available, and otherwise use a functionally equivalent implementation. Apart from its relative simplicity, this will also resolve issues such as #2065.
2. The `memory_efficient_backward` kwarg for `Linear8bitLt` will be removed in the upcoming release, as per a longstanding [deprecation notice](https://github.com/bitsandbytes-foundation/bitsandbytes/blob/7dca70049566b5b1c55cbd67e1cb191729a98152/bitsandbytes/nn/modules.py#L908). This PR removes usage of the kwarg for compatibility.

### DIFF
--- a/src/peft/tuners/adalora/model.py
+++ b/src/peft/tuners/adalora/model.py
@@ -174,7 +174,6 @@ class AdaLoraModel(LoraModel):
             kwargs.update(
                 {
                     "has_fp16_weights": target_base_layer.state.has_fp16_weights,
-                    "memory_efficient_backward": target_base_layer.state.memory_efficient_backward,
                     "threshold": target_base_layer.state.threshold,
                     "index": target_base_layer.index,
                 }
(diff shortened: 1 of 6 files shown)


BenjaminBossan (approved):
Thanks for this PR to update bnb 8bit in PEFT. The changes LGTM and after the internal discussion, this should be good to merge (some regression tests will need adjusting once bnb v0.45.0 is released).

Issue #2252, opened 2024-12-03
 Adapter name conflict with tuner prefix leads to unclear warning during model loading


pzdkn:
### System Info

- peft=0.13.3.dev0
- transformers=4.46.3
- Python=3.12.6

### Who can help?

@BenjaminBossan 
@stevhliu 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [X] My own task or dataset (give details below)

### Reproduction

```python
import peft 
import torch
from pathlib import Path
from transformers import AutoModelForCausalLM, AutoTokenizer
from peft import LoraConfig, TaskType
from tempfile import TemporaryDirectory


def get_adapter_state_dict(model, adapter_name):
    adapter_state_dict = model.state_dict()
    adapter_weights = {key: value for key, value in adapter_state_dict.items() if adapter_name in key}
    return adapter_weights

def test_adapter_name():
    MODEL_ID = "openai-community/gpt2"
    ADAPTER_NAME = "lora"
    
    # Load the base model
    base_model = AutoModelForCausalLM.from_pretrained(MODEL_ID)
    
    # Define LoraConfig without base_model_name_or_path
    peft_config = LoraConfig(
        task_type=TaskType.CAUSAL_LM, 
        inference_mode=False, 
        r=8
    )
    
    # Get the PEFT model with the adapter name
    peft_model = peft.get_peft_model(base_model, peft_config, adapter_name=ADAPTER_NAME)
    
    # Saving and loading the model
    with TemporaryDirectory() as temp_dir:
        temp_dir = Path(temp_dir)

        adapter_weights = get_adapter_state_dict(peft_model, ADAPTER_NAME)
        peft_model.save_pretrained(temp_dir)
        
        loaded_peft_model = peft.PeftModel.from_pretrained(model=base_model,
                                                            model_id= temp_dir / ADAPTER_NAME,
                                                            adapter_name=ADAPTER_NAME)
        loaded_adapter_weights = get_adapter_state_dict(loaded_peft_model, ADAPTER_NAME)
       
        # Assertion fails due to adapter weights been newly intialized
        for key in adapter_weights:
            assert torch.isclose(adapter_weights[key], loaded_adapter_weights[key]).all()

```

### Expected behavior

A clear error or warning message indicating that the adapter_name and tuner_prefix should not be the same.

pzdkn:
[...]

PR #2254, opened 2024-12-03
Fix: Add warning for adapter_name conflict with tuner 

pzdkn:
This pull request addresses an [issue ](https://github.com/huggingface/peft/issues/2252) where adapter weights are incorrectly initialized when the adapter_name conflicts with the tuner_prefix during model loading. Specifically, it introduces a warning message to notify users when this conflict occurs, as this could lead to confusion or unintentional weight reinitialization.

**Changes:**

- Added a warning message when adapter_name matches the tuner_prefix, indicating a potential issue with missing adapter keys during model loading.
- The warning clarifies that this conflict is likely caused by the adapter_name being contained in the tuner_prefix, helping users identify and resolve the issue more easily.

**Impact:**

- Improves user experience by providing clearer feedback when model loading doesn't match expectations, preventing silent failures and making debugging easier.

### DIFF
--- a/src/peft/mapping.py
+++ b/src/peft/mapping.py
@@ -72,6 +72,7 @@ from .tuners import (
 )
 from .tuners.tuners_utils import BaseTuner
 from .utils import _prepare_prompt_learning_config
+from .utils.constants import PEFT_TYPE_TO_PREFIX_MAPPING
 
 
 if TYPE_CHECKING:
@@ -202,6 +203,13 @@ def get_peft_model(
         warnings.warn(
             "lora with eva initialization used with low_cpu_mem_usage=False. "
             "Setting low_cpu_mem_usage=True can improve the maximum batch size possible for eva initialization."
+        )
+
+    prefix = PEFT_TYPE_TO_PREFIX_MAPPING.get(peft_config.peft_type)
+    if prefix and adapter_name in prefix:
+        warnings.warn(
+            f"Adapter name {adapter_name} should not be contained in the prefix {prefix}."
+            "This may lead to reinitialization of the adapter weights during loading."
         )
 
     if mixed:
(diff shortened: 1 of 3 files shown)


BenjaminBossan (commented):
Thanks for this PR. It is certainly not a bad idea to warn users if they come across this issue. I'd go even further and also add a warning when users first create a model with an overlapping adapter name. For that, we'd have to adjust:

- [`get_peft_model`](https://github.com/huggingface/peft/blob/c05758989d1a1bc650ab1f41fe1de34c88838731/src/peft/mapping.py#L146)
- [`add_adapter`](https://github.com/huggingface/peft/blob/c05758989d1a1bc650ab1f41fe1de34c88838731/src/peft/mapping.py#L146)

Would you be interested in working on those too? If not, I can later tackle this in a separate PR.

Regardless of that question, it would also be great to add a unit test for the warning (i.e. that there is no warning on the happy path but a warning when an overlapping name is chosen). [Here](https://github.com/huggingface/peft/blob/c05758989d1a1bc650ab1f41fe1de34c88838731/tests/test_initialization.py#L1756) is an example of a test that could be used as a starting point.

pzdkn:
Yes I'd happy to take that!

pzdkn:
Thanks for the guidance @BenjaminBossan ! I have added the warnings to both creation paths and the loading path plus an unit test.

BenjaminBossan (changes requested):
Thanks a lot for adding the other warnings and for providing the unit tests. In general, this all looks very good. I just have a small request, which is to break the one long test into multiple smaller tests. This improves readability and makes the tests useful (right now, if one assert fails, the checks further below are not executed).

One way is just to split this function into multiple smaller function. This results in a bit of code duplication but that's okay. Another suggestion is to create a class for all these tests. Inside the class, you can have define a pytest fixture for the duplicate code (like the `model` and `config`, check [this test](https://github.com/huggingface/peft/blob/f86522e011af21697ef1477da4d232e74de83232/tests/test_initialization.py#L1514) as an example). I'll leave this decision up to you. LMK if you have any questions.

pzdkn:
Hi thanks for the tips @BenjaminBossan ! Hope this does the job now 😄

pzdkn:
Done :+1:  Happy for another round of review @BenjaminBossan

BenjaminBossan (approved):
Thanks for adding the warnings for this edge case in naming, and for adding the nice unit tests. The PR LGTM.

Please ignore the failing CI, they're unrelated to the PR. We currently have trouble with the CI runners.

Issue #2261, opened 2024-12-05
Make module imports / re-export conforming with typing specs for proper type checker support


bluenote10:
### Feature request

Make peft's modules type checker friendly.

### Motivation

Currently, using peft with a type checker is somewhat painful, because its module re-exports are not conforming to the [typing specs for library interfaces](https://github.com/python/typing/blob/ad4cbc45f868f47031494694bdb7349119302f59/docs/spec/distributing.rst#library-interface-public-and-private-symbols). This means that very basic things do not type check properly. For instance the following is a type error in both pyright and mypy:

```py
from peft import LoraConfig
```

Mypy error is for instance:
```
example.py:1: error: Module "peft" does not explicitly export attribute "LoraConfig"  [attr-defined]
```

This is obviously not great, because the code works fine as runtime. Suppressing all these type errors causes a lot of noise on user side, and even worse, fails to make any benefit from the type annotations.

### Your contribution

As the specs say: _Imported symbols are considered private by default._

There are basically two ways to get proper library interfaces:
- Using an explicit `__all__` in the modules to mark the re-exports.
- Using the patterns shown in the [Import Conventions](https://github.com/python/typing/blob/ad4cbc45f868f47031494694bdb7349119302f59/docs/spec/distributing.rst#import-conventions) section.

Using an explicit `__all__` would also allow you to remove the linter suppressions like this one

https://github.com/huggingface/peft/blob/f86522e011af21697ef1477da4d232e74de83232/src/peft/__init__.py#L1-L3

because the linter would then understand that the symbol is being used for re-export purposes.

BenjaminBossan:
Thanks for bringing this up. I agree that the type annotation situation in PEFT is lackluster, we don't perform any automatic checks. Could you please tell us which type checker you're using (and which config, if any)?

Regarding your suggestion, I'd be fine with defining an `__all__` with all "public" objects in the `__init__.py`. Using `import X as X` looks terrible though. If you're interested, feel free to create a PR and check if the solution solves your issue.

bluenote10:
> Could you please tell us which type checker you're using (and which config, if any)?

As mentioned, I'm using both pyright and mypy, and it is a type error in both. This is mostly with default configs, but I don't think it depends a lot on the config, because that is just a general convention (of course, you could come up with a config that suppresses these errors). Note that these typing specs are not specific to any type checker, but rather _all_ type checkers in the ecosystem should stick to them -- so if a type checker adheres to the standard, it has to report a type error.

[...]

PR #2265, opened 2024-12-07
ENH fix library interface

bluenote10:
This PR fixes peft's library interface to adhere to the typing specification as described in #2261.

The easiest way to fix the interface was basically to remove the existing `flake8: noqa` lints, which exactly have lead to a non-spec conforming interface. After enabling linting of these files, ruff ensures that the symbols _must be used_ e.g. within `__all__`. Thus it is technically impossible to import a symbol, but forget to add it to `__all__`, because ruff will report that.

Note that when enabling linting these files, ruff would also like to re-order the imports, because they are currently not ordered properly. To simplify the reviewing, I have temporarily added a local `# noqa: I001` suppression so that ruff does not _yet_ insist on the proper order. This makes it easier for you to see that the exported symbols have not changed. The only symbol that I have removed is `VBLoRAConfig`, because it was actually appearing twice in the list, and ruff will also complain about the redundant double import.

After merging this PR, I would then open a follow-up PR that removes these `# noqa: I001` suppressions, along with the required import re-orderings. In this follow-up PR the listed symbols in `__all__` will remain stable (I have already sorted them alphabetically). So in that second PR, there is no mental effort required that the re-ordering is correct :wink: -- ruff will ensure that the imported and exported symbols have to match.


BenjaminBossan:
Thanks for the PR and for laying out your plan. This sounds reasonable to me and I'm happy to hear that ruff enforces this, so that the lists are staying up-to-date.

Just one question regarding the sorting: Usually, capital letters are ordered to be before lower case letters. So e.g. for `utils/__init__.py`, I'd expect the list to be `['CONFIG_NAME', 'INCLUDE_LINEAR_LAYERS_SHORTHAND', 'ModulesToSaveWrapper', ...]`. Or is the logic here different for some reason?

bluenote10:
> Just one question regarding the sorting: Usually, capital letters are ordered to be before lower case letters.

To clarify: I believe currently ruff only cares about the order of the imports (`I001`), but I don't think it cares about the order of the exports in `__all__` (there may be another optional rule for that -- but I haven't checked that yet). So there is currently some freedom in how to order the symbols in `__all__`. What I did to avoid having to sort them manually was to rely on the VSCode feature (select the rows, go to command palette, and do "sort lines ascending"). So this just happens to be the ordering semantics of VSCode. If you prefer, I could manually re-order them. However, in case other developers would also use this VSCode feature, it might be easier to go for that way of ordering to allow for automation.

BenjaminBossan:
> I found the ruff rule to order `__all__`

How about for this PR, only order the newly added `__all__`s according to that logic, we can leave the others for a later PR.

bluenote10:
> How about for this PR, only order the newly added __all__s according to that logic, we can leave the others for a later PR.

Yep that sounds good. In fact, I simply temporarily enabled `RUF022`, let ruff do its thing to get the sorting right, and only committed the changes in the `__init__.py` affected by this PR.

PR #2268, opened 2024-12-09
Attempt to fix performance issue on CI runners

githubnemo:
**DO NOT MERGE YET**

MacOS and Windows CI runners / tests are taking ~4x as long as the Linux tests do.

For MacOS there is already an issue that reflects the numbers we get (https://github.com/actions/runner-images/issues/1336) but to make sure that it is not us, removing the lines to install `numpy < 2` for Windows and MacOS helps to debug.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,7 +62,7 @@ jobs:
         run: |
           #make test
           # TODO: also test with --no-gc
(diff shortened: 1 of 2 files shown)


githubnemo:
With https://github.com/huggingface/transformers/pull/35236 merged this is no longer needed.

PR #2276, opened 2024-12-12
FIX Adoption prompt: New way to obtain position embeddings

BenjaminBossan:
This PR resolves the [failing adoption prompt tests](https://github.com/huggingface/peft/actions/runs/12278356833/job/34259845086#step:5:1003) in the CI using transformers installed from source.

In this transformers PR:

https://github.com/huggingface/transformers/pull/34858/

the `module.rotary_emb` attribute has been removed, which adoption prompt so far assumed was present. Instead, the `position_embeddings` are now already computed and can be taken directly from the `kwargs`.


Cyrilvallez (approved):
Makes sense to me! If needed, the `rotary_emb` module can be retrieved from the `Model` module, not the `Attention` module anymore!

Cyrilvallez:
But `cos` and `sin` are guaranteed to be present during the model's forward anyway 👌

githubnemo (approved):

PR #2279, opened 2024-12-13
MAINT apply sorting of imports

bluenote10:
As discussed in #2265, this PR removes the temporarily introduced `# noqa: I001`, fixing the sorting of these imports.


BenjaminBossan (approved):
Thanks for implementing the rule for sorting imports, LGTM. Failing CI is unrelated.

PR #2172, opened 2024-10-23
Support Bone

JL-er:
Re pr https://github.com/huggingface/peft/pull/2148

PR #2282, opened 2024-12-13
Refactor: PEFT method registration function

BenjaminBossan:
This is a refactor of how new PEFT methods are _registered_. It got bigger than I initially expected.

## Goal

The goal of this refactor is the following: Right now, when a new PEFT method is added, a new directory is created in `src/peft/tuners/<name>` with a config, model, etc. This is fine and self-contained.

However, in addition to that, a couple of other places in the PEFT code base need to be touched for this new PEFT method to become usable.

As an example, take the recently added Bone method (#2172). Ignoring tests, docs, and examples, we have the additions to `src/peft/tuners/bone`, but also need to:

1. Add an entry to `PEFT_TYPE_TO_CONFIG_MAPPING` in `mapping.py`.
2. Add an entry to `PEFT_TYPE_TO_TUNER_MAPPING` in `mapping.py`.
3. Add an entry to `PEFT_TYPE_TO_MODEL_MAPPING` in `peft_model.py`
4. Add an entry to `PEFT_TYPE_TO_PREFIX_MAPPING` in `utils/constants.py`
5. Add some code to `get_peft_model_state_dict`in `utils.save_and_load.py`

With the changes in this PR, all these steps can be omitted.

On top of that, we also have the re-imports to `peft/__init__.py` and `peft/tuners/__init__.py` but those are still required (I'm hesitant to mess with the import system). Furthermore, it's still required to add an entry to `PeftType` in `utils.peft_types.py`. Since this is an `enum`, it can't be easily generated automatically. Therefore, adding a new PEFT method is still not 100% self-contained.

## Changes in this PR

With this PR, less book-keeping is required. Instead of the 5 steps described above, contributors now only need to call

```python
# example for the Bone method
register_peft_method(name="bone", config_cls=BoneConfig, model_cls=BoneModel)
```

in the `__init__.py` of their PEFT method. In addition to registering the method, this also performs a couple of sanity checks (e.g. no duplicate names, method name and method prefix being identical).

Moreover, since so much book keeping is removed, this PR reduces the number of lines of code overall (at the moment +317, - 343).

## Implementation

The real difficulty of this task is that the module structure in PEFT is really messy, easily resulting in circular imports. This has been an issue in the past but has been especially painful here. For this reason, some stuff had to be moved around:

- `MODEL_TYPE_TO_PEFT_MODEL_MAPPING` is now in `auto.py` instead of `mapping.py`
- `PEFT_TYPE_TO_PREFIX_MAPPING` has been moved to `mapping.py` from `constants.py`
- `get_peft_model` had to be moved out of `mapping.py` and is now in its own module, `func.py` (better name suggestions welcome). This should be safe, as the function is re-imported to the main PEFT namespace, which all examples use.

The `PEFT_TYPE_TO_MODEL_MAPPING` dict could be completely removed, as it was basically redundant with `PEFT_TYPE_TO_TUNER_MAPPING`. The `get_peft_model_state_dict` could be simplified, as a lot of code was almost duplicated.

Overall, I think this is a cleaner module structure, but still not very clean overall.

## Open questions

I'm not 100% sure if this should be merged. AFAICT it should be a safe refactor that does not affect user code. There could be other packages out there that use some PEFT internals that could break with this refactor. If we decide to merge this, we should consider alerting potentially affected packages to test it.

I'm also open to the argument that the benefits are not outweighing the cost of the refactor.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -43,26 +43,11 @@ from peft.utils.constants import DUMMY_MODEL_CONFIG, PEFT_TYPE_TO_PREFIX_MAPPING
 from . import __version__
 from .config import PeftConfig
 from .tuners import (
-    AdaLoraModel,
-    AdaptionPromptModel,
-    BOFTModel,
-    BoneModel,
     CPTEmbedding,
-    FourierFTModel,
-    HRAModel,
-    IA3Model,
-    LNTuningModel,
-    LoHaModel,
-    LoKrModel,
-    LoraModel,
     MultitaskPromptEmbedding,
-    OFTModel,
-    PolyModel,
     PrefixEncoder,
     PromptEmbedding,
     PromptEncoder,
-    VBLoRAModel,
-    VeraModel,
     XLoraConfig,
     XLoraModel,
 )
@@ -85,30 +70,6 @@ from .utils import (
     set_peft_model_state_dict,
     shift_tokens_right,
 )
-
-
-PEFT_TYPE_TO_MODEL_MAPPING = {
-    PeftType.LORA: LoraModel,
-    PeftType.LOHA: LoHaModel,
-    PeftType.LOKR: LoKrModel,
-    PeftType.PROMPT_TUNING: PromptEmbedding,
-    PeftType.P_TUNING: PromptEncoder,
-    PeftType.PREFIX_TUNING: PrefixEncoder,
-    PeftType.ADALORA: AdaLoraModel,
-    PeftType.BOFT: BOFTModel,
-    PeftType.ADAPTION_PROMPT: AdaptionPromptModel,
-    PeftType.IA3: IA3Model,
-    PeftType.OFT: OFTModel,
-    PeftType.POLY: PolyModel,
-    PeftType.LN_TUNING: LNTuningModel,
-    PeftType.VERA: VeraModel,
-    PeftType.FOURIERFT: FourierFTModel,
-    PeftType.XLORA: XLoraModel,
-    PeftType.HRA: HRAModel,
-    PeftType.VBLORA: VBLoRAModel,
-    PeftType.CPT: CPTEmbedding,
-    PeftType.BONE: BoneModel,
-}
 
 
 class PeftModel(PushToHubMixin, torch.nn.Module):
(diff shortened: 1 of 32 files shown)


src/peft/peft_model.py
```diff
@@ -1784,9 +1737,7 @@ def forward(
             inputs_embeds = torch.cat((prompts, inputs_embeds), dim=1)
             return self.base_model(inputs_embeds=inputs_embeds, **kwargs)
 
-    def _cpt_forward(
-        self, input_ids=None, inputs_embeds=None, peft_config=None, task_ids=None, batch_size=None, **kwargs
-    ):
+    def _cpt_forward(self, input_ids, inputs_embeds, peft_config, task_ids, batch_size, **kwargs):
```
githubnemo, line 1740: Why this change?

githubnemo (commented):
As for the name of `func.py`: Maybe `mapping_func.py` is a more appropriate name?

src/peft/peft_model.py
BenjaminBossan: I don't remember my reasoning back then, but checking this now, it wouldn't make sense to use the default arguments, e.g.

`device = [i.device for i in [input_ids, inputs_embeds, labels] if i is not None][0]`

would fail if the arguments were `None`.

BenjaminBossan:
> As for the name of `func.py`: Maybe `mapping_func.py` is a more appropriate name?

Good idea, I renamed the module.

githubnemo (commented):
In general I do like static mapping more than dynamic, makes the code more readable without interpreting too much. However, in this case I can see that it is easy to forget certain mappings and keeping them at one place is not always easy. It also makes for a good public interface.

If you could add the doc string for `register_peft_method` we can merge this, I think.

src/peft/utils/peft_types.py
```diff
@@ -90,7 +90,35 @@ class TaskType(str, enum.Enum):
...
+        config_cls:
+            The configuration class of the PEFT method.
+        model_cls:
+            The model class of the PEFT method.
+        prefix (Optional[str], optional):
+            The prefix of the PEFT method. It should be unique. If not provided, the name of the PEFT method is used as
+            the prefix.
+        is_mixed_compatible (bool, optional):
+            Whether the PEFT method is compatible with `PeftMixedModel`. If you're not sure, leave it as False
+            (default).
```
githubnemo, line 111: Is there a better place to document when a model is compatible with `PeftMixelModel`? If not I would use the option to document what is necessary to be compatible.

BenjaminBossan: Models are never automatically compatible with `PeftMixelModel`, support needs to be added explicitly. I didn't think this was a good place to explain it, which is why I wrote to just ignore this option if unsure.

githubnemo (approved):

Issue #2283, opened 2024-12-15
TypeError when inference with different LoRA adapters in the same batch


yuxiang-guo:
### System Info

transformers 4.41.0
peft 0.13.2

### Who can help?

@BenjaminBossan
I tried to adopt [Inference with different LoRA adapters in the same batch] to an encoder-decoder T5 model.
Specifically, I load the base model, the first LoRA, and the second LoRA adapters,  and perform inference with these three models in the same batch. However, some errors occurred. 

BTW, does [inference with different LoRA adapters in the same batch] support beam search when using generate()?

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [X] My own task or dataset (give details below)

### Reproduction

Code:

```python
base_model = MT5ForConditionalGeneration.from_pretrained(base_model_path, cache_dir='cache')
peft_model = PeftModel.from_pretrained(base_model,<lora_path1> ,adapter_name="l1")
peft_model.load_adapter(<lora_path2>, adapter_name="l2")
adapter_names = ["__base__", "l1", "l2"]
output = model.generate(
                input_ids=inputs['input_ids'],
                adapter_names=adapter_names,
                max_length=20,
                prefix_allowed_tokens_fn=self.restrict_decode_vocab,
                early_stopping=True
            )
```

The error message:
```
Traceback (most recent call last):
  File "/home/user/user1/GR/trainer.py", line 1025, in prediction_step
    doc_ids = model.generate(
  File "/home/user/anaconda3/envs/test/lib/python3.8/site-packages/peft/peft_model.py", line 1972, in generate
    with self._enable_peft_forward_hooks(**kwargs):
  File "/home/user/anaconda3/envs/test/lib/python3.8/contextlib.py", line 113, in __enter__
    python-BaseException
return next(self.gen)
  File "/home/user/anaconda3/envs/test/lib/python3.8/site-packages/peft/peft_model.py", line 798, in _enable_peft_forward_hooks
    with self.base_model._enable_peft_forward_hooks(*args, **kwargs):
  File "/home/user/anaconda3/envs/test/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/home/user/anaconda3/envs/test/lib/python3.8/site-packages/peft/tuners/lora/model.py", line 441, in _enable_peft_forward_hooks
    handle = module.register_forward_pre_hook(pre_forward, with_kwargs=True)
[...]

PR #2287, opened 2024-12-17
FIX: Generating with mixed adapter batches and with beam search enabled

BenjaminBossan:
See #2283

Right now, using [mixed adapter batches](https://huggingface.co/docs/peft/developer_guides/lora#inference-with-different-lora-adapters-in-the-same-batch) (introduced in #1558) with beam search generations does not work. This is because users need to pass the adapter names associated with each sample, i.e. the number of adapter names should be identical to the number of samples in the input.

When applying beam search, transformers internally repeats the samples once per beam (or so it looks like). Therefore, we have more samples during generation than samples in the input. Consequently, the adapter names have to be extended accordingly. This is now taken care of.

For encoder-decoder models, we need to be careful. I seems like only the decoder needs to be extended, whereas the encoder receives the original number of inputs. Therefore, when an encoder-decoder model is identified, the extension is only applied to the decoder part.


githubnemo:
*not stale*

src/peft/tuners/lora/model.py
```diff
@@ -444,13 +444,34 @@ def _enable_peft_forward_hooks(self, *args, **kwargs):
...
         hook_handles = []
         for module in self.modules():
             if isinstance(module, LoraLayer) or isinstance(module, ModulesToSaveWrapper):
                 pre_forward = partial(_adapter_names_pre_forward_hook, adapter_names=adapter_names)
                 handle = module.register_forward_pre_hook(pre_forward, with_kwargs=True)
                 hook_handles.append(handle)
 
+        if uses_beam_search and hasattr(self.model, "get_encoder"):
+            # For encooder-decoder models, even when applying beam search, the encoder part of the model should not use
+            # the extended adapter_names.
```
githubnemo, line 468: Please extend the comment to answer the "why not".

Also, isn't the section above executed anyway and therefore we're just applying additional hooks to the decoder?

tests/testing_common.py
```diff
@@ -941,6 +941,86 @@ def _test_mixed_adapter_batches(self, model_id, config_cls, config_kwargs):
...
+            for sample0, sample1 in zip(gen0, gen1):
+                sample0 = remove_padding(sample0, pad_value)
+                sample1 = remove_padding(sample1, pad_value)
+                if (len(sample0) != len(sample1)) or (sample0 != sample1):
+                    # at least one sample differs, the generations are not identical
+                    return False
+            return True
+
+        # sanity check that there are enough outputs and that they are different
+        assert len(gen_base) == len(gen_adapter0) == len(gen_adapter1) >= 3
```
githubnemo, line 1008:
```suggestion
        assert len(gen_base) == len(gen_adapter0) == len(gen_adapter1) 
        assert len(gen_adapter1) >= 3
```

githubnemo (approved):
LGTM :)

PR #2288, opened 2024-12-17
Add zizmor for CI (security) linting

githubnemo:
To add a bit of a safety net to our CI jobs it might make sense to add a CI security linting tool such as [zizmor](https://woodruffw.github.io/zizmor/).

The linting run should be green at the moment since I fixed all reported issues:

- setting `persist-credentials: false` in all checkout runs
- changing template substitutions to environment variable substitutions

I added an ignore rule for `dangerous-triggers` to ignore the `upload_pr_to_documentation` workflow as our actions are configured to only run such steps on approval which should already have seen at least maintainer eyes and the zizmor run.


.github/workflows/zizmor.yaml
```diff
@@ -0,0 +1,24 @@
+name: CI security linting
+
+on:
+  push:
+    branches: ["main"]
```
BenjaminBossan, line 5: How about only executing this workflow if anything was changed inside of `.github` using the [`path argument`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-including-paths)? Or could we be missing something that way?

.github/zizmor.yml
```diff
@@ -0,0 +1,5 @@
+rules:
+  dangerous-triggers:
+    ignore:
+      # this workflow is only triggered after maintainer approval
+      - upload_pr_documentation.yml:3:1
```
BenjaminBossan, line 5: It's not beautiful but I don't know if there is anything else that we can do here.

BenjaminBossan (commented):
Thanks for improving the security of the PEFT workflows.

> changing template substitutions to environment variable substitutions

Could you explain why this is more secure?

> `persist-credentials: false`

Just to clarify, this is not a security issue on the repo as we're currently not using ssh, this is just a precaution, as that right?

githubnemo:
> > changing template substitutions to environment variable substitutions
> 
> Could you explain why this is more secure?

Sure. With `env: FOO=${{ foo }}` we're separating code and data. Environment variables are just a place to store strings and no interpretation of the contents is happening. You'd have to force an interpretation using `eval $FOO` or something similar.

Contrast this to the behavior before: `run: echo "something ${{ foo }}"` - here we're modifying the code directly with whatever the content of `${{ foo }}` is - there is no conversion to a shell string, there is no separation of code and data and therefore there is no way for the shell to know what is right or wrong.

> > `persist-credentials: false`
> 
> Just to clarify, this is not a security issue on the repo as we're currently not using ssh, this is just a precaution, as that right?

Yes, this is just to make the linter happy and avoid accidents should we ever (for whatever reason) use manual triggers using SSH.

.github/workflows/zizmor.yaml
githubnemo: Good idea, added accordingly.

.github/zizmor.yml
githubnemo: The [zizmor documentatin](https://woodruffw.github.io/zizmor/audits/#dangerous-triggers) recommends switching to the `workflow_call` trigger which has the security benefit of not passing environment variables from one workflow to the other - they're more isolated.

FWIW it seems to be the better scheme for this workflow anyway but I'm not sure if there are unforeseen consequences.

githubnemo:
@McPatate might be interested in this too, perhaps?

BenjaminBossan (approved):
LGTM, thanks.

PR #2294, opened 2024-12-23
Add MoSLoRA (EMNLP 2024)

wutaiqiang:
Previous PR:
https://github.com/huggingface/peft/pull/2013
https://github.com/huggingface/peft/pull/1905

Paper: https://aclanthology.org/2024.emnlp-main.450/
Blog: https://zhuanlan.zhihu.com/p/704821936
TL, DR:  We decompose LoRA into subspaces via structural re-parameterization and propose a simple yet effective MoSLoRA method, employing a learnable mixer to fuse more subspaces and more flexibly.

Sorry for the late submission. Based on the discussion in the previous PRs, I have implemented MoSLoRA in **separate files**. Also, I added test cases and expanded the documents with guidance on applying MoSLoRA. 

Please help review the code. Thanks for your kind work and Merry Christmas.

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -4210,7 +4210,7 @@ class TestHotSwapping:
             output1 = model(inputs).logits
 
         # sanity check:
-        tol = 1e-5
+        tol = 1e-4
         assert not torch.allclose(output0, output1, atol=tol, rtol=tol)
 
         with tempfile.TemporaryDirectory() as tmp_dirname:
@@ -4351,7 +4351,9 @@ class TestHotSwapping:
 
     @pytest.mark.skipif(not is_diffusers_available(), reason="Test requires diffusers to be installed")
     @pytest.mark.xfail(
-        strict=True, reason="Requires hotswap to be implemented in diffusers", raises=torch._dynamo.exc.RecompileError
+        strict=True,
+        reason="Requires hotswap to be implemented in diffusers",
+        raises=ValueError,
     )
     # it is important to check hotswapping small to large ranks and large to small ranks
     @pytest.mark.parametrize("ranks", [(11, 11), (7, 13), (13, 7)])


BenjaminBossan:
Thank you for resuming the work on MoSLoRA. I just want to let you know that we're currently on vacation and will review this next year.

docs/source/developer_guides/lora.md
```diff
@@ -54,6 +54,17 @@ lora_config = LoraConfig(init_lora_weights="pissa_niter_[number of iters]", ...)
 ```
 For detailed instruction on using PiSSA, please follow [these instructions](https://github.com/huggingface/peft/tree/main/examples/pissa_finetuning).
 
+### MoSLoRA
+[MoSLoRA](https://arxiv.org/abs/2406.11909) initializes the lora branch with an extra mixer layer. The vanilla LoRA can be viewed as the special case that the mixer is a _fixed_ identity metrix mixing $r$ subspaces. MoSLoRA try to employ a _learnable_ matrix to mix the information from $r^2$ subspaces.
```
BenjaminBossan, line 58: > MoSLoRA try to employ a _learnable_ matrix to mix the information from $r^2$ subspaces.

Let's reword this, there is no need for having "try to", right?

src/peft/utils/constants.py
```diff
@@ -120,6 +120,42 @@ def starcoder_model_postprocess_past_key_value(past_key_values):
     "qwen2": ["q_proj", "v_proj"],
 }
 
+TRANSFORMERS_MODELS_TO_MOSLORA_TARGET_MODULES_MAPPING = {
```
BenjaminBossan, line 123: Can we just make this a copy of `TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING`? I think each time that LoRA works, MosLoRA should also work.

src/peft/tuners/moslora/model.py
```diff
@@ -0,0 +1,617 @@
...
+        >>> model = prepare_model_for_kbit_training(model)
+        >>> lora_model = get_peft_model(model, config)
+        ```
+
+    **Attributes**:
+        - **model** ([`~transformers.PreTrainedModel`]) -- The model to be adapted.
+        - **peft_config** ([`MoSLoraConfig`]): The configuration of the Lora model.
+    """
+
+    prefix: str = "lora_"
```
BenjaminBossan, line 146: The prefix should be changed to `moslora` to avoid confusion.

src/peft/tuners/moslora/layer.py
```diff
@@ -0,0 +1,489 @@
...
+from peft.utils.other import transpose
+
+
+from peft.tuners.lora import LoraLayer, Embedding, Conv2d, Conv3d
+from .config import MoSLoraConfig
+
+
+class MoSLoraLayer(LoraLayer):
+    # All names of layers that may contain (trainable) adapter weights
+    adapter_layer_names = ("lora_A", "lora_B", "lora_mixer", "lora_embedding_A", "lora_embedding_B")
```
BenjaminBossan, line 39: We have all these parameter names that are still `lora_*`. I think it will be better to rename them all to `moslora_*` to avoid confusion. WDYT?

src/peft/tuners/moslora/layer.py
```diff
@@ -0,0 +1,489 @@
...
+    # All names of other parameters that may contain adapter-related parameters
+    other_param_names = ("r", "lora_alpha", "scaling", "lora_dropout")
+
+    def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, **kwargs) -> None:
+        super().__init__(base_layer)
+        self.use_moslora: dict[str, bool] = {}
```
BenjaminBossan, line 45: Do we still need this argument? If I use `MosLoraConfig`, it's clear that I want to use it.

src/peft/tuners/moslora/config.py
```diff
@@ -0,0 +1,77 @@
...
+    Args:
+        use_moslora: (`bool` | `Literal["kai", "orth"]`):
+
+    """
+    use_moslora: (bool | Literal["kai", "orth"]) = field(
+        default=True,
+        metadata={
+            "help": (
+                "Whether to enable 'Mixture-of-Subspaces in Low-Rank Adaptation' (MoSLoRA)."
+                "This technique employs a learnable mixer to fuse more subspaces in vanilla LoRA and more flexibly."
```
BenjaminBossan, line 37:
```suggestion
                "This technique employs a learnable mixer to fuse more subspaces in vanilla LoRA and is more flexibly."
```

src/peft/tuners/moslora/config.py
```diff
@@ -0,0 +1,77 @@
...
+@dataclass
+class MoSLoraConfig(LoraConfig):
+    """
+    This is the configuration class to store the configuration of a [`~peft.MoSLoRA`].
+
+    Args:
+        use_moslora: (`bool` | `Literal["kai", "orth"]`):
+
+    """
+    use_moslora: (bool | Literal["kai", "orth"]) = field(
```
BenjaminBossan, line 32: IMO it would be better to rename the parameter to `moslora_init`, as this is more about how to initialize the MosLoRA component.

src/peft/tuners/moslora/config.py
```diff
@@ -0,0 +1,77 @@
...
+        metadata={
+            "help": (
+                "Whether to enable 'Mixture-of-Subspaces in Low-Rank Adaptation' (MoSLoRA)."
+                "This technique employs a learnable mixer to fuse more subspaces in vanilla LoRA and more flexibly."
+                "In code implement, MoSLoRA inserts a mixer of r*r size between lora_A and lora_B."
+                "For more information, see https://arxiv.org/pdf/2406.11909."
+                "Passing `'False'` to disable mixer and thus it would be same as vanilla LoRA"
+                "Passing `'kai'` results in Kaiming Uniform initialization for Mixer."
+                "Passing `'orth'` results in Orthogonal initialization for Mixer."
+                "Passing `'True'` would enable Kaiming Uniform initialization which is default"
```
BenjaminBossan, line 43: The `"kai"` and `True` options are identical, right?  I would suggest to remove the redundancy. Also, I don't think we need the `False` option -- if users want LoRA, they can already do that. Moreover, I prefer the names to be fully spelled out for clarity (remember to update the docs & examples too).

In sum, I would change the argument to: `moslora_init: Literal["kaiming", "orthogonal"]`

src/peft/tuners/moslora/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@dataclass
+class MoSLoraConfig(LoraConfig):
+    """
+    This is the configuration class to store the configuration of a [`~peft.MoSLoRA`].
+
+    Args:
+        use_moslora: (`bool` | `Literal["kai", "orth"]`):
+
```
BenjaminBossan, line 30: As is, if users inspect the docstring of MosLoRA, they would only see this one argument. You should copy the full docstring of the LoRA config and make the MosLoRA-specific changes.

examples/moslora_finetuning/README.md
```diff
@@ -0,0 +1,104 @@
...
+```python
+import torch
+from peft import MoSLoraConfig, get_peft_model
+from transformers import AutoTokenizer, AutoModelForCausalLM, Trainer
+from datasets import load_dataset
+
+model = AutoModelForCausalLM.from_pretrained("TinyLlama/TinyLlama-1.1B-Chat-v1.0", device_map="cuda")
+tokenizer = AutoTokenizer.from_pretrained("TinyLlama/TinyLlama-1.1B-Chat-v1.0")
+dataset = load_dataset("timdettmers/openassistant-guanaco", split="train")
+lora_config = MoSLoraConfig(
```
BenjaminBossan, line 19:
```suggestion
moslora_config = MoSLoraConfig(
```

Code below also needs adjusting.

BenjaminBossan (changes requested):
Thank you for this very thorough PR. I reviewed parts of the code for now, but did not do a full deep dive. For this, it would be really great if you could do the following:

1. Quickly summarize the changes you made compared to LoRA. E.g. you included all the modules for the quantized layers, which is great. But it would take a lot of time to review them all when there is probably only very little change. Could you summarize those changes.
2. Similarly, I assume that most of `model.py` and `layer.py` is copied 1:1. Could you summarize the parts that were changed so that I can review them specifically?
3. On top of that, it would be also great if you could mark those changes in the code with comments. This will also help with keeping MosLoRA up to date when there are future changes in LoRA.

Regarding the parts that I did review, I had a couple of comments but most should be quick to fix. I think the biggest change I consider would be about the parameter names: Right now, you mostly still use `lora` all throughout, e.g. we still have `lora_A`, `lora_B`, etc. I'm a bit conflicted about this: On the one hand, those parameters have the same function as in LoRA, on the other hand it could be confusing when checking the model what type of adapter we have. Since you probably put some thoughts into this, could you share your reasoning?

Apart from those comments, could you please:
- Run `make style`
- Update the year in the copyright notices to 2025

BenjaminBossan:
I only skimmed it, but the method described in [this paper](https://arxiv.org/abs/2411.19557) looks very similar to MosLoRA, only that LoRA A and B are fixed and that a specific init method is chosen.

src/peft/tuners/moslora/layer.py
wutaiqiang: Thanks for your advice. Since MoSLoRA=LoRA+mixer, the only difference is the learnable mixer matrix between lora_A and lora_B. So I follow Adalora to use the same names as LoRA.  In Adalora:

https://github.com/huggingface/peft/blob/bf186edc5babcdec7f33982b631f78eba844752c/src/peft/tuners/adalora/layer.py#L37

They also inherit the same names as LoRA.

src/peft/tuners/moslora/layer.py
wutaiqiang: Thanks for your advice. Since there are two initialization methods for the mixer in MoSLoRA, so I introduced the use_moslora to be "kai" or "orth" for the Kaiming initialization and orthogonal initialization. Moreover, by setting use_moslora as a bool parameter, the user can run the vanilla LoRA quickly.

src/peft/tuners/moslora/config.py
wutaiqiang: Thanks for your kind advice. I think it would be convenient for the user to run vanilla LoRA, so I named this parameter "use_moslora". Anyway, I can rename it and remove the boolean choice if you think it is necessary :)

wutaiqiang:
> I only skimmed it, but the methods described in [this paper](https://arxiv.org/pdf/2405.17604) and [this paper](https://arxiv.org/abs/2411.19557) look very similar to MosLoRA, only that LoRA A and B are fixed and that a specific init method is chosen.

Thanks for pointing out these two papers. They look similar but they also differ in the initialization of A/B and Mixer.

The Arxiv time for these three papers is:

| Method | Time |
|----|-----|
| MoSLoRA | 16 Jun 2024 |
| LoRA-XS | 27 May 2024 |
| LoRA-SB | 29 Nov 2024 |

The structure is similar but the detailed design differs. In LoRA-XS, they try to decompose frozen pre-train weights via SVD and initialize AB using these decomposed vectors. In MoSLoRA, we do not need the decomposition process and introduce the mixer to fuse subspaces.

src/peft/tuners/moslora/config.py
wutaiqiang: I see. I would follow your suggestion.

wutaiqiang:
> Heads up: We recently merged PR #2282, which facilitates the addition of new PEFT methods. This is what causes the merge conflicts. It should be straightforward to adapt the code to the new way, as it reduces book keeping. Check for instance the diff for LoRA: https://github.com/huggingface/peft/pull/2282/files#diff-1c7921c8a3305a64be1fe90184fd27e1d32547caba0037e5ead8cd8927e93ebc. You can do the same changes, while removing the need to update all those mappings (see the [description](https://github.com/huggingface/peft/pull/2282#issue-2738270979) of said PR). If you have any doubts, just ask me.

It is quite a big change. I think I would need more time to modify the code. Please wait for a while. 

It is okay if you think it is necessary to close this PR and open a new one next time. Thanks for your time and effort :)

BenjaminBossan:
> Thanks for pointing out these two papers. They look similar but they also differ in the initialization of A/B and Mixer.

Yes, the initialization differs and what parameters are trainable, but the rest looks the same for me. I wonder, if MoSLoRA is added, we can add options to use LoRA-XS and LoRA-SB too, e.g. `config = MosloraConfig(init_weights="lora-sb")`. This way, the 3 methods could be implemented without too much extra code.

> It is quite a big change. I think I would need more time to modify the code. Please wait for a while.

It really shouldn't be too big. Please ask if anything is unclear.

> It is okay if you think it is necessary to close this PR and open a new one next time.

Whatever works best for you.

PR #2297, opened 2024-12-26
Fix typo in lora config.py

innerlee:
Mmodule -> Module


BenjaminBossan (approved):
Thanks for fixing the typos in the type annotations.

Failing CI is unrelated, so this is good to be merged.

PR #2300, opened 2024-12-30
FIX: Make CorDA example work

5eqn:
Previously when changing `load_dataset` logic in #2231 , I've forgotten `traindataset = []` for dataset `codefeedback` and `wizardlm`, so the example doesn't work. In this PR I've fixed this bug, and tested that the example works for all datasets.

### DIFF
--- a/examples/corda_finetuning/datautils.py
+++ b/examples/corda_finetuning/datautils.py
@@ -167,6 +167,7 @@ def get_calib_data(name, tokenizer, model_id, nsamples, seqlen=2048, seed=3):
         selected_data_dict = (
             load_dataset("iboing/CodeFeedback-Filtered-Instruction", split="train").shuffle(seed=seed).take(nsamples)
         )
+        traindataset = []
         for example in selected_data_dict:
             if example.get("input", "") == "":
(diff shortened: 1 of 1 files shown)


BenjaminBossan (commented):
Thanks for catching this. I wonder if it would be easier if we move this line:

https://github.com/huggingface/peft/blob/6d458b300fc2ed82e19f796b53af4c97d03ea604/examples/corda_finetuning/datautils.py#L199

to the beginning of the function body. That way, we can ensure that `traindataset` is always initialized and don't need to call it in each if-else branch. WDYT?

5eqn:
> Thanks for catching this. I wonder if it would be easier if we move this line:
> 
> https://github.com/huggingface/peft/blob/6d458b300fc2ed82e19f796b53af4c97d03ea604/examples/corda_finetuning/datautils.py#L199
> 
> to the beginning of the function body. That way, we can ensure that `traindataset` is always initialized and don't need to call it in each if-else branch. WDYT?

Thanks for your suggestion! After this move, if someone were to add a dataset, it will indeed be less error-prone. I've updated the code and it's ready for review.

PR #2306, opened 2025-01-06
FIX #2295: Warn when user reloads modified model

### LINKED ISSUE #2295: PEFT model doesn't update params when having changed LoRA config
### System Info

I have noticed that when updated the `target_modules` settings in the LoRA config, the PEFT model params remain unchanged. Might affect other PEFT settings too.

My assumption is that `get_peft_model()` does not re-instantiate/update its settings once it has been initialized before.

System: Windows 11
Python: 3.11
peft: 0.14.0

### Who can help?

@BenjaminBossan @sayakpaul

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

For reproduction in a Jupyter Notebook:

```py
from peft import LoraConfig, get_peft_model, TaskType
from transformers import AutoTokenizer, AutoModelForTokenClassification
import torch

label_list = ['B-LOC', 'B-MISC', 'B-ORG', 'B-PER', 'I-LOC', 'I-MISC', 'I-ORG', 'I-PER', 'O']

# Initialize tokenizer
tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-3.2-1B")
tokenizer.padding_side = "right"

model = AutoModelForTokenClassification.from_pretrained(
    "meta-llama/Llama-3.2-1B",
    pad_token_id=tokenizer.eos_token_id,
    torch_dtype=torch.bfloat16,
    device_map="auto", 
    num_labels=len(label_list)
)

    print(name)
```

```py
lora_config = LoraConfig(
    task_type=TaskType.TOKEN_CLS,
    r=16,             
    lora_alpha=32, 
    lora_dropout=0.1
)
```

```py
```

This outputs 

```
trainable params: 1,722,377 || all params: 1,237,555,218 || trainable%: 0.1392
```

But when changing the above code without restarting the kernel to:

```py
lora_config = LoraConfig(
    task_type=TaskType.TOKEN_CLS,
    r=16,             
    lora_alpha=32, 
    target_modules=["layers.0.self_attn.q_proj", "layers.0.self_attn.v_proj"], # changed to specific heads
    lora_dropout=0.1
)
```

and retrieving the trainable params again:

```py
``` 

it outputs again

```
trainable params: 1,722,377 || all params: 1,237,555,218 || trainable%: 0.1392
``` 

but after the update it should be 

```
trainable params: 124,937 || all params: 1,235,957,778 || trainable%: 0.0101
```

### Expected behavior

When having updated `lora_config`, `get_peft_model()` should retrieve the current config.

githubnemo:
This is a fix for https://github.com/huggingface/peft/issues/2295.

When modifying a model with `get_peft_model` that was already modified in the same way, even specifying a different config may not change the trainable parameter count, e.g. when specifying target modules that are only a subset of the previous target modules.

With this patch a warning will be issued with a hint to `.unload()` when calling `get_peft_model` on an already modified model.

### DIFF
--- a/src/peft/mapping.py
+++ b/src/peft/mapping.py
@@ -31,7 +31,6 @@ from .peft_model import (
     PeftModelForSeq2SeqLM,
     PeftModelForSequenceClassification,
     PeftModelForTokenClassification,
-    get_layer_status,
 )
 from .tuners import (
     AdaLoraConfig,
@@ -71,7 +70,7 @@ from .tuners import (
     VeraModel,
     XLoraConfig,
 )
-from .tuners.tuners_utils import BaseTuner
+from .tuners.tuners_utils import BaseTuner, BaseTunerLayer
 from .utils import _prepare_prompt_learning_config
 
 
@@ -182,20 +181,14 @@ def get_peft_model(
     new_name = model.__dict__.get("name_or_path", None)
     peft_config.base_model_name_or_path = new_name
 
-    # Especially in notebook environments there could be a case that a user
-    # wants to experiment with different configuration values. However, it
-    # is likely that there won't be any changes for new configs on an already
+    # Especially in notebook environments there could be a case that a user wants to experiment with different
+    # configuration values. However, it is likely that there won't be any changes for new configs on an already
     # initialized PEFT model. The best we can do is warn the user about it.
-    try:
-        if len(get_layer_status(model)) > 0:
-            warnings.warn(
-                "You are trying to modify a model with PEFT for a "
-                "second time. If you want to reload the model with a "
-                "different config, make sure to call `.unload()` before."
-            )
-    except ValueError:
-        # not a PEFT model or no adapters in use
-        pass
+    if any(isinstance(module, BaseTunerLayer) for module in model.modules()):
+        warnings.warn(
+            "You are trying to modify a model with PEFT for a second time. If you want to reload the model with a "
+            "different config, make sure to call `.unload()` before."
+        )
 
     if (old_name is not None) and (old_name != new_name):
         warnings.warn(
--- a/src/peft/mapping.py
+++ b/src/peft/mapping.py
@@ -31,6 +31,7 @@ from .peft_model import (
     PeftModelForSeq2SeqLM,
     PeftModelForSequenceClassification,
     PeftModelForTokenClassification,
+    get_layer_status,
 )
 from .tuners import (
     AdaLoraConfig,
@@ -180,6 +181,21 @@ def get_peft_model(
     old_name = peft_config.base_model_name_or_path
     new_name = model.__dict__.get("name_or_path", None)
     peft_config.base_model_name_or_path = new_name
+
+    # Especially in notebook environments there could be a case that a user
+    # wants to experiment with different configuration values. However, it
+    # is likely that there won't be any changes for new configs on an already
+    # initialized PEFT model. The best we can do is warn the user about it.
+    try:
+        if len(get_layer_status(model)) > 0:
+            warnings.warn(
+                "You are trying to modify a model with PEFT for a "
+                "second time. If you want to reload the model with a "
+                "different config, make sure to call `.unload()` before."
+            )
+    except ValueError:
+        # not a PEFT model or no adapters in use
+        pass
 
     if (old_name is not None) and (old_name != new_name):
         warnings.warn(
--- a/tests/test_mapping.py
+++ b/tests/test_mapping.py
@@ -1,0 +1,39 @@
+import pytest
+import torch
+
+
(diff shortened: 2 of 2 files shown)


src/peft/mapping.py
```diff
@@ -181,6 +182,21 @@ def get_peft_model(
     new_name = model.__dict__.get("name_or_path", None)
     peft_config.base_model_name_or_path = new_name
 
+    # Especially in notebook environments there could be a case that a user
```
BenjaminBossan, line 185: We have 120 char limits in the project, could you please configure your formatter accordingly?

src/peft/mapping.py
```diff
@@ -181,6 +182,21 @@ def get_peft_model(
...
+    # wants to experiment with different configuration values. However, it
+    # is likely that there won't be any changes for new configs on an already
+    # initialized PEFT model. The best we can do is warn the user about it.
+    try:
+        if len(get_layer_status(model)) > 0:
```
BenjaminBossan, line 190: Nice idea to use `get_layer_status`. I wonder if in this case, a simple

`if any(isinstance(module, BaseTunerLayer) for module in model.modules())`

would not serve the purpose better. This check would stop once the first PEFT layer is found, while `get_layer_status` would do a bunch of more work unnecessarily.

tests/test_mapping.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+import torch
+
+
+class TestGetPeftModel:
+    RELOAD_WARNING_EXPECTED_MATCH = r"You are trying to modify a model .*"
+
+    @pytest.fixture
```
BenjaminBossan, line 8: In this project, we just import the PEFT functions and classes directly on the root of the test files, so it's not really necessary to use a fixture in this case.

tests/test_mapping.py
```diff
@@ -0,0 +1,39 @@
+import pytest
```
BenjaminBossan, line 1: Let's add the copyright notice.

tests/test_mapping.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @pytest.fixture
+    def base_model(self):
+        return torch.nn.Sequential(torch.nn.Linear(10, 2))
+
+    def test_get_peft_model_warns_when_reloading_model(self, get_peft_model, lora_config, base_model):
+        get_peft_model(base_model, lora_config)
+
+        with pytest.warns(UserWarning, match=self.RELOAD_WARNING_EXPECTED_MATCH):
+            get_peft_model(base_model, lora_config)
```
BenjaminBossan, line 28: How about also adding a check where the user calls `get_peft_model` on the `PeftModel` instance itself?

tests/test_mapping.py
```diff
@@ -0,0 +1,39 @@
...
+    def test_get_peft_model_proposed_fix_in_warning_helps(self, get_peft_model, lora_config, base_model, recwarn):
+        peft_model = get_peft_model(base_model, lora_config)
+        peft_model.unload()
+        get_peft_model(base_model, lora_config)
+
+        warning_checker = pytest.warns(UserWarning, match=self.RELOAD_WARNING_EXPECTED_MATCH)
+
+        for warning in recwarn:
+            if warning_checker.matches(warning):
+                pytest.fail("Warning raised even though model was unloaded.")
```
BenjaminBossan, line 39: Maybe a bit of an edge case, but currently we would not detect if a user tries to create, say, a LoRA model based on a base model that was already modified with a prompt learning method, like prefix tuning. This is because those methods don't add any `BaseTunerLayer`s (which is what `get_layer_status` relies on). Implementing such a check is probably not easy and out of scope for this PR. We could still add an xfail-ing test though.

BenjaminBossan (changes requested):
Thanks for adding this warning. I have a couple of small comments, but nothing major. Please check.

src/peft/mapping.py
githubnemo: For some reason I thought that p-tuning/prompt-tuning are also layer tuners (which they aren't) so I thought it was worthwhile to use the more complex `get_layer_status`. But you're correct, a simple check suffices.

tests/test_mapping.py
BenjaminBossan: Honestly, I'm not quite sure if it would just work or not. Coincidentally, there is a recent issue asking a similar question (#2307).

githubnemo: For our sanity I'd suggest to go forward with the current state of things and once we know more about the interplay between soft-prompting vs. soft-prompting and soft-prompting vs. lora we adapt tests and/or implementation. WDYT?

githubnemo:
@BenjaminBossan please review again

Issue #2310, opened 2025-01-07
Comparison of Different Fine-Tuning Techniques for Conversational AI


ImamaDev:
### Feature request

It would be incredibly helpful to have a clear comparison or support for various fine-tuning techniques specifically for conversational AI. This feature could include insights into their strengths, limitations, and ideal use cases, helping practitioners choose the right approach for their needs.

Here’s a list of techniques to consider:

LoRa
AdaLoRa
BONE
VeRa
XLora
LN Tuning
VbLora
HRA (Hyperparameter Regularization Adapter)
IA3 (Input-Aware Adapter)
Llama Adapter
CPT (Conditional Prompt Tuning)etc

### Motivation

With the growing number of fine-tuning techniques for conversational AI, it can be challenging to identify the most suitable approach for specific use cases. A comprehensive comparison of these techniques—highlighting their strengths, limitations, and ideal scenarios—would save time, reduce trial-and-error, and empower users to make informed decisions. This feature would bridge the gap between research and practical application, enabling more effective model customization and deployment.

### Your contribution

I’d be happy to collaborate on this! While I might not have a complete solution right now, I’m willing to contribute by gathering resources, reviewing papers, or helping organize comparisons. If others are interested in teaming up, we could work together on a PR to make this feature happen. Let’s connect and brainstorm how we can tackle this effectively!

BenjaminBossan:
Thanks for coming up with this proposal. Indeed, this is something we have on our backlog for a long time. As you can imagine, providing objective and useful information on this is a huge undertaking, since relying on the paper results can often be problematic.

As a long term project, we plan to provide some kind of benchmark that compares all these methods in terms of runtime, memory usage, performance, etc. but I can't give any concrete date yet.

In the meantime, we have started to be more rigorous when new methods are being added in requiring a clear description of what the best use cases are. There is still a lot of room for improvement, especially when it comes to methods that were added some time ago.

If you (and others) want to contribute, I think a good place to start would be to go through the individual methods in the [PEFT docs](https://huggingface.co/docs/peft/index) and help improve the descriptions. If we can make them more uniform, with more details on the best uses cases, pros and cons, this would already be a nice improvement.

![image](https://github.com/user-attachments/assets/f2e1fe78-e0ec-4b4f-831d-aabbc4118efa)

There are other places that could benefit from such a clean up, e.g. the description of all the [LoRA initialization methods](https://huggingface.co/docs/peft/developer_guides/lora#initialization).

[...]

PR #2312, opened 2025-01-08
The paper on the Bone structure has been updated

JL-er:
The new version of the paper is more reliable and easier for readers to understand.
https://arxiv.org/abs/2409.15371


BenjaminBossan:
Thanks for the update.

I haven't compared the new paper, but checking the PR: Is DiSHA a method that could be used on its own, but is not available in PEFT, or is it the union of Bone and Bat? What I'd like to avoid is for users to be confused when they want to use DiSHA but don't find any method with that name in PEFT. Maybe this could be clarified in the description.

JL-er:
Simply put, DiSHA is the overall framework, while Bone and Bat are just subsets of it. So I even want to rename Bone to DiSHA, and then choose either Bone or Bat during init_weights.

BenjaminBossan:
> So I even want to rename Bone to DiSHA, and then choose either Bone or Bat during init_weights.

When it comes to the PEFT code, let's avoid any renaming, as this would break backwards compatibility.

JL-er:
> > So I even want to rename Bone to DiSHA, and then choose either Bone or Bat during init_weights.
> 
> When it comes to the PEFT code, let's avoid any renaming, as this would break backwards compatibility.

So how should I modify the description of Bone? Because DiSHA explains in detail how Bone came about, the theoretical support provided in the older Bone paper is insufficient.

BenjaminBossan:
I think rewriting the descriptions of the method as you did is okay, as the existing code is not affected by it. I just think you should highlight that in the PEFT code base, the method is still referred to as `Bone` and users should use `BoneConfig` etc. Also, I think it's worth updating the docstring, for instance for `BoneConfig`, to mention the relationship to DiSHA. Imagine a user reads the new docs and then searches for DiSHA in the code base, they would currently not get any hits.

In theory, we can also allow renaming the method in code. However, for that, we would need a long deprecation period, so that users who use `BoneConfig` know that they should switch to `DishaConfig`. If the prefixes of the adapter weights are changed, we also need a way to convert existing checkpoints. I think it is not worth the hassle and we should keep the existing names.

JL-er:
> I think rewriting the descriptions of the method as you did is okay, as the existing code is not affected by it. I just think you should highlight that in the PEFT code base, the method is still referred to as `Bone` and users should use `BoneConfig` etc. Also, I think it's worth updating the docstring, for instance for `BoneConfig`, to mention the relationship to DiSHA. Imagine a user reads the new docs and then searches for DiSHA in the code base, they would currently not get any hits.
> 
> In theory, we can also allow renaming the method in code. However, for that, we would need a long deprecation period, so that users who use `BoneConfig` know that they should switch to `DishaConfig`. If the prefixes of the adapter weights are changed, we also need a way to convert existing checkpoints. I think it is not worth the hassle and we should keep the existing names.

I plan not to modify the structure names, but to add explanations in the examples so that users can understand how to use them.

BenjaminBossan:
@JL-er do you still plan on finishing this PR?

JL-er:
> @JL-er do you still plan on finishing this PR?

I'll finish it tomorrow

JL-er:
> @JL-er do you still plan on finishing this PR?

Completed

BenjaminBossan (approved):
Thanks for the updates.

The CI is acting up at the moment, so I just canceled, since these changes should not affect the tests. Note that on hf/papers, there is still the old version of the paper, not sure if/how it can be updated.

BenjaminBossan:
Hey @JL-er I just wanted to let you know that we added a [PEFT method comparison suite](https://github.com/huggingface/peft/tree/main/method_comparison) and the [results are now in](https://github.com/huggingface/peft/issues/2310#issuecomment-2988652008). Besides LoRA, Bone (Bat) is on the Pareto frontier when it comes to test score vs memory:

image

I thought you might be happy to hear that :)

Joluck:
> Hey @JL-er I just wanted to let you know that we added a [PEFT method comparison suite](https://github.com/huggingface/peft/tree/main/method_comparison) and the [results are now in](https://github.com/huggingface/peft/issues/2310#issuecomment-2988652008). Besides LoRA, Bone (Bat) is on the Pareto frontier when it comes to test score vs memory:
> 
> image
> 
> I thought you might be happy to hear that :)

Haha, this is exactly the PEFT arena I’ve always wanted. I believe Bone will be incredibly powerful, and I’m really grateful to you all. At the same time, I also want to share some good news with you: the paper on this method has been accepted by ESFOMO (ICML workshop) and is also being submitted to NeurIPS. However, there is one not-so-great piece of news—my method has been renamed again, from Bone to DiSHA, and now to MiSS. But I can assure you that the name will not change anymore, and I’m quite satisfied with this version of the paper :)

Joluck:
@BenjaminBossan I plan to officially rename Bone in PEFT to MiSS, and this will be the final name with no further changes. I’d like to ask: what are your thoughts on this change, and do you foresee any potential issues it might cause?（To better promote this method

BenjaminBossan:
@JL-er Okay, as long as backwards compatibility is preserved, like in this PR, a final name change should be fine.

BenjaminBossan:
@JL-er Okay, as long as backwards compatibility is preserved, like in this PR, a final name change should be fine.

PR #2313, opened 2025-01-08
[DOC] Updated documentation for `get_peft_model()` for in-place base model modification

### LINKED ISSUE #2295: PEFT model doesn't update params when having changed LoRA config

d-kleine:
closes #2295 along with FIX #2306 

Updates documentation for `get_peft_model()` that the base model will be modified in-place:
- tutorial/peft_model_config
- `get_peft_model()` docstring

### DIFF
--- a/docs/source/tutorial/peft_model_config.md
+++ b/docs/source/tutorial/peft_model_config.md
@@ -138,7 +138,7 @@ lora_model.print_trainable_parameters()
 > [!WARNING]
 > When calling [`get_peft_model`], the base model will be modified *in-place*. That means, when modifying a model with [`get_peft_model`] that was already modified in the same way before, even specifying a different configuration may not change the trainable parameter count (e.g., when specifying target modules that are only a subset of the previous target modules).
 ><br>
-> Therefore, if you would like to modify your PEFT configuration after having called [`get_peft_model()`] before, you would first have unload the model with [`~LoraModel.unload`] and then call [`get_peft_model()`] with your new configuration.
+> Therefore, if you would like to modify your PEFT configuration after having called [`get_peft_model()`] before, you would first have to unload the model with [`~LoraModel.unload`] and then call [`get_peft_model()`] with your new configuration.
 
 Now you can train the [`PeftModel`] with your preferred training framework! After training, you can save your model locally with [`~PeftModel.save_pretrained`] or upload it to the Hub with the [`~transformers.PreTrainedModel.push_to_hub`] method.
 
--- a/docs/source/tutorial/peft_model_config.md
+++ b/docs/source/tutorial/peft_model_config.md
@@ -135,6 +135,11 @@ lora_model.print_trainable_parameters()
 "trainable params: 1,572,864 || all params: 332,769,280 || trainable%: 0.472659014678278"
 ```
 
+> [!WARNING]
+> When calling [`get_peft_model`], the base model will be modified *in-place*. That means, when modifying a model with [`get_peft_model`] that was already modified in the same way before, even specifying a different configuration may not change the trainable parameter count (e.g., when specifying target modules that are only a subset of the previous target modules).
+><br>
+> Therefore, if you would like to modify your PEFT configuration after having called [`get_peft_model()`] before, you would first have unload the model with [`~LoraModel.unload`] and then call [`get_peft_model()`] with your new configuration.
+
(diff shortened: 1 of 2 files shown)


docs/source/tutorial/peft_model_config.md
```diff
@@ -135,6 +135,9 @@ lora_model.print_trainable_parameters()
 "trainable params: 1,572,864 || all params: 332,769,280 || trainable%: 0.472659014678278"
 ```
 
+> [!WARNING]
+> When calling [`get_peft_model`], the base model will be modified *in-place*. That means, when modifying a model with [`get_peft_model`] that was already modified in the same way before, even specifying a different configuration may not change the trainable parameter count (e.g., when specifying target modules that are only a subset of the previous target modules). Therefore, if you would like to modify your PEFT configuration after having called [`get_peft_model()`] before, you would first have to unload the model with [`~LoraModel.unload`] and then call [`get_peft_model()`] with your new configuration.
```
BenjaminBossan, line 139: > may not change the trainable parameter count

I would not focus so much on this specific aspect. The reason why in the original example, the parameter count was not changed is because the new target modules were a subset of the old target modules. This is not generally the case and users should not think that just because the parameter count was changed, everything is fine. IMO the gist of the issue is that calling `get_peft_model` with the already modified base model will apply LoRA on top of the existing LoRA.

> you would first have to unload the model

Alternatively, the user can just re-initialize the base model, let's add this as info. (They can also create a deepcopy before the first `get_peft_model` call, but let's not overcomplicate things)

BenjaminBossan (changes requested):
Thanks for creating this PR to make it very clear in the docs that `get_peft_model` modifies the base model in place. I made a few suggestions which should hopefully better clarify what's going, please check.

docs/source/tutorial/peft_model_config.md
d-kleine: Yeah, I see what you mean, good points! 

I just edited the text to a more general statement that the model will be further modified, avoiding to go too much into details. Furthermore, I have also implemented your request on adding the re-initialization info, trying to keep the box as both informative but also concise as possible.

BenjaminBossan (approved):
Thanks for improving the docs for `get_peft_model`, this should help avoid confusion in the future.

Failing CI is unrelated, so merging is fine.

PR #2320, opened 2025-01-10
FIX: Package checks for torchao, EETQ

BenjaminBossan:
### Torchao

Under some unknown circumstances, it can happen that even though `importlib.util.find_spec("torchao") is not None`, `importlib_metadata.version("torchao")` still fails. This error is now caught, same as in [diffusers](https://github.com/huggingface/diffusers/blob/9f06a0d1a4a998ac6a463c5be728c892f95320a8/src/diffusers/utils/import_utils.py#L351-L357). The error was noticed in the [diffusers CI](https://github.com/huggingface/diffusers/actions/runs/12700836694/job/35404283533#step:7:313).

### EETQ

This is basically a revert of #2226. That PR had to add a check to the EETQ import as EETQ was broken for some time with latest transformers (see https://github.com/NetEase-FuXi/EETQ/issues/34 for context) but that has been fixed.

### DIFF
--- a/src/peft/import_utils.py
+++ b/src/peft/import_utils.py
@@ -81,19 +81,7 @@ def is_auto_awq_available():
 
 @lru_cache
 def is_eetq_available():
-    if importlib.util.find_spec("eetq") is None:
-        return False
-
-    is_available = True
-    try:
-        from eetq import EetqLinear  # noqa: F401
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
The one failing CI run is independent of this PR and fixed via #2314.

PR #2323, opened 2025-01-13
[tests] make cuda-only tests device-agnostic

faaany:
This PR enables current cuda-only tests to work on non-cuda accelerators. Below are some of the results on XPU after this fix:

```bash
PASSED tests/test_common_gpu.py::PeftGPUCommonTests::test_r_odd_hra_inference
PASSED tests/test_common_gpu.py::PeftGPUCommonTests::test_modules_to_save_grad
PASSED tests/test_gpu_examples.py::TestBOFT::test_boft_half_conv
PASSED tests/test_gpu_examples.py::TestBOFT::test_boft_half_linear
PASSED tests/test_gpu_examples.py::TestEvaInitializationGPU::test_eva_initialization_consistency[model]
```

### DIFF
--- a/src/peft/import_utils.py
+++ b/src/peft/import_utils.py
@@ -13,9 +13,11 @@
 # limitations under the License.
 import importlib
 import importlib.metadata as importlib_metadata
+import platform
 from functools import lru_cache
 
 import packaging.version
+import torch
 
 
 @lru_cache
@@ -111,3 +113,24 @@ def is_torchao_available():
             f"but only versions above {TORCHAO_MINIMUM_VERSION} are supported"
         )
     return True
+
+
+@lru_cache
+def is_xpu_available(check_device=False):
+    """
+    Checks if XPU acceleration is available either via `intel_extension_for_pytorch` or via stock PyTorch (>=2.4) and
+    potentially if a XPU is in the environment
+    """
+
+    system = platform.system()
+    if system == "Darwin":
+        return False
+    else:
+        if check_device:
+            try:
+                # Will raise a RuntimeError if no XPU is found
+                _ = torch.xpu.device_count()
+                return torch.xpu.is_available()
+            except RuntimeError:
+                return False
+        return hasattr(torch, "xpu") and torch.xpu.is_available()
--- a/src/peft/import_utils.py
+++ b/src/peft/import_utils.py
@@ -118,8 +118,7 @@ def is_torchao_available():
 @lru_cache
 def is_xpu_available(check_device=False):
     """
-    Checks if XPU acceleration is available either via `intel_extension_for_pytorch` or via stock PyTorch (>=2.4) and
-    potentially if a XPU is in the environment
+    Checks if XPU acceleration is available and potentially if a XPU is in the environment
     """
 
     system = platform.system()
--- a/tests/test_common_gpu.py
+++ b/tests/test_common_gpu.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import gc
 import tempfile
+import unittest
 
 import pytest
 import torch
@@ -79,7 +80,8 @@ if is_bnb_available():
         from peft.tuners.vera import Linear4bit as VeraLinear4bit
 
 
-class PeftGPUCommonTests:
+@require_non_cpu
+class PeftGPUCommonTests(unittest.TestCase):
     r"""
     A common tester to run common operations that are performed on GPU such as generation, loading in 8bit, etc.
     """
--- a/tests/test_common_gpu.py
+++ b/tests/test_common_gpu.py
@@ -53,7 +53,7 @@ from peft import (
     get_peft_model,
     prepare_model_for_kbit_training,
 )
-from peft.import_utils import is_bnb_4bit_available, is_bnb_available
+from peft.import_utils import is_bnb_4bit_available, is_bnb_available, is_xpu_available
 from peft.tuners.lora.config import LoraRuntimeConfig
 from peft.utils import infer_device
 
@@ -100,7 +100,7 @@ class PeftGPUCommonTests(unittest.TestCase):
         gc.collect()
         if torch.cuda.is_available():
             torch.cuda.empty_cache()
-        elif torch.xpu.is_available():
+        elif is_xpu_available():
             torch.xpu.empty_cache()
         gc.collect()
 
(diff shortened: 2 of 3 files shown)


BenjaminBossan:
Thanks for the PR. I ran the GPU tests locally with NVIDIA GPUs to confirm that they still work as expected. The failing CI appears to be unrelated, I will get back to this PR after investigating.

BenjaminBossan:
Update: The failing CI is caused by the changes of this PR, I was mistaken in my first assessment. Changes like the following are the cause:

```diff
- @pytest.mark.skipif(not torch.cuda.is_available(), reason="test requires a GPU")
+ @require_non_cpu
```

For some reason, the test is **not skipped** on GitHub CI, even though the GitHub runners are using only CPU. Checking `require_non_cpu`, it looks correct:

https://github.com/huggingface/peft/blob/b345a6e41521b977793cbdcaf932280081b18141/tests/testing_utils.py#L33-L41

Locally, without GPU accelerator, the decorator works correctly for me, so it is unclear why it is failing on GitHub.

faaany:
> Update: The failing CI is caused by the changes of this PR, I was mistaken in my first assessment. Changes like the following are the cause:
> 
> ```diff
> - @pytest.mark.skipif(not torch.cuda.is_available(), reason="test requires a GPU")
> + @require_non_cpu
> ```
> 
> For some reason, the test is **not skipped** on GitHub CI, even though the GitHub runners are using only CPU. Checking `require_non_cpu`, it looks correct:
> 
> https://github.com/huggingface/peft/blob/b345a6e41521b977793cbdcaf932280081b18141/tests/testing_utils.py#L33-L41
> 
> Locally, without GPU accelerator, the decorator works correctly for me, so it is unclear why it is failing on GitHub.

I saw from the error log that the "torch_device" seems to be "none" in GitHub CI, which is strange. For the BNB error, I added a `require_bitsandbytes` maker. could you trigger the CI again and see what the error says?

BenjaminBossan:
Thanks for the update. Some tests are still failing:

```
FAILED tests/test_common_gpu.py::TestSameAdapterDifferentDevices::test_ln_tuning_add_new_adapter_does_not_change_device
FAILED tests/test_gpu_examples.py::TestPrefixTuning::test_prefix_tuning_multiple_devices_decoder_model - RuntimeError: Cannot access accelerator device when none is available.
FAILED tests/test_gpu_examples.py::TestPrefixTuning::test_prefix_tuning_multiple_devices_encoder_decoder_model - RuntimeError: Cannot access accelerator device when none is available.
FAILED tests/test_gpu_examples.py::TestHotSwapping::test_hotswapping_compiled_model_does_not_trigger_recompilation - assert 1 == 0
ERROR tests/test_gpu_examples.py::TestEvaInitializationGPU::test_eva_initialization_consistency[model_bnb] - importlib.metadata.PackageNotFoundError: No package metadata was found for bitsandbytes
```

- For `test_ln_tuning_add_new_adapter_does_not_change_device` and `test_hotswapping_compiled_model_does_not_trigger_recompilation`, those probably fail because the `@require_non_cpu` does not work correctly on GH.
- For `test_prefix_tuning_multiple_devices_decoder_model` and `test_prefix_tuning_multiple_devices_encoder_decoder_model`, these define `device_map`. This should be solved by `@require_torch_multi_gpu`.
- For the Eva test, we should add a `require_bitsandbytes` to the `model_bnb` fixture.

faaany:
@BenjaminBossan, I know why the first test case fails. This is because some tests are not child classes of " unittest.TestCase". Code updated. "require_torch_multi_accelerators" and "require_bitsandbytes" are also added.

BenjaminBossan:
> I know why the first test case fails. This is because some tests are not child classes of " unittest.TestCase".

Ah I see, well spotted. However, we can't change the tests to inherit from `unittest.TestCase`. These are "pytest-style" test classes, not unit tests (we are slowly transitioning towards pytest-style). This allows us to use some pytest features like `pytest.fixture` which don't work with unittest. Therefore, making it a `TestCase` will break the test.

Unfortunately, this means we can't use `@require_non_cpu` decorator on the test class and it's why the `pytest.mark.skipif` decorator was being used.

There are thus 2 options:

1. Implement a `require_non_cpu` decorator using `pytest.mark.skipif`
2. Move the decorator to each method instead of having it on the class.

I tested the 2nd option and it looks like the decorator should still work on methods, but it doesn't work on classes:

```python
import unittest
import pytest

def skip_deco(test_case):
    return unittest.skipUnless(False, "test is skipped")(test_case)

class TestUnitTestStyle(unittest.TestCase):
    @pytest.fixture  # fixtures don't work with TestCase
    def one(self):
        return 1

    def test_it(self, one):
        assert one == 2

@skip_deco  # does not work
class TestPytestStyleClassDeco:
    @pytest.fixture
    def one(self):
        return 1

    def test_it(self, one):
        assert one == 2

class TestPytestTyleMethodDeco:
    @pytest.fixture
    def one(self):
        return 1

    @skip_deco  # works
    def test_it(self, one):
        assert one == 2
```

faaany:
That's correct. I think we can go with option 2. Code updated.

BenjaminBossan (changes requested):
Thanks for the updates, we're almost there. We still get these errors:

```
FAILED tests/test_gpu_examples.py::TestLowCpuMemUsageDifferentDevices::test_low_cpu_mem_usage_with_quantization[bnb-4bit] - importlib.metadata.PackageNotFoundError: bitsandbytes
FAILED tests/test_gpu_examples.py::TestLowCpuMemUsageDifferentDevices::test_low_cpu_mem_usage_with_quantization[bnb-8bit] - ImportError: Using `bitsandbytes` 8-bit quantization requires the latest version of bitsandbytes: `pip install -U bitsandbytes`
```

This is because of this line:

https://github.com/huggingface/peft/pull/2323/files#diff-037b6550b3d063119b60801a751c3c5c97eab8e912ae8370c2dd5569d38305adL3868

which results in these tests being run in normal CI. However, in normal CI, bitsandbytes is not installed, causing the error. Let's add the `require_bitsandbytes` decorator (probably needs to be on method level, since it's not a `TestCase` class).

Moreover, all the instances of `torch.xpu.is_available()` are failing on MacOS with:

`AttributeError: module 'torch' has no attribute 'xpu'`

I assume that this function is not implemented for Mac. What we probably need to do is write a custom `is_xpu_available` function that checks the platform and returns `False` on Darwin and `torch.xpu.is_available()` otherwise.

I saw that there is such a function [in accelerate](https://github.com/huggingface/accelerate/blob/fbfa53bc5eaabf13359d7a8edc91bc08ea145080/src/accelerate/utils/imports.py#L380) but there is no platform check, so I don't think we can just use that.

faaany:
I see, thanks for the feedback! updated.

BenjaminBossan (approved):
Thanks for making more tests compatible with XPU. Changes LGTM.

PR #2324, opened 2025-01-13
FIX: reduce CorDA memory consumption + docs

5eqn:
- Added documentation about selection of `run_model` sample size
- Added documentation about memory consumption of preprocessing
- Pruned temporary fields after preprocessing and building LoRA adapter, so that CorDA does not consume more memory during training compared to other LoRA methods

### DIFF
--- a/examples/corda_finetuning/README.md
+++ b/examples/corda_finetuning/README.md
@@ -100,7 +100,12 @@ lora_config = LoraConfig(
     init_lora_weights="corda",
     corda_config=corda_config,
 )
+
+# Call `preprocess_corda` first to collect covariance matrix and build SVD result for model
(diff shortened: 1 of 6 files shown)


BenjaminBossan (changes requested):
Thanks for the updates to the docs and the improvements to memory consumption. The changes LGTM.

Before merging, could you please fix the following line:

https://github.com/huggingface/peft/blob/3289134524d39f5b11718e221c5ec40daef8fe1a/examples/corda_finetuning/preprocess.py#L24

The import is no longer valid after merging #2282. Please always import like so: `from peft import get_peft_model`.

BenjaminBossan (approved):
Thanks for the updates, LGTM.

PR #2325, opened 2025-01-13
FIX low_cpu_mem_usage=True with 8bit bitsandbytes

BenjaminBossan:
There was a bug in PEFT that occurred when trying to use the `low_cpu_mem_usage=True` option with 8bit bitsandbytes quantized models. This bug is fixed now.

See https://github.com/huggingface/diffusers/issues/10550 for the bug report.

### DIFF
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -248,15 +248,6 @@ class LoraModel(BaseTuner):
             child = child.base_layer
 
         meta = torch.device("meta")
-
-        # handle .state attribute of 8 bit bnb
-        if (getattr(child, "state", None) is not None) and not any(p.device == meta for p in new_module.parameters()):
-            if hasattr(new_module, "base_layer"):
-                new_module.base_layer.state = child.state
-            else:
-                new_module.state = child.state
-            new_module.to(child.weight.device)
-
         # dispatch to correct device
         for name, module in new_module.named_modules():
             if (self.prefix in name) or ("ranknum" in name):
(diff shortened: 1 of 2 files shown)


src/peft/tuners/lora/model.py
```diff
@@ -247,14 +247,16 @@ def _replace_module(self, parent, child_name, new_module, child):
         if hasattr(child, "base_layer"):
             child = child.base_layer
 
-        if getattr(child, "state", None) is not None:
+        meta = torch.device("meta")
+
+        # handle .state attribute of 8 bit bnb
```
githubnemo, line 252: Can you elaborate a bit more what "handle" means in this case? What is the reason behind ignoring meta device parameters?

BenjaminBossan: Thanks a lot for nudging me to take a closer look at this piece of code. Indeed, what I found is that it can most likely be removed completely.

I tried to understand where it came from and I think this is actually a remnant from the first days of PEFT where there was no `base_layer` and instead the PEFT layer would "integrate" the base layer. After #1106, this is no longer the case. I confirmed this be checking that actually `new_module.base_layer.state is child.state` -- since they're the same object, all that casting is no longer needed. The `else` branch was not reached in my tests.

To double check, I ran all GPU tests locally (filtered by `-k 8bit`) and they passed. The newly added tests also still pass. Thus I think it's safe to remove the whole block.

githubnemo: Cool! Just to understand the change:

1. `new_module.to(child.weight.device)` is not needed because that's handled a few lines after, right?
2. I understand that `child == new_module.base_layer` and therefore `state` is equal but what is `new_module` when it does not have a `base_layer` attribute?

sayakpaul (approved):

PR #2331, opened 2025-01-16
React on new zizmor version findings

githubnemo:
Zizmor detected a potential cache poisoning attack via `setup-docker-buildx`. There is an argument to this (an attacker with a valid github token could modify the cache, change the buildx binary and tamper with Docker build releases) but there is also an argument against it: the buildx cache would prevent general information leaks when a new buildx release is tampered with. Since there is no obvious benefit from either side, we ignore this hint and deem it uncritical.

We also change the trigger of zizmor runs to pushes on main, regardless of whether workflow files are changed or not to catch new audits from more recent zizmor versions.

### DIFF
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -3,8 +3,6 @@ name: CI security linting
 on:
   push:
     branches: ["main"]
-    paths:
-      - '.github/**'
   pull_request:
     branches: ["*"]
     paths:
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -3,3 +3,13 @@ rules:
     ignore:
       # this workflow is only triggered after maintainer approval
       - upload_pr_documentation.yml:3:1
+  cache-poisoning:
+    ignore:
+      # the docker buildx binary is cached and zizmor warns about a cache poisoning attack.
+      # OTOH this cache would make us more resilient against an intrusion on docker-buildx' side.
+      # There is no obvious benefit so we leave it as it is.
+      - build_docker_images.yml:35:9
+      - build_docker_images.yml:68:9
+      - build_docker_images.yml:101:9
+      - build_docker_images.yml:134:9
+      - build_docker_images.yml:167:9


BenjaminBossan:
I want to understand the implications a bit better here before continuing. The zizmor errors are a bit obscure to me. Are they suggesting that we shouldn't be using any cache for docker at all? Or what would be the alternative to fix the error instead of ignoring it?

githubnemo:
> Are they suggesting that we shouldn't be using any cache for docker at all? Or what would be the alternative to fix the error instead of ignoring it?

They are arguing that you shouldn't use any caching when building release artifacts to avoid tempered releases due to cache poisoning. You can read about their take on it [here](https://woodruffw.github.io/zizmor/audits/#cache-poisoning).

The fix would look like this:

```diff
diff --git a/.github/workflows/build_docker_images.yml b/.github/workflows/build_docker_images.yml
index 5b93f80..3a01fea 100644
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -21,6 +21,8 @@ jobs:
     steps:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
+        with:
+          cache-binary: false
       - name: Check out code
         uses: actions/checkout@v3
         with:
@@ -54,6 +56,8 @@ jobs:
     steps:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
+        with:
+          cache-binary: false
       - name: Check out code
         uses: actions/checkout@v3
         with:
@@ -87,6 +91,8 @@ jobs:
     steps:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
+        with:
+          cache-binary: false
       - name: Check out code
         uses: actions/checkout@v3
         with:
@@ -120,6 +126,8 @@ jobs:
     steps:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
+        with:
+          cache-binary: false
       - name: Check out code
         uses: actions/checkout@v3
         with:
@@ -153,6 +161,8 @@ jobs:
     steps:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
+        with:
+          cache-binary: false
       - name: Check out code
         uses: actions/checkout@v3
         with:
```

BenjaminBossan (approved):
Okay, so the consequence would be that the docker build would be severely slowed down. The images are not intended for public use, so the damage would be contained, although of course attackers could pull off some shenanigans if they had access to our GPU workers.

Of course, if an attacker had access to the GitHub token of PEFT, they could do quite some damage (no releases though), so one could argue that cache poisoning would not be our biggest worry in that case. It is also not still unclear to me how the cache poisoning would be performed concretely, would the steps be auditable?

Anyway, I trust your judgment here as I'm not so knowledgeable in this area.

githubnemo:
> Okay, so the consequence would be that the docker build would be severely slowed down. 

Not severely, the cache is only for the `buildx` binary docker plugin. The `docker/build-push-action` does not perform caching by default.

PR #2332, opened 2025-01-17
TST: Update torch.compile tests and docs

BenjaminBossan:
We have tests to check if `torch.compile` works for various PEFT methods and "advanced" features (QLoRA, merging, ...). These tests are not run on a regular basis, but are triggered manually. As such, it was time to revisit them.

So far, a few of these tests were marked as `xfail`ing. All these tests are passing now. The reasons for this:

- Presumably: New PyTorch version (I haven't checked older); to wit, running the `torch.compile` tests without changes leads to [failures](https://github.com/huggingface/peft/actions/runs/12829812187/job/35776634643) because of strict `xfail`
- Loosening some tolerances
- Remove a spurious argument added by `torch.compile`
- Slightly adjust order of when `torch.compile` is called

The docs have been updated to reflect these new findings.

Note that the `torch.compile` tests don't run on regular CI. I ran the tests locally and they passed for me.

### DIFF
--- a/docs/source/developer_guides/torch_compile.md
+++ b/docs/source/developer_guides/torch_compile.md
@@ -59,7 +59,7 @@ Below are some of the more advanced PEFT features that **work**. They were all t
 - Disabling adapters (i.e. using `with model.disable_adapter()`)
 - Unloading (i.e. calling `model.merge_and_unload()`)
 - Mixed adapter batches (i.e. calling `model(batch, adapter_names=["__base__", "default", "other", ...])`)
-- Inference with multiple adapters
+- Inference with multiple adapters (i.e. using `model.add_adapter` or `model.load_adapter` to load more than 1 adapter); for this, only call `torch.compile` _after_ loading all adapters
 
 Generally, we can expect that if a feature works correctly with LoRA and is also supported by other adapter types, it should also work for that adapter type.
 


tests/test_torch_compile.py
```diff
@@ -349,10 +351,10 @@ def test_causal_lm_multiple_lora_adapter_compile(self, tokenizer, data):
             output_base = model(sample)
 
         config = LoraConfig(task_type=TaskType.CAUSAL_LM, init_lora_weights=False)
-        model = get_peft_model(model, config).eval()
-        model = self.compile(model, {})
+        model = get_peft_model(model, config)
         model.add_adapter("other", config)
         model = self.compile(model, {})
+        model.eval()
```
githubnemo, line 357: Is it a requirement to only call compile once?

githubnemo (approved):
LGTM, one question out of interest

tests/test_torch_compile.py
BenjaminBossan: From my testing, `torch.compile` needs to be called after loading all adapters. I extended the docs to say so in my [last commit](https://github.com/huggingface/peft/pull/2332/commits/d9cb9f9aa70b63f397e67407c81fac58bc8c71d7).

PR #2333, opened 2025-01-17
Adding Lora implementation for nn.Conv1d

### LINKED ISSUE #2241: Request for adding the lora implementation for Conv1d rather than transormers.utils.Conv1d
### Feature request

Hi, I found that Lora does not support the model with torch.nn.Conv1d as convolution layers, which limits the use-case for models pre-trained with this class (for example, Enformer). I wonder if it is possible to add an implementation based on this class.

### Motivation

To finetune enformer.

### Your contribution

If you need I can open a PR.

CCLDArjun:
Resolves #2241

My comment shows that the shapes match in Enformer model: https://github.com/huggingface/peft/issues/2241#issuecomment-2596949160

Unsure how to test further it other than to run it in some training loop

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -67,6 +67,8 @@ class LoraLayer(BaseTunerLayer):
         base_layer = self.get_base_layer()
         if isinstance(base_layer, nn.Linear):
             in_features, out_features = base_layer.in_features, base_layer.out_features
+        elif isinstance(base_layer, nn.Conv1d):
+            in_features, out_features = base_layer.in_channels, base_layer.out_channels
         elif isinstance(base_layer, nn.Conv2d):
             in_features, out_features = base_layer.in_channels, base_layer.out_channels
         elif isinstance(base_layer, nn.Conv3d):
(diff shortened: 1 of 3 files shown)


CCLDArjun:
`make style` was editing 22 different files so I ran ruff on my changed files directly I think it should be good

BenjaminBossan:
> `make style` was editing 22 different files so I ran ruff on my changed files directly I think it should be good

I think the most likely explanation is that you were using a different ruff version from what is used on CI. This would explain why CI still fails. Could you please ensure that the same version is used: ruff-0.6.9?

CCLDArjun:
@BenjaminBossan Yep, 0.6.9 works much better `make style` is happy

BenjaminBossan (changes requested):
Thanks a lot for the updates, the PR LGTM. I tested it Hubert to have a more realistic test and it worked too (with the exception being the `groups` argument, but that is yet to be added by all conv layers).

Before merging, however, I just noticed one small change still needed, namely this error message which lists all supported layer types for LoRA:

https://github.com/huggingface/peft/blob/bbb112841bd4e7d6ef4d11739f22da81f17e65d6/src/peft/tuners/lora/model.py#L347-L351

BenjaminBossan (approved):
Thanks for adding LoRA support for Conv1d, LGTM.

PR #2334, opened 2025-01-20
FIX Bug with modules_to_save loading if substring

### LINKED ISSUE #2289: Inconsistent Parameter Mismatches After Merging PEFT and Base Models
### System Info

peft 0.14.0, transformers 4.45.2, accelerate 1.0.1, Python 3.11.9, windows

### Who can help?

@BenjaminBossan @sayakpaul

### Information

- [X] My own modified scripts

### Tasks

### Reproduction
```python
from transformers_custom.modeling impor6t CustomConfig
from transformers_custom.tokenization import CustomTokenizer
from transformers_custom.multitask_model import CustomForSequenceClassificationMultitask
from peft import PeftModel
import torch    

def compare_model_params(model1, model2):
    # Extract state dictionaries
    sd1 = model1.state_dict()
    sd2 = model2.state_dict()
    
    # First, check if they have the same keys
    keys1 = set(sd1.keys())
    keys2 = set(sd2.keys())
    
    # Find parameters that are not present in both
    missing_in_model2 = keys1 - keys2
    missing_in_model1 = keys2 - keys1
    
    if missing_in_model2:
        print("Parameters missing in model2:", missing_in_model2)
    if missing_in_model1:
        print("Parameters missing in model1:", missing_in_model1)
        
    # Now compare parameters that exist in both
    mismatch_names = []
    for key in sorted(keys1.intersection(keys2)):
        param1 = sd1[key]
        param2 = sd2[key]
        
        # Check for shape mismatch
        if param1.shape != param2.shape:
            mismatch_names.append(key)
            continue
        
        # Check for value mismatch
        if not torch.allclose(param1, param2):
            print("Mismatched values for parameter:", key, f"model1: {param1}", f"model2: {param2}")
            mismatch_names.append(key)
    
    # Print out results
    if mismatch_names:
        print("Mismatched parameters:", mismatch_names)
    else:
        print("All parameters match perfectly.")

base_model_path = r"C:\models\tms\download\base2"
peft_path = r"C:\models\tms\download\adapter2"
merged_model_path = r"C:\models\tms\download\adapter2_merged\peft_merged"

config = CustomConfig.from_pretrained(
    base_model_path,
    num_labels=8,
    finetuning_task=None,
    cache_dir=None,
    revision="main",
)

base_model = CustomForSequenceClassificationMultitask.from_pretrained(
    base_model_path,
    config=config,
    cache_dir=None,
    revision="main",
) 

peft_model = PeftModel.from_pretrained(base_model, peft_path)

peft_model_merged = peft_model.merge_and_unload()
peft_model_merged.eval()

merged_config = CustomConfig.from_pretrained(
    merged_model_path,
    num_labels=8,
    finetuning_task=None,
    cache_dir=None,
    revision="main",
)

merged_model = CustomForSequenceClassificationMultitask.from_pretrained(
    merged_model_path,
    config=merged_config,
    cache_dir=None,
    revision="main",
)
merged_model.eval()

compare_model_params(peft_model_merged, merged_model)
```
### Expected behavior
I saved the base model and the merged model (using save_pretrained) after training and calling merge_and_unload(). I also saved the PEFT model (via trainer.save_model). After loading the PEFT parameters on top of the base model and calling merge_and_unload(), I compared the newly merged model with the previously saved merged model. Some parameters do not match, and the specific mismatches change with each run to compare models. For example, sometimes the mismatched parameters are ['classifier2.class_dense.bias', 'classifier2.class_dense.weight', ...] and other times ['custom.encoder.layer.19.attention.self.query.weight'].

How can I resolve this issue? Ideally, there should be no mismatches, or at least the mismatches should be consistent across runs.

BenjaminBossan:
Fixes #2289

This bug was the result of an error in the logic of modifying the `state_dict` for `modules_to_save` in `set_peft_model_state_dict`. The error in the logic was that it was checked if an entry from `modules_to_save` (a set of strings) is a substring of a key of the `state_dict`. If it was, a new name was assigned to that key in the `state_dict`, which would allow to load the weight later.

The issue that stems from the substring check occurs if there are multiple `modules_to_save`, and one of them has a name that is a substring of another. So e.g. if one is named `"classifier"` and the other is named `"classifier2"`, there could be a false match.

This PR fixes the issue by enclosing the string with `"."`, i.e. we now check if `".classifier."` is a substring instead, which avoid false matches.

What made this bug even harder to debug was that `modules_to_save` is a set and therefore has no predetermined order. Therefore, the bug would be flaky. To address this, `modules_to_save` is now sorted before iterating over it. That doesn't contribute to resolving the bug, but it makes the bug non-flaky, allowing future debugging to be easier.

### DIFF
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -335,10 +335,11 @@ def set_peft_model_state_dict(
     state_dict = {}
     if getattr(model, "modules_to_save", None) is not None:
         for key, value in peft_model_state_dict.items():
-            if any(module_name in key for module_name in model.modules_to_save):
-                for module_name in model.modules_to_save:
-                    if module_name in key:
-                        key = key.replace(module_name, f"{module_name}.modules_to_save.{adapter_name}")
+            if any(f".{module_name}." in key for module_name in model.modules_to_save):
+                # sort to make order deterministic, but should not affect overall logic
+                for module_name in sorted(model.modules_to_save):
+                    if f".{module_name}." in key:
+                        key = key.replace(f".{module_name}.", f".{module_name}.modules_to_save.{adapter_name}.")
                         break
             state_dict[key] = value
     else:
--- a/tests/test_other.py
+++ b/tests/test_other.py
@@ -12,12 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 import pytest
 import torch
 from torch import nn
 from transformers import AutoModelForCausalLM, AutoModelForSequenceClassification
 
-from peft import LoraConfig, get_peft_model
+from peft import LoraConfig, PeftModel, get_peft_model
 from peft.utils.other import ModulesToSaveWrapper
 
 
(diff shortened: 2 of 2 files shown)


src/peft/utils/save_and_load.py
```diff
@@ -335,10 +335,11 @@ def set_peft_model_state_dict(
     state_dict = {}
     if getattr(model, "modules_to_save", None) is not None:
         for key, value in peft_model_state_dict.items():
-            if any(module_name in key for module_name in model.modules_to_save):
-                for module_name in model.modules_to_save:
-                    if module_name in key:
-                        key = key.replace(module_name, f"{module_name}.modules_to_save.{adapter_name}")
+            if any(f".{module_name}." in key for module_name in model.modules_to_save):
```
githubnemo, line 338: This does not cover the case where the module uses a `nn.Parameter` instead of `nn.Linear` as classification head. Do we want to support that as well?

githubnemo (commented):
Great debugging, thanks! As discussed offline we will accept the flakiness of the test as the chance is sufficiently low that we'll miss a regression with how many CI runs we do regularly.

src/peft/utils/save_and_load.py
BenjaminBossan: The targeted object for `modules_to_save` must be an `nn.Module`, it's not possible to target an `nn.Parameter`. Or am I misunderstanding?

githubnemo: Ah, perfect, nevermind then.

githubnemo (approved):

PR #2345, opened 2025-01-23
TST Use different diffusion model for testing

BenjaminBossan:
So far, tests are using `hf-internal-testing/tiny-stable-diffusion-torch` for testing diffusion models. However, this model has some issues:

- still uses pickle (.bin) instead of safetensors
- there is a `FutureWarning` because of the config

Now, using `hf-internal-testing/tiny-sd-pipe` (thanks Sayak for suggesting) instead which doesn't have those issues.

On top of that, there is often a Hub error when loading the diffusion model, although it's unclear why this particular model is so prone to that. Maybe the new model eschews this error, but even if it doesn't, it is better to make the switch.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -179,7 +179,7 @@ class TestScalingAdapters:
         assert torch.allclose(logits_base_model, logits_lora_model)
 
     def test_diffusers_pipeline(self):
-        model_id = "hf-internal-testing/tiny-stable-diffusion-torch"
+        model_id = "hf-internal-testing/tiny-sd-pipe"
         pipeline = StableDiffusionPipeline.from_pretrained(model_id)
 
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
> On top of that, there is often a Hub error when loading the diffusion model, although it's unclear why this particular model is so prone to that. Maybe the new model eschews this error, but even if it doesn't, it is better to make the switch.

Update: 3 of 12 CI runs [failed](https://github.com/huggingface/peft/actions/runs/12933697246/job/36073052080?pr=2345#step:8:1331) because they could not load the model. So nothing improved in this regard.

sayakpaul:
I see the entire CI to be green. So, is the failure random?

sayakpaul (approved):

PR #2346, opened 2025-01-24
Improve invalid peft config error message

thedebugger:
Print required keys to help w/ troubleshooting

### DIFF
--- a/src/peft/config.py
+++ b/src/peft/config.py
@@ -157,7 +157,10 @@ class PeftConfigMixin(PushToHubMixin):
 
             filtered_kwargs, unexpected_kwargs = _check_and_remove_unused_kwargs(cls, kwargs)
             if not MIN_EXPECTED_CONFIG_KEYS.issubset(set(filtered_kwargs.keys())):
-                raise TypeError(f"The config that is trying to be loaded is not a valid {cls.__name__} config.")
+                raise TypeError(
+                    f"The {cls.__name__} config that is trying to be loaded is missing required keys: "
+                    f"{MIN_EXPECTED_CONFIG_KEYS}."
+                )
 
             warnings.warn(
                 f"Unexpected keyword arguments {sorted(unexpected_kwargs)} for class {cls.__name__}, these are "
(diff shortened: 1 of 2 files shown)


src/peft/config.py
```diff
@@ -157,7 +157,7 @@ def from_peft_type(cls, **kwargs):
 
             filtered_kwargs, unexpected_kwargs = _check_and_remove_unused_kwargs(cls, kwargs)
             if not MIN_EXPECTED_CONFIG_KEYS.issubset(set(filtered_kwargs.keys())):
-                raise TypeError(f"The config that is trying to be loaded is not a valid {cls.__name__} config.")
+                raise TypeError(f"The config that is trying to be loaded is invalid {cls.__name__} config and missing required keys: {MIN_EXPECTED_CONFIG_KEYS}.")
```
BenjaminBossan, line 160: I think the wording would not be correct after the change. Before:

> The config that is trying to be loaded is not a valid LoraConfig config

After

> The config that is trying to be loaded is invalid LoraConfig config and ...

So let's revert the wording or make it "is **an** invalid".

Moreover, let's add a line break (120 chars max).

tests/test_config.py
```diff
@@ -365,7 +365,7 @@ def test_from_pretrained_sanity_check(self, config_class, tmp_path):
         with open(tmp_path / "adapter_config.json", "w") as f:
             json.dump(non_peft_json, f)
 
-        msg = f"The config that is trying to be loaded is not a valid {config_class.__name__} config"
+        msg = f"The config that is trying to be loaded is invalid {config_class.__name__} config and missing required keys: {{'peft_type'}}."
```
BenjaminBossan, line 368: Same comment as above.

BenjaminBossan (changes requested):
Thanks for improving the error message. Let's just make a small change to the message, otherwise this looks good.

src/peft/config.py
thedebugger: ok, I made the error message more concise. sounds okay?

BenjaminBossan:
Thanks for the rewording. Could you please run `make style` (requires ruff v0.9.3)?

BenjaminBossan (approved):
Thanks for improving the error message when trying to load invalid configs.

PR #2350, opened 2025-01-28
Use locked install for zizmor

githubnemo:
To be on the safe side when installing zizmor and it's dependencies we're using a locked installation, meaning that the dependencies and their versions are taken from the Cargo.lock file.

This will hopefully reduce the chances of having the pipeline randomly fail due to updated dependencies down the line.


BenjaminBossan (approved):
I think it makes sense in our context. Can be merged once CI is green.

PR #2358, opened 2025-02-03
[docs] rename link to PEFT Quicktour

makelinux:
The "Get started" link currently points to the "Quicktour" article,
while "Get started" is also the first title in the TOC, causing
confusion.

Rename the "Get started" link to "Quicktour" to match the article and
ensure consistency.


BenjaminBossan (approved):
Thanks for clarifying.

PR #2363, opened 2025-02-06
CI: Remove EETQ from Dockerfile

BenjaminBossan:
At the moment, EETQ is broken after the PyTorch 2.6 release. There is a PR to fix this but it's not clear when it's merged. For the time being, comment out the EETQ install, otherwise our Docker image is not being updated, thus nightly PEFT CI runs with packages that are out of date.


BenjaminBossan:
The EETQ patch has been merged, so this PR should no longer be necessary.

PR #2371, opened 2025-02-08
Optimization for ConvNd if dropout=0.

gslama12:
As discussed in [#2153](https://github.com/huggingface/peft/pull/2153)

### DIFF
--- a/src/peft/tuners/lora/dora.py
+++ b/src/peft/tuners/lora/dora.py
@@ -139,7 +139,7 @@ class _DoraConvNdLayer(DoraLinearLayer):
         weight_norm = weight.norm(p=2, dim=dim, keepdim=True).transpose(1, 0)
         return weight_norm
 
-    def forward(self, x, *, lora_A, lora_B, scaling, base_layer):
+    def forward(self, x, *, lora_A, lora_B, scaling, base_layer, base_result=None):
         """
         For DoRA, calculate the extra output from LoRA with DoRA applied. This should be added on top of the base layer
         output.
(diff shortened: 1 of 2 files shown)


src/peft/tuners/lora/dora.py
```diff
@@ -157,8 +157,9 @@ def forward(self, x, *, lora_A, lora_B, scaling, base_layer):
         # during backpropagation"
         weight_norm = weight_norm.detach()
         mag_norm_scale = magnitude / weight_norm
-        result_dora = (mag_norm_scale - 1) * (
-            self.conv_fn(
+
+        if base_result is None:
```
BenjaminBossan, line 161: Wondering: Do we not need to account for the possibility that the base layer has a bias term, same as we do for linear?

https://github.com/huggingface/peft/blob/40fe16644697564c8d0b3996830d682630469e70/src/peft/tuners/lora/dora.py#L91-L94

BenjaminBossan (commented):
Thanks for adding this DoRA optimization to conv layers. Generally, this looks good, but I have one comment about the bias term.

src/peft/tuners/lora/dora.py
gslama12: Yes i believe you are right. Since the previous case always used bias=None i think we should subtract it to keep the implementation consistent. I updated the code.

BenjaminBossan:
Thanks for the update. Let's call `make style` or remove the trailing white space to make the linter happy.

gslama12:
Should be fixed.

src/peft/tuners/lora/dora.py
```diff
@@ -167,8 +168,12 @@ def forward(self, x, *, lora_A, lora_B, scaling, base_layer):
                 dilation=base_layer.dilation,
                 groups=base_layer.groups,
             )
-        ) + mag_norm_scale * lora_B(lora_A(x)) * scaling
+        else:
+            bias = base_layer.bias
+            if bias is not None:
+                base_result = base_result - bias
```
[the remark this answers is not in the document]
gslama12: Looks good to me, thanks for the fix!

BenjaminBossan (approved):
Thanks for bringing this DoRA optimization to conv layers.

PR #2376, opened 2025-02-13
Standalone Custom Tokens Tuner and integrated into LoRA

githubnemo:
This PR is based on the nifty addition of @marcusinthesky from https://github.com/huggingface/peft/pull/1541.

I took the liberty of bringing the branch up-to-date and do concept where we not only have `CustomTokens` as a PEFT method for selectively re-training tokens but also have a `trainable_token_indices` parameter in LoRA to combine both approaches (and possibly with other methods in the future).

### What is this

When adding tokens or fine-tuning the representation of specific tokens we currently have little choice but to retrain the whole embedding matrix which can be huge and adds to the memory footprint (in RAM but also on disk). This method creates a sparse matrix of shape `(n, embed_dim)` where n is the number of tokens to be customized and only trains these few values.

### How to use this

Two possibilities:

1. use the `CustomTokens` PEFT method
```
peft_config = CustomTokensConfig(target_modules=['embed_tokens'], token_indices=[0, 1, 2])
peft_model = get_peft_model(model, peft_config)
```
2. use in conjunction with LoRA
```
peft_config = LoraConfig(
    target_modules='all-linear',
    trainable_token_indices={'embed_tokens': [0, 1, 2]},
)
peft_model = get_peft_model(model, peft_config)
``` 

### Implementation details

This is an early draft since I found no better way of implementing it without touching the `modules_to_save` infrastructure.

The idea is to abstract the `ModulesToSaveWrapper` into an `AuxiliaryTrainingWrapper` that allows for more functionality than simply setting `requires_grad_(True)` on specific modules and saving them alongside other modules. There are now three classes, 
- `AuxiliaryTrainingWrapper` the base class that provides a common interface for wrapping modules and forwarding getattr/forward calls from said modules
- `ModulesToSaveWrapper` is the same as before but extended by having a method to get the state dict from the wrapped models for the given adapter so that we know which modules to save without having to match the state dict names
- `NewTokensWrapper` is a thin wrapper around `CustomTokensLayer` that can be applied to layers specified by the `trainable_token_indices` parameter from `LoraConfig` (and others in the future)

To load and save these modules we iterate over the model's `named_modules` to filter all `AuxiliaryTrainingWrapper` instances, get their state dicts and - depending on load or save - read adapter-specific names and write them out to be adapter-less or vice versa. In theory this should handle saving `modules_to_save` as well as `trainable_token_indices` but that's one point that needs verification and careful review.

Things that I did not explicitly address as of yet:
- I'm unsure about how weight-tying comes into play here, writing an explicit test for this is one of my immediate next steps but I think that it should be fine as long as we restore the embedding weight matrix properly
- `get_peft_model_state_dict` will probably also mark the embedding layer as target since it is a valid embedding layer name. this is useless. we could prevent this by overriding the default setting for `save_embedding_layers` but unsure if it is a good idea. We could also just tell the user that they can delete the weights if they want to. Not sure about this yet.

### Open tasks

- [ ] add documentation and example on how to use this method (with LoRA and standalone)
- [x] Add custom model tests with `TrainableTokensConfig` being used directly
- [x] Add test that runs on GPU
- [x] allow for different `token_indices` per adapter
- [x] tests that cover having multiple custom token tuners at once
- [x] tests that cover having multiple targets for custom tokens


marcusinthesky:
Thanks for the shout-out. Looks super cool.

src/peft/utils/other.py
```diff
@@ -397,13 +465,65 @@ def set_adapter(self, adapter_name: str):
...
+        from peft.tuners.custom_tokens import CustomTokensLayer
+
+        # since super().__init__() calls update before we have a chance to initialise the adapter we would
+        # need here, we do the initialization here.
+        self.token_adapter = CustomTokensLayer(self.original_module, adapter_name, self.token_indices)
+
+    def _forward_wrapped(self, x, active_adapter, *args, **kwargs):
+        self.token_adapter.set_adapter(active_adapter)
+
+        y = self.token_adapter.forward(x)
```
BenjaminBossan, line 504: Better?

```suggestion
        y = self.token_adapter(x)
```

src/peft/utils/other.py
```diff
@@ -418,19 +538,36 @@ def _freeze_adapter(model, adapter_name):
...
-def _set_trainable(model, adapter_name, modules_to_save):
+def _set_trainable(model, adapter_name, module_names, mode="retrain", **wrapper_kwargs):
+    """Wraps modules that are supposed to be re-trained either normally, i.e. marking them to require gradients and
+    saving them alongside other modules, or with certain methods that go alongside PEFT methods, such as retraining
+    specific token indices using sparse matrices.
+
+    Note that you need to validate beforehand if there are layers shared between modes, e.g. if the 'embedding' layer
+    is configured for both mode `retrain` and `new_tokens` which would result in conflicts down the line.
+    """
+    if mode == "retrain":
```
BenjaminBossan, line 549: I wonder: instead of the mode argument, can we pass the `target_cls` directly? If setting `ModulesToSaveWrapper` as a default value results in circular imports, it could be set to `None` and then default to `ModulesToSaveWrapper` here.

src/peft/utils/save_and_load.py
```diff
@@ -134,6 +135,10 @@ def renamed_dora_weights(k):
         else:
             raise NotImplementedError
 
+    elif config.peft_type == PeftType.CUSTOM_TOKENS:
+        # TODO maybe add "{adapter_name}." as prefix to the filter string?
+        to_return = {k: state_dict[k] for k in state_dict if "custom_tokens_delta_tokens" in k}
```
BenjaminBossan, line 140: Is this block needed or would not already be covered by the `elif config.peft_type in list(PeftType):` below?

src/peft/utils/save_and_load.py
```diff
@@ -200,6 +205,11 @@ def renamed_dora_weights(k):
             if any(f"{module_name}.modules_to_save.{adapter_name}" in key for module_name in model.modules_to_save):
                 to_return[key.replace("modules_to_save.", "")] = value
 
+    # ADDITIONAL TRAINING MODULES
+    for name, module in model.named_modules():
+        if isinstance(module, AuxiliaryTrainingWrapper):
```
BenjaminBossan, line 210: If `AuxiliaryTrainingWrapper` is a `ModulesToSaveWrapper`, is this not a duplication of the explicit `modules_to_save` handling directly above?

src/peft/utils/save_and_load.py
```diff
@@ -345,6 +356,23 @@ def set_peft_model_state_dict(
     else:
         state_dict = peft_model_state_dict
 
+    # handle auxiliary training wrappers by getting each of them and translating saved state dict key
+    # (which does not include the adapter name) to loaded state dict key (which includes the adapter name).
+    for name, module in model.named_modules():
+        if isinstance(module, AuxiliaryTrainingWrapper):
+            for k, _ in module.adapter_state_dict(adapter_name).items():
```
BenjaminBossan, line 363:
```suggestion
            for k in module.adapter_state_dict(adapter_name).keys():
```

src/peft/utils/save_and_load.py
```diff
@@ -345,6 +356,23 @@ def set_peft_model_state_dict(
...
+                # each saved state dict is adapter specific, i.e. does not contain the adapter name
+                # but the loaded state dict does include adapter names since we can have multiple.
+                k_no_adapter = k.replace(f".{adapter_name}", "")
+                store_key = f"{name}.{k}"
+                lookup_key = f"{name}.{k_no_adapter}"
+
+                state_dict[store_key] = peft_model_state_dict[lookup_key]
+
+                # delete the old key from the previous `state_dict = peft_model_state_dict` statement.
+                # TODO can be removed once the model_to_save code is refactored
```
BenjaminBossan, line 373:
```suggestion
                # TODO can be removed once the modules_to_save code is refactored
```
?

Also, what would need to be refactored?

tests/test_custom_tokens.py
```diff
@@ -0,0 +1,97 @@
+import copy
```
BenjaminBossan, line 1: Missing copyright notice.

tests/test_custom_tokens.py
```diff
@@ -0,0 +1,97 @@
...
+        output_trn = peft_model.forward(output_hidden_states=True, **X)
+
+        peft_model.save_pretrained(save_path)
+        peft_model_org = peft_model
+
+        # check whether the token indices differ from the base model after loading the model
+        # from the checkpoint.
+        peft_model = AutoPeftModel.from_pretrained(save_path)
+        output_mod = peft_model.forward(output_hidden_states=True, **X)
+        output_org = original_model.forward(output_hidden_states=True, **X)
```
BenjaminBossan, line 44:
```suggestion
        output_mod = peft_model(output_hidden_states=True, **X)
        output_org = original_model(output_hidden_states=True, **X)
```

I was also confused about `_org`. Maybe change to `_orig`?

tests/test_custom_tokens.py
```diff
@@ -0,0 +1,97 @@
...
+        peft_model = get_peft_model(model, peft_config)
+        save_path = tmp_path / "stand_alone_usage"
+
+        # simulate normal use but take care to use the tokens that we expect to be modified
+        # (+1 that we don't expect to be modified)
+        X = {
+            "input_ids": torch.tensor([[0, 1, 2, 3]]),
+            "attention_mask": torch.tensor([[1, 1, 1, 1]]),
+        }
+        output_trn = peft_model.forward(output_hidden_states=True, **X)
```
BenjaminBossan, line 35: `_trn`?

tests/test_custom_tokens.py
```diff
@@ -0,0 +1,97 @@
...
+        W_trn = output_trn.hidden_states[0]
+
+        # all PEFT model embed outputs must equal the outputs during 'training' to make sure
+        # that saving/loading works properly.
+        assert torch.allclose(W_mod, W_trn)
+
+        assert not torch.allclose(W_mod[:, :3], W_org[:, :3])
+        assert torch.allclose(W_mod[:, 3:], W_org[:, 3:])
+
+    def test_combined_with_lora_usage(self, model, tokenizer, tmp_path):
```
BenjaminBossan, line 63: Would it make sense to refactor the test to avoid most duplication? Especially if we plan on supporting other PEFT methods too.

src/peft/peft_model.py
```diff
@@ -950,7 +951,30 @@ def set_additional_trainable_modules(self, peft_config, adapter_name):
             else:
                 self.modules_to_save.update(peft_config.modules_to_save)
             # this may add a new ModulesToSaveWrapper
-            _set_trainable(self, adapter_name, modules_to_save=peft_config.modules_to_save)
+            _set_trainable(self, adapter_name, module_names=peft_config.modules_to_save, mode="retrain")
+
+        if getattr(peft_config, "trainable_token_indices", None) is not None:
```
BenjaminBossan, line 956:
```suggestion
        if getattr(peft_config, "trainable_token_indices", None):
```

The problem is that `peft_config.trainable_token_indices` is `[]` by default, so we end up in this branch inadvertently. Fixing this will hopefully resolve the failing CI caused by `LNTuning`.

tests/test_custom_tokens.py
```diff
@@ -0,0 +1,97 @@
...
+    def model(self, model_id):
+        return AutoModelForCausalLM.from_pretrained(model_id)
+
+    @pytest.fixture
+    def tokenizer(self, model_id):
+        return AutoTokenizer.from_pretrained(model_id)
+
+    def test_stand_alone_usage(self, model, tokenizer, tmp_path):
+        original_model = copy.deepcopy(model)
+        peft_config = CustomTokensConfig(target_modules=["embed_tokens"], token_indices=[0, 1, 2])
```
BenjaminBossan, line 25: Just wondering: Would the test cover corner cases a little better if `token_indices` were not consecutive tokens starting at 0? So e.g. `[1, 3]` instead?

BenjaminBossan (changes requested):
Thanks a lot for picking up this feature, having this should be very useful for many PEFT users.

At this point, I have only done a quick review to iterate fast. In addition to my inline comments, I have some more general points:

1. Naming: I wonder if "custom tokens" is the right name for the feature. WDYT about "extra tokens", "trainable tokens", or "additional tokens"? Or even something like "sparse embedding update" or so? Let's just hold on a for a sec and ensure we find the best name, as we can't change it once the feature is out.
2. I'm just wondering out loud about whether a sparse matrix is the best way to implement this. If we have a high embedding dimension, the matrix will contain a lot of items, not sure if this could be inefficient. If it's implemented as a dense matrix (of course, only for the relevant columns), would that be possible? Perhaps via usage of `index_add` or `scatter_add`. I haven't investigated this option, just throwing some ideas out there.
3. Let's try to address as many TODOs as possible before merging or else they tend to stick around.
4. Did you run any realistic tests to ensure that this saves memory and reduces file size? I can help with that.
5. We should have updates to the docs and examples to show the standalone version and the LoRA integration. It would be fine to do that in a separate PR after this one but ideally it'll be added here.
Regarding testing:

Python 3.9 tests are failing because the `foo: bar | baz` type annotation syntax is not yet supported. Please add a `from __future__ import annotations` import where necessary.

Moreover, let's add a test case to `test_custom_models.py`, like here:

https://github.com/huggingface/peft/blob/6d033600e731cca368e69349e85a924b5cd72392/tests/test_custom_models.py#L71

This should result in nice bump in test coverage.

src/peft/peft_model.py
```diff
@@ -128,7 +128,8 @@ def __init__(
             ctx = init_empty_weights if low_cpu_mem_usage else nullcontext
             with ctx():
                 self.base_model = cls(model, {adapter_name: peft_config}, adapter_name)
-            self.set_additional_trainable_modules(peft_config, adapter_name)
+
+        self.set_additional_trainable_modules(peft_config, adapter_name)
```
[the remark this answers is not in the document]
githubnemo: Correct. I can move it back since it is currently out of scope but since there is currently no way for the user to specify neither `modules_to_save` nor `trainable_token_indices` it should be fine.

src/peft/tuners/custom_tokens/layer.py
```diff
@@ -0,0 +1,135 @@
...
+        self.num_total_embeddings = base_layer.num_embeddings  # total number of tokens in the vocabulary
+
+        # we set the number of trainable tokens
+        self.update_layer(adapter_name)
+
+    def update_layer(self, adapter_name):
+        # we initialize the delta embedding weights and store them in a trainable parameter
+        values = torch.rand(
+            (self.num_trainable_embeddings * self.base_layer.weight.shape[-1],)
+        )  # we initialize the values from a normal distribution N(0, 1), as in https://pytorch.org/docs/stable/generated/torch.nn.Embedding.html
```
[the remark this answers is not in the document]
BenjaminBossan: I think I'm wrong, since this would add the same value twice, right? So to keep the initial values, this would have to be zeros.

I'm wondering: If a user extends the embedding for new tokens and then uses the custom token tuner, after training time when they load the model, they would need to ensure that when they resize the embedding again, the seed is exactly the same right? Or else they would need to save the original embedding, but that would mean much larger file sizes for the adapter, which we want to avoid.

This is okay I guess, but not super user friendly. For instance, with `modules_to_save`, we don't have this issue as the adapter weights contain all the info we need (but of course it's a full copy of the original weights, so quite large).

In an ideal world, the extra params for the custom tokens would _replace_ to params of the original dict, so that users won't have to worry about restoring those. Not sure if that's possible. As an alternative, I wonder if we can save a checksum of the weights that are being replaced as a buffer and then, when loading, we can raise an error if the checksum does not match?

tests/test_custom_tokens.py
```diff
@@ -0,0 +1,97 @@
...
+        W_mod = output_mod.hidden_states[0]
+        W_org = output_org.hidden_states[0]
+        W_trn = output_trn.hidden_states[0]
+
+        # all PEFT model embed outputs must equal the outputs during 'training' to make sure
+        # that saving/loading works properly.
+        assert torch.allclose(W_mod, W_trn)
+
+        assert not torch.allclose(W_mod[:, :3], W_org[:, :3])
+        assert torch.allclose(W_mod[:, 3:], W_org[:, 3:])
```
BenjaminBossan, line 97: We should also ensure that there are tests that cover:

- multiple targets for custom tokens
- having multiple custom token tuners at once

githubnemo:
>     1. Naming: I wonder if "custom tokens" is the right name for the feature. WDYT about "extra tokens", "trainable tokens", or "additional tokens"? Or even something like "sparse embedding update" or so? Let's just hold on a for a sec and ensure we find the best name, as we can't change it once the feature is out.

Agreed. A part of me wants it to be more general but I think that `TrainableTokens` is general enough without being too specific. I'll change it.

>     2. I'm just wondering out loud about whether a sparse matrix is the best way to implement this. If we have a high embedding dimension, the matrix will contain a lot of items, not sure if this could be inefficient. If it's implemented as a dense matrix (of course, only for the relevant columns), would that be possible? Perhaps via usage of `index_add` or `scatter_add`. I haven't investigated this option, just throwing some ideas out there.

Naïvely I would expect that the sparse implementation is able to cope with this but I agree, it is not certain that this is the best way for implementing it (or the best single way, depending on the conditions). Let's skip this discussion before we don't have a benchmark in place.

>     3. Let's try to address as many TODOs as possible before merging or else they tend to stick around.

Yep. Most of the TODOs are points where I was unsure about how to proceed before the initial review(s). Getting on these now.
 
>     4. Did you run any realistic tests to ensure that this saves memory and reduces file size? I can help with that.

Nope, just functional tests. It would be great if you could do a bit of benchmarking, especially with the points from above regarding efficiency with larger embedding sizes.

>     5. We should have updates to the docs and examples to show the standalone version and the LoRA integration. It would be fine to do that in a separate PR after this one but ideally it'll be added here.

Yes, adding it as a to do item in the PR description.

BenjaminBossan:
> Agreed. A part of me wants it to be more general but I think that TrainableTokens is general enough without being too specific. I'll change it.

:+1: 

> It would be great if you could do a bit of benchmarking, especially with the points from above regarding efficiency with larger embedding sizes.

I'll do a comparison with what would be the current approach, adding the embedding to `modules_to_save`. I plan to check that tomorrow.

src/peft/tuners/custom_tokens/layer.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
+        if self.disable_adapters:
+            if self.merged:
+                self.unmerge()
+            result = self.base_layer(x, *args, **kwargs)
+        elif self.merged:
+            result = self.base_layer(x, *args, **kwargs)
+        else:
+            deltas = None
```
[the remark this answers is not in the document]
githubnemo: Wouldn't it be better to check `if self.disable_adapters or not self.active_adapters`? Otherwise `result` would be undefined.

src/peft/tuners/custom_tokens/model.py
```diff
@@ -0,0 +1,145 @@
...
+    @staticmethod
+    def _create_new_module(peft_config, adapter_name, target, **kwargs):
+        # Collect dispatcher functions to decide what backend to use for the replaced LoRA layer. The order matters,
+        # because the first match is always used. Therefore, the default layers should be checked last.
+        new_module = CustomTokensLayer(target, adapter_name, **kwargs)
+
+        return new_module
+
+    def _replace_module(self, parent, child_name, new_module, child):
+        # see https://github.com/huggingface/peft/blob/e5973883057b723b3f0fe3982bfa9d1e0c0fd8ec/src/peft/tuners/lycoris_utils.py#L300C4-L300C47
```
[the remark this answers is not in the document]
githubnemo: :shrug: deleted

src/peft/utils/save_and_load.py
githubnemo: Ah, of course it is :)

src/peft/utils/save_and_load.py
githubnemo: Yes. I thought about explicitly excluding that case but I don't see a benefit. I wanted to discuss if we can drop the code before that and do all the `modules_to_save` handling in the additional training modules section. WDYT?

src/peft/utils/save_and_load.py
githubnemo: In a perfect world I'd drop the `modules_to_save` handling above completely and just use `AuxiliaryTrainingWrapper` - that's what I meant with refactoring. Looking at it now it is, however, unlikely that the commented code will differ much. So the question is just: do we want to remove the `modules_to_save` handling and streamline it or do we want to keep it? I think it is safe to drop it but I wanted a second opinion beforehand.

src/peft/utils/save_and_load.py
BenjaminBossan: Hmm, would the benefit not be that we avoid doing the same work twice? From that perspective, I'd say let's drop it. Alternatively, we can add a comment to drop it later if we plan to refactor to unify the trainable module handling.

src/peft/peft_model.py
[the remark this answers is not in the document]
BenjaminBossan: Got it. I think it's fine, I just wanted to ensure it was not accidental.

BenjaminBossan:
Just noting this so we don't forget:

TODO
- Add custom model tests with `TrainableTokensConfig` being used directly
- Add test that runs on GPU

BenjaminBossan:
I tried a quick experiment that changes the implementation to avoid sparse matrices by using `index_add`. The new forward method looks like this:

```python
        if self.disable_adapters or not self.active_adapters:
            if self.merged:
                self.unmerge()
        elif self.merged:
        else:
            deltas = None
                #delta = self.get_sparse_delta_tokens(active_adapter)
                delta = self.trainable_tokens_delta_tokens[active_adapter]
                if deltas is None:
                    deltas = delta
                else:
                    deltas += delta

            W = self.base_layer.weight
            W = W.index_add(dim=0, index=torch.tensor(self.token_indices), source=deltas.reshape(len(self.token_indices), -1))

            result = F.embedding(
                input=x,
                weight=W,
                padding_idx=self.base_layer.padding_idx,
                max_norm=self.base_layer.max_norm,
                norm_type=self.base_layer.norm_type,
                scale_grad_by_freq=self.base_layer.scale_grad_by_freq,
                sparse=self.base_layer.sparse,
            )

        return result
```

`self.sparse_delta_tokens` can be removed. With this method, the same tests pass and fail as before. Thus it looks like a viable alternative. Once all tests pass, we should benchmark what works better.

I also found something that needs changing in `TrainableTokensLayer`. Right now, if we have multiple trainable token adapters at the same time, they cannot have different `token_indices`:

```
self.token_indices = token_indices
```

`self.token_indices` should instead be a dict that maps from adapter name to `token_indices`. Having different `token_indices` may be a bit niche, but we should still support it.

src/peft/tuners/custom_tokens/layer.py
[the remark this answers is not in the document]
githubnemo: I changed the initialization of the delta values to zero. I don't think that it is important to initialize them randomly as they're probably used in different contexts so random is not important. This is also nice because most tests assume that initializing a PEFT model does not change the parameters.

src/peft/utils/save_and_load.py
githubnemo: Dropped it.

src/peft/utils/save_and_load.py
githubnemo: I thought about this and I don't it makes sense to keep both code pieces in there. They should do the same and if they don't it is a bug either way. If this is making trouble in other tests we can discuss working around it / excluding modules to save explicitly.

tests/test_custom_tokens.py
githubnemo: Parametrized `peft_config`. The model is not parametrized as we can also test such combinations in testing common at a later point.

githubnemo:
I think I addressed most of the review points and also fixed quite a few test failures that arose from adding lora + `trainable_token_indices` in `test_custom_models`.

I've updated the list of unaddressed to dos which I will continue to work on in the PR's description.

BenjaminBossan (changes requested):
Thanks for the updates, this PR looks quite mature already. I left a few more comments, but they are mostly minor things.

There is, however, still the question if the approach of using sparse matrices and adding them is the best one. I prototyped a different approach based in `index_copy`, which I think could be better:

1. No handling of sparse matrices, which are not fully supported for all ops (e.g. order of addition matters).
2. Arguably, code is a bit simplified (no need for `get_sparse_delta_tokens`).
3. IMO, replacing the columns instead of adding on top is more user friendly: When the user extends the embedding matrix for new tokens, it is initialized randomly. The delta on these random tokens is then trained. When they later load the model, they would have to ensure that the tokens are again initialized the exact same way as before training, otherwise they're adding on top of other random numbers. Replacing the values sidesteps this problem.

I worked on a prototype for this change, here is the patch:

```diff
diff --git a/src/peft/tuners/trainable_tokens/layer.py b/src/peft/tuners/trainable_tokens/layer.py
index 9c754d73..be7d6500 100644
--- a/src/peft/tuners/trainable_tokens/layer.py
+++ b/src/peft/tuners/trainable_tokens/layer.py
@@ -25,6 +25,11 @@ from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
 
 
 class TrainableTokensLayer(nn.Module, BaseTunerLayer):
+    # All names of layers that may contain (trainable) adapter weights
+    adapter_layer_names = ("trainable_tokens_delta_tokens",)
+    # All names of other parameters that may contain adapter-related parameters
+    other_param_names = ("token_indices",)
+
     def __init__(
         self,
         base_layer: nn.Module,
@@ -60,25 +65,32 @@ class TrainableTokensLayer(nn.Module, BaseTunerLayer):
         # that some tests assume that after init should be the same as disabled adapters.
         # so either we let this break the assumptions or we initialize the deltas to zero.
         # since we're adding on top of the existing embeddings, I think that zero is sensible.
-        values = torch.zeros(self.num_trainable_embeddings * self.base_layer.weight.shape[-1])
+        # FIXME token_indices -> dict
+        values = self.get_base_layer().weight[self.token_indices].clone()
 
         # cause safetensors doesn't support sparse tensors, we need to store the values in a dense tensor and then convert it to a sparse tensor when called
         self.trainable_tokens_delta_tokens[adapter_name] = nn.Parameter(values, requires_grad=True)
+        self._move_adapter_to_device_of_base_layer(adapter_name)
+        1
+
+    def get_delta_weights(self, adapter_names: Optional[list[str]] = None):
+        adapter_names = check_adapters_to_merge(self, adapter_names)
+
+        if not adapter_names:
+            # no adapter to merge
+            return
+
+        for adapter_name in adapter_names:
+            deltas = None
+            for active_adapter in self.active_adapters:
+                delta = self.trainable_tokens_delta_tokens[adapter_name]
+                if deltas is None:
+                    deltas = delta
+                else:
+                    deltas += delta
 
-    def get_sparse_delta_tokens(self, adapter_name):
-        # we created the indices of the sparse tensor
-        r = torch.Tensor(self.token_indices).long()
-        c = torch.arange(self.embedding_dim)
-        indices = torch.stack(
-            [r.repeat(self.embedding_dim), c.repeat(self.num_trainable_embeddings, 1).t().reshape(-1)], dim=1
-        ).T
-
-        # we create the sparse tensor from our `delta_tokens` and `indices
-        return torch.sparse_coo_tensor(
-            indices=indices,
-            values=self.trainable_tokens_delta_tokens[adapter_name],
-            size=(self.num_total_embeddings, self.embedding_dim),
-        )
+        # FIXME token_indices -> dict
+        return deltas
 
     def merge(self, safe_merge: bool = False, adapter_names: Optional[list[str]] = None) -> None:
         adapter_names = check_adapters_to_merge(self, adapter_names)
@@ -87,24 +99,33 @@ class TrainableTokensLayer(nn.Module, BaseTunerLayer):
             # no adapter to merge
             return
 
-        for adapter_name in adapter_names:
-            orig_weights = self.base_layer.weight.data
-            merged = orig_weights + self.get_sparse_delta_tokens(adapter_name)
+        deltas = self.get_delta_weights(adapter_names)
+        if deltas is None:
+            return
+
+        orig_weights = self.base_layer.weight.data
+        # FIXME token_indices -> dict
+        index = torch.tensor(self.token_indices).to(orig_weights.device)
+        merged = orig_weights.index_copy(dim=0, index=index, source=deltas)
 
-            if safe_merge and not torch.isfinite(merged).all():
-                raise ValueError(f"NaNs detected in the merged weights. The adapter {adapter_name} seems to be broken")
-            else:
-                self.base_layer.weight.data = merged
-                self.merged_adapters.append(adapter_name)
+        if safe_merge and not torch.isfinite(merged).all():
+            raise ValueError(f"NaNs detected in the merged weights. The adapter(s) {adapter_names} seem to be broken")
+        else:
+            self.base_layer.weight.data = merged
+
+        for adapter_name in adapter_names:
+            self.merged_adapters.append(adapter_name)
 
     def unmerge(self) -> None:
         if not self.merged:
             warnings.warn("Already unmerged. Nothing to do.")
             return
 
-        while len(self.merged_adapters) > 0:
-            adapter_name = self.merged_adapters.pop()
-            self.base_layer.weight.data -= self.get_sparse_delta_tokens(adapter_name)
+        deltas = self.get_delta_weights(self.merged_adapters)
+        if deltas is not None:
+            index = torch.tensor(self.token_indices).to(self.base_layer.weight.device)
+            self.base_layer.weight.data = self.base_layer.weight.data.index_copy(dim=0, index=index, source=-deltas)
+        self.merged_adapters.clear()
 
     def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
         if self.disable_adapters or not self.active_adapters:
@@ -114,19 +135,15 @@ class TrainableTokensLayer(nn.Module, BaseTunerLayer):
         elif self.merged:
             result = self.base_layer(x, *args, **kwargs)
         else:
-            deltas = None
-            for active_adapter in self.active_adapters:
-                delta = self.get_sparse_delta_tokens(active_adapter)
-                if deltas is None:
-                    deltas = delta
-                else:
-                    deltas += delta
-
             W = self.base_layer.weight
+            deltas = self.get_delta_weights(self.active_adapters)
+            assert deltas is not None, "should not happen"
+            index = torch.tensor(self.token_indices).to(W.device)
+            W = W.index_copy(dim=0, index=index, source=deltas)
 
             result = F.embedding(
                 input=x,
-                weight=W + deltas,
+                weight=W,
                 padding_idx=self.base_layer.padding_idx,
                 max_norm=self.base_layer.max_norm,
                 norm_type=self.base_layer.norm_type,
```

WDYT? For me, the custom model tests pass the same as with the current PR. The new trainable tokens tests would need to be adapted a bit.

src/peft/utils/other.py
```diff
@@ -355,32 +372,117 @@ def forward(self, x: torch.Tensor, *args, **kwargs):
...
+        if adapter_name == self.active_adapter:
+            self.modules_to_save[adapter_name].requires_grad_(True)
+
+    def enable_adapters(self, enabled: bool):
+        """Takes care of setting the required_grad flag on the wrapped module.
+        If adapters are enabled, gradients for the module are required as well.
+        """
+        if self._disable_adapters is not enabled:
+            # already in the desired state, do nothing
+            return
```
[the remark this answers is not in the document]
githubnemo: Yeah, this bothers me and already caused a bug. If we say that the result of `enable_adapters` is idempotent we do not need the early return and we can simply remove the early return, call `super().enable_adapters` and be done with it. The overhead should be minimal and if it wasn't the person implementing could always opt for early returns. I'm going to remove the early return.

src/peft/utils/other.py
```diff
@@ -397,13 +499,108 @@ def set_adapter(self, adapter_name: str):
...
+        self, merge: bool, safe_merge: bool, adapter_names: Optional[list[str]]
+    ) -> torch.nn.Module:
+        """Unloading in case of `ModulesToSave` means to simply return the wrapped module.
+
+        However, if the wrapped module is itself a tuner, we'll call merge on it before.
+        """
+        new_module = self.modules_to_save[self.active_adapter]
+
+        # TODO: not sure if this is still a sensible thing to do. We would basically have to
+        # do the same checks as `_unload_and_optionally_merge` to support MHA, for example.
```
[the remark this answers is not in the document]
githubnemo: I thought about this a bit more. Since `AuxiliaryTrainingWrapper` not implements `unload_and_optionally_merge_module` this code would probably never be called. So not sure if keeping this is doing much but I may be missing a case here.

src/peft/utils/other.py
```diff
@@ -317,6 +325,15 @@ def _check_forward_args(self, x, *args, **kwargs):
             )
             raise ValueError(msg)
 
+    def _forward_wrapped(self, x: torch.Tensor, active_adpater: str, *args: Any, **kwargs: Any) -> torch.Tensor:
```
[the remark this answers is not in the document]
githubnemo: I tend to agree but then we would have to change adapters during `forward_mixed`. Would that be better?

tests/test_custom_models.py
```diff
@@ -103,6 +103,12 @@
         LoraConfig,
         {"target_modules": ["emb", "conv1d"], "use_dora": True},
     ),
+    (
+        "Embedding + transformers Conv1D 1 LoRA trainable_tokens",
+        "EmbConv1D",
+        LoraConfig,
+        {"target_modules": ["conv1d"], "trainable_token_indices": {"emb": [0, 10]}},
+    ),
```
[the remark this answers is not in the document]
githubnemo: Currently we only support `nn.Embedding` layers and the tested `MLP` class doesn't have one. Should we add one or add the ability to specify the model?

src/peft/utils/other.py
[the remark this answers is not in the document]
BenjaminBossan: Oh, that's actually problematic right now. After calling `_mixed_batch_forward`, the active adapters may be different from what they were before. Probably what we need is a dedicated `_mixed_batch_forward` method for `TrainableTokensLayer`. It would be mostly identical to `forward`, but I think a bit of duplication would be fine here. Then we should be able to avoid calling `set_adapter` completely during any kind of forward pass.

tests/test_custom_models.py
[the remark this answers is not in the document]
BenjaminBossan: Hmm, true. I think this requires slight refactoring of `MultipleActiveAdaptersTester` to add something like `get_model(tuner_method)` that returns a model with an embedding layer in this case.

src/peft/utils/other.py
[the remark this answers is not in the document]
githubnemo: Yep, fixed.

tests/test_trainable_tokens.py
```diff
@@ -0,0 +1,272 @@
...
+        return AutoTokenizer.from_pretrained(model_id)
+
+    def simulate_training(self, trainable_tokens_layer, adapter_name="default"):
+        """Simulates training of trainable_tokens adapter layer by assigning random
+        values to the delta tokens.
+        """
+        trained_values = torch.rand(
+            trainable_tokens_layer.num_trainable_embeddings * trainable_tokens_layer.base_layer.weight.shape[-1]
+        )
+        trainable_tokens_layer.trainable_tokens_delta_tokens[adapter_name].data = trained_values
```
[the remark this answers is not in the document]
githubnemo: We discussed this and settled on a WONT DO for now.

src/peft/utils/other.py
```diff
@@ -203,16 +203,36 @@ def shift_tokens_right(input_ids: torch.Tensor, pad_token_id: int, decoder_start
...
-        self.modules_to_save = torch.nn.ModuleDict({})
-        self._active_adapter = adapter_name
+        self._active_adapter = [adapter_name]
         self._disable_adapters = False
-        self.update(adapter_name)
+        self._adapters = set()
+
+        self.init_modules(adapter_name, **kwargs)
+
+        self.update(adapter_name, **kwargs)
```
BenjaminBossan, line 222: There is an issue with `TrainableTokensLayer.update_layer` being called twice. Once via `init_modules`, since `TrainableTokensLayer.__init__` calls `self.udpate_layer`. And then a second time via the `self.update` function. Either `TrainableTokensLayer.__init__` should not call `self.udpate_layer` or we should not call `update` here.

I think it should be the former proposal, since we generally init the layer params in a separate call and not directly through `__init__` of PEFT layers. If you make that change, you also need to ensure that `update_layer` is called when creating `TrainableTokensModel` directly (i.e. not using the feature via LoRA). For this, in the `_create_new_module` method, call

`new_module.update_layer(adapter_name, token_indices=kwargs["token_indices"])`

after initializing the `new_module`.

src/peft/utils/other.py
```diff
@@ -395,15 +529,123 @@ def set_adapter(self, adapter_name: str):
...
-        self.modules_to_save[self.active_adapter].requires_grad_(False)
+        if len(adapter_names) > 1:
+            raise ValueError(f"Attempted to set multiple ({adapter_names}) adapters at once for modules_to_save.")
+
+        adapter_name = adapter_names[0]
+
+        if adapter_name not in self._adapters:
+            raise ValueError(f"Adapter {adapter_name} not found in {self._adapters}")
+
+        self.modules_to_save[self.active_adapter[0]].requires_grad_(False)
```
BenjaminBossan, line 545:
```suggestion
        self.modules_to_save[self.active_adapters[0]].requires_grad_(False)
```

src/peft/utils/other.py
```diff
@@ -395,15 +529,123 @@ def set_adapter(self, adapter_name: str):
...
+    solely updating a few columns) using the `TrainableTokensLayer` PEFT method.
+    """
+
+    def __init__(
+        self,
+        module_to_save: torch.nn.Module,
+        adapter_name: str,
+        token_indices: list[int],
+    ) -> None:
+        super().__init__(module_to_save, adapter_name, token_indices=token_indices)
```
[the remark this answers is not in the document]
githubnemo: I wonder if we can just set `self.original_module = self.token_adapter.base_layer`. Disabling the adapter should make the base layer equal to the original module anyway. WDYT?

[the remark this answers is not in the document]
BenjaminBossan: Would it work via a `@property`?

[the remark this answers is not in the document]
githubnemo: Possibly. I just assigned it on init since it is a bit less indirection.

src/peft/tuners/custom_tokens/layer.py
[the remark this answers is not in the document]
marcusinthesky: @githubnemo I agree. 

As [Reparameterized PEFT](https://arxiv.org/pdf/2312.12148) aims train the delta's, initializing at zero-may help when the user accidentally initializes tokens which are not in their post-training/FT corpora. This may also be a safer option with ['Token Drag'](https://github.com/jmward01/lmplay/wiki/Unified-Embeddings).

src/peft/tuners/lora/config.py
```diff
@@ -273,6 +273,13 @@ class LoraConfig(PeftConfig):
             parameter when you want to apply LoRA to the ColumnParallelLinear and RowParallelLinear layers of megatron.
         megatron_core (`Optional[str]`):
             The core module from Megatron to use, defaults to `"megatron.core"`.
+        trainable_token_indices (`Optional[Union[List[int], dict[str, List[int]]]]`)
+            Lets you specify which token indices to selectively fine-tune without requiring to re-train the whole
+            embedding matrix using the `peft.TrainableTokensModel` method. You can either specify a list of indices
```
BenjaminBossan, line 278: > using the `peft.TrainableTokensModel` method.

I wonder if that's important info or can just be dropped.

src/peft/tuners/trainable_tokens/config.py
```diff
@@ -0,0 +1,84 @@
...
+from peft.utils import PeftType
+
+
+@dataclass
+class TrainableTokensConfig(PeftConfig):
+    """
+    Configuration for the `TrainableTokens` method.
+
+    This method makes certain tokens (identified by their indices) trainable without as large a memory footprint as
+    training the target matrix fully. The following arguments are need.
```
BenjaminBossan, line 30:
```suggestion
    training the target matrix fully.
```

I think the docstring should mention prominently that this is intended for training new tokens in the embedding matrix.

src/peft/utils/other.py
```diff
@@ -395,15 +529,125 @@ def set_adapter(self, adapter_name: str):
...
+            # check if the module is itself a tuner layer
+            if merge:
+                new_module.merge(safe_merge=safe_merge, adapter_names=adapter_names)
+            new_module = new_module.get_base_layer()
+
+        return new_module
+
+
+class TrainableTokensWrapper(AuxiliaryTrainingWrapper):
+    """Wraps a module (typically an embedding layer) that is supposed to be re-trained sparsely (i.e.
```
BenjaminBossan, line 577: "sparsely" does not quite fit anymore.

src/peft/utils/other.py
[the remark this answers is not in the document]
BenjaminBossan: What I mean is: If we use `@property`, would it prevent PyTorch from considering this to be 2 separate modules? Ideally, we would not get back the same embedding matrix twice when calling `model.state_dict()`. `named_modules()` already does the right thing it seems.

src/peft/utils/other.py
```diff
@@ -418,19 +662,54 @@ def _freeze_adapter(model, adapter_name):
...
+    model,
+    adapter_name,
+    module_names,
+    strict_module_check=False,
+    wrapper_cls: Optional[AuxiliaryTrainingWrapper] = None,
+    **wrapper_kwargs,
+):
+    """Wraps modules that are supposed to be re-trained either normally, i.e. marking them to require gradients and
+    saving them alongside other modules, or with certain methods that go alongside PEFT methods, such as retraining
+    specific token indices using sparse matrices.
```
BenjaminBossan, line 675: "sparse" no longer fits

tests/test_custom_models.py
```diff
@@ -2068,6 +2112,17 @@ def test_multiple_active_adapters_forward(
...
+        # the assumption that the output of the combined output of two adapters is != to the output of one
+        # adapter is not true for unmodified trainable tokens as they just mimic the existing embedding matrix.
+        # therefore, we modify the weights so that the adapter weights differs from the embedding weights.
+        if "trainable_tokens" in tuner_method:
+            peft_model.emb.token_adapter.trainable_tokens_delta["adapter_1"].data = torch.rand_like(
+                peft_model.emb.token_adapter.trainable_tokens_delta["adapter_1"].data
+            )
+            peft_model.emb.token_adapter.trainable_tokens_delta["adapter_2"].data = torch.rand_like(
+                peft_model.emb.token_adapter.trainable_tokens_delta["adapter_2"].data
+            )
```
BenjaminBossan, line 2124: This is necessary because, when used with LoRA, trainable tokens has no option to set `init_weights=False`, right?

tests/test_custom_models.py
```diff
@@ -2127,13 +2185,26 @@ def test_multiple_active_adapters_merge_and_unmerge(
...
         model = get_peft_model(model, config_1)
 
+        # the assumption that the output of the combined output of two adapters is != to the output of one
+        # adapter is not true for unmodified trainable tokens as they just mimic the existing embedding matrix.
+        # therefore, we modify the weights so that the adapter weights differs from the embedding weights. in this
+        # case we even use 20*rand to be very distinct to adapter 2 since we're comparing outputs and not embeddings
+        # with rather high tolerance values. this is also the reason why `init_weights` is not sufficient here and
+        # when using `<peft method>.trainable_token_indices` we do not have the utility of `init_weights` anyway.
+        if "trainable_tokens" in tuner_method:
+            model.emb.token_adapter.trainable_tokens_delta["default"].data = 20 * torch.rand_like(
```
BenjaminBossan, line 2204: So I tried out this test again and it passed for me with `2 *` instead of `20 *`. Not sure what is different than last time, but let's go with `2`? Test on CPU and GPU.

tests/test_trainable_tokens.py
```diff
@@ -0,0 +1,505 @@
...
+        output_train = peft_model.forward(output_hidden_states=True, **X)
+
+        peft_model.save_pretrained(save_path)
+        peft_model_org = peft_model
+
+        # check whether the token indices differ from the base model after loading the model
+        # from the checkpoint.
+        peft_model = AutoPeftModel.from_pretrained(save_path)
+        output_load = peft_model.forward(output_hidden_states=True, **X)
+        output_orig = original_model.forward(output_hidden_states=True, **X)
```
BenjaminBossan, line 93:
```suggestion
        output_load = peft_model(output_hidden_states=True, **X)
        output_orig = original_model(output_hidden_states=True, **X)
```

tests/test_trainable_tokens.py
```diff
@@ -0,0 +1,505 @@
...
+            "input_ids": torch.tensor([[0, 1, 2, 3, 4]]),
+            "attention_mask": torch.tensor([[1, 1, 1, 1, 1]]),
+        }
+
+        for step in range(3):
+            optimizer.zero_grad()
+            y_pred = model(**X)
+            loss = y_pred.logits.mean()
+            loss.backward()
+            optimizer.step()
```
BenjaminBossan, line 178: Add a check that the trainable tokens were updated?

BenjaminBossan (changes requested):
Thanks for the updates, this is almost good to go. I have some smaller comments, but those should not require big changes. Apart from those, two points:

1. Do we know what happens when embedding weights are tied?
2. Please add the GPU test that I shared with you to `test_gpu_examples.py`

docs/source/package_reference/trainable_tokens.md
```diff
@@ -0,0 +1,42 @@
...
+-->
+
+# Trainable Tokens
+
+The Trainable Tokens method provides a way to target specific token embeddings for fine-tuning without resorting to
+training the full embedding matrix or using a low-rank adapter. It is based on the initial implementation from
+[here](https://github.com/huggingface/peft/pull/1541).
+
+The method only targets specific tokens and selectively trains the token indices you specify. Consequently the
+required disk memory is also significantly lower than storing the full fine-tuned embedding matrix.
```
BenjaminBossan, line 24: You mention that disk memory is *also* lower but did not mention yet that VRAM is lower.

docs/source/package_reference/trainable_tokens.md
```diff
@@ -0,0 +1,42 @@
...
+
+Some preliminary benchmarks acquired with [this script](https://github.com/huggingface/peft/blob/main/scripts/train_memory.py)
+suggest that for `gemma-2-2b` (which has a rather large embedding matrix) you can save 4.8GiB VRAM with Trainable Tokens
+over fully fine-tuning. While LoRA will use even less memory (-6.3GiB total over fine-tuning) it might also target
```
BenjaminBossan, line 28:
```suggestion
over fully fine-tuning the embedding matrix. While LoRA will use even less memory (-6.3GiB total over fine-tuning) it might also target
```

BenjaminBossan: I ran the check again (all with LoRA rank 32 and bfloat16):

1. `modules_to_save=['embed_tokens']`

cuda memory avg: 9621MB
cuda memory max: 10880MB
total time: 11.78s
file size of checkpoint: 1149.4MB

2. LoRA on embedding

cuda memory avg: 5245MB
cuda memory max: 6988MB
total time: 9.60s
file size of checkpoint: 1180.9MB

3. Trainable tokens (6 indices)

cuda memory avg: 5117MB
cuda memory max: 6890MB
total time: 10.28s
file size of checkpoint: 24.4MB

So LoRA on embedding vs trainable tokens is pretty much on par when it comes to VRAM.

tests/test_gpu_examples.py
```diff
@@ -1550,6 +1550,98 @@ def on_optimizer_step(self, args, state, control, **kwargs):
             # assert loss is not None
             assert trainer.state.log_history[-1]["train_loss"] is not None
 
+    @pytest.mark.single_gpu_tests
```
BenjaminBossan, line 1553: You put this test into wrong class, which results in it trying to load a GPTQ quantized base model. Please put it in the previous test class at line ~1393.

tests/test_gpu_examples.py
```diff
@@ -4353,3 +4445,97 @@ def test_hotswapping_compiled_diffusers_model_does_not_trigger_recompilation(sel
...
+            bnb_config = BitsAndBytesConfig(
+                load_in_4bit=True,
+                bnb_4bit_quant_type="nf4",
+                bnb_4bit_compute_dtype=torch.float16,
+                bnb_4bit_quant_storage=torch.float16,
+                bnb_4bit_use_double_quant=True,
+            )
+
+            model = AutoModelForCausalLM.from_pretrained(
+                self.causal_lm_model_id,
```
BenjaminBossan, line 4466: This attribute is undefined, you can use `"facebook/opt-350m"`.

tests/test_gpu_examples.py
```diff
@@ -4353,3 +4445,97 @@ def test_hotswapping_compiled_diffusers_model_does_not_trigger_recompilation(sel
...
+                quantization_config=bnb_config,
+                device_map="auto",
+            )
+
+            # add 2 new tokens
+            tokenizer = AutoTokenizer.from_pretrained(self.causal_lm_model_id)
```
BenjaminBossan, line 4472: Same

BenjaminBossan (changes requested):
I did a pass on the testing and doc changes specifically. Overall looks good, just some smaller comments.

src/peft/tuners/lora/config.py
githubnemo: I like pointers like these because I am incentivized to follow up on this without looking at the code and learning something new.

tests/test_custom_models.py
githubnemo: Yes, added a clarifying comment.

tests/test_custom_models.py
githubnemo: If it was flaky before it will probably be flaky some point later. I don't see why we couldn't continue with 20. It is rather arbitrary.

docs/source/package_reference/trainable_tokens.md
githubnemo: Is this gemma2 2b again?

tests/test_gpu_examples.py
githubnemo: That's from being in the wrong test class.

githubnemo:
Thanks for the review :) Addressed your comments.

Weight-tying will be handled in a follow-up PR: https://github.com/huggingface/peft/pull/2399

docs/source/package_reference/trainable_tokens.md
```diff
@@ -0,0 +1,42 @@
...
+required RAM will be lower and disk memory is also significantly lower than storing the full fine-tuned embedding matrix.
+
+Some preliminary benchmarks acquired with [this script](https://github.com/huggingface/peft/blob/main/scripts/train_memory.py)
+suggest that for `gemma-2-2b` (which has a rather large embedding matrix) you can save 4.8GiB VRAM with Trainable Tokens
+over fully fine-tuning the embedding matrix. While LoRA will use even less memory (-6.3GiB total over fine-tuning) it might also target
+tokens you don't want to be changed. With less extreme embedding matrixes the difference might come out shorter as well.
+
+Note that this method does not add tokens for you, you have to add tokens to the tokenizer yourself and resize the
+embedding matrix of the model accordingly. This method will only re-train the embeddings for the tokens you specify.
+This method can also be used in conjunction with LoRA layers! See [`~peft.LoraConfig.trainable_token_indices`].
```
BenjaminBossan, line 33: Link does not appear to be working :/

image

docs/source/developer_guides/lora.md
```diff
@@ -272,6 +272,50 @@ trainer = Trainer(
 )
 ```
 
+## Efficiently train tokens alongside LoRA
+
+Sometimes it is necessary to not only change some layer's weights but to add new tokens as well. With larger models this can be a memory-costly endeavour. PEFT LoRA adapters support the `trainable_token_indices` parameter which allows tuning of other tokens alongside fine-tuning of specific layers with LoRA. This method only trains the tokens you specify and leaves all other tokens untouched. This saves memory and doesn't throw away learned context of existing token embeddings in contrast to when training the whole embedding matrix. Under the hood this method uses the [`~TrainableTokenLayer`].
```
BenjaminBossan, line 277: This link is also broken :/

image

src/peft/tuners/lora/config.py
BenjaminBossan: Makes sense.

docs/source/developer_guides/lora.md
```diff
@@ -272,6 +272,50 @@ trainer = Trainer(
...
+
+```py
+# for layer 'embedding'
```
BenjaminBossan, line 280:
```suggestion
# for layer 'embed_tokens'
```

BenjaminBossan (approved):
Great work, really thorough PR and it should be helpful to many users. I have noticed a few minor issues still with the docs, up to you if you want to fix them. Anyway, feel free to merge once the CI is green.

PR #2382, opened 2025-02-17
ENH Allow rank/alpha keys to be "fully qualified"

BenjaminBossan:
See https://github.com/huggingface/diffusers/pull/10808 for context.

Right now, if we have a key in `rank_pattern` or `alpha_pattern`, it is interpreted as a pattern to be matched against the module names of the model (basically it is an `endswith` match). The issue with this logic is that we may sometimes have false matches. E.g. if we have a model with modules `"foo"` and `"bar"`, and if `"bar"` also has a sub-module called `"foo"`, it is impossible to target just the outer `"foo"` but not `"bar.foo"`. (Note: It is already possible target only `"bar.foo"` and not `"foo"` by using `"bar.foo"` as key)

This PR adds the possibility to indicate to PEFT that a key should be considered to be the "fully qualified" key, i.e. a strict match should be performed instead of a pattern match. For this, users need to prefix the string `"FULL-NAME-"` before the actual key. In our example, that would be `"FULL-NAME-foo"`.

Notice that since the prefix contains `"-"`, there is no possibility that it could accidentally match a valid module name.

For now, this is a draft PR, until we can verify that it is sufficient to solve the original diffusers issue.


BenjaminBossan:
@sayakpaul I forgot about this PR, is it good from the diffusers point of view?

sayakpaul (approved):
It slipped through the crack from my end too. All good.

shethaadit (approved):

BenjaminBossan:
@sayakpaul I have converted this PR to draft to put it on hold for now, given what @githubnemo suggested and the discussion in #2413. Sorry for the inconvenience, but I think it's worth it to invest a bit more time to figure out the best way forward.

sayakpaul:
Then I guess the `diffusers` PR should be reverted as well?

BenjaminBossan:
There is no final decision yet, but maybe we can check if the solution proposed above is enough or not.

> why not use the caret operator to force a full match?

sayakpaul:
I will temporarily remove the conditional [branch](https://github.com/huggingface/diffusers/blob/363d1ab7e24c5ed6c190abb00df66d9edb74383b/src/diffusers/loaders/peft.py#L196) and settle with what we have in the [other branch](https://github.com/huggingface/diffusers/blob/363d1ab7e24c5ed6c190abb00df66d9edb74383b/src/diffusers/loaders/peft.py#L268C21-L271C49) in `diffusers`. Once there's a decision and this PR is merged, we can revisit that. Would that work for you?

BenjaminBossan:
Whatever works best for you is fine with me.

BenjaminBossan:
Superseded by #2419.

PR #2383, opened 2025-02-18
FIX: Skip X-LoRA tests for transformers >= 4.49.0

BenjaminBossan:
Since the latest transformers release of v4.49.0, X-LoRA tests are broken. The PR that caused it was:

https://github.com/huggingface/transformers/pull/35724

For the time being, let's skip the X-LoRA tests if this transformers version is detected and also advice users against using X-LoRA with this transformers version.


BenjaminBossan:
Closing in favor of #2384.

PR #2389, opened 2025-02-20
orthogonal lora layer init

winglian:
see: https://datta0.github.io/posts/rethink-lora-init/

### DIFF
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -349,7 +349,7 @@ class LoraConfig(PeftConfig):
         },
     )
     init_lora_weights: (
-        bool | Literal["gaussian", "eva", "olora", "pissa", "pissa_niter_[number of iters]", "corda", "loftq"]
+        bool | Literal["gaussian", "eva", "olora", "pissa", "pissa_niter_[number of iters]", "corda", "loftq", "orthogonal"]
     ) = field(
         default=True,
         metadata={
@@ -369,6 +369,7 @@ class LoraConfig(PeftConfig):
                 "nonnegative integer. "
                 "Passing `'corda'` results in CorDA initialization. "
                 "Pass `'loftq'` to use LoftQ initialization."
+                "Pass `'orthogonal'` to use orthogonal initialization."
             ),
         },
     )
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -188,6 +188,18 @@ class LoraLayer(BaseTunerLayer):
                 nn.init.kaiming_uniform_(self.lora_A[adapter_name].weight, a=math.sqrt(5))
             elif init_lora_weights.lower() == "gaussian":
                 nn.init.normal_(self.lora_A[adapter_name].weight, std=1 / self.r[adapter_name])
+            elif init_lora_weights.lower() == "orthogonal":
+                with torch.no_grad():
+                    rank = self.r[adapter_name]
+                    X = torch.randn(rank, rank)
+                    Q, _ = torch.linalg.qr(X)
+                    set1 = Q[0::2,:]  # Odd rows
+                    set2 = Q[1::2,:]  # Even rows
+                    dtype = self.get_base_layer().weight.dtype
+                    a_wt = torch.randn(self.in_features, rank//2).mm(set1).to(dtype).T/10.0
+                    b_wt = torch.randn(rank//2, self.out_features).T.mm(set2).to(dtype)/10.0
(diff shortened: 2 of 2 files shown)


src/peft/tuners/lora/layer.py
```diff
@@ -188,6 +188,18 @@ def reset_lora_parameters(self, adapter_name, init_lora_weights):
                 nn.init.kaiming_uniform_(self.lora_A[adapter_name].weight, a=math.sqrt(5))
             elif init_lora_weights.lower() == "gaussian":
                 nn.init.normal_(self.lora_A[adapter_name].weight, std=1 / self.r[adapter_name])
+            elif init_lora_weights.lower() == "orthogonal":
+                with torch.no_grad():
+                    rank = self.r[adapter_name]
+                    X = torch.randn(rank, rank)
+                    Q, _ = torch.linalg.qr(X)
+                    set1 = Q[0::2,:]  # Odd rows
+                    set2 = Q[1::2,:]  # Even rows
```
BenjaminBossan, line 197: `r` needs to be even for this to work, right? Let's check it and raise an error with a helpful message if it's not.

BenjaminBossan (changes requested):
Thanks a lot for adding an option for orthogonal initialization of LoRA weights.

Note that the [OLoRA initialization](https://huggingface.co/docs/peft/v0.14.0/en/developer_guides/lora#olora) is also aimed at orthogonal initialization. Maybe it would be worth it to compare the two. A disadvantage of OLoRA is, however, that the base weights are also modified, which requires users to take some extra steps if they want to load the model with other LoRA adapters, for instance. Pinging @tokenizer-decode just in case they wanna check this PR.

Before merging, we would also need some more additions to this PR:
- Update the docstring of `LoraConfig`, similar to the help. How about also adding a link to the blog post (AFAICT there is no paper?).
- Add a unit test. Check out [the tests in this test class](https://github.com/huggingface/peft/blob/bf186edc5babcdec7f33982b631f78eba844752c/tests/test_initialization.py#L67).
- Let's run `make style` to satisfy the linter.

src/peft/tuners/lora/config.py
```diff
@@ -369,6 +369,7 @@ class LoraConfig(PeftConfig):
                 "nonnegative integer. "
                 "Passing `'corda'` results in CorDA initialization. "
                 "Pass `'loftq'` to use LoftQ initialization."
+                "Pass `'orthogonal'` to use orthogonal initialization."
```
tokenizer-decode, line 372: I think this is confusing to the user.

BenjaminBossan:
@tokenizer-decode Thanks for commenting. It would indeed by nice to see a comparison with OLoRA or PiSSA, which the linked blog post didn't test. I could see an argument for the proposed initialization method being easier to use, as the base weights are unchanged, so even if it's not as good, there could be some value. WDYT?

tokenizer-decode:
I honestly don't see the performance benefit. But if you think there is an ease of use benefit, there could be some value. 

This goes for every other decomposition method, SVD e.g.. If the value is not updating the base weights, we can always let the user use the method with a parameter like `no_update` and we would turn off the part where we update the base weights.

But I might add, for future readers who are confused, updating base weights is generally where you get the performance.

winglian:
Screenshot 2025-02-20 at 10 36 09 PM
here's GRPO + PEFT. olora initialization goes straight to 0.0 rewards after the first step. orthogonal outperforms dora too.

If it's easier, I can convert this so that the init_lora accepts a callable and users can provide their own initialization function

BenjaminBossan:
> here's GRPO + PEFT. olora initialization goes straight to 0.0 rewards after the first step.

Thanks for running the tests :tada: Is the script open so that we can check what might be going on with OLoRA?

> If it's easier, I can convert this so that the init_lora accepts a callable and users can provide their own initialization function

In general, we would like to avoid this, even though it could be practical. The reason is that we wouldn't be able to serialize the `LoraConfig` into JSON with values that are Python code.

In sum, I think we can still proceed with the orthogonal weight initialization method. As I mentioned, even if it did not outperform OLoRA or similar methods, it could still be valuable as a more user friendly option.

BenjaminBossan:
@winglian Do you have time to finish the PR? If not, let us know so that one of us can take over.

BenjaminBossan:
@winglian I finished up the PR in #2498, would be grateful if you could take a look. Of course, I would add you as a co-author (we could add `@datta0` as well).

BenjaminBossan:
@winglian I merged #2498, which supersedes this PR, so I'm closing it now. I added you as co-author.

PR #2398, opened 2025-02-25
Enable grad checkpointing after get_peft_model

SunMarc:
# What does this PR do ?

Fixes https://github.com/huggingface/transformers/issues/35826
This PR enables grad checkpointing even if the model is already converted to a PeftModel. I can add a test if needed, but where should I put it ? I see that you have `_test_training_gradient_checkpointing` but it is a common test. 

The following works:
```python
model = ...
model.enable_gradient_checkpointing()
config = ...
```

but not that (with this PR, it should work now) : 
```python
model = ...
config = ...
model.enable_gradient_checkpointing() # doesn't return an error or a warning, so the users think that it worked
```

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -1378,6 +1378,13 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
         card.text = "\n".join(lines)
         card.save(filename)
 
+    def gradient_checkpointing_enable(self):
+        if hasattr(self.base_model, "gradient_checkpointing_enable"):
+            self.base_model.gradient_checkpointing_enable()
+            self.base_model = self._prepare_model_for_gradient_checkpointing(self.base_model)
+        else:
+            raise AttributeError("gradient_checkpointing_enable is not defined")
+
 
 class PeftModelForSequenceClassification(PeftModel):
     """


BenjaminBossan:
Thanks for the PR @SunMarc. For my better understanding, under what circumstances would this come up?

So far, I thought we had subsumed this functionality under `prepare_model_for_kbit_training`:

https://github.com/huggingface/peft/blob/3dd26682f49244645fe18e12052ec12b10661edc/src/peft/utils/other.py#L96-L181

Maybe that's not a smart idea and having a separate method is preferable, I just need to understand.

BenjaminBossan:
Ah yes, I see, sorry for missing that. Indeed, it makes sense to enable the possibility like this since users may not use `prepare_model_for_kbit_training`.

I checked the existing `_test_training_gradient_checkpointing`. My first idea was as follows: add a new argument `enable_gradient_checkpointing_before_peft=True`. Thus, by default, the test would do the same thing as it does now. But when setting the argument to `False`, gradient checkpointing would only be enabled after calling `get_peft_model`. From my understanding, this should result in the error.

However, when I tried this, there was no error, even when moving `model.gradient_checkpointing_enable()` down after `get_peft_model`. So I'm not quite sure how to trigger the error. Is it required to use `Trainer`?

SunMarc:
Could you try with the following script ? : 

```python

import os
import torch
from transformers import  AutoModelForCausalLM, AutoTokenizer
from peft import get_peft_model, LoraConfig, TaskType
import copy

def main():
    train_data = {"input": "input test", "output": "output test"}
    model_name = "codellama/CodeLlama-13b-Instruct-hf"
    
    model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.float16,device_map="auto")
    model.config.pad_token_id = model.config.eos_token_id

    input_ids = tokenizer.encode(train_data["input"])
    output_ids = tokenizer.encode(train_data["output"])
    model_inputs_output = input_ids + output_ids + [tokenizer.eos_token_id]
    model_inputs_output = torch.tensor(model_inputs_output, dtype=torch.int64)
    labels = copy.deepcopy(model_inputs_output)
    labels[: len(input_ids)] = -1 # 
    example_mask = model_inputs_output.ge(0)
    label_mask = labels.ge(0)
    model_inputs_output[~example_mask] = 0
    labels[~label_mask] = -100
    train_dataset = {
            "input_ids": model_inputs_output.unsqueeze(0).to("cuda"),
            "attention_mask": example_mask.unsqueeze(0).to("cuda"),
            "labels": labels.unsqueeze(0).to("cuda")
        }

    lora_config = LoraConfig(
            r=8,  
            lora_alpha=16,  
            target_modules=["q_proj", "gate_proj", "v_proj", "o_proj", "up_proj", "k_proj", "down_proj"],  # 与llama-factory一致
            lora_dropout=0.05,  
            task_type= TaskType.CAUSAL_LM  
        )
    model.train()
    model.to("cuda")
    model.gradient_checkpointing_enable()

    output = model(**train_dataset)
    loss = output["loss"]
    print(f"loss: {loss.requires_grad}")


if __name__ == "__main__":
    main()
```

BenjaminBossan:
Yes, I can confirm that `loss.requires_grad is False`. But I don't see how this example differs from the unit test:

https://github.com/huggingface/peft/blob/3dd26682f49244645fe18e12052ec12b10661edc/tests/testing_common.py#L1214-L1246

(as mentioned, I moved the `model.gradient_checkpointing_enable()` call)

I checked if it could be the model, or if the loss needs to be calculated by the `forward` call, but both made no difference. Any idea?

BenjaminBossan:
@SunMarc Did you have time to investigate further yet?

BenjaminBossan:
ping @SunMarc

PR #2399, opened 2025-02-25
Trainable Tokens: Support for Weight Tying

githubnemo:
This is a follow-up PR of https://github.com/huggingface/peft/pull/2376 to add support for weight-tying. Do not merge before the other is not merged.

## What is this

Some models, such as gpt2, tie the weights between the LM head and the input embeddings for various reasons. If we use the trainable tokens adapter, we're changing the result of the `forward()` of the input embeddings but we do not change the weights (unless we `merge()`). This means that the changes are **not reflected** in the tied weights, such as the LM head, leading to wrong results when training.

## How it is solved

The current approach is searching for tied layers and putting `TrainableTokensLayer` adapters on them as well but initialized to use the parameters from the embedding layer's `TrainableTokensLayer`. This is done via the `tied_adapter` argument of `TrailableTokensLayer.__init__()`.

## What needs to be done

- [x] encoder-decoder model tests
- [x] support for standalone `TrainableTokens` adapter
- [x] more tests


src/peft/tuners/trainable_tokens/layer.py
```diff
@@ -37,24 +37,40 @@ def __init__(
...
-        self.trainable_tokens_original = BufferDict({})
+        #
+        # In case we have weight tying with another token adapter, we'll have no actual
+        # references on our own but use everything from the tied adapter.
+        if not self.tied_adapter:
+            self.trainable_tokens_delta = nn.ParameterDict({})
+            self.trainable_tokens_original = BufferDict({})
+            self.token_indices = {}
+        else:
+            self.trainable_tokens_delta = self.tied_adapter.trainable_tokens_delta
```
BenjaminBossan, line 61: IIRC, you mentioned that stuff like `model.state_dict()` already correctly works for this case thanks to transformers. Could you add a comment for that? When the model is not a transformers model, would we get the same param twice?

src/peft/tuners/trainable_tokens/layer.py
```diff
@@ -37,24 +37,40 @@ def __init__(
...
+            self.trainable_tokens_original = self.tied_adapter.trainable_tokens_original
+            self.token_indices = self.tied_adapter.token_indices
 
         # Mark the weight as unmerged
         self.merged_adapters = []
 
     def update_layer(self, adapter_name, **kwargs):
+        if kwargs.get("tied_adapter", None):
+            # in this case we don't have any say, we're just following whatever the tied
```
BenjaminBossan, line 70:
```suggestion
            # in this case we don't have any because we're just following whatever the tied
```

?

src/peft/tuners/trainable_tokens/layer.py
```diff
@@ -140,22 +166,31 @@ def forward_adapters(self, x: torch.Tensor, active_adapters, *args, **kwargs) ->
...
+                result = F.embedding(
+                    input=x,
+                    weight=W,
+                    padding_idx=self.base_layer.padding_idx,
+                    max_norm=self.base_layer.max_norm,
+                    norm_type=self.base_layer.norm_type,
+                    scale_grad_by_freq=self.base_layer.scale_grad_by_freq,
+                    sparse=self.base_layer.sparse,
+                )
+            elif isinstance(self.base_layer, torch.nn.Linear):
```
BenjaminBossan, line 184: This would not necessarily work with quantized models, right? I wonder if we can find a more robust way of handling this, but I'm not sure how exactly.

src/peft/tuners/trainable_tokens/model.py
```diff
@@ -41,51 +41,109 @@ def __getattr__(self, name: str):
...
-            self.model.get_input_embeddings(), TrainableTokensLayer
+        # In case of weight-tying we need to adapt the tied weights as well and use tie the embedding adapter.
+        #
+        # The TrainableTokensLayer supports being tied to another TrainableTokensLayer meaning that the layer will
+        # not do any changes on its own but solely rely on the weights from the tied adapter. We will search for the
+        # tied weights and put tied TrainableTokensLayer adapters on them, all tied to the adapter of the embedding
+        # matrix.
+        if (
+            model_config.get("tie_word_embeddings", False)
+            and self.model._tied_weights_keys is not None
```
BenjaminBossan, line 71: Same comment as above.

src/peft/tuners/tuners_utils.py
```diff
@@ -626,7 +626,8 @@ def _get_tied_target_modules(self, model: nn.Module) -> list[str]:
         model_config = self.get_model_config(model)
         if model_config.get("tie_word_embeddings"):
             for target_module in self.targeted_module_names:
-                if target_module in EMBEDDING_LAYER_NAMES:
+                # TODO discuss in PR if reasonable change
```
BenjaminBossan, line 629: Okay, let's discuss :) So what does this change?

src/peft/utils/other.py
```diff
@@ -583,8 +582,12 @@ def __init__(
         module_to_save: torch.nn.Module,
         adapter_name: str,
         token_indices: list[int],
+        tied_adapter=None,
     ) -> None:
-        super().__init__(module_to_save, adapter_name, token_indices=token_indices)
+        """Supports weight-tying to another adapter when passed a `tied_adapter` which is expected to be a
+        `TrainableTokensLayer`.
+        """
```
BenjaminBossan, line 589: Should be a comment?

src/peft/utils/other.py
```diff
@@ -595,17 +598,28 @@ def original_module(self):
...
     def _error_message_name(self):
         return "trainable_token_indices"
 
+    def _hasattr_wrapped(self, name, modules):
+        return name == "weight"
+
+    def _getattr_wrapped(self, name, modules):
+        # some models query self.wte.weight.dtype, some may query the weights directly. for the first case it is not
+        # necessary to do anything special but we don't know if is going to be `.dtype`. so we need to get the merged
+        # weits from the adapter.
```
BenjaminBossan, line 618:
```suggestion
        # weights from the adapter.
```

src/peft/utils/other.py
```diff
@@ -595,17 +598,28 @@ def original_module(self):
...
+        if name == "weight":
+            return modules["token_adapter"].get_merged_weights(self.token_adapter.active_adapters)
+        assert False, f"should never be reached, bad check in _hasattr_wrapped for {name}"
```
BenjaminBossan, line 621: Raise a proper error? Maybe with suggestion to open an issue on PEFT.

src/peft/utils/other.py
```diff
@@ -655,6 +675,18 @@ def unload_and_optionally_merge_module(
         return self.token_adapter.get_base_layer()
 
 
+def _get_input_embeddings_name(model):
+    if not hasattr(model, "get_input_embeddings"):
+        return None
+
+    input_embeddings = model.get_input_embeddings()
+    for name, module in model.named_modules():
+        if module == input_embeddings:
```
BenjaminBossan, line 684:
```suggestion
        if module is input_embeddings:
```
?

tests/test_trainable_tokens.py
```diff
@@ -533,18 +588,55 @@ def model_weight_tied(self, model_id_weight_tied):
...
+        assert hasattr(peft_model.model.model.embed_tokens, "token_adapter")
+        assert not hasattr(peft_model.model.lm_head, "token_adapter")
+
+    @pytest.mark.parametrize(
+        "peft_config",
+        [
+            LoraConfig(
+                target_modules="all-linear",
+                trainable_token_indices={"embed_tokens": [0, 1, 3]},
+            ),
```
BenjaminBossan, line 606: Did you plan to add more configs? If not, can the decorator be removed?

tests/test_trainable_tokens.py
```diff
@@ -533,18 +588,55 @@ def model_weight_tied(self, model_id_weight_tied):
...
+
+        # we have to find out if the input embedding tying is doing its job during forward.
+        # for this we can leverage the fact that  emb_out(1/emb_in(x))  is  embed_dim  on the
+        # diagonal iff  emb_in.weight == emb_out.weight.
+        token_indices = [0, 1, 2, 3]
+        emb_dim = 768
+        emb_in = peft_model.model.model.decoder.embed_tokens(torch.tensor([token_indices]))
+        emb_out = peft_model.model.lm_head(1 / emb_in)
+
+        assert all(torch.diag(emb_out[0]) == torch.tensor([emb_dim] * len(token_indices)))
```
BenjaminBossan, line 628: Is it better to use `torch.allclose`?

tests/test_trainable_tokens.py
```diff
@@ -533,18 +588,55 @@ def model_weight_tied(self, model_id_weight_tied):
...
-            peft_model = get_peft_model(model_weight_tied, peft_config)
+        peft_model = get_peft_model(model_weight_tied, peft_config)
+
+        # make it so that the input embeddings diverge. when the weights are tied this should
+        # reflect in the output embeddings as well.
+        self.simulate_training(peft_model.model.model.decoder.embed_tokens.token_adapter)
+
+        # we have to find out if the input embedding tying is doing its job during forward.
+        # for this we can leverage the fact that  emb_out(1/emb_in(x))  is  embed_dim  on the
+        # diagonal iff  emb_in.weight == emb_out.weight.
```
BenjaminBossan, line 622: Oh, hmm, okay, oh, ...

Also:
```suggestion
        # diagonal iff emb_in.weight == emb_out.weight.
```

tests/test_trainable_tokens.py
```diff
@@ -555,10 +647,86 @@ def test_weight_tying_raises_when_detected_standalone(self, model_weight_tied):
...
+
+    @pytest.mark.parametrize(
+        "peft_config",
+        [
+            LoraConfig(
+                target_modules="all-linear",
+                trainable_token_indices={"shared": [0, 1, 3]},
+            ),
+        ],
+    )
```
BenjaminBossan, line 692: Same argument as above

tests/test_trainable_tokens.py
```diff
@@ -555,10 +647,86 @@ def test_weight_tying_raises_when_detected_standalone(self, model_weight_tied):
...
+
+        # we have to find out if the input embedding tying is doing its job during forward.
+        # for this we can leverage the fact that  emb_out(1/emb_in(x))  is  embed_dim  on the
+        # diagonal iff  emb_in.weight == emb_out.weight.
+        token_indices = [0, 1, 2, 3]
+        emb_dim = base_model.config.d_model
+        emb_in = peft_model.model.encoder.embed_tokens(torch.tensor([token_indices]))
+        emb_out = peft_model.model.lm_head(1 / emb_in)
+
+        assert all(torch.diag(emb_out[0]) == torch.tensor([emb_dim] * len(token_indices)))
```
BenjaminBossan, line 711: Same argument as above

tests/test_trainable_tokens.py
```diff
@@ -619,3 +787,78 @@ def test_original_module_not_in_state_dict(self, model):
...
+    @pytest.fixture
+    def model_embed_multiple(self):
+        return ModelEmbedMultiple()
+
+    @pytest.mark.parametrize(
+        "model_fixture_name, getter",
+        [
+            ("model_emb", lambda model: model.emb),
+            ("model_embed_in", lambda model: model.embed_in),
+            ("model", lambda model: model.model.model.embed_tokens),
```
BenjaminBossan, line 812: It could be more prudent to use `operator.attrgetter` than `lambda` but maybe it's unproblematic here.

src/peft/utils/other.py
```diff
@@ -655,6 +675,18 @@ def unload_and_optionally_merge_module(
         return self.token_adapter.get_base_layer()
 
 
+def _get_input_embeddings_name(model):
```
BenjaminBossan, line 678: The function could get a `default` argument that it returns instead of `None`, like `getattr`, but no strong opinion.

tests/testing_common.py
```diff
@@ -263,6 +274,32 @@ def check_config_json(self, tmp_dirname, model):
         if hasattr(model, "config"):  # custom models don't have a config attribute
             assert config["base_model_name_or_path"] == model.config.to_dict()["_name_or_path"]
 
+    def perturb_trainable_token_weights_if_used(self, model, config_kwargs, adapter_name="default", weight=1.0):
```
BenjaminBossan, line 277:
```suggestion
    def perturb_trainable_token_weights_if_used(self, model, config_kwargs, adapter_name="default", scale=1.0):
```

Maybe more precise name?

tests/testing_common.py
```diff
@@ -615,9 +656,12 @@ def _test_merge_layers_nan(self, model_id, config_cls, config_kwargs):
             base_model_name_or_path=model_id,
             **config_kwargs,
         )
+
         model = get_peft_model(model, config)
         model = model.to(self.torch_device)
 
+        self.perturb_trainable_token_weights_if_used(model, config_kwargs)
```
BenjaminBossan, line 663: Ah, a bit unfortunate that this is necessary. WDYT about using an auto fixture that monkey patches `TrainableTokensLayer.update_layer` so that it calls "super" but then perturbs the weight?

BenjaminBossan (changes requested):
Thanks for adding support for trainable tokens with tied embeddings and enhancing the tests. This was more complex than I expected. Good work covering this many edge cases.

I have a couple of comments, but I think there is nothing major.

src/peft/tuners/trainable_tokens/layer.py
githubnemo: I meant to express that, as a tied layer, we don't have anything to do but to return. I'll clarify.

src/peft/tuners/tuners_utils.py
githubnemo: Ah, yes, this function was essentially broken for any but the most trivial case. Take a model like T5 which has more than one embedding layer, nested in the encoder and the decoder. There is also one embedding layer on the top-level of the module (`shared`) which ties `shared`, `encoder.embed_tokens` and `decoder.embed_tokens`. 

`_get_tied_target_modules` will not warn about training `encoder.embed_tokens` even though it is clearly a tied weight because `encoder.embed_tokens` is not in `EMBEDDING_LAYER_NAMES`.

From that perspective I think this change is reasonable. At the time I was worried that the ominous `EMBEDDING_LAYER_NAMES` may lead to false-positive warnings in case we're having a layer that is called `embedding` and are also using weight-tying for the input embedding layer (e.g., `embed_tokens`).

But this case is probably very niche.

tests/test_trainable_tokens.py
githubnemo: This is just in preparation for supporting other PEFT methods. I was thinking that all tests that are combined (peft method + trainable tokens) should be parametrized by config.

tests/test_trainable_tokens.py
githubnemo: In hindsight yes, I didn't know if there are models where it is a bit more complicated so I left it as lambdas. I can change it if you want.

src/peft/tuners/tuners_utils.py
BenjaminBossan: I see, that makes a lot of sense. Maybe we can add a dedicated test for that? I agree about the potential for false positives, but AFAICT, this only affects warnings*, so better a false positive than a false negative here. Let's add a short comment that summarizes what you just mentioned.

*a small danger could be if the method will be used for something more in the future, but arguably that will be an issue for then.

tests/testing_common.py
githubnemo: I think that something that changes weights should be plainly visible, even though it is not doing anything 80% of the time. If we change something about this arrangement, then it should focus on getting `init_weights` to work with trainable tokens IMO (or something else that involves less implicit magic than there already is).

src/peft/tuners/trainable_tokens/layer.py
githubnemo: I think at least for bnb that `isinstance(self.base_layer, torch.nn.Linear)` holds true. Or what do you mean?

src/peft/tuners/tuners_utils.py
BenjaminBossan: The test would mostly be to catch regressions on the functionality, but I agree it's not a priority. Having a good comment should get us 90% of the way there.

tests/test_trainable_tokens.py
BenjaminBossan: Makes sense.

tests/test_trainable_tokens.py
BenjaminBossan: I know that `lambda` has some weird scoping rules, but I can never remember what exactly it was. If it's not relevant here, it's okay to leave it as is.

tests/testing_common.py
BenjaminBossan: I can see that argument. Fine with leaving it as is.

src/peft/utils/other.py
githubnemo: I moved it to the class docstring but I disasgree that it should be a comment.

src/peft/tuners/trainable_tokens/layer.py
githubnemo: It turns out it is more subtle than that. While transformers keeps an eye out for tied weights on saving/loading we still have to deal with the tied weights we produce by using the tied adapter. I've added a test and a solution (wrapping the tied adapter in a list so that it is excluded from `.*modules()` calls).

My first idea was to wrap the parameters from the tied adapter in a `BufferDict` which turned out to be unnecessary with the list wrapping solution. That's the reason why there are now some fixes + tests for the buffer dict in this PR as well...

src/peft/tuners/trainable_tokens/layer.py
githubnemo: Hmm is it realistic that we address the issue in this PR? It seems rather like a separate can of worms.

BenjaminBossan: Okay, let's keep it for later. Then let's add a comment here so that we don't forget.

BenjaminBossan (approved):
Thanks for the last changes and for the fixes to `BufferDict`. PR LGTM.

PR #2403, opened 2025-02-27
FIX for ConvNd layers using the groups argument.

gslama12:
As discussed in [#2153](https://github.com/huggingface/peft/pull/2153).

Code [example](https://gist.github.com/gslama12/0a853f43418b1c104286b4bb4dd745f3)

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -1061,7 +1061,7 @@ class _ConvNd(nn.Module, LoraLayer):
         conv_layer = type(base_layer)
         out_kernel = out_stride = (1,) * (self._kernel_dim - 2)
         self.lora_A[adapter_name] = conv_layer(self.in_features, r, kernel_size, stride, padding, bias=False)
-        self.lora_B[adapter_name] = conv_layer(r, self.out_features, out_kernel, out_stride, bias=lora_bias)
+        self.lora_B[adapter_name] = conv_layer(r, self.out_features // self.base_layer.groups, out_kernel, out_stride, bias=lora_bias)
         self.lora_bias[adapter_name] = lora_bias
 
         if use_rslora:
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -1243,13 +1243,12 @@ class _ConvNd(nn.Module, LoraLayer):
                 3
             ) * self.scaling[adapter]
         else:
-            output_tensor = (
-                self.conv_fn(
-                    weight_A.transpose(0, 1),
-                    weight_B,
-                ).transpose(0, 1)
-                * self.scaling[adapter]
-            )
+            output_tensor = self.conv_fn(weight_A.transpose(0, 1), weight_B)
+
+            if self.base_layer.groups > 1:
+                output_tensor = output_tensor * self.scaling[adapter]
+            else:
+                output_tensor = output_tensor.transpose(0, 1) * self.scaling[adapter]
 
         if cast_to_fp32:
             output_tensor = output_tensor.to(dtype=dtype)
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -1245,7 +1245,7 @@ class _ConvNd(nn.Module, LoraLayer):
         else:
             output_tensor = self.conv_fn(weight_A.transpose(0, 1), weight_B)
 
-            if self.base_layer.groups > 1:
+            if self.get_base_layer().groups > 1:
                 output_tensor = output_tensor * self.scaling[adapter]
             else:
                 output_tensor = output_tensor.transpose(0, 1) * self.scaling[adapter]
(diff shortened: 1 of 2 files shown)


src/peft/tuners/lora/layer.py
```diff
@@ -1061,7 +1061,7 @@ def update_layer(
         conv_layer = type(base_layer)
         out_kernel = out_stride = (1,) * (self._kernel_dim - 2)
         self.lora_A[adapter_name] = conv_layer(self.in_features, r, kernel_size, stride, padding, bias=False)
-        self.lora_B[adapter_name] = conv_layer(r, self.out_features, out_kernel, out_stride, bias=lora_bias)
+        self.lora_B[adapter_name] = conv_layer(r, self.out_features // self.base_layer.groups, out_kernel, out_stride, bias=lora_bias)
```
BenjaminBossan, line 1064:
```suggestion
        self.lora_B[adapter_name] = conv_layer(r, self.out_features // base_layer.groups, out_kernel, out_stride, bias=lora_bias)
```

src/peft/tuners/lora/layer.py
```diff
@@ -1243,13 +1243,12 @@ def get_delta_weight(self, adapter) -> torch.Tensor:
...
-            output_tensor = (
-                self.conv_fn(
-                    weight_A.transpose(0, 1),
-                    weight_B,
-                ).transpose(0, 1)
-                * self.scaling[adapter]
-            )
+            output_tensor = self.conv_fn(weight_A.transpose(0, 1), weight_B)
+
+            if self.base_layer.groups > 1:
```
BenjaminBossan, line 1248:
```suggestion
            if self.get_base_layer().groups > 1:
```

BenjaminBossan (changes requested):
Thanks for the PR. The change seems to be much smaller than I initially thought, which is great. Before we proceed, could we do the following:

Let's add a test case for this. First, let's create an entry like this one:

https://github.com/huggingface/peft/blob/f51203f3e4d8b3856596d3296dae7180f15910c1/tests/test_custom_models.py#L114

Then we need to define a model with a conv layer that uses groups. Something similar to this with `groups=5` should work:

https://github.com/huggingface/peft/blob/f51203f3e4d8b3856596d3296dae7180f15910c1/tests/test_custom_models.py#L864-L880

Then make sure that the model is being used when it's model ID is passed by adding an entry similar to this one:

https://github.com/huggingface/peft/blob/f51203f3e4d8b3856596d3296dae7180f15910c1/tests/test_custom_models.py#L967-L968

LMK if anything is unclear.

Moreover, don't forget to run `make style` for the linter.

gslama12:
thanks for the fast reply. I implemented the test as you described. I added one for lora and one for dora. LMK if there is something missing.

gslama12:
I made an error in the test model, i fixed it now and the TC you stated should work. Let's see if anything else fails.

gslama12:
It seems like there is an issue with the merging TCs. I will try to look into it.. if you have any suggestions LMK.

BenjaminBossan:
Thanks for the updates but some tests are still failing on CI (ignore those caused by timeouts, that's a HF Hub issue). Checking the values, it doesn't look like it's just a matter of precision/tolerance but that there's something else going on. Do these tests pass locally for you?

gslama12:
No they don't. It seems like something with the merging procedure is off. I will try to look into it.

gslama12:
I tried to recreate the test `test_merge_layers_021_Conv2d_Groups_LoRA`, which is failing in the pipeline. Maybe you can checkout this [gist](https://gist.github.com/gslama12/7f2196900f733e99643bafcbdb1e5e17), which runs fine on my machine. I wonder if there is maybe something going on within the testing pipeline. 🤔

BenjaminBossan:
You don't need to create a standalone script to reproduce the error, just run pytest like so:

```sh
pytest tests/test_custom_models.py -k test_merge_layers_021_Conv2d_Groups_LoRA
```

With this, I can reproduce the error locally. Dropping into the debugger, I see:

```
(Pdb) logits
tensor([-0.0053, -5.2435], grad_fn=<SelectBackward0>)
(Pdb) logits_merged
tensor([-7.0452e-04, -7.2583e+00])
```

To run all the groups-related PRs, call this:

```sh
pytest tests/test_custom_models.py -k groups -v
```

gslama12:
Yes i am aware of that. I also get the same assertion error when running the test so my first thought was that i messed up the merging with my changes. But when i create a similar scenario and run it as a local file without the pipeline, the assertion seems to work. 

So i'm currently trying to figure out what the difference is.

gslama12:
Ahh ok, thanks for the help. I will try to debug this.

BenjaminBossan:
Yes, sounds good @gslama12, thanks. At the code location where the error is raised, let's add a comment with a short explanation and a link to this PR.

gslama12:
I have added the warning at init and the error message for merging. LMK if there are things left to do for this PR.

BenjaminBossan:
Thanks for the update. Unsurprisingly, now all the tests that involve merging are failing for the model using `groups`:

https://github.com/huggingface/peft/actions/runs/13751050130/job/38480944098?pr=2403#step:8:1232

I think there is no other solution but to edit all of these tests and add a skip (or early return) if this new model is being used.

gslama12:
I added the skip statements to the tests. Lets see if there are any other skips that are missing.

BenjaminBossan (approved):
Thanks for your great work on this, LGTM.

Issue #2342, opened 2025-01-23
CI: Add gptqmodel to the CI


BenjaminBossan:
### Feature request

This issue is to track the TODO from [this comment](https://github.com/huggingface/peft/pull/2247#pullrequestreview-2569656574). Once optimum 1.24.0 and transformers 4.49.0 are released, we should enable gptqmodel in the CI (and remove auto-gptq).

### Motivation

-

### Your contribution

-

BenjaminBossan:
Not stale. New optimum and transformers releases are out, we should be good to proceed.

Qubitium:
@BenjaminBossan  Affirmative. To be clear, should we clear out all autogptq specific base code or just remove autogptq ci tests and replace with gptqmodel tests only?

BenjaminBossan:
Let's go with the latter.

BenjaminBossan:
@Qubitium @jiqing-feng We forgot a bit about this, I wonder if it's safe now to remove gptq from PEFT completely and fully rely on gptqmodel. What is your opinion?

Also, since https://github.com/ModelCloud/GPTQModel/pull/1703 is merged, I wonder if we can remove auto-awq and also use gptqmodel instead.

jiqing-feng:
I am ok with it. How about you @Qubitium ?

Qubitium:
@BenjaminBossan  @jiqing-feng  Yes. I am in agreement with both AutoGPTQ and AutoAWQ deprecation. On the AutoAWQ front, not only have we merged AWQ support but we have sinced added two new AWQ kernels: `AwqTorch` (100% torch) and `AwqTorchFused` which is based on the work by @jiqing-feng for optimized `cpu` ops using Intel `avx`/`xmx`/`amx` instructions (we have since validated the kernel on AMD as well).

There may be `Peft` compat issues due to latest `GPT-QModel` refractors but I will work with `Peft` and @jiqing-feng  to fix any issues regarding `Peft` regardes to `kernel` selection, training, etc so we have a stable pinned release `Peft` can count on.

BenjaminBossan:
Fantastic, thanks you two. Let us know if we can help with anything.

Qubitium:
@BenjaminBossan  @jiqing-feng  Work has begun on the autogptq and autoawq deprecation. 

https://github.com/huggingface/peft/pull/2917
https://github.com/huggingface/transformers/pull/41567

I will ping when the PRs is ready for review and CI are passing.

[...]

PR #2404, opened 2025-02-28
FIX GPTQModel Lora Compat

Qubitium:
PR Changes:

* FIX GPTQ linear layers from GPTQModel is not compatible with PEFT Lora wrapper
* Skip `Scaling`  multiply ops if `Scaling == 1` (Unsure gpu is smart enough to no-op this micro optimizaton so doing this manually). The loras we are testing for GPTQ does not use Scale so `scale = r / lora_alpha` where `r == lora_alpha`

Notes:

* `GPTQLoraLinear` copies most of the code and structure from `AwqLoraLinear`. 

TODO:

* [ ] Add CI test for GPTQmodel + Lora

### DIFF
--- a/src/peft/import_utils.py
+++ b/src/peft/import_utils.py
@@ -52,7 +52,7 @@ def is_auto_gptq_available():
 @lru_cache
 def is_gptqmodel_available():
     if importlib.util.find_spec("gptqmodel") is not None:
-        GPTQMODEL_MINIMUM_VERSION = packaging.version.parse("1.9.0")
+        GPTQMODEL_MINIMUM_VERSION = packaging.version.parse("1.9.99")
         OPTIMUM_MINIMUM_VERSION = packaging.version.parse("1.23.99")
         version_gptqmodel = packaging.version.parse(importlib_metadata.version("gptqmodel"))
         if GPTQMODEL_MINIMUM_VERSION <= version_gptqmodel:
--- a/src/peft/import_utils.py
+++ b/src/peft/import_utils.py
@@ -13,12 +13,14 @@
 # limitations under the License.
 import importlib
 import importlib.metadata as importlib_metadata
+import logging
 import platform
 from functools import lru_cache
 
 import packaging.version
 import torch
 
+log = logging.getLogger(__name__)
 
 @lru_cache
 def is_bnb_available() -> bool:
@@ -50,9 +52,9 @@ def is_auto_gptq_available():
 
 
 @lru_cache
-def is_gptqmodel_available():
+def is_gptqmodel_available(prompt_install: bool = False):
     if importlib.util.find_spec("gptqmodel") is not None:
-        GPTQMODEL_MINIMUM_VERSION = packaging.version.parse("1.7.0")
+        GPTQMODEL_MINIMUM_VERSION = packaging.version.parse("1.9.0")
         OPTIMUM_MINIMUM_VERSION = packaging.version.parse("1.23.99")
         version_gptqmodel = packaging.version.parse(importlib_metadata.version("gptqmodel"))
         if GPTQMODEL_MINIMUM_VERSION <= version_gptqmodel:
@@ -62,18 +64,21 @@ def is_gptqmodel_available():
                     return True
                 else:
                     raise ImportError(
-                        f"gptqmodel requires optimum version {OPTIMUM_MINIMUM_VERSION} or higher. Found version {version_optimum}, "
-                        f"but only versions above {OPTIMUM_MINIMUM_VERSION} are supported"
+                        f"gptqmodel requires optimum version `{OPTIMUM_MINIMUM_VERSION}` or higher. Found version `{version_optimum}`, "
+                        f"but only versions above `{OPTIMUM_MINIMUM_VERSION}` are supported"
                     )
             else:
                 raise ImportError(
-                    f"gptqmodel requires optimum version {OPTIMUM_MINIMUM_VERSION} or higher to be installed."
+                    f"gptqmodel requires optimum version `{OPTIMUM_MINIMUM_VERSION}` or higher to be installed."
                 )
         else:
             raise ImportError(
-                f"Found an incompatible version of gptqmodel. Found version {version_gptqmodel}, "
-                f"but only versions above {GPTQMODEL_MINIMUM_VERSION} are supported"
+                f"Found an incompatible version of gptqmodel. Found version `{version_gptqmodel}`, "
+                f"but only versions above `{GPTQMODEL_MINIMUM_VERSION}` are supported"
             )
+    elif prompt_install:
+        log.info("Please install GPTQModel for required functionality: `pip install -U gptqmodel --no-build-isolation -v`.")
+
 
 
 @lru_cache
--- a/src/peft/import_utils.py
+++ b/src/peft/import_utils.py
@@ -13,15 +13,11 @@
 # limitations under the License.
 import importlib
 import importlib.metadata as importlib_metadata
-import logging
 import platform
 from functools import lru_cache
 
 import packaging.version
 import torch
-
-
-log = logging.getLogger(__name__)
 
 
 @lru_cache
@@ -54,7 +50,7 @@ def is_auto_gptq_available():
 
 
 @lru_cache
-def is_gptqmodel_available(prompt_install: bool = False):
+def is_gptqmodel_available():
     if importlib.util.find_spec("gptqmodel") is not None:
         GPTQMODEL_MINIMUM_VERSION = packaging.version.parse("1.9.0")
         OPTIMUM_MINIMUM_VERSION = packaging.version.parse("1.23.99")
@@ -78,10 +74,6 @@ def is_gptqmodel_available(prompt_install: bool = False):
                 f"Found an incompatible version of gptqmodel. Found version `{version_gptqmodel}`, "
                 f"but only versions above `{GPTQMODEL_MINIMUM_VERSION}` are supported"
             )
-    elif prompt_install:
-        log.info(
-            "Please install GPTQModel for required functionality: `pip install -U gptqmodel --no-build-isolation -v`."
-        )
 
 
 @lru_cache
--- a/src/peft/import_utils.py
+++ b/src/peft/import_utils.py
@@ -20,7 +20,9 @@ from functools import lru_cache
 import packaging.version
 import torch
 
+
 log = logging.getLogger(__name__)
+
 
 @lru_cache
 def is_bnb_available() -> bool:
@@ -77,8 +79,9 @@ def is_gptqmodel_available(prompt_install: bool = False):
                 f"but only versions above `{GPTQMODEL_MINIMUM_VERSION}` are supported"
             )
     elif prompt_install:
-        log.info("Please install GPTQModel for required functionality: `pip install -U gptqmodel --no-build-isolation -v`.")
-
+        log.info(
+            "Please install GPTQModel for required functionality: `pip install -U gptqmodel --no-build-isolation -v`."
+        )
 
 
 @lru_cache
--- a/src/peft/import_utils.py
+++ b/src/peft/import_utils.py
@@ -52,8 +52,8 @@ def is_auto_gptq_available():
 @lru_cache
 def is_gptqmodel_available():
     if importlib.util.find_spec("gptqmodel") is not None:
-        GPTQMODEL_MINIMUM_VERSION = packaging.version.parse("1.9.99")
-        OPTIMUM_MINIMUM_VERSION = packaging.version.parse("1.23.99")
+        GPTQMODEL_MINIMUM_VERSION = packaging.version.parse("2.0.0")
+        OPTIMUM_MINIMUM_VERSION = packaging.version.parse("1.24.0")
         version_gptqmodel = packaging.version.parse(importlib_metadata.version("gptqmodel"))
         if GPTQMODEL_MINIMUM_VERSION <= version_gptqmodel:
             if is_optimum_available():
--- a/src/peft/tuners/lora/gptq.py
+++ b/src/peft/tuners/lora/gptq.py
@@ -115,8 +115,11 @@ def dispatch_gptq(
     cfg = kwargs.get("gptq_quantization_config", None)
 
     if is_gptqmodel_available():
-        new_module = GPTQLoraLinear(target, adapter_name, **kwargs)
-        target.qweight = target_base_layer.qweight
+        from gptqmodel.nn_modules.qlinear import BaseQuantLinear
+
+        if isinstance(target_base_layer, BaseQuantLinear):
+            new_module = GPTQLoraLinear(target, adapter_name, **kwargs)
+            target.qweight = target_base_layer.qweight
     else:
         quant_linear = get_auto_gptq_quant_linear(cfg)
 
--- a/src/peft/tuners/lora/gptq.py
+++ b/src/peft/tuners/lora/gptq.py
@@ -83,12 +83,7 @@ class GPTQLoraLinear(torch.nn.Module, LoraLayer):
                 expected_dtype = result.dtype
                 x = self._cast_input_dtype(x, lora_A.weight.dtype)
 
-            # lora_dropout float value is not stored so we need to check for cls
-            if isinstance(dropout, torch.nn.Dropout):
-                output = lora_B(lora_A(dropout(x)))
-            else:
-                # dropout == Identity which is no-op if lora_dropout == 0.0
-                output = lora_B(lora_A(x))
+            output = lora_B(lora_A(dropout(x)))
 
             if requires_conversion:
                 output = output.to(expected_dtype)
--- a/src/peft/tuners/lora/gptq.py
+++ b/src/peft/tuners/lora/gptq.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from functools import lru_cache
 from typing import Any, Optional
 
 import torch
@@ -57,11 +56,6 @@ class GPTQLoraLinear(torch.nn.Module, LoraLayer):
             lora_bias=lora_bias,
         )
 
-    @lru_cache
-    def _adapter_in_lora_keys(self, adapter: str):
-        # only need to check lora_A as result is same for lora_B
-        return adapter in self.lora_A.keys()
-
     def forward(self, x: torch.Tensor):
         # note: logic differs from default Linear because merging is not supported
         result = self.quant_linear_module(x)
@@ -69,8 +63,9 @@ class GPTQLoraLinear(torch.nn.Module, LoraLayer):
         if self.disable_adapters:
             return result
 
+        lora_A_keys = self.lora_A.keys()
         for active_adapter in self.active_adapters:
-            if not self._adapter_in_lora_keys(active_adapter):
+            if active_adapter not in lora_A_keys:
                 continue
 
             lora_A = self.lora_A[active_adapter]
--- a/src/peft/tuners/lora/gptq.py
+++ b/src/peft/tuners/lora/gptq.py
@@ -184,7 +184,10 @@ class GPTQLoraLinear(torch.nn.Module, LoraLayer):
             output = lora_B(lora_A(dropout(x)))
             if requires_conversion:
                 output = output.to(expected_dtype)
-            output = output * scaling
+
+            if scaling != 1:
+                output = output * scaling
+
             result = result + output
         return result
 
--- a/src/peft/tuners/lora/gptq.py
+++ b/src/peft/tuners/lora/gptq.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+from functools import lru_cache
 from typing import Any, Optional
 
 import torch
@@ -22,7 +22,7 @@ from peft.tuners.tuners_utils import BaseTunerLayer
 from peft.utils import get_auto_gptq_quant_linear
 
 
-class QuantLinear(torch.nn.Module, LoraLayer):
+class GPTQLoraLinear(torch.nn.Module, LoraLayer):
     def __init__(
         self,
         base_layer,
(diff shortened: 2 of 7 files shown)


Qubitium:
@BenjaminBossan  @SunMarc  PR ready for review.  My co-worker is writing up the ci-test for this but I want to get the review started early, if possible, before `peft` ci test is ready. Our GPTQmodel test is passing.

src/peft/tuners/lora/layer.py
```diff
@@ -489,7 +489,12 @@ def _mixed_batch_forward(
             # getting the sub-batch, passing it to LoRA layers and updating the corresponding indices of the linear
             # layer output
             sub_batch = x[sub_batch_indices_list[i]].to(lora_A.weight.dtype)
-            lora_output = lora_B(lora_A(dropout(sub_batch))) * scaling
+
+            if scaling == 1:  # no scaling
+                lora_output = lora_B(lora_A(dropout(sub_batch)))
+            else:
+                lora_output = lora_B(lora_A(dropout(sub_batch))) * scaling
+
```
SunMarc, line 497: not sure to see the benefit of doing that

src/peft/tuners/lora/layer.py
```diff
@@ -721,7 +726,10 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
                 x = self._cast_input_dtype(x, lora_A.weight.dtype)
 
                 if not self.use_dora[active_adapter]:
-                    result = result + lora_B(lora_A(dropout(x))) * scaling
+                    if scaling == 1:  # no scaling
+                        result = result + lora_B(lora_A(dropout(x)))
+                    else:
+                        result = result + lora_B(lora_A(dropout(x))) * scaling
```
SunMarc, line 732: same

src/peft/tuners/lora/gptq.py
```diff
@@ -19,7 +19,7 @@
 from peft.import_utils import is_gptqmodel_available
 from peft.tuners.lora.layer import LoraLayer
 from peft.tuners.tuners_utils import BaseTunerLayer
-from peft.utils import get_auto_gptq_quant_linear, get_gptqmodel_quant_linear
```
SunMarc, line 22: since you don't use get_gptqmodel_quant_linear anymore, you can remove the function

src/peft/tuners/lora/gptq.py
```diff
@@ -109,14 +109,87 @@ def dispatch_gptq(
...
+        adapter_name,
+        r: int = 0,
+        lora_alpha: int = 1,
+        lora_dropout: float = 0.0,
+        init_lora_weights: bool = True,
+        use_rslora: bool = False,
+        use_dora: bool = False,
+        lora_bias: bool = False,
+        **kwargs,
+    ):
```
SunMarc, line 141: what's the difference with Quantlinear ?

SunMarc (commented):
Thanks ! Left a couple of comments

src/peft/tuners/lora/layer.py
Qubitium: Following test show that pytorch can't optimize out the obvious no-op. Doing the * 1 math is 2x slower for the 3 shapes I tested to roughly simluate the logic.  I think the same would be true if we ask the tensors to multiply by 0. It maybe be faster to zero all tensors then multiply by 0, or in this case a no-op of * 1

Benchmark Result:
```
Benchmarking for tensor shape: (256, 256)
Operation A: 0.000012 seconds per iteration
Operation B: 0.000005 seconds per iteration
----------------------------------------
Benchmarking for tensor shape: (512, 512)
Operation A: 0.000012 seconds per iteration
Operation B: 0.000006 seconds per iteration
----------------------------------------
Benchmarking for tensor shape: (1024, 1024)
Operation A: 0.000014 seconds per iteration
Operation B: 0.000006 seconds per iteration
----------------------------------------
```


```py
import torch
import timeit

tensor_shapes = [(256, 256), (512, 512), (1024, 1024)]
repeats = 100

def benchmark_operation(operation, x, y):
    # Warm-up
    for _ in range(10):
        _ = operation(x, y)

    timer = timeit.Timer(lambda: operation(x, y))
    time_taken = timer.timeit(number=repeats) / repeats
    return time_taken

scale = 1
def operation_A(x, y):
    return x + y * scale

def operation_B(x, y):
    if scale == 1:
        return x + y
    else:
        return x + y * scale

CUDA = torch.device("cuda:0")

for shape in tensor_shapes:
    print(f"Benchmarking for tensor shape: {shape}")

    x = torch.rand(shape).to(CUDA)
    y = torch.rand(shape).to(CUDA)

    # Benchmark A
    time_A = benchmark_operation(operation_A, x, y)
    print(f"Operation A: {time_A:.6f} seconds per iteration")

    # Benchmark B
    time_B = benchmark_operation(operation_B, x, y)
    print(f"Operation B: {time_B:.6f} seconds per iteration")

    print("-" * 40)
    ```

File: src/peft/tuners/lora/gptq.py (hunk 4 above)
Comment (Qubitium) [reply src/peft/tuners/lora/gptq.py:141]: You're right. There may be no need to for the new GPTQLoraLinear cls. Checking.

File: src/peft/tuners/lora/gptq.py [hunk 5]
```diff
@@ -65,8 +70,9 @@ def forward(self, x: torch.Tensor):
             return result
 
         for active_adapter in self.active_adapters:
-            if active_adapter not in self.lora_A.keys():
+            if not self._adapter_in_lora_keys(active_adapter):
                 continue
```
Qubitium, line 74: @SunMarc  Found another micro optimization opportunity. In the loop in `forward` it is allocating an iterable `keys` every single time. I don't see anywhere that updates the `self.lora_A` so I adde a `lru_cache` to remove this allocation after first foward pass.

src/peft/tuners/lora/gptq.py
```diff
@@ -77,10 +83,19 @@ def forward(self, x: torch.Tensor):
                 expected_dtype = result.dtype
                 x = self._cast_input_dtype(x, lora_A.weight.dtype)
 
-            output = lora_B(lora_A(dropout(x)))
+            # lora_dropout float value is not stored so we need to check for cls
+            if isinstance(dropout, torch.nn.Dropout):
+                output = lora_B(lora_A(dropout(x)))
+            else:
+                # dropout == Identity which is no-op if lora_dropout == 0.0
+                output = lora_B(lora_A(x))
```
Qubitium, line 91: @SunMarc  Another micro optimization no-op skip. `lora_dropout: float` is not stored so if `0.0` the `update_layers` assign an `Identity` module which does nothing no-op.  Only do dropout if dropout is activated `lora_dropout: float > 0.0`

src/peft/tuners/lora/gptq.py
Qubitium: Replaced with `QuantLinear` but I renamed to `GPTQLoraLinear` as the name is much clearer .

src/peft/tuners/lora/layer.py
```diff
@@ -712,16 +723,26 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
...
+                    if scaling == 1:  # no scaling
+                        if isinstance(dropout, nn.Dropout):
+                            result = result + lora_B(lora_A(dropout(x)))
+                        else:
+                            result = result + lora_B(lora_A(x))
+                    else:
+                        if isinstance(dropout, nn.Dropout):
+                            result = result + lora_B(lora_A(dropout(x))) * scaling
+                        else:
+                            result = result + lora_B(lora_A(x)) * scaling
```
Qubitium, line 745: Same micro optimzations from `gptq` changes.

src/peft/tuners/lora/layer.py
Qubitium: @BenjaminBossan  Our upcoming feature of GPTQModel + Lora will be 100% scale = 1 so that's why I want to make sure it is as performant as possible. Lora apply is already an extra cost on top of the normal module forward so any bit helps. I would predict, the faster your gpu, the more this helps, since faster gpus would iterate the module/layers even faster and result even more loops that hit this useless math.  

I will add a note on how some Loras, such as ones GPTQModel will generate, will for now, always be scale == 1.

src/peft/tuners/lora/layer.py
```diff
@@ -696,6 +702,11 @@ def get_delta_weight(self, adapter) -> torch.Tensor:
 
         return output_tensor
 
+    @lru_cache
```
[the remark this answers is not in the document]
Qubitium: @BenjaminBossan  I was looking for this code that may modify the self.adapters info post `GPTQLoraLinear` init. but I was not able to find it likely because I don't see the whole picture of PEFT as you do. Can you point me to me how user dynamically add new adapters, and remove them? Thanks.

src/peft/tuners/lora/gptq.py
Qubitium: Just did a benchmark and it does not affect runtime. Looke pytorch is able to optimize it away. Will remove.

src/peft/tuners/lora/layer.py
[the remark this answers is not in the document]
BenjaminBossan: Yeah, it's not easy to figure out where that happens. The actual logic is here:

https://github.com/huggingface/peft/blob/f51203f3e4d8b3856596d3296dae7180f15910c1/src/peft/tuners/tuners_utils.py#L789-L804

The relevant attributes for LoRA are defined here:

https://github.com/huggingface/peft/blob/f51203f3e4d8b3856596d3296dae7180f15910c1/src/peft/tuners/lora/layer.py#L42

src/peft/tuners/lora/gptq.py
```diff
@@ -110,13 +115,16 @@ def dispatch_gptq(
     cfg = kwargs.get("gptq_quantization_config", None)
 
     if is_gptqmodel_available():
-        device_map = kwargs.get("device_map", None)
-        quant_linear = get_gptqmodel_quant_linear(cfg, device_map=device_map)
+        from gptqmodel.nn_modules.qlinear import BaseQuantLinear
+
+        if isinstance(target_base_layer, BaseQuantLinear):
+            new_module = GPTQLoraLinear(target, adapter_name, **kwargs)
+            target.qweight = target_base_layer.qweight
```
Qubitium, line 122: > What exactly was broken?

Using transformer to load the model and then use `model.use_adapter()` we hit a problem where the pre-PR code would select a quant linear with wrong device_map == none == `cpu` when  the model was loaded on `cuda` so the `gptqmodel_select_quant_linear()` returns TorchQuantLinear for device_map==none==`cpu` which mismatches the target_base_linear which is `MarlinQuantLinear` (correct) causing this line to mistach ```if quant_linear is not None and isinstance(target_base_layer, quant_linear):```


```
DEBUG: selected_quant_linear=`<class 'gptqmodel.nn_modules.qlinear.torch.TorchQuantLinear'>`
DEBUG: target_base_layer=`MarlinQuantLinear`
```

```py
# old code:
if is_gptqmodel_available():
        device_map = kwargs.get("device_map", None)
        quant_linear = get_gptqmodel_quant_linear(cfg, device_map=device_map)
    else:
        quant_linear = get_auto_gptq_quant_linear(cfg)

    if quant_linear is not None and isinstance(target_base_layer, quant_linear):
        new_module = QuantLinear(target, adapter_name, **kwargs)
        target.qweight = target_base_layer.qweight
```


So the weird thing is device_map is wrong. But instead of fixing this, I saw that at this point, the quant linear is always BaseQuantLInear so I just blindy did the GPTQLoraLInear wrap as that would always work. 

Would the GPTQ `dispatcher` be called on non-gptq modules? Maybe I am not getting the full picture of when/whom the GPTQ dispatcher is called upon. Is the dispatcher like a serial prober peft sends to all modules to say, hey, is this module yours (GPTQ)? If so, do th lora thing..if not, do nothing.

src/peft/tuners/tuners_utils.py
```diff
@@ -168,7 +168,7 @@ def __init__(
         if not hasattr(self, "peft_config"):
             self.peft_config = {adapter_name: peft_config} if isinstance(peft_config, PeftConfig) else peft_config
         else:
-            logger.info(
+            logger.warning(
```
Qubitium, line 171: @BenjaminBossan  Based on youur previous msg, peft should only do warning, and it looks like a warning message to me, not level == info.

src/peft/tuners/lora/layer.py
```diff
@@ -711,17 +717,24 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
         else:
             result = self.base_layer(x, *args, **kwargs)
             torch_result_dtype = result.dtype
+
+            lora_A_keys = self.lora_A.keys()
             for active_adapter in self.active_adapters:
-                if active_adapter not in self.lora_A.keys():
+                if active_adapter not in lora_A_keys:
                     continue
```
Qubitium, line 724: @BenjaminBossan  I did micro benchmarks on this code too. And lru cache and old code had exact same performance which is very strange as the `dict.keys` must make a copy/clone of at the very least the `list` for the keys. But benchmarks shows there is no `speed` difference. Still, I am a firm believer that nothing is magic so I removed the unsafe `lru_cache` with a var outsid the loop so the same values list object is not allocated in a loop.

src/peft/tuners/lora/layer.py
```diff
@@ -711,17 +717,24 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
...
                 scaling = self.scaling[active_adapter]
                 x = self._cast_input_dtype(x, lora_A.weight.dtype)
 
                 if not self.use_dora[active_adapter]:
-                    result = result + lora_B(lora_A(dropout(x))) * scaling
+                    # Loras such as EoRA will always be scaling == 1 so we can skip the no-op math
+                    if scaling == 1:
+                        result = result + lora_B(lora_A(dropout(x)))
+                    else:
+                        result = result + lora_B(lora_A(dropout(x))) * scaling
```
Qubitium, line 737: @BenjaminBossan  Comments added to why Loras, depending on implementation and target, may well be `scale == 1`

src/peft/tuners/lora/gptq.py
Qubitium: Ok. I see the code in `dispatcher` logic where the module is serially sent to all dispatcher and first one that returns not None `new_module` wins.

src/peft/tuners/lora/gptq.py
BenjaminBossan: Yes, within the `forward` call, this should be safe to do. I wouldn't expect any measurable speed up, but it shouldn't hurt.

BenjaminBossan (commented):
> Ouch. No wonder my ide can't find it. These types of dynamic code makes things real tough. Python's find_usage is nearly useless here.

I agree that this type of dynamism should be avoided if possible. However, I don't really see how else this can be implemented in a generic fashion, since each PEFT method can have different relevant attributes.

src/peft/tuners/lora/gptq.py
Qubitium: I have already removed the lru cache for all instance and just moved the allocation once outside the loop.

Qubitium:
@BenjaminBossan  Looks like everything is cleared for merge. But please don't merge until we have added CI tests.

src/peft/tuners/tuners_utils.py
Qubitium: done

tests/test_gptqmodel.py
```diff
@@ -347,3 +348,30 @@ def test_non_default_adapter_name(self):
...
+        assert isinstance(gate_proj_module, GPTQLoraLinear)
+        assert gate_proj_module.lora_A["default"].weight.data.shape[0] == 256
+        assert gate_proj_module.lora_B["default"].weight.data.shape[1] == 256
+
+        tokenizer = AutoTokenizer.from_pretrained(model_id)
+        inp = tokenizer("Capital of France is", return_tensors="pt").to(model.device)
+        tokens = model.generate(**inp)[0]
+        result = tokenizer.decode(tokens)
+
+        print("result: ", result)
```
Qubitium, line 377: @BenjaminBossan Unit test added which combines both simple lora plus  `LoraConfig.rank_pattern` which aligns with GPTQModel's `dynamic` per module or in this case, per-lora `rank` override. So using GPTQModel you can automatically generate HF compatible `lora` weight that have different `rank` per module/lora combo via the `rank_pattern` dict.

Qubitium:
@BenjaminBossan This PR is `green` on our end. Need to be paired with GPTQModel `main` to run. Let me know if you have another requests. We would love to this get this merged as we may have even more PRs in the pipeline. =) 

Also, is there a planned date for a new official release of `PEFT` on pypi?

tests/test_gptqmodel.py
```diff
@@ -347,3 +348,30 @@ def test_non_default_adapter_name(self):
         # sanity check
         assert n_trainable_default == n_trainable_other
         assert n_total_default == n_total_other
+
+    @staticmethod
+    def test_load_lora():
+        model_id = "ModelCloud/Llama-3.2-1B-gptqmodel-ci-4bit"
```
BenjaminBossan, line 354: Do you have a smaller model that could be used here? That would reduce the risk of getting a network timeout or full disk error on CI. If not, we can try how this one works out.

BenjaminBossan (changes requested):
Thanks for the updates. Overall, this looks good, I just added a few minor comments.

A bigger issue for now is the requirement for an as yet unreleased gptqmodel version. What is the plan here, do we want to wait for that release first before proceeding with this PR?

Also a minor note, building the wheel is relatively slow, do you have prebuilt wheels?

> Also, is there a planned date for a new official release of PEFT on pypi?

We just rolled out a bigger feature which requires a few follow up changes, after that I think we can do a release. I'd say probably start of next week.

tests/test_gptqmodel.py
Qubitium: Anything below 1B have massive quantization errors and would cause inference to be highly unstable which cause ci tests to be wildly unstable too. Let's stick with 1B unless we get errors.

Qubitium:
> Thanks for the updates. Overall, this looks good, I just added a few minor comments.
> 
> A bigger issue for now is the requirement for an as yet unreleased gptqmodel version. What is the plan here, do we want to wait for that release first before proceeding with this PR?
> 
> Also a minor note, building the wheel is relatively slow, do you have prebuilt wheels?

I am cutting and release GPTQModel v2.0 tonight. 

We also build and auto-download prebuild wheels for our pypi release. But post source release to pypi, it takes like 4-6 hours to generate all the different wheels. 

I will work to clear off your suggestions and we can proceed to merge. v2.0 will be release in the next 2-3 hours, followed by downloadable wheels another 3-4 hours after that.

src/peft/import_utils.py
```diff
@@ -52,8 +52,8 @@ def is_auto_gptq_available():
 @lru_cache
 def is_gptqmodel_available():
     if importlib.util.find_spec("gptqmodel") is not None:
-        GPTQMODEL_MINIMUM_VERSION = packaging.version.parse("1.9.99")
-        OPTIMUM_MINIMUM_VERSION = packaging.version.parse("1.23.99")
+        GPTQMODEL_MINIMUM_VERSION = packaging.version.parse("2.0.0")
+        OPTIMUM_MINIMUM_VERSION = packaging.version.parse("1.24.0")
```
Qubitium, line 56: @BenjaminBossan  Updated versions.

tests/test_gptqmodel.py
```diff
@@ -374,4 +371,4 @@ def test_load_lora():
         tokens = model.generate(**inp)[0]
         result = tokenizer.decode(tokens)
 
-        print("result: ", result)
+        assert "paris" in result.lower()
```
Qubitium, line 374: @BenjaminBossan We now assert on expected output. 1B is pretty stable so don't have to worry about flakiness.

Qubitium (commented):
@BenjaminBossan  Ci test updated with requested changes. 

However there is a AdaLora test failure that we inherited. Have no idea what it does. The test is failing at the AdaLora config state so there is probably changes to PEFT that is making it no llonger compatible. 

How to fix this?

```
FAILED test_gptqmodel.py::PeftGPTQModelTests::test_adalora_causalLM - ValueError: AdaLoRA does not work when `total_step` is None, supply a value > 0.
```

BenjaminBossan:
> Ci test updated with requested changes.

Nice. Let's wait for the release and the wheels and I'll give it a spin.

> However there is a AdaLora test failure that we inherited. Have no idea what it does. The test is failing at the AdaLora config state so there is probably changes to PEFT that is making it no llonger compatible.
> 
> How to fix this?

Could you try adding `total_step=6` when the `AdaLoraConfig` is initialized?

Qubitium:
> > Ci test updated with requested changes.
> 
> Nice. Let's wait for the release and the wheels and I'll give it a spin.
> 
> > However there is a AdaLora test failure that we inherited. Have no idea what it does. The test is failing at the AdaLora config state so there is probably changes to PEFT that is making it no llonger compatible.
> > How to fix this?
> 
> Could you try adding `total_step=6` when the `AdaLoraConfig` is initialized?

@BenjaminBossan  Pushed AdaLora fix. Had to tweak total_steps, init, and final_init values to get it to pass. 

GPTQModel v2.0.0 just released. Wheels are building: https://github.com/ModelCloud/GPTQModel/actions/runs/13641508684/job/38132256878

BenjaminBossan (changes requested):
Thanks, I installed the latest versions of gqpqmodel and optimum and the tests were generally running.

There was one small issue with `test_adalora_causalLM` on 2 GPUs failing. This is not a gptq specific issue, I've seen that frequently when using `Trainer`, as it will default to Data Parallel (the old one) when detecting multiple GPUs. So let's mark this test with `@pytest.mark.single_gpu_tests`.

Regarding the addition of gptqmodel dependency (and removal of auto-gptq) on the nightly CI (#2342), do we want to tackle this in a separate PR?

Qubitium:
> There was one small issue with `test_adalora_causalLM` on 2 GPUs failing. This is not a gptq specific issue, I've seen that frequently when using `Trainer`, as it will default to Data Parallel (the old one) when detecting multiple GPUs. So let's mark this test with `@pytest.mark.single_gpu_tests`.

Done. Marked. 

> Regarding the addition of gptqmodel dependency (and removal of auto-gptq) on the nightly CI (#2342), do we want to tackle this in a separate PR?

Yes, we will complete on a separate PR.

BenjaminBossan (approved):
Thanks for the fix and the additional test. CI is not compliant today, will merge despite the one failing run (after 6 attempts).

PR #2406, opened 2025-03-04
Weight LoRA

Vepricov:
**This PR brings a new method -- Weight LoRA.**

# WeightLoRA

Weight LoRA is a less complex, but important, PEFT method that adds a weight $w_i$ to each LoRA adapter (here i -- adapter number). This is done in order to perform, in addition to the classical optimisation over all LoRAs $A_1, B_1, ..., A_n, B_n$, an alternative optimisation over a vector of weights $w := (w_1, ..., w_n)^T \in R^n$ with a wide variety of constraints. In our research paper, we consider two approaches: 1) the vector $w$ must be in simplex $\Delta_{n-1}$, and 2) the vector $w$ has only $K$ non-zero coordinates. Both of these methods solve the problem of finding the most important LoRA adapters in the model and concentrating training on them while disabling the rest.

The abstract from the paper is:

The widespread utilization of language models in modern applications is inconceivable without Parameter-Efficient Fine-Tuning techniques, such as low-rank adaptation (LoRA), which adds trainable adapters to selected layers. Although LoRA may obtain accurate solutions, it requires significant memory to train large models and intuition on which layers to add adapters. In this paper, we propose a novel method, WeightLoRA, which overcomes this issue by adaptive selection of the most critical LoRA heads throughout the optimization process. As a result, we can significantly reduce the number of trainable parameters while maintaining the capability to obtain consistent or even superior metric values. Finally, we conduct experiments for the series of competitive benchmarks and DeBERTa and BART models, comparing our approach with the most popular LoRA modifications. The experimental results demonstrate the efficacy of WeightLoRA and the superior performance of WeightLoRA+ in comparison to the baselines in nearly all cases.

[Original code](https://github.com/brain-mmo-lab/WLoRA)

### DIFF
--- a/docs/source/package_reference/weightlora.md
+++ b/docs/source/package_reference/weightlora.md
@@ -1,0 +1,31 @@
+<!--Copyright 2023 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# WeightLoRA
+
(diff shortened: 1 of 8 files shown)


BenjaminBossan:
Thanks for this PR that proposes to add Weight LoRA to PEFT. Do you have a link to the full paper? I only skimmed the implementation, but from what I saw, this is basically LoRA but with the only difference being that the scaling parameter is trainable? Just from the abstract you pasted, it appears that there should be additional constraints on `w` that I don't see in the implementation.

Vepricov:
> I only skimmed the implementation, but from what I saw, this is basically LoRA but with the only difference being that the scaling parameter is trainable? 

Yes, you are right. However, the fact that the parameters `w_i` are trainable opens a great potential for methods that use alternative optimization with constraints on the weights `w`.

> Just from the abstract you pasted, it appears that there should be additional constraints on `w` that I don't see in the implementation.

These constraints should not be taken into account in the WeightLoRA method, but in the implementation of the optimizer step (e.g. SGD with projection). In our paper, we provide the [WeightAdam](https://github.com/brain-mmo-lab/WLoRA/blob/0da20c5c5d227fe09e7e1e23481e57690339579f/src/optimizers.py#L501C1-L501C18) optimizer, where we project weights `w` onto the desired set.

> Do you have a link to the full paper?

Unfortunately we submitted this paper to the ACL 2025 conference and it has double-blind review, therefore I cannot send the full text, but I can share, for example, the results of the experiment in the form of a table.

PR #2408, opened 2025-03-05
[tests] add the missing `.eval()` for inference

faaany:


tests/test_gpu_examples.py
```diff
@@ -1740,7 +1737,7 @@ def test_offload_merge(self):
         assert torch.allclose(post_unload_merge_olayer, pre_merge_olayer)
 
 
-@pytest.mark.skipif(not torch.cuda.is_available(), reason="test requires a GPU")
+@pytest.mark.skipif(not (torch.cuda.is_available() or torch.xpu.is_available()), reason="test requires a GPU")
```
BenjaminBossan, line 1740:
```suggestion
@pytest.mark.skipif(not (torch.cuda.is_available() or is_xpu_available()), reason="test requires a GPU")
```

BenjaminBossan (changes requested):
Thanks for extending the tests and making them XPU compatible. I found one small issue, the rest looks good.

Regarding the testing, did you run the bnb regression tests and the LoftQ tests successfully on XPU?

faaany:
> Thanks for extending the tests and making them XPU compatible. I found one small issue, the rest looks good.
> 
> Regarding the testing, did you run the bnb regression tests and the LoftQ tests successfully on XPU?

No, not all bnb and LoftQ tests pass. We only focus on cases that pass on CUDA but fail on XPU. If they fail on CUDA, we are not trying to fix them now.

BenjaminBossan:
> No, not all bnb and LoftQ tests pass. We only focus on cases that pass on CUDA but fail on XPU. If they fail on CUDA, we are not trying to fix them now.

Could you please share which CUDA tests are failing? Normally, that shouldn't happen.

BenjaminBossan (approved):
Thanks for adding the missing `.eval()` calls to the tests.

PR #2411, opened 2025-03-06
FIX: Reset the FP32 matmul precision in tests

BenjaminBossan:
Fixes [currently failing](https://github.com/huggingface/peft/actions/runs/13689803593/job/38280830007#step:6:1768) hotswap+compile tests that fail because outputs are not close enough before vs after compilation.

In `test_gpu_examples.py`, some tests run torchao, which sets the float32 matmul precision to "high". This in turn results in some models producing different outputs when compiled (but only for some seeds). Therefore, we need to ensure that the precision is reset to "highest", which is the default.

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -4263,6 +4263,15 @@ class TestHotSwapping:
 
     torch_device = infer_device()
 
+    @pytest.fixture(scope="class", autouse=True)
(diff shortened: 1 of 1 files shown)


githubnemo (approved):
Nice debugging :)

BenjaminBossan:
> Nice debugging :)

Same same.

PR #2414, opened 2025-03-09
Use baddbmm_() in LoRA code for memory savings

rationalism:
Use in-place baddbmm_() and addmm_() operations in LoRA forward() method, to save memory by doing in-place operations without creating new tensors. VRAM usage, fine-tuning Llama-3-8B on sequences of length 131,072 with QLoRA:

Before: 23456 MiB
After: 20386 MiB

Ran the tests on my local desktop


rationalism:
Never mind, have to fix a bug here

PR #2418, opened 2025-03-10
Merge LoRA Adapters with AWQ BaseModels [Experimental]

Whadup:
This PR extends the AwqLoraLinear class to allow merging in of LoRA Adapters. 
Instead of re-quantizing the whole model, we use the original quantization scales and zeros.

### DIFF
--- a/src/peft/tuners/lora/awq.py
+++ b/src/peft/tuners/lora/awq.py
@@ -21,7 +21,7 @@ from peft.import_utils import is_auto_awq_available
 from peft.tuners.lora.layer import LoraLayer
 from peft.tuners.tuners_utils import BaseTunerLayer
 
-from awq.utils.packing_utils import unpack_awq, reverse_awq_order, dequantize_gemm
+from awq.utils.packing_utils import unpack_awq, reverse_awq_order
 from awq.modules.linear.gemm import WQLinear_GEMM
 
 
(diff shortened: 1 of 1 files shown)


BenjaminBossan (changes requested):
Thanks for adding merging capabilities to AWQ. I only skimmed the PR so far, but could you please:

- Also implement the `unmerge` method? It should be very similar to the `merge` method, but remove the delta weight
- There should be a unit test to ensure that merging works, e.g. similar to [this test](https://github.com/huggingface/peft/blob/8edaae9460e4b76bce9431dc187402178ff7b689/tests/test_common_gpu.py#L1202-L1253) (without DoRA).
- Let's run `make style` on your changes.

Whadup:
@BenjaminBossan Thanks for looking into it already! Your three points are on my agenda, I will give you a ping when I commit the changes.

BenjaminBossan:
Great, thanks a lot.

BenjaminBossan:
@Whadup Do you still plan on working on this?

BenjaminBossan:
It's not quite clear to me, but it appears like AutoAWQ will be integrated into [llm-compressor](https://github.com/vllm-project/llm-compressor):

> AutoAWQ Integration: Perform low-bit weight-only quantization efficiently using AutoAWQ, now part of LLM Compressor. Note: This integration should be considered experimental for now.

PR #2419, opened 2025-03-11
ENH: Extend the regex for rank/alpha pattern

BenjaminBossan:
Supersedes #2382

Right now, the regex used to match the keys passed for `rank_pattern` and `alpha_pattern` requires that either:

1. The module name is identical to the key
2. The module name having a prefix and then ending on the key

This is restrictive, since it doesn't allow to disambiguate between all cases. E.g. if we have a model with these attributes:

- `model.foo`
- `model.bar.foo`

Then we:

- can target both by passing `"foo"` to the `rank_pattern` / `alpha_pattern` dict
- can target only `model.bar.foo` by passing `"bar.foo"`
- *cannot*  target only `model.foo`

This PR makes it possible to pass "^foo" as a key. This way, only `model.foo` but not `model.bar.foo` is targeted, as the latter's name does not start with `"foo"`.

As a general rule for users, if they intend to have a full match, they should pass the full name of the module preceded by a `^`. This is the least ambiguous way.

When running the test case with the old code, all the test cases with `^` will fail, which is fine, since `^` was not working anyway. At the same time, all test cases not using `^` pass, which means they are backwards compatible.

TODO: update docs

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -19,7 +19,7 @@ import os
 import re
 import warnings
 from contextlib import nullcontext
-from typing import Any, Optional, Union
+from typing import Any, Optional, Sequence, Union
 
 import accelerate
 import torch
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
@sayakpaul @hlky After some internal discussion, Marian and I came up with this new solution which would replace the initial proposal of having "fully qualified" keys from #2382. With this suggestion, it should be enough to add the `^` prefix to a layer name to make it "fully qualified" while not requiring any flag to toggle the semantics of the keys. If you could check that this new solution works to solve the initial problem in https://github.com/huggingface/diffusers/pull/10808, that would be great.

sayakpaul:
@BenjaminBossan thanks! Posted an update [here](https://github.com/huggingface/diffusers/pull/10985/commits/1c98875cbcdf8a0861b917362d3c717cea8cdbf5). 

TL;DR: works like a charm.

githubnemo (approved):
LGTM, thanks for the revision :)

sayakpaul:
Thanks, @BenjaminBossan. Would be nice to get a nudge once that happens. I will get my PR out of DRAFT.

PR #2423, opened 2025-03-11
Fixes the bug with not being able to handle loras targeting different sets of modulestosave

### LINKED ISSUE #2422: load_adapter Fails When modules_to_save Are Different for Each Adapter
### System Info

Python 3.11.9
transformers==4.40.2
peft==0.11.2

### Who can help?

@BenjaminBossan 

### Information

- [x] My own modified scripts

### Tasks

## Describe the Bug

When trying to **sequentially load multiple LoRA adapters** into a `PeftModel`, an error occurs if the `modules_to_save` setting is **different** between adapters.

## Steps to Reproduce

```python
import torch
from transformers import AutoModelForSequenceClassification, AutoTokenizer

# Load tokenizer and base model
model_name = "bert-base-uncased"
base_model = AutoModelForSequenceClassification.from_pretrained(model_name, num_labels=2)

# Create Adapter 1 (Saving 'classifier')
config1 = LoraConfig(
    r=8, lora_alpha=16, lora_dropout=0.1, bias="none",
    modules_to_save=["classifier"]
)
lora_model_1 = get_peft_model(base_model, config1)
lora_model_1.save_pretrained("lora_adapter_1")

# Create Adapter 2 (Saving 'classifier' again)
config2 = LoraConfig(
    r=8, lora_alpha=16, lora_dropout=0.1, bias="none",
    modules_to_save=["classifier"]
)
lora_model_2 = get_peft_model(base_model, config2)
lora_model_2.save_pretrained("lora_adapter_2")

# Create Adapter 3 (Saving 'pooler' instead of 'classifier')
config3 = LoraConfig(
    r=8, lora_alpha=16, lora_dropout=0.1, bias="none",
    modules_to_save=["pooler"]  # Different modules_to_save
)
lora_model_3 = get_peft_model(base_model, config3)
lora_model_3.save_pretrained("lora_adapter_3")

# Load base model
final_model = AutoModelForSequenceClassification.from_pretrained(model_name, num_labels=2)

# Load Adapter 1 into PeftModel
final_model = PeftModel.from_pretrained(final_model, "lora_adapter_1", adapter_name="adapter_1")

# Load Adapter 2 successfully
final_model.load_adapter("lora_adapter_2", adapter_name="adapter_2")

# Load Adapter 3 (Fails here!)
final_model.load_adapter("lora_adapter_3", adapter_name="adapter_3")  # ERROR
```

Traceback:

```
  File "/home/cc/polymorph/fine-tuning/hf_commit.py", line 47, in <module>
    final_model.load_adapter("lora_adapter_3", adapter_name="adapter_3")  # ERROR
  File "/home/cc/peft/src/peft/peft_model.py", line 1276, in load_adapter
  File "/home/cc/peft/src/peft/utils/save_and_load.py", line 347, in set_peft_model_state_dict
    key_map = module.adapter_state_dict_load_map(adapter_name)
  File "/home/cc/peft/src/peft/utils/other.py", line 550, in adapter_state_dict_load_map
    return {k: f"modules_to_save.{adapter_name}.{k}" for k in self.adapter_state_dict(adapter_name)}
  File "/home/cc/peft/src/peft/utils/other.py", line 553, in adapter_state_dict
    return self.modules_to_save[adapter_name].state_dict()
  File "/home/cc/miniconda3/envs/central/lib/python3.10/site-packages/torch/nn/modules/container.py", line 502, in __getitem__
    return self._modules[key]
KeyError: 'adapter_3'
```

## Expected Behavior

- The model should **allow loading multiple LoRA adapters** even if they have **different ****`modules_to_save`** settings.
- Instead, the PEFT library assumes **all adapters must have the same ****`modules_to_save`**, leading to a **`KeyError`** when adding an adapter with a new `modules_to_save` module.

## Possible Fix

This is because of the following line https://github.com/huggingface/peft/blob/37266c1baba367a9a31fc31f22e992a486e84f19/src/peft/utils/save_and_load.py#L342 that checks for all the loras in al the modules_to_save limiting it to only present loras in each modules to save wrapper should solve the problem. Also, https://github.com/huggingface/peft/blob/37266c1baba367a9a31fc31f22e992a486e84f19/src/peft/utils/other.py#L778 should be modified to make sure the modules_to_save is not disabled through a run of non existant lora.

saeid93:
Fixes #2422

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -776,6 +776,7 @@ def _set_adapter(model, adapter_name):
             # if the adapter is found in this module, set it as the active adapter, else disable the adapters of this
             # module
             if adapter_name in module._adapters:
+                module.enable_adapters(True)
                 module.set_adapter(adapter_name)
             else:
                 module.enable_adapters(False)
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -339,7 +339,7 @@ def set_peft_model_state_dict(
     # them and translating saved state dict key (which does not include the adapter name) to loaded state dict key
     # (which includes the adapter name).
     for name, module in model.named_modules():
-        if isinstance(module, AuxiliaryTrainingWrapper):
+        if isinstance(module, AuxiliaryTrainingWrapper) and adapter_name in module.modules_to_save.keys():
             # Not every module has a 1:1 mapping. ModulesToSaveWrapper, for example, removes the
             # `modules_to_save.{adapter_name}.` prefix. This prefix must be restored when loading the model from the
             # saved state dict which is why we fetch a load key map from the wrapper.


BenjaminBossan:
Thanks for the bug report and for providing a fix. I think this is a regression introduced by #2376, since your example works when I checkout PEFT v0.14.0.

Before proceeding, could you please create a unit test to catch the bug? It could be based on your original example and be placed [here](https://github.com/huggingface/peft/blob/main/tests/test_other.py). Just note that you should not re-use the same base model for `get_peft_model` but create a new one each time, since the base model is mutated, which can have unintended side effects.

githubnemo:
Hey @saeid93, thanks for your fix. I spent some time working on that issue. Good catch on the disabled adapter! I think that I've found a fix that doesn't limit the fix to `ModulesToSaveWrapper` and also supports other `AuxiliaryTrainingWrapper` such as `TrainableTokensWrapper`. It'd be great if you take a look at #2430 as well :)

saeid93:
Hi @BenjaminBossan and @githubnemo

Thank you both for your feedback.
I'll add the test and check #2340 when I get a chance. Just a note about your comment this is partially related to the recent changes by #2376 but even in older versions although it doesn't return an error it returns the wrong response as the LoRAs get disabled when iterated on a non-existant moduletosave at https://github.com/huggingface/peft/blob/8edaae9460e4b76bce9431dc187402178ff7b689/src/peft/utils/other.py#L781 . Even in previous version I had to access the module to save layer manually to enable the lora like below to make it enable it in my maincode:
```python
            model.classifiers.enable_adapters(True)
```
I just added a new line to ensure the lora is always enabled as @githubnemo also mentioned.

githubnemo:
> Hi @BenjaminBossan and @githubnemo
> 
> Thank you both for your feedback. I'll add the test and check #2340 when I get a chance. 

Before doing that, maybe take a look at PR #2430 first to avoid doing more work than necessary, I think it fixes both the issues you mentioned.

PR #2430, opened 2025-03-14
Fix #2422: Modules to save with multiple adapters

### LINKED ISSUE #2422: load_adapter Fails When modules_to_save Are Different for Each Adapter

githubnemo:
Using multiple adapters with different `modules_to_save` values leads to a scenario where it is implicitly assumed that each `ModulesToSaveWrapper` has a module for every loaded adapter. Since the adapters have different `modules_to_save` values this is not the case and retrieving the state dict fails with a key lookup error.

In addition to that, after disabling a `ModulesToSaveWrapper`, setting the adapter as active does not re-enable said adapter.

The new test is sensitive to both of these issues.

This change builds upon #2423.

### DIFF
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -1236,6 +1236,7 @@ class PeftCustomModelTester(unittest.TestCase, PeftCommonTester):
         outputs_base = model(**X)
         if issubclass(config_cls, (FourierFTConfig, TrainableTokensConfig)):
             config_kwargs = config_kwargs.copy()
+            # override the default value and make PEFT operation a no-op
             config_kwargs["init_weights"] = True
         config = config_cls(
             base_model_name_or_path=model_id,
@@ -1255,9 +1256,9 @@ class PeftCustomModelTester(unittest.TestCase, PeftCommonTester):
         model.train()
         # EmbConv1D is slow to learn for some reason
         lr = 0.01 if model_id != "EmbConv1D" else 1.0
-        if isinstance(config_cls, LNTuningConfig):
-            # LayerNorm tuning is slow to learn
-            lr = 1.0
+        if isinstance(config, TrainableTokensConfig):
+            # TrainableTokens is only changing a small subset, so we need a higher lr to see the difference
+            lr = 2.0
         optimizer = torch.optim.SGD(model.parameters(), lr=lr)
 
         # train at least 3 steps for all parameters to be updated (probably this is required because of symmetry


tests/test_other.py
```diff
@@ -107,6 +108,84 @@ def test_get_peft_model_revision_warning(tmp_path):
         _ = get_peft_model(base_model, lora_config, revision=overwrite_revision)
 
 
+def test_load_multiple_adapters_different_modules_to_save(tmp_path):
+    # This tests the error described in #2422 where loading multiple adapters with different modules_to_save
+    # attributes fails (due to a regression from #2376.
```
BenjaminBossan, line 113:
```suggestion
    # attributes fails (due to a regression from #2376).
```

BenjaminBossan (approved):
Thanks, not only for the fix but also for the additional comments. LGTM.

Let's wait 1 day to see if we get a response on #2423, and merge then.

PR #2353, opened 2025-01-29
ENH Allow disabling input dtype casting for LoRA

BenjaminBossan:
WIP

PR #2433, opened 2025-03-18
Fix multiple methods model dtype issues

BenjaminBossan:
As a user, it should be possible to manually cast the base model to a lower precision dtype, float16 or bfloat16, and still have the different PEFT methods work correctly. Currently, this is not the case for many PEFT methods, as can be replicated by the added tests.

To understand the problem, it helps to take a step back. By default, PEFT will treat the adapter weights with high precision, i.e. with float32. When the base model is lower precision, the user needs to pass inputs in lower precision too, as otherwise `self.base_layer(x)` would fail. However, this low precision input clashes with the high precision adapter weights.

The solution implemented in this PR is to cast the input to a higher dtype [1]. That way, the whole adapter operation is conducted in high precision. Only once that has finished will the final result be cast to the original dtype if necessary. This should lead to better results thanks to the high precision, but it may require more memory. Note that this is how LoRA is implemented, so the changes in this PR bring the other methods more in line with what LoRA does.

If the user does not want the adapter to be in float32, they can always pass `autocast_adapter_dtype=False` when calling `get_peft_model` or `PeftModel.from_pretrained` etc. This is also tested.

Besides adjusting the forward method to account for these changes, the `merge` and `unmerge` methods also often had to be adjusted, as they did not correctly account for the base model dtype. Now, those methods should always conserve the original dtype of the base model.

Note that if, for whatever reason, the input casting in [1] is not desired, users can use the `disable_input_dtype_casting` context manager to disable it (more context information on this feature can be found in PR #2353). I updated the corresponding code to be agnostic to the specific PEFT method (beforehand, it was only for LoRA). This is why I had to move the `_cast_input_dtype` method from LoRA to the `BaseTunerLayer`.

TODO: Notify the contributors of the individual methods. Their approval is not required for this PR to be merged, but they might be able to better identify any potential issues.

### DIFF
--- a/src/peft/tuners/ia3/layer.py
+++ b/src/peft/tuners/ia3/layer.py
@@ -234,6 +234,7 @@ class _ConvNd(nn.Module, IA3Layer):
         for active_adapter in adapter_names:
             if active_adapter in self.ia3_l.keys():
                 base_layer = self.get_base_layer()
+                orig_dtype = base_layer.weight.data.dtype
                 ia3_scaling = self.ia3_l[active_adapter].data
                 if not self.is_feedforward:
                     ia3_scaling = ia3_scaling.transpose(0, 1)
@@ -246,13 +247,13 @@ class _ConvNd(nn.Module, IA3Layer):
                             f"NaNs detected in the merged weights. The adapter {active_adapter} seems to be broken"
                         )
 
-                    base_layer.weight.data = output_weight
+                    base_layer.weight.data = output_weight.to(orig_dtype)
                 else:
-                    base_layer.weight.data = torch.mul(base_layer.weight.data, ia3_scaling)
+                    base_layer.weight.data = torch.mul(base_layer.weight.data, ia3_scaling).to(orig_dtype)
 
                 if not self.is_feedforward and (base_layer.bias is not None):
                     scaling = self.ia3_l[active_adapter].reshape(base_layer.bias.shape)
-                    base_layer.bias.data = torch.mul(base_layer.bias.data, scaling.data)
+                    base_layer.bias.data = torch.mul(base_layer.bias.data, scaling.data).to(orig_dtype)
 
                 self.merged_adapters.append(active_adapter)
 
@@ -269,15 +270,17 @@ class _ConvNd(nn.Module, IA3Layer):
             active_adapter = self.merged_adapters.pop()
             if active_adapter in self.ia3_l.keys():
                 base_layer = self.get_base_layer()
+                orig_dtype = base_layer.weight.data.dtype
                 # divide by (IA)^3 vector. Add tolerace to avoid division by zero
                 ia3_scaling = self.ia3_l[active_adapter].data
                 if not self.is_feedforward:
                     ia3_scaling = ia3_scaling.transpose(0, 1)
-                base_layer.weight.data = torch.div(base_layer.weight.data, ia3_scaling + 1e-8)
+                base_layer.weight.data = torch.div(base_layer.weight.data, ia3_scaling + 1e-8).to(orig_dtype)
 
                 if not self.is_feedforward and (base_layer.bias is not None):
                     scaling = self.ia3_l[active_adapter].reshape(base_layer.bias.shape)
-                    base_layer.bias.data = torch.mul(base_layer.bias.data, scaling.data)
+                    orig_dtype = base_layer.bias.data.dtype
+                    base_layer.bias.data = torch.mul(base_layer.bias.data, scaling.data).to(orig_dtype)
 
     def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
         dtype = previous_dtype = x.dtype
--- a/src/peft/helpers.py
+++ b/src/peft/helpers.py
@@ -22,6 +22,7 @@ from torch import nn
 
 from .peft_model import PeftConfig, PeftModel
 from .tuners.lora import LoraLayer
+from .tuners.tuners_utils import BaseTunerLayer
 
 
 def update_forward_signature(model: PeftModel) -> None:
@@ -218,8 +219,6 @@ def disable_input_dtype_casting(model: nn.Module, active: bool = True):
     """
     Context manager disables input dtype casting to the dtype of the weight.
 
-    Currently specifically works for LoRA.
-
     Parameters:
         model (nn.Module):
             The model containing PEFT modules whose input dtype casting is to be adjusted.
@@ -237,7 +236,7 @@ def disable_input_dtype_casting(model: nn.Module, active: bool = True):
 
     original_values = {}
     for name, module in model.named_modules():
-        if not isinstance(module, LoraLayer):
+        if not isinstance(module, BaseTunerLayer):
             continue
         original_values[name] = module.cast_input_dtype_enabled
         module.cast_input_dtype_enabled = False
@@ -246,7 +245,7 @@ def disable_input_dtype_casting(model: nn.Module, active: bool = True):
         yield
     finally:
         for name, module in model.named_modules():
-            if not isinstance(module, LoraLayer):
+            if not isinstance(module, BaseTunerLayer):
                 continue
             if name in original_values:
                 module.cast_input_dtype_enabled = original_values[name]
--- a/src/peft/tuners/boft/layer.py
+++ b/src/peft/tuners/boft/layer.py
@@ -220,6 +220,8 @@ class BOFTLayer(BaseTunerLayer):
         # Mark the weight as unmerged
         self._disable_adapters = False
         self.merged_adapters = []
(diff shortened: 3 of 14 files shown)


BenjaminBossan:
Pinging the contributors of the different methods if they could be so kind as to review the changes that affect them:

- BOFT: @yfeng95 @zqiu24 @YuliangXiu
- Bone: @JL-er
- FourierFT: @Phoveran
- HRA: @DaShenZi721 
- IA³: @SumanthRH
- LN Tuning: @DTennant 
- LoHa and LoKr: @kovalexal @yaswanth19
- DoRA: @nbasyl
- OFT: @okotaku @zqiu24
- VeRA: @dkopi @vvvm23

If I forgot anyone, please feel free to ping them.

If there is general agreement, my plan is to merge this shortly after the next PEFT release (v0.15.0), so aiming for next week. That way, if this does introduce regressions, we have more time to fix them.

yaswanth19:
@BenjaminBossan AFAIK the changes seems to okay for LoKr :hugs:. Also is precision loss the reason to convert the inputs to float32 rather than downcasting the adapter weights :thinking:

nbasyl:
Hi @BenjaminBossan, I recently found an issue with DoRA training where the magnitude vector isn’t being updated when training on A100, although it works fine on H100. Could this be related to the casting issue? Also could you point me to any test cases to verify this pull request? Additionally, I recently ran into a problem using DoRA with DeepSpeed ZeRO-3. I’m not sure if this is related as well, but I’ll open an issue soon with steps to reproduce the error. Thanks a lot!

BenjaminBossan:
> AFAIK the changes seems to okay for LoKr

Thanks for checking.

> Also is precision loss the reason to convert the inputs to float32 rather than downcasting the adapter weights

Exactly, although I'm not sure how much of a difference it really makes. It's just been like that for LoRA from the start and I think it makes sense to consolidate the other methods.

> I recently found an issue with DoRA training where the magnitude vector isn’t being updated when training on A100, although it works fine on H100

I haven't worked with those GPUs, so I can't say if there could be a relationship.

> The torchtune team also found a similar issue

Is it this one: https://github.com/pytorch/torchtune/issues/2250? If yes, and if it's indeed an underflow issue, then this PR could indeed fix it.

> Also could you point me to any test cases to verify this pull request?

The tests I added are here: https://github.com/huggingface/peft/pull/2433/files#diff-8e23036752ec7b8a68889069e72c29171c53652df5e7d390afa4d223f22c77d2

You should be able to check out my branch and then call

`pytest tests/test_custom_models.py -k "(test_forward_bfloat16 or test_forward_float16) and dora" -v`

to check them. Note, however, that those are very simple tests, no multi GPU etc.

> Additionally, I recently ran into a problem using DoRA with DeepSpeed ZeRO-3. I’m not sure if this is related as well, but I’ll open an issue soon with steps to reproduce the error.

Okay, feel free to ping me once the issue is done.

nbasyl:
Hi @BenjaminBossan, I have ran the test on DoRA using the test file you wrote on H100, and the following cases did not pass: 
image

I also wanted to verify if the magnitude vector is being updated correctly, so I tried modifying the function def test_training_works:, 
image
I printed out the before and after lora_A weight and got NaN 
image. 

Am I missing something when running pytest tests/test_custom_models.py -k "test_training_works" -v -s, or could you help me write a simple test function to check whether the magnitude vector is being updated correctly? Thank you!

BenjaminBossan:
> I have ran the test on DoRA using the test file you wrote on H100, and the following cases did not pass:

Hmm, strange, does the error only occur on H100 or was that the only device you tested? Could you please also show the full error message?

> I also wanted to verify if the magnitude vector is being updated correctly, so I tried modifying the function def test_training_works:,

Please note that this a very specific test inside the `TestDynamicDispatch` class, it should not be used to test DoRA. I checked this test and it is indeed faulty (lr is too high, leading to nan), I will fix it in a separate PR. But this is totally independent of DoRA.

> could you help me write a simple test function to check whether the magnitude vector is being updated correctly?

To check this, I think the test here and/or the tests below it are best suited:

https://github.com/huggingface/peft/blob/e5e7b73fcf7f0d9fc370adbb016390454ef6eb09/tests/test_gpu_examples.py#L855

You can run it with:

`pytest tests/test_gpu_examples.py -k test_causal_lm_training_4bit_dora`

Just to do a quick check, I added these lines of code:

```python
...
mv_before = model.base_model.model.model.decoder.layers[0].self_attn.v_proj.lora_magnitude_vector["default"].weight.data.clone()

...

trainer.train()
mv_after = model.base_model.model.model.decoder.layers[0].self_attn.v_proj.lora_magnitude_vector["default"].weight.data
assert not torch.allclose(mv_before, mv_after, atol=1e-5, rtol=1e-5)
```

WDYT?

src/peft/tuners/ia3/layer.py
SumanthRH, line 1: @BenjaminBossan The changes for (IA)^3 seem fine!

Qubitium:
@BenjaminBossan I have a question regarding Lora. 

After PR:

Model is bfloat16 and lora is stored as bfloat16. Peft would now upcast, by default the x input to float32 and the lora to float32 correct?

Before PR:

Peft did not do any up/down casting and used as is, by default. 

Is above correct summation?

Also, can you do some prelim tests to see what the performance costs are for the autocast to float32 and lora adapter apply if any? If there is no underflow issue for some cases of adapter, it would be good to let user know in code/doc that autocast resolves the accuracy issue but at XX cost so users a little bit more informed.,

BenjaminBossan:
> After PR:
> 
> Model is bfloat16 and lora is stored as bfloat16. Peft would now upcast, by default the x input to float32 and the lora to float32 correct?
> 
> Before PR:
> 
> Peft did not do any up/down casting and used as is, by default.
> 
> Is above correct summation?

No, for LoRA there is no change in this PR (a few small changes around merging and MHA, but otherwise no change). The situation for LoRA has always been: By default, the LoRA weights are loaded in float32. If the input is lower precision, it is upcast to float32 when the LoRA part is calculated during `forward`, and after this downcast to the previous dtype. See these parts of the current code:

https://github.com/huggingface/peft/blob/e5e7b73fcf7f0d9fc370adbb016390454ef6eb09/src/peft/tuners/tuners_utils.py#L321

https://github.com/huggingface/peft/blob/e5e7b73fcf7f0d9fc370adbb016390454ef6eb09/src/peft/tuners/lora/layer.py#L724-L744

This PR just makes it so that this behavior is consistent between the different layer types and PEFT methods. The reason why I worked on this is because I found many PEFT methods break when loading the model in lower dtype.

> Also, can you do some prelim tests to see what the performance costs are for the autocast to float32 and lora adapter apply if any? If there is no underflow issue for some cases of adapter, it would be good to let user know in code/doc that autocast resolves the accuracy issue but at XX cost so users a little bit more informed.,

I did a very quick experiment with `meta-llama/Llama-3.2-3B` (unquantized, bfloat16) on a 4090 with LoRA rank 32 for 100 steps. What I got:

|  | LoRA float32 | LoRA bfloat16 |
| -- | -- | -- |
| train time | 11.7 s | 11.2 s |
| eval time | 17.4 s | 17.0 s |
| train loss | 1.1884 | 1.2519 |
| memory max | 18424MB | 17198MB |
| memory reserved avg | 11340MB | 10831MB |
| memory reserved 99th perc. | 17782MB | 16617MB |
| checkpoint file size | 35.0MB | 17.5MB |

So there are small but noticeable differences, with lower LoRA precision resulting in slightly faster training, less memory, and smaller file size, at the cost of slightly higher loss. Of course, results will vary a lot depending on the use case, and longer/bigger training runs could be more prone to under/overflow.

At the end of the day, users can easily change the behavior by passing `autocast_adapter_dtype=False`. Given that LoRA is the most commonly used PEFT method by far, I think it is better to make the LoRA behavior the default for the other PEFT methods than vice versa.

Regarding the documentation, yes, we can certainly highlight this option better. I'll check if I can find a good place in the docs to put this.

JL-er:
@BenjaminBossan  The structure of Bone is very simple, so it shouldn't have much impact. Do I only need to test it with `pytest tests/ -k "bone"`?

BenjaminBossan:
> The structure of Bone is very simple, so it shouldn't have much impact. Do I only need to test it with `pytest tests/ -k "bone"`?

No need for you to test, just a quick check if the code changes for Bone look right to you. We already run the tests and I also did a full training run with Bone and a bfloat16 base model and it worked.

DTennant:
Hi,

The changes for LN Tuning seems fine by me.
Thanks.

githubnemo (approved):
LGTM, minor nitpicks

BenjaminBossan:
Thanks everyone for your feedback. The PR is now merged, but the next PEFT release will not be very soon. So if you happen to find an issue, please report it so that we can fix it before the next release.

PR #2443, opened 2025-03-20
[WIP] Refactor to better support LoRA variants

BenjaminBossan:


src/peft/tuners/lora/variants.py
```diff
@@ -0,0 +1,291 @@
...
+        if module.lora_magnitude_vector is None:
+            # first dora layer being added, add lora_magnitude_vector to the list of learnable parameters
+            module.adapter_layer_names = module.adapter_layer_names[:] + ("lora_magnitude_vector",)
+
+        dora_layer = DoraConv2dLayer(fan_in_fan_out=False)
+        lora_A = module.lora_A[adapter_name].weight
+        lora_B = module.lora_B[adapter_name].weight
+        scaling = module.scaling[adapter_name]
+        dora_layer.update_layer(base_layer=module.get_base_layer(), lora_A=lora_A, lora_B=lora_B, scaling=scaling)
+        module.lora_magnitude_vector[adapter_name] = dora_layer
```
githubnemo, line 276: this could probably be refactored using a factory for `DoraConv*dLayer` that gets called in `_DoraConvNdVariant.init` to safe up on some repetition.

githubnemo (commented):
This already looks quite good. Some comments.

src/peft/tuners/lora/layer.py
```diff
@@ -57,13 +95,14 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
         # Mark the weight as unmerged
         self._disable_adapters = False
         self.merged_adapters = []
-        self.use_dora: dict[str, bool] = {}
+        self.use_dora: dict[str, bool] = {}  # not actively used anymore after #2443, keep it for BC
```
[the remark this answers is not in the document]
BenjaminBossan: I would generally refrain from making backwards incompatible changes even with a deprecation period unless there is a big gain. Too many users/package authors miss the deprecation messages and when something breaks they just fix a previous PEFT version.

src/peft/tuners/lora/layer.py
```diff
@@ -116,6 +155,21 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
         self.in_features = in_features
         self.out_features = out_features
 
+    def resolve_lora_variant(self, *, use_dora: bool, **kwargs) -> Optional[LoraVariant]:
+        """
+        Return any possible LoRA variant for this layer type.
```
githubnemo, line 160:
```suggestion
        Return a matching LoRA variant for this layer type.
```

To me this suggesated multiple and/or arbitrary LoRA variants.

src/peft/tuners/lora/layer.py
```diff
@@ -116,6 +155,21 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
...
+
+        Given the init arguments of this layer, return the correct LoRA variant, if any. E.g., if `use_dora=True`, this
+        method should return the DoRA variant for the given layer.
+
+        If there is no fitting variant, return None.
+
+        Note: If this layer type does not support the LoRA variant at all, please raise an error during __init__ and
+        not here.
```
githubnemo, line 168: Why? Please explain the benefit in the doc string.

githubnemo (commented):
Nice :) Two nitpicks and then we can merge this.

githubnemo (approved):

src/peft/tuners/lora/layer.py
gapsong, line 1: Why not just remove `self.use_dora` entirely?

```python
self.use_dora: dict[str, bool] = {}  # not actively used anymore after #2443, keep it for BC
```

If it's no longer used and backward compatibility isn't a concern, it can be safely removed to clean up the codebase.
If BC is still important, consider issuing a deprecation warning instead to make the transition clearer.

githubnemo: IIRC X-LoRA is currenlty looking at the `use_dora` attribute of a layer to raise a warning. There might be other, 3rd party implementations that do something similar which would break. Raising a warning for a dict access is probably adding more code than what is worth IMO.

PR #2444, opened 2025-03-20
Fix incorrect link in docs

kenning:
The broken link can be found [here](https://huggingface.co/docs/peft/accelerate/deepspeed#use-peft-qlora-and-deepspeed-with-zero3-for-finetuning-large-models-on-multiple-gpus) in the docs.


BenjaminBossan (approved):
Thanks for fixing the link.

Issue #2450, opened 2025-03-25
modules_to_save resulting in empty tensor with deepspeed zero3 LoRA training


agokrani:
### System Info

peft==0.15.0
transformers==4.49.0

### Who can help?

@BenjaminBossan I am training a model with LoRA using DeepSpeed zero3, and I have set modules_to_save=["embed_tokens", "lm_head"]. The model trained successfully without any errors. However, while merging the model, I am encountering the following error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/agokrani/Documents/git/zett/env-zett-311/lib/python3.11/site-packages/transformers/integrations/peft.py", line 239, in load_adapter
    incompatible_keys = set_peft_model_state_dict(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/agokrani/Documents/git/zett/env-zett-311/lib/python3.11/site-packages/peft/utils/save_and_load.py", line 444, in set_peft_model_state_dict
    load_result = model.load_state_dict(peft_model_state_dict, strict=False)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/agokrani/Documents/git/zett/env-zett-311/lib/python3.11/site-packages/torch/nn/modules/module.py", line 2581, in load_state_dict
    raise RuntimeError(
RuntimeError: Error(s) in loading state_dict for Phi3ForCausalLM:
	size mismatch for model.embed_tokens.modules_to_save.default.weight: copying a param with shape torch.Size([0]) from checkpoint, the shape in current model is torch.Size([151669, 3072]).
```

Upon inspecting the adapter_model.safetensors file, I noticed that both embed_tokens and lm_head tensors were saved, but they are empty tensors. I have debugged the root cause, and it seems that the following code was removed from the get_peft_model_state_dict function:

```
if getattr(model, "modules_to_save", None) is not None:
    for key, value in state_dict.items():
        if any(f"{module_name}.modules_to_save.{adapter_name}" in key for module_name in model.modules_to_save):
            to_return[key.replace("modules_to_save.", "")] = value
```

Instead, this code was added:

```
# ADDITIONAL TRAINING MODULES / MODULES_TO_SAVE
for name, module in model.named_modules():
    if isinstance(module, AuxiliaryTrainingWrapper):
[...]

PR #2451, opened 2025-03-25
FIX: modules_to_save resulting in empty tensor with deepspeed

agokrani:
see: https://github.com/huggingface/peft/issues/2450


githubnemo:
Closing as discussed in #2456

PR #2456, opened 2025-03-26
Fix #2450: Revamp adapter_state_dict_* methods

### LINKED ISSUE #2450: modules_to_save resulting in empty tensor with deepspeed zero3 LoRA training

githubnemo:
`AuxiliaryTrainingWrapper.adapter_state_dict` now utilizes an external state dict for the computation of the module state dict to avoid problems with DeepSpeed (or FSDP) when dealing with distributed parameters.

It is not possible to simply wrap everything in `GatheredParameters` context managers since doing that leads to a deadlock when running on more than one process (reasons unclear). Since transformers, or more specifically, accelerate already handles state dict fetching for the whole model, it is more economical to use that state dict and rewrite the methods that before depended on `state_dict()` calls.

### DIFF
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -201,7 +201,9 @@ def get_peft_model_state_dict(
             # keys that the module thinks need to be saved. We cannot rely on `.state_dict()` internally of the
             # module since accelerators like DeepSpeed require special handling which is done for the model
             # state dict from above but most likely not in the module itself. See #2450.
-            module_state_dict = {k[len(f"{name}.") :]: v for k, v in state_dict.items() if k.startswith(f"{name}.")}
+            module_state_dict = {
+                k.removeprefix(f"{name}."): v for k, v in state_dict.items() if k.startswith(f"{name}.")
+            }
             to_return.update(
                 {f"{name}.{k}": v for k, v in module.adapter_state_dict(adapter_name, module_state_dict).items()}
             )
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -197,7 +197,14 @@ def get_peft_model_state_dict(
     # ADDITIONAL TRAINING MODULES / MODULES_TO_SAVE
     for name, module in model.named_modules():
         if isinstance(module, AuxiliaryTrainingWrapper):
-            to_return.update({f"{name}.{k}": v for k, v in module.adapter_state_dict(adapter_name).items()})
+            # Compute the module-relative state dict to make it easier for the adapter to fetch the appropriate
+            # keys that the module thinks need to be saved. We cannot rely on `.state_dict()` internally of the
+            # module since accelerators like DeepSpeed require special handling which is done for the model
+            # state dict from above but most likely not in the module itself. See #2450.
+            module_state_dict = {k[len(f"{name}.") :]: v for k, v in state_dict.items() if k.startswith(f"{name}.")}
+            to_return.update(
+                {f"{name}.{k}": v for k, v in module.adapter_state_dict(adapter_name, module_state_dict).items()}
+            )
 
     # DEAL WITH EMBEDDINGS
     # check the common embedding layers in `target_modules` to reset `save_embedding_layers` if necessary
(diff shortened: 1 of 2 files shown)


src/peft/utils/save_and_load.py
```diff
@@ -197,7 +197,14 @@ def renamed_dora_weights(k):
     # ADDITIONAL TRAINING MODULES / MODULES_TO_SAVE
     for name, module in model.named_modules():
         if isinstance(module, AuxiliaryTrainingWrapper):
-            to_return.update({f"{name}.{k}": v for k, v in module.adapter_state_dict(adapter_name).items()})
+            # Compute the module-relative state dict to make it easier for the adapter to fetch the appropriate
+            # keys that the module thinks need to be saved. We cannot rely on `.state_dict()` internally of the
+            # module since accelerators like DeepSpeed require special handling which is done for the model
+            # state dict from above but most likely not in the module itself. See #2450.
+            module_state_dict = {k[len(f"{name}.") :]: v for k, v in state_dict.items() if k.startswith(f"{name}.")}
```
BenjaminBossan, line 204: Since we're Python 3.9+, would [`k.removeprefix`](https://docs.python.org/3/library/stdtypes.html#str.removeprefix)  work?

BenjaminBossan (approved):
Thanks for fixing this, I could confirm that the sft DeepSpeed example passes on my end with these changes. The solution is also more elegant than what I proposed, so I closed my PR.

@agokrani, if you have the opportunity, could you please test if this branch fixes your initial issue?

OutSightAI:
@BenjaminBossan Give me couple of hours and will get back to you.

OutSightAI:
@BenjaminBossan @githubnemo I can confirm that this solves my original problem. The solution is better than my PR, so I will close my PR as well. Thanks for the help

githubnemo:
@OutSightAI Thanks for testing :)

PR #2459, opened 2025-03-27
Prepare 0.15.1 patch

githubnemo:
This release is a patch release to release a fix for #2450 which might result in loss of `modules_to_save` when trained with deepspeed ZeRO stage 3.

Only the squashed commit from #2456 and the version bump are included in this PR.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@
 from setuptools import find_packages, setup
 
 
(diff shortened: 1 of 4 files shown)


BenjaminBossan (approved):
Thanks, LGTM.

Issue #2441, opened 2025-03-19
Integarting RandLoRA a full rank PEFT algorithm


PaulAlbert31:
### Feature request

Include RandLoRA as an option for PEFT
Code: https://github.com/PaulAlbert31/RandLoRA/tree/main/peft/src/peft/tuners/randlora
Paper:  https://openreview.net/pdf?id=Hn5eoTunHN

### Motivation

RandLoRA is a recently published paper https://openreview.net/forum?id=Hn5eoTunHN that challenges the limitations of LoRA's low-rank setting. RandLoRA combines random bases to create a full rank parameter-efficient updates. An custom gradient function is proposed to optimize memory use.

In summary:
RandLoRA can outperform LoRA for complex tasks (e.g. vision-language or LLM finetuning)
RandLoRA reduces memory usage over LoRA thanks to custom gradient functions that share non-trainable random bases across layers
RandLoRA's parameter count is dictated by the rank of the random basis.
RandLoRA can take slightly longer to train when scaling up the amount of trainable parameters.

To illustrate RandLoRA's capabilities, here are key results from the paper:

![Image](https://github.com/user-attachments/assets/88a4085d-776b-498f-88c4-62b62e5468ba)

CLIP results (vision-language classification) with GPU VRAM usage during training:

![Image](https://github.com/user-attachments/assets/df294ed2-d044-4cf4-9791-681d75c9d67c)

Loss landscape connectivity of LoRA vs RandLoRA vs Std. Finetuning for CLIP. RandLoRA achieves a deeper minima than LoRA for equal trainable parameters.

![Image](https://github.com/user-attachments/assets/ad5e3ed9-a9bf-4d0e-8951-789ff7aab723)

### Your contribution

I have implemented RandLoRA as part of a local copy of the peft package https://github.com/PaulAlbert31/RandLoRA
More specifically: https://github.com/PaulAlbert31/RandLoRA/tree/main/peft/src/peft/tuners/randlora

I am happy to help or refactor part of the existing code to fit the current PEFT structure.

githubnemo:
Hey, thanks for the suggestion, at the first glance this looks promising.

We're currently in the process of deciding how to implement such variations, starting with [SineLoRA](https://github.com/huggingface/peft/issues/2434#issuecomment-2740943926). We'll get back to you, hopefully quite soon, on how to implement this and/or with follow-up questions.

BenjaminBossan:
[...]

PR #2464, opened 2025-03-28
Feature: RandLora integration into peft

PaulAlbert31:
Follow up to issue 2441: https://github.com/huggingface/peft/issues/2441
This is an implementation of RandLoRA into peft, including some support for quantized training.

Some work is left to do on proper BnB configurations, the current implementation still works on --load_4bit configurations. Would need some guidance on this.

The 4bit and 8bit layer classes are however implemented.

Similarly to Vera, RandLoRA uses random bases which are shared between layers. These bases are saved into BufferDict objects but I had trouble using safetensors to save them. This may be due to a deprecated transformers version (4.46.3) during testing.

Happy to iterate as per feedback.


BenjaminBossan (changes requested):
Thanks for the PR. I've only skimmed it so far but it looks already quite good, well done (good job using the new method registration).

I think the best next step would be to add RandLoRA to the unit testing harness. This will allow us to easily see what already works and what doesn't. For this, check for instance how VeRA was added here:

https://github.com/huggingface/peft/blob/2ee02af9d445cda0410c43a6e72160df59ad971c/tests/test_custom_models.py#L422-L440

Let's add similar entries for RandLoRA. If we then find that some tests are failing, I'll guide you through the steps needed to make them pass.

Later, we'll probably also have to add some dedicated RandLoRA tests to ensure that the weight sharing works (see https://github.com/huggingface/peft/blob/main/tests/test_vera.py), maybe also something for the special grad handling. But it's fine to leave this for later.

> Some work is left to do on proper BnB configurations, the current implementation still works on --load_4bit configurations. Would need some guidance on this.

I'm not quite sure what you mean by that, could you explain further? But we can also leave bnb integration for later, it might be better to get the normal RandLoRA fully working first.

> but I had trouble using safetensors to save them. This may be due to a deprecated transformers version (4.46.3) during testing.

Could you give a small snippet to reproduce this?

Below, I'll list a few more steps required before merging the PR. This is just to get an overview and can be left for the later, for now let's focus on the points mentioned above.

- [ ] Documentation
- [ ] Example script/notebook
- [x] Full test integration
- [ ] Optional: Add an experiment to be able to [compare it to other PEFT methods](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA)

BenjaminBossan:
@PaulAlbert31 Could you please run `make style`, otherwise the CI won't run the tests. Also, please always ping me once your changes are ready.

PaulAlbert31:
Hey @BenjaminBossan , apologies for the previous update I am still learning my way around these tools.
I have updated the draft after running make style and fixing errors. 

Regarding the safetensor error I was mentioning earlier, this seems to go away after upgrading transformers so nothing to worry about there. This happens when saving the model after training so I guess this would be covered in the existing tests. Here is a small snippet I used to test.
```python 
import torch
from peft import ( 
    RandLoraConfig,
    get_peft_model
)
import transformers
from transformers import AutoModelForCausalLM, AutoTokenizer, TrainingArguments, Trainer, TextDataset, DataCollatorForLanguageModeling

config = RandLoraConfig(
            r=16,
            randlora_alpha=32,
            randlora_dropout=0.05,
            target_modules=["k_proj"],
            bias="none",
            task_type="CAUSAL_LM",
            projection_prng_key=int(torch.exp(torch.tensor(1))*3.1415*1000),
)

            "Qwen/Qwen2.5-0.5B",
            torch_dtype=torch.float16,
            trust_remote_code=True,
)

tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen2.5-0.5B", trust_remote_code=True)
tokenizer.pad_token = tokenizer.eos_token # Set pad token if it's missing

dummy_text = "This is a dummy text for training.\nAnother line of dummy text.\nAnd a third line."
dummy_data_path = "dummy_data.txt" 
with open(dummy_data_path, "w") as f:
    f.write(dummy_text)

train_dataset = TextDataset(
    tokenizer=tokenizer,
    file_path=dummy_data_path,
    block_size=1
)

data_collator = DataCollatorForLanguageModeling(
    tokenizer=tokenizer,
    mlm=False
)

    output_dir="./output_trainer",
    num_train_epochs=1,
    per_device_train_batch_size=1,
    optim="adamw_torch", 
    save_steps=1000,
    logging_steps=10, 
    learning_rate=1e-4,
    weight_decay=0.01,
    fp16=True, 
    bf16=False,
    max_steps=-1, 
    logging_dir="./logs", 
    dataloader_num_workers=0, 
    save_total_limit=1, 
    save_safetensors=True,
)

trainer = Trainer(
    model=model,
    args=training_args,
    train_dataset=train_dataset,
    data_collator=data_collator,
    tokenizer=tokenizer,
)

trainer.train()
model.save_pretrained("test_randlora")
```

Regarding quantized training, I meant that I use the load_4_bit argument to the AutoModelForCausalLM.from_pretrained to quantize the base model [here in my code](https://github.com/PaulAlbert31/RandLoRA/blob/e3f0e5c4809416a032edbbc34cc4b7cdbaf8468d/commonsense_reasoning/finetune.py#L182C1-L189C10). My understanding is that a [BnbQuantizationConfig](https://huggingface.co/docs/accelerate/v1.5.2/en/package_reference/utilities#accelerate.utils.BnbQuantizationConfig) is the updated way of doing it. 

Since this config interacts with the base model and not the adapter, I guess this is not important for peft ?

Paul

src/peft/tuners/randlora/config.py
```diff
@@ -0,0 +1,191 @@
...
+        save_projection (`bool`):
+            Whether to save the global basis_A / basis_B random basis in the state dict alongside per layer lambda /
+            gamma diagonal matrices. weights. This will increase the size of the checkpoint, but guarantee that we can
+            reload the checkpoint on all system configurations. Defaults to `True`.
+        sparse (`bool`):
+            Whether to use sparse random bases as described in the RandLora paper. The current implementation is a
+            proof of concept where the sparseness is not used to improve speed or memory usage. Defaults to `False`.
+        very_sparse (`bool`):
+            Whether to use very sparse random bases. The current implementation is a proof of concept where the
+            sparseness is not used to improve speed or memory usage. Defaults to `False`.
```
BenjaminBossan, line 49: As a user, should I never use sparseness?

src/peft/tuners/randlora/layer.py
```diff
@@ -0,0 +1,340 @@
...
+        self.r[adapter_name] = r
+        if randlora_dropout > 0.0:
+            randlora_dropout_layer = nn.Dropout(p=randlora_dropout)
+        else:
+            randlora_dropout_layer = nn.Identity()
+
+        self.randlora_dropout.update(nn.ModuleDict({adapter_name: randlora_dropout_layer}))
+
+        # Actual trainable parameters
+        n = min(self.in_features, self.out_features) / r
```
BenjaminBossan, line 112: Can we use a more descriptive name than `n` please?

src/peft/tuners/randlora/layer.py
```diff
@@ -0,0 +1,340 @@
...
+        self.n = int(n) if n.is_integer() else int(n) + 1  # Full rank
+        self.randlora_lambda[adapter_name] = nn.Parameter(torch.zeros(r, self.n), requires_grad=True)
+        self.randlora_gamma[adapter_name] = nn.Parameter(
+            torch.ones(self.n, min(self.out_features, self.in_features)) / max(self.out_features, self.in_features),
+            requires_grad=True,
+        )
+
+        self.scaling[adapter_name] = randlora_alpha / r / math.sqrt(self.n)  # * 10#/ math.sqrt(self.n)
```
BenjaminBossan, line 120: What about this comment?

src/peft/tuners/randlora/layer.py
```diff
@@ -0,0 +1,340 @@
...
+            self.reset_randlora_parameters(adapter_name)
+
+        self._move_adapter_to_device_of_base_layer(adapter_name)
+        self.set_adapter(self.active_adapters)
+
+    def reset_randlora_parameters(self, adapter_name):
+        if adapter_name in self.randlora_lambda.keys():
+            with torch.no_grad():
+                nn.init.zeros_(self.randlora_lambda[adapter_name])
+                nn.init.zeros_(self.randlora_gamma[adapter_name]).fill_(
```
BenjaminBossan, line 169: Is the `zeros_` required if it is followed by a `fill_`?

src/peft/tuners/randlora/layer.py
```diff
@@ -0,0 +1,340 @@
...
+            # no adapter to merge
+            return
+
+        for active_adapter in adapter_names:
+            if active_adapter in self.randlora_lambda.keys():
+                base_layer = self.get_base_layer()
+                if safe_merge:
+                    # Note that safe_merge will be slower than the normal merge
+                    # because of the copy operation.
+                    orig_weights = base_layer.weight.data.clone()
```
BenjaminBossan, line 222: We're missing `+ self.get_delta_weight(active_adapter)` here, right?

src/peft/tuners/randlora/layer.py
```diff
@@ -0,0 +1,340 @@
...
+        update = (update_B.T @ update_A.T).T
+        output_tensor = transpose(update, self.fan_in_fan_out)
+
+        if dtype != self.randlora_B[adapter].dtype:
+            output_tensor = output_tensor.to(dtype=dtype)
+
+            # cast back the weights
+            # TODO: why?, taken from the VeRA implementation
+            self.randlora_lambda[adapter].data = self.randlora_lambda[adapter].to(dtype)
+            self.randlora_gamma[adapter].data = self.randlora_gamma[adapter].to(dtype)
```
BenjaminBossan, line 310: This can indeed be removed.

BenjaminBossan (changes requested):
Thanks for the updates, the PR is making good progress.

A lot of the added tests are still failing. You can run them locally with:

`pytest tests/test_custom_models.py -k randlora -v`

The error most of the time is something like:

`RuntimeError: The size of tensor a (20) must match the size of tensor b (10) at non-singleton dimension 1`

I'm fairly certain that is due to how the slices are determined in `get_scaled_bases` but I'm not sure what the correct fix is. Could you please investigate?

> Regarding the safetensor error I was mentioning earlier, this seems to go away after upgrading transformers so nothing to worry about there.

Nice. I tested your snippet and it passes for me too.

> Regarding quantized training, I meant that I use the load_4_bit argument to the AutoModelForCausalLM.from_pretrained

Ah I see, no worries about that. In PEFT, we use the proper arguments. We will later just copy one of the existing PEFT bnb tests and make adjustments for RandLoRA. But as mentioned above, let's first get the non-quantized code fully running.

BenjaminBossan:

src/peft/tuners/randlora/layer.py
```diff
@@ -0,0 +1,340 @@
...
+            error_tmpl = (
+                "{} has a size of {} but {} or greater is required; this probably happened because an additional RandLora "
+                "adapter was added after the first one with incompatible shapes."
+            )
+            # check input size
+            if randlora_A_param.shape[-1] < self.in_features:
+                raise ValueError(error_tmpl.format("randlora_A", randlora_A_param.shape[1], self.in_features))
+            # check output size
+            if randlora_B_param.shape[0] < self.out_features:
+                raise ValueError(error_tmpl.format("randlora_B", randlora_B_param.shape[0], self.out_features))
```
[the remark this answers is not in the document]
PaulAlbert31: updated_A will always be used to obtain the smallest dimension of the weight matrix to reduce trainable parameters (randlora_gamma is applied to the A random base and contains most of the trainable parameters).
The size selection for randlora_A and randlora_B was wrong so I corrected and added a comment here.

src/peft/tuners/randlora/layer.py
```diff
@@ -0,0 +1,340 @@
...
+        # As adapted layers may have different shapes and RandLora contains a single shared pair of A and B matrices,
+        # we initialize these matrices with the largest required size for each dimension.
+        # During the forward pass, required submatrices are sliced out from the shared randlora_A and randlora_B.
+        sliced_A = randlora_A[:, : self.n, :min_dim]
+        sliced_B = randlora_B[:max_dim, : self.n, :]
+
+        # Flattening the matrices over the rank and number of bases dimensions is more memory efficient
+        update_B = sliced_B.flatten(start_dim=1)
+        update_A = UniqueBaseGrad.apply(sliced_A, randlora_lambda, randlora_gamma).flatten(end_dim=1)
+        if min_dim == self.in_features:
```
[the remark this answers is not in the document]
PaulAlbert31: This is related to a previous comment where updated_A will always be used to obtain the smallest dimension of the weight matrix to reduce trainable parameters.

I have added a comment to explain

tests/test_custom_models.py
```diff
@@ -3578,6 +3607,148 @@ def forward(self, X):
             "base_model.model.lin2.vera_lambda_d.adapter1",
         )
 
+    def test_requires_grad_randlora_different_targets(self):
```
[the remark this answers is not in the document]
PaulAlbert31: Yes good catch, thanks

tests/test_custom_models.py
```diff
@@ -499,6 +500,25 @@
...
+        "MLP",
+        RandLoraConfig,
+        {"target_modules": ["lin0"], "modules_to_save": ["lin1"]},
+    ),
+    (
+        "Embedding + transformers Conv1D 1 RandLora",
+        "EmbConv1D",
+        RandLoraConfig,
+        {"target_modules": ["conv1d"]},
+    ),
```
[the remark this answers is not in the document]
PaulAlbert31: Removed

tests/test_custom_models.py
```diff
@@ -499,6 +500,25 @@
         TrainableTokensConfig,
         {"target_modules": ["emb"], "token_indices": [0, 1, 3], "init_weights": False},
     ),
+    ########
+    # RandLora #
+    ########
+    ("Vanilla MLP 1 RandLora", "MLP", RandLoraConfig, {"target_modules": "lin0"}),
+    ("Vanilla MLP 2 RandLora", "MLP", RandLoraConfig, {"target_modules": ["lin0"]}),
+    ("Vanilla MLP 3 RandLora", "MLP", RandLoraConfig, {"target_modules": ["lin1"]}),
+    ("Vanilla MLP 4 RandLora", "MLP", RandLoraConfig, {"target_modules": ["lin0", "lin1"]}),
```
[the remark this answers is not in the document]
PaulAlbert31: Yes, this is now added

src/peft/tuners/randlora/config.py
PaulAlbert31: It could be beneficial for other reason such as reducing overfitting.

The sparse bases ternary bases could in theory be used to compute matmul free updates via value aggregation, similarly to this paper: https://arxiv.org/abs/2406.02528 . I don't believe this is currently supported by CUDA though. 
Do you think is this too experimental to include ?

src/peft/tuners/randlora/layer.py
PaulAlbert31: Changed to n -> num_bases

src/peft/tuners/randlora/layer.py
PaulAlbert31: Good catch, thanks

src/peft/tuners/randlora/layer.py
PaulAlbert31: This is what is done in the vera implementation. I agree this is not very efficient.
Do you have another suggestion ?

PaulAlbert31:
Hey @BenjaminBossan, thanks a lot for all your comments and catching some errors. 
I have now updated the code according to your comments and all tests are passing.
I have also added tests for the sparse and very sparse bases.

Let me know if more changes are needed and we can move to more specific tests or to documentation/example scripts

examples/fp4_finetuning/finetune_fp4_opt_bnb_peft.py
```diff
@@ -41,7 +41,7 @@
 max_memory = f"{free_in_GB - 2}GB"
 
 n_gpus = torch.cuda.device_count()
-max_memory = {i: max_memory for i in range(n_gpus)}
+max_memory = dict.fromkeys(range(n_gpus), max_memory)
```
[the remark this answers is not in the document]
PaulAlbert31: Oh I missed these, I'll update my commits

src/peft/tuners/randlora/config.py
BenjaminBossan: Since you've already done the work, I'm fine with keeping this. We should just clarify to the user what they should do with these parameters. If you add the info from this comment to the docstrings, it would help.

BenjaminBossan:
@PaulAlbert31 You may have seen the merge conflicts in `test_decoder_models.py` and `test_encoder_decoder_models.py`. This is because we refactored the test files to get rid of unittest and only use pytest instead. Overall, this should simplify the test structure. Don't hesitate to ping me if you have questions when resolving the conflict.

PaulAlbert31:
Hey @BenjaminBossan 
This new update runs the gpu tests correctly, I have also updated the default hyper-parameters of the RandLoRA config to be more comparable to a rank 32 LoRA. I haven't run the benchmark yet, I'll try and do that next week to ensure a fair comparison.
I haven't rebased on the current main yet to include the new bfloat16 tests, still working on that as well.

PaulAlbert31:
Hey @BenjaminBossan 
I have now merged with main and updated hyper-parameters to have better performance on the [PEFT method comparison suite](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA).

Here are the numbers for equal trainable parameters to LoRA:
      
| Metric                         | LoRA rank 32 | RandLoRA rank 32 |
|---------------------------------|--------------|-------------------|
| cuda memory max                | 22.3 GB      | 22.8 GB           |
| cuda memory reserved avg       | 11.9 GB      | 12.7 GB           |
| cuda memory reserved 99th percentile | 17.7 GB      | 18.4 GB           |
| train loss                     | 0.607        | 0.577             |
| test accuracy                  | 47.8%        | 52.3%             |

I had to pause the LoRA training halfway so I do not have a accurate training time for it but it should be close to the results you reported. 

I am surprised that the memory usage for RandLoRA is higher than LoRA, this goes against what I saw in my paper but that could also be due to specific matrix sizes. I think that memory savings of RandLora are not as interesting on square matrices  (q in the attention) but I have to write a small script to test.

Hopefully I conducted the merge with main the right way, let me know if this is not appropriate.

PaulAlbert31:
Hey @BenjaminBossan,
I have reset and redone the rebase properly. Things should be more readable now. 

All contributions should still be here but let me know I am missing some tests. The bfloat16-float16 tests are passing as well.

Let me know if this is ok and we can move on to the next step 👍

src/peft/tuners/randlora/model.py
```diff
@@ -0,0 +1,563 @@
...
+            if module_shape != largest_shape:
+                largest_shape = tuple(max(a, b) for a, b in zip(largest_shape, module_shape))
+
+        if largest_shape is None:
+            msg = "No layers types compatible with RandLora were found. Please check `peft_config.target_modules`."
+            raise ValueError(msg)
+
+        return largest_shape
+
+    def _init_randlora_A_randlora_B_sparse(self, config: RandLoraConfig, adapter_name: str, s: int = 3) -> None:
```
BenjaminBossan, line 140: Let's avoid one letter variable names like `s` and `n`. Please provide a more expressive name.

src/peft/tuners/randlora/model.py
```diff
@@ -0,0 +1,563 @@
...
+        key_list = [key for key, _ in self.model.named_modules() if "randlora" not in key]
+        new_adapter = None
+        for key in key_list:
+            _, target, _ = _get_submodules(self.model, key)
+            if isinstance(target, RandLoraLayer):
+                target.delete_adapter(adapter_name)
+                if new_adapter is None:
+                    new_adapter = target.active_adapter[:]
+
+        self.active_adapter = new_adapter or []
```
BenjaminBossan, line 523: Could you please add the line 
`self._delete_auxiliary_adapter(adapter_name, new_active_adapters=new_adapter)`
here?

tests/test_custom_models.py
```diff
@@ -3780,7 +3816,7 @@ def forward(self, X):
         # active adapter is still "default"
         self.check_requires_grad(
             peft_model,
-            "base_model.model.lin1.vera_lambda_b.default",
+            "nobase_model.model.lin1.vera_lambda_b.default",
```
BenjaminBossan, line 3819: This change is probably not supposed to be there?

tests/test_initialization.py
```diff
@@ -2661,7 +2661,7 @@ def fn(x, *args):
         if prepare_layer_inputs_keys is None:
             prepare_layer_inputs_fn = fn
         else:
-            prepare_layer_inputs_fn = {k: fn for k in prepare_layer_inputs_keys}
+            prepare_layer_inputs_fn = dict.fromkeys(prepare_layer_inputs_keys, fn)
```
BenjaminBossan, line 2664: Please revert, this ruff rule is excluded in the `pyproject.toml` so should not be changed by ruff.

BenjaminBossan (changes requested):
Thank you very much for rebasing again, otherwise it would have been very hard for me to do another in-depth review.

Overall, this PR is in a very good state, most things look very good. I mostly added some smaller comments, then there is an issue with multi-GPU training. Finally, I think we need some tests along the line of  `test_vera.py` because of the shared variables in RandLoRA. I think those tests can just be copied with small edits and we should be fine. Test rest looks good.

I will also re-run the method comparison suite and report the results later.

BenjaminBossan:
Here are the new results I got, which are pretty much in line with what you posted above.

| metric | LoRA rank 32 | _new_ RandLoRA rank 32 |
| -- | -- | -- |
| cuda memory max | 22.3 GB | 22.8 GB |
| cuda memory reserved avg | 11.9 GB | 12.7 GB |
| cuda memory reserved 99th percentile | 17.7 GB | 18.4 GB |
| train loss | 0.607 | 0.578 |
| test accuracy | 47.8% | 51.9% |
| total time (sec) | 1847 | 1949 |

src/peft/tuners/randlora/layer.py
```diff
@@ -0,0 +1,346 @@
...
+
+        # As adapted layers may have different shapes and RandLora contains a single shared pair of A and B matrices,
+        # we initialize these matrices with the largest required size for each dimension.
+        # During the forward pass, required submatrices are sliced out from the shared randlora_A and randlora_B.
+        sliced_A = randlora_A[:, : self.num_bases, :min_dim]
+        sliced_B = randlora_B[:max_dim, : self.num_bases, :]
+
+        # Flattening the matrices over the rank and number of bases dimensions is more memory efficient
+        update_B = sliced_B.flatten(start_dim=1)
+        update_A = UniqueBaseGrad.apply(sliced_A, randlora_lambda, randlora_gamma).flatten(end_dim=1)
```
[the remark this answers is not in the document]
PaulAlbert31: Good point, thanks for catching that.
I have followed your fix. 
Not sure what to do about the OOM error. I guess moving tensors between devices exceeds the per-GPU sharding budget which causes the OOM error. 
I guess a fix would be to assume the random bases and maybe the learnable parameters should be on both GPUs in the device_map. Not sure how/how easy it is to do.

PaulAlbert31:
Hey @BenjaminBossan, great to see you were able to reproduce the results !

I have updated the pull request with your suggested changes and added a test_randlora.py test file.

As I commented above, I am not sure how to proceed with the multi-GPU OOM error, let me know if you find an easy fix.

src/peft/tuners/randlora/model.py
```diff
@@ -0,0 +1,564 @@
...
+        # deterministic init of randlora_A and randlora_B if we know the key
+        generator = torch.Generator(device="cpu").manual_seed(config.projection_prng_key)
+
+        # The gamma matrix is applied on A meaning it can be unique (shared) accross the n scaling matrices.
+        # We also set randlora_A as the smallest matrix to reduce trainable parameters.
+        randlora_A = torch.rand((config.r, 1, min_dim), generator=generator)
+
+        # Number of bases to ensure full rank
+        num_bases = min_dim / config.r
+        num_bases = int(n_bases) if num_bases.is_integer() else int(num_bases) + 1  # Ensure full rank
```
BenjaminBossan, line 161:
```suggestion
        num_bases = int(num_bases) if num_bases.is_integer() else int(num_bases) + 1  # Ensure full rank
```

tests/test_randlora.py
```diff
@@ -0,0 +1,301 @@
...
+        X = self.lin1(X)
+        X = self.relu(X)
+        X = self.lin2(X)
+        X = self.relu(X)
+        X = self.lin3(X)
+        X = self.sm(X)
+        return X
+
+
+class TestRandLora:
```
BenjaminBossan, line 51: Let's add a comment here that this is copied from `TestVera` in `test_vera.py` and changes there should be reflected here.

BenjaminBossan (changes requested):
Thanks for the latest changes and for adding the new tests. I only skimmed them, as they are mostly copy-pasted, right?

Apart from that, I only discovered a small mistake, the rest looks good.

> As I commented above, I am not sure how to proceed with the multi-GPU OOM error, let me know if you find an easy fix.

Let's not worry too much about this, since it's only DP, which very few people still use. My suggestion is to just move to a smaller base model in the GPU tests, e.g. `facebook/opt-350m`, with a comment explaining that 6.7b results in OOM.

PaulAlbert31:
Hi @BenjaminBossan,
Great to see we are getting close ! 

Yes test_randlora.py is a direct copy/paste of test_vera.py

I have not been able to reproduce the OOM error with DP as I only have 1 GPU available. Could you let me know where I can add a condition to switch to `facebook/opt-350m` for RandLoRA's tests ?

BenjaminBossan:
@PaulAlbert31 I tried running the tests from `test_gpu_examples.py` again and this time they passed without OOM. I'm not sure what's different this time around, but I think we're good now and no further changes are required.

For the CI to run, could you please call `make style` and fix the remaining linter issue?

PaulAlbert31:
@BenjaminBossan  Apologies, I keep forgetting to run `make style`
Great to hear the OOM is gone for now !

PaulAlbert31:
@BenjaminBossan 
Damn, so close. I don't have access to a windows either. 

I have had a similar problem before and reducing the randlora_alpha coefficient helped. I have now reduced it even further. Is it ok to try the windows tests again ? 

If the test still fails I'll add a skip for windows.

PaulAlbert31:
@BenjaminBossan thanks for the suggestion, I have documented the possible instabilities with larger randlora_alpha values and moved the PR status to ready 🎉

PR #2466, opened 2025-03-28
FIX Deleting adapters on auxiliary modules

### LINKED ISSUE #2381: Bug when deleting adapters of a model with modules_to_save
### System Info

All PEFT versions.

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

```python
from transformers import AutoModelForSequenceClassification

config = LoraConfig(task_type="SEQ_CLS")

model = AutoModelForSequenceClassification.from_pretrained(model_id)
adapter_to_delete = "delete_me"
model.add_adapter(adapter_to_delete, config)
# sanity check
assert "delete_me" in model.base_model.model.score.modules_to_save

model.delete_adapter(adapter_to_delete)
assert "delete_me" not in model.base_model.model.score.modules_to_save
```

### Expected behavior

When adding, say, a LoRA adapter with `modules_to_save`, then deleting the adapter, the LoRA part is correctly removed but the `modules_to_save` part is not removed.

BenjaminBossan:
Resolves #2381

For `modules_to_save` and `trainable_token_indices`, i.e. for the auxiliary modules, we did not handle adapter deletion yet. That means that if a user calls, for instance, `lora_model.delete_adapter(name)`, the LoRA adapter was deleted, but if it had `modules_to_save`, those were kept. This is inconsistent and can lead to memory leaks. This PR fixes this situation.

As a related change, the `peft_model.modules_to_save` attribute is now a `@property` that dynamically collects its entries from `peft_model.peft_config`. Otherwise, this attribute would also require bookkeeping.

I'm pretty sure that there are more edge cases where adapter deletion does not work completely as intended. The goal of this PR is not to fix all of them, but to improve on the current status.


src/peft/utils/other.py
```diff
@@ -550,6 +556,41 @@ def set_adapter(self, adapter_names: Union[str, list[str]]):
         self.modules_to_save[adapter_name].requires_grad_(True)
         self._active_adapter = adapter_name
 
+    def delete_adapter(self, adapter_name: str, new_active_adapters: Optional[list[str]]) -> None:
+        """
+        Delete the adapter if present.
+
+        This method will also set a new active adapter if the deleted adapter was the active adapter. It is important
+        that the new adapter is chosen in a deterministic way, so that the same adapter is chosen on all layers.
```
githubnemo, line 564:
```suggestion
        This method will also set a new active adapter if the deleted adapter was the active adapter. It is important
        that the new adapter is chosen by the caller in a deterministic way, so that the same adapter is chosen on all layers.
```

This is something that the caller has to watch out for, as I understand it. Or am I wrong?

tests/testing_common.py
```diff
@@ -1417,11 +1414,24 @@ def _test_delete_adapter(self, model_id, config_cls, config_kwargs):
             for attr in attributes_to_check:
                 assert adapter_to_delete not in getattr(target, attr)
 
+        has_auxiliary = bool(getattr(config, "modules_to_save", False))
+        if has_auxiliary:
+            for module in model.modules():
+                if isinstance(module, AuxiliaryTrainingWrapper):
+                    assert adapter_to_delete not in module.modules_to_save
+                    assert module.active_adapters == ["default"]
+
```
githubnemo, line 1423: This looks like it tests auxiliary training wrappers but it only tests `modules_to_save`. Maybe the following would be more general?

```suggestion
        for module in model.modules():
            if isinstance(module, AuxiliaryTrainingWrapper):
                assert adapter_to_delete not in module._adapters
                assert module.active_adapters == ["default"]

```

githubnemo (commented):
Thanks :)
Looks good, there are some minor comments but nothing fancy

tests/testing_common.py
BenjaminBossan: Yes, totally right. I changed the check and of course this revealed new bugs. Those should now be fixed too, but I'm sure they're not the last ones :-/

src/peft/utils/other.py
```diff
@@ -749,16 +754,20 @@ def delete_adapter(self, adapter_name: str, new_active_adapters: Optional[list[s
...
-        if new_active_adapters[0] not in self.active_adapters:
+        if new_active_adapters[0] not in self.token_adapter.trainable_tokens_delta:
             # a new active adapter was chosen but it seems like it has no trainable_tokens
-            self._active_adapters = []
+            self._active_adapter = []
             return
 
         new_active_adapter = new_active_adapters[0]
+        self.set_adapter(new_active_adapter)
         self.token_adapter.set_adapter(new_active_adapter)
```
githubnemo, line 771:
```suggestion
        self.set_adapter(new_active_adapter)
```

Should be enough?

githubnemo (commented):
I like the adapted tests, legible and seem to cover the specifics. Minor comment in addition to the response re `forward_wrapped_disabled`.

PR #2471, opened 2025-04-01
fix gpu tests

jiqing-feng:
Hi @BenjaminBossan . This PR fixed gpu tests. To reproduce it: `pytest tests/test_common_gpu.py::PeftGPUCommonTests::test_8bit_dora_inference`

The lora 8bit model q_proj:
image

The Dora 8bit model q_proj:
image

You can see that Dora model have an extra linear: `lora.dora.DoraLinearLayer`. 
The output cannot be the same unless the weights of `lora.dora.DoraLinearLayer` are all zero.


BenjaminBossan:
Hey, I don't think these changes are correct. As the comment suggests, the purpose of this test is:

> check for same result with and without DoRA when initializing with init_lora_weights=False

By setting `init_lora_weights=False`, we ensure that the LoRA weights are initialized such that LoRA is *not* a no-op (with `init_lora_weights=True`, the default, the `lora_B` weight is all zeros). Setting `init_lora_weights="eva"` doesn't make sense in this context. Also, the tests pass both on CI and locally, so it's unclear to me what you're trying to fix.

jiqing-feng:
That's weird. I got totally different result on A100, will check it again.

BenjaminBossan:
I get:

```
tensor([[[-1.3665, -1.3774,  2.7763,  ..., -1.3983, -1.3762, -1.4330],
         [-2.7419, -2.7511,  0.5346,  ..., -2.8071, -2.7081, -2.8664],
         [-3.5760, -3.5855,  0.5904,  ..., -3.6521, -3.5360, -3.6858],
         [-3.8018, -3.8119,  0.6282,  ..., -3.8884, -3.7535, -3.9122],
         [-3.9198, -3.9294,  0.6355,  ..., -4.0114, -3.8708, -4.0364],
         [-3.8760, -3.8866,  0.6693,  ..., -3.9737, -3.8277, -3.9972]]],
       device='cuda:0', grad_fn=<UnsafeViewBackward0>)
tensor([[[-1.3665, -1.3774,  2.7763,  ..., -1.3983, -1.3762, -1.4330],
         [-2.7419, -2.7511,  0.5346,  ..., -2.8071, -2.7081, -2.8664],
         [-3.5760, -3.5855,  0.5904,  ..., -3.6521, -3.5360, -3.6858],
         [-3.8018, -3.8119,  0.6282,  ..., -3.8884, -3.7535, -3.9122],
         [-3.9198, -3.9294,  0.6355,  ..., -4.0114, -3.8708, -4.0364],
         [-3.8760, -3.8866,  0.6693,  ..., -3.9737, -3.8277, -3.9972]]],
       device='cuda:0', grad_fn=<UnsafeViewBackward0>)
```

On a 4090, both with latest main of PEFT and v0.15.1. I tried transformers main and v4.50.3, bnb 0.45.3, torch 2.6.0. `assert torch.allclose(logits_lora, logits_dora)` passes for me too.

jiqing-feng:
I see, it's because of the numerical loss [here](https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/dora.py#L98):
image

There is some numerical loss in A100 so the results are slightly different. I don't know why no numerical loss in 4090....

jiqing-feng:
Hi @BenjaminBossan . Do you have any idea to deal with this case?

BenjaminBossan:
> I don't know why these 2 results can be the same because the computation is different [here](https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/bnb.py#L241-L254)

The DoRA part itself is a no-op without any updates to the DoRA params. The LoRA part is not a no-op since `init_lora_weights=False` but the results should still be the same because of the seed.

> There is some numerical loss in A100 so the results are slightly different. I don't know why no numerical loss in 4090....

I'm also not sure, did you check that the dtype is the same between the A100 and 4090? Also, are the torch versions identical?

jiqing-feng:
Both 4090 and A100 is float32 dtype. The torch version is 2.8.0.dev20250401+cu128 for both.

BenjaminBossan:
Could you try out how much you need to improve tolerance for the script to pass on A100?

jiqing-feng:
> Could you try out how much you need to increase tolerance for the script to pass on A100?

atol=0.2, rtol=0.1 could pass the assert.

I also selected the max and mean value of abs(logits_lora - logits_dora)
max: 0.2207
mean: 0.0235

BenjaminBossan:
Hmm, this is quite high, I don't think it makes sense to set these tolerances, as that would make the test almost meaningless. Instead, could we check the hardware and skip on A100? Maybe it's the architecture (Ampere vs Ada)?

jiqing-feng:
Hi @BenjaminBossan . As I only test A100 and 4090, don't know what kind of GPU can pass the tests. But I know XPU cannot pass so I disabled XPU only. It is enough for me. Please review the new changes. Thanks!

BenjaminBossan (approved):
Thanks for working on this.

PR #2472, opened 2025-04-01
CI: More caching in tests

BenjaminBossan:
Apply more caching in the tests, especially with regard to datasets, to avoid getting 429 errors. Mark some tests as xfail for now that would otherwise permanently fail.


BenjaminBossan:
The last run failed 4 times in a row with different Hub errors, so I'll merge despite the one run being red.

PR #2473, opened 2025-04-01
Fix compare results

jiqing-feng:
Hi @BenjaminBossan . I tested `tests/bnb/test_bnb_regression.py` on A100 and XPU with the latest `multi-backend-refactor` branch of bitsandbytes. The results might be slightly different as bnb updates. Please review this PR. Thanks!


BenjaminBossan:
Thanks for the PR. I don't think we should change the `expected_bytes`, otherwise the test would start failing on other setups, right?

Also, just to give more context: These are all pure bitsandbytes tests, they don't really have anything to do with PEFT. We added them here because bitsandbytes didn't have their own CI at the time and we wanted to help out. This will soon change and we will probably phase out the bnb tests in PEFT. Therefore, I wonder if we should not just leave the file as it is and you can skip this test on your internal XPU CI. WDYT?

jiqing-feng:
Hi @BenjaminBossan . Yes, it should be easier for me. Please review the new changes. Thanks!

BenjaminBossan:

jiqing-feng:
> We had some issues with the CI that should now be fixed, could you please merge with/rebase on the latest main?

Done. Already merged the main branch.

BenjaminBossan (approved):
Thanks, LGTM.

PR #2475, opened 2025-04-02
fix error_factor for xpu

jiqing-feng:
The output between Cuda and XPU is slightly different due to numerical loss. We need to apply a different error scale for XPU.

Hi @BenjaminBossan . Please review this PR, thanks!


tests/test_gpu_examples.py
```diff
@@ -2198,7 +2198,7 @@ class TestLoftQ:
     # The error factor indicates by how much the quantization error should be decreased when using LoftQ compared to
     # quantization without LoftQ. Thus 1.03 means that the error should be decreased by 3% at least. This is a very
     # conservative value to prevent flakiness, in practice most gains are > 1.5
-    error_factor = 1.03
+    error_factor = 1.03 if torch.cuda.is_available() else 1.005
```
BenjaminBossan, line 2201: I would suggest to change the logic a bit: If the `self.device` is XPU, set `error_factor` to 1.005, otherwise use 1.03. This way, we won't affect other possible devices that are not CUDA and the code is more self-documenting.

BenjaminBossan (changes requested):
Thanks for the PR. I would suggest a small change, please check my comment.

tests/test_gpu_examples.py
jiqing-feng: OK, cpu also need 1.005.

jiqing-feng:
Hi @BenjaminBossan . I have fixed it. Please check it. Thanks!

BenjaminBossan:

jiqing-feng:
> We had some issues with the CI that should now be fixed, could you please merge with/rebase on the latest main?

Done. Already merged the main branch.

PR #2462, opened 2025-03-27
TST Refactor tests to make them simpler

BenjaminBossan:
This PR rewrites `test_decoder_models.py` to use pure pytest and remove unittest. Besides small quality of life improvements, this allows us to use `pytest.mark.parametrize` instead of the parameterized package. As the former allows us to have nested parametrization, we can also get rid of the whole `ClassInstantier` machinery, the tests are now much easier to understand and more accessible for new contributors.

The `ClassInstantier` class still exists, as other test files need to be refactored too.

PR #2478, opened 2025-04-04
TST Refactor (continued) of encoder tests

BenjaminBossan:
Follow up to #2462 using the same approach as there.

Test coverage before vs after the refactor is 100% identical.

### DIFF
--- a/tests/test_encoder_decoder_models.py
+++ b/tests/test_encoder_decoder_models.py
@@ -12,15 +12,31 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import tempfile
-import unittest
 
+import pytest
 import torch
-from parameterized import parameterized
 from transformers import AutoModelForSeq2SeqLM, AutoModelForTokenClassification
 
-from peft import LoraConfig, PromptEncoderConfig, TaskType, get_peft_model
+from peft import (
+    AdaLoraConfig,
+    BOFTConfig,
+    BoneConfig,
+    FourierFTConfig,
+    HRAConfig,
+    IA3Config,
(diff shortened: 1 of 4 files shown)


tests/test_encoder_decoder_models.py
```diff
@@ -12,35 +12,171 @@
...
 ]
 
-FULL_GRID = {"model_ids": PEFT_ENCODER_DECODER_MODELS_TO_TEST, "task_type": "SEQ_2_SEQ_LM"}
-
-
-class PeftEncoderDecoderModelTester(unittest.TestCase, PeftCommonTester):
-    r"""
-    Test if the PeftModel behaves as expected. This includes:
-    - test if the model has the expected methods
+# Note: Missing from this list are LoKr, LoHa, LN Tuning
```
githubnemo, line 47: Any particular reason? If so, this would be helpful to add.

githubnemo (commented):
Nice :)
One question, otherwise good to go.

tests/test_encoder_decoder_models.py
BenjaminBossan: I think it was just forgotten or not considered important when those methods were added. My goal with the refactors was to replicate as close as possible the existing tests (hence I check the line coverage), modifying the test matrix would counteract that goal (and who knows if these methods pass). I agree though that eventually they should be added.

githubnemo: Then this note should also state that they are supposed to be added when there's time. Without that the reader won't benefit from this note IMO

githubnemo (approved):

PR #2482, opened 2025-04-07
Update nightly-bnb.yml

mostafawnsh48:


githubnemo:
This is an old change and the changes are already applied.

PR #2483, opened 2025-04-08
Use Python 3.9 as RUFF target version and apply fixes

cyyever:
Apply RUFF fixes based on Python 3.9 features. Most changes are about type annotations. Rule `C420` has been disabled for readability. Overall, it's safe to merge them.

### DIFF
--- a/examples/boft_controlnet/utils/light_controlnet.py
+++ b/examples/boft_controlnet/utils/light_controlnet.py
@@ -14,7 +14,7 @@
 
 
 from dataclasses import dataclass
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Optional, Union
 
 import torch
 from diffusers.configuration_utils import ConfigMixin, register_to_config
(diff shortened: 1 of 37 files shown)


tests/test_initialization.py
```diff
@@ -2661,7 +2661,7 @@ def fn(x, *args):
         if prepare_layer_inputs_keys is None:
             prepare_layer_inputs_fn = fn
         else:
-            prepare_layer_inputs_fn = {k: fn for k in prepare_layer_inputs_keys}
+            prepare_layer_inputs_fn = dict.fromkeys(prepare_layer_inputs_keys, fn)
```
BenjaminBossan, line 2664: Not a fan of this change but it's not worth it to fight ruff over this.

BenjaminBossan (approved):
Thanks for updating the ruff config and applying the changes.

tests/test_initialization.py
cyyever: Can revert them

BenjaminBossan:
> Sorry, too many changes to revert.

No worries, just something to keep in mind for the future.

PR #2488, opened 2025-04-10
FIX: X-LoRA error when targeting different modules

### LINKED ISSUE #2485: X-LoRA keyerror if expert project not exists
### System Info

X-LoRA Linear layer check if the expert is dora at here:

https://github.com/huggingface/peft/blob/13c81df8438d892be457b84175b4f303caec62b1/src/peft/tuners/xlora/layer.py#L111

but the active_adapter may not in the target if it do not change the linear layer here.

for example:

```
# expert one
{
  "target_modules": [
    "v_proj",
    "k_proj",
    "q_proj"
  ],
}
# expert two
{
  "target_modules": [
    "v_proj",
    "q_proj"
  ],
}
```

The dora check code may change to use `self.target.use_dora.get(active_adapter)`

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

Use the example at https://huggingface.co/docs/peft/package_reference/xlora#peft.XLoraModel with two lora adapter with different target modules.

### Expected behavior

Forward as expected

BenjaminBossan:
Resolves #2485

Fixes an issue with X-LoRA that resulted in an error when the individual LoRA adapters were targeting different modules.

### DIFF
--- a/src/peft/tuners/xlora/layer.py
+++ b/src/peft/tuners/xlora/layer.py
@@ -107,11 +107,11 @@ class XLoraLinearLayer(XLoraLayer):
         # Ignore if disabled. We want to make sure this is always run.
         if not self.target.merged:
             for adapter_n, active_adapter in enumerate(self.target.active_adapters):
+                if active_adapter not in self.target.lora_A.keys():
+                    continue
                 # TODO: implement X-LoRA with Lora+Dora layers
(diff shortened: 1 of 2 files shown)


EricLBuehler (approved):
Looks great.

PR #2489, opened 2025-04-10
FIX: Error when merging LoRA bias with scale != 1

### LINKED ISSUE #2486: LoRA Bias Merge Not Done Correctly
### System Info

N/A

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

N/A

### Expected behavior

I am looking at the code in: `src/peft/tuners/lora/layer.py`

If the `lora_bias` argument is set to `True`, then the forward pass for a Linear layer produces:
`output = Wx + b + scaling*(BAx + l)` 
where W, b are the weight and bias of the original linear layer, BA are the LoRA weight updates, and l is the LoRA bias term. 

In the merge() function, it is correctly merging the regular and LoRA weights as:
`W-new = W + scaling*BA`

But it seems to be incorrectly merging the bias terms as:
`b-new = b + l`
Don't we instead want:
`b-new = b + scaling*l` because during the training of the LoRA bias term it was always scaled in the forward() function?

BenjaminBossan:
Resolves #2486

When merging with LoRA bias (i.e. setting lora_bias=True), the scaling was not considered, leading to incorrect results when scaling is not 1. This is now fixed.

Note: This was missed by the unit tests since the default values of `r` and `lora_alpha` are both 8, resulting in a scale of 1. The tests have now been altered to use a scale != 1 for `lora_bias=True`.

### DIFF
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -123,6 +123,7 @@ TEST_CASES = [
     ("Conv3d 1 LoRA with DoRA", "Conv3d", LoraConfig, {"target_modules": ["conv3d"], "use_dora": True}),
     ("Conv3d 2 LoRA with DoRA", "Conv3d", LoraConfig, {"target_modules": ["conv3d", "lin0"], "use_dora": True}),
     # LoRA with lora_B bias enabled (note: embedding is not supported)
+    # It's important to set lora_alpha != r to ensure that scaling is taken into account correctly
     (
         "Vanilla MLP 1 LoRA with lora_b bias",
         "MLP",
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Heads up @sayakpaul since LoRA bias was introduced for Flux LoRA IIRC.

tests/test_custom_models.py
```diff
@@ -127,10 +127,20 @@
         "Vanilla MLP 1 LoRA with lora_b bias",
         "MLP",
         LoraConfig,
-        {"target_modules": ["lin0", "lin1"], "lora_bias": True},
+        {"target_modules": ["lin0", "lin1"], "lora_bias": True, "lora_alpha": 32},
```
sayakpaul, line 130: Perhaps I would add a comment to reflect the note you mentioned in the OP.

BenjaminBossan: Good point, I added a comment:

> # It's important to set lora_alpha != r to ensure that scaling is taken into account correctly

githubnemo (approved):

PR #2490, opened 2025-04-10
Testing common uses situational HF_HUB_OFFLINE

githubnemo:
Employ offline mode when the model was already accessed once from the hub in order to speed up the CI and make the process less prone to rate limiting.

The idea here is that we can mark contexts that, once they were visited once for a specific model id, we can assume that they are cached locally and can set `HF_HUB_OFFLINE=1` for this context. This PR tests this concept for `testing_common` which is already a big chunk of the tests and probably has the biggest gain given the amount of change.

We already saw that the assumption does not always hold true: for the prompt tuning tests (`_test_prepare_input_for_generation`) there is a case where one time the tokenizer is not used for model X and after that time the tokenizer is used - since we're setting the hub to offline for the second time the tokenizer `from_pretrained` call will fail. This problem is alleviated by adding the tokenizer name to the model id as cache identifier.

### DIFF
--- a/tests/testing_common.py
+++ b/tests/testing_common.py
@@ -233,22 +233,45 @@ class ClassInstantier(OrderedDict):
 
 
 @contextmanager
-def model_cache_context(kind, model_ids):
+def hub_online_once(model_id: str):
+    """Set env[HF_HUB_OFFLINE]=1 (and patch transformers/hugging_face_hub to think that it was always that way)
+    for model ids that were seen already so that the hub is not contacted twice for the same model id in said context.
+    The cache (`HUB_MODEL_ACCESSES`) also tracks the number of cache hits per model id.
+
+    The reason for doing a context manager and not patching specific methods (e.g., `from_pretrained`) is that there
+    are a lot of places (`PeftConfig.from_pretrained`, `get_peft_state_dict`, `load_adapter`, ...) that possibly
+    communicate with the hub to download files / check versions / etc.
+
+    Note that using this context manager can cause problems when used in code sections that access different resources.
+    Example:
+
+    ```
+    def test_something(model_id, config_kwargs):
+        with hub_online_once(model_id):
+            model = ...from_pretrained(model_id)
+            self.do_something_specific_with_model(model)
+    ```
+    It is assumed that `do_something_specific_with_model` is an absract method that is implement by several tests.
+    Imagine the first test simply does `model.generate([1,2,3])`. The second call from another test suite however uses
+    a tokenizer (`AutoTokenizer.from_pretrained(model_id)`) - this will fail since the first pass was online but didn't
+    use the tokenizer and we're now in offline mode and cannot fetch the tokenizer. The recommended workaround is to
+    extend the cache key (`model_id` passed to `hub_online_once` in this case) by something in case the tokenizer is
+    used, so that these tests don't share a cache pool with the tests that don't use a tokenizer.
+    """
     global HUB_MODEL_ACCESSES
     override = {}
 
-    if isinstance(model_ids, str):
-        model_ids = [model_ids]
     try:
-        if all((kind, m) in HUB_MODEL_ACCESSES for m in model_ids):
+        if model_id in HUB_MODEL_ACCESSES:
             override = {"HF_HUB_OFFLINE": "1"}
-            for m in model_ids:
-                HUB_MODEL_ACCESSES[(kind, m)] += 1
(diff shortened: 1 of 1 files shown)


tests/testing_common.py
```diff
@@ -228,6 +232,45 @@ def get_grid_parameters(self, grid_parameters, filter_params_func=None):
         return generated_tests
 
 
+@contextmanager
+def model_cache_context(kind, model_ids):
```
BenjaminBossan, line 236: Let's add a nice docstring.

tests/testing_common.py
```diff
@@ -228,6 +232,45 @@ def get_grid_parameters(self, grid_parameters, filter_params_func=None):
...
+        for m in model_ids:
+            if HUB_MODEL_ACCESSES.get((kind, m)) == 0:
+                del HUB_MODEL_ACCESSES[(kind, m)]
+        raise
+    finally:
+        pass
+
+
+@contextmanager
+def hub_online_once(model_ids):
```
BenjaminBossan, line 269: Let's also add a short docstring.

tests/testing_common.py
```diff
@@ -228,6 +232,45 @@ def get_grid_parameters(self, grid_parameters, filter_params_func=None):
         return generated_tests
 
 
+@contextmanager
+def model_cache_context(kind, model_ids):
+    global HUB_MODEL_ACCESSES
+    override = {}
+
+    if isinstance(model_ids, str):
+        model_ids = [model_ids]
```
BenjaminBossan, line 241: Isn't `model_ids` always a str? I couldn't see a different case. There is `model_id + config_kwargs.get("tokenizer_name_or_path", "")` but that's also a string, right?

tests/testing_common.py
```diff
@@ -63,6 +65,8 @@
 from .testing_utils import get_state_dict
 
 
+HUB_MODEL_ACCESSES = {}
```
BenjaminBossan, line 68: Could some logic be simplified if this where a defaultdict with 0 default? Theoretically, `False` / `True` would also be enough but I see the value in counting the cache hits.

tests/testing_common.py
```diff
@@ -228,6 +232,45 @@ def get_grid_parameters(self, grid_parameters, filter_params_func=None):
...
+            for m in model_ids:
+                HUB_MODEL_ACCESSES[(kind, m)] += 1
+        else:
+            for m in model_ids:
+                if (kind, m) not in HUB_MODEL_ACCESSES:
+                    HUB_MODEL_ACCESSES[(kind, m)] = 0
+        with (
+            mock.patch.dict(os.environ, override),
+            mock.patch("huggingface_hub.constants.HF_HUB_OFFLINE", override.get("HF_HUB_OFFLINE", False) == "1"),
+            mock.patch("transformers.utils.hub._is_offline_mode", override.get("HF_HUB_OFFLINE", False) == "1"),
```
BenjaminBossan, line 254: damn

BenjaminBossan (changes requested):
Great job, this really helps a lot. I only have a few small comments.

tests/testing_common.py
githubnemo: I thought that it might be possible to have sections where there are several models in the context but so far I didn't find something like that. It also turned out to be useless in its current form for the tokenizer case since the tokenizer default value ("") would still produce a cache miss, even if `model_id` was already hit (-> you'd have to ignore empty entries).

I think I'm going to remove it and make it only accept a string.

tests/testing_common.py
githubnemo: Yes, counting cache hits was my goal and I think it can still be of value. I'm inclined to keep it like that.

tests/testing_common.py
```diff
@@ -228,6 +232,59 @@ def get_grid_parameters(self, grid_parameters, filter_params_func=None):
         return generated_tests
 
 
+@contextmanager
```
BenjaminBossan, line 235: I'm wondering if the cm should be moved to `testing_utils.py` in case we want to apply it elsewhere (i.e. outside of  `PeftCommonTester`).

BenjaminBossan (approved):
Thanks for cleaning up. LGTM.

Just a small comment, but this can also be ignored or done later.

tests/testing_common.py
githubnemo: Yep. I'd do this once we apply this to other tests.

PR #2493, opened 2025-04-11
make test_lora_use_dora_linear pass on XPU

yao-matrix:
**command**
`pytest -rA tests/test_initialization.py::TestLoraInitialization::test_lora_use_dora_linear`

**symptom**
This test is flaky on XPU, `torch.use_deterministic_algorithms()` can make it always pass by enabling deterministic algorithms.

**proposed fix**
I added a `require_deterministic` decorator for such cases: while entering this decorator, it will save current deterministic state and enable deterministic algorithm for target test case; while exiting from this decorator, it will restore before deterministic state.

**impact**
XPU: before this PR **fail**, after this PR **pass**
A100: before this PR **pass**, after this PR **pass**

Also the new decorator can fix below 2 fails mentioned in https://github.com/huggingface/peft/pull/2471, @jiqing-feng, so I remove them from skip cases.

### DIFF
--- a/tests/test_common_gpu.py
+++ b/tests/test_common_gpu.py
@@ -60,6 +60,7 @@ from .testing_utils import (
     device_count,
     load_cat_image,
     require_bitsandbytes,
+    require_deterministic,
     require_multi_accelerator,
     require_non_cpu,
 )
@@ -1124,8 +1125,8 @@ class PeftGPUCommonTests(unittest.TestCase):
 
     @require_non_cpu
     @pytest.mark.single_gpu_tests
+    @require_deterministic
     @require_bitsandbytes
-    @pytest.mark.skipif(not torch.cuda.is_available(), reason="XPU have numerial errors")
     def test_4bit_dora_inference(self):
         # check for same result with and without DoRA when initializing with init_lora_weights=False
         bnb_config = BitsAndBytesConfig(
@@ -1164,8 +1165,8 @@ class PeftGPUCommonTests(unittest.TestCase):
 
     @require_non_cpu
     @pytest.mark.single_gpu_tests
+    @require_deterministic
     @require_bitsandbytes
-    @pytest.mark.skipif(not torch.cuda.is_available(), reason="XPU have numerial errors")
     def test_8bit_dora_inference(self):
         # check for same result with and without DoRA when initializing with init_lora_weights=False
         model = AutoModelForCausalLM.from_pretrained(
--- a/tests/test_common_gpu.py
+++ b/tests/test_common_gpu.py
@@ -60,7 +60,7 @@ from .testing_utils import (
     device_count,
     load_cat_image,
     require_bitsandbytes,
-    require_deterministic,
+    require_deterministic_for_xpu,
     require_multi_accelerator,
     require_non_cpu,
 )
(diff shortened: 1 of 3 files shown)


yao-matrix:
@BenjaminBossan, pls help review, thx

yao-matrix:
@faaany , FYI.

tests/test_common_gpu.py
```diff
@@ -1124,8 +1125,8 @@ def test_serialization_shared_tensors(self):
 
     @require_non_cpu
     @pytest.mark.single_gpu_tests
+    @require_deterministic
```
BenjaminBossan, line 1128: Let's add a comment here that this is required for the test to pass on XPU.

tests/test_common_gpu.py
```diff
@@ -1164,8 +1165,8 @@ def test_4bit_dora_inference(self):
...
```
BenjaminBossan, line 1168: Let's add a comment here that this is required for the test to pass on XPU.

tests/test_initialization.py
```diff
@@ -1053,6 +1053,7 @@ def test_lora_rslora_scaling_pattern(self):
         assert model.embed.scaling["default"] == expected_scaling["embed"]
         assert model.conv2d.scaling["default"] == expected_scaling["conv2d"]
 
+    @require_deterministic
```
BenjaminBossan, line 1056: Let's add a comment here that this is required for the test to pass on XPU.

BenjaminBossan (commented):
Generally, I'm fine with the PR. But I'm a bit surprised that a very straightforward test like `test_lora_use_dora_linear` would require setting the algos to deterministic. Do you have an idea why that would be?

I assume that the error is quite small and loosening tolerances would also make the test pass or is it a big difference?

yao-matrix:
> pytest -rA tests/test_initialization.py::TestLoraInitialization::test_lora_use_dora_linear

It's not very big for this case, and our first thought is to loose the tolerance to `1e-6` in this case. And then we test more such cases, and see we need different tolerances for different cases. Then, we are thinking it's may be needed to change tolerance in a very cautious way, if we want to target XPU as CUDA-equivalent device. And we found while switching to deterministic way, we'll meet current criteria, so our first plan is to find the ops which lead to such variance and discuss w/ our operator team to see whether we can fix them before we change tolerance.  That's the logic behind this PR, we have one work now, we need work more to make it work and performant, once it's done, we will remove this decorator case-by-case. There maybe some cases that we cannot meet current tolerance and we think we need work w/ you to lift it to a higher value, we will discuss with you then. Hope this explanation helps.

tests/test_common_gpu.py
yao-matrix: changes to use `require_deterministic_for_xpu` for explicity

yao-matrix:
@BenjaminBossan , done, thx

BenjaminBossan:
Thanks for the update @yao-matrix, could you please run `make style`?

yao-matrix:
done, @BenjaminBossan , sorry for forgetting checking style

yao-matrix:
the ci failure is not relevant to this PR.

yao-matrix:
@BenjaminBossan , pls help review, thx

BenjaminBossan (approved):
Thanks for the PR, LGTM.

PR #2496, opened 2025-04-12
Expose the config network error.

Iemand005:
I was getting an error "Can't find 'adapter_config.json' at 'unsloth/mistral-7b-instruct-v0.3-bnb-4bit'" but I had no clue why I was getting this error or what the problem was.

This error message is not useful as end users don't know where it's trying to find adapter_config.json, I thought it was trying to find it locally but it turned out it's a web request that's failing which makes way more sense.

I added the error message that was caught to the thrown error message so that the user of the library can actually see what the network error was and which url it failed to fetch. I don't know the convention you follow for error messages, feel free to change it but please include the HTTP error code and the URL that was attempted in the thrown errors.

### DIFF
--- a/src/peft/config.py
+++ b/src/peft/config.py
@@ -201,7 +201,7 @@ class PeftConfigMixin(PushToHubMixin):
                     pretrained_model_name_or_path, CONFIG_NAME, subfolder=subfolder, **hf_hub_download_kwargs
                 )
             except Exception as exc:
-                raise ValueError(f"Can't find '{CONFIG_NAME}' at '{pretrained_model_name_or_path}'") from exc
+                raise ValueError(f"Can't find '{CONFIG_NAME}' at '{pretrained_model_name_or_path}' ({exc.args[0]})") from exc
 
         loaded_attributes = cls.from_json_file(config_file)
         kwargs = {**class_kwargs, **loaded_attributes}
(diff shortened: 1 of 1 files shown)


src/peft/config.py
```diff
@@ -201,7 +201,7 @@ def from_pretrained(cls, pretrained_model_name_or_path: str, subfolder: Optional
                     pretrained_model_name_or_path, CONFIG_NAME, subfolder=subfolder, **hf_hub_download_kwargs
                 )
             except Exception as exc:
-                raise ValueError(f"Can't find '{CONFIG_NAME}' at '{pretrained_model_name_or_path}'") from exc
+                raise ValueError(f"Can't find '{CONFIG_NAME}' at '{pretrained_model_name_or_path}' ({exc.args[0]})") from exc
```
BenjaminBossan, line 204:
```suggestion
                raise ValueError(f"Can't find '{CONFIG_NAME}' at '{pretrained_model_name_or_path}'") from exc
```

src/peft/config.py
```diff
@@ -264,8 +264,8 @@ def _get_peft_type(
                     CONFIG_NAME,
                     **hf_hub_download_kwargs,
                 )
-            except Exception:
-                raise ValueError(f"Can't find '{CONFIG_NAME}' at '{model_id}'")
+            except Exception as exc:
+                raise ValueError(f"Can't find '{CONFIG_NAME}' at '{model_id}' ({exc.args[0]})") from exc
```
BenjaminBossan, line 268:
```suggestion
                raise ValueError(f"Can't find '{CONFIG_NAME}' at '{model_id}'") from exc
```

BenjaminBossan (changes requested):
Thanks for the PR, it's a good idea to re-raise from the original error. Since re-raising will show the error message from the first exception, there is no need to include it in the PEFT error message too. Otherwise, the PR LGTM.

BenjaminBossan:
@Iemand005 Did you have time to look at my reply yet?

BenjaminBossan:
@Iemand005 Do you still plan to work on this? If not, I'll probably open a new PR including your code and my suggestion sometime next week.

Issue #2477, opened 2025-04-02
Bug with 'PromptEncoderConfig' in Newer Versions of PEFT


nikiduki:
### System Info

peft==0.15.1
transformers==4.50.2

### Who can help?

@benjaminbossan @sayakpaul

When creating a `PromptEncoderConfig` with `task_type="SEQ_CLS"` on newer versions of **peft (>=0.15.0)**, the `get_peft_model` method fails with the error:
`AttributeError: 'PromptEncoderConfig' object has no attribute 'modules_to_save'`

When running the code on version **0.14.0**, no errors occur.

### Information

- [x] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [x] My own task or dataset (give details below)

### Reproduction

```python
from peft import PromptEncoderConfig, get_peft_model
from transformers import AutoModelForSequenceClassification, AutoTokenizer

model_name = "google-bert/bert-base-uncased"
tokenizer = AutoTokenizer.from_pretrained(model_name)
base_model = AutoModelForSequenceClassification.from_pretrained(model_name)

ptuning_config = PromptEncoderConfig(task_type="SEQ_CLS", num_virtual_tokens=20)

model = get_peft_model(base_model, ptuning_config)
```

### Expected behavior

When running the code on **peft >=0.15.0**, the `get_peft_model` method fails with the following error:
`AttributeError: 'PromptEncoderConfig' object has no attribute 'modules_to_save'`

In contrast, running the code on **peft 0.14.0** does not produce any errors.

BenjaminBossan:
Thanks for the report, this is indeed a regression.

@githubnemo Before the change, we have used `model.modules_to_save`, now `peft_config.modules_to_save`. Should we go back to that?

githubnemo:
We released v0.15.2 from a patch branch as a quickfix for this. #2481 is still open but should be ready soon as well :)

PR #2497, opened 2025-04-14
Breakout fix of #2481 for #2477, patch release prep

githubnemo:
This change represents the fixing part without tests and their complications to be merged as fix for #2477 and to be released as patch 0.15.2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@
 from setuptools import find_packages, setup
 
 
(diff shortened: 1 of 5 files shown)


BenjaminBossan:
Could you please merge with/rebase on the latest main with the hf hub offline context, so that we have a chance for the tests to pass?

PR #2498, opened 2025-04-15
ENH: Orthogonal LoRA layer initialization (2)

BenjaminBossan:
Continuation of, and supersedes #2389

Check discussion there for further info.


BenjaminBossan:
I compared the results with orthogonal init vs normal and Gaussian LoRA init, with all other parameters kept equal, using the MetaMathQA method comparison suite. Test accuracy on GSMK8K is 47.8% for default, 49.3% for Gaussian, and 48.9% for orthogonal. As expected, memory usage and runtime are practically identical.

I also plotted the train loss:

image

There seems to be a slight advantage for orthogonal initialization there is well, though for the first half of the run, it lags behind the other methods.

BenjaminBossan:
ping @githubnemo

githubnemo (approved):
LGTM, thanks for finishing this!

Do you think it would be worthwhile to directly add an experiment to the method comparison suite?

PR #2499, opened 2025-04-15
Add INC dispatcher

dsocek:
# What does this PR do?

Adds Intel Neural Compressor (INC) dispatcher to support loading LoRA into INC quantized model.  

Intel Neural Compressor modifies layer names or introduces new layers during quantization. Names can be seen [here](https://github.com/intel/neural-compressor/blob/main/neural_compressor/torch/algorithms/fp8_quant/_quant_common/helper_modules.py) (INC adds prefix “Patched“).  HuggingFace PEFT does not have a proper dispatcher to deal with the new names.

**Before this PR** we get the following error when trying to load LoRA weights into INC quantized model:
```bash
  File "/root/./test.py", line 124, in <module>
    pipeline.load_lora_weights("lora_model", adapter_name="user_lora")
  File "/usr/local/lib/python3.10/dist-packages/diffusers/loaders/lora_pipeline.py", line 1846, in load_lora_weights
  File "/usr/local/lib/python3.10/dist-packages/diffusers/loaders/lora_pipeline.py", line 1948, in load_lora_into_transformer
    inject_adapter_in_model(lora_config, transformer, adapter_name=adapter_name, **peft_kwargs)
  File "/usr/local/lib/python3.10/dist-packages/peft/mapping.py", line 76, in inject_adapter_in_model
    peft_model = tuner_cls(model, peft_config, adapter_name=adapter_name, low_cpu_mem_usage=low_cpu_mem_usage)
  File "/usr/local/lib/python3.10/dist-packages/peft/tuners/lora/model.py", line 142, in __init__
    super().__init__(model, config, adapter_name, low_cpu_mem_usage=low_cpu_mem_usage)
  File "/usr/local/lib/python3.10/dist-packages/peft/tuners/tuners_utils.py", line 181, in __init__
    self.inject_adapter(self.model, adapter_name, low_cpu_mem_usage=low_cpu_mem_usage)
  File "/usr/local/lib/python3.10/dist-packages/peft/tuners/tuners_utils.py", line 509, in inject_adapter
    self._create_and_replace(peft_config, adapter_name, target, target_name, parent, current_key=key)
  File "/usr/local/lib/python3.10/dist-packages/peft/tuners/lora/model.py", line 237, in _create_and_replace
    new_module = self._create_new_module(lora_config, adapter_name, target, device_map=device_map, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/peft/tuners/lora/model.py", line 348, in _create_new_module
    raise ValueError(
ValueError: Target module PatchedLinear(
  original=Linear, in_features=3072, out_features=3072, bias=True, scale_input dtype=float, scale_weight dtype=float
  (quant_input): QuantInput(lp_dtype=torch.float8_e4m3fn, hp_dtype=torch.bfloat16, scale_inv dtype=float)
  (dequant_output): QuantDequantNone(lp_dtype=torch.float8_e4m3fn, hp_dtype=torch.bfloat16, doesn't quantize nor dequantize)
) is not supported. 
Currently, only the following modules are supported: `torch.nn.Linear`, `torch.nn.Embedding`, `torch.nn.Conv1d`, 
`torch.nn.Conv2d`, `torch.nn.Conv3d`, `transformers.pytorch_utils.Conv1D`, `torch.nn.MultiheadAttention.`.
```

**After this PR**, LoRA weights are loaded into INC quantized model without any issues, and performance of the model is similar to that of original quantized model ✅

### DIFF
--- a/examples/stable_diffusion/inc_flux_lora_hpu.py
+++ b/examples/stable_diffusion/inc_flux_lora_hpu.py
@@ -1,10 +1,3 @@
-import importlib
-
-import torch
-from neural_compressor.torch.quantization import FP8Config, convert, finalize_calibration, prepare
-from optimum.habana.diffusers import GaudiFluxPipeline
-
-
 """
 This exampe demonstrates loading of LoRA adapter (via PEFT) into an FP8 INC-quantized FLUX model.
 
@@ -14,6 +7,11 @@ https://github.com/intel/neural-compressor/tree/master/examples/helloworld/fp8_e
 Requirements:
 pip install optimum-habana sentencepiece neural-compressor[pt] peft
 """
+
+import importlib
+
+import torch
+from neural_compressor.torch.quantization import FP8Config, convert, finalize_calibration, prepare
 
 
 # Checks if HPU device is available
@@ -31,9 +29,10 @@ def is_hpu_available():
 
 
 # Ensure HPU device is available before proceeding
-if not is_hpu_available():
+if is_hpu_available():
+    from optimum.habana.diffusers import GaudiFluxPipeline
+else:
     raise RuntimeError("HPU device not found. This code requires Intel Gaudi device to run.")
-
 
 # Example: FLUX model inference on HPU via optimum-habana pipeline
 hpu_configs = {
--- a/examples/stable_diffusion/inc_flux_lora_hpu.py
+++ b/examples/stable_diffusion/inc_flux_lora_hpu.py
@@ -1,0 +1,65 @@
+import os
+import torch
+import importlib
+"""
+This exampe demonstrates loading of LoRA adapter (via PEFT) into an FP8 INC-quantized FLUX model.
+
+More info on Intel Neural Compressor (INC) FP8 quantization is available at:
+https://github.com/intel/neural-compressor/tree/master/examples/helloworld/fp8_example
+
+Requirements:
+pip install optimum-habana sentencepiece neural-compressor[pt] peft
+"""
+
+# Checks if HPU device is available
+# Adapted from https://github.com/huggingface/accelerate/blob/b451956fd69a135efc283aadaa478f0d33fcbe6a/src/accelerate/utils/imports.py#L435
+def is_hpu_available():
+    if (
+        importlib.util.find_spec("habana_frameworks") is None
+        or importlib.util.find_spec("habana_frameworks.torch") is None
+    ):
+        return False
+
+    import habana_frameworks.torch  # noqa: F401
+
+    return hasattr(torch, "hpu") and torch.hpu.is_available()
+
+
+# Ensure HPU device is available before proceeding
+if not is_hpu_available():
+    raise RuntimeError("HPU device not found. This code requires Intel Gaudi device to run.")
+
+
+# Example: FLUX model inference on HPU via optimum-habana pipeline
+from optimum.habana.diffusers import GaudiFluxPipeline
+hpu_configs = {
+    "use_habana": True,
+    "use_hpu_graphs": True,
+    "sdp_on_bf16": True,
+    "gaudi_config": "Habana/stable-diffusion",
+}
+pipe = GaudiFluxPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", torch_dtype=torch.bfloat16, **hpu_configs)
+prompt = "A picture of sks dog in a bucket"
+
+# Quantize FLUX transformer to FP8 using INC (Intel Neural Compressor)
+from neural_compressor.torch.quantization import FP8Config, convert, prepare, finalize_calibration
+quant_configs = {
+    "mode": "AUTO",
+    "observer": "maxabs",
+    "scale_method": "maxabs_hw",
+    "allowlist": {"types": [], "names":  []},
+    "blocklist": {"types": [], "names":  []},
+    "dump_stats_path": "/tmp/hqt_output/measure"
+}
+config = FP8Config(**quant_configs)
+pipe.transformer = prepare(pipe.transformer, config)
+pipe(prompt)
+finalize_calibration(pipe.transformer)
+pipe.transformer = convert(pipe.transformer)
+
+# Load LoRA weights with PEFT
+pipe.load_lora_weights("dsocek/lora-flux-dog", adapter_name="user_lora")
+
+# Run inference
+image = pipe(prompt).images[0]
+image.save("dog.png")
--- a/examples/stable_diffusion/inc_flux_lora_hpu.py
+++ b/examples/stable_diffusion/inc_flux_lora_hpu.py
@@ -1,6 +1,10 @@
-import os
-import torch
 import importlib
+
+import torch
+from neural_compressor.torch.quantization import FP8Config, convert, finalize_calibration, prepare
+from optimum.habana.diffusers import GaudiFluxPipeline
+
+
 """
 This exampe demonstrates loading of LoRA adapter (via PEFT) into an FP8 INC-quantized FLUX model.
 
@@ -10,6 +14,7 @@ https://github.com/intel/neural-compressor/tree/master/examples/helloworld/fp8_e
 Requirements:
 pip install optimum-habana sentencepiece neural-compressor[pt] peft
 """
+
 
 # Checks if HPU device is available
 # Adapted from https://github.com/huggingface/accelerate/blob/b451956fd69a135efc283aadaa478f0d33fcbe6a/src/accelerate/utils/imports.py#L435
@@ -31,7 +36,6 @@ if not is_hpu_available():
 
 
 # Example: FLUX model inference on HPU via optimum-habana pipeline
-from optimum.habana.diffusers import GaudiFluxPipeline
 hpu_configs = {
     "use_habana": True,
     "use_hpu_graphs": True,
@@ -42,14 +46,13 @@ pipe = GaudiFluxPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", torch_d
 prompt = "A picture of sks dog in a bucket"
 
 # Quantize FLUX transformer to FP8 using INC (Intel Neural Compressor)
-from neural_compressor.torch.quantization import FP8Config, convert, prepare, finalize_calibration
 quant_configs = {
     "mode": "AUTO",
     "observer": "maxabs",
     "scale_method": "maxabs_hw",
-    "allowlist": {"types": [], "names":  []},
-    "blocklist": {"types": [], "names":  []},
-    "dump_stats_path": "/tmp/hqt_output/measure"
+    "allowlist": {"types": [], "names": []},
+    "blocklist": {"types": [], "names": []},
+    "dump_stats_path": "/tmp/hqt_output/measure",
 }
 config = FP8Config(**quant_configs)
 pipe.transformer = prepare(pipe.transformer, config)
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -145,6 +145,9 @@ class LoraLayer(BaseTunerLayer):
         elif hasattr(base_layer, "W_q") and base_layer.__class__.__name__ == "HQQLinear":
             # HQQ layers
             in_features, out_features = base_layer.in_features, base_layer.out_features
+        elif base_layer.__class__.__name__.startswith("Patched"):
+            # INC layers
+            in_features, out_features = base_layer.in_features, base_layer.out_features
         else:
             # possibly support user provided custom layer types using dynamic dispatch
             if hasattr(base_layer, "in_features") and hasattr(base_layer, "out_features"):
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -145,7 +145,7 @@ class LoraLayer(BaseTunerLayer):
         elif hasattr(base_layer, "W_q") and base_layer.__class__.__name__ == "HQQLinear":
             # HQQ layers
             in_features, out_features = base_layer.in_features, base_layer.out_features
-        elif base_layer.__class__.__name__.startswith("Patched"):
+        elif base_layer.__class__.__name__ == "PatchedLinear":
             # INC layers
             in_features, out_features = base_layer.in_features, base_layer.out_features
         else:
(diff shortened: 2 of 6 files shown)


dsocek:
@BenjaminBossan Thanks for the review.  A `helloworld` type of example can be found at [Intel Neural Compressor (INC) repo](https://github.com/intel/neural-compressor/tree/master/examples/helloworld/fp8_example#one-step-to-get-quantized-model), but this does not have lora loading example. 

To actually test this PR with PEFT, here's an example of FLUX model inference on HPU via optimum-habana pipeline (his demonstrates loading a PEFT-tuned model with FP8 quantization):

```py
#!/usr/bin/env python
import os
import torch

# Example: FLUX model inference on HPU via optimum-habana pipeline
from optimum.habana.diffusers import GaudiFluxPipeline
hpu_configs = {
    "use_habana": True,
    "use_hpu_graphs": True,
    "sdp_on_bf16": True,
    "gaudi_config": "Habana/stable-diffusion",
}
pipe = GaudiFluxPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", torch_dtype=torch.bfloat16, **hpu_configs)
prompt = "A picture of sks dog in a bucket"

# Quantize FLUX transformer to FP8 using INC
from neural_compressor.torch.quantization import FP8Config, convert, prepare, finalize_calibration
quant_configs = {
    "mode": "AUTO",
    "observer": "maxabs",
    "scale_method": "maxabs_hw",
    "allowlist": {"types": [], "names":  []},
    "blocklist": {"types": [], "names":  []},
    "dump_stats_path": "/tmp/hqt_output/measure"
}
config = FP8Config(**quant_configs)
pipe.transformer = prepare(pipe.transformer, config)
pipe(prompt)
finalize_calibration(pipe.transformer)
pipe.transformer = convert(pipe.transformer)

# Load LoRA weights with PEFT
pipe.load_lora_weights("dsocek/lora-flux-dog", adapter_name="user_lora") # <--- FAILS WITHOUT INC DISPATCHER

# Run inference
image = pipe(prompt).images[0]
image.save("dog.png")
```

Requirements:

```bash
pip install optimum-habana sentencepiece neural-compressor[pt] peft
```

dsocek:
> I assume some features like merging of LoRA weights would not work yet?

We could address merging and other additional features in a different PR, if that's ok

src/peft/tuners/lora/layer.py
```diff
@@ -145,6 +145,9 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
         elif hasattr(base_layer, "W_q") and base_layer.__class__.__name__ == "HQQLinear":
             # HQQ layers
             in_features, out_features = base_layer.in_features, base_layer.out_features
+        elif base_layer.__class__.__name__.startswith("Patched"):
+            # INC layers
+            in_features, out_features = base_layer.in_features, base_layer.out_features
```
BenjaminBossan, line 150: Is this not already covered by the `else` clause below?

BenjaminBossan (changes requested):
Thanks for providing the example. I have a small comment. Could you please also run `make style`?

I think it would be great to add it to `peft/examples/stable_diffusion/`. Let's also add a check at the top of the script if the required hardware is present, + a small description what the script does.

> We could address merging and other additional features in a different PR, if that's ok

Generally, that would be fine, we have other quantization without merging support. However, right now, the implementation dispatches to the vanilla LoRA layers, so if a user calls `merge`, the normal merge method is called, which would result in an error (I assume, I haven't tested) or may even fail silently. This is not very user friendly. IMO it would be better to dispatch to special INC layers, similar to what we do with other quantization methods. The layers would not implement `merge`  and `unmerge`, or raise a `NotImplementedError`, so that users know what's going on. Later, it will be easier to add those methods to the INC layers. WDYT?

src/peft/tuners/lora/layer.py
dsocek: It looks like its covered, but we get this warning:
```bash
/root/peft/src/peft/tuners/lora/layer.py:157: UserWarning: Unsupported layer type '<class 'neural_compressor.torch.algorithms.fp8_quant._quant_common.helper_modules.PatchedLinear'>' encountered, proceed at your own risk.
```
With added `elif` we avoid the `Unsupported later type` warning.

dsocek:
> Thanks for providing the example. I have a small comment. Could you please also run `make style`?
> 
> I think it would be great to add it to `peft/examples/stable_diffusion/`. Let's also add a check at the top of the script if the required hardware is present, + a small description what the script does.

Sounds good, I will fix style with `make style` in next commit. Are you thinking to add a standalone HPU INC quantized Flux LoRA inference example py file, or somehow integrate into existing example script?  A standalone can be done quickly.  Do you also want to have example with and without quantization (somehow control via args or so)?

A check `is_hpu_available()` was recently added to `accelerate` (see [here](https://github.com/huggingface/accelerate/blob/b451956fd69a135efc283aadaa478f0d33fcbe6a/src/accelerate/utils/imports.py#L435)), should we consider adding similar check in `peft/import_utils` and just call this from the example?

> > We could address merging and other additional features in a different PR, if that's ok
> 
> Generally, that would be fine, we have other quantization without merging support. However, right now, the implementation dispatches to the vanilla LoRA layers, so if a user calls `merge`, the normal merge method is called, which would result in an error (I assume, I haven't tested) or may even fail silently. This is not very user friendly. IMO it would be better to dispatch to special INC layers, similar to what we do with other quantization methods. The layers would not implement `merge` and `unmerge`, or raise a `NotImplementedError`, so that users know what's going on. Later, it will be easier to add those methods to the INC layers. WDYT?

I see, so you would like to add specific INC LoRA layers which define `merge` and `unmerge` (like say in HQQ) at least as placeholders with `NotImplementedError`, correct?

I am not sure when/how is merge/unmerge called with diffusers (sorry for my misunderstanding here) to even test what happens currently :) I know we can call something like `pipe.set_adapters(["user_lora", "user_lora2"], adapter_weights=[0.7, 0.8])` for example to "merge" 2 adapters but I think maybe this is not the same merge/unmerge on layer level.  I also know one can use `pipe.fuse_lora()` which will fuse lora with actual model (but we can't unfuse then), will this trigger `merge`?

BenjaminBossan:
> Are you thinking to add a standalone HPU INC quantized Flux LoRA inference example py file, or somehow integrate into existing example script? A standalone can be done quickly. Do you also want to have example with and without quantization (somehow control via args or so)?

Standalone is fine.

> A check `is_hpu_available()` was recently added to `accelerate` (see [here](https://github.com/huggingface/accelerate/blob/b451956fd69a135efc283aadaa478f0d33fcbe6a/src/accelerate/utils/imports.py#L435)), should we consider adding similar check in `peft/import_utils` and just call this from the example?

Normally, we would just use accelerate, but since it's so recent, I think it's better to copy+paste with a comment about the source.

> I see, so you would like to add specific INC LoRA layers which define `merge` and `unmerge` (like say in HQQ) at least as placeholders with `NotImplementedError`, correct?

Yes, that's the suggestion.

> I am not sure when/how is merge/unmerge called with diffusers (sorry for my misunderstanding here) to even test what happens currently :)

It looks like it's used here: https://github.com/huggingface/diffusers/blob/026507c06cdabfb0c13ddeb1ac33f5d8e244361f/src/diffusers/loaders/peft.py#L698-L737

But even if it weren't: This is not exclusive to diffusion models, right? People could use INC for normal LLMs and then later may want to merge the weights, for instance for faster inference. So having the option would be good in the long run, but not necessary for a start. If you need help with testing, LMK.

dsocek:
@BenjaminBossan  Thanks for quick reply and clarification.  All sounds good.  I will implement the placeholders and should be able to test them via merge/unmerge with an LLM case, rather than diffusers case.

> Normally, we would just use accelerate, but since it's so recent, I think it's better to copy+paste with a comment about the source.

I assume you mean copy paste this function directly in the example script and not add to `peft/import_utils`, correct?

BenjaminBossan:
> I will implement the placeholders and should be able to test them via merge/unmerge with an LLM case, rather than diffusers case.

Thanks, sounds good.

> I assume you mean copy paste this function directly in the example script and not add to `peft/import_utils`, correct?

If it's only needed there, then yes.

examples/stable_diffusion/inc_flux_lora_hpu.py
```diff
@@ -0,0 +1,68 @@
+import importlib
+
+import torch
+from neural_compressor.torch.quantization import FP8Config, convert, finalize_calibration, prepare
+from optimum.habana.diffusers import GaudiFluxPipeline
```
BenjaminBossan, line 5: This import fails for me:

```
  File "<string>", line 1, in <module>
  File "/home/name/work/forks/peft/examples/stable_diffusion/inc_flux_lora_hpu.py", line 5, in <module>
    from optimum.habana.diffusers import GaudiFluxPipeline
  File "/home/name/anaconda3/envs/peft/lib/python3.12/site-packages/optimum/habana/__init__.py", line 34, in <module>
    check_synapse_version()
  File "/home/name/anaconda3/envs/peft/lib/python3.12/site-packages/optimum/habana/utils.py", line 207, in check_synapse_version
    habana_frameworks_version_number = get_habana_frameworks_version()
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/name/anaconda3/envs/peft/lib/python3.12/site-packages/optimum/habana/utils.py", line 245, in get_habana_frameworks_version
    return version.parse(output.stdout.split("\n")[0].split()[-1])
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^
```

This is because I don't seem to have "habana-torch-plugin" in my env:

```sh
$ pip list | rg habana
optimum-habana            1.17.0
```

However, I did install `optimum-habana` and `neural-compressor[pt]` so not sure what the issue is.

examples/stable_diffusion/inc_flux_lora_hpu.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+"""
```
BenjaminBossan, line 8: Let's move this to the top of the file.

BenjaminBossan (changes requested):
Thanks for your continued work on this PR. From first inspection, this looks good. I tried out the example but ran into an issue, please check my comment.

Furthermore, I assume that to use INC, I would need the correct hardware, i.e. I wouldn't be able to test this locally or on the PEFT CI. Would the plan be that Intel is running its own CI to check that it works? In that case, we should add a unit test, even if it's just a very functional one. For instance, we could copy the [bnb test here](https://github.com/huggingface/peft/blob/d3ff1334a7fec4f4f17ed28a5a6cbcb048645cf1/tests/test_gpu_examples.py#L238) and make adjustments for INC. Of course, the test should be skipped if the required packages and/or hardware are not available. Intel could then run this test on their CI and inform us if there is any regression in PEFT that breaks the integration.

examples/stable_diffusion/inc_flux_lora_hpu.py
dsocek: Ah good catch, I should have moved the import if Gaudi pipeline after HPU is available passes.  But, yes you need torch.hpu hook which is not simple to setup (basically best to use officially released docker for Gaudi pytorch).  Will fix this small issue, and also move the short example description section on top.

dsocek:
> I assume that to use INC, I would need the correct hardware, i.e. I wouldn't be able to test this locally or on the PEFT CI

Yes, you would need Intel Gaudi (HPU) for this

> Would the plan be that Intel is running its own CI to check that it works? In that case, we should add a unit test, even if it's just a very functional one. For instance, we could copy the [bnb test here](https://github.com/huggingface/peft/blob/d3ff1334a7fec4f4f17ed28a5a6cbcb048645cf1/tests/test_gpu_examples.py#L238) and make adjustments for INC. Of course, the test should be skipped if the required packages and/or hardware are not available. Intel could then run this test on their CI and inform us if there is any regression in PEFT that breaks the integration.

Yes we can easily add such a test in Optimum-Habana, which uses PEFT in a number of samples.  I am contributing to OH actively so should be possible to add a CI test which will run this or similar example and if there is future failure it will be caught quickly and addressed.  We already have a number of PEFT related tests in OH for [llms](https://github.com/huggingface/optimum-habana/blob/main/tests/test_peft_inference.py) or [diffusers](https://github.com/huggingface/optimum-habana/blob/main/tests/test_diffusers.py#L2865).

dsocek:
@BenjaminBossan I fixed the small issue you indicated, see my earlier comment for your other concern.
Thanks again for quick and thorough review!

Also I did a final test and the updated example works on HPU as before, and on a system without HPU device now we get error as expected:

```bash
$ python inc_flux_lora_hpu.py
2025-04-25 20:43:06 [WARNING][auto_accelerator.py:418] Auto detect accelerator: CUDA_Accelerator.
  File "/root/peft/examples/stable_diffusion/inc_flux_lora_hpu.py", line 35, in <module>
    raise RuntimeError("HPU device not found. This code requires Intel Gaudi device to run.")
RuntimeError: HPU device not found. This code requires Intel Gaudi device to run.
```

BenjaminBossan (changes requested):
Thanks for the fixes, I can confirm that the check now works as expected.

> Yes we can easily add such a test in Optimum-Habana, which uses PEFT in a number of samples. I am contributing to OH actively so should be possible to add a CI test which will run this or similar example and if there is future failure it will be caught quickly and addressed. We already have a number of PEFT related tests in OH for [llms](https://github.com/huggingface/optimum-habana/blob/main/tests/test_peft_inference.py) or [diffusers](https://github.com/huggingface/optimum-habana/blob/main/tests/test_diffusers.py#L2865).

Sounds good, then I think we don't need dedicated tests in PEFT as well. Could you please add a comment to `inc.py` with a reference to the tests in optimum-habana so that a reader knows why there are no dedicated PEFT tests?

Apart from this, the PR looks pretty much ready. I think for easier discovery, it would be nice to have an entry in the [PEFT quantization docs](https://huggingface.co/docs/peft/developer_guides/quantization) to mention INC and the caveat that merging is not yet supported, WDYT?

dsocek:
@BenjaminBossan Thanks again for a thorough review. I incorporated your final suggestions into the PR:

- added note about external handling of CI tests to `inc.py`
- added section on `INC quantization` in the docs (with caveat subsection)

BenjaminBossan (approved):
Thanks for adding support for INC, LGTM.

PR #2503, opened 2025-04-18
TST make 3 flaky test cases always pass on XPU

yao-matrix:
PASSED tests/test_initialization.py::TestAdaLoraInitialization::test_adalora_default_init_identity
PASSED tests/test_initialization.py::TestLokrInitialization::test_lokr_linear_init_default
PASSED tests/test_gpu_examples.py::TestPTuningReproducibility::test_p_tuning_exactly_reproducible_after_loading

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -85,6 +85,7 @@ from .testing_utils import (
     require_auto_awq,
     require_auto_gptq,
     require_bitsandbytes,
+    require_deterministic_for_xpu,
     require_eetq,
     require_hqq,
     require_multi_accelerator,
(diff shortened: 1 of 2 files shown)


yao-matrix:
@BenjaminBossan, pls help review, thx.

BenjaminBossan:
This means that `require_deterministic_for_xpu` can be removed completely, right?

yao-matrix:
> This means that `require_deterministic_for_xpu` can be removed completely, right?

@BenjaminBossan seems no, actually I added 3 more cases which `require_deterministic_for_xpu` in this PR, so we still it now. Good news is seems we've found the operator leads to such numerical differences, but we still need time to see how to fix and validate, op numerical things are usually pretty complex. Once they are fixed and validated, i will submit PRs and remove `require_deterministic_for_xpu` if needed.

BenjaminBossan (approved):
Thanks for the PR, LGTM.

PR #2507, opened 2025-04-22
FIX: CPT should not be tested with sequence classification

BenjaminBossan:
PR #2481 added sequence classification tests to PEFT. The test matrix included CPT. However, CPT only supports the task type `CAUSAL_LM`. These tests still passed but now [started failing](https://github.com/huggingface/peft/actions/runs/14592717272/job/40931414213) with:

> AttributeError: object has no attribute 'prepare_inputs_for_generation'

This is probably a change in transformers but since causal LM was never meant to work, the actual fix is to remove CPT from the seq cls test matrix.

Since CPT automatically changes the task type to `CAUSAL_LM`, this mistake can be hard to spot. Therefore, this PR also adds a warning if users pass the wrong task type.

### DIFF
--- a/src/peft/tuners/cpt/config.py
+++ b/src/peft/tuners/cpt/config.py
@@ -82,7 +82,8 @@ class CPTConfig(PromptLearningConfig):
         self.peft_type = PeftType.CPT  # Specifies that the PEFT type is CPT.
         if self.task_type != TaskType.CAUSAL_LM:
             warnings.warn(
-                f"{self.__class__.__name__} only supports task_type = {TaskType.CAUSAL_LM}, setting it automatically."
+                f"{self.__class__.__name__} only supports task_type = {TaskType.CAUSAL_LM.value}, "
+                "setting it automatically."
             )
             self.task_type = "CAUSAL_LM"  # Ensures task type is causal language modeling.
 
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
FYI @tsachiblau

src/peft/tuners/cpt/config.py
```diff
@@ -79,7 +80,12 @@ def __post_init__(self):
         self.num_attention_heads = None  # Number of attention heads (if applicable).
         self.num_transformer_submodules = 1  # Number of transformer submodules used.
         self.peft_type = PeftType.CPT  # Specifies that the PEFT type is CPT.
-        self.task_type = "CAUSAL_LM"  # Ensures task type is causal language modeling.
+        if self.task_type != TaskType.CAUSAL_LM:
+            warnings.warn(
+                f"{self.__class__.__name__} only supports task_type = {TaskType.CAUSAL_LM.value}, "
+                "setting it automatically."
+            )
+            self.task_type = TaskType.CAUSAL_LM  # Ensures task type is causal language modeling.
```
sayakpaul, line 88: Is erroring out not a better option here?

sayakpaul (approved):
Thanks! I just have one question.

src/peft/tuners/cpt/config.py
sayakpaul: Okay. If that is the case, I think it might make sense to start a deprecation cycle (can be done in a future PR).

BenjaminBossan: This is a good idea, I changed the code to give a `FutureWarning`

sayakpaul (approved):
Lovely. Thanks.

PR #2512, opened 2025-04-23
LORA/MODEL: Use max rank of pattern for `add_weighted_adapter`

Beinsezii:
Otherwise it's possible that the new merged adapter won't allocate enough space when a patterned rank is larger than config.r.

For example, the `cat` method uses `sum(adapters_ranks)` for the empty tensor alloc, so if `cat(loras)` for a given layer is ever > the sum of `config.r` the assignment will overflow.

Local test fails but it's all entirely the TestEvaInitialization module which appears completely unrelated.

### DIFF
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -562,7 +562,12 @@ class LoraModel(BaseTuner):
         # if there is only one adapter, we can only use linear merging
         combination_type = "linear" if len(adapters) == 1 else combination_type
 
-        adapters_ranks = [self.peft_config[adapter].r for adapter in adapters]
+        adapters_ranks: list[int] = [
+            # When allocating tensors for the new adapter, we need the maximum possible rank to not overflow
+            config.r if not config.rank_pattern else max([config.r, *config.rank_pattern.values()])
+            for config in (self.peft_config[adapter] for adapter in adapters)
+        ]
+
         if combination_type in ("linear", "ties", "dare_ties", "dare_linear", "magnitude_prune"):
             # all adapters ranks should be same, new rank is just this value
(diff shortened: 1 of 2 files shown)


bghira:
thanks Ben, i'd requested this pull request be opened since it seems trivial enough but I'm not sure we'll have time to add unit tests here, especially considering none of us are particularly familiar with the unit test ecosystem for PEFT.

BenjaminBossan:
Thanks for providing more context @bghira. I wrote the code required to test this change:

```python
    @parameterized.expand([IA3Config, LoHaConfig, LoKrConfig, LoraConfig, HRAConfig, BoneConfig])
    def test_add_weighted_adapter_cat_with_rank_pattern(self, config_cls):
        # Fixes a bug described in #2512, which resulted from the rank_pattern not being taken into account
        config0 = LoraConfig(target_modules=["lin0", "lin1"], r=8)
        config1 = LoraConfig(target_modules=["lin0", "lin1"], r=8, rank_pattern={"lin0": 16})
        model = MLP()
        model = get_peft_model(model, config0).to(self.torch_device)
        model.add_weighted_adapter(["default", "other"], weights=[1.0, 1.0], adapter_name="merged", combination_type="cat")
```

I can best be added [in this line](https://github.com/huggingface/peft/blob/d3ff1334a7fec4f4f17ed28a5a6cbcb048645cf1/tests/test_custom_models.py#L2065). @Beinsezii would you be so kind to add it to your PR?

Beinsezii:
Verified it fails on `main` and passes on `beinsezii/weighted_adapter_max_rank_pattern`

BenjaminBossan (approved):
Thanks for adding the test and for the PR overall. LGTM.

PR #2514, opened 2025-04-24
Update Docker image builds for torch 2.7+cu126

matthewdouglas:
PyTorch 2.7 was just released. The default wheel from PyPI is built with CUDA toolkit 12.6, compared to 12.4 with earlier PyTorch releases. This resulted in broken tests for bitsandbytes, as the Docker image that builds bitsandbytes from source only builds with CUDA 12.4.

We'll update the Dockerfiles to use CUDA 12.6, matching the PyTorch 2.7 that is now used in the tests. This will result in building the libbitsandbytes_cuda126.so version that we need instead of libbitsandbytes_cuda124.so.

Secondly, I'm removing the multi-backend-refactor version of this workflow as we're starting to move that effort into main.

cc: @BenjaminBossan


BenjaminBossan:
Thanks for taking care of this. Since the multi-backend tests are removed, we can also completely remove https://github.com/huggingface/peft/tree/main/docker/peft-gpu-bnb-multi-source, right?

As a reminder for myself, once the PR is merged, I should also update the internal Notion docs.

Regarding the failing zizmor job, this can be ignored for now. The failing trufflehog job also seems unrelated.

matthewdouglas:
Absolutely. I've updated to clean that up!

PR #2519, opened 2025-04-28
fix typo for skipping test

jiqing-feng:
Hi @BenjaminBossan . I just found an typo in #2473 . We should skip the test if not cuda available.


BenjaminBossan (approved):
Good catch, thanks for fixing.

PR #2522, opened 2025-04-29
ENH: Add default Qwen3 target modules

BenjaminBossan:


sayakpaul (approved):
Do we have to test this?

BenjaminBossan:
> Do we have to test this?

I tested this locally, but we don't generally have a test for each architecture with predefined `target_modules`.

PR #2524, opened 2025-04-30
Randlora documentation and some example usage

PaulAlbert31:
Hi @BenjaminBossan and others,
This is a follow up to https://github.com/huggingface/peft/pull/2464 and issue https://github.com/huggingface/peft/issues/2441.

I have drafted a documentation for RandLora and slightly updated the example usage in the model.py docstring.

Since RandLora performs well compared to Lora on the PEFT model comparison suite, is it also possible to add RandLora to a PEFT leader board or is that something you don't do at the moment ? 

Happy to iterate or give more example usages.


PaulAlbert31:
Hi @BenjaminBossan,
Sorry about that, I missed those. I have now set up pre-commit as per the docs so to hopfully avoid things like this happening again. 

I have removed the diff lines and added the MetaMathQA config.

docs/source/package_reference/randlora.md
```diff
@@ -0,0 +1,45 @@
...
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# RandLora: Full-rank parameter-efficient fine-tuning of large models 
+[RandLora](https://huggingface.co/papers/2502.00987) is a parameter-efficient fine-tuning technique that is similar to LoRA and VeRA but performs full rank updates to improve performance. RandLora can be particulary usefull when adapting large model to hard tasks that require complex updates while preserving the parameter efficiency of LoRA. The full rank update of RandLora is acheived by linearly scaling random bases. The random bases are a collection of multiple low rank matrices such that the summation of their ranks if greater or equal to the full rank of the parameter matrices. The trainable parameters of RandLora are two diagonal matrices (vectors) that get multiplied with the right hand low rank random bases, in a similar way to VeRA's update. To maintain low memory usage, RandLora uses a custom function that prevents storing unnecessary bases in memory for backpropagation.
```
BenjaminBossan, line 18:
```suggestion
[RandLora](https://huggingface.co/papers/2502.00987) is a parameter-efficient fine-tuning technique that is similar to [LoRA](https://huggingface.co/papers/2106.09685) and [VeRA](https://huggingface.co/papers/2310.11454) but performs full rank updates to improve performance. RandLora can be particulary usefull when adapting large model to hard tasks that require complex updates while preserving the parameter efficiency of LoRA. The full rank update of RandLora is achieved by linearly scaling random bases. The random bases are a collection of multiple low rank matrices such that the summation of their ranks if greater or equal to the full rank of the parameter matrices. The trainable parameters of RandLora are two diagonal matrices (vectors) that get multiplied with the right hand low rank random bases, in a similar way to VeRA's update. To maintain low memory usage, RandLora uses a custom function that prevents storing unnecessary bases in memory for backpropagation.
```

docs/source/package_reference/randlora.md
BenjaminBossan, line 1: Could you please also add an entry to `_toctree.yml`?

docs/source/package_reference/randlora.md
```diff
@@ -0,0 +1,45 @@
...
+
+RandLora presents the noteworthy difference that contrary to other LoRA-like PEFT algorithm, increasing RandLora's random base ranks increases the amount of trainable parameters. Because number of bases x bases rank is constant in RandLora, reducing the rank will increase the number of random bases, hence the number of base-specific trainable diagonal bases.
+
+Because reducing the rank RandLora's random bases will increase their number, RandLora can become slower to train than LoRA for very small ranks where typically, ranks below 4 with result in a large training time increase. This does not affect inference though as the RandLora adapters can be merged into the pretrained weight matrices.
```
BenjaminBossan, line 22:
```suggestion
Because reducing the rank of RandLora's random bases will increase their number, RandLora can become slower to train than LoRA for very small ranks where typically, ranks below 4 with result in a large training time increase. This does not affect inference though as the RandLora adapters can be merged into the pretrained weight matrices.
```

I think that is what you mean?

BenjaminBossan (changes requested):
Thanks for adding the RandLora documentation and experiment config. The docs are really well written, well done.

I only found some minor issues that should be easily resolved, please check.

For better adoption, I would also recommend adding a full example. This can be as easy as copying one from the `examples/` directory and making the necessary adjustments for RandLora. This can also be done in a later PR if you prefer.

PaulAlbert31:
@BenjaminBossan 
Thanks for catching the typos in the docs, I have now added an example in the `example/` folder following DoRA's structure.

I am still investigating the large memory usage of RandLora I observed when running randlora_finetune.py. This goes against what I have observed outside of the peft library. 
My current guess is that the random basis are copied across layers or cast from fp16 to fp32 at some point which causes the large usages or even OOM in some cases. I'll open another PR if I find a fix.

Please let me know in case I missed something.

examples/randlora_finetuning/README.md
```diff
@@ -0,0 +1,120 @@
...
+peft_model = get_peft_model(model, lora_config)
+trainer = transformers.Trainer(
+    model=peft_model,
+    train_dataset=dataset,
+    dataset_text_field="text",
+    max_seq_length=2048,
+    tokenizer=tokenizer,
+)
+trainer.train()
+peft_model.save_pretrained("randlora-llama-3-8b")
```
BenjaminBossan, line 30: The name doesn't fit the base model.

examples/randlora_finetuning/README.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+There is no additional change needed to your standard PEFT training procedure, simply swap your LoRAConfig for a RandLoraConfig. Note however that RandLora's trainable parameter count is **inversely proportional** to the rank parameter `r`. Lower `r` to increase and increase it to reduce trainable parameters of RandLora.
```
BenjaminBossan, line 33:
```suggestion
There is no additional change needed to your standard PEFT training procedure, simply swap your `LoraConfig` for a `RandLoraConfig`. Note however that RandLora's trainable parameter count is **inversely proportional** to the rank parameter `r`. Lower `r` to increase and increase it to reduce trainable parameters of RandLora.
```

examples/randlora_finetuning/README.md
```diff
@@ -0,0 +1,120 @@
...
+```bash
+python examples/randlora_finetuning/randlora_finetuning.py --base_model meta-llama/Meta-Llama-3-8B --data_path timdettmers/openassistant-guanaco
+```
+This 👆🏻 by default will load the model in peft set up with RandLora config. Now if you wanna quickly compare it with Lora, all you need to do is to input ` --use_lora` in the command line and reduce `--randlora_alpha` to 2x the rank. So same above example would be 👇🏻;
+
+```bash
+python examples/randlora_finetuning/randlora_finetuning.py --base_model meta-llama/Meta-Llama-3-8B --data_path timdettmers/openassistant-guanaco --use_lora --randlora_alpha
+```
+
+RandLora can be made to use sparse or very sparse random bases. These sparse matrices can help reduce overfitting. To add `--very_sparse` to run with very sparse matrice or run the following for sparse matrices:
```
BenjaminBossan, line 45:
```suggestion
RandLora can be made to use sparse or very sparse random bases. These sparse matrices can help reduce overfitting. Add `--very_sparse` to run with very sparse matrices or `--sparse` for sparse matrices:
```

examples/randlora_finetuning/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+```bash
+python examples/randlora_finetuning/randlora_finetuning.py --base_model meta-llama/Meta-Llama-3-8B --quantize --sparse
```
BenjaminBossan, line 48:
```suggestion
python examples/randlora_finetuning/randlora_finetuning.py --base_model meta-llama/Meta-Llama-3-8B --sparse
```

Let's remove it here as the option is discussed in the example below.

examples/randlora_finetuning/README.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+RandLora also supports quantization. To use 4-bit quantization try:
+
+```bash
+python examples/randlora_finetuning/randlora_finetuning.py --base_model meta-llama/Meta-Llama-3-8B --quantize
+```
+
+By default the RandLora layers are the key and value layers of LLama model. Adding adapters on more layers will increase memory usage. If you whish to choose a different set of layers for RandLora to be applied on, you can simply define it using:
```
BenjaminBossan, line 57:
```suggestion
By default the RandLora layers are the key and value layers of LLama model. Adding adapters on more layers will increase memory usage. If you wish to choose a different set of layers for RandLora to be applied on, you can simply define it using:
```

examples/randlora_finetuning/README.md
```diff
@@ -0,0 +1,120 @@
...
+    --randlora_target_modules "k_proj,v_proj" \
+    --hub_model_id "YOUR_HF_REPO" \
+    --push_to_hub
+```
+
+## Use the model on 🤗
+You can load and use the model as any other 🤗 models.
+```python
+from transformers import AutoModel
+model = AutoModel.from_pretrained("ShirinYamani/huggyllama-llama-7b-finetuned")
```
BenjaminBossan, line 89: I think this example can be removed, as there is no connection to RandLora.

examples/randlora_finetuning/README.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+## RandLora vs. LoRA
+RandLora differs from LoRA and other related low rank approximation algorithms by chanllenging the low rank paradigm. RandLora adapters learn **full-rank** updates as the [paper](https://huggingface.co/papers/2502.00987) shows that the low rank constraint of LoRA can constrain performance gains as trainable parameters increase (with higher ranks). As a result, using RandLora is specifically recommended for difficult tasks that are underfit by LoRA. RandLoRA however also often improves performance for common tasks. If increasing LoRA's rank improves performance for your task, RandLora will most likely outperform.
+
+RandLora is expected to:
+
+1. Increase performance over LoRA for equivalent amounts of trainable parameters, mostly for larger equivalent amounts (> LoRA rank 4). This performance increase is dependent on using a large randlora_alpha scaling parameter (usually 20x the basis rank). This large parameter can sometimes make training the update unstable, reduce the learning rate or the scaling parameter if this is the case.
```
BenjaminBossan, line 97:
```suggestion
1. Increase performance over LoRA for equivalent amounts of trainable parameters, mostly for larger equivalent amounts (> LoRA rank 4). This performance increase is dependent on using a large `randlora_alpha` scaling parameter (usually 20x the basis rank). This large parameter can sometimes make training the update unstable, reduce the learning rate or the scaling parameter if this is the case.
```

examples/randlora_finetuning/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+2. Increase training time when using very low RandLora basis ranks. RandLora can increase training time when a large amount of random bases is used (low randlora rank).
```
BenjaminBossan, line 99: The two sentences are a bit redundant, could you please clarify?

examples/randlora_finetuning/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+RandLora is expected to:
+
+1. Increase performance over LoRA for equivalent amounts of trainable parameters, mostly for larger equivalent amounts (> LoRA rank 4). This performance increase is dependent on using a large randlora_alpha scaling parameter (usually 20x the basis rank). This large parameter can sometimes make training the update unstable, reduce the learning rate or the scaling parameter if this is the case.
+
+2. Increase training time when using very low RandLora basis ranks. RandLora can increase training time when a large amount of random bases is used (low randlora rank). 
+
+## RandLora vs. VeRA
+RandLora shares similarities with VeRA in that both algorithms use random basis combinations to address some of LoRA's limitations. The limitations addressed by each algorithm is however different.
+VeRA aims to reduce trainable parameters beyond rank 1 LoRAs while RandLoRA reduces the performance limitation due the low rank of the update as the trainable parameter count increases.
```
BenjaminBossan, line 103:
```suggestion
VeRA aims to reduce trainable parameters beyond rank 1 LoRAs while RandLoRA reduces the performance limitation due to the low rank of the update as the trainable parameter count increases.
```

examples/randlora_finetuning/randlora_finetuning.py
```diff
@@ -0,0 +1,223 @@
+import os
```
BenjaminBossan, line 1: Could you please add a comment that this is based on `examples/dora_finetuning/dora_finetuning.py`?

examples/randlora_finetuning/randlora_finetuning.py
```diff
@@ -0,0 +1,223 @@
...
+
+    # Define training arguments
+    training_args = TrainingArguments(
+        output_dir=output_dir,
+        num_train_epochs=num_epochs,
+        per_device_train_batch_size=batch_size,
+        per_device_eval_batch_size=batch_size,
+        warmup_steps=100,
+        weight_decay=0.01,
+        max_steps=10,
```
BenjaminBossan, line 125: I would like to see `max_steps` a configurable argument via the command line, as 10 steps is quite low.

examples/randlora_finetuning/randlora_finetuning.py
```diff
@@ -0,0 +1,223 @@
...
+    # Data collator to dynamically pad the batched examples
+    data_collator = DataCollatorForLanguageModeling(tokenizer, mlm=False)
+
+    # Define training arguments
+    training_args = TrainingArguments(
+        output_dir=output_dir,
+        num_train_epochs=num_epochs,
+        per_device_train_batch_size=batch_size,
+        per_device_eval_batch_size=batch_size,
+        warmup_steps=100,
```
BenjaminBossan, line 123: Instead of hard-coding `warmup_steps`, would it make more sense to make it, say, 10% of `max_steps`?

examples/randlora_finetuning/randlora_finetuning.py
```diff
@@ -0,0 +1,223 @@
...
+        weight_decay=0.01,
+        max_steps=10,
+        logging_dir="./logs",
+        logging_steps=eval_step,
+        save_steps=save_step,
+        save_total_limit=2,
+        push_to_hub=push_to_hub,
+        hub_model_id=hub_model_id,
+        gradient_accumulation_steps=16,
+        fp16=True,
```
BenjaminBossan, line 133: Should this not depend on the `torch_dtype` that was chosen earlier?

examples/randlora_finetuning/README.md
```diff
@@ -0,0 +1,120 @@
...
+There is no additional change needed to your standard PEFT training procedure, simply swap your LoRAConfig for a RandLoraConfig. Note however that RandLora's trainable parameter count is **inversely proportional** to the rank parameter `r`. Lower `r` to increase and increase it to reduce trainable parameters of RandLora.
+
+Run the finetuning script simply by running:
+```bash
+python examples/randlora_finetuning/randlora_finetuning.py --base_model meta-llama/Meta-Llama-3-8B --data_path timdettmers/openassistant-guanaco
+```
+This 👆🏻 by default will load the model in peft set up with RandLora config. Now if you wanna quickly compare it with Lora, all you need to do is to input ` --use_lora` in the command line and reduce `--randlora_alpha` to 2x the rank. So same above example would be 👇🏻;
+
+```bash
+python examples/randlora_finetuning/randlora_finetuning.py --base_model meta-llama/Meta-Llama-3-8B --data_path timdettmers/openassistant-guanaco --use_lora --randlora_alpha
```
BenjaminBossan, line 42: `--randlora_alpha` is missing a value.

examples/randlora_finetuning/randlora_finetuning.py
```diff
@@ -0,0 +1,223 @@
...
+        warmup_steps=100,
+        weight_decay=0.01,
+        max_steps=10,
+        logging_dir="./logs",
+        logging_steps=eval_step,
+        save_steps=save_step,
+        save_total_limit=2,
+        push_to_hub=push_to_hub,
+        hub_model_id=hub_model_id,
+        gradient_accumulation_steps=16,
```
BenjaminBossan, line 132: I'd say either remove this argument or make it configurable.

src/peft/tuners/randlora/model.py
```diff
@@ -129,7 +132,11 @@ def _find_dim(self, config) -> tuple[int, int]:
                 continue
 
             if module_shape != largest_shape:
-                largest_shape = tuple(max(a, b) for a, b in zip(largest_shape, module_shape))
+                # largest_shape = tuple(max(a, b) for a, b in zip(largest_shape, module_shape))
```
BenjaminBossan, line 135: Remove

src/peft/tuners/randlora/model.py
```diff
@@ -129,7 +132,11 @@ def _find_dim(self, config) -> tuple[int, int]:
...
+                largest_shape = (
+                    max(max(module_shape), max(largest_shape)),
+                    max(min(module_shape), min(largest_shape)),
+                )
```
BenjaminBossan, line 139: Could you please explain this change?

BenjaminBossan (dismissed):
Thanks for adding the examples. Overall, they look good, but they still need some "fine-tuning". Please check my comments.

Regarding the notebook, I get an error when trying to open it on GitHub. Other people seem to face the same error, maybe [this fix](https://github.com/orgs/community/discussions/155944#discussioncomment-12898793) works.

> I am still investigating the large memory usage of RandLora I observed when running randlora_finetune.py. This goes against what I have observed outside of the peft library.

Thanks for investigating, please create a PR as soon as you find the underlying issue. Is the example you're comparing it to also using `Trainer`? In my experience, comparing a vanilla PyTorch training loop vs `Trainer` can be quite difficult, as there are so many things going on under the hood.

examples/randlora_finetuning/README.md
PaulAlbert31: I have reformulated to be clearer. I want to carry the message that RandLora can increase performance over LoRA but can (1) be unstable (nan gradients) because of the large scaling parameter (2) be slower to train if the rank of the RandLora bases is reduced a lot.

examples/randlora_finetuning/randlora_finetuning.py
PaulAlbert31: Sorry forgot to remove after debugging. I have now removed the parameter from the training_args

examples/randlora_finetuning/randlora_finetuning.py
PaulAlbert31: Changed to 16//batch_size to ensure minimum size after accumulation is 16 is that suitable ?

PaulAlbert31:
Thanks for the feedback @BenjaminBossan, I have implemented your suggested changes.
I also had a bit of a fight to fix the notebook but I thankfully came out on top in the end.

Here is a command I used in case the issue happens with other contributions:
`jupyter nbconvert --clear-output --inplace qrandlora_finetuning.ipynb`

This fix is suggested in the same thread you linked: https://github.com/orgs/community/discussions/155944#discussioncomment-12856952

Let me know if there is more to improve

examples/randlora_finetuning/randlora_finetuning.py
githubnemo: Because you found that it has to be a 16 accumulation steps to work properly? Maybe it's worthwhile mentioning that as a comment.

githubnemo (commented):
Thanks for the fixes. I'm taking over the review from @BenjaminBossan but there's not much left to do as it seems :)

 Just a few nitpicks from my side.

docs/source/package_reference/randlora.md
```diff
@@ -0,0 +1,45 @@
...
+
+Because reducing the rank of RandLora's random bases will increase their number, RandLora can become slower to train than LoRA for very small ranks where typically, ranks below 4 with result in a large training time increase. This does not affect inference though as the RandLora adapters can be merged into the pretrained weight matrices.
+
+RandLora additionally supports training with sparse, unary random bases (only containing -1, 0 and 1). These bases are as described in [Bingham et al.](https://cs-people.bu.edu/evimaria/cs565/kdd-rp.pdf) and [Ping et al.](https://hastie.su.domains/Papers/Ping/KDD06_rp.pdf) and could theoretically be used to reduce compute needs by performing aggregations instead of matrix multiplications to create the weight update. This is not currently supported. Although it does not currently reduce compute, using sparse random bases in RandLora can reduce overfitting in some cases. For users intersted in using sparse unary bases, the `sparse` option is recommended over the `very_sparse` one that can reduce perfromance.
```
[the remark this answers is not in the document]
PaulAlbert31: Yes good point thanks, ternary is the correct term

PaulAlbert31:
Hi @githubnemo, thanks for your comment and catching the typos.
This new update addresses the comments, let me know if there is more to do !

githubnemo (approved):
This is great, thanks a lot for the thorough documentation, example and integration into the method comparison suite.

PR #2531, opened 2025-05-05
added support for Conv1d for DoRA

EskildAndersen:


BenjaminBossan (changes requested):
Thanks for working on the support of `nn.Conv1d`* for DoRA. Before proceeding, let's add unit tests for this feature. This is as simple as adding a line like this:

`("Conv1d LoRA Conv1d with DoRA", "Conv1d", LoraConfig, {"target_modules": ["conv1d"], "use_dora": True}),`

here:

https://github.com/huggingface/peft/blob/6c054d0ff2c92ac59a4c0ce0d66991b393b6a4a8/tests/test_custom_models.py#L115

*Just to be clear, we already support `transformers.pytorch_utils.Conv1D`, which is a different layer type.

githubnemo:
@EskildAndersen please make sure to use the same ruff version as specified in `setup.py` or otherwise `make style` will potentially differ from the repository style.

either run

```
# in peft directory
pip install .[quality]
```

or

```
pip install ruff~=0.9.2
```

githubnemo (approved):
LGTM. I suggest to not delete the clarifying comment about the handling of Conv layers but otherwise I think that this can be merged. I'll add the suggestion and once the pipeline is green, I'll merge it.

Thanks for the contribution, good work :)

PR #2536, opened 2025-05-08
perf: optimize RandLora dimension calculation with shape caching

srijondasgit:
## Description
This PR optimizes the RandLora dimension calculation by implementing shape caching and improving the dimension finding logic. 

### Changes
- Added module shape caching to avoid redundant shape calculations
- Optimized dimension finding by tracking max dimensions separately

### Performance Impact
- Reduces redundant shape calculations for each module
- More efficient handling of layers with different dimensions
- Better memory usage by avoiding repeated shape computations

### Testing
- All RandLora tests pass

### DIFF
--- a/src/peft/tuners/randlora/model.py
+++ b/src/peft/tuners/randlora/model.py
@@ -114,7 +114,9 @@ class RandLoraModel(BaseTuner):
         peft_config = self._prepare_adapter_config(config, model_config)
         peft_config = _maybe_include_all_linear_layers(peft_config, self.model)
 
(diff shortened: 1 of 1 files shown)


githubnemo:
Hi, thanks for the pull request. Maybe I'm missing something but I fail to see the impact of this change regarding performance. Especially the cache would only be useful if it was shared across `_find_dim` calls, which it isn't right now.

Do you have a test or script to measure the impact of this change? I'd suggest explaining the gravity of the issue first before attempting a fix.

Issue #2516, opened 2025-04-25
Call for contribution: KaSA: Knowledge-Aware Singular-Value Adaptation of Large Language Models


BenjaminBossan:
### Feature request

There is an [existing implementation of KaSA](https://github.com/juyongjiang/KaSA) that can be used as reference for the PEFT implementation. AFAICT, the changes mainly come down to [this new parameter](https://github.com/juyongjiang/KaSA/blob/f85e88c22d0fa4cb8ab2923d7c2bf1bbec152da3/peft/src/peft/tuners/lora/layer.py#L130). For the PEFT integration, the new LoRA variant pattern introduced in #2443 should be leveraged (i.e. use a similar approach to extend vanilla LoRA as is used for DoRA).

### Motivation

[KaSA](https://arxiv.org/abs/2412.06071) is a variation of LoRA that promises improved task adaption compared to vanilla LoRA or PiSSA.

### Your contribution

I will review the PR and guide the contributor(s) if there are questions.

nsbg:
I'm interested in this PR! Is the opportunity still open?

BenjaminBossan:
Thanks @nsbg, the opportunity is indeed still open. Don't hesitate to create an early draft PR to get quick feedback.

BenjaminBossan:
Yes, that would be the idea, translating the code changes from the existing implementation into the LoRA variant approach. I can't guarantee that it will be a 100% smooth ride, perhaps some things will not quite fit. But just open a draft PR and we can discuss how to resolve those impediments if you encounter them.

nsbg:
I opend draft PR 🤗

rusdes:
@BenjaminBossan Hi! This looks interesting. If this issue is stagnant, I would love to quickly close this. I am a grad student and a researcher at a lab at Georgia Tech, and have worked extensively with LoRA. Feel free to assign it to me.

iambogeumkim:
@rusdes hi! I'm currently working on this task. I recently submitted a PR related to this, and since most of the work after that PR is almost finished, another PR will be submitted soon.

robbiebusinessacc:
Hi @iambogeumkim, are you still planning to finish KaSA? Thread's been quiet for a few weeks, so just wanted to check in (no rush at all), only wondering where things stand.

iambogeumkim:
[...]

PR #2543, opened 2025-05-17
Add KaSA implementation to layer.py

nsbg:
cc @BenjaminBossan 

I've added KaSA implementation part to `layer.py` by referring to [the existing KaSA code](https://github.com/juyongjiang/KaSA). This is a very early draft of the PR, and there's a chance I misunderstood something, so a lot might need to be revised/added.

I'm not entirely sure what the sentence "For the PEFT integration, the new LoRA variant pattern introduced in #2443 should be leveraged (i.e. use a similar approach to extend vanilla LoRA as is used for DoRA)" - from https://github.com/huggingface/peft/issues/2516 - means exactly. If you could provide some guidance on this part, it would be really helpful for me.

I'm open to any kind of feedback!

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -33,7 +33,7 @@ from peft.utils.integrations import (
 from peft.utils.other import transpose
 
 from .config import LoraConfig
-
+from .dora import DoraConv2dLayer, DoraConv3dLayer, DoraEmbeddingLayer, DoraLinearLayer, _DoraConvNdLayer
 
 class LoraVariant:
     """
@@ -107,6 +107,9 @@ class LoraLayer(BaseTunerLayer):
         self.cast_input_dtype_enabled: bool = True
         self.lora_variant: dict[str, LoraVariant] = {}
         self.kwargs = kwargs
+        
+        # Diag value
+        self.lora_diag = nn.ParameterDict({})
 
         base_layer = self.get_base_layer()
         if isinstance(base_layer, nn.Linear):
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Thanks @nsbg for working on KaSA. Sorry for the late reply, I was out of office.

Regarding the implementation, I don't think this works correctly yet, as your code only updates the base weight and it also does it unconditionally, when we want users to be able to enable/disable KaSA.

The proper way to implement this is:

1. Add a new argument to [`LoraConfig`](https://github.com/huggingface/peft/blob/5a42bb773f71064de5e8be0a079456c2f7df1925/src/peft/tuners/lora/config.py#L200) called `use_kasa` (`False` by default).
2. Extend the [LoRA variant resolution](https://github.com/huggingface/peft/blob/5a42bb773f71064de5e8be0a079456c2f7df1925/src/peft/tuners/lora/layer.py#L197) to include `use_kasa`.
3. Implement a `KasaVariant` similar to the existing [`DoraVariant`](https://github.com/huggingface/peft/blob/5a42bb773f71064de5e8be0a079456c2f7df1925/src/peft/tuners/lora/variants.py#L28). To see what changes are needed, search for the string `diag` in the [existing KaSA code](https://github.com/juyongjiang/KaSA/blob/f85e88c22d0fa4cb8ab2923d7c2bf1bbec152da3/peft/src/peft/tuners/lora/layer.py) and port it over to the variant.

Later, we'll also need docs and tests, but this can wait until the implementation is done.

iambogeumkim:
Thank you for your review. I'll include your feedback in the next step.

BenjaminBossan:
@nsbg No hurry, just asking if you still intend on working on this.

PR #2550, opened 2025-05-20
LORA/MODEL: Discard `rank_pattern`, `rank_alpha` for `add_weighted_adapter`

Beinsezii:
Oversight from #2512 where the 0 config's rank/alpha pattern was inherited despite the newly adapter needing uniform based on `_check_add_weighted_adapter`'s calcs.

Can `pytest tests -k test_add_weighted_adapter_cat_with_rank_pattern` against `d6948f3192600e641907df0e4ada7c18581a664d` to repro

cc @BenjaminBossan

### DIFF
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -675,6 +675,8 @@ class LoraModel(BaseTuner):
             r=new_rank,
             lora_alpha=new_rank,
             target_modules=new_target_modules,
+            alpha_pattern={},
+            rank_pattern={},
         )
         self.inject_adapter(self.model, adapter_name)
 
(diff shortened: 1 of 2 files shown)


Beinsezii:
I think the failures are all unrelated. Something in the CI changed, so torch itself is failing now.

BenjaminBossan:
> I think the failures are all unrelated. Something in the CI changed, so torch itself is failing now.

Yes, the changes are unrelated, the tests related to your change look good. We're just discussing internally about the approach and will be back with a review soon.

BenjaminBossan (approved):
Thanks for this fix. It should indeed be generally okay to set `alpha_pattern` and `rank_pattern` to `{}` here.

PR #2555, opened 2025-05-28
enable peft examples on XPU

yao-matrix:


BenjaminBossan:
@yao-matrix Thanks for creating the PR. I know it's not ready but wanted to give some quick feedback.

This PR changes a lot of examples to make them device-agnostic. In general, I agree that not hard-coding CUDA is a good idea and we should pay more intention in the future when new examples are added. That said, this is a huge change and basically impossible for us to check if the code still works after the change (examples are not automatically tested, for the proper PEFT code, this is of course different). Often, the changes involve more than just a single line, so a quick glance may not be enough. Therefore, I'm reluctant to accept such a big change.

On top of that, the changes in notebooks are hard to review, there is some diff noise probably caused by automatic linting, and lots of code still will only work for CUDA or XPU but not other devices.

I wonder if instead of making this big change, we can add a section to the docs to explain the steps users need to take to adopt CUDA-specific PEFT code to make it work on XPU or to be device-agnostic in general.

This is just a suggestion and perhaps we can still continue with this PR. But in the future, it's a good idea to open an issue first to discuss the change before opening such big PRs.

yao-matrix:
@BenjaminBossan , yes, that's why i  put at draft, during the enabling, i found it's hard to review since it's huge. It includes notebook cleaning, deprecated API upgrade, and device enabling, which makes reasonable review im-practical. On one hand, we want the users of peft can run examples on Intel XPU out-of-the-box; on the other hand, the changes will be many beyond device expansion, it also includes fixing some out-of-maintain issues because some of the cases cannot run for a long time(from this perspective it's maybe a good thing, since we can take this opportunity to revisit the examples, and to make them work again if not). Do you think it's feasible that we split the thing into 2 stages:
1. to clean all notebook's output cells as what you guys did in this https://github.com/huggingface/notebooks/pull/569 of notebook
2. submit PR per adapter or example, per my experience, the PR will not that much, and be review-friendly. We can also take this opportunity to revisit all the examples too.

I will close this PR, it's definitely in-appropriate. Thx very much.

BenjaminBossan:
> 2\. submit PR per adapter or example, per my experience, the PR will not that much, and be review-friendly. We can also take this opportunity to revisit all the examples too.

Yes, I think this is the way to go. Apart from splitting up the work and making reviews easier, this also allows to focus on the more popular PEFT methods first and get greater leverage. Hopefully, this also allows you to make use of all the work you already put into this PR.

> 1. to clean all notebook's output cells as what you guys did in this [Remove outputs to enable Github rendering notebooks#569](https://github.com/huggingface/notebooks/pull/569) of notebook

Let's not do that, the notebook output is valuable for readers. Also, from a quick check, most notebooks seem to be rendering just fine. Instead, if we find individual notebooks that don't render, let's try to fix those specifically, e.g. by rerunning them with an up to date jupyter package.

Issue #2556, opened 2025-05-29
[LORA Mamba] Raise Error When out_proj or conv1d Are Included in LoRA Config for Mamba-Based Models (e.g. Falcon-H1)


dhiaEddineRhaiem:
### Feature request

Raise an error (or at minimum, a warning) in **PEFT** when users include `out_proj` or `conv1d` as target_modules in the LoraConfig for **Mamba-based models** (e.g. Falcon-H1), as these modules bypass the forward() of their wrapped LoRALinear or LoRAConv1D, and cause silent failures or merge-time crashes.

### Motivation

1. Including `out_proj` or `conv1d` in the LoRA config causes merge_and_unload() to fail with shape mismatch errors, because the LoRA paths are never called during training or inference. Instead, the Falcon-H1 I (and other hybrid and mamba based models) implementation passes the raw weights directly into a fused kernel implemented in a custom autograd function (`**mamba_split_conv1d_scan_combined**`).

Traceback:
```
RuntimeError: output with shape [2048, 1, 4] doesn't match the broadcast shape [2048, 2048, 4]
```
**Code to reproduce:**
```py
from transformers import AutoModelForCausalLM
from peft import PeftModel, LoraConfig, TaskType
import torch

model = AutoModelForCausalLM.from_pretrained("tiiuae/Falcon-H1-1.5B-Deep-Base", torch_dtype=torch.bfloat16, device_map="auto")
model = PeftModel.from_pretrained(model, "kristaller486/sokol-1.5b-01-lora")
model = model.merge_and_unload() # <-- fails
``` 

**Config:**
```py
peft_config = LoraConfig(
    r=32,
    lora_alpha=64,
    lora_dropout=0.1,
    bias="none",
    task_type=TaskType.CAUSAL_LM,
    target_modules=[
        "q_proj", "k_proj", "v_proj", "o_proj",
        "gate_proj", "up_proj", "down_proj",
        "conv1d", "in_proj", "out_proj"  # <- these two break merging
    ]
)
```

**Explanation**
From Falcon-H1’s modeling code, both conv1d and out_proj weights are passed directly to the fused kernel during forward, like this:
```py
out = mamba_split_conv1d_scan_combined(
    projected_states,
    self.conv1d.weight.squeeze(1),
    ...,
    outproj_weight=self.out_proj.weight,
    outproj_bias=self.out_proj.bias,
    ...
)
```

This means:

No .forward() method of the LoRA-wrapped modules is called

[...]

PR #2562, opened 2025-06-02
Prevent applying LoRA to disallowed modules in Mamba-based architectures

dhiaEddineRhaiem:
This PR fixes this [issue](https://github.com/huggingface/peft/issues/2556), preventing applying LoRA to **disallowed modules**  which are `out_proj` and  `conv1d` in **Mamba-based architectures**.
@BenjaminBossan


dhiaEddineRhaiem:
Thanks for the thorough feedback @BenjaminBossan — I completely agree with your suggestions!

Let me add the logic you suggested.

For testing, I can provide Mamba-based models like:
https://huggingface.co/state-spaces/mamba-790m-hf

I'll also include a small unit test in `test_initialization.py`.

Thanks again!

src/peft/tuners/tuners_utils.py
```diff
@@ -398,6 +398,26 @@ def _check_merge_allowed(self):
...
+            peft_config.peft_type in lora_like_types
+            and hasattr(model, "config")
+            and getattr(model.config, "model_type", None) in mamba_model_types
+        ):
+            if any(mod in forbidden_modules for mod in peft_config.target_modules):
+                raise ValueError(
+                    f"[PEFT:{peft_config.peft_type}] target_modules {peft_config.target_modules} contain forbidden modules "
+                    f"for Mamba-based models (model_type={model.config.model_type}): {forbidden_modules}. "
+                    "Please exclude them to avoid compatibility issues."
+                )
```
BenjaminBossan: Let's move this function body into a standalone function (no need for it to be a method), something like `_check_target_modules_mamba` or so. Then `_check_target_module_compatiblity` will just call `_check_target_modules_mamba`. IMO, this improves clarity and makes it easier to extend in the future.

src/peft/tuners/tuners_utils.py
```diff
@@ -398,6 +398,26 @@ def _check_merge_allowed(self):
...
+        mamba_model_types = {"falcon_h1", "mamba", "mamba2", "falcon_mamba"}
+
+        if (
+            peft_config.peft_type in lora_like_types
+            and hasattr(model, "config")
+            and getattr(model.config, "model_type", None) in mamba_model_types
+        ):
+            if any(mod in forbidden_modules for mod in peft_config.target_modules):
+                raise ValueError(
+                    f"[PEFT:{peft_config.peft_type}] target_modules {peft_config.target_modules} contain forbidden modules "
```
BenjaminBossan: I find "forbidden" a bit strong as a wording, how about just "incompatible"?
```suggestion
                    f"[PEFT:{peft_config.peft_type}] target_modules {peft_config.target_modules} contain incompatible modules "
```

src/peft/mapping_func.py
```diff
@@ -120,6 +120,7 @@ def get_peft_model(
 
     if peft_config.is_prompt_learning:
         peft_config = _prepare_prompt_learning_config(peft_config, model_config)
+
```
BenjaminBossan: Let's remove this change.

tests/test_initialization.py
```diff
@@ -1264,6 +1264,23 @@ def test_lora_with_bias_incompatible_arguments(self, extra_kwargs):
...
+            task_type="CAUSAL_LM",
+            inference_mode=False,
+            r=8,
+            lora_alpha=16,
+            lora_dropout=0.1,
+            target_modules=["out_proj", "conv1d"],  # Forbidden modules for Mamba-based models
+        )
+
+        # This should raise the error if your patched PEFT is in effect
+        model = get_peft_model(model, peft_config)
```
BenjaminBossan: With the new changes, this should raise an error, right? So let's do:

```python
msg = ...
```

tests/test_initialization.py
```diff
@@ -1264,6 +1264,23 @@ def test_lora_with_bias_incompatible_arguments(self, extra_kwargs):
         with pytest.raises(ValueError, match=msg):
             LoraConfig(target_modules=["linear"], lora_bias=True, **extra_kwargs)
 
+    def test_lora_forbidden_mamba_modules(self):
```
BenjaminBossan: The used model is still relatively big, so I think it's better not to use it in the normal unit tests. I'd suggest to move the test to `test_gpu_examples.py`, that way it runs only nightly, on more powerful workers than the default github runners.

BenjaminBossan (changes requested):
Thanks for making the changes. There are still some smaller issues but they should be easy to address. Please take a look.

src/peft/tuners/tuners_utils.py
dhiaEddineRhaiem: sure

src/peft/tuners/tuners_utils.py
dhiaEddineRhaiem: sounds cleaner for me also

src/peft/tuners/tuners_utils.py
```diff
@@ -113,7 +113,27 @@ def onload_layer(layer):
             offload_state_dict(safetensors_filename, layer.base_layer._hf_hook.weights_map)
         layer.base_layer._hf_hook.post_forward(layer.base_layer, torch.tensor([]))
 
-
+def _check_lora_target_modules_mamba(peft_config: PeftConfig, model: nn.Module, target_name: str):
+    """
+    Prevent applying LoRA to incompatible modules in specific architectures (e.g., Mamba).
+    """
+    import pdb;pdb.set_trace()
```
BenjaminBossan: :scissors:

src/peft/tuners/tuners_utils.py
```diff
@@ -398,6 +418,12 @@ def _check_merge_allowed(self):
                 + example_code
             )
 
+    def _check_target_module_compatiblity(self, peft_config: PeftConfig, model: nn.Module, target_name: str):
+        """
+        Prevent applying LoRA to incompatible modules in specific architectures (e.g., Mamba).
+        """
+        return _check_lora_target_modules_mamba(peft_config, model, target_name)
```
```suggestion
        _check_lora_target_modules_mamba(peft_config, model, target_name)
```

No need to return here.

tests/test_initialization.py
```diff
@@ -1264,6 +1264,23 @@ def test_lora_with_bias_incompatible_arguments(self, extra_kwargs):
         with pytest.raises(ValueError, match=msg):
             LoraConfig(target_modules=["linear"], lora_bias=True, **extra_kwargs)
 
+    def test_lora_incompatible_mamba_modules(self):
+        # Ensure LoRA raises an error when applying to forbidden modules
+        # ('out_proj', 'conv1d') in Mamba-based architectures like Falcon-Mamba tiny.
+        model = AutoModelForCausalLM.from_pretrained("tiiuae/falcon-mamba-tiny-dev")
+
+        peft_config = LoraConfig(
```
```suggestion
        config = LoraConfig(
```

same name as used below.

tests/test_initialization.py
```diff
@@ -1264,6 +1264,23 @@ def test_lora_with_bias_incompatible_arguments(self, extra_kwargs):
...
+            task_type="CAUSAL_LM",
+            inference_mode=False,
+            r=8,
+            lora_alpha=16,
+            lora_dropout=0.1,
```
BenjaminBossan: Let's remove all of these arguments, as they distract from what's important.

tests/test_initialization.py
```diff
@@ -1264,6 +1264,23 @@ def test_lora_with_bias_incompatible_arguments(self, extra_kwargs):
         with pytest.raises(ValueError, match=msg):
             LoraConfig(target_modules=["linear"], lora_bias=True, **extra_kwargs)
 
+    def test_lora_incompatible_mamba_modules(self):
```
BenjaminBossan, line 1267: As mentioned earlier, let's move the test to `test_gpu_examples.py` (can be standalone test, i.e. no need to add it to one of the existing classes).

tests/test_initialization.py
```diff
@@ -1264,6 +1264,23 @@ def test_lora_with_bias_incompatible_arguments(self, extra_kwargs):
...
+
+        peft_config = LoraConfig(
+            task_type="CAUSAL_LM",
+            inference_mode=False,
+            r=8,
+            lora_alpha=16,
+            lora_dropout=0.1,
+            target_modules=["out_proj", "conv1d"],  # Forbidden modules for Mamba-based models
+        )
+        msg = "incompatible modules for mamba based models"
```
BenjaminBossan: The regex does not quite match, it should be:

```suggestion
        msg = "incompatible modules for Mamba-based models"
```

src/peft/tuners/tuners_utils.py
```diff
@@ -496,6 +522,7 @@ def inject_adapter(
                 unmatched_modules.append(key)
             else:
                 self.targeted_module_names.append(key)
+                self._check_target_module_compatiblity(peft_config, model, key)
```
BenjaminBossan: I missed this in the last review. For this to properly work, move this one line down by one line pass `target_name` instead of `key`, as `key` is the fully qualified name, like `'backbone.layers.0.mixer.conv1d'`.

src/peft/tuners/tuners_utils.py
```diff
@@ -113,7 +113,27 @@ def onload_layer(layer):
...
+    lora_like_types = {"LORA", "ADALORA", "XLORA", "RANDLORA"}
+    incompatible_modules = {"out_proj", "conv1d"}
+    mamba_model_types = {"falcon_h1", "mamba", "mamba2", "falcon_mamba"}
+
+    if (
+        peft_config.peft_type in lora_like_types
+        and hasattr(model, "config")
+        and getattr(model.config, "model_type", None) in mamba_model_types
+    ):
+        if any(mod in incompatible_modules for mod in peft_config.target_modules):
```
BenjaminBossan: Here, we should use `target_name` and not `peft_config.target_modules`:
```suggestion
        if target_name in incompatible_modules:
```

This line can actually be added to the `if` above, i.e. just have one `if` condition, not 2.

BenjaminBossan (changes requested):
Thanks again for the quick updates. I still discovered a few small issues, but nothing major. Please take a look.

Once you finish your changes, please remember to call `make style` for the linter.

tests/test_initialization.py
dhiaEddineRhaiem: I thought we could keep it there as the model falcon mamba tiny is extremely small model 

let me jnow what do you think

BenjaminBossan: Yes, generally it is small, but it's still 1GB. As CI is already flaky because we hit rate limits, and as GH workers have relatively little memory, I would prefer to move this to the larger workers or use a smaller model.

dhiaEddineRhaiem: The previous model was falcon h1 0.5B which is effectively 1GB

But the current one is falcon mamba tiny dev is a 8.77M paramter model , so it should be good right ?

BenjaminBossan: Oh I missed that. Yes, this size is absolutely fine.

tests/test_initialization.py
```diff
@@ -1264,6 +1264,19 @@ def test_lora_with_bias_incompatible_arguments(self, extra_kwargs):
...
+    def test_lora_incompatible_mamba_modules(self):
+        # Ensure LoRA raises an error when applying to forbidden modules
+        # ('out_proj', 'conv1d') in Mamba-based architectures like Falcon-Mamba tiny.
+        model = AutoModelForCausalLM.from_pretrained("tiiuae/falcon-mamba-tiny-dev")
+
+        config = LoraConfig(
+            task_type="CAUSAL_LM",
+            target_modules=["out_proj", "conv1d"],  # Forbidden modules for Mamba-based models
+        )
+        msg = "incompatible modules for Mamba-based models"
```
BenjaminBossan: Tests are failing because the regex doesn't match the error message. This should work:
```suggestion
        msg = "is incompatible with Mamba-based models"
```

dhiaEddineRhaiem: Great, Thanks for your help !

commited the suggested fix

BenjaminBossan (approved):
Thanks for adding this checks to ensure `target_modules` compatibility, LGTM. Issues with MacOS tests are unrelated.

PR #2564, opened 2025-06-02
Test if -n1 solves memory issue no MacOS

githubnemo:
ARM64 runners only have 7GB RAM which was not enough previously. Maybe running only one test thread at a time works, though?


githubnemo:
Even with -n1 it seems that we're running into OOM errors. Python 3.11 and 3.10 experience a segfault pretty much immediately after starting the test. I think this needs a lot more work to make it happen.

Issue #2549, opened 2025-05-20
DoRA RuntimeError:  shape is invalid when Conv2d(groups=2)


victolee0:
### System Info

peft==0.15.2
torch==2.2.2

### Who can help?

@BenjaminBossan @sayakpaul 

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

```py
from peft import LoraConfig, get_peft_model
import torch

class MyModel(torch.nn.Module):
    def __init__(self):
        super(MyModel, self).__init__()
        self.conv = torch.nn.Conv2d(96, 192, (3, 3), stride=(2, 2), padding=(1, 1), bias=False, groups=2)
        self.fc = torch.nn.Linear(192, 10)

    def forward(self, x):
        x = self.conv(x)
        x = self.fc(x)
        return x


model = MyModel()

lora_config = LoraConfig(
    r=16,
    target_modules=["conv"],
    use_dora=True
)

model = get_peft_model(model, lora_config)
```

### Expected behavior

RuntimeError: shape '[192, 48, 3, 3]' is invalid for input of size 165888

165888 / (192 * 48 * 3 * 3) = 2

when `groups=2` got an error
if `groups=1` works fine

BenjaminBossan:
Thanks for the report. The snippet is not quite complete, could you please show the (shape of the) input you feed into the model? Also, which PEFT version are you using?

SP1029:
@BenjaminBossan I think the snippet by @victolee0 is complete. I tried reproducing the error and was able to successfully reproduce on google colab. Following is my understanding of why the error is occurring in `peft==0.15.2`.

The PR #2403 fixes the `groups` argument which was being ignored in `layers.py` for LoRA. This is not present in `peft 0.15.2`, which likely is the source of this issue. Before this PR the `lora_B` had a incorrect size, but for lora it was working fine since `lora_B @ lora_A` is never directly added with the `weight` of the base layer. This is because during forward pass both of these terms are first multiplied with the input and then added. Also `merge` operation is not supported.

Before PR: 
`self.lora_B[adapter_name] = conv_layer(r, self.out_features, out_kernel, out_stride, bias=lora_bias)`

After PR: 
```
self.lora_B[adapter_name] = conv_layer(Add commentMore actions
        r, self.out_features // base_layer.groups, out_kernel, out_stride, bias=lora_bias
    )
[...]

PR #2567, opened 2025-06-03
FIX for #2549 - modify lora_B definition for conv layers with groups

SP1029:
Issue #2549 is about error in initializing the model with peft approach 
of dora and lora and errors in the forward pass for conv layers with
`groups>1`.

This fix is one possible approach to resolve the errors by including 
`groups` as part of the definition of `lora_B` parameter. While this 
fix resolves the error, it introduces a restriction of `in_features` and 
`out_features` to be divisible by rank `r`.

### DIFF
--- a/tests/test_initialization.py
+++ b/tests/test_initialization.py
@@ -1280,6 +1280,7 @@ class TestLoraInitialization:
     def get_model_conv2d_groups(self):
         class ModelConv2DGroups(nn.Module):
             """For testing when groups argument is used in conv layer"""
+
             def __init__(self):
                 super().__init__()
                 self.conv2d = nn.Conv2d(16, 32, 3, padding=1, groups=2)
@@ -1303,25 +1304,31 @@ class TestLoraInitialization:
         return ModelConv2DGroups().eval().to(self.torch_device)
 
     @pytest.mark.parametrize(
-        "test_name, config_cls, config_kwargs",
+        "config_cls, config_kwargs",
         [
-            ("lora with rank divisible by groups", LoraConfig, {"r": 8, "target_modules": ["conv2d"]}),
-            ("lora with rank equal to groups", LoraConfig, {"r": 2, "target_modules": ["conv2d"]}),
-            ("lora with rank not divisible by groups", LoraConfig, {"r": 1, "target_modules": ["conv2d"]}),
-            (
-                "dora with rank divisible by groups",
+            pytest.param(LoraConfig, {"r": 8, "target_modules": ["conv2d"]}, id="lora with rank divisible by groups"),
+            pytest.param(LoraConfig, {"r": 2, "target_modules": ["conv2d"]}, id="lora with rank equal to groups"),
+            pytest.param(
+                LoraConfig, {"r": 1, "target_modules": ["conv2d"]}, id="lora with rank not divisible by groups"
+            ),
+            pytest.param(
                 LoraConfig,
                 {"r": 8, "target_modules": ["conv2d"], "use_dora": True},
+                id="dora with rank divisible by groups",
             ),
-            ("dora with rank equal to groups", LoraConfig, {"r": 2, "target_modules": ["conv2d"], "use_dora": True}),
-            (
-                "dora with rank not divisible by groups",
+            pytest.param(
+                LoraConfig,
+                {"r": 2, "target_modules": ["conv2d"], "use_dora": True},
+                id="dora with rank equal to groups",
+            ),
+            pytest.param(
                 LoraConfig,
                 {"r": 1, "target_modules": ["conv2d"], "use_dora": True},
+                id="dora with rank not divisible by groups",
             ),
         ],
     )
-    def test_error_raised_if_rank_not_divisible_by_groups(self, test_name, config_cls, config_kwargs):
+    def test_error_raised_if_rank_not_divisible_by_groups(self, config_cls, config_kwargs):
         # This test checks if error is raised when rank is not divisible by groups for conv layer since
         # currently, support is limited to conv layers where the rank is divisible by groups in lora and dora
         base_model = self.get_model_conv2d_groups()


src/peft/tuners/lora/dora.py
```diff
@@ -145,7 +146,8 @@ def forward(self, x, *, lora_A, lora_B, scaling, base_layer, base_result=None):
         output.
         """
         weight = base_layer.weight
-        lora_weight = torch.mm(lora_B.weight.flatten(start_dim=1), lora_A.weight.flatten(start_dim=1))
+        r = lora_A.shape[0]
```
BenjaminBossan:
```suggestion
        r = lora_A.weight.shape[0]
```

BenjaminBossan:
I wanted to add a small test case that would fail with the current implementation. This is what I came up with:

```diff
diff --git a/tests/test_custom_models.py b/tests/test_custom_models.py
index 3b1283a7..1ff2a5f4 100644
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -120,6 +120,8 @@ TEST_CASES = [
     ("Conv2d 2 LoRA with DoRA", "Conv2d", LoraConfig, {"target_modules": ["conv2d", "lin0"], "use_dora": True}),
     ("Conv2d Groups LoRA", "Conv2dGroups", LoraConfig, {"target_modules": ["conv2d"]}),
     ("Conv2d Groups LoRA with DoRA", "Conv2dGroups", LoraConfig, {"target_modules": ["conv2d"], "use_dora": True}),
+    ("Conv2d Groups2 LoRA", "Conv2dGroups2", LoraConfig, {"target_modules": ["conv2d"], "use_dora": False}),
+    ("Conv2d Groups2 LoRA with DoRA", "Conv2dGroups2", LoraConfig, {"target_modules": ["conv2d"], "use_dora": True}),
     ("Conv3d 1 LoRA", "Conv3d", LoraConfig, {"target_modules": ["conv3d"]}),
     ("Conv3d 2 LoRA", "Conv3d", LoraConfig, {"target_modules": ["conv3d", "lin0"]}),
     ("Conv3d 1 LoRA with DoRA", "Conv3d", LoraConfig, {"target_modules": ["conv3d"], "use_dora": True}),
@@ -984,6 +986,28 @@ class ModelConv2DGroups(nn.Module):
         return X
 
 
+class ModelConv2DGroups2(nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.conv2d = nn.Conv2d(16, 32, 3, padding=1, groups=2)
+        self.relu = nn.ReLU()
+        self.flat = nn.Flatten()
+        self.lin0 = nn.Linear(12800, 2)
+        self.sm = nn.LogSoftmax(dim=-1)
+        self.dtype = torch.float
+
+    def forward(self, X):
+        # Note: needs a different input shape, thus ignore original input
+        X = torch.zeros([9, 16, 20, 20]).to(self.conv2d.weight.device)
+        X = X.to(self.dtype)
+        X = self.conv2d(X)
+        X = self.relu(X)
+        X = self.flat(X)
+        X = self.lin0(X)
+        X = self.sm(X)
+        return X
+
+
 class ModelConv3D(nn.Module):
     def __init__(self):
         super().__init__()
@@ -1054,6 +1078,9 @@ class MockTransformerWrapper:
         if model_id == "Conv2dGroups":
             return ModelConv2DGroups().to(torch_dtype)
 
+        if model_id == "Conv2dGroups2":
+            return ModelConv2DGroups2().to(torch_dtype)
+
         if model_id == "Conv3d":
             return ModelConv3D().to(torch_dtype)
 
@@ -1126,7 +1153,7 @@ class PeftCustomModelTester(unittest.TestCase, PeftCommonTester):
     @parameterized.expand(TEST_CASES)
     def test_merge_layers(self, test_name, model_id, config_cls, config_kwargs):
         # https://github.com/huggingface/peft/pull/2403
-        if model_id in ["Conv2dGroups"]:
+        if model_id in ["Conv2dGroups", "Conv2dGroups2"]:
             pytest.skip(
                 f"Skipping test for {model_id} as merging is not supported. (See https://github.com/huggingface/peft/pull/2403 for details)"
             )
@@ -1149,7 +1176,7 @@ class PeftCustomModelTester(unittest.TestCase, PeftCommonTester):
     @parameterized.expand(TEST_CASES)
     def test_merge_layers_fp16(self, test_name, model_id, config_cls, config_kwargs):
         # https://github.com/huggingface/peft/pull/2403
-        if model_id in ["Conv2dGroups"]:
+        if model_id in ["Conv2dGroups", "Conv2dGroups2"]:
             pytest.skip(
                 f"Skipping test for {model_id} as merging is not supported. (See https://github.com/huggingface/peft/pull/2403 for details)"
             )
@@ -1164,7 +1191,7 @@ class PeftCustomModelTester(unittest.TestCase, PeftCommonTester):
     @parameterized.expand(TEST_CASES)
     def test_merge_layers_is_idempotent(self, test_name, model_id, config_cls, config_kwargs):
         # https://github.com/huggingface/peft/pull/2403
-        if model_id in ["Conv2dGroups"]:
+        if model_id in ["Conv2dGroups", "Conv2dGroups2"]:
             pytest.skip(
                 f"Skipping test for {model_id} as merging is not supported. (See https://github.com/huggingface/peft/pull/2403 for details)"
             )
@@ -1180,7 +1207,7 @@ class PeftCustomModelTester(unittest.TestCase, PeftCommonTester):
     @parameterized.expand(TEST_CASES)
     def test_safe_merge(self, test_name, model_id, config_cls, config_kwargs):
         # https://github.com/huggingface/peft/pull/2403
-        if model_id in ["Conv2dGroups"]:
+        if model_id in ["Conv2dGroups", "Conv2dGroups2"]:
             pytest.skip(
                 f"Skipping test for {model_id} as merging is not supported. (See https://github.com/huggingface/peft/pull/2403 for details)"
             )
@@ -1274,7 +1301,7 @@ class PeftCustomModelTester(unittest.TestCase, PeftCommonTester):
         # check that none of this raises an error
         model(**X)
 
-        if model_id in ["Conv2dGroups"]:
+        if model_id in ["Conv2dGroups", "Conv2dGroups2"]:
             # this model does not support merging
             return
 
@@ -1316,7 +1343,7 @@ class PeftCustomModelTester(unittest.TestCase, PeftCommonTester):
         # check that none of this raises an error
         model(**X)
 
-        if model_id in ["Conv2dGroups"]:
+        if model_id in ["Conv2dGroups", "Conv2dGroups2"]:
             # this model does not support merging
             return
 
@@ -1357,7 +1384,7 @@ class PeftCustomModelTester(unittest.TestCase, PeftCommonTester):
         # check that none of this raises an error
         model(**X)
 
-        if model_id in ["Conv2dGroups"]:
+        if model_id in ["Conv2dGroups", "Conv2dGroups2"]:
             # this model does not support merging
             return
 
@@ -1398,7 +1425,7 @@ class PeftCustomModelTester(unittest.TestCase, PeftCommonTester):
         # check that none of this raises an error
         model(**X)
 
-        if model_id in ["Conv2dGroups"]:
+        if model_id in ["Conv2dGroups", "Conv2dGroups2"]:
             # this model does not support merging
             return
 
@@ -1569,7 +1596,7 @@ class PeftCustomModelTester(unittest.TestCase, PeftCommonTester):
     @parameterized.expand(TEST_CASES)
     def test_disable_adapters_with_merging(self, test_name, model_id, config_cls, config_kwargs):
         # https://github.com/huggingface/peft/pull/2403
-        if model_id in ["Conv2dGroups"]:
+        if model_id in ["Conv2dGroups", "Conv2dGroups2"]:
             pytest.skip(
                 f"Skipping test for {model_id} as merging is not supported. (See https://github.com/huggingface/peft/pull/2403 for details)"
             )
```

As expected, this fails with the current implementation. When switching to the branch of this PR, the new example works (except a handful of tests that fail, I haven't checked further yet). However, also as expected, now the old `groups` test fails because of `in_channels must be divisible by groups`.

src/peft/tuners/lora/dora.py
SP1029: Thank you for pointing out, updated.

SP1029:
Thank you for helping with the test case.
There could be a possible alternate approach to tackle the error.

https://github.com/microsoft/LoRA/blob/c4593f060e6a368d7bb5af5273b8e42810cdef90/loralib/layers.py#L290-L297

If we look at the `conv` layer implementation in `loralib`. I think the main idea here is to view the weights of the conv layer as a matrix of dimensions `(in_channels * kernel_size)` `x` `((out_channels * kernel_size)  // groups)` and apply lora on this with `lora_A`, `lora_B` with rank = `r * kernel_size`. Finally, the updated weight matrix is re-shaped, and normal convolution is applied.

If a similar implementation is used in this case, then this does not have the restriction of `in_features` and `out_features` to be divisible by `rank`. Also, it would enable `merge_weights` for `conv` layers which are currently not supported. But mostly, there would have been reasons for choosing to apply lora by using two conv layers, which might have to be compared with, and I am not completely aware of. 

What would be your suggestion on this?

SP1029:
The input has been changed from `torch.zeros([9, 16, 20, 20])` to non-zero so that test-cases requiring parameters to be updated also pass.

SP1029:
The architecture of `ModelConv2DGroups` has been modified such that `groups` divides `rank`. This is done by up-projecting the input `X` of size 90 and modifying the `conv` layer to satisfy the constrain.

SP1029:
After the changes, `pytest tests/test_custom_models.py -k groups -v` on this branch passes the test-cases.
For the `main` branch, there are test-cases failing with `ModelConv2DGroups2` and the new `ModelConv2DGroups`.
So the new `ModelConv2DGroups` becomes a case with `groups=in,out` while `ModelConv2DGroups2` is more generic.

@BenjaminBossan I think the basic implementation-side for this approach has been completed, so could you please review it.

SP1029:
@BenjaminBossan Could you please review the implementation for this approach.

The `groups` tests pass on the new branch.
Some `groups` test with new `ModelConv2DGroups` (`rank=groups`) and `ModelConv2DGroups` (`rank%groups==0`) fail successfully for the main branch.

SP1029:
@BenjaminBossan Gentle Reminder

src/peft/tuners/lora/layer.py
```diff
@@ -1098,7 +1098,7 @@ def update_layer(
         out_kernel = out_stride = (1,) * (self._kernel_dim - 2)
         self.lora_A[adapter_name] = conv_layer(self.in_features, r, kernel_size, stride, padding, bias=False)
         self.lora_B[adapter_name] = conv_layer(
-            r, self.out_features // base_layer.groups, out_kernel, out_stride, bias=lora_bias
+            r, self.out_features, out_kernel, out_stride, groups=base_layer.groups, bias=lora_bias
```
BenjaminBossan, line 1108: As discussed, this introduces the requirement that `in_channels must be divisible by groups`. If this is not true, users could potentially get confused by the error message, since they might not even be aware that the base model uses `groups`. Therefore, I would suggest that we first check if `in_channels % groups == 0` and if it isn't, raise a helpful error message.

The message could be something along the lines of:

> `Targeting a {conv_layer.__name__} with groups={base_layer.groups} and rank {r}. However, the rank must be divisible by groups. Either choose a different rank or don't target this specific layer.`

Let's also add a test for that error in `test_initialization.py`.

tests/test_custom_models.py
```diff
@@ -964,16 +966,43 @@ def forward(self, X):
 class ModelConv2DGroups(nn.Module):
     def __init__(self):
         super().__init__()
-        self.conv2d = nn.Conv2d(5, 5, 3, groups=5)
+        # groups is set as 8 since default r=8
+        # hence to make r divisible by groups
+        self.conv2d = nn.Conv2d(16, 16, 3, groups=8)
```
BenjaminBossan, line 972: Could you please re-order the assignment of modules here so that they fit the order they're being used in `forward`? This is just for clarity.

BenjaminBossan (changes requested):
Thanks a lot for the updates. Overall, this looks quite good and I just have some smaller comments, please check.

I think that even though we have the limitation with regard to the rank being divisible, this implementation is still more flexible that what we currently and thus more useful for PEFT users. But I'd like to also hear your opinion too @gslama12.

src/peft/tuners/lora/layer.py
SP1029: Thanks a lot for pointing out, had this in mind but missed while implementing.

tests/test_custom_models.py
SP1029: Yes sure.

SP1029: Done. Is it fine now?

src/peft/tuners/lora/layer.py
SP1029: I have added a `ValueError` to be raised under the circumstance of `rank` not being divisible by `groups`. This is because original definition of conv layer ensures that `groups` divides `in_channels` and `out_channels`. We introduce additional constrain of `rank` being one of these in the definition of `lora_B`.

I have added the test for verifying if error is being raised in `test_lora_variants.py` since the `test_custom_models.py` included several tests (parameter-update/quantisation etc) while we only wanted to check for this error specific to lora and dora for conv layers. But I was a bit unsure if this would be best place to have the test, @BenjaminBossan in case you feel any other place would be better please suggest.

gslama12:
I have added a brief comment in #2549 . Unfortunately I do not have time for this rn but I will try look at this solution in detail by the end of the week.

@SP1029 Does the current implementation support the merging of the adapter for conv layers that use groups?

SP1029:
The current implementation (and this fix) does not support merging.

gslama12:
Ok, in this case we may even want to think about maybe combining both solutions if that is feasible somehow. 🤔

src/peft/tuners/lora/layer.py
BenjaminBossan: Nicely done. However, let's move these tests to `test_initialization.py`, which is where the tests live that check the method init. Ideally, add the tests to [this class](https://github.com/huggingface/peft/blob/d6dbbc919576d92c31d00994e41cb503ff132c94/tests/test_initialization.py#L70), at the end after the existing tests.

tests/test_lora_variants.py
```diff
@@ -80,6 +104,16 @@ def forward(self, input_ids, dummy_image_input):
 ]
 
 
+CONV_GROUPS_TEST_CASES = [
```
BenjaminBossan: I think for this case, we don't need to assign `CONV_GROUPS_TEST_CASES`, just put the settings below directly inside of `pytest.mark.parametrize`.

tests/test_lora_variants.py
```diff
@@ -80,6 +104,16 @@ def forward(self, input_ids, dummy_image_input):
...
+    ("lora with rank divisible by groups", LoraConfig, {"r": 8, "target_modules": ["conv2d"]}),
```
BenjaminBossan: The `test_name` is not used by the test, it can be safely removed. Since it does contain useful information, how about turning it into a comment instead?

tests/test_lora_variants.py
SP1029: Ok

tests/test_lora_variants.py
SP1029: @BenjaminBossan, Yes, it is true that the `test_name` is not being explicitly used by the test function, but by viewing how the other tests using `test_name` have been written, I was having a query regarding its presence. 

In the `tests` directory, the `test_name` argument has previously been used in `test_custom_models.py,` `test_common.py,` and `test_regression.py.` `test_custom_models.py` and `test_common.py` also seem to not have been explicitly used by the test function, but it seemed that their purpose was to make it easier to understand the test, which fails when `pytest` is run.

For example, if the `ValueError` code for checking `rank` divisible by `groups` is commented, the following is seen on `pytest`:
```
FAILED tests/test_lora_variants.py::TestConvGroups::test_error_raised_if_rank_not_divisible_by_groups[lora with rank not divisible by groups-LoraConfig-config_kwargs2] - AssertionError: Regex pattern did not match.
FAILED tests/test_lora_variants.py::TestConvGroups::test_error_raised_if_rank_not_divisible_by_groups[dora with rank not divisible by groups-LoraConfig-config_kwargs5] - AssertionError: Regex pattern did not match.
```
which contains the `test_name` making it easier in the future for someone not fully familiar with the test to gain more insight on test-failing without looking at the code. While the test can be figured out even in the absence of `test_name,` maybe its presence acts as a `comment.` 

This is what felt to be a use of `test_name,` but if we feel that a comment would be better, then we can convert it into a comment.

tests/test_lora_variants.py
SP1029: Moved the test and updated.

src/peft/tuners/lora/layer.py
SP1029: Ok. The test has been moved.

SP1029:
I have a small query in the existing (before fix) implementation @gslama12 . It works for the case of `groups=out_channels.` In this case, as in the previous test cases being passed, the forward pass is successful. But there is a broadcast operation (copying along one dimension) happening in the first term along the `channel` dimension when `lora_B(lora_A(X)) + base_layer(X)` is done, which might make all channels of `lora_B(lora_A(X))` same. Just wanted to confirm that is this a part of previous idea to support `groups`?

> Ok, in this case we may even want to think about maybe combining both solutions if that is feasible somehow. 🤔

One simple way would be to add `if` block for the case of `groups=out_channels`.

tests/test_initialization.py
```diff
@@ -1277,6 +1277,72 @@ def test_lora_incompatible_mamba_modules(self):
...
+                {"r": 8, "target_modules": ["conv2d"], "use_dora": True},
+            ),
+            ("dora with rank equal to groups", LoraConfig, {"r": 2, "target_modules": ["conv2d"], "use_dora": True}),
+            (
+                "dora with rank not divisible by groups",
+                LoraConfig,
+                {"r": 1, "target_modules": ["conv2d"], "use_dora": True},
+            ),
+        ],
+    )
```
BenjaminBossan, line 1330: > In the tests directory, the test_name argument has previously been used in test_custom_models.py, test_common.py, and test_regression.py. test_custom_models.py and test_common.py also seem to not have been explicitly used by the test function, but it seemed that their purpose was to make it easier to understand the test, which fails when pytest is run.

Previously, we were using unittest style tests but now we're moving everything to pytest style. This means that we can use `@pytest.mark.paramtrize` which has a [dedicated option to name tests](https://docs.pytest.org/en/stable/example/parametrize.html#different-options-for-test-ids) and does not require this awkward workaround. Just as a quick example, here are 2 ways:

```python
@pytest.mark.parametrize("x, y", [
    (1, 2),
    (3, 4),
], ids=["this works", "this does not work"])
def test_with_id(x, y):
    assert x + y < 5

@pytest.mark.parametrize("x, y", [
    pytest.param(1, 2, id="this works"),
    pytest.param(3, 4, id="this does not work"),
])
def test_with_id2(x, y):
    assert x + y < 5
```

SP1029: Thank you for rectifying. Have updated it according to the second way.

SP1029:
So how should we proceed, do we want to combine both approaches or no?

gslama12:
I ran some examples and noticed that the fix might only be needed for `dora`. It seems like `lora` works fine for all settings of the `groups` argument if we use `self.lora_B[adapter_name] = conv_layer(r, self.out_features, out_kernel, out_stride, bias=lora_bias)`, which was actually the base implementation before the changes that I proposed. @SP1029 Can you maybe confirm that this is the case for your usage too? If so we may just want to change the  `use_dora=True` case.

@SP1029 As for the broadcast, you are right. There probably should not be a broadcasting operation when combining the base layer result with the adapter. As my approach actually just makes the `lora_B` adapter smaller to suit the smaller weight of the grouped base convolution layer, this side-effect occurs.

@BenjaminBossan I think you are right - combining the two approaches might not be possible without creating some really confusing constraints.

I would opt for your proposed solution for now and maybe only implement it for dora if that is possible. Maybe we can then keep this topic as an open issue.

SP1029:
Yes @gslama12, the original `lora` should mostly work. If the change is only to be kept for `dora` then there would be minor updates in the test-cases. 
Just one small thing is that this might mean slightly different handling for `lora` and `dora` for `groups` in terms of params/layer-definitions. If this concern seems small then necessary updates can be made.

BenjaminBossan:
> I would opt for your proposed solution for now and maybe only implement it for dora if that is possible. Maybe we can then keep this topic as an open issue.

> Just one small thing is that this might mean slightly different handling for `lora` and `dora` for `groups` in terms of params/layer-definitions. If this concern seems small then necessary updates can be made.

I think it would not be ideal to have two conceptually different approaches when dealing with groups depending on whether DoRA is used or not. This would be unexpected behavior, so I would strongly prefer a single solution. As I wrote earlier, I don't think there is a single "correct" solution, as this goes beyond the LoRA paper, so we should take the most useful approach. As the implementation from this PR is more flexible overall, I have a slight preference for it. If we can agree on that, I would suggest to go with it for LoRA and DoRA.

SP1029:
> prefer a single solution.

I would also be slightly inclined towards this (single).

BenjaminBossan (approved):
Thanks for this PR and everyone for the fruitful debate. The changes LGTM and from my point of view, we can merge it.

LMK @SP1029 and @gslama12 if you agree and I'll go ahead.

BenjaminBossan:
Okay, I think we have discussed this PR long enough, IMO it's time to merge it. Thanks to everyone involved in the implementations and the discussion.

SP1029:
Thank you everyone.

Issue #1510, opened 2024-02-26
Implementing GLora


aravindhv10:
### Feature request

Implementing GLORA in PEFT as a superset of LORA, reference:

https://arxiv.org/abs/2306.07967

### Motivation

This is a superset of regular LORA and performs better in some circumstances.

### Your contribution

I can help with implementation if given guidance.

BenjaminBossan:
Duplicate of #780. Do you have experience with GLoRA, did it work well for your use cases?

Also, note that there is a PEFT fork that contains a [GLoRA implementation](https://github.com/Arnav0400/peft/blob/0bcbe4549b54dbcc82c9354e7ba00b386a47b3eb/src/peft/tuners/glora.py) but it the fork is a bit out of date by now. Maybe you could still try to give it a go and see if it leads to improvements on your use cases.

aravindhv10:
I had used GLora with mixed results in Kohya SS SDXL dreambooth scripts. The results are mostly qualitative and some models were better than Lora. Thanks for the pointers! will have a look.

Issue #780, opened 2023-08-03
Add new method: GLoRA


HsunGong:
### Feature request

See: https://github.com/Arnav0400/ViT-Slim/tree/master/GLoRA

### Motivation

See: https://github.com/Arnav0400/ViT-Slim/tree/master/GLoRA
A new method for PEFT

### Your contribution

Could be

kumar-devesh:
Hi @HsunGong, will you be working on contributing GLoRA? I would like to work on it as well!

kumar-devesh:
Hello @younesbelkada @BenjaminBossan, I would like to work towards contributing `GLoRA`, should be a nice addition to the library

kumar-devesh:
@younesbelkada @BenjaminBossan I would like to work towards adding this to peft.

githubnemo:
This was adressed in https://github.com/huggingface/peft/pull/2568 but wasn't completed. To [quote](https://github.com/huggingface/peft/pull/2568#issuecomment-3406489263) @not-lain:

>  imo the only good thing to retain from this pr [#2568] is the config file, other scripts are too messy and unusable.

Therefore this is now open again and contributions are welcome. I think it's a good idea to take the config file as suggested and start from there. Some other things from that PR can probably be salvaged as well.

kumar-devesh:
Hey @githubnemo , I would like to take this up this weekend. Whats the closet to GLoRA peft method that i can use for reference?

githubnemo:
*not stale*

PR #2568, opened 2025-06-03
add GLORA 

not-lain:
based on https://github.com/huggingface/peft/compare/main...Arnav0400:peft:main 
and #1510 as well as #780


src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,185 @@
...
+        ]
+
+    def make_param(self, shape, config=None):
+        if "LoRA" in config:
+            out_feature = shape[0]
+            in_feature = shape[1]
+            try:
+                rank = int(config.split("_")[1])
+            except Exception:
+                rank = 4
```
githubnemo, line 83: This should simply raise an exception since it indicates an user error. Assuming a default will lead to unexpected behavior.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,185 @@
...
+    def merge(self):
+        if self.merged:
+            return
+
+        # If eval_config is not set, use a random config for merging
+        if self.eval_config is None:
+            warnings.warn("eval_config not set for GLora layer, using a random config for merge.")
+            path_config = random.choice(self.configs)
+        else:
+            path_config = self.eval_config
```
githubnemo: I think this should raise an error explaining what to do to fix it instead of simply assuming a random config which might or might not work.

Ideally there's a test in `tests/test_glora.py` to test this behavior.

tests/test_glora.py
```diff
@@ -0,0 +1,221 @@
...
+            self.base_model = AutoModelForCausalLM.from_pretrained(self.model_id)
+        except Exception:
+            print("Failed to load HF tiny model, using SimpleTransformer for tests.")
+            self.base_model = SimpleTransformer()
+            self.tokenizer = DummyTokenizer()
+
+    def tearDown(self):
+        gc.collect()
+        torch.cuda.empty_cache()
+        gc.collect()
```
githubnemo, line 77: We have a test file explicitly for GPU/accelerator tests - all other tests assume to run on CPU, therefore this is not necessary.

tests/test_glora.py
```diff
@@ -0,0 +1,221 @@
...
+        with torch.no_grad():
+            merged_model.eval()
+            _ = merged_model(dummy_input)
+
+    @unittest.skipIf(
+        not torch.cuda.is_available()
+        or not hasattr(torch.cuda, "is_bf16_supported")
+        or not torch.cuda.is_bf16_supported(),
+        "BF16 not supported or no CUDA",
+    )
```
githubnemo, line 187: To also support XPU/Intel:

```
from accelerate.utils.imports import is_xpu_available
from accelerate.utils.imports import is_bf16_available

[...]

    @unittest.skipIf(
        not torch.cuda.is_available()
        or not is_xpu_available()
        or not is_bf16_available()
        "BF16 not supported or no CUDA/XPU",
    )

File: tests/test_glora.py [hunk 5]
```diff
@@ -0,0 +1,221 @@
...
+    eos_token = 0
+
+    def __call__(self, *args, **kwargs):
+        return {"input_ids": torch.randint(0, 100, (2, 5))}
+
+    def batch_decode(self, *args, **kwargs):
+        return ["decoded text"]
+
+
+class GLORATester(unittest.TestCase):
```
githubnemo, line 63: Great that you added tests! We're in the process of migrating existing tests to pytest. It's totally OK to keep these as is but if you want to convert these already to pytest I wouldn't complain :)

src/peft/utils/constants.py
```diff
@@ -129,6 +129,10 @@ def starcoder_model_postprocess_past_key_value(past_key_values):
     "qwen3": ["q_proj", "v_proj"],
 }
 
+TRANSFORMERS_MODELS_TO_GLORA_TARGET_MODULES_MAPPING = (
+    TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING  # need to check this later
+)
```
githubnemo: the paper writes

> We apply GLoRA for LLMs by solely tuning the attention layers

while this is probably not meant as advice it is - as far as I understand - the only practice for which we have data. Since this is very similar for LoRA I'm fine with keeping GLoRA in line with LoRA.

Could you document this decision in a comment above this line?

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,185 @@
...
+        self.glora_Ad, self.glora_Au = self.make_param((out_features, in_features), f"LoRA_{r}")
+        self.glora_Bd, self.glora_Bu = self.make_param((out_features, in_features), f"LoRA_{r}")
+        self.glora_Cd, self.glora_Cu = self.make_param((in_features, 1), f"LoRA_{r}")
+        self.glora_D = nn.Parameter(torch.zeros(out_features))
+        self.glora_E = nn.Parameter(torch.zeros(out_features))
+        # Initialize the parameters for GLora
+        self.eval_config = kwargs.get("eval_config", None)
+        nn.init.kaiming_uniform_(self.glora_Au, a=math.sqrt(5))
+        nn.init.kaiming_uniform_(self.glora_Bu, a=math.sqrt(5))
+        nn.init.kaiming_uniform_(self.glora_Cu, a=math.sqrt(5))
```
githubnemo: Let's move this to a separate `reset_parameters` method for clarity.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,185 @@
...
+import random
+import warnings
+from typing import Optional
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+
+random.seed(56)
```
githubnemo: Remove; seeding should be up to the user. If it is a problem there should be a proper seed method.

src/peft/tuners/glora/config.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    This is the configuration class to store the configuration of a [`GLoraModel`].
+
+    Args:
+        r (`int`): GLora attention dimension.
+        target_modules (`Optional[Union[List[str], str]]`): The names of the modules to apply GLora to.
+        # Add other GLORA specific parameters from your sample if they belong in the config
+    """
+
+    r: int = field(default=4, metadata={"help": "Lora attention dimension"})
```
githubnemo: Maybe something more akin to

> Default rank of the LoRA matrices if the config contains LoRA parametrization.

?

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,232 @@
...
+
+        if isinstance(config, GLoraConfig):
+            self.peft_config[adapter_name] = config
+            self.add_adapter(adapter_name, config)
+        elif isinstance(config, dict):
+            for name, cfg in config.items():
+                self.peft_config[name] = cfg
+                self.add_adapter(name, cfg)
+        else:
+            raise TypeError(f"Unsupported config type: {type(config)}")
```
githubnemo: Maybe I'm overlooking something but can't this be replaced with `super().__init__(model, config, adapter_name, low_cpu_mem_usage=low_cpu_mem_usage)`? (`low_cpu_mem_usage` needs to be added in the function parameter list, of course)

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,232 @@
...
+
+        if hasattr(current_config, "inference_mode") and current_config.inference_mode:
+            _freeze_adapter(self.model, adapter_name)
+
+    def _check_target_module_exists(self, glora_config: GLoraConfig, key: str) -> bool:
+        if isinstance(glora_config.target_modules, str):
+            return bool(re.fullmatch(glora_config.target_modules, key))
+        elif isinstance(glora_config.target_modules, list):
+            return any(key.endswith(target_key) for target_key in glora_config.target_modules)
+        return False
```
githubnemo, line 100: Let's use the default implementation of `check_target_module_exists`:

```python
from peft.tuners.tuners_utils import check_target_module_exists
[...]
    def _check_target_module_exists(self, glora_config: GLoraConfig, key: str) -> bool:
        return check_target_module_exists(glora_config, key)
```

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,232 @@
...
+            new_module.to(old_module.weight.device)
+
+    def __getattr__(self, name: str):
+        try:
+            return super().__getattr__(name)
+        except AttributeError:
+            return getattr(self.model, name)
+
+    @staticmethod
+    def _prepare_glora_config(peft_config: GLoraConfig, model_config: dict) -> GLoraConfig:
```
githubnemo, line 178: this should be `_prepare_adapter_config`

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,232 @@
...
+
+        if getattr(old_module, "state", None) is not None:
+            new_module.state = old_module.state
+            new_module.to(old_module.weight.device)
+
+    def __getattr__(self, name: str):
+        try:
+            return super().__getattr__(name)
+        except AttributeError:
+            return getattr(self.model, name)
```
githubnemo, line 175: In LoRA we catch this case in reference to #1892, I think this code is susceptible to this as well.

```suggestion
        except AttributeError:
            if name == "model":
                raise
            return getattr(self.model, name)
```

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,232 @@
...
+                target.merge()
+                new_module = nn.Linear(target.in_features, target.out_features, bias=(target.bias is not None))
+                new_module.weight.data = target.weight.data.clone()  # Get merged weight
+                if target.bias is not None:
+                    new_module.bias.data = target.bias.data.clone()  # Get merged bias
+
+                self._replace_module(parent, target_name, new_module.to(target.weight.device), target)
+
+            if isinstance(target, ModulesToSaveWrapper):
+                pass
```
githubnemo, line 255: I think it makes sense to do implement this similar to how LoRA does it:

```python
                if hasattr(target, "unload_and_optionally_merge_module"):
                    # if layers have special unloading method, like MultiheadAttention, use that
                    unloaded_module = target.unload_and_optionally_merge_module(
                        merge=merge, safe_merge=safe_merge, adapter_names=adapter_names
                    )
                    self._replace_module(parent, target_name, unloaded_module, target)
                elif hasattr(target, "base_layer"):
                    if merge:
                        target.merge(safe_merge=safe_merge, adapter_names=adapter_names)
                    self._replace_module(parent, target_name, target.get_base_layer(), target)
```

Of course this assumes that the target (e.g., `glora.Linear`) implements `unload_and_optionally_merge_module` but I would have suggested to implement it anyway since it is a good interface to have.

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,232 @@
...
+        config_dict = {}
+        for adapter_name, peft_config_obj in self.peft_config.items():
+            config = asdict(peft_config_obj)
+            if inference:
+                config["inference_mode"] = True
+            for k, v in config.items():
+                if isinstance(v, Enum):
+                    config[k] = v.value
+            config_dict[adapter_name] = config
+        return config_dict
```
githubnemo, line 283: Remove?

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,232 @@
...
+    def print_trainable_parameters(self):
+        trainable_params = 0
+        all_param = 0
+        for _, param in self.named_parameters():
+            all_param += param.numel()
+            if param.requires_grad:
+                trainable_params += param.numel()
+        print(
+            f"trainable params: {trainable_params} || all params: {all_param} || trainable%: {100 * trainable_params / all_param:.4f}"
+        )
```
githubnemo: Remove?

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,232 @@
...
+                )
+            elif isinstance(target, nn.Linear):
+                new_module = self._create_new_module(glora_config, adapter_name, target)
+                self._replace_module(parent, target_name, new_module, target)
+
+        if not is_target_modules_in_base_model:
+            raise ValueError(
+                f"Target modules {glora_config.target_modules} not found in the base model. "
+                f"Please check the target modules and try again."
+            )
```
githubnemo, line 155: This functionality is already implemented in `BaseTuner.inject_adapter`. I think `_find_and_replace` as well as `add_adapter` can be removed entirely?

githubnemo (commented):
Thanks for the PR, this looks already quite mature. I especially like that you've gone through the trouble to add tests! I've added some comments on things that I think can be improved still.

I think it would make sense to also add a test case in `test_custom_models.py` since this will give a good coverage over the existing functionality in PEFT. Feel free to ask if you need assistance in doing so.

As I understand the paper they use random sampling + evolutionary search to select the best configuration for the different parameters (i.e., how to structure, say, `A`: LoRA, vector, scalar, none). PEFT aims to not interfere with the training process, as such I would suggest to *not* use random sampling of this configuration during init but to give the user the option to set the config once (via `GloraConfig`, similar to how `eval_config` works). If the user is keen to try search methods, such as evolutionary search, they can still use either `GloraModel.set_adapter_config` or, this would need to be added, something like `glora.Linear.set_adapter_config`, so that the user can do this for each layer individually (to support the ES case). Maybe there's a better name instead of `set_adapter_config` to set it apart from PEFT adapters. WDYT?

src/peft/tuners/glora/layer.py
not-lain:
```suggestion
            rank = int(config.split("_")[1])
```

src/peft/tuners/glora/model.py
not-lain:
```suggestion
```

src/peft/tuners/glora/layer.py
not-lain: agreed
```suggestion
```

src/peft/tuners/glora/model.py
not-lain: I think this should be discussed in another PR, as it's not part of the BaseTuner and is part of each model in the library.
IMO we should make a different PR to shift this method to the BaseTuner.

src/peft/tuners/glora/config.py
not-lain: LGTM
```suggestion
    r: int = field(default=4, metadata={"help": "Default rank of the LoRA matrices if the config contains LoRA parametrization."})
```

src/peft/utils/constants.py
not-lain: I did not test this yet, I was waiting for the workflow to run and see how it goes, this choice so far is based on intuition and nothing more, will inspect this in more detail further in the future.

not-lain (commented):
thanks a lot for the review! 
will ping you once I resolve all the pointers

src/peft/tuners/glora/layer.py
not-lain: (leaving this comment for future me to reflect on this one) 
IMO rank should be extracted from `config["r"]` I may need to go back to the original implementation and check this in more depth

src/peft/tuners/glora/layer.py
not-lain: I think it would be better if we move `eval_config` to the `GLoraConfig` class and set a default value for that over there

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,179 @@
...
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+
+class GLoraLayer:
+    def __init__(self, adapter_name: str, in_features: int, out_features: int, r: int = 0, **kwargs):
+        self.r = {}
+        self.r[adapter_name] = r
```
not-lain: this is confusing, this needs to be changed to something else

not-lain: after some checks, this is the default behaviour in other loras for the case when multiple adapters are present

src/peft/tuners/glora/layer.py
not-lain: Done, now eval_config is part of the `GLoraConfig` with default values

BenjaminBossan:
@not-lain What is the state of the PR, is it ready for review or do you still need to work on it?

not-lain:
Hi @BenjaminBossan this PR needs some minor modifications, will request a review once it's ready (apologies for the slow development on my end)

githubnemo:
@not-lain are you still working on this? Otherwise we'd convert this into an issue so someone else can take this :)

not-lain:
@githubnemo 
Thanks a lot for suggesting this, I'm with you in this one, feel free to open a separate issue for this, i cannot pick this pr any time soon because of a busy schedule and i think it's the sensible thing to retire it for good.
imo the only good thing to retain from this pr is the config file, other scripts are too messy and unusable.

PR #2569, opened 2025-06-03
`prepare_model_for_gradient_checkpointing` protected to public

qgallouedec:
It's quite convenient to be able to enable gradient checkpointing after initialization.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -1317,8 +1317,16 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
         ):
             device_map = kwargs.get("device_map", "auto")
             max_memory = kwargs.get("max_memory", None)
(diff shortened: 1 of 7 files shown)


BenjaminBossan (changes requested):
Thanks for the PR. I agree that we can make the method public. Instead of renaming it, could you please create a new `prepare_model_for_gradient_checkpointing` method that just calls the existing method? That way, if there is any code that uses the private method, it continues to work. And speaking of which, this call should also be updated: https://github.com/huggingface/peft/blob/6d133307adf28e11d061c20aa1f040c1d26d2012/src/peft/peft_model.py#L138

BenjaminBossan (approved):
Thanks for the update, LGTM. Failing MacOS tests are unrelated.

PR #2570, opened 2025-06-04
FIX: Correctly determine no_split_modules

BenjaminBossan:
See discussion in https://github.com/huggingface/transformers/pull/38141 for context.

In the PEFT `fsdp_auto_wrap` policy, we determine the `_no_split_modules`. However, this currently neglects to visit the children of the model, which can be required for some architectures. This PR fixes that.

Note that the `_get_no_split_modules` function is largely [copied from transformers](https://github.com/huggingface/transformers/blob/1e921a3a9cea92b383ca4b0484ee45596bbdadc3/src/transformers/modeling_utils.py#L2677-L2704). One change is that it doesn't take the `device_map` argument. That argument is used in transformers inside an error message but not for the logic proper. I think it's safe to remove.

Moreover, I made an unrelated change to `fsdp_auto_wrap_policy`, namely making local imports global (there was no reason for them to be local).

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -14,6 +14,7 @@
 from __future__ import annotations
 
 import copy
+import functools
 import inspect
 import os
 import re
@@ -24,12 +25,14 @@ from typing import Any, Optional, Union
 
 import accelerate
 import torch
+from accelerate import FullyShardedDataParallelPlugin
 from accelerate.hooks import add_hook_to_module, remove_hook_from_module
 from accelerate.utils import is_npu_available, is_xpu_available
 from huggingface_hub import file_exists
 from huggingface_hub.errors import EntryNotFoundError, HFValidationError
 from packaging import version
 from safetensors.torch import storage_ptr, storage_size
+from transformers import PreTrainedModel
 
 from ..import_utils import is_auto_gptq_available, is_gptqmodel_available, is_torch_tpu_available
 from .constants import (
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
> Thanks for fixing! I don't think there are any FSDP tests right, but I assume it should fix the user's issue

No end to end tests @zucchini-nlp, only very light testing [here](https://github.com/huggingface/peft/blob/62c9cf30319ca219e1d6754626c17f510fc76441/tests/test_gpu_examples.py#L4174). As the code is mostly copied from transformers, it should hopefully work as expected.

src/peft/utils/other.py
```diff
@@ -942,12 +945,33 @@ def _prepare_prompt_learning_config(peft_config, model_config):
...
+    the underlying `_no_split_modules`.
+
+    Returns:
+        `List[str]`: List of modules that should not be split
+    """
+    # After discussion in https://github.com/huggingface/transformers/pull/38141, based on:
+    # https://github.com/huggingface/transformers/blob/1e921a3a9cea92b383ca4b0484ee45596bbdadc3/src/transformers/modeling_utils.py#L2677-L2704
+    _no_split_modules: set[str] = set()
+    if not hasattr(model, "_no_split_modules"):
+        return _no_split_modules
```
githubnemo, line 960: It can't be the case that the container model doesn't have the attribute but children do?

BenjaminBossan: I think we should be safe as it's an attribute on `PreTrainedModel`:

https://github.com/huggingface/transformers/blob/9300728665aaeb0ebf4db99f9d9fbce916b4a183/src/transformers/modeling_utils.py#L2052

PR #2571, opened 2025-06-04
Add support for Quantization-Aware Low-Rank Adaptation (QALoRA)

gapsong:
This pull request introduces QALoRA (Quantization-Aware Low-Rank Adaptation), a new fine-tuning technique for quantized large language models, along with its implementation in the PEFT library. The changes include updates to documentation, configuration, and core logic to support QALoRA's memory-efficient and performance-preserving features.

### Documentation Updates
* [`examples/qalora_finetuning/README.md`](diffhunk://#diff-6e8694feed4e98d9caf8f6a187aeb945e989bf6e00392be4b7674bb0b6be96d7R1-R178): Added detailed documentation for QALoRA, including its introduction, implementation details, usage examples, command-line instructions, and comparison with other techniques like LoRA and DoRA.

### Configuration Enhancements
* [`src/peft/tuners/lora/config.py`](diffhunk://#diff-13380145401d203d5935c5189dd09879f990b81aa63e8e3aaff8ce9110333f0eR497-R519): Introduced two new configuration parameters: `use_qalora` to enable QALoRA and `qalora_group_size` to control the pooling group size for memory-performance tradeoffs.

### Core Logic for QALoRA
* [`src/peft/tuners/lora/gptq.py`](diffhunk://#diff-c65b84680649207b662504c2e0bcf65d8478ae27ed4ff6d7d58f06b615b025a2R22): Updated the GPTQ LoRA implementation to support QALoRA, including logic for resolving QALoRA variants and passing group size parameters. [[1]](diffhunk://#diff-c65b84680649207b662504c2e0bcf65d8478ae27ed4ff6d7d58f06b615b025a2R22) [[2]](diffhunk://#diff-c65b84680649207b662504c2e0bcf65d8478ae27ed4ff6d7d58f06b615b025a2R36) [[3]](diffhunk://#diff-c65b84680649207b662504c2e0bcf65d8478ae27ed4ff6d7d58f06b615b025a2R58-R74)
* [`src/peft/tuners/lora/layer.py`](diffhunk://#diff-24a141c266b7b714ae8fcc470f31bc283f7b0f5a671bbf6d5f092741fc374104R187-R208): Enhanced the layer update logic to initialize QALoRA-specific parameters and handle adapter-specific configurations. [[1]](diffhunk://#diff-24a141c266b7b714ae8fcc470f31bc283f7b0f5a671bbf6d5f092741fc374104R187-R208) [[2]](diffhunk://#diff-24a141c266b7b714ae8fcc470f31bc283f7b0f5a671bbf6d5f092741fc374104L242-R253)
* [`src/peft/tuners/lora/model.py`](diffhunk://#diff-4331e1b00e557c2af6e682bc221d1517f554d7ca24b9149520237b431d8d797fR203-R204): Incorporated QALoRA-specific parameters into the model creation and replacement process.

### QALoRA Variant Implementation
* [`src/peft/tuners/lora/variants.py`](diffhunk://#diff-a14f8c2e9d3ff608a836f59d47401dc0e66c19a8e7ea5c25434ab857da6c0eccR133-R321): Added the `QALoraLinearVariant` class, implementing QALoRA-specific logic for initialization, delta weight computation, merging, unmerging, and forward propagation. This includes pooling input features and scaling them for efficient adaptation.

### DIFF
--- a/examples/qalora_finetuning/README.md
+++ b/examples/qalora_finetuning/README.md
@@ -1,7 +1,7 @@
 # QALoRA: Quantization-Aware Low-Rank Adaptation
 
 ## Introduction
-[QALoRA](https://huggingface.co/papers/2309.14717) is a quantization-aware version of Low-Rank Adaptation that enables efficient fine-tuning of quantized large language models. QALoRA uses input feature pooling and a specialized grouping technique to work with quantized weights, significantly reducing memory requirements while preserving performance. QALoRA enables fine-tuning of models that would otherwise be too large for consumer GPUs.
+[QALoRA](https://huggingface.co/papers/2309.14717) is a quantization-aware version of Low-Rank Adaptation that enables efficient fine-tuning of quantized large language models. QALoRA uses input feature pooling and a specialized grouping technique to work with quantized weights, significantly reducing memory requirements while preserving performance. QALoRA enables fine-tuning of models that would otherwise be too large for consumer GPUs. In PEFT it only works for GPTQ.
 
 ## Quick start
 ```python
@@ -70,7 +70,7 @@ python examples/qalora_finetuning/qalora_gptq_finetuning.py \
     --qalora_group_size 8
 ```
 
-QALoRA also works with different quantization methods (GPTQ, EETQ, AWQ, etc.):
+QALoRA works with GPTQ:
 ```bash
 python examples/qalora_finetuning/qalora_gptq_finetuning.py \
     --base_model TheBloke/Llama-2-13b-GPTQ \
@@ -146,7 +146,6 @@ While the QA-LoRA paper describes a direct weight modification technique using "
 2. The QALoRA adapter weights are then merged with the dequantized model
 3. The merged model must be re-quantized if quantization is still desired
 
-This implementation choice was made because **it yielded better performance in practice**, despite being less memory-efficient than the direct modification approach described in the paper.
 
 ### Memory Considerations
 
--- a/examples/qalora_finetuning/README.md
+++ b/examples/qalora_finetuning/README.md
@@ -65,13 +65,6 @@ python examples/qalora_finetuning/qalora_gptq_finetuning.py \
     --qalora_group_size 8
 ```
 
-QALoRA works with GPTQ:
-```bash
-python examples/qalora_finetuning/qalora_gptq_finetuning.py \
-    --base_model TheBloke/Llama-2-13b-GPTQ \
-    --use_qalora
-```
-
 ### Full example of the script 
 ```bash
 python qalora_gptq_finetuning.py \
@@ -81,7 +74,6 @@ python qalora_gptq_finetuning.py \
     --num_epochs 3 \
     --learning_rate 3e-4 \
     --cutoff_len 512 \
-    --val_set_size 500 \
     --use_qalora \
     --qalora_group_size 8 \
     --eval_step 10 \
@@ -91,7 +83,6 @@ python qalora_gptq_finetuning.py \
     --lora_alpha 32 \
     --lora_dropout 0.05 \
     --lora_target_modules "q_proj,k_proj,v_proj,o_proj,gate_proj,up_proj,down_proj" \
-    --hub_model_id "YOUR_HF_REPO" \
     --push_to_hub
 ```
 
--- a/examples/qalora_finetuning/README.md
+++ b/examples/qalora_finetuning/README.md
@@ -1,0 +1,178 @@
+# QALoRA: Quantization-Aware Low-Rank Adaptation
+
+## Introduction
+[QALoRA](https://huggingface.co/papers/2309.14717) is a quantization-aware version of Low-Rank Adaptation that enables efficient fine-tuning of quantized large language models. QALoRA uses input feature pooling and a specialized grouping technique to work with quantized weights, significantly reducing memory requirements while preserving performance. QALoRA enables fine-tuning of models that would otherwise be too large for consumer GPUs.
+
+## Quick start
+```python
+import torch
+from peft import LoraConfig, get_peft_model
+from transformers import AutoTokenizer, AutoModelForCausalLM, Trainer
+from datasets import load_dataset
+
+# Load a quantized model (example with GPTQ quantization)
+model = AutoModelForCausalLM.from_pretrained(
+    "TheBloke/Llama-2-7b-GPTQ", 
+    revision="gptq-4bit-32g-actorder_True", 
+    device_map="auto"
+)
+tokenizer = AutoTokenizer.from_pretrained("TheBloke/Llama-2-7b-GPTQ")
+dataset = load_dataset("timdettmers/openassistant-guanaco", split="train")
+
+# Configure QALoRA parameters
+lora_config = LoraConfig(
+    use_qalora=True,
+    qalora_group_size=8,
+    r=16,
+    lora_alpha=32,
+    target_modules=["q_proj", "k_proj", "v_proj", "o_proj", "gate_proj", "up_proj", "down_proj"],
+    lora_dropout=0.05,
+)
+
+# Create the PEFT model
+peft_model = get_peft_model(model, lora_config)
+
+# Set up trainer and train
+trainer = Trainer(
+    model=peft_model,
+    train_dataset=dataset,
+    args=TrainingArguments(
+        per_device_train_batch_size=1,
+        gradient_accumulation_steps=4,
+        num_train_epochs=3,
+        learning_rate=3e-4,
+        output_dir="qalora-llama-2-7b"
+    ),
+    data_collator=DataCollatorForLanguageModeling(tokenizer, mlm=False),
+)
+trainer.train()
+peft_model.save_pretrained("qalora-llama-2-7b")
+```
+
+To use QALoRA, simply set `use_qalora = True` and specify a `qalora_group_size` in your LoRA configuration. The group size controls the memory/performance tradeoff - smaller values use less memory but may affect performance.
+
+## Command Line Examples
+
+Run the finetuning script with a GPTQ quantized model:
+```bash
+python examples/qalora_finetuning/qalora_gptq_finetuning.py \
+    --base_model TheBloke/Llama-2-7b-GPTQ \
+    --data_path timdettmers/openassistant-guanaco \
+    --use_qalora \
+    --quantize
+```
+
+You can customize the pooling group size (default is 16):
+```bash
+python examples/qalora_finetuning/qalora_gptq_finetuning.py \
+    --base_model TheBloke/Llama-2-7b-GPTQ \
+    --use_qalora \
+    --qalora_group_size 8
+```
+
+QALoRA also works with different quantization methods (GPTQ, EETQ, AWQ, etc.):
+```bash
+python examples/qalora_finetuning/qalora_gptq_finetuning.py \
+    --base_model TheBloke/Llama-2-13b-GPTQ \
+    --use_qalora
+```
+
+### Full example of the script 
+```bash
+python qalora_gptq_finetuning.py \
+    --base_model "TheBloke/Llama-2-13b-GPTQ" \
+    --data_path "PATH_TO_DATASET" \
+    --output_dir "PATH_TO_OUTPUT_DIR" \
+    --batch_size 1 \
+    --num_epochs 3 \
+    --learning_rate 3e-4 \
+    --cutoff_len 512 \
+    --val_set_size 500 \
+    --use_qalora \
+    --qalora_group_size 8 \
+    --eval_step 10 \
+    --save_step 100 \
+    --device "cuda:0" \
+    --lora_r 16 \
+    --lora_alpha 32 \
+    --lora_dropout 0.05 \
+    --lora_target_modules "q_proj,k_proj,v_proj,o_proj,gate_proj,up_proj,down_proj" \
+    --hub_model_id "YOUR_HF_REPO" \
+    --push_to_hub
+```
+
+## Use the model on 🤗
+You can load and use the finetuned model like any other PEFT model:
+```python
+from peft import PeftModel, PeftConfig
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+# Load the base quantized model
+base_model = AutoModelForCausalLM.from_pretrained(
+    "TheBloke/Llama-2-7b-GPTQ",
+    device_map="auto"
+)
+tokenizer = AutoTokenizer.from_pretrained("TheBloke/Llama-2-7b-GPTQ")
+
+# Load the PEFT adapter
+peft_model_id = "YOUR_HF_REPO"
+model = PeftModel.from_pretrained(base_model, peft_model_id)
+
+# Generate text
+input_text = "Hello, I'm a language model"
+inputs = tokenizer(input_text, return_tensors="pt").to(model.device)
+outputs = model.generate(**inputs, max_length=100)
+print(tokenizer.decode(outputs[0], skip_special_tokens=True))
+```
+
+## QALoRA vs. LoRA vs. DoRA
+
+QALoRA offers several advantages over standard LoRA:
+
+1. **Memory efficiency**: QALoRA works directly with quantized models, reducing memory usage by up to 60-70% compared to standard LoRA.
+
+2. **Hardware accessibility**: Enables fine-tuning of larger models (13B, 70B) on consumer GPUs that would be impossible with standard LoRA.
+
+3. **Performance preservation**: Despite quantization, QALoRA can achieve comparable performance to full-precision LoRA in many tasks.
+
+Compared to DoRA:
+- DoRA focuses on improved expressivity through weight decomposition
+- QALoRA focuses on memory efficiency through quantization awareness
+- Both can be used with quantized models, but QALoRA is specifically designed for this use case
+
+
+## Implementation Details: Merging with Quantized Models
+
+> **Note:** The current implementation differs from the original QA-LoRA paper's approach.
+
+While the QA-LoRA paper describes a direct weight modification technique using "beta shift" to modify quantized weights without full dequantization, this implementation uses a different approach:
+
+1. The quantized model is first dequantized to full precision
+2. The QALoRA adapter weights are then merged with the dequantized model
+3. The merged model must be re-quantized if quantization is still desired
+
+This implementation choice was made because **it yielded better performance in practice**, despite being less memory-efficient than the direct modification approach described in the paper.
+
+### Memory Considerations
+
+This process requires significant memory (enough to hold the full dequantized model) and additional computation for the re-quantization step. For large models, this may not be possible on consumer hardware.
+
+For most use cases, we recommend keeping the base quantized model and the QALoRA adapter separate, loading them with `PeftModel.from_pretrained()` as shown in the usage example above. This approach maintains the memory efficiency benefits of quantization throughout the deployment pipeline.
+
+
+## Citation
+```
+@article{dettmers2023qlora,
+  title={QLoRA: Efficient Finetuning of Quantized LLMs},
+  author={Dettmers, Tim and Pagnoni, Artidoro and Holtzman, Ari and Zettlemoyer, Luke},
+  journal={arXiv preprint arXiv:2305.14314},
+  year={2023}
+}
+
+@article{xu2023qalora,
+  title={QA-LoRA: Quantization-Aware Low-Rank Adaptation of Large Language Models},
+  author={Xu, Yuhui and Liu, Lingxi and Rao, Longhui and Zhao, Teng and Xiong, Zhiwei and Gao, Mingkui},
+  journal={arXiv preprint arXiv:2309.14717},
+  year={2023}
+}
+```
--- a/examples/qalora_finetuning/README.md
+++ b/examples/qalora_finetuning/README.md
@@ -1,7 +1,9 @@
 # QALoRA: Quantization-Aware Low-Rank Adaptation
 
 ## Introduction
-[QALoRA](https://huggingface.co/papers/2309.14717) is a quantization-aware version of Low-Rank Adaptation that enables efficient fine-tuning of quantized large language models. QALoRA uses input feature pooling and a specialized grouping technique to work with quantized weights, significantly reducing memory requirements while preserving performance. QALoRA enables fine-tuning of models that would otherwise be too large for consumer GPUs. In PEFT it only works for GPTQ.
+[QALoRA](https://huggingface.co/papers/2309.14717) is a quantization-aware version of Low-Rank Adaptation that enables efficient fine-tuning of quantized large language models. 
+QALoRA uses input feature pooling and a specialized grouping technique to work with quantized weights, significantly reducing memory requirements while preserving performance. 
+QALoRA enables fine-tuning of models that would otherwise be too large for consumer GPUs. In PEFT it only works for GPTQ.
 
 ## Quick start
 ```python
@@ -54,13 +56,6 @@ To use QALoRA, simply set `use_qalora = True` and specify a `qalora_group_size` 
 ## Command Line Examples
 
 Run the finetuning script with a GPTQ quantized model:
-```bash
-python examples/qalora_finetuning/qalora_gptq_finetuning.py \
-    --base_model TheBloke/Llama-2-7b-GPTQ \
-    --data_path timdettmers/openassistant-guanaco \
-    --use_qalora \
-    --quantize
-```
 
 You can customize the pooling group size (default is 16):
 ```bash
@@ -81,7 +76,6 @@ python examples/qalora_finetuning/qalora_gptq_finetuning.py \
 ```bash
 python qalora_gptq_finetuning.py \
     --base_model "TheBloke/Llama-2-13b-GPTQ" \
-    --data_path "PATH_TO_DATASET" \
     --output_dir "PATH_TO_OUTPUT_DIR" \
     --batch_size 1 \
     --num_epochs 3 \
--- a/examples/qalora_finetuning/README.md
+++ b/examples/qalora_finetuning/README.md
@@ -62,7 +62,7 @@ You can customize the pooling group size (default is 16):
 python examples/qalora_finetuning/qalora_gptq_finetuning.py \
     --base_model TheBloke/Llama-2-7b-GPTQ \
     --use_qalora \
-    --qalora_group_size 8
+    --qalora_group_size 32
 ```
 
 ### Full example of the script 
@@ -75,7 +75,7 @@ python qalora_gptq_finetuning.py \
     --learning_rate 3e-4 \
     --cutoff_len 512 \
     --use_qalora \
-    --qalora_group_size 8 \
+    --qalora_group_size 32 \
     --eval_step 10 \
     --save_step 100 \
     --device "cuda:0" \
--- a/examples/qalora_finetuning/README.md
+++ b/examples/qalora_finetuning/README.md
@@ -125,7 +125,7 @@ outputs = model.generate(**inputs, max_length=100)
 print(tokenizer.decode(outputs[0], skip_special_tokens=True))
 ```
 
-## QALoRA vs. LoRA vs. DoRA
+## QALoRA vs. LoRA
 
 QALoRA offers several advantages over standard LoRA:
 
@@ -134,11 +134,6 @@ QALoRA offers several advantages over standard LoRA:
 2. **Hardware accessibility**: Enables fine-tuning of larger models (13B, 70B) on consumer GPUs that would be impossible with standard LoRA.
 
 3. **Performance preservation**: Despite quantization, QALoRA can achieve comparable performance to full-precision LoRA in many tasks.
-
-Compared to DoRA:
-- DoRA focuses on improved expressivity through weight decomposition
-- QALoRA focuses on memory efficiency through quantization awareness
-- Both can be used with quantized models, but QALoRA is specifically designed for this use case
 
 
 ## Implementation Details: Merging with Quantized Models
--- a/examples/qalora_finetuning/qalora_gptq_finetuning.py
+++ b/examples/qalora_finetuning/qalora_gptq_finetuning.py
@@ -1,0 +1,423 @@
+#!/usr/bin/env python3
+"""
+Training script for fine-tuning language models with QALoRA using GPTQ quantization.
+This script supports cached quantization to avoid repeating expensive quantization processes.
+"""
+
+import argparse
+import os
+
+import torch
+from datasets import load_dataset
+from transformers import (
+    AutoModelForCausalLM,
+    AutoTokenizer,
+    DataCollatorForLanguageModeling,
+    GPTQConfig,
+    Trainer,
+    TrainingArguments,
+)
+
+from peft import LoraConfig, get_peft_model
+
+
+def check_adapter_gradients(model) -> None:
+    """
+    Verify that QALoRA adapters are properly initialized with correct gradient settings.
+    Checks if adapter parameters are trainable while base model weights are frozen.
+
+    Args:
+        model: The PEFT model with QALoRA adapters
+    """
+    print("\n=== QALoRA Implementation Verification ===")
+
+    total_params = 0
+    trainable_params = 0
+    qalora_modules = 0
+
+    # Check trainable parameters
+    for name, param in model.named_parameters():
+        total_params += param.numel()
+        if param.requires_grad:
+            trainable_params += param.numel()
+
+    # Check for QALoRA modules
+    for name, module in model.named_modules():
+        if hasattr(module, "qalora_group_size"):
+            qalora_modules += 1
+            adapter_name = next(iter(module.qalora_group_size.keys()), None)
+            if adapter_name:
+                print(f"\nQALoRA module found: {name}")
+                print(f"  - Group size: {module.qalora_group_size[adapter_name]}")
+                print(f"  - Scaling factor: {module.qalora_scaling_factor.get(adapter_name, 'N/A')}")
+
+                # Check adapter matrices
+                if hasattr(module, "lora_A") and adapter_name in module.lora_A:
+                    a_shape = module.lora_A[adapter_name].weight.shape
+                    a_grad = module.lora_A[adapter_name].weight.requires_grad
+                    print(f"  - lora_A shape: {a_shape}, requires_grad: {a_grad}")
+
+                if hasattr(module, "lora_B") and adapter_name in module.lora_B:
+                    b_shape = module.lora_B[adapter_name].weight.shape
+                    b_grad = module.lora_B[adapter_name].weight.requires_grad
+                    print(f"  - lora_B shape: {b_shape}, requires_grad: {b_grad}")
+
+                # Check base weights
+                if hasattr(module, "weight"):
+                    base_grad = module.weight.requires_grad
+                    print(f"  - Base weight requires_grad: {base_grad} (should be False)")
+
+    # Print summary
+    print("\nParameter summary:")
+    print(f"  - Total parameters: {total_params:,}")
+    print(f"  - Trainable parameters: {trainable_params:,} ({trainable_params / total_params * 100:.2f}%)")
+    print(f"  - QALoRA modules found: {qalora_modules}")
+
+    if qalora_modules > 0 and trainable_params > 0:
+        print("✅ QALoRA implementation verification: PASSED")
+    else:
+        print("❌ QALoRA implementation verification: FAILED")
+
+
+def print_parameter_memory(model, optimizer_name: str = "adam") -> None:
+    """
+    Print detailed memory usage information for model parameters.
+
+    Args:
+        model: The model to analyze
+        optimizer_name: The optimizer type to estimate memory for ("adam" or "sgd")
+    """
+    total_param_mem = 0
+    total_grad_mem = 0
+    total_opt_mem = 0
+
+    print(f"{'Name':50} {'Shape':20} {'Dtype':10} {'Param MB':10} {'Grad MB':10} {'Requires Grad'}")
+    print("-" * 110)
+
+    for name, param in model.named_parameters():
+        param_size = param.numel() * param.element_size() / 1024**2  # in MB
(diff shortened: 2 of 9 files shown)


gapsong:
Hi @BenjaminBossan ,

Thank you for your review and the helpful feedback!

I've addressed the main points you raised:
*   I have scraped the merge/unmerge logic completely from `QALoraLinearVariant` for the `GPTQLoraLinear` context, as you correctly pointed out it's not supported there and was effectively dead code.
*   Thank you especially for pointing out that the QALoRA variant's forward function was never actually being called within `GPTQLoraLinear`. This was a critical oversight, and I've now updated `GPTQLoraLinear.forward` to correctly dispatch to the QALoRA variant's forward pass when QALoRA is active.

Ready for another look when you have a moment!

src/peft/tuners/lora/config.py
```diff
@@ -494,6 +494,29 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    use_qalora: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Enable <a href='https://huggingface.co/papers/2309.14717'>Quantization-Aware Low-Rank Adaptation (QALoRA)</a>. This technique combines quantization-aware training "
```
BenjaminBossan: Let's mention that it is only implemented for GPTQ for now. Also, please update the docstring (can use the same text).

src/peft/tuners/lora/config.py
```diff
@@ -494,6 +494,29 @@ class LoraConfig(PeftConfig):
...
+                "with LoRA to improve performance for quantized models. This can improve the performance of LoRA, "
+                "especially at low ranks. Right now, QALoRA only supports linear layers. QALoRA introduces a bigger "
+                "overhead than pure LoRA, so it is recommended to merge weights for inference."
```
BenjaminBossan: This recommendation is a bit moot as merging is not supported for GPTQ. Let's remove this sentence.

src/peft/tuners/lora/gptq.py
```diff
@@ -53,9 +55,23 @@ def __init__(
...
+        if use_qalora:
+            from .variants import QALoraLinearVariant
+
+            return QALoraLinearVariant()
+        if not use_dora:
+            return None
+
+        from .variants import DoraLinearVariant
+
+        return DoraLinearVariant()
```
BenjaminBossan: Let's change the check a bit for completeness to basically:

```python
if use_dora and use_qalora:
    NotImplementedError
elif use_dora:
    variant = ...
elif use_qalora:
    variant = ...
else:
    variant = None
return variant
```

src/peft/tuners/lora/gptq.py
```diff
@@ -64,29 +80,33 @@ def forward(self, x: torch.Tensor):
             return result
 
         lora_A_keys = self.lora_A.keys()
+        torch_result_dtype = result.dtype
```
BenjaminBossan: This is not needed, right?

src/peft/tuners/lora/gptq.py
```diff
@@ -64,29 +80,33 @@ def forward(self, x: torch.Tensor):
...
-            requires_conversion = not torch.is_autocast_enabled()
-            if requires_conversion:
-                expected_dtype = result.dtype
-                x = self._cast_input_dtype(x, lora_A.weight.dtype)
-
-            output = lora_B(lora_A(dropout(x)))
-
-            if requires_conversion:
-                output = output.to(expected_dtype)
+            # requires_conversion = not torch.is_autocast_enabled()
```
BenjaminBossan: remove the comment?

src/peft/tuners/lora/layer.py
```diff
@@ -184,17 +184,28 @@ def update_layer(
...
         use_rslora,
         use_dora: bool = False,
+        use_qalora: bool = False,
         lora_bias: bool = False,
+        **kwargs
     ):
-        # collect the kwargs
-        kwargs = locals().copy()
-        del kwargs["self"]
+        params_for_variant_init = {
```
BenjaminBossan: Why was this change necessary?

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,186 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
...
+
+        Args:
+            module (Linear): The linear module to be adapted.
+            adapter_name (str): The name of the adapter.
+            **kwargs: Additional keyword arguments.
+                qalora_group_size (int, optional): The size of groups for pooling. Defaults to 16.
+        """
+        # Store the qalora_group_size as a module attribute
+        # It's expected that qalora_group_size is passed in kwargs from LoraConfig
+        qalora_group_size = kwargs.get("qalora_group_size", 16)  # Default to 16 if not specified in kwargs
```
BenjaminBossan: We should ensure the argument is always there, thus `kwargs.get` should not be necessary.

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,186 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
...
+        if not hasattr(module, "qalora_group_size"):
+            module.qalora_group_size = {}
```
BenjaminBossan: We need to add: `module.other_param_names = module.other_param_names[:] + ("qalora_group_size",)`

This corresponds to this DoRA code:

https://github.com/huggingface/peft/blob/e67052b18cb3c58370e0339ecf76a5faf447f2ff/src/peft/tuners/lora/variants.py#L33

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,186 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
...
+        module.qalora_group_size[adapter_name] = qalora_group_size
+
+        # Store original dimensions for use in get_delta_weight
+        if not hasattr(module, "orig_in_features"):
+            module.orig_in_features = {}
```
BenjaminBossan: Same comment as above regarding `other_param_names`

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,186 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
...
+        module.orig_in_features[adapter_name] = module.in_features
+
+        # Create and store pooling factor for scaling
+        if not hasattr(module, "qalora_scaling_factor"):
+            module.qalora_scaling_factor = {}
```
BenjaminBossan: Same comment as above regarding `other_param_names`. But do we even need `qalora_scaling_factor`, as it can be calculated on the fly based on `module.in_features` and `qalora_group_size` anyway?

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,186 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
...
+            # depending on interpretation.
+            # Let's assume paper_scaling_factor is related to how x_pooled is scaled before LoRA.
+            # The current implementation calculates it as module.in_features / qalora_group_size,
+            # which is the number of groups (L).
+            # If the intention is to scale by group_size, this should be `qalora_group_size`.
+            # For now, keeping the existing logic for qalora_scaling_factor:
+            module.qalora_scaling_factor[adapter_name] = module.in_features / qalora_group_size
+        else:
+            # No special scaling if dimensions don't align
+            module.qalora_scaling_factor[adapter_name] = 1.0
```
BenjaminBossan: Would this not lead to very different results than `qalora_scaling_factor = module.in_features / qalora_group_size`?

I wonder if perhaps it makes more sense to raise an error here and require that `module.in_features % qalora_group_size == 0`? I think this would simplify the code and when `module.in_features % qalora_group_size != 0` we just get standard LoRA?

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,186 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
...
+
+        Args:
+            module (Linear): The adapted linear module.
+            active_adapter (str): The currently active adapter.
+
+        Returns:
+            torch.Tensor: The calculated delta weight.
+        """
+        if (
+            not hasattr(module, "qalora_group_size")
```
BenjaminBossan: Is it even possible to hit this condition? Same question about `not hasattr(module, "qalora_scaling_factor")`.

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,186 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
...
+            # which is the number of groups (L).
+            # If the intention is to scale by group_size, this should be `qalora_group_size`.
+            # For now, keeping the existing logic for qalora_scaling_factor:
+            module.qalora_scaling_factor[adapter_name] = module.in_features / qalora_group_size
+        else:
+            # No special scaling if dimensions don't align
+            module.qalora_scaling_factor[adapter_name] = 1.0
+
+    @staticmethod
+    def get_delta_weight(module: Linear, active_adapter: str) -> torch.Tensor:
```
BenjaminBossan: Just wondering, as merging is not implemented, do we even need this method? I guess we could leave it here in case merging is added later, but at least a comment should be added in that case.

src/peft/tuners/lora/layer.py
gapsong: This change was originally made to prevent `kwargs` from being nested (so we wouldn't have to use `kwargs.kwargs["qalora_group_size"])`. However, since `qalora_group_size` is now always a parameter, this change is obsolete.

src/peft/tuners/lora/variants.py
gapsong: You are right. We will never reach in here

gapsong:
@BenjaminBossan I have integrated all your feedback into the code, except for the tests.
I'm currently working on those. I will structure the tests similarly to the dora logic and integrate them in the same way.

examples/qalora_finetuning/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+You can customize the pooling group size (default is 16):
+```bash
+python examples/qalora_finetuning/qalora_gptq_finetuning.py \
+    --base_model TheBloke/Llama-2-7b-GPTQ \
+    --use_qalora \
+    --qalora_group_size 8
+```
+
+QALoRA also works with different quantization methods (GPTQ, EETQ, AWQ, etc.):
```
BenjaminBossan: Currently, that's not true, is it?

examples/qalora_finetuning/README.md
```diff
@@ -0,0 +1,178 @@
...
+model = PeftModel.from_pretrained(base_model, peft_model_id)
+
+# Generate text
+input_text = "Hello, I'm a language model"
+inputs = tokenizer(input_text, return_tensors="pt").to(model.device)
+outputs = model.generate(**inputs, max_length=100)
+print(tokenizer.decode(outputs[0], skip_special_tokens=True))
+```
+
+## QALoRA vs. LoRA vs. DoRA
```
BenjaminBossan: I don't think there is a specific reason to compare QALoRA to DoRA, is there? I think the more interesting question for users would be: What's the difference between QALoRA and QLoRA? When should I use which, what trade offs are there?

examples/qalora_finetuning/README.md
```diff
@@ -0,0 +1,178 @@
+# QALoRA: Quantization-Aware Low-Rank Adaptation
+
+## Introduction
+[QALoRA](https://huggingface.co/papers/2309.14717) is a quantization-aware version of Low-Rank Adaptation that enables efficient fine-tuning of quantized large language models. QALoRA uses input feature pooling and a specialized grouping technique to work with quantized weights, significantly reducing memory requirements while preserving performance. QALoRA enables fine-tuning of models that would otherwise be too large for consumer GPUs.
+
```
BenjaminBossan, line 7: Let's add a note right at the beginning that right now, only GPTQ is supported and merging does not work yet.

examples/qalora_finetuning/README.md
```diff
@@ -0,0 +1,178 @@
...
+> **Note:** The current implementation differs from the original QA-LoRA paper's approach.
+
+While the QA-LoRA paper describes a direct weight modification technique using "beta shift" to modify quantized weights without full dequantization, this implementation uses a different approach:
+
+1. The quantized model is first dequantized to full precision
+2. The QALoRA adapter weights are then merged with the dequantized model
+3. The merged model must be re-quantized if quantization is still desired
+
+This implementation choice was made because **it yielded better performance in practice**, despite being less memory-efficient than the direct modification approach described in the paper.
```
BenjaminBossan: Out of curiosity: Is this your personal experience or is there a reference for this?

examples/qalora_finetuning/README.md
```diff
@@ -0,0 +1,178 @@
+# QALoRA: Quantization-Aware Low-Rank Adaptation
+
+## Introduction
+[QALoRA](https://huggingface.co/papers/2309.14717) is a quantization-aware version of Low-Rank Adaptation that enables efficient fine-tuning of quantized large language models. QALoRA uses input feature pooling and a specialized grouping technique to work with quantized weights, significantly reducing memory requirements while preserving performance. QALoRA enables fine-tuning of models that would otherwise be too large for consumer GPUs.
```
BenjaminBossan: I think for many users, it would be great if you could mention in a sentence what the advantage of quantization aware training is vs normal training with quantized models.

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,119 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
...
+                # This case should ideally not happen if input x matches layer's expected in_features
+                # Or, if dynamic in_features is truly supported, then pooling should adapt.
+                # For now, let's assume x_2d.shape[-1] is the correct in_features for this specific forward pass.
+                # If module.in_features is the strict definition, then an error or warning might be needed.
+                # Re-evaluate if in_features for pooling should be x_2d.shape[-1] or module.in_features
+                # Sticking to module.in_features as it's used for the divisibility check.
+                pass # Assuming x_2d.shape[-1] == module.in_features for this QALoRA path
+
+            x_grouped = x_2d.reshape(x_2d.shape[0], pooled_dim, group_size)
+            x_pooled = x_grouped.mean(dim=-1)
```
BenjaminBossan: Could [`nn.functional.avg_pool1d`](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.avg_pool1d.html) be used here?

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,119 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
...
+            # 2. Apply QALoRA specific scaling
+            x_pooled_scaled = x_pooled * paper_scaling_factor
+
+            lora_A_weight_reshaped = lora_A_weight.reshape(lora_A_weight.shape[0], pooled_dim, group_size)
+            lora_A_pooled_weight = lora_A_weight_reshaped.mean(dim=2)
+
+            intermediate = x_pooled_scaled @ lora_A_pooled_weight.t()
+            delta = intermediate @ lora_B_weight.t()
+            delta = delta * lora_scaling_coefficient # Apply LoRA scaling (alpha/r)
+        else:
```
BenjaminBossan: The `else` branch cannot be reached, right?

examples/qalora_finetuning/qalora_gptq_finetuning.py
```diff
@@ -0,0 +1,422 @@
...
+    DataCollatorForLanguageModeling,
+    GPTQConfig,
+    Trainer,
+    TrainingArguments,
+)
+
+from peft import LoraConfig, get_peft_model
+
+
+def check_adapter_gradients(model) -> None:
```
BenjaminBossan: I think this type of check could be moved to a unit test instead.

examples/qalora_finetuning/qalora_gptq_finetuning.py
```diff
@@ -0,0 +1,422 @@
...
+    total_grad_mem = 0
+    total_opt_mem = 0
+
+    print(f"{'Name':50} {'Shape':20} {'Dtype':10} {'Param MB':10} {'Grad MB':10} {'Requires Grad'}")
+    print("-" * 110)
+
+    for name, param in model.named_parameters():
+        param_size = param.numel() * param.element_size() / 1024**2  # in MB
+        grad_size = param_size if param.requires_grad else 0
+        opt_size = 0
```
BenjaminBossan: If the user uses neither Adam nor SGD, this would be shown as 0, which would be confusing. Instead, you could check if the value is 0 and then replace it, for instance, with `"unknown"`.

examples/qalora_finetuning/qalora_gptq_finetuning.py
```diff
@@ -0,0 +1,422 @@
...
+
+        if param.requires_grad:
+            if optimizer_name.lower() == "adam":
```
BenjaminBossan: How about also checking `"adamw"` which is quite popular?

examples/qalora_finetuning/qalora_gptq_finetuning.py
```diff
@@ -0,0 +1,422 @@
...
+        f"Parameter dtype is per-column above. "
+        f"\nTotal parameter memory: {total_param_mem:.2f} MB"
+        f"\nTotal gradient memory (training): {total_grad_mem:.2f} MB"
+        f"\nTotal optimizer state memory (training): {total_opt_mem:.2f} MB"
+        f"\nTotal memory for training: {total_param_mem + total_grad_mem + total_opt_mem:.2f} MB"
+        f"\nTotal memory for inference: {total_param_mem:.2f} MB"
+    )
+
+
+def load_or_quantize_model(
```
BenjaminBossan, line 24: Hmm, when I load a model that is already GPTQ-quantized like `"TheBloke/Llama-2-7b-GPTQ"`, this function still create a new folder and saves a copy of the model there, right? Let's avoid this.

examples/qalora_finetuning/qalora_gptq_finetuning.py
```diff
@@ -0,0 +1,422 @@
...
+    tokenizer = AutoTokenizer.from_pretrained(base_model, token=hf_token)
+    if tokenizer.pad_token is None:
+        tokenizer.pad_token = tokenizer.eos_token
+
+    # Load or quantize model
+    model = load_or_quantize_model(base_model, tokenizer)
+
+    # Configure LoRA
+    target_modules = (
+        lora_target_modules.split(",")
```
BenjaminBossan, line 189: Not super important, but `target_modules="all-linear"` is also a valid argument in PEFT.

examples/qalora_finetuning/qalora_gptq_finetuning.py
```diff
@@ -0,0 +1,422 @@
...
+        r=lora_r,
+        lora_alpha=lora_alpha,
+        target_modules=target_modules,
+        lora_dropout=lora_dropout,
+        bias="none",
+    )
+
+    # Get PEFT model with adapters
+    model = get_peft_model(model, lora_config)
+    model.save_pretrained(output_dir)
```
BenjaminBossan: Why save the checkpoint here?

examples/qalora_finetuning/qalora_gptq_finetuning.py
```diff
@@ -0,0 +1,422 @@
...
+    device = torch.device(device)
+    print(f"Using device: {device}")
+
+    # Load tokenizer
+    tokenizer = AutoTokenizer.from_pretrained(base_model, token=hf_token)
+    if tokenizer.pad_token is None:
+        tokenizer.pad_token = tokenizer.eos_token
+
+    # Load or quantize model
+    model = load_or_quantize_model(base_model, tokenizer)
```
BenjaminBossan: IMO it would make sense to expose the `bits` argument to the CLI.

examples/qalora_finetuning/qalora_gptq_finetuning.py
```diff
@@ -0,0 +1,422 @@
...
+        if param.requires_grad:
+            trainable_params += param.numel()
+
+    # Check for QALoRA modules
+    for name, module in model.named_modules():
+        if hasattr(module, "qalora_group_size"):
+            qalora_modules += 1
+            adapter_name = next(iter(module.qalora_group_size.keys()), None)
+            if adapter_name:
+                print(f"\nQALoRA module found: {name}")
```
BenjaminBossan: Hmm, printing all this information for each layer is a bit much IMO. How about only printing the summary?

examples/qalora_finetuning/qalora_gptq_finetuning.py
```diff
@@ -0,0 +1,422 @@
...
+
+    print("-" * 110)
+    print(f"{'TOTAL':50} {'':20} {'':10} {total_param_mem:10.2f} {total_grad_mem:10.2f}")
+    print(
+        f"Parameter dtype is per-column above. "
+        f"\nTotal parameter memory: {total_param_mem:.2f} MB"
+        f"\nTotal gradient memory (training): {total_grad_mem:.2f} MB"
+        f"\nTotal optimizer state memory (training): {total_opt_mem:.2f} MB"
+        f"\nTotal memory for training: {total_param_mem + total_grad_mem + total_opt_mem:.2f} MB"
+        f"\nTotal memory for inference: {total_param_mem:.2f} MB"
```
BenjaminBossan: I think there is an error somewhere in the calculation. When running this with `TheBloke/Llama-2-7b-GPTQ`, I get:

```
Parameter dtype is per-column above. 
Total parameter memory: 576.76 MB
Total gradient memory (training): 76.25 MB
Total optimizer state memory (training): 152.50 MB
Total memory for training: 805.51 MB
Total memory for inference: 576.76 MB
```

These values are too low, both from a theoretical viewpoint and from what I measure in practice (of course, in practice we need memory for hidden states, which are difficult to estimate).

When I use [my own script](https://github.com/BenjaminBossan/presentations/blob/main/2024-07-11-europython/utils.py) for checking, with int4 quantized `meta-llama/Llama-2-7b-hf`, Adam, and LoRA rank 16, what I get is:

```json
{
  "number of parameters": {
    "Embedding.weight": 131072000,
    "Linear.weight": 6476005376,
    "lora": 39976960,
    "LlamaRMSNorm.weight": 266240
  },
  "number of bytes": {
    "Embedding.weight": 262144000,
    "Linear.weight": 3238002688,
    "lora": 159907840,
    "LlamaRMSNorm.weight": 532480
  },
  "number of bytes (readable)": {
    "Embedding.weight": "250.0 MB",
    "Linear.weight": "3.02 GB",
    "lora": "152.5 MB",
    "LlamaRMSNorm.weight": "520.0 KB"
  },
  "total number of parameters w/o LoRA": 6607343616,
  "total number of parameters w/  LoRA": 6647320576,
  "total size w/o LoRA": 3500679168,
  "total size w/  LoRA": 3660587008,
  "total size w/o LoRA (readable)": "3.26 GB",
  "total size w/  LoRA (readable)": "3.41 GB",
  "memory required for full fine-tuning": "13.04 GB*",
  "memory required for LoRA fine-tuning": "3.71 GB"
}
```
This does not take into account QALoRA, but I don't think the difference can be that big.

examples/qalora_finetuning/qalora_gptq_finetuning.py
```diff
@@ -0,0 +1,422 @@
...
+        grad_size = param_size if param.requires_grad else 0
+        opt_size = 0
+
+        if param.requires_grad:
+            if optimizer_name.lower() == "adam":
+                opt_size = param_size * 2  # Adam keeps 2 extra states (m, v)
+            elif optimizer_name.lower() == "sgd":
+                opt_size = param_size  # SGD with momentum keeps 1 extra state
+
+        print(
```
BenjaminBossan: Again, I think printing this for each layer is a bit of information overkill.

examples/qalora_finetuning/README.md
```diff
@@ -0,0 +1,178 @@
...
+peft_model.save_pretrained("qalora-llama-2-7b")
+```
+
+To use QALoRA, simply set `use_qalora = True` and specify a `qalora_group_size` in your LoRA configuration. The group size controls the memory/performance tradeoff - smaller values use less memory but may affect performance.
+
+## Command Line Examples
+
+Run the finetuning script with a GPTQ quantized model:
+```bash
+python examples/qalora_finetuning/qalora_gptq_finetuning.py \
```
BenjaminBossan: When I run this locally, I get a train loss of 0.0, can you replicate?

examples/qalora_finetuning/qalora_gptq_finetuning.py
```diff
@@ -0,0 +1,422 @@
...
+        if lora_target_modules
+        else ["q_proj", "k_proj", "v_proj", "o_proj", "gate_proj", "up_proj", "down_proj"]
+    )
+
+    print("use_qalora", use_qalora)
+    lora_config = LoraConfig(
+        task_type="CAUSAL_LM",
+        use_dora=use_dora,
+        use_qalora=use_qalora,
+        qalora_group_size=8,  # Explicitly set group size for QALoRA
```
BenjaminBossan: Would it make sense to expose this argument to the CLI?

src/peft/tuners/lora/gptq.py
```diff
@@ -53,9 +56,25 @@ def __init__(
...
             use_rslora=use_rslora,
             use_dora=use_dora,
+            use_qalora=use_qalora,
             lora_bias=lora_bias,
+            qalora_group_size=qalora_group_size,
         )
 
+    def resolve_lora_variant(self, *, use_dora: bool, use_qalora: bool, **kwargs) -> Optional[LoraVariant]:
+        if use_dora and use_qalora:
+            NotImplementedError
```
BenjaminBossan:
```python
            raise NotImplementedError(...)
```

src/peft/tuners/lora/gptq.py
```diff
@@ -53,9 +56,25 @@ def __init__(
...
+        elif use_dora:
+            from .variants import DoraLinearVariant
+            variant = DoraLinearVariant()
+        elif use_qalora:
+            if self.in_features % kwargs["qalora_group_size"] == 0:
```
BenjaminBossan: Let's not perform the check here. Instead, let's move it inside `QALoraLinearVariant.init`.

BenjaminBossan (changes requested):
Thanks a lot for all the updates. I still found a couple of issues, please check my comments. I also checked and ran the script, where I also found some areas for improvement.

As for testing, I think it would be enough to have something very similar to [this one](https://github.com/huggingface/peft/blob/759bb70acea8486b3a493792a5b6ba80b5217fb4/tests/test_gpu_examples.py#L1748) with QALoRA enabled.

Also, before pushing, don't forget to run `make style` to make the linter happy.

gapsong:
More tests are still needed for comprehensive coverage and some of the existing comments are a work in progress and will be refined.

BenjaminBossan:
> More tests are still needed for comprehensive coverage and some of the existing comments are a work in progress and will be refined.

Okay, ich warte dann, bis diese fertig sind bevor ich erneut reviewe ;)

gapsong:
@BenjaminBossan du kannst wieder reinschauen.

Ich habe die Initialisierung nochmal überarbeitet. Jetzt wird lora_A direkt beim Init durch eine neue Instanz mit der gepoolten Eingabedimension ersetzt. Vorher hatten wir lora_A noch in der alten Shape und haben das Pooling erst im Forward angewendet, so wie es auch ursprünglich im Paper gemacht wurde.

Mit der neuen Variante übernimmt lora_A das Pooling bereits beim Init, wodurch die Dimensionen und damit die Anzahl der trainierbaren Parameter bei QALoRA tatsächlich kleiner werden als bei QLoRA. Vorher waren die "trainable params" bei QLoRA und QALoRA identisch, jetzt ist QALoRA sparsamer.

Eine interessante Beobachtung: Der Speicherverbrauch ist trotzdem sehr ähnlich zu QLoRA, obwohl weniger trainierbare Parameter vorhanden sind. Das scheint aber durch die zusätzlichen Pooling-Tensoren im Forward zustande zu kommen. Vielleicht kannst du mir nochmal bestätigen, ob das so korrekt und zu erwarten ist.

examples/qalora_finetuning/README.md
BenjaminBossan: Still missing?

examples/qalora_finetuning/README.md
BenjaminBossan: Question still open

examples/qalora_finetuning/README.md
BenjaminBossan: Question still open

examples/qalora_finetuning/README.md
BenjaminBossan: Question still open

src/peft/tuners/lora/config.py
```diff
@@ -494,6 +494,28 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    use_qalora: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "It is only implemented in GPTQ for now. Enable <a href='https://huggingface.co/papers/2309.14717'>Quantization-Aware Low-Rank Adaptation (QALoRA)</a>. This technique combines quantization-aware training "
```
BenjaminBossan: Let's ensure that each line is 120 chars max.

src/peft/tuners/lora/gptq.py
BenjaminBossan: Still open

tests/test_gpu_examples.py
```diff
@@ -1168,201 +1168,32 @@ def test_initialize_dora_with_bnb_on_cpu(self, kbit):
...
-            trainer.train()
-
-            model.cpu().save_pretrained(tmp_dir)
-
-            assert "adapter_config.json" in os.listdir(tmp_dir)
-            assert SAFETENSORS_WEIGHTS_NAME in os.listdir(tmp_dir)
-
-            # assert loss is not None
-            assert trainer.state.log_history[-1]["train_loss"] is not None
+    # In der PeftGPTQGPUTests Klasse hinzufügen:
```
BenjaminBossan: Remove

tests/test_gpu_examples.py
```diff
@@ -1168,201 +1168,32 @@ def test_initialize_dora_with_bnb_on_cpu(self, kbit):
         weights_not_cpu = [name for name, p in peft_model.named_parameters() if p.device != torch.device("cpu")]
         assert not weights_not_cpu
 
-    @pytest.mark.single_gpu_tests
-    def test_causal_lm_training_vera(self):
```
BenjaminBossan: Hmm, it looks like a bunch of unrelated tests where moved around. I don't think that should be necessary and it also makes reviewing a lot harder. Could you please ensure that the existing tests remain in place and only new tests for GPTQ-QALoRA are added?

BenjaminBossan:
@gapsong Please ping me once the PR is ready for the next review.

gapsong:
@BenjaminBossan I have adjusted the code expect for the `resolve_lora_variant` 
Could you convince me that  `if self.in_features % kwargs["qalora_group_size"] == 0:`
belongs into the `init()`?

BenjaminBossan:
Thanks for the updates. I tried the example script again for 5 steps, using `TheBloke/Llama-2-7B-GPTQ`, and found that the `train_loss` is 0.0. Can you confirm that? It doesn't matter if `use_qalora` is enabled or not.

> Could you convince me that `if self.in_features % kwargs["qalora_group_size"] == 0:` belongs into the `init()`?

For, mostly for these reasons:

1. These types of checks are typically done during init methods in PEFT, so it is consistent.
2. It is a QALoRA related piece of code, thus it makes sense to put in the QALoRA specific code.
3. `resolve_lora_variant` should focus on what the name says, deciding which LoRA variant is used.

gapsong:
@BenjaminBossan I’m done and I found the “bug”:
The qalora_group_size was set to 8, which was too small. I’ve changed the default value to 32.

I also adjusted the resolve_lora_variant function:
Training with "lora" is no longer the default if the dimensions do not fit—specifically, when module.in_features % kwargs["qalora_group_size"] != 0. Instead, an error is now thrown if the parameters do not match.

I hope this time works out :)

examples/qalora_finetuning/README.md
```diff
@@ -0,0 +1,157 @@
...
+```bash
+python qalora_gptq_finetuning.py \
+    --base_model "TheBloke/Llama-2-13b-GPTQ" \
+    --output_dir "PATH_TO_OUTPUT_DIR" \
+    --batch_size 1 \
+    --num_epochs 3 \
+    --learning_rate 3e-4 \
+    --cutoff_len 512 \
+    --use_qalora \
+    --qalora_group_size 32 \
```
BenjaminBossan, line 78: Thanks for updating to a better default. Do you think that `qalora_group_size=16` is still a good choice for the default value in `LoraConfig`?

src/peft/tuners/lora/gptq.py
BenjaminBossan: This check can now be removed, right?

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,98 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
...
+
+        Args:
+            module (Linear): The linear module to be adapted.
+            adapter_name (str): The name of the adapter.
+            **kwargs: Additional keyword arguments.
+                qalora_group_size (int): The size of groups for pooling. This is expected to be passed.
+        """
+        if "qalora_group_size" not in kwargs:
+            raise ValueError(
+                "QALoraLinearVariant.init expects 'qalora_group_size' to be provided in kwargs."
```
BenjaminBossan:
```suggestion
                "`use_qalora=True` requires 'qalora_group_size' to be provided in kwargs."
```

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,98 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
...
+                " Please ensure it is passed from the LoraConfig."
+            )
+
+        if module.in_features is not None and module.in_features % kwargs["qalora_group_size"] != 0:
+            raise ValueError(
+                f"QALoraLinearVariant.init expects module.in_features ({module.in_features}) to be divisible by 'qalora_group_size' ({kwargs['qalora_group_size']})"
```
BenjaminBossan:
```suggestion
                f"`use_qalora=True` requires `module.in_features` ({module.in_features}) to be divisible by 'qalora_group_size' ({kwargs['qalora_group_size']})"
```

gapsong:
@BenjaminBossan included your suggestions!
Thank you

gapsong:
@BenjaminBossan something seems off with the GPU memory, when using QA_Lora I am investigating the problem at the moment.

gapsong:
@BenjaminBossan GPU problem was resolved. I removed `torch.nn.functional.avg_pool1d` since the peak memory needed for the unsqueezed layers were to high

tests/test_custom_models.py
```diff
@@ -2820,6 +2820,51 @@ def test_requires_grad_lora_different_targets(self):
             "base_model.model.lin1.lora_B.adapter1.weight",
         )
 
+    def test_requires_grad_qalora_same_targets(self):
```
BenjaminBossan: Hmm, for some reason I haven't really noticed this test. IMO it is not necessary and can be removed. This is because QALoRA does not change the handling of multiple adapters, so there is not really a reason to believe there could be anything wrong there. We also don't check other quantization methods here. Unless you had a specific reason to add this test, I'd suggest to just remove it.

BenjaminBossan (commented):
Thanks for the latest update. I just have a small comment, otherwise the PR LGTM.

I also ran the example locally with and without QALoRA and the difference in memory usage was negligible, which I think is what we hoped to see.

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,101 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
...
+        """
+        if "qalora_group_size" not in kwargs:
+            raise ValueError(
+                "`use_qalora=True` requires 'qalora_group_size' to be provided in kwargs."
+                " Please ensure it is passed from the LoraConfig."
+            )
+
+        if module.in_features is not None and module.in_features % kwargs["qalora_group_size"] != 0:
+            raise ValueError(
+                f"`use_qalora=True` requires `module.in_features` ({module.in_features}) to be divisible by 'qalora_group_size' ({kwargs['qalora_group_size']})"
```
BenjaminBossan: Could you please add line breaks to this long line?

src/peft/tuners/lora/variants.py
```diff
@@ -130,6 +130,101 @@ def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.
         return result
 
 
+class QALoraLinearVariant(LoraVariant):
```
BenjaminBossan: Please move this class to the bottom of the file, as is it sits between different DoRA classes.

BenjaminBossan (changes requested):
I did a final pass and discovered two small things that need fixing, otherwise LGTM. Thanks for your patience.

gapsong:
@BenjaminBossan what do I have to do here to resolve the errors or is it something out of my scope?

BenjaminBossan:
> what do I have to do here to resolve the errors or is it something out of my scope?

You mean the MacOS errors? Yeah, they're unrelated and should be fixed with the next transformers release.

gapsong:
@BenjaminBossan shall we merge it then?

BenjaminBossan:
> shall we merge it then?

Did you see my [last comment](https://github.com/huggingface/peft/pull/2571#pullrequestreview-2957301813)?

gapsong:
> > shall we merge it then?
> 
> Did you see my [last comment](https://github.com/huggingface/peft/pull/2571#pullrequestreview-2957301813)?

I have missed it. It should be ready now @BenjaminBossan

thistleknot:
thank you, now training with qa-lora =D

PR #2572, opened 2025-06-05
Simple variant application test

githubnemo:
This change adds a place for future lora variant tests to linger and also adds a basic test that checks whether the variant was correctly applied to the lora layer if so requested. This does not cover quantized layers yet but it should be similar to add thanks to the mapping.

Since we're expecting a lot more variants to be added in the future it is probably sensible to start early in establishing a place for testing.

### DIFF
--- a/tests/test_lora_variants.py
+++ b/tests/test_lora_variants.py
@@ -120,7 +120,7 @@ class TestLoraVariants:
         """Ensure that the parameters added by the DoRA variant are participating in the output computation."""
         layer_names = ["linear1", "linear2", "conv1d", "conv2d", "embedding"]
         peft_config = LoraConfig(target_modules=layer_names, use_dora=True)
-        base_model, peft_config, peft_model = self.custom_model_with_loss_backpropagated(peft_config)
+        base_model, peft_model = self.custom_model_with_loss_backpropagated(peft_config)
 
         for layer in layer_names:
             assert getattr(peft_model.base_model.model, layer).lora_magnitude_vector["default"].weight.grad is not None
--- a/tests/test_lora_variants.py
+++ b/tests/test_lora_variants.py
@@ -1,0 +1,90 @@
+import pytest
+from torch import nn
+
+from peft import LoraConfig, get_peft_model
+from peft.tuners.lora.layer import Conv1d as LoraConv1d
+from peft.tuners.lora.layer import Conv2d as LoraConv2d
+from peft.tuners.lora.layer import Embedding as LoraEmbedding
+from peft.tuners.lora.layer import Linear as LoraLinear
+from peft.tuners.lora.variants import (
+    DoraConv1dVariant,
+    DoraConv2dVariant,
+    DoraEmbeddingVariant,
+    DoraLinearVariant,
+)
+
+
+class CustomModel(nn.Module):
+    """pytorch module that contains common targetable layers (linear, embedding, conv, ...)"""
+
+    def __init__(self, num_embeddings=100, embedding_dim=16, num_classes=10):
+        super().__init__()
+        self.embedding = nn.Embedding(num_embeddings, embedding_dim)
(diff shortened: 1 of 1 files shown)


tests/test_lora_variants.py
```diff
@@ -0,0 +1,90 @@
...
+        [
+            (
+                "DoRA 1",
+                "dora",
+                LoraConfig,
+                {"target_modules": ["linear0", "linear1", "conv1d", "conv2d", "embedding"], "use_dora": True},
+            ),
+        ],
+    )
+    def test_variant_is_applied_to_layers(self, test_name, variant_name, config_cls, config_kwargs):
```
BenjaminBossan: `test_name` is not used

tests/test_lora_variants.py
```diff
@@ -0,0 +1,90 @@
+import pytest
```
BenjaminBossan, line 15: Copyright notice is missing.

githubnemo:
I've opted for a variant-specific test, the generic test would be too complicated I feel.

tests/test_lora_variants.py
```diff
@@ -88,3 +103,24 @@ def test_variant_is_applied_to_layers(self, test_name, variant_name, config_cls,
                 continue
 
             assert isinstance(module.lora_variant["default"], expected_variant_type)
+
+    def custom_model_with_loss_backpropagated(self, config_cls, config_kwargs):
```
BenjaminBossan: Why not pass the `config` instance directly?

tests/test_lora_variants.py
```diff
@@ -55,19 +71,18 @@ def forward(self, input_ids, dummy_image_input):
 }
 
 
+TEST_CASES = [
+    (
+        "dora",
+        LoraConfig,
+        {"target_modules": ["linear1", "linear2", "conv1d", "conv2d", "embedding"], "use_dora": True},
```
BenjaminBossan, line 78: Would it make sense to just have the config instance here?

tests/test_lora_variants.py
```diff
@@ -88,3 +103,24 @@ def test_variant_is_applied_to_layers(self, test_name, variant_name, config_cls,
...
+        x, y = torch.ones(10, 10).long(), torch.ones(10, 1, 10, 10)
+        out = peft_model(x, y)
+        loss = out.sum()
+        loss.backward()
+
+        return base_model, peft_config, peft_model
+
+    def test_dora_params_have_gradients(self):
+        config_cls = LoraConfig
+        config_kwargs = {"target_modules": ["linear1", "linear2", "conv1d", "conv2d", "embedding"], "use_dora": True}
```
BenjaminBossan: How about assigning the `target_modules` to a variable that can be used both here and below for the loop?

tests/test_lora_variants.py
```diff
@@ -88,3 +103,24 @@ def test_variant_is_applied_to_layers(self, test_name, variant_name, config_cls,
...
+        peft_model = get_peft_model(base_model, peft_config)
+
+        x, y = torch.ones(10, 10).long(), torch.ones(10, 1, 10, 10)
+        out = peft_model(x, y)
+        loss = out.sum()
+        loss.backward()
+
+        return base_model, peft_config, peft_model
+
+    def test_dora_params_have_gradients(self):
```
BenjaminBossan, line 119: Let's add a small comment, like: `# This test ensures that the the DoRA components participate in the output computation` or so.

tests/test_lora_variants.py
githubnemo: This pattern is used in all other test suites as well and I think it has the benefit that you don't have to worry about leaking modifications between tests because someone forgot a copy. So I'd rather keep it.

BenjaminBossan (approved):
Thanks, LGTM. Can be merged when CI is green (or at least the parts that can be green).

PR #2574, opened 2025-06-06
FIX: Transformers VLM architecture changes

BenjaminBossan:
Follow up to #2554
See discussion in https://github.com/huggingface/transformers/pull/38627

## Description

To quote:

> transformers PR #37033 re-arranges the way visual language models are built by moving the LM head from the language model to the top-level VLM (among other things).

A consequence of this is that the keys in the PEFT `state_dict` now also follow the new architecture. This means that:

1. If a PEFT checkpoint was saved with the old architecture but is loaded with the new architecture, loading fails.
2. If a PEFT checkpoint was saved with the new architecture but is loaded with the old architecture, loading fails.

Point 1. can be addressed by making use of the newly added `_checkpoint_conversion_mapping` attribute for models with the new architecture. In transformers, this is used to map old model `state_dict`s to the new `state_dict` format. In PEFT, with some fiddling, we can use the same mapping to make old PEFT `state_dict`s compatible with the new architecture (backwards compatibility).

However, 2. is not easily addressed. We would need a reverse mapping for this. This could be easily derived from `_checkpoint_conversion_mapping`, but since this attribute doesn't exist on old models, we cannot do that. Therefore, new checkpoints created with PEFT on these models won't load successfully when users use old transformers (forward compatibility).

These cases are covered by the added unit tests, which means that the test covering case 2 currently fails.

If we could reliably detect that we are in case 2, we could warn the user and advise them to upgrade transformers, but I don't know if it's possible to figure this out.

## Notes

We skip prompt learning methods when applying the mapping. This is because they don't have the `"base_model.model."` prefix, which we need to remove before mapping. Instead just using `"base_model."`. This could be fine, we could only remove `"base_model."`, However, the subsequent sub-module could also be called `"model"`, resulting in what looks like `"base_model.model."`. To avoid this confusion, we skip prompt learning. Since it should be applied to the language model part directly and applies itself on the outer model (unlike LoRA et al.), skipping should be fine.

We also allow users to pass their own `key_mapping` to `from_pretrained` and `load_adapter`, though the documentation advises against it. This argument could theoretically be used as a workaround in case there is indeed an issue with prompt learning s`tate_dicts`.

Apart from these changes, I also made a small change to account for https://github.com/huggingface/transformers/issues/38017#issuecomment-2935889679.

### DIFF
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -583,10 +583,14 @@ def load_peft_weights(
         remapped_adapters_weights = adapters_weights
     else:
         # See discussion in https://github.com/huggingface/transformers/pull/38627
+        # Remap adapter weight names according to the provided key_mapping.
         remapped_adapters_weights = {}
-        prefix = "base_model.model."
         for key, val in adapters_weights.items():
-            if not key.startswith(prefix):
+            if key.startswith("base_model.model."):
+                prefix = "base_model.model."
+            elif key.startswith("base_model."):
+                prefix = "base_model."
+            else:
                 raise ValueError(
                     "An error occurred while trying to load a PEFT state_dict with key_mapping. This should not "
                     "happen. Please open an issue on https://github.com/huggingface/peft/issues and report the error."
--- a/tests/test_initialization.py
+++ b/tests/test_initialization.py
@@ -3770,8 +3770,8 @@ class TestLoadPeftKeyMapping:
     # loaded with the changed architecture. Unfortunately, new checkpoints cannot be loaded with the old architecture,
     # the corresponding test is marked as xfail.
 
-    # Note: We only test prefix tuning (prompt learning method) and LoRA (non-prompt learning method) as the other PEFT
-    # methods should work the same way. It would be excessive to test all of them here.
+    # Note: We only test prefix tuning (prompt learning method), LoRA (non-prompt learning method), and VBLoRA (shared
+    # parameters) as the other PEFT methods should work the same way. It would be excessive to test all of them here.
 
     @pytest.fixture
     def fake_model_config(self):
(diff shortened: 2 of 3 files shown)


tests/test_initialization.py
```diff
@@ -3757,3 +3759,299 @@ def test_scaling_multiple_adapters(self, model):
...
+                if item in self:
+                    return self[item]
+                raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{item}'")
+
+        return FakeConfig()
+
+    @pytest.fixture
+    def old_model(self, fake_model_config):
+        # create a small model that mimics the old architecture of, for instance, Qwen/Qwen2-VL-2B-Instruct
+        # Qwen2VLForConditionalGeneration(
```
zucchini-nlp, line 3793: maybe we can host old and new style tiny models on the hub with custom code, WDYT? And re-use the same models to add tests on transformers repo

zucchini-nlp (approved):
Meant to approve

tests/test_initialization.py
BenjaminBossan: Sounds like a good idea for PRs like https://github.com/huggingface/transformers/pull/38627.

src/peft/utils/save_and_load.py
```diff
@@ -572,4 +579,27 @@ def get_hub_filename(use_safetensors=True):
     else:
         adapters_weights = torch_load(filename, map_location=torch.device(device))
 
-    return adapters_weights
+    if not key_mapping:
+        remapped_adapters_weights = adapters_weights
+    else:
+        # See discussion in https://github.com/huggingface/transformers/pull/38627
```
githubnemo, line 585:
```suggestion
        # Remap adapter weights according to the provided key_mapping.
        # Useful for providing backward compatibility with old checkpoints when the model architecture changed.
        # See discussion in https://github.com/huggingface/transformers/pull/38627
```

src/peft/utils/save_and_load.py
```diff
@@ -572,4 +579,27 @@ def get_hub_filename(use_safetensors=True):
...
+        remapped_adapters_weights = {}
+        prefix = "base_model.model."
+        for key, val in adapters_weights.items():
+            if not key.startswith(prefix):
+                raise ValueError(
+                    "An error occurred while trying to load a PEFT state_dict with key_mapping. This should not "
+                    "happen. Please open an issue on https://github.com/huggingface/peft/issues and report the error."
+                )
```
githubnemo: This would fail for VBLoRA, I think, since that implementation stores weights in `base_model.vblora_vector_bank`. Also not sure about adapters that target encoder/decoder models like T5 where there is no `base_model.model` but `base_model.{shared,encoder,decoder}`.

This remembers me of the auxiliary training wrapper saving mechanism. What we're doing there is to iterate over all the possible prefixes of the model (via `model.named_modules()`) and use those as prefix. Would that work here as well?

BenjaminBossan:
@githubnemo I think the concerns should be addressed, please check again.

githubnemo (approved):
Thanks for the fix and the additions. LGTM

steveepreston:
Guys, is this bugfix includes in peft `0.17.1`?
I'm still getting missing adapter keys on latest version

githubnemo:
> Guys, is this bugfix includes in peft `0.17.1`? I'm still getting missing adapter keys on latest version

~~Yep, it should be included in the release. Can you give a bit more context / a reproducer so we can help you better?~~

Nevermind, apparently the discussion is happening in #2699.

PR #2575, opened 2025-06-08
OFT: several improvements to make OFT faster and more memory efficient

zqiu24:
Hi there,

I want to make some changes to OFT. The major problem of OFT is it is relatively slow and not memory efficient for large models, the commit does not change the training logic but makes several improvements to make it faster.

 **Core Improvements:**
  - Enhanced OFT layer implementation with the option of using Cayley-Neumann parametrization
  - Input-centric implementation to reduce memory and compute
  - Adding support to different quantization: bnb, awe, aqlm, hqq, eetq, gptq etc.
  
Hopefully, you can let me add these improvements to the OFT training code.

The below is some test on when performing fine-tuning with Qwen2.5-7B.

oftv2

Best,


zqiu24:
@BenjaminBossan Hi, can you please make a review? It would be great. Best,

src/peft/tuners/oft/layer.py
```diff
@@ -189,20 +442,16 @@ def update_layer(self, adapter_name, r, oft_block_size, module_dropout, coft, ep
...
-        self.coft[adapter_name] = coft
-        self.block_share[adapter_name] = block_share
-        self.eps[adapter_name] = eps * math.ceil(self.out_features / r) * math.ceil(self.out_features / r)
-
         # Create weights with provided shape
+        n_elements = oft_block_size * (oft_block_size - 1) // 2
         if block_share:
-            self.oft_r[adapter_name] = nn.Parameter(
-                torch.empty(1, math.ceil(self.in_features / r), math.ceil(self.in_features / r))
+            self.oft_R[adapter_name] = OFTRotationModule(
```
BenjaminBossan: Would this change still allow to load old checkpoints and produce the same results or is it backwards incompatible?

src/peft/tuners/lora/layer.py
```diff
@@ -795,6 +795,8 @@ def __init__(
             lora_bias=lora_bias,
         )
 
+        breakpoint()
```

src/peft/tuners/oft/aqlm.py
```diff
@@ -0,0 +1,101 @@
+# Copyright 2024-present the HuggingFace Inc. team.
```
BenjaminBossan:
```suggestion
# Copyright 2025-present the HuggingFace Inc. team.
```

Here and below.

src/peft/tuners/oft/config.py
```diff
@@ -67,9 +67,9 @@ class OFTConfig(PeftConfig):
             Whether to share the OFT parameters between blocks or not. This is `False` by default.
     """
 
-    r: int = field(default=8, metadata={"help": "OFT rank, number of OFT blocks per injected layer."})
+    r: int = field(default=0, metadata={"help": "OFT rank, number of OFT blocks per injected layer."})
```
BenjaminBossan, line 70: Let's only change the defaults if really necessary to keep the change backwards compatible. So let's leave this value at 8, `oft_block_size` at 0, and set `use_cayley_neumann` to `False`.

src/peft/tuners/oft/layer.py
```diff
@@ -89,15 +297,13 @@ def __init__(self, base_layer: nn.Module, **kwargs) -> None:
         base_layer: the pretrained model layer
         """
         self.base_layer = base_layer
-        # OFT info
-        self.oft_r = nn.ParameterDict({})
-        self.oft_s = nn.ParameterDict({})
+        self.oft_R = nn.ParameterDict({})
```
BenjaminBossan: If we want to rename parameters, we would need to ensure that during model loading, names are also remapped, otherwise the change is backwards incompatible. As an example where we remapped parameter names, check this DoRA code:

https://github.com/huggingface/peft/blob/759bb70acea8486b3a493792a5b6ba80b5217fb4/src/peft/utils/save_and_load.py#L108-L119

Note that ideally, we don't rename the parameters at all, even if the new names are more descriptive.

BenjaminBossan (changes requested):
Thanks a lot for this big update to OFT. I didn't have time for an in-depth review yet, but I did a high level review and have some comments, please check.

My biggest concern right now is that the PR is backwards incompatible for the following reasons:

1. a new module, `OFTRotationModule`, is being used
2. a new option, `use_cayley_neumann` is added and enabled by default
3. some parameters are renamed

We should instead strife to make the change backwards compatible. This would mean:

1. Not sure if possible, but initialize `OFTRotationModule` in a way that the module works the same as previously. Remap `state_dict` parameters if necessary.
2. Default to `use_cayley_neumann=False`.
3. Don't rename the parameters or ensure that `state_dict`s are mapped accordingly.

I think it's fine not to concern ourselves with forwards compatibility. That means it's not necessary for OFT checkpoints created after this PR is merged to be compatible with older PEFT versions.

As for the quantized layers, also thanks a lot for adding all of them. I haven't checked the details yet. Maybe it would be good to move those to a separate, self-contained PR, WDYT?

Apart from this, we'll also have to add some tests for the new parameters and quantized layers, but let's first deal with the points mentioned above.

zqiu24:
> Thanks a lot for this big update to OFT. I didn't have time for an in-depth review yet, but I did a high level review and have some comments, please check.
> 
> My biggest concern right now is that the PR is backwards incompatible for the following reasons:
> 
> 1. a new module, `OFTRotationModule`, is being used
> 2. a new option, `use_cayley_neumann` is added and enabled by default
> 3. some parameters are renamed
> 
> We should instead strife to make the change backwards compatible. This would mean:
> 
> 1. Not sure if possible, but initialize `OFTRotationModule` in a way that the module works the same as previously. Remap `state_dict` parameters if necessary.
> 2. Default to `use_cayley_neumann=False`.
> 3. Don't rename the parameters or ensure that `state_dict`s are mapped accordingly.
> 
> I think it's fine not to concern ourselves with forwards compatibility. That means it's not necessary for OFT checkpoints created after this PR is merged to be compatible with older PEFT versions.
> 
> As for the quantized layers, also thanks a lot for adding all of them. I haven't checked the details yet. Maybe it would be good to move those to a separate, self-contained PR, WDYT?
> 
> Apart from this, we'll also have to add some tests for the new parameters and quantized layers, but let's first deal with the points mentioned above.
> 
> PS: If the prospect of making these changes backwards compatible is too much, we can also consider creating a new PEFT method like "OFTv2" and leave the current OFT as is.

Dear Benjamin,

Thank you so much for this quick reply and the suggestions. I am actually the author of both OFT and this newer version of OFT (the paper will be released on arxiv once the PR gets accepted). 

I totally understand the concerns with the backward compatibility, here are some explanations for the changes:

I changed to add the module `OFTRotationModule` and the name to oft_R to be more consistent with the naming of LoRA (lora_A and lora_B). I locally tested and it seems this is also necessary if we do peft with qlora/qoft (fine-tune quantized layers) + fsdp, otherwise it gets some data type problem with the distributed setting. It also avoids the definition of OFT training operations for each differently quantized layers seperately.

Personally I hope and think that it would be great if we could leave it as one OFTConfig (instead of adding another OFTConfigV2) to avoid confusions for the user, because essentially the algorithm does not change but is simply much faster and also supports the fine-tuning quantized layers. Is it possible to simply adding a check and detect if the checkpoint is an old one, the user should instead pip install an older version? 

Best,

src/peft/tuners/oft/layer.py
```diff
@@ -89,15 +305,13 @@ def __init__(self, base_layer: nn.Module, **kwargs) -> None:
...
+        self.oft_block_size = {}
         self.r = {}
         self.oft_block_size = {}
         self.oft_dropout = nn.ModuleDict({})
-        self.coft = {}
-        self.eps = {}
-        self.block_share = {}
+        # For Embedding layer
+        self.oft_embedding_R = nn.ParameterDict({})
```
BenjaminBossan: Is not used, right?

src/peft/tuners/oft/layer.py
```diff
@@ -106,20 +320,48 @@ def __init__(self, base_layer: nn.Module, **kwargs) -> None:
...
         base_layer = self.get_base_layer()
-
         if isinstance(base_layer, nn.Linear):
             in_features, out_features = base_layer.in_features, base_layer.out_features
         elif isinstance(base_layer, nn.Conv2d):
             in_features, out_features = base_layer.in_channels, base_layer.out_channels
+        elif isinstance(base_layer, nn.MultiheadAttention):
+            if not base_layer._qkv_same_embed_dim:
+                raise ValueError(f"Only same dim for query/key/value is supported as of now for {self.__class__}.")
+            in_features, out_features = base_layer.embed_dim, 3 * base_layer.embed_dim
```
BenjaminBossan: MHA is not supported, so this can be removed, right?

src/peft/tuners/oft/layer.py
```diff
@@ -133,19 +375,50 @@ def scale_layer(self, scale: float) -> None:
...
     def unscale_layer(self, scale=None) -> None:
         for active_adapter in self.active_adapters:
-            if active_adapter not in self.oft_r.keys():
+            if active_adapter not in self.oft_R.keys():
                 continue
 
             warnings.warn("Unscaling operation for OFT not supported! Keeping scale to 1.")
 
-    def update_layer(self, adapter_name, r, oft_block_size, module_dropout, coft, eps, block_share, init_weights):
+    def _check_forward_args(self, x, *args, **kwargs):
```
BenjaminBossan: This method is used in LoRA to check that we can perform mixed adapter predictions (i.e. in the same batch, having some samples with adapter X and other samples with adapter Y. See:
https://github.com/huggingface/peft/blob/a27406c26da0a1a4ceba2a98ac85c94e30824694/src/peft/tuners/lora/layer.py#L537
As this is not supported by OFT, we can remove this method.

src/peft/tuners/oft/layer.py
```diff
@@ -405,23 +649,13 @@ def get_delta_weight(self, adapter_name) -> tuple[torch.Tensor, torch.Tensor]:
...
-
-        orth_rotate = self._cayley_batch(oft_r)
-        weight = self._block_diagonal(orth_rotate, rank)
-
-        return weight, oft_s
+        return self.oft_R[adapter_name].get_weight()
 
     def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
+        self._check_forward_args(x, *args, **kwargs)
+        adapter_names = kwargs.pop("adapter_names", None)
```
BenjaminBossan: As mentioned above, since OFT does not support `_mixed_batch_forward`, these lines can be safely removed.

src/peft/tuners/oft/model.py
```diff
@@ -126,27 +134,37 @@ def _create_and_replace(
...
         if not isinstance(target, OFTLayer):
-            new_module = self._create_new_module(oft_config, adapter_name, target, **kwargs)
+            device_map = self.model.hf_device_map if hasattr(self.model, "hf_device_map") else None
+            new_module = self._create_new_module(oft_config, adapter_name, target, device_map=device_map, **kwargs)
             if adapter_name not in self.active_adapters:
                 # adding an additional adapter: it is not automatically trainable
                 new_module.requires_grad_(False)
             self._replace_module(parent, target_name, new_module, target)
         else:
+            breakpoint()
```
BenjaminBossan: Remove

src/peft/tuners/oft/layer.py
```diff
@@ -220,63 +501,16 @@ def reset_oft_parameters(self, adapter_name, init_weights):
         Reset the OFT parameters.
         """
         if init_weights is False:
-            nn.init.normal_(self.oft_r[adapter_name], mean=0.0, std=0.1)
-            nn.init.normal_(self.oft_s[adapter_name], mean=1.0, std=0.1)
+            nn.init.normal_(self.oft_R[adapter_name], mean=0.0, std=0.1)
```
BenjaminBossan: This fails for me because `self.oft_R[adapter_name]` is a `OFTRotationModule`, did you mean:
```suggestion
            nn.init.normal_(self.oft_R[adapter_name].weight, mean=0.0, std=0.1)
```

BenjaminBossan (changes requested):
Thanks for offering the perspective, I understand your argument. After some further exploration of the code, it looks like there is no easy way to make the transition backwards compatible. I think the least we should do, however, is to check if a user tries to load an old OFT checkpoint and raise an error with a helpful error message. For this, we could for instance add some code after this line:

https://github.com/huggingface/peft/blob/a27406c26da0a1a4ceba2a98ac85c94e30824694/src/peft/utils/save_and_load.py#L435

The check could be something like this:

```python
elif config.peft_type == PeftType.OFT:
    # new OFT adapter is backwards incompatible, see #2575
    if any(".oft_s." in key for key in peft_model_state_dict):
        raise ValueError("Trying to load old OFT checkpoint, which is no longer supported. Please install PEFT <= v0.15.2 to load it or train a new OFT adapter.")
```

Next week, I should hopefully find the opportunity to do some testing with the new OFT method to check if I can replicate the improvements that you mentioned.

Before we can merge, we also need to update some of the tests. For instance, this test no longer works with the new OFT parameters:

https://github.com/huggingface/peft/blob/a27406c26da0a1a4ceba2a98ac85c94e30824694/tests/test_common_gpu.py#L1882-L1898

Furthermore, since quantized layers are now supported, let's add GPU tests for them. As an example, here is a LoRA test for GPTQ. You can essentially copy this test and switch the method from LoRA to OFT.

src/peft/tuners/oft/layer.py
zqiu24: Yes, updated

src/peft/tuners/oft/layer.py
zqiu24: Yes, updated.

src/peft/tuners/oft/layer.py
zqiu24: OK, updated

src/peft/tuners/oft/layer.py
zqiu24: OK, updated.

src/peft/tuners/oft/model.py
zqiu24: OK, updated

src/peft/tuners/oft/layer.py
zqiu24: Updated

zqiu24:
@BenjaminBossan Thank you so much for the detailed reply and the suggestions. I just pushed a version with the required changes and added tests.

src/peft/tuners/oft/awq.py
```diff
@@ -0,0 +1,145 @@
+# Copyright 2024-present the HuggingFace Inc. team.
```
BenjaminBossan: Let's update all the years to 2025.

src/peft/tuners/oft/config.py
BenjaminBossan: WDYT of this suggestion?

src/peft/tuners/oft/model.py
```diff
@@ -126,22 +134,31 @@ def _create_and_replace(
...
+            "loaded_in_8bit": getattr(self.model, "is_loaded_in_8bit", False),
+            "loaded_in_4bit": getattr(self.model, "is_loaded_in_4bit", False),
         }
-        kwargs["bias"] = bias
+
+        quant_methods = ["gptq", "aqlm", "awq"]
+        for quant_method in quant_methods:
+            quantization_config = get_quantization_config(self.model, method=quant_method)
+            if quantization_config is not None:
+                kwargs[f"{quant_method}_quantization_config"] = quantization_config
```
BenjaminBossan, line 156: It's not clear to me why this is needed, could you please add a comment? Why is it only performed for some quantization methods and not others?

tests/test_custom_models.py
BenjaminBossan, line 1: There is now the option to use OFT with `use_cayley_neumann=True` and `use_cayley_neumann=False`. Let's ensure that we have at least one test case for each one of those.

BenjaminBossan (changes requested):
Thanks for the updates, there isn't much missing for this PR.

Apart from the smaller comments I left, ideally we should also have one test per quantization method which was added in this PR. For this, please check [`test_gpu_examples.py`](https://github.com/huggingface/peft/blob/main/tests/test_gpu_examples.py), where you can see the test classes for each quantization method. Could you please add OFT examples?

I would also like to do some of my own testing, but this will have to wait until next week. I hope this is fine.

src/peft/tuners/oft/awq.py
zqiu24: OK, updated.

src/peft/tuners/oft/model.py
zqiu24: For the above one, I change to default use the oft_block_size to be consistent with lora (currently, the r in OFT, if it gets bigger, the # params actually gets smaller), therefore I am trying to let the user to default set oft_block_size (bigger value corresponds to larger # trainable params).

For the below comment, I adapted it from the lora/model.py, is it not necessary?

BenjaminBossan: > For the above one, I change to default use the oft_block_size to be consistent with lora (currently, the r in OFT, if it gets bigger, the # params actually gets smaller), therefore I am trying to let the user to default set oft_block_size (bigger value corresponds to larger # trainable params).

I see, so this is because it's more intuitive for the user that increasing this hyper-parameter increases the total number of parameters. Then I think it's okay to change the default. Nonetheless, could you please extend the docs for `r` and `oft_block_size` to mention how they impact the model parameter count?

> For the below comment, I adapted it from the lora/model.py, is it not necessary?

Ah I see, it is indeed requires for GPTQ here:

https://github.com/huggingface/peft/pull/2575/files#diff-93456276d55a0a0bd19039b99cbebb83ec8ca68275732b81d865742cb7e263deR99

However, I don't see it being used for AQLM or AWQ. Probably, this was just blindly copy-pasted for LoRA and could be removed there too.

zqiu24: Sure, I will update the OFT doc.

tests/test_custom_models.py
zqiu24: @BenjaminBossan Where do you suggest to add this test? Best,

BenjaminBossan: It should be sufficient to extend the existing OFT test cases here:

https://github.com/huggingface/peft/blob/bd893a8a36a86cd5f09a2935cf06ba82501af28b/tests/test_custom_models.py#L300-L322

zqiu24:
@BenjaminBossan Hi Benjamin, just updated the test, would be great if you could provide some feedback this week? We would really like to publish the paper to Arxiv :)

BenjaminBossan:
> Hi Benjamin, just updated the test, would be great if you could provide some feedback this week? We would really like to publish the paper to Arxiv :)

@zqiu24 Okay, let's try our best to get this merged this week. I tried to run the new tests with:

`pytest tests/test_custom_models.py -k "oft and not boft" -v`

but I get an error:

> tests/test_custom_models.py - ValueError: You can only specify either r (2) or oft_block_size (32), but not both simultaneously, because r x oft_block_size == in_features.

Could you please check if the tests pass for you and if they don't, fix the remaining issues?

I'm also currently running some experiments and get an issue when the base model dtype != float32. I think I know the error, but I'll get back to you once I'm sure I pinned it down.

PS: I think I found the cause, see my comment below.

src/peft/tuners/oft/layer.py
```diff
@@ -431,42 +637,15 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
...
-            orig_weight = self.get_base_layer().weight.data
-            orig_weight = torch.transpose(orig_weight, 0, 1)
-            rotated_weight = torch.mm(oft_rotation, orig_weight.to(oft_rotation.dtype))
-            rotated_weight = torch.transpose(rotated_weight, 0, 1)
-            scaled_rotated_weight = rotated_weight * oft_scale
-
-            x = self._cast_input_dtype(x, scaled_rotated_weight.dtype)
-            bias = self._cast_input_dtype(self.get_base_layer().bias, scaled_rotated_weight.dtype)
-            result = F.linear(input=x, weight=scaled_rotated_weight, bias=bias)
+            result = self.base_layer(x, *args, **kwargs)
```
BenjaminBossan: So this line errors for me when the base layer is bfloat16. The reason is that x has been upcast to float32 due to `x = self._cast_input_dtype(x, oft_R.weight.dtype)`. I could fix the problem by changing this line to:

```suggestion
            result = self.base_layer(x.to(previous_dtype), *args, **kwargs)
```

zqiu24: Thanks, updated.

zqiu24:
@BenjaminBossan Thank you so much:) I will do my best to make the changes as fast as I get the review. Yes, the above issue is because we changed the default now.

for some reasons I have the following error when I perform test locally, it breaks:

INTERNALERROR>   File "/home/wliu/anaconda3/envs/peft_pr/lib/python3.12/site-packages/coverage/sqlitedb.py", line 138, in _execute
INTERNALERROR>     raise DataError(f"Couldn't use data file {self.filename!r}: {msg}") from exc
INTERNALERROR> coverage.exceptions.DataError: Couldn't use data file '/lustre/fast/fast/wliu/zqiu/peft_pr/peft/.coverage.i102.2091071.XYUWDkvx.c': disk I/O error

So I cannot see the error logs,

zqiu24:
platform linux -- Python 3.12.10, pytest-8.4.0, pluggy-1.6.0 -- /home/wliu/anaconda3/envs/peft_pr/bin/python3.12
cachedir: .pytest_cache
rootdir: /lustre/fast/fast/wliu/zqiu/peft_pr/peft
configfile: pyproject.toml
plugins: xdist-3.7.0, cov-6.2.1

This is my testing env

BenjaminBossan:
Some good news, I tested the new OFT implementation on our [MetaMathQA benchmark](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA) and there are significant memory improvements, going down from 28GB to 22GB. However, the test accuracy also decreased from 48.5% to 44.5% and train loss was worse, 0.596 vs 0.636. I assume that it has to do with the new default values. Would you recommend changing the hyper-parameters to get better results?

> for some reasons I have the following error when I perform test locally, it breaks:

Hmm, I haven't seen that error. Could you try re-installing pytest and pytest-cov? If that doesn't help, maybe just uninstall pytest-cov and locally remove [this config line](https://github.com/huggingface/peft/blob/bd893a8a36a86cd5f09a2935cf06ba82501af28b/pyproject.toml#L44). We're not interested in test coverage right now, so this would be fine.

zqiu24:
@BenjaminBossan Thanks for the suggestions. I run the local tests and passed all tests for test_custom_models, feel free to tell me whether there are any other test problems :)

Maybe we could first arrive a version where it is ready to be merged (no errors / all tests passed), then I will locally adjust the hyperparameters a little bit for the performance?

Best,

BenjaminBossan:
@zqiu24 Could you please run `make style`?

zqiu24:
@BenjaminBossan Hi, just pushed a version with make style :)

BenjaminBossan:
@zqiu24 The code quality checks still fail, could you try explicitly running

`doc-builder style src/peft tests docs/source --max_len 119 --check_only`

zqiu24:
@BenjaminBossan Hi, thanks, when I run the above command, it says that:

```
doc-builder style src/peft tests docs/source --max_len 119 --check_only
  File "/home/wliu/anaconda3/envs/peft_pr/bin/doc-builder", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/wliu/anaconda3/envs/peft_pr/lib/python3.12/site-packages/doc_builder/commands/doc_builder_cli.py", line 47, in main
    args.func(args)
  File "/home/wliu/anaconda3/envs/peft_pr/lib/python3.12/site-packages/doc_builder/commands/style.py", line 28, in style_command
    raise ValueError(f"{len(changed)} files should be restyled!")
ValueError: 2 files should be restyled!
```

which is the same as running

```make quality```

but the two files are not related to oft?

**Comment by zqiu24** [comment]
@BenjaminBossan OK, I ran ```make style``` and it changed two oft-related files, now it should be working.

**Comment by BenjaminBossan** [comment]
@zqiu24 Thanks for the updates. Now the tests run, but a [bunch of them are failing](https://github.com/huggingface/peft/actions/runs/15851912573/job/44688264464?pr=2575). `test_custom_models.py` looks good but the other tests fail. You should be able to reproduce the errors with:

`pytest tests/test_decoder_models.py tests/test_encoder_decoder_models.py tests/test_feature_extraction_models.py tests/test_stablediffusion.py -k "oft and not boft"`

**Comment by zqiu24** [comment]
@BenjaminBossan Thanks for the feedback, just resolved the errors in:

```
pytest tests/test_decoder_models.py tests/test_encoder_decoder_models.py tests/test_feature_extraction_models.py tests/test_stablediffusion.py -k "oft and not boft"
```

**Comment by zqiu24** [comment]
@BenjaminBossan Do I need to do something? it seems some versions the tests were passed and for some versions not

**Comment by BenjaminBossan** [comment]
MacOS errors are unrelated

**Comment by zqiu24** [comment]
> MacOS errors are unrelated

@BenjaminBossan So if I determined the best the of default hyperparameters, it is ready to be merged? Best,

**Comment by BenjaminBossan** [comment]
> So if I determined the best the of default hyperparameters, it is ready to be merged? Best,

I guess we can merge as is, but normally we would not want to change the default parameters later. Thus, ideally, we have them set before merging. If you have enough time to test better defaults, please go ahead.

**Comment by zqiu24** [comment]
Yes, thanks. I am just curious about the performance gap. I will test it locally and give me the go from my side. Best,

**Comment by zqiu24** [comment]
@BenjaminBossan would you mind to test the current version again and see if the performance is better?

> Some good news, I tested the new OFT implementation on our [MetaMathQA benchmark](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA) and there are significant memory improvements, going down from 28GB to 22GB. However, the test accuracy also decreased from 48.5% to 44.5% and train loss was worse, 0.596 vs 0.636. I assume that it has to do with the new default values. Would you recommend changing the hyper-parameters to get better results?
> 
> > for some reasons I have the following error when I perform test locally, it breaks:
> 
> Hmm, I haven't seen that error. Could you try re-installing pytest and pytest-cov? If that doesn't help, maybe just uninstall pytest-cov and locally remove [this config line](https://github.com/huggingface/peft/blob/bd893a8a36a86cd5f09a2935cf06ba82501af28b/pyproject.toml#L44). We're not interested in test coverage right now, so this would be fine.

File: src/peft/tuners/oft/layer.py [hunk 15]
```diff
@@ -322,7 +543,20 @@ def __init__(
 
         self._active_adapter = adapter_name
 
-        self.update_layer(adapter_name, r, oft_block_size, module_dropout, coft, eps, block_share, init_weights)
+        breakpoint()
```
BenjaminBossan: Please remove

zqiu24: Sorry, forget to delete this.

BenjaminBossan:
> would you mind to test the current version again and see if the performance is better?

I ran it again, with the updated code and otherwise same settings. The final results are very similar (showing most relevant metrics):

Before:

```json
    "test accuracy": 0.4450341167551175,
    "train loss": 0.6357718685865402,
    "cuda_memory_reserved_avg": 11804291891,
    "cuda_memory_max": 22189965312,
    "cuda_memory_reserved_99th": 17720976343,
    "train_time": 1967.2921348049422,
```

After:

```json
    "test accuracy": 0.4473085670962851,
    "train loss": 0.6358395928144455,
    "cuda_memory_reserved_avg": 11783439070,
    "cuda_memory_max": 22198353920,
    "cuda_memory_reserved_99th": 17720934400,
    "train_time": 2134.552857758019,
```

If you want, you should be quickly able to run the benchmark yourself. You need this file: `method_comparison/MetaMathQA/experiments/oft/llama-3.2-3B-oft_block_size-32/adapter_config.json`

```json
{
  "auto_mapping": null,
  "bias": "none",
  "block_share": false,
  "coft": false,
  "eps": 6e-05,
  "exclude_modules": null,
  "fan_in_fan_out": false,
  "inference_mode": false,
  "init_weights": true,
  "layers_pattern": null,
  "layers_to_transform": null,
  "module_dropout": 0.0,
  "modules_to_save": null,
  "num_cayley_neumann_terms": 5,
  "oft_block_size": 32,
  "peft_type": "OFT",
  "r": 0,
  "revision": null,
  "target_modules": null,
  "task_type": null,
  "use_cayley_neumann": false
}
```

and then run `python run.py -v method_comparison/MetaMathQA/experiments/oft/llama-3.2-3B-oft_block_size-32/`

(you might need some extra requirements: `pip install -r method_comparison/MetaMathQA/requirements.txt`)

At the end of the day, we can still merge and figure out if there are better settings later.

zqiu24:
@BenjaminBossan Thank you for the help and running of OFT tests, I really appreciate it:)
I tested locally, I think the performance gap may come from the oft_s, which currently does not exist anymore, I tried locally there is no way I can think of to use that, without significantly slower the training and increasing the memory usage. But for other tasks, OFT yields competitive performance. It can also be that for this task, we might need a bigger oft_block_size or another learning rate to get better performance. For now, I think maybe we just keep as it is for PR? Best,

BenjaminBossan:
Okay, then let's deal with potentially better experimental settings later. LMK if the PR is ready for a final review.

zqiu24:
@BenjaminBossan Please have a final review. Best,

docs/source/conceptual_guides/oft.md
```diff
@@ -74,6 +86,52 @@ specify either `boft_block_size` or `boft_block_num`, but not both simultaneousl
...
+## OFT Example Usage
+
+For using OFT for quantized finetuning with [TRL](https://github.com/huggingface/trl) for `SFT`, `PPO`, or `DPO` fine-tuning, follow the following outline:
+
+```py
+from transformers import AutoTokenizer, AutoModelForCausalLM, BitsAndBytesConfig
+from trl import SFTTrainer
+from peft import OFTConfig
+
+if use_quantization:
```
BenjaminBossan, line 98: This snippet uses a mix of 2 and 4 spaces for indentation. Let's use 4 spaces consistently.

src/peft/tuners/oft/bnb.py
```diff
@@ -0,0 +1,388 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan: Here and below, the date is sometimes wrong.

```suggestion
# Copyright 2025-present the HuggingFace Inc. team.
```

tests/test_gptqmodel.py
```diff
@@ -315,6 +405,68 @@ def test_causal_lm_training_multi_accelerator(self):
             # assert loss is not None
             assert trainer.state.log_history[-1]["train_loss"] is not None
 
+    @pytest.mark.single_gpu_tests
```
BenjaminBossan:
```suggestion
    @pytest.mark.multi_gpu_tests
```

tests/test_common_gpu.py
```diff
@@ -1884,18 +1884,18 @@ def test_oft_add_new_adapter_does_not_change_device(self, mlp):
         config = OFTConfig(target_modules=["lin0"])
         model = get_peft_model(mlp, config)
         model = model.to(self.device)
-        model.lin0.oft_r.cpu()
+        model.lin0.oft_R.weight.cpu()
```
BenjaminBossan:
```suggestion
        model.lin0.oft_R.cpu()
```

src/peft/tuners/oft/bnb.py
zqiu24: Updated.

tests/test_gptqmodel.py
zqiu24: Updated.

docs/source/conceptual_guides/oft.md
zqiu24: Updated.

tests/test_common_gpu.py
zqiu24: @BenjaminBossan Locally it seems a bit different, I don't have this oft_r any more, can you check in the newest commit?

BenjaminBossan: It's not about `oft_r`, it's about removing the `.weight`. The issue with this line is that `oft_R` now has a `ModuleDict`, so the correct call would be `model.lin0.oft_R.default.weight.cpu()`. But my suggestion would also work and is shorter.

zqiu24: @BenjaminBossan Thanks for the clarification. updated

zqiu24:
@BenjaminBossan Hi, is there any other changes required from my side? Best,

BenjaminBossan (approved):
Thanks for the improvement to, and extension of, OFT. The changes LGTM.

PR #2576, opened 2025-06-09
Optimize isinstance Check in LoraParallelLinear

JavaZeroo:
The self.is_parallel_a attribute is [set](https://github.com/huggingface/peft/pull/2576/commits/1c96361517881aae57291d890f0b1b226dca80fa#diff-5e62802c7cd4e536708b9ad1b5bfe866d307630079d95d234c2e5cd022aabd9eL66) during initialization based on whether base_layer is an instance of backend.RowParallelLinear. Reperforming the isinstance check is redundant and slightly less performant.


BenjaminBossan (approved):
Thanks. This is a microscopic optimization, but still valid, so it can be merged. Failing CI is unrelated.

PR #2577, opened 2025-06-09
[FEAT] Add C3A Support

Phoveran:
[C3A Paper](https://arxiv.org/abs/2407.19342) is here.

currently, I fail to run the test by the command `make test`, and the output is:

image

Thank you for your time and guidance!


BenjaminBossan:
Hi, thanks for the PR. I didn't have time to look at it yet or to read the paper, but regarding your question, I assume that you would like to add unit tests and that's where you encounter it. The reason is that you're missing the `pytest-cov` package in your environment.

Phoveran:
Hi Benjamin,

I've passed the test cases and done make style. Would be kind to review the code? Thank you!

src/peft/tuners/c3a/__init__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan: Let's update all dates to 2025

src/peft/tuners/c3a/config.py
```diff
@@ -0,0 +1,112 @@
...
+from peft.utils import PeftType
+
+
+@dataclass
+class C3AConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`C3AModel`].
+
+    Args:
+        block_size (`int`): block_size for C3A, must be divisible by both the input number and the output number.
```
BenjaminBossan: Let's mention the default value and how increasing/decreasing the block size influences the number of trainable parameters. Also, the wording is a bit imprecise IMO, how about:

> block_size for C3A, must be divisible by both the input size and the output size of the target layer.

src/peft/tuners/c3a/config.py
```diff
@@ -0,0 +1,112 @@
...
+            and saved in the final checkpoint.
+        layers_to_transform (`Union[list[int],int]`):
+            The layer indexes to transform, if this argument is specified, it will apply C3A on the layer indexes that
+            are specified in this list. If a single integer is passed, it will apply C3A on the layer at this index.
+        layers_pattern (`str`):
+            The layer pattern name, used only if `layers_to_transform` is different from `None` and if the layer
+            pattern is not in the common layers pattern.
+        block_size_pattern (`dict`):
+            The mapping from layer names or regexp expression to block_size which are different from the default
+            specified. For example, `{model.decoder.layers.0.encoder_attn.k_proj: 1280`}
```
BenjaminBossan:
```suggestion
            specified. For example, `{"model.decoder.layers.0.encoder_attn.k_proj": 1280`}
```

src/peft/tuners/c3a/layer.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def update_layer(self, adapter_name, block_size, init_weights):
+        if block_size <= 0:
+            raise ValueError(f"`block_size` should be a positive integer value but the value passed is {block_size}")
+        assert self.in_features % block_size == 0, (
+            f"The block size should be a factor of the input size. However, the input size is {self.in_features} and the block size is {block_size}"
+        )
+        assert self.out_features % block_size == 0, (
+            f"The block size should be a factor of the output size. However, the output size is {self.out_features} and the block size is {block_size}"
+        )
```
BenjaminBossan: Let's raise proper `ValueError`s here. Also, let's add tests for this to `test_initialization.py`.

src/peft/tuners/c3a/layer.py
```diff
@@ -0,0 +1,203 @@
...
+
+        self.block_size[adapter_name] = block_size
+
+        weight = getattr(self.get_base_layer(), "weight", None)
```
BenjaminBossan:
```suggestion
        weight = self.get_base_layer().weight
```

Since only linear layers are supported, the attribute should exist. If it doesn't, the code below would fail anyway.

src/peft/tuners/c3a/config.py
```diff
@@ -0,0 +1,112 @@
...
+        init_weights (`bool`):
+            The initialization of the C3A weights. Set this to False if the weights should be initialized to a commonly
+            used distribution. Set this to True if the weights should be initialized to zeros.
```
BenjaminBossan: Here and below, let's mention the possible values for `init_weights`. Also, the type is better expressed as `Literal[True, "gaussian", "xavier_uniform", "kaiming_uniform"]`. AFAICT, `False` is not a valid option, hence `bool` is not a correct type.

src/peft/tuners/c3a/config.py
```diff
@@ -0,0 +1,112 @@
...
+    block_size_pattern: Optional[dict] = field(
+        default_factory=dict,
+        metadata={
+            "help": (
+                "The mapping from layer names or regexp expression to block_size which are different from the default specified. "
+                "For example, `{model.decoder.layers.0.encoder_attn.k_proj: 1280`}"
+            )
+        },
+    )
+    init_weights: Optional[bool] = field(
```
BenjaminBossan:
```suggestion
    init_weights: Literal[True, "gaussian", "xavier_uniform", "kaiming_uniform"] = field(
```

src/peft/tuners/c3a/layer.py
```diff
@@ -0,0 +1,203 @@
...
+                a = 1.0 * math.sqrt(1.0 / float(fan_in))
+                nn.init.uniform_(self.c3a_kernel[adapter_name], -a, a)
+            else:
+                raise ValueError(f"Unknown init_weights: {init_weights}")
+
+
+#  ------------------------------------------------------------------------------------------
+#  Copyright (c) Microsoft Corporation. All rights reserved.
+#  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+#  ------------------------------------------------------------------------------------------
```
BenjaminBossan: Remove

src/peft/tuners/c3a/layer.py
```diff
@@ -0,0 +1,203 @@
...
+            return
+
+        for active_adapter in adapter_names:
+            if active_adapter in self.c3a_kernel.keys():
+                base_layer = self.get_base_layer()
+                if safe_merge:
+                    # Note that safe_merge will be slower than the normal merge
+                    # because of the copy operation.
+                    orig_weights = base_layer.weight.data.clone()
+                    # orig_weights += self.get_delta_weight(active_adapter)
```
BenjaminBossan: Remove?

src/peft/tuners/c3a/utils.py
```diff
@@ -0,0 +1,44 @@
+import torch
+from torch.autograd import Function
+from torch.fft import fft, ifft
+
+
+def circulant(tensor, dim=-1):
```
BenjaminBossan: Is this used anywhere?

BenjaminBossan (changes requested):
Thanks a lot for adding C³A to PEFT. The method looks very interesting and promising. I'll try to find some time next week to do a bit of my own testing.

The implementation itself is already nearly perfect, I just added a few smaller comments, please check.

On top of that, before we merge, let's also work:

1. Add documentation about C³A
2. Add one example
3. Extend the tests a bit: Please add a test case to `test_decoder_models.py`, `test_encoder_decoder_models.py`, `test_feature_extraction_models.py`, `test_seq_classifier.py`, and `test_stablediffusion.py`. This should be straightforward as it only requires adding a test config similar to what you did in `test_custom_models.py`. However, due to the shape constraint of C³A, some of these models may not work. In that case, let's just add a comment that says something like "C³A could not be tested because the input and output shapes are ...".

Out of curiosity, is there any relation to FourierFT?

Phoveran:
Hi Benjamin,

I have added more testing, a doc and an example according to your guidance. Besides, issues you listed should have been addressed. Let me know if you need further changes.

Actually, one of the original goal of this work is to make FourierFT mathematically elegant and efficient. :D

src/peft/tuners/c3a/config.py
```diff
@@ -0,0 +1,117 @@
...
+    block_size_pattern: Optional[dict] = field(
+        default_factory=dict,
+        metadata={
+            "help": (
+                "The mapping from layer names or regexp expression to block_size which are different from the default specified. "
+                "For example, `{model.decoder.layers.0.encoder_attn.k_proj: 1280`}"
+            )
+        },
+    )
+    init_weights: Optional[Literal[True, "gaussian", "kaiming_uniform", "xavier_uniform"]] = field(
```
BenjaminBossan: Since you made `False` a valid option, the correct type would now be:

`Union[bool, Literal[True, "gaussian", "kaiming_uniform", "xavier_uniform"]]`

or

`bool | Literal[True, "gaussian", "kaiming_uniform", "xavier_uniform"]]`

src/peft/tuners/c3a/layer.py
```diff
@@ -0,0 +1,198 @@
...
+
+    @torch.no_grad()
+    def reset_c3a_parameters(self, adapter_name, init_weights):
+        if init_weights is True:
+            return
+
+        if adapter_name in self.c3a_kernel.keys():
+            if init_weights == "gaussian":
+                nn.init.normal_(self.c3a_kernel[adapter_name])
+            elif init_weights in ["xavier_uniform", False]:  # Support test cases where False presents
```
BenjaminBossan, line 96: Could you explain why you chose `False == 'xavier_uniform'`? In general, we have the convention that `True` means the adapter is initialized as a no-op and `False` means it's not a no-op. Using Xavier is indeed consistent with that, but in LoRA we use Kaiming for `False`. So is there a specific reason to use Xavier here?

src/peft/tuners/c3a/layer.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+class C3ALinear(nn.Module, C3ALayer):
+    # Lora implemented in a dense layer
+    def __init__(
+        self,
+        base_layer,
+        adapter_name: str,
+        block_size: int,
+        init_weights: str,
```
BenjaminBossan: Could you please adjust the type annotation to be consistent with what's in the config?

tests/test_initialization.py
```diff
@@ -1530,6 +1531,40 @@ def test_vblora_with_incompatible_vector_length_with_out_features(self):
             get_peft_model(model, config)
 
 
+class TestC3AInitialization:
```
BenjaminBossan, line 1581: Nice

tests/test_decoder_models.py
```diff
@@ -201,12 +202,20 @@
...
         },
     ),
+    (
+        C3AConfig,
+        {
+            "task_type": "CAUSAL_LM",
+            "block_size": 1,
+            "target_modules": None,
+        },
+    ),
```
BenjaminBossan, line 213: I see, `block_size=1` makes it so that it should always work, no matter the shape of the weights. However, it's else somewhat of an edge case, right? Could you please add another config with a `block_size>1` here and in the other tests? Something like 2, 4, or 8 should most likely work. I think this would give us more robustness in the C³A tests.

src/peft/tuners/c3a/layer.py
```diff
@@ -0,0 +1,198 @@
...
+            result = self.base_layer(x, *args, **kwargs)
+        elif self.merged:
+            result = self.base_layer(x, *args, **kwargs)
+        else:
+            result = self.base_layer(x, *args, **kwargs)
+            for active_adapter in self.active_adapters:
+                if active_adapter not in self.c3a_kernel.keys():
+                    continue
+                c3a_kernel = self.c3a_kernel[active_adapter]
+                x = BlockCircularConvolution.apply(x.float(), c3a_kernel.float()) / x.size(-1)
```
BenjaminBossan: I'm wondering about the dypes here, is there a reason why `x` needs to be cast to float? Does the block circular convolution work with fp16 and bf16? If yes, would it make sense to keep the dtype?

If it only works with fp32, could we cast `c3a_kernel` to fp32 during initialization instead of having to do it in each `forward` and `get_delta_weight` call?

examples/sequence_classification/C3A.ipynb
```diff
@@ -0,0 +1,511 @@
...
+   "execution_count": null,
+   "id": "e3b13308",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "batch_size = 32\n",
+    "model_name_or_path = \"roberta-base\"\n",
+    "task = \"mrpc\"\n",
+    "peft_type = PeftType.C3A\n",
+    "device = \"cuda\" if torch.cuda.is_available() else \"cpu\"\n",
```
BenjaminBossan: Going forward, we attempt to make our code and examples more device agnostic. Could you please use `infer_device` which you can import from `peft.utils`?

tests/test_decoder_models.py
Phoveran: Ahhh Actually I found in your test case that one of the layer is of the shape 37, which is a prime number and makes all tests with `block_size>1` fail.

src/peft/tuners/c3a/layer.py
Phoveran: Xavier has a better performance in our experiments ahhh. It is 100% experimental.

src/peft/tuners/c3a/layer.py
Phoveran: I have added comments in the code for the supporting status of FFT. Thank you!

src/peft/tuners/c3a/layer.py
BenjaminBossan: I see, then it's okay like this, just add a comment to explain the non-obvious choices.

tests/test_decoder_models.py
BenjaminBossan: Too bad, then having `block_size=1` here is fine, as long as there are other tests with `block_size>1`. Let's just add a comment that other block sizes are not possible here because of the damn prime numbers.

Phoveran:
Hi Benjamin,

Thank you for these comments! I believe I have properly addressed them :D. Please feel free to leave messages for further concerns!

BenjaminBossan:
@Phoveran With Python 3.9, there is the following error:

```
src/peft/tuners/c3a/layer.py:115: in C3ALinear
    init_weights: bool | Literal["gaussian", "kaiming_uniform", "xavier_uniform"],
E   TypeError: unsupported operand type(s) for |: 'type' and '_LiteralGenericAlias'
```

I think a `from __future__ import annotations` import should fix it.

src/peft/tuners/c3a/config.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+@dataclass
+class C3AConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`C3AModel`].
+
+    Args:
+        block_size (`int`):
+            block size for C3A, must be divisible by both the input size and the output size of the target layer. If
```
BenjaminBossan, line 30: Could you please ensure that the character limit of 120 is enforced in this docstring? ruff does not automatically fix strings.

BenjaminBossan (commented):
Thanks for fixing the last issue.

IMO the PR is ready to be merged, but before doing that, I wanted to discuss one matter. In PEFT, we have now added a [benchmark for all PEFT method](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA). I added C³A using the default config settings and started a run but then canceled after 2500 steps. The issue was that the loss was stagnating at around 1.3 and the valid accuracy stayed at 0. Other PEFT methods have much better scores. Probably the default settings are not working well. Do you have suggestions for how to adjust the hyper-parameters? Perhaps, it also means we need to update the default values.

(Btw. we have similarly bad scores for [FourierFT](https://github.com/huggingface/peft/blob/5fe7f8f8abe914d313fc3751f2ea92de7718fbaf/method_comparison/MetaMathQA/results/fourierft--llama-3.2-3B-default.json), maybe you know some better settings for this model as well)

Phoveran:
Thanks for raising the benchmarking issue.

Indeed, we found that the usually used training config does not fit FourierFT nor C3A. The main difference should be the choice on LR. We refer to the Appendix part of these papers [FourierFT](https://arxiv.org/pdf/2405.03003) & [C3A](https://arxiv.org/pdf/2407.19342) for more LR details. In general, you might try 5e-1 for C3A and 3e-3 for FourierFT.

Thank you and please let me know if there's any further issue!

src/peft/tuners/c3a/config.py
Phoveran: I'm not quite familiar with this, but have done `docformatter config.py --wrap-descriptions 119 --wrap-summaries 112 --in-place` to format the docstrings. If it is not correct, could you please provide more guidance? Thanks!

BenjaminBossan: Hmm, I think the issue is that the long lines are not in the docstring, but long strings in the code itself, thus they are not touched by any formatter. Probably the only solution is to wrap them by hand. It's not a big deal, you can keep it as is, it just makes reading the code harder.

BenjaminBossan:
> The main difference should be the choice on LR. We refer to the Appendix part of these papers [FourierFT](https://arxiv.org/pdf/2405.03003) & [C3A](https://arxiv.org/pdf/2407.19342) for more LR details. In general, you might try 5e-1 for C3A and 3e-3 for FourierFT.

Thanks for the guidance. Could you please update the C³A docs to mention that a much higher learning rate than usual is needed? Otherwise, many users who try it might bounce because the model doesn't train well.

 I ran an experiment locally with lr 0.5 and got much better results. That said, they're still not on par with LoRA (rank 32):

| metric | LoRA rank 32 | C³A w/ lr 0.5 |
| -- | -- | -- |
| total time | 1847 sec | 2102 sec |
| trainable params | 9175040 | 1376256 |
| cuda max reserved memo | 22.3 GB | 22.0 GB |
| train loss | 0.607 | 0.659 |
| test accuracy | 0.478 | 0.412 |

It could be due to the number of parameters, but that's just my hunch.

Anyway, it is not critical for this PR, but I think going forward, it would be nice to provide good experimental settings and defaults so that users can get started as quickly as possible.

Phoveran:
Thanks for the feedback! I have manually wrapped strings and updated the doc. I will work on a proper experimental setting ASAP.

Phoveran:
Hi Benjamin,

I have done some experiments and found out one satisfying experimental setting. The setting and results are uploaded. (You might find the meta info part blank since I'm using a machine that is not allowed to connect to the public network and I'm using offline HF.) Besides, more instructions are added.

In my experiment, I found out that C3A excels in accuracy with only nearly half parameters compared with LoRA, and this results in a much smaller adapter file.

BenjaminBossan:
Thanks a lot @Phoveran for running those experiments, discovering a good setting, and documenting how users can choose better hyper-parameters. As for the result file, could you please remove it? We will run the experiment on our machines to ensure that all experiments run under comparable conditions.

BenjaminBossan:
Could you please merge with/rebase on the latest PEFT main, then the CI should pass and we can merge.

BenjaminBossan (approved):
Thanks for the great work, the PR LGTM. Hopefully it finds a lot of adoption.

PR #2579, opened 2025-06-10
FIX Account for attention mask being a dict, fix generate issues with gemma

BenjaminBossan:
See also #2580

Resolves CI errors such as this one:

https://github.com/huggingface/peft/actions/runs/15481482956/job/43588020111#step:5:53182

This PR resolves 2 issues:

## 1. attention mask being a dict

After the transformers change in https://github.com/huggingface/transformers/pull/37866, it can happen that:

> Models using different types of attention in different layers (i.e. gemma3) will now have a dict returned by prepare_inputd_for_generation (one dict entry per attention type)

As PEFT operates on the attention mask for prompt learning methods, we need to adjust the code for the possibility of `attention_mask` being a dict. Right now, I simply extract the single value if the dict is just one element. For other sizes, I just raise an error, as I don't know how to deal with that. For our tests, this is enough but we might need to find a better solution in the future.

## 2. torch.compile errors during generation

#2458 fixed an issue with 4d attention masks and added gemma3 to the test suite, which uses 4d attention masks. However, the solution was insufficient, as it involves replacing the 4d attention mask with a 2d mask and handing it off to the model to create the correct 4d attention mask. The problem is that mask creation triggers an error with `torch.compile` and thus needs to be performed outside of the compile context, i.e. during  `prepare_inputs_for_generation`. This PR now uses the same logic as transformers to do exactly that.

There are still issues with prefix tuning and incorrect shapes, which may be solvable, but require further work. Similarly, there is an issue with VBLoRA because this line is not `torch.compile` friendly:

https://github.com/huggingface/peft/blob/e67052b18cb3c58370e0339ecf76a5faf447f2ff/src/peft/tuners/vblora/layer.py#L191

The corresponding tests are skipped for now.

Finally, for these fixes to work, two more changes are needed on the transformers side:

1. Await a new transformers release (>4.52) so that we can use [`create_masks_for_generate`](https://github.com/huggingface/transformers/blob/380e6ea4065afc4f639d7b1bed990e18020b7270/src/transformers/masking_utils.py#L921).
2. ~For prompt learning, we [remove the `cache_position` argument](https://github.com/huggingface/peft/blob/e67052b18cb3c58370e0339ecf76a5faf447f2ff/src/peft/peft_model.py#L1997), I'm not quite sure if there is not a better solution. Anyway, because of this it needs to be recomputed but models like [gemma recompute in a way that is not torch.compile-friendly](https://github.com/huggingface/transformers/blob/8ff22e9d3b0d14a4ae18890fbabe5ffc2d9a037b/src/transformers/models/gemma3/modeling_gemma3.py#L544-L548). They should use a [compile friendly method](https://github.com/huggingface/transformers/blob/8ff22e9d3b0d14a4ae18890fbabe5ffc2d9a037b/src/transformers/generation/utils.py#L1796) instead. When I locally patch transformers to do so, the tests pass.~ `cache_position` is no longer being removed from the `model_kwargs`, thus the aforementioned problem does not occur.

For these reasons, this PR stays in draft status for now and #2580 is used to make the CI green for the time being.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -41,7 +41,7 @@ from transformers.utils import PushToHubMixin
 from peft.tuners.tuners_utils import BaseTuner, BaseTunerLayer
 from peft.utils.constants import DUMMY_MODEL_CONFIG
 from peft.utils.integrations import init_empty_weights
-from peft.utils.other import set_additional_trainable_modules
+from peft.utils.other import create_attention_mask, set_additional_trainable_modules
 
 from . import __version__
 from .config import PeftConfig
@@ -739,7 +739,7 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
                     dtype=past_key_values[0].dtype,
                     device=past_key_values[0].device,
                 )
-                cache_position = torch.arange(peft_config.num_virtual_tokens)
+                cache_position = torch.arange(peft_config.num_virtual_tokens, device=past_key_values[0].device)
                 for layer_idx in range(peft_config.num_layers):
                     key_states, value_states = past_key_values[0][layer_idx], past_key_values[1][layer_idx]
                     new_cache.update(
@@ -1961,7 +1961,22 @@ class PeftModelForCausalLM(PeftModel):
                     # to   [batch_size, total_sequence_length]
                     bs = attention_mask.shape[0]
                     total_seq_len = prefix_attention_mask.shape[1] + attention_mask.shape[2]
-                    model_kwargs["attention_mask"] = torch.ones((bs, total_seq_len), dtype=attention_mask.dtype)
+                    attention_mask_2d = torch.ones((bs, total_seq_len), dtype=attention_mask.dtype)
+                    if model_kwargs["cache_position"][0] == 0:
+                        # heuristic to determine if we're in prefill stage
+                        cache_position_ = torch.arange(total_seq_len, device=model_kwargs["input_ids"].device)
+                    else:
+                        cache_position_ = model_kwargs["cache_position"]
+                    attention_mask_new = create_attention_mask(
+                        self.get_base_model(),
+                        model_input=None,
+                        attention_mask=attention_mask_2d,
+                        past_key_values=model_kwargs.get("past_key_values"),
+                        cache_position=cache_position_,
+                        batch_size=bs,
+                        sequence_length=total_seq_len,
+                    )
+                    model_kwargs["attention_mask"] = attention_mask_new
                 else:
                     # 2d attention mask
                     model_kwargs["attention_mask"] = torch.cat((prefix_attention_mask, attention_mask), dim=1)
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -2018,15 +2018,19 @@ class PeftModelForCausalLM(PeftModel):
                 model_kwargs["input_ids"] = None
 
         # if cache_position exists and if we're in the prefill stage
-        if (model_kwargs.get("cache_position") is not None) and (model_kwargs["cache_position"][0] == 0):
-            if peft_config.peft_type == PeftType.PREFIX_TUNING:
-                # for prefix tuning, the past_key_values have been prefilled
-                model_kwargs["cache_position"] += peft_config.num_virtual_tokens
-            else:
-                # for other prompt learning methods, the inputs_embeds have been extended
-                device = model_kwargs["cache_position"].device
-                new_seq_length = model_kwargs["cache_position"].shape[0] + peft_config.num_virtual_tokens
-                model_kwargs["cache_position"] = torch.arange(new_seq_length).to(device)
+        if (
+            (model_kwargs.get("cache_position") is not None)
+            and (model_kwargs["cache_position"][0] == 0)
+            and (peft_config.peft_type == PeftType.PREFIX_TUNING)
+        ):
+            # for prefix tuning, the past_key_values have been prefilled
+            model_kwargs["cache_position"] += peft_config.num_virtual_tokens
+        else:
+            # For transformers>=4.38.0 - for some architectures such as Llama, `cache_position` is passed in the forward
+            # pass to keep track of the position ids of the cache. We have to pop that from `model_kwargs` as
+            # `cache_position` is properly created by the model, using the passed `inputs_embeds`:
+            # https://github.com/huggingface/transformers/blob/593230f0a1150ea9c0477b9d859f25daf73c8c33/src/transformers/models/llama/modeling_llama.py#L956
+            _ = model_kwargs.pop("cache_position", None)
 
         return model_kwargs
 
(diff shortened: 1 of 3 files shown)


src/peft/peft_model.py
```diff
@@ -1951,7 +1961,27 @@ def prepare_inputs_for_generation(self, *args, task_ids: Optional[torch.Tensor]
                     # to   [batch_size, total_sequence_length]
                     bs = attention_mask.shape[0]
                     total_seq_len = prefix_attention_mask.shape[1] + attention_mask.shape[2]
-                    model_kwargs["attention_mask"] = torch.ones((bs, total_seq_len), dtype=attention_mask.dtype)
+                    attention_mask_2d = torch.ones((bs, total_seq_len), dtype=attention_mask.dtype)
+
+                    # heuristic to determine if we're in prefill stage
```
githubnemo: in the comment:  please explain what the prefill stage is for and what it relates to (kv cache)

src/peft/peft_model.py
```diff
@@ -1951,7 +1961,27 @@ def prepare_inputs_for_generation(self, *args, task_ids: Optional[torch.Tensor]
...
+                    is_prefill = model_kwargs["cache_position"][0] == 0
+                    if is_prefill and (peft_config.peft_type != PeftType.PREFIX_TUNING):
+                        # if in prefill stage, for prompt learning methods that are not prefix tuning, new tokens
+                        # (embeddings) are inserted
```
githubnemo: in the comment: please explain why prefix tuning is exempt here

currently it reads as if prefix tuning doesn't insert inputs at all which is confusing because it does, but, IIUC, into the kv cache of all layers.

src/peft/peft_model.py
```diff
@@ -1987,11 +2017,20 @@ def prepare_inputs_for_generation(self, *args, task_ids: Optional[torch.Tensor]
...
 
-        # For transformers>=4.38.0 - for some architectures such as Llama, `cache_position` is
-        # passed in the forward pass to keep track of the position ids of the cache. We have to
-        # pop that from `model_kwargs` as `cache_position` is properly created by the model, using the passed
-        # `inputs_embeds`: https://github.com/huggingface/transformers/blob/593230f0a1150ea9c0477b9d859f25daf73c8c33/src/transformers/models/llama/modeling_llama.py#L956
-        _ = model_kwargs.pop("cache_position", None)
+        # if cache_position exists and if we're in the prefill stage
+        if (
+            (model_kwargs.get("cache_position") is not None)
+            and (model_kwargs["cache_position"][0] == 0)
```
githubnemo: it might make sense to move the `cache position not None and cache position == 0` part higher and make that into `is_prefill` with a proper comment to repeat repetition and add a bit of semantic context.

src/peft/utils/other.py
```diff
@@ -1272,3 +1272,49 @@ def set_additional_trainable_modules(model, peft_config, model_config, adapter_n
                 token_indices=token_adapter.token_indices[adapter_name],
                 tied_adapter=model.get_input_embeddings().token_adapter,
             )
+
+
+def create_attention_mask(
+    model, *, model_input, attention_mask, past_key_values, cache_position, batch_size, sequence_length
+):
+    # adapted from:
+    # https://github.com/huggingface/transformers/blob/cb4c56ce0dfa1350267ed28e57760986a58a9ba4/src/transformers/generation/utils.py#L644-L680
```
githubnemo, line 1281: in the comment: Please add a bit of context what the purpose of this function is and in what context it is supposed to be used / why it is necessary to do what it does.

this is a very specific function which does things that are a bit intransparent and subtle so I think aiding the reader with context info is a good thing here.

BenjaminBossan:
Note: I checked locally and some Gemma3 tests are failing when run on GPU due to compile errors:

`pytest tests/test_decoder_models.py::TestDecoderModels::test_generate[CPTConfig-config_kwargs3-hf-internal-testing/tiny-random-Gemma3ForCausalLM] tests/test_decoder_models.py::TestDecoderModels::test_generate[VBLoRAConfig-config_kwargs13-hf-internal-testing/tiny-random-Gemma3ForCausalLM] tests/test_decoder_models.py::TestDecoderModels::test_generate[PromptEncoderConfig-config_kwargs11-hf-internal-testing/tiny-random-Gemma3ForCausalLM] tests/test_decoder_models.py::TestDecoderModels::test_generate[PromptTuningConfig-config_kwargs12-hf-internal-testing/tiny-random-Gemma3ForCausalLM] tests/test_decoder_models.py::TestDecoderModels::test_generate_pos_args[CPTConfig-config_kwargs3-hf-internal-testing/tiny-random-Gemma3ForCausalLM] tests/test_decoder_models.py::TestDecoderModels::test_generate_pos_args[PromptTuningConfig-config_kwargs12-hf-internal-testing/tiny-random-Gemma3ForCausalLM] tests/test_decoder_models.py::TestDecoderModels::test_generate_pos_args[VBLoRAConfig-config_kwargs13-hf-internal-testing/tiny-random-Gemma3ForCausalLM] tests/test_decoder_models.py::TestDecoderModels::test_generate_pos_args[PromptEncoderConfig-config_kwargs11-hf-internal-testing/tiny-random-Gemma3ForCausalLM]`

The error is:

```
E               torch._dynamo.exc.Unsupported: Data dependent operator
E                 Explanation: Operator `aten._local_scalar_dense.default` has a non-Tensor output whose value is dependent on the data of Tensor inputs.
E                 Hint: Enable tracing of data-dependent output operators with `torch._dynamo.config.capture_scalar_outputs = True`
E               
E                 Developer debug context: aten._local_scalar_dense.default
E               
E               
E               from user code:
E                  File "/home/name/work/forks/transformers/src/transformers/utils/generic.py", line 943, in wrapper
E                   output = func(self, *args, **kwargs)
E                 File "/home/name/work/forks/transformers/src/transformers/models/gemma3/modeling_gemma3.py", line 681, in forward
E                   outputs: BaseModelOutputWithPast = self.model(
E                 File "/home/name/work/forks/transformers/src/transformers/utils/generic.py", line 943, in wrapper
E                   output = func(self, *args, **kwargs)
E                 File "/home/name/work/forks/transformers/src/transformers/models/gemma3/modeling_gemma3.py", line 525, in forward
E                   cache_position = torch.arange(
```

The reason is that Gemma uses [some](https://github.com/huggingface/transformers/blob/371c4711136386075bfb272692860c1d4ee9c1d2/src/transformers/models/gemma3/modeling_gemma3.py#L525-L529) [code](https://github.com/huggingface/transformers/blob/371c4711136386075bfb272692860c1d4ee9c1d2/src/transformers/models/gemma3/modeling_gemma3.py#L889-L891) that is not `torch.compile` friendly to generate `cache_position`. If that code was switched to a [compile friendly version](https://github.com/huggingface/transformers/blob/8ff22e9d3b0d14a4ae18890fbabe5ffc2d9a037b/src/transformers/generation/utils.py#L1796), the tests should pass. Ping @Cyrilvallez 

I could confirm that these failures are not caused by this PR, but rather that these failures were masked by the error that is fixed in this PR.

PR #2593, opened 2025-06-18
Add Makefile + results for MetaMathQA task

githubnemo:
These are the first results for the MetaMathQA task and also the first test of the Makefile used to run these tests.

The Makefile offers the functionality to run individual experiments by specifying the result you want to have, e.g.
`make results/adalora--llama-3.2-3B-rank32[...].json`. Alternatively you can simply run `make` for `make all` which runs all experiments that don't have a result yet or which have outdated configs (comparing result timestamp and config timestamp).

The results are from the main branch. No errors happened during the run. There were errors with a compute instance that used a A10G 24GB because of OOM. L40S with 48GB RAM was fine.


method_comparison/app.py
```diff
@@ -349,7 +349,7 @@ def reset_filter(task_name, model_id, metric_x, metric_y):
 
 
 # TODO only 1 task, using temporary results for now
-path = os.path.join(os.path.dirname(__file__), "MetaMathQA", "temporary_results")
+path = os.path.join(os.path.dirname(__file__), "MetaMathQA", "results")
```
BenjaminBossan, line 351: Let's remove the comment above.

BenjaminBossan (changes requested):
Thanks for adding the Makefile and the results. These results look consistent with the tentative results we had previously.

Could you please add a small section to the README that explains the make commands? Something akin to what you added to the PR description. Let's also remove the `results/Makefile` placeholder, as we now have one Makefile per experiment.

Regarding the Makefile, I have no experience with that, so can't really review it. As long as it functionally works, I'm fine, it's not core PEFT code anyway.

That said, maybe I'm missing something, but when I run `make all`, it seems like all experiments are being run. For instance, the LoKr experiment is run even though the timestamp for LoKr's `experiments/lokr/llama-3.2-3B-rank32/adapter_config.json` is from June 17 11:34, and for `lokr--llama-3.2-3B-rank32--2025-06-17T17-17-37+00-00.json` it is June 18 15:34.

githubnemo:
> That said, maybe I'm missing something, but when I run `make all`, it seems like all experiments are being run. For instance, the LoKr experiment is run even though the timestamp for LoKr's `experiments/lokr/llama-3.2-3B-rank32/adapter_config.json` is from June 17 11:34, and for `lokr--llama-3.2-3B-rank32--2025-06-17T17-17-37+00-00.json` it is June 18 15:34.

I think that is to be expected. The date in the file name is not relevant to make, only the date on the FS. It is only useful if you have a machine that is persisted and doesn't override the results every time you want to execute new/missing experiments.

BenjaminBossan:
> I think that is to be expected. The date in the file name is not relevant to make, only the date on the FS.

The reported dates are from the FS, so the experiment file for LoKr is older than the result file. Therefore, I would have thought that it does not need to run.

2catycm:
I think the parameter efficiency is also an important param to be shown in the app.py. It should be compared in the pareto figure. 

And the performance of Full Finetuning is also very important for comparison

BenjaminBossan:
> I think the parameter efficiency is also an important param to be shown in the app.py. It should be compared in the pareto figure.

It is somewhat of an intentional choice not to report parameter efficiency. The reason is that it would draw a lot of attention but is actually not that informative in practice. This is because the number of trainable parameters does not have a linear relationship with memory requirement, runtime, checkpoint size etc. But those are the outcomes people really care about in practice. I guess we could still log the value but showing it in the plot is not as helpful as it may sound and could even be counter productive.

> And the performance of Full Finetuning is also very important for comparison

We haven't done that yet because of the memory requirement but we should definitely give it a shot.

githubnemo:
@BenjaminBossan there was a small bug in the Makefile that assumed that the `/` separators were replaced with `-` instead of `--`. This is fixed and alongside I removed the timestamps from the results. We can look back at past results via `git restore -s <rev> results` anyway.

Re trainable parameters: I'll create a separate PR for this.

method_comparison/MetaMathQA/utils.py
```diff
@@ -566,7 +566,7 @@ def log_to_console(log_data: dict[str, Any], print_fn: Callable[..., None]) -> N
 def log_to_file(
     *, log_data: dict, save_dir: str, experiment_name: str, timestamp: str, print_fn: Callable[..., None]
 ) -> None:
-    file_name = f"{experiment_name.replace(os.path.sep, '--')}--{timestamp.replace(':', '-')}.json"
```
BenjaminBossan, line 569: With this change, let's remove the `timestamp` argument from `log_to_file` completely.

BenjaminBossan:
@2catycm For full fine-tuning, see #2597.

2catycm:
> @2catycm For full fine-tuning, see #2597.

Thanks a lot! It is very useful.

2catycm:
I am also wondering if it is possible for the script to use transformers's Trainer or lightning's  Trainer. 

One important use case for me is the gradient accumulation, because sometimes my available gpu memory is too small (well, even though these methods are parameter efficient) the batch size written in the default config, while it is important to keep the results equivalent. It can be easily implemented by Trainers by just changing arguments. But it is a little tricky for me to add this function to the script alone. I am also interested in using deepspeed to improve the training speed. 

In addition, with a trainer, the saving logic may be easier to manage.

githubnemo:
> I am also wondering if it is possible for the script to use transformers's Trainer or lightning's Trainer.

No, we decidedly chose to keep as much of the training process under our control to keep the complexity down and have as much flexibility as possible. This also comes with the drawback that certain things that are subtle to implement (like gradient accumulation) have to be repeated but it was a cost we decided to take for the aforementioned benefits.

> One important use case for me is the gradient accumulation, because sometimes my available gpu memory is too small (well, even though these methods are parameter efficient) the batch size written in the default config, while it is important to keep the results equivalent.

If the memory consumption is too large for your local compute, you can reduce the batch size via `training_params.json` in the experiment folder and remove that file when submitting a PR for your experiment. Currently experiments are run on an NVIDIA L40S with 48GB VRAM. You don't have to supply the results of a run, only the experiment configs since we aim to run all the experiments on the same hardware.

> I am also interested in using deepspeed to improve the training speed.

Good to hear but this benchmark is not the place to use DeepSpeed. Feel free to provide a DeepSpeed example with your implementation, though!

PR #2594, opened 2025-06-18
Input sanitizer for benchmark result renderer

githubnemo:
Since `DataFrame.query` is potentially vulnerable we limit the possible filter input to a fixed grammar that is roughly like this:

```
expr = left op right
left = ( expr ) | literal
right = ( expr ) | literal
op = in | >= | < | <= | == | and | or
```

this will give us boolean operations and basic comparisons. Note that `literal` can be arbitrary python literals (strings, tuples, ...).

### DIFF
--- a/method_comparison/sanitizer.py
+++ b/method_comparison/sanitizer.py
@@ -1,0 +1,91 @@
+import ast
+
+import pandas as pd
+
+
+def _evaluate_node(df, node):
+    """
+    Recursively evaluates an AST node to generate a pandas boolean mask.
+    """
+    # Base Case: A simple comparison like 'price > 100'
+    if isinstance(node, ast.Compare):
+        if not isinstance(node.left, ast.Name):
+            raise ValueError("Left side of comparison must be a column name.")
+        col = node.left.id
+        if col not in df.columns:
+            raise ValueError(f"Column '{col}' not found in DataFrame.")
+
+        if len(node.ops) > 1:
+            raise ValueError("Chained comparisons like '10 < price < 100' are not supported.")
+
+        op_node = node.ops[0]
+        val_node = node.comparators[0]
+        try:
+            value = ast.literal_eval(val_node)
+        except ValueError:
+            raise ValueError("Right side of comparison must be a literal (number, string, list).")
+
+        operator_map = {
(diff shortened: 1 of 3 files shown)


method_comparison/test_sanitizer.py
```diff
@@ -0,0 +1,39 @@
...
+def df_products():
+    data = {
+        'product_id': [101, 102, 103, 104, 105, 106],
+        'category': ['Electronics', 'Books', 'Electronics', 'Home Goods', 'Books', 'Electronics'],
+        'price': [799.99, 19.99, 49.50, 120.00, 24.99, 150.00],
+        'stock': [15, 300, 50, 25, 150, 5]
+    }
+    return pd.DataFrame(data)
+
+def test_lower(df_products):
```
BenjaminBossan: Can all the succeeding tests be subsumed into one with extended parametrization?

method_comparison/test_sanitizer.py
```diff
@@ -0,0 +1,39 @@
...
+    mask1 = parse_and_filter(df_products, "price < 50")
+    assert sorted(df_products[mask1].product_id) == [102, 103, 105]
+
+def test_exploit_fails(df_products):
```
BenjaminBossan, line 18: Let's add a comment that without parsing, this succeeds. Although technically, it would not work in the app as pandas is not imported, but for instance `total_time < 2500 and @os.system("/bin/echo password")` works.

method_comparison/sanitizer.py
```diff
@@ -0,0 +1,91 @@
...
+        expression_node = tree.body
+    except (SyntaxError, ValueError) as e:
+        raise ValueError(f"Invalid filter syntax: {e}")
+
+    # The recursive evaluation starts here
+    mask = _evaluate_node(df, expression_node)
+    return mask
+
+
+
```
BenjaminBossan: del del del

BenjaminBossan (changes requested):
Thanks for discovering and fixing this vulnerability. Overall, LGTM, just some small comments. Moreover, some valid expressions are rejected, e.g. `(peft_type == "LORA") & (train_time < 2500)` or `not (peft_type == "LORA")`. Not a huge deal, but if it's not too difficult, it would be nice to add them.

PR #2596, opened 2025-06-19
enable ut cases on XPU, all enabled cases pass 

yao-matrix:


tests/test_gpu_examples.py
```diff
@@ -3418,14 +3419,15 @@ def test_causal_lm_training_awq(self):
     # TODO remove marker if/once issue is resolved, most likely requiring a fix in AutoAWQ:
     # https://github.com/casper-hansen/AutoAWQ/issues/754
     @pytest.mark.xfail(
+        condition=is_torch_version("==", "2.7"),
```
yao-matrix, line 3422: PT 2.8 works in my env, so I added a condition here to only XFAIL 2.7

BenjaminBossan: Interesting. Let's change the condition to `<=` then, unless you know it works with 2.6 and below.

Overall, [AutoAWQ is now archived](https://github.com/casper-hansen/AutoAWQ), so I think we don't need to put too much effort in keeping it running in the future.

BenjaminBossan (changes requested):
Thanks for the PR. Just a minimal change requirement from my side.

tests/test_gpu_examples.py
yao-matrix: From https://github.com/casper-hansen/AutoAWQ/issues/754, it seems per your test "When switching to an env with PyTorch 2.6, the test passes". So I just XFAIL 2.7.

Sure, do you have a preference to use which library when we want to use awq algorithm? `optimum-quanto`, `torchao` or maybe `llm-compressor`?

BenjaminBossan: Oh, good catch, then we can leave it as is.

> Sure, do you have a preference to use which library when we want to use awq algorithm? `optimum-quanto`, `torchao` or maybe `llm-compressor`?

For now, we can leave it as is.

BenjaminBossan (approved):
Thanks for the PR and for your further explanations. I think we can merge it as is and deal with AWQ at a later point.

PR #2598, opened 2025-06-19
Track number of (trainable) parameters for MetaMathQA

githubnemo:
This PR adds tracking for the number of (trainable) parameters for each experiment

Tracking the number of parameters, trainable and total, will make the results
much more transparent regarding model capacity. If a method was accidentally
trained with a lot more or less trainable parameters it would make for unfair
results. Having these numbers will also make benchmarking parameter efficiency
easier.

This is a draft based on https://github.com/huggingface/peft/pull/2593.


BenjaminBossan (approved):
LGTM

PR #2599, opened 2025-06-19
Auto-tagging of PEFT models

githubnemo:
Features like inference need correctly set tags on the repo / the model card in order to be available. Also the Hub uses tags to index the models and make them searchable.

With this change PEFT tags models automatically as lora if they happen to be trained with LoRA, the base model and a custom `peft:method:<the method>` tag.

@Vaibhavs10 as discussed. This would ensure that new models are always tagged correctly and work out-of-the-box with, say, inference providers. We also took the liberty to introduce a `peft:method:<the_method>` tag (e.g., `peft:method:bone` or `peft:method:lora`). This is not final, feel free to discuss but we thought it may make sense to have this so that users can explore adapters a bit more deeply.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -1408,6 +1408,25 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
     def active_peft_config(self):
         return self.peft_config[self.active_adapter]
 
+    def _get_peft_specific_model_tags(self):
+        """Derive tags for the model card from the adapter's config. For example, setting the
+        base model is important for enabling support for inference providers but it also makes models more searchable
+        on the hub.
+        """
+        peft_method = self.active_peft_config.peft_type.value
+
+        tags = [
+            f"peft:method:{peft_method.lower()}",
+        ]
+
+        if peft_method == "LORA":
+            tags.append("lora")
+
+        if hasattr(self.base_model, "name_or_path"):
+            tags.append(f"base_model:adapter:{self.base_model.name_or_path}")
+
+        return tags
+
     def create_or_update_model_card(self, output_dir: str):
         """
         Updates or create model card to include information about peft:
@@ -1422,6 +1441,12 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
         card = ModelCard.load(filename) if os.path.exists(filename) else ModelCard.from_template(ModelCardData())
 
         card.data["library_name"] = "peft"
+
+        tags = self._get_peft_specific_model_tags()
+        if not card.data["tags"]:
+            card.data["tags"] = tags
+        else:
+            card.data["tags"] = list(set(card.data["tags"]).add(tags))
 
         model_config = BaseTuner.get_model_config(self)
         model_config = None if model_config == DUMMY_MODEL_CONFIG else model_config
--- a/tests/test_hub_features.py
+++ b/tests/test_hub_features.py
@@ -15,9 +15,10 @@ import copy
 
 import pytest
 import torch
+from huggingface_hub import ModelCard
 from transformers import AutoModelForCausalLM
 
-from peft import AutoPeftModelForCausalLM, LoraConfig, PeftConfig, PeftModel, get_peft_model
+from peft import AutoPeftModelForCausalLM, BoneConfig, LoraConfig, PeftConfig, PeftModel, get_peft_model
 
 
 PEFT_MODELS_TO_TEST = [("peft-internal-testing/test-lora-subfolder", "test")]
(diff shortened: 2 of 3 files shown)


src/peft/peft_model.py
```diff
@@ -1408,6 +1408,25 @@ def base_model_torch_dtype(self):
     def active_peft_config(self):
         return self.peft_config[self.active_adapter]
 
+    def _get_peft_specific_model_tags(self):
+        """Derive tags for the model card from the adapter's config. For example, setting the
+        base model is important for enabling support for inference providers but it also makes models more searchable
+        on the hub.
```
BenjaminBossan:
```suggestion
        base model is important for enabling support for HF inference providers but it also makes models more
        searchable on the HF hub.
```

src/peft/peft_model.py
```diff
@@ -1408,6 +1408,25 @@ def base_model_torch_dtype(self):
...
+
+        tags = [
+            f"peft:method:{peft_method.lower()}",
+        ]
+
+        if peft_method == "LORA":
+            tags.append("lora")
+
+        if hasattr(self.base_model, "name_or_path"):
+            tags.append(f"base_model:adapter:{self.base_model.name_or_path}")
```
BenjaminBossan, line 1427: Is there a specific reason for this adapter format?

src/peft/peft_model.py
```diff
@@ -1423,6 +1442,14 @@ def create_or_update_model_card(self, output_dir: str):
 
         card.data["library_name"] = "peft"
 
+        tags = set()
+        if hasattr(self.base_model, "model_tags"):
```
BenjaminBossan:
```suggestion
        base_model = self.get_base_model()
        if hasattr(base_model, "model_tags"):
```

src/peft/peft_model.py
```diff
@@ -1423,6 +1442,14 @@ def create_or_update_model_card(self, output_dir: str):
...
+            tags = tags.union(self.base_model.model_tags or [])
+
+        tags = tags.union(self._get_peft_specific_model_tags())
+        if tags:
+            card.data["tags"] = list(tags)
```
BenjaminBossan: WDYT? This makes the order deterministic, which can help, for instance, with keeping diff noise low.

```suggestion
            card.data["tags"] = sorted(tags)
```

tests/test_hub_features.py
```diff
@@ -112,3 +113,65 @@ def test_load_different_peft_and_base_model_revision(self, tmp_path):
...
+        ],
+    )
+    def test_model_card_has_expected_tags(self, model_id, peft_config, tags, excluded_tags, pre_tags, tmp_path):
+        """Make sure that PEFT sets the tags in the model card automatically and correctly.
+        This is important so that a) the models are searchable on the Hub and also 2) some features depend on it to
+        decide how to deal with them (e.g., inference).
+
+        Makes sure that the base model tags are still present (if there are any).
+        """
+        base_model = AutoModelForCausalLM.from_pretrained(model_id)
```
BenjaminBossan: Should we add `with hub_online_once(model_id)` here?

tests/test_hub_features.py
```diff
@@ -112,3 +113,65 @@ def test_load_different_peft_and_base_model_revision(self, tmp_path):
...
+        peft_model = get_peft_model(base_model, peft_config)
+        save_path = tmp_path / "adapter"
+
+        peft_model.save_pretrained(save_path)
+
+        model_card = ModelCard.load(save_path / "README.md")
+        assert set(tags).issubset(set(model_card.data.tags))
+
+        if excluded_tags:
+            assert not set(excluded_tags).issubset(set(model_card.data.tags))
```
BenjaminBossan: I wonder if this should be:

```suggestion
            assert set(excluded_tags).isdisjoint(set(model_card.data.tags))
```

The issue with `issubset` is that it requires each item to be in the 2nd set. Thus if `excluded_tags = {"a", "b"}` and `model_card.data.tags = {"b", "c"}`, the assertion would not fail.

src/peft/peft_model.py
githubnemo: I'm just implementing this as instructed by @Vaibhavs10 :) Pinging for answers

githubnemo:
I've addressed the comments. From the commit:

> We remove the `peft:method:*` tag because this change needs more discussion and is partially unrelated to this change. It is replaced by the necessary `transformers` tag if the model is based on transformers.
>
> We're also trying to resolve the pipeline tag automatically if it isn't set.  While there is the `transformers.pipelines.base.SUPPORTED_PEFT_TASKS` mapping it is not sufficient to resolve the pipeline tag automatically since it is not a 1:1 mapping. Only the causal LM case is a unique mapping.


There's also a test model with task type `CAUSAL_LM` uploaded [here](https://huggingface.co/hubnemo/tiny-Qwen2ForCausalLM-2.5-lora).

src/peft/peft_model.py
```diff
@@ -1423,6 +1443,20 @@ def create_or_update_model_card(self, output_dir: str):
 
         card.data["library_name"] = "peft"
 
+        tags = set()
+        base_model = self.get_base_model()
+        if hasattr(base_model, "model_tags"):
+            tags = tags.union(self.base_model.model_tags or [])
```
BenjaminBossan:
```suggestion
            tags = tags.union(base_model.model_tags or [])
```

BenjaminBossan (approved):
PR LGTM, just a small comment.

src/peft/peft_model.py
```diff
@@ -1408,6 +1408,26 @@ def base_model_torch_dtype(self):
...
+        searchable on the HF hub.
+        """
+        peft_method = self.active_peft_config.peft_type.value
+
+        tags = []
+
+        if hasattr(self.base_model, "model") and isinstance(self.base_model.model, transformers.PreTrainedModel):
+            tags.append("transformers")
+
+        if peft_method == "LORA":
```
Vaibhavs10, line 1423: Should we do it for all the other methods too i.e. qlora etc

githubnemo: As discussed in slack, we're going to skip extending this feature for now until we know what tags we need.

PR #2602, opened 2025-06-20
Results with number of parameters + full fine tuning

githubnemo:
This change updates all results with their respective number of parameters (trained + absolute) and adds the newly introduced full-finetuning.

In addition to these results there was also an issue with the Makefile as it didn't consider the possibility of having experiments that don't have an adapter config (e.g., full fine-tuning).


BenjaminBossan (approved):
Thanks for the updated values and for enabling full fine-tuning runs. I checked a couple of diffs for the run and found that the metrics are very close, though unfortunately not 100% identical. However, the train times were significantly lower. This is almost certainly not related to changes in PEFT, so it has to do with the EC2 instance. Hopefully, in the future, we will see more consistent times.

As an example, for [method_comparison/MetaMathQA/results/lora--llama-3.2-3B-rank64.json](https://github.com/huggingface/peft/pull/2602/files#diff-91c14b1882501bd0d7678a40828481cb4ac96d417468f72de8d416700327773b), we have:

| metric | before | after |
| -- | -- | -- |
| test accuracy | 0.488248673237301 | 0.4890068233510235 |
| train loss | 0.5928513034582138 | 0.5928755496740341 |
| train_time | 2204.731052237 | 1853.4967184819961 |
| total time | 2386.7308763650003 | 2017.2277705579982 |

PR #2604, opened 2025-06-23
Add MiSS as a replacement for Bone.

JL-er:
Is it okay to directly add MiSS and inform users that Bone will be removed in a future version, recommending them to use MiSS instead? Currently, Bone doesn't seem to have many users.


Joluck:
> Is it okay to directly add MiSS and inform users that Bone will be removed in a future version, recommending them to use MiSS instead? Currently, Bone doesn't seem to have many users.

Alternatively, would it be acceptable to remove Bone entirely at this stage, given that it likely has few existing users?

docs/source/conceptual_guides/adapter.md
```diff
@@ -122,12 +122,23 @@ HRA constructs a chain of `r` trainable Householder reflections (HRs). Because t
...
+
+<small><a href="https://huggingface.co/papers/2409.15371">MiSS: Balancing LoRA Performance and Efficiency with Simple Shard Sharing</a></small>
+
+Intuitively, the shape of a single trainable matrix in MiSS is consistent with `lora_B`, so the `r` parameter in MiSS is less than the `r` in LoRA by (`in_feature * r`).
+
+Note: Bat's r (b) is special and requires that weight W satisfies the conditions `in_features % r == 0` and `out_features % r == 0`. Additionally, when `in_features == out_features` and MiSS-r equals LoRA-r, MiSS's number of trainable parameters is only half that of LoRA.
+
+Although the nonlinear updates of Bat bring some performance improvements, they also increase computational overhead. Its main purpose is to provide researchers with a direction for improvement. Therefore, we recommend fine-tuning the comprehensive MiSS model instead.
```
BenjaminBossan, line 137: Duplicate paragraph with l. 133.

docs/source/package_reference/miss.md
```diff
@@ -0,0 +1,32 @@
+<!--Copyright 2024 The HuggingFace Team. All rights reserved.
```
BenjaminBossan:
```suggestion
<!--Copyright 2025 The HuggingFace Team. All rights reserved.
```

docs/source/package_reference/miss.md
```diff
@@ -0,0 +1,32 @@
...
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# MiSS
+
+MiSS: Balancing LoRA Performance and Efficiency with Simple Shard Sharing([MiSS](https://huggingface.co/papers/2409.15371)) is a novel PEFT method that features a low-rank structure but introduces a new update mechanism distinct from LoRA, achieving an excellent balance between performance and efficiency.
```
BenjaminBossan: Instead of having a separate file for MiSS, I think it would be better to update `bone.md`.

examples/miss_finetuning/miss_finetuning.py
```diff
@@ -0,0 +1,105 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan:
```suggestion
# Copyright 2025-present the HuggingFace Inc. team.
```

src/peft/tuners/miss/config.py
```diff
@@ -0,0 +1,123 @@
...
+        metadata={
+            "help": "List of modules apart from MiSS layers to be set as trainable and saved in the final checkpoint. "
+            "For example, in Sequence Classification or Token Classification tasks, "
+            "the final layer `classifier/score` are randomly initialized and as such need to be trainable and saved."
+        },
+    )
+
+    def __post_init__(self):
+        super().__post_init__()
+        self.peft_type = PeftType.BONE
```
BenjaminBossan: Is this supposed to still be `BONE`?

BenjaminBossan (commented):
Thanks for the PR. I haven't done a full review yet, as I'm not quite clear about the difference between Bone and MiSS. Are the two methods identical, just the name differs, or are there differences in the implementation?

IIUC, the two are the same (which is what it looks like, except that MiSS adds `miss_dropout`). In that case, we should find a way to transition without breaking backwards compatibility.

One way to achieve that is to keep the code exactly as it is on the main branch and just update the documentation to mention that "Bone" is the old name and "MiSS" the new name, but the users should still use `BoneConfig` etc. The new `dropout` parameter can still be added.

If it is really important to switch the name in code too, we need to add extra measures to ensure backwards compatibility. Sure, there might only be a few Bone users, but we can't be sure of that and even if, breaking their code would not be nice. Instead, we should proceed as follows:

1. In `BoneConfig.__post_init__`, add a `FutureWarning`. Let's give at least 3 versions notice, i.e. let's say that Bone will be removed in v0.19.0.
2. In `save_and_load.py`, we need to detect Bone checkpoints and remap the keys from Bone to MiSS.
3. In `BoneConfig.from_pretrained`, we need to return a `MissConfig`.
4. We need a test that ensures that old Bone checkpoints can successfully be loaded, produce the same output, and load a `MissModel` instance with `MissLayer`s.
5. We need `# TODO` comments on all the Bone code/files that can be removed once we reach v0.19.0.

Probably there are more required steps that I'm missing, but this would be the very least. IMHO, this extra work compared to just keeping the old code and only updating the docs is quite high compared to the advantage. LMK what you think.

Joluck:
> Thanks for the PR. I haven't done a full review yet, as I'm not quite clear about the difference between Bone and MiSS. Are the two methods identical, just the name differs, or are there differences in the implementation?
> 
> IIUC, the two are the same (which is what it looks like, except that MiSS adds `miss_dropout`). In that case, we should find a way to transition without breaking backwards compatibility.
> 
> One way to achieve that is to keep the code exactly as it is on the main branch and just update the documentation to mention that "Bone" is the old name and "MiSS" the new name, but the users should still use `BoneConfig` etc. The new `dropout` parameter can still be added.
> 
> If it is really important to switch the name in code too, we need to add extra measures to ensure backwards compatibility. Sure, there might only be a few Bone users, but we can't be sure of that and even if, breaking their code would not be nice. Instead, we should proceed as follows:
> 
> 1. In `BoneConfig.__post_init__`, add a `FutureWarning`. Let's give at least 3 versions notice, i.e. let's say that Bone will be removed in v0.19.0.
> 2. In `save_and_load.py`, we need to detect Bone checkpoints and remap the keys from Bone to MiSS.
> 3. In `BoneConfig.from_pretrained`, we need to return a `MissConfig`.
> 4. We need a test that ensures that old Bone checkpoints can successfully be loaded, produce the same output, and load a `MissModel` instance with `MissLayer`s.
> 5. We need `# TODO` comments on all the Bone code/files that can be removed once we reach v0.19.0.
> 
> Probably there are more required steps that I'm missing, but this would be the very least. IMHO, this extra work compared to just keeping the old code and only updating the docs is quite high compared to the advantage. LMK what you think.

I'm sorry — this was my mistake. However, I believe renaming it is necessary to encourage more users to adopt and improve it in the future.

Joluck:
I believe we should accelerate this process as much as possible — v0.19.0 is a long release cycle, and if more users adopt it in the meantime, it could become problematic. I'll complete the remaining tests as soon as possible.

Joluck:
MiSS will also include a mini implementation, allowing users to control the sizes of the trainable matrices r1 and r2 independently.

BenjaminBossan:
> I believe we should accelerate this process as much as possible — v0.19.0 is a long release cycle

It's not that long. We will probably do a release soon, i.e. v0.16. Then we have new release approximately every 2-3 months, so 0.19.0 could be end of this year or early next year. I think we can also remove it in v0.18.0 but definitely not earlier than that.

Joluck:
> > I believe we should accelerate this process as much as possible — v0.19.0 is a long release cycle
> 
> 
> 
> It's not that long. We will probably do a release soon, i.e. v0.16. Then we have new release approximately every 2-3 months, so 0.19.0 could be end of this year or early next year. I think we can also remove it in v0.18.0 but definitely not earlier than that.

Alright, let’s complete the miss first and guide the user on how to use it. Thanks again!

BenjaminBossan:
> Alright, let’s complete the miss first and guide the user on how to use it. Thanks again!

Great, let's do it. To be on the same boat, do you want to first implement MiSS (as more a less a copy of Bone) and then in a separate PR deprecate Bone and add the code for backwards compatibility?

Joluck:
> > Alright, let’s complete the miss first and guide the user on how to use it. Thanks again!
> 
> Great, let's do it. To be on the same boat, do you want to first implement MiSS (as more a less a copy of Bone) and then in a separate PR deprecate Bone and add the code for backwards compatibility?

Bone offers the same functionality as MiSS, but all future updates will be centered around MiSS. For users currently using Bone, we strongly recommend switching to the new name.

Joluck:
> @BenjaminBossan

I need to add some clarification。

Joluck:
@BenjaminBossan 
I have completed the testing. Could you please check if there are any issues?

src/peft/tuners/miss/config.py
```diff
@@ -0,0 +1,139 @@
...
+
+from dataclasses import dataclass, field
+from typing import Literal, Optional, Union
+
+from peft.config import PeftConfig
+from peft.utils import PeftType
+
+
+@dataclass
+class MiSSConfig(PeftConfig):
```
BenjaminBossan: IMO, `MissConfig` etc. is preferable to `MiSSConfig` etc. First of all, it's easier to type, second, it's more consistent with the rest of the code base (apart from a few exceptions), e.g. we have `LoraConfig`, not `LoRAConfig`.

src/peft/tuners/miss/__init__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2024-present the HuggingFace Inc. team.
```
BenjaminBossan:
```suggestion
# Copyright 2025-present the HuggingFace Inc. team.
```

Please ensure the right date is used everywhere

src/peft/tuners/miss/config.py
```diff
@@ -0,0 +1,139 @@
...
+
+    Args:
+        r (`int`):
+            The rank of MiSS across different layers. It is best to set 'r' to an even number; otherwise, the default
+            initialization method will not work. The rank of MiSS corresponds to a low-rank decomposition along the
+            in_features dimension.
+        miss_dropout (`float`):
+            The dropout probability for Lora layers.
+        mini_r (`int`):
+            When you set `init_weights=mini`, you need to set `mini_r`. Please make sure that `out_features` is
```
BenjaminBossan: This doesn't really explain what the parameter does.

src/peft/tuners/miss/layer.py
```diff
@@ -0,0 +1,388 @@
...
+    # All names of layers that may contain (trainable) adapter weights
+    adapter_layer_names = ("miss_block",)
+    # All names of other parameters that may contain adapter-related parameters
+    other_param_names = ("miss_r", "miss_dropout", "miss_mini_r")
+
+    def __init__(self, base_layer: nn.Module, **kwargs) -> None:
+        self.base_layer = base_layer
+        self.miss_r = {}
+        self.miss_dropout = nn.ModuleDict({})
+        self.miss_mini_r = {}
```
BenjaminBossan, line 35: AFAICT, `self.miss_mini_r` is not being used. Instead, you use `self.mini_r`. This means that it's not possible to use multiple MiSS adapters with different values for `mini_r`. Please always use `self.miss_mini_r[adapter_name]`.

src/peft/tuners/miss/layer.py
```diff
@@ -0,0 +1,388 @@
...
+            raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
+
+        self.miss_r[adapter_name] = r
+        self.miss_mini_r[adapter_name] = mini_r
+        if miss_dropout > 0.0:
+            miss_dropout_layer = nn.Dropout(p=miss_dropout)
+        else:
+            miss_dropout_layer = nn.Identity()
+
+        self.miss_dropout.update(nn.ModuleDict({adapter_name: miss_dropout_layer}))
```
BenjaminBossan:
```suggestion
        self.miss_dropout[adapter_name] = miss_dropout_layer
```

Simpler :)

src/peft/tuners/miss/layer.py
```diff
@@ -0,0 +1,388 @@
...
+
+        # Initialize weights
+        if init_weights == "bat":
+            if self.in_features % r != 0 or self.out_features % r != 0:
+                raise ValueError("The weight matrix must be fully divisible into [r, r] blocks.")
+            self.reset_bat_parameters(adapter_name, r)
+        elif init_weights == "mini":
+            if self.out_features % mini_r != 0:
+                raise ValueError(
+                    "mini_r is divided along the out_features dimension. For optimal performance and implementation simplicity, it is recommended that out_features be divisible by mini_r."
```
BenjaminBossan: Let's add the value of `out_features` to the error message, as it may not be quite easy to figure out for the user. Also, please break the long line at 120 chars.

Let's also add a test to `test_initialization.py` for `mini_r`.

src/peft/tuners/miss/layer.py
```diff
@@ -0,0 +1,388 @@
...
+        miss_dropout: float = 0.0,
+        init_weights: Union[bool, str] = True,
+        **kwargs,
+    ) -> None:
+        super().__init__()
+        MiSSLayer.__init__(self, base_layer, **kwargs)
+        self._active_adapter = adapter_name
+        self.update_layer(adapter_name, r, mini_r, miss_dropout, init_weights, **kwargs)
+        self.miss_fn = init_weights
+        self.mini_r = mini_r
```
BenjaminBossan: Let's not store this attribute, that way we can be sure that `self.miss_mini_r` is always being used.

tests/test_custom_models.py
```diff
@@ -797,6 +814,20 @@
...
+        MiSSConfig,
+        {"target_modules": ["lin0"], "init_weights": False, "r": 2},
+        {"target_modules": ["lin0"], "init_weights": False, "r": 2},
+    ),
+    (
+        "MiSS Different",
+        "miss",
+        MiSSConfig,
+        {"target_modules": ["lin0"], "init_weights": False, "r": 2},
+        {"target_modules": ["lin1"], "init_weights": False, "r": 2},
```
BenjaminBossan, line 885: Let's add an example with `mini_r`

docs/source/conceptual_guides/adapter.md
```diff
@@ -122,12 +122,23 @@ HRA constructs a chain of `r` trainable Householder reflections (HRs). Because t
...
 
 ## Bone
-[DiSHA](https://huggingface.co/papers/2409.15371) A novel PEFT technique distinct from LoRA, called Dimension-Sharding Adaptation (DiSHA). By dividing the original weights into multiple subspaces that share a single matrix for weight updates, DiSHA simplifies the process by requiring the trainable matrix to be initialized to zero, eliminating the need for complex initialization as in some LoRA variants. Bone and Bat are derivative structures of DiSHA. Bone significantly improves computational efficiency while saving memory, whereas Bat addresses the limitation of Bone's linear update by employing a non-linear update to break through the upper bound.
+[DiSHA](https://huggingface.co/papers/2409.15371) New version of paper(MiSS: Balancing LoRA Performance and Efficiency with Simple Shard Sharing)
```
BenjaminBossan: As is, DiSHA is not explained any further, which is confusing to the reader. Please explain the relationship between DiSHA, Bone and MiSS. Also, please add a note that users should switch to MiSS from Bone.

BenjaminBossan (changes requested):
Thanks for your wok on the PR. I did a mostly light weight review, as most of the code is identical to the Bone code, and focused on the differences. Please check my comments.

Also, I think we should put in a concrete plan to help transition users from Bone to MiSS (even if it's done in a separate PR). What would be your suggestion?

1. Deprecate Bone.
2. Automatically convert Bone checkpoints into Miss when loading.
3. Provide a conversion script for Bone checkpoints and `adapter_config.json`.

src/peft/tuners/miss/layer.py
Joluck: image
Is this feasible?

Joluck:
> Thanks for your wok on the PR. I did a mostly light weight review, as most of the code is identical to the Bone code, and focused on the differences. Please check my comments.
> 
> Also, I think we should put in a concrete plan to help transition users from Bone to MiSS (even if it's done in a separate PR). What would be your suggestion?
> 
> 1. Deprecate Bone.
> 2. Automatically convert Bone checkpoints into Miss when loading.
> 3. Provide a conversion script for Bone checkpoints and `adapter_config.json`.

Thank you very much for your detailed answer. Do you have an example? Is it feasible to directly replace the name "Bone" with "MiSS" in the weight files and in adapter_config.json?

BenjaminBossan:
> Thank you very much for your detailed answer. Do you have an example? Is it feasible to directly replace the name "Bone" with "MiSS" in the weight files and in adapter_config.json?

We didn't have this exact case in the past, so I might not have considered everything. But off the top of my head, these steps would be necessary.

1. Remapping the `state_dict`

An example for this would be here:

https://github.com/huggingface/peft/blob/35000fda88fdf29db785783f8a40431cee2cab10/src/peft/utils/save_and_load.py#L427-L436

A similar approach could be taken for Bone->MiSS

2. Changing the config type

I think this could be dealt with here:

https://github.com/huggingface/peft/blob/35000fda88fdf29db785783f8a40431cee2cab10/src/peft/peft_model.py#L439-L448

3. Change the model type from `BoneModel` to `MissModel`

For this, we would need to change the `PEFT_TYPE_TO_TUNER_MAPPING`. It is usually automatically populated here: 

https://github.com/huggingface/peft/blob/35000fda88fdf29db785783f8a40431cee2cab10/src/peft/utils/peft_types.py#L167

So we would need some manual intervention to override Bone with MiSS.

4. Adding tests

All of this would need to be unit tested to ensure that nothing breaks during conversion. For this, I created a Bone checkpoint based on `trl-internal-testing/tiny-random-LlamaForCausalLM` in our PEFT testing repo on HF, which can be used as a basis for testing.

https://huggingface.co/peft-internal-testing/bone-tiny-llama

**Alternatively** we could think about not auto-converting Bone, only deprecating it. To facilitate the switch for users, we could provide a conversion script. This would be less work overall but also less user friendly. I would be fine with any of the two approaches.

Joluck:
> **Alternatively** we could think about not auto-converting Bone, only deprecating it. To facilitate the switch for users, we could provide a conversion script. This would be less work overall but also less user friendly. I would be fine with any of the two approaches.

I think we should just deprecate it directly, as that would indeed reduce a lot of work. Providing a conversion script would be more convenient. Do we need to first test whether the MiSS implementation is correct and feasible? I tested it locally and didn't encounter any issues.

src/peft/tuners/miss/layer.py
BenjaminBossan: I don't see why not.

BenjaminBossan:
> Do we need to first test whether the MiSS implementation is correct and feasible? I tested it locally and didn't encounter any issues.

We can work on that first and do the conversion script last. But I think it's not a bad idea to tackle this soon because it might surface potential issues which might be easier to fix the earlier we know about them.

Joluck:
> > Do we need to first test whether the MiSS implementation is correct and feasible? I tested it locally and didn't encounter any issues.
> 
> We can work on that first and do the conversion script last. But I think it's not a bad idea to tackle this soon because it might surface potential issues which might be easier to fix the earlier we know about them.

Would it be feasible if we only convert the checkpoint used in training? Could you specify exactly which parts I need to convert?

BenjaminBossan:
> Would it be feasible if we only convert the checkpoint used in training?

I'm not sure what you mean by that.

> Could you specify exactly which parts I need to convert?

My idea is the following: Let's say a user has trained a Bone model and saved it locally or on HF Hub. They call the conversion script, something like `python scripts/convert-bone-to-miss.py <path/to/checkpoint>`. The script then loads the `state_dict` and does the remapping of the keys. Furthermore, the script loads the `adapter_config.json` and rewrites it to use MiSS. Then, the results are written to a new directory. When the user loads from this new checkpoint, they will automatically get a `MissModel`. Does that make sense?

Joluck:
> My idea is the following: Let's say a user has trained a Bone model and saved it locally or on HF Hub. They call the conversion script, something like `python scripts/convert-bone-to-miss.py <path/to/checkpoint>`. The script then loads the `state_dict` and does the remapping of the keys. Furthermore, the script loads the `adapter_config.json` and rewrites it to use MiSS. Then, the results are written to a new directory. When the user loads from this new checkpoint, they will automatically get a `MissModel`. Does that make sense?

I conducted tests at [https://huggingface.co/peft-internal-testing/bone-tiny-llama](https://huggingface.co/peft-internal-testing/bone-tiny-llama).

BenjaminBossan:
LMK when this is ready for review.

Joluck:
> LMK when this is ready for review.

All set! ：）

scripts/convert-bone-to-miss.py
```diff
@@ -0,0 +1,41 @@
+import json
```
BenjaminBossan, line 19: Let's add the copyright notice and turn this into a proper CLI. That means, let's add argparse for the `input_dir` and `output_dir` and a small help. Also, please manually run ruff on this file once you're finished.

scripts/convert-bone-to-miss.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+import os
+from safetensors import safe_open
+from safetensors.torch import save_file
+
+input_path = os.path.join(input_dir, 'adapter_model.safetensors')
+output_path = os.path.join(output_dir, 'adapter_model.safetensors')
+
+os.makedirs(output_dir, exist_ok=True)
```
BenjaminBossan: Duplicated with line 8.

scripts/convert-bone-to-miss.py
```diff
@@ -0,0 +1,41 @@
+import json
+import os
+
+input_dir = "xxx/bone"
+output_dir = "xxx/miss"
+
+
+output_path = os.path.join(output_dir, "adapter_config.json")
```
BenjaminBossan: Let's use `peft.utils.CONFIG_NAME`

scripts/convert-bone-to-miss.py
```diff
@@ -0,0 +1,41 @@
...
+with open(output_path, 'w', encoding='utf-8') as f:
+    json.dump(config, f, indent=2, ensure_ascii=False)
+
+
+import os
+from safetensors import safe_open
+from safetensors.torch import save_file
+
+input_path = os.path.join(input_dir, 'adapter_model.safetensors')
+output_path = os.path.join(output_dir, 'adapter_model.safetensors')
```
BenjaminBossan: Let's use `peft.utils.SAFETENSORS_WEIGHTS_NAME`.

scripts/convert-bone-to-miss.py
```diff
@@ -0,0 +1,41 @@
...
+
+os.makedirs(output_dir, exist_ok=True)
+new_data = {}
+with safe_open(input_path, framework='pt') as f:
+    for old_key in f.keys():
+        tensor = f.get_tensor(old_key)
+        print(f"{old_key}: {tensor.shape}")
```
BenjaminBossan: Remove

scripts/convert-bone-to-miss.py
```diff
@@ -0,0 +1,41 @@
...
+        
+        new_key = old_key.replace("bone", "miss")
```
BenjaminBossan: I wonder if we should be more precise here, e.g. by doing this:

```suggestion
        new_key = old_key.replace(".bone_", ".miss_")
```

Just as an example, if there is a module called `backbone`, we would currently replace it with `backmiss`.

scripts/convert-bone-to-miss.py
```diff
@@ -0,0 +1,41 @@
...
+        new_data[new_key] = tensor
+
+save_file(new_data, output_path)
+
+
+print(f"\nfile save: {output_path}")
```
BenjaminBossan:
```suggestion
print(f"Converted checkpoint saved at {output_path}")
```

docs/source/conceptual_guides/adapter.md
```diff
@@ -122,12 +122,15 @@ HRA constructs a chain of `r` trainable Householder reflections (HRs). Because t
...
```
BenjaminBossan: I think we should mention the conversion script here, along with a small example of how to invoke it.

scripts/convert-bone-to-miss.py
BenjaminBossan, line 1: Did you try out if it works correctly? E.g. use a trained Bone adapter, convert it to MiSS, load the MiSS adapter, then check if it still produces the same output.

src/peft/tuners/bone/config.py
```diff
@@ -121,3 +122,7 @@ def __post_init__(self):
         # if target_modules is a regex expression, then layers_pattern should be None
         if isinstance(self.target_modules, str) and self.layers_pattern is not None:
             raise ValueError("`layers_pattern` cannot be used when `target_modules` is a str.")
+
+        warnings.warn(
+            "Bone will be removed in v0.18.0, you need to use MiSS"
+        )
```
BenjaminBossan:
```suggestion
        # TODO: remove Bone from PEFT in v0.18.0
        warnings.warn(
            "Bone will be removed in v0.18.0 of PEFT, use `MissConfig` instead.", FutureWarning
        )
```

I think here too, we should mention the conversion script.

src/peft/tuners/miss/config.py
```diff
@@ -0,0 +1,140 @@
+# Copyright 2024-present the HuggingFace Inc. team.
```
BenjaminBossan:
```suggestion
# Copyright 2025-present the HuggingFace Inc. team.
```

Let's ensure the correct year in each copyright notice.

src/peft/tuners/miss/layer.py
```diff
@@ -0,0 +1,392 @@
...
+        weight_miss = self.miss_block[adapter]
+
+        if cast_to_fp32:
+            weight_miss = weight_miss.float()
+
+        in_features = orig_weight.size(-1)
+        out_features = orig_weight.size(0)
+        r = weight_miss.size(0)
+        if self.miss_fn == "mini":
+            weight_miss = weight_miss.repeat(1, out_features // self.mini_r)
```
BenjaminBossan: This should also use `self.miss_mini_r[adapter]`

src/peft/tuners/miss/layer.py
```diff
@@ -0,0 +1,392 @@
...
+                bias = self._cast_input_dtype(self.base_layer.bias, orig_weight.dtype)
+                result = F.linear(input=x, weight=orig_weight, bias=bias)
+            else:
+                result = self.base_layer(x, *args, **kwargs)
+                for active_adapter in self.active_adapters:
+                    if active_adapter not in self.miss_block.keys():
+                        continue
+                    miss = self.miss_block[active_adapter]
+                    if self.miss_fn == "mini":
+                        # miss = miss.repeat(1, self.base_layer.out_features // self.mini_r)
```
BenjaminBossan: Remove

BenjaminBossan (changes requested):
Thanks for the updates. Overall not much is missing, but I think we can do a better job with the conversion script and advertising it to users. Please check my comments.

scripts/convert-bone-to-miss.py
Joluck: Yes, I have aligned their outputs.

Joluck:
@BenjaminBossan Thank you so much for your detailed review! Really appreciate you! ❤️

docs/source/conceptual_guides/adapter.md
```diff
@@ -122,12 +122,16 @@ HRA constructs a chain of `r` trainable Householder reflections (HRs). Because t
...
+"If you already have a Bone checkpoint, you can use `/scripts/convert-bone-to-miss.py` to convert it into a MiSS checkpoint and proceed with training using MiSS."
```
BenjaminBossan:
```suggestion
If you already have a Bone checkpoint, you can use `/scripts/convert-bone-to-miss.py` to convert it into a MiSS checkpoint and proceed with training using MiSS.
```

src/peft/tuners/bone/config.py
```diff
@@ -121,3 +122,9 @@ def __post_init__(self):
         # if target_modules is a regex expression, then layers_pattern should be None
         if isinstance(self.target_modules, str) and self.layers_pattern is not None:
             raise ValueError("`layers_pattern` cannot be used when `target_modules` is a str.")
+
+        warnings.warn(
+            "Bone will be removed in v0.18.0 of PEFT, use `MissConfig` instead." \
+            "If you already have a Bone checkpoint, you can use `/scripts/convert-bone-to-miss.py` to convert it into " \
```
BenjaminBossan:
```suggestion
            "Bone will be removed in v0.18.0 of PEFT, use `MissConfig` instead. "
            "If you already have a Bone checkpoint, you can use `/scripts/convert-bone-to-miss.py` to convert it into "
```

Not necessary to use `\`

scripts/convert-bone-to-miss.py
```diff
@@ -0,0 +1,74 @@
...
+from safetensors import safe_open
+from safetensors.torch import save_file
+
+from peft.utils import CONFIG_NAME, SAFETENSORS_WEIGHTS_NAME
+
+
+def convert_bone_to_miss(bone_dir: Path, miss_dir: Path) -> None:
+    """Convert Bone checkpoint files to MiSS format."""
+    bone_config_path = os.path.join(bone_dir, CONFIG_NAME)
+    miss_config_path = os.path.join(miss_dir, CONFIG_NAME)
```
BenjaminBossan: Since the paths are `Path` instances, we can simplify the code:

```suggestion
    bone_config_path = bone_dir / CONFIG_NAME
    miss_config_path = miss_dir / CONFIG_NAME
```

scripts/convert-bone-to-miss.py
```diff
@@ -0,0 +1,74 @@
...
+        config = json.load(f)
+
+    config["peft_type"] = "MISS"
+
+    with open(miss_config_path, 'w', encoding='utf-8') as f:
+        json.dump(config, f, indent=2, ensure_ascii=False)
+
+
+    bone_weight_path = os.path.join(bone_dir, SAFETENSORS_WEIGHTS_NAME)
+    miss_weight_path = os.path.join(miss_dir, SAFETENSORS_WEIGHTS_NAME)
```
BenjaminBossan:
```suggestion
    bone_weight_path = bone_dir / SAFETENSORS_WEIGHTS_NAME
    miss_weight_path = miss_dir / SAFETENSORS_WEIGHTS_NAME
```

scripts/convert-bone-to-miss.py
```diff
@@ -0,0 +1,74 @@
...
+from safetensors.torch import save_file
+
+from peft.utils import CONFIG_NAME, SAFETENSORS_WEIGHTS_NAME
+
+
+def convert_bone_to_miss(bone_dir: Path, miss_dir: Path) -> None:
+    """Convert Bone checkpoint files to MiSS format."""
+    bone_config_path = os.path.join(bone_dir, CONFIG_NAME)
+    miss_config_path = os.path.join(miss_dir, CONFIG_NAME)
+
```
BenjaminBossan: The `miss_dir` might not exist yet:

```suggestion
    if not os.path.exists(miss_dir):
        os.makedirs(miss_dir, exist_ok=True)
```

BenjaminBossan (changes requested):
Thanks for the updates, I still found some small issues but nothing big.

I also tested the conversion script and can confirm on my end that the converted model produces the same output.

On top, I tested a different script to see if two newly initialized models, one with Bone and one with MiSS, would yield the same result, and this also passed. So I think the friction to switch from Bone to MiSS should be minimal to users.

Joluck:
finish:)

src/peft/tuners/bone/config.py
```diff
@@ -121,3 +122,8 @@ def __post_init__(self):
         # if target_modules is a regex expression, then layers_pattern should be None
         if isinstance(self.target_modules, str) and self.layers_pattern is not None:
             raise ValueError("`layers_pattern` cannot be used when `target_modules` is a str.")
+
+        warnings.warn(
+            "Bone will be removed in v0.18.0 of PEFT, use `MissConfig` instead. "
```
BenjaminBossan:
```suggestion
            "Bone will be removed in v0.19.0 of PEFT, use `MissConfig` instead. "
```

We plan to make a release, v0.17.0, next week, I think we can get MiSS merged before that! That means there would be very little time to migrate away from Bone. So let's bump the deprecation time to v0.19.0.

Joluck:
@BenjaminBossan finish！

examples/causal_language_modeling/peft_lora_clm_accelerate_ds_zero3_offload.py
```diff
@@ -297,9 +297,9 @@ def test_preprocess_function(examples):
 
         correct = 0
         total = 0
-        assert len(eval_preds) == len(dataset["train"][label_column]), (
-            f"{len(eval_preds)} != {len(dataset['train'][label_column])}"
-        )
+        assert len(eval_preds) == len(
```
BenjaminBossan: Undo

examples/conditional_generation/peft_lora_seq2seq_accelerate_ds_zero3_offload.py
```diff
@@ -247,9 +247,9 @@ def collate_fn(examples):
 
         correct = 0
         total = 0
-        assert len(eval_preds) == len(dataset["train"][label_column]), (
-            f"{len(eval_preds)} != {len(dataset['train'][label_column])}"
-        )
+        assert len(eval_preds) == len(
```
BenjaminBossan: Undo

BenjaminBossan (changes requested):
For some reason, the formatting didn't work as expected, could you please try again? Ensure that you use the same ruff version as is used in the CI, `0.9.10`. Also, you need to add `miss.md` to the `_toctree.yml`.

BenjaminBossan:
@JL-er ruff is still [complaining](https://github.com/huggingface/peft/actions/runs/16650702499/job/47186447597?pr=2604#step:5:17). Could you please double-check that your local setup is correct and then push the fixes?

BenjaminBossan:
Here is the required diff:

```diff
diff --git a/tests/test_custom_models.py b/tests/test_custom_models.py
index 2db07f75..7d72eee5 100644
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -2737,9 +2737,9 @@ class TestMultiRankAdapter:
                 if isinstance(module, BaseTunerLayer):
                     rank_expected = rank_pattern.get(key, r)
                     rank_current = module.lora_A[adapter].weight.shape[0]
-                    assert (
-                        rank_current == rank_expected
-                    ), f"Rank {rank_current} is not equal to expected {rank_expected}"
+                    assert rank_current == rank_expected, (
+                        f"Rank {rank_current} is not equal to expected {rank_expected}"
+                    )
 
 
 class TestLayerRepr:
diff --git a/tests/test_decoder_models.py b/tests/test_decoder_models.py
index ed2d9832..9880745f 100644
--- a/tests/test_decoder_models.py
+++ b/tests/test_decoder_models.py
@@ -563,9 +563,9 @@ class TestDecoderModels(PeftCommonTester):
             and layers[2].mlp.up_proj.lora_A.default.weight.data.storage().data_ptr()
             != layers[3].mlp.up_proj.lora_A.default.weight.data.storage().data_ptr()
         ), "Expected all LoRA adapters to have distinct weights"
-        assert (
-            len([n for n, _ in model.named_parameters() if ".lora_A." in n]) == 8
-        ), "Expected 8 LoRA adapters since we are adding one each for up and down."
+        assert len([n for n, _ in model.named_parameters() if ".lora_A." in n]) == 8, (
+            "Expected 8 LoRA adapters since we are adding one each for up and down."
+        )
         self._test_prepare_for_training(model_id, LoraConfig, config_kwargs.copy())
         self._test_generate(model_id, LoraConfig, config_kwargs.copy())
 
```

JL-er:
> ```diff
> +        assert len([n for n, _ in model.named_parameters() if ".lora_A." in n]) == 8, (
> +            "Expected 8 LoRA adapters since we are adding one each for up and down."
> +        )
> ```

I ran `make style` and these changes appeared.

Joluck:
> @JL-er ruff is still [complaining](https://github.com/huggingface/peft/actions/runs/16650702499/job/47186447597?pr=2604#step:5:17). Could you please double-check that your local setup is correct and then push the fixes?

I ran `make style` and these changes appeared.

Joluck:
@BenjaminBossan  It was my mistake — I was using the wrong version of ruff.

BenjaminBossan:
Hmm, something is wonky with the linter lately, I'm not quite sure what the issue is, but it's not just you who has issues.

Regarding your fix, the diff was not applied correctly. The `+` is still in the code.

JL-er:
> Hmm, something is wonky with the linter lately, I'm not quite sure what the issue is, but it's not just you who has issues.
> 
> Regarding your fix, the diff was not applied correctly. The `+` is still in the code.

> I will revert to the version before running `make style` and run `make style` again.

Joluck:
> Hmm, something is wonky with the linter lately, I'm not quite sure what the issue is, but it's not just you who has issues.
> 
> Regarding your fix, the diff was not applied correctly. The `+` is still in the code.

I tested it on my device, and it should be fixed now.

BenjaminBossan:
Hmm, some [tests are failing](https://github.com/huggingface/peft/actions/runs/16672757067/job/47198147731?pr=2604#step:8:15310) @JL-er, could you please investigate?

Joluck:
> Hmm, some [tests are failing](https://github.com/huggingface/peft/actions/runs/16672757067/job/47198147731?pr=2604#step:8:15310) @JL-er, could you please investigate?

I will check it.

BenjaminBossan:
@JL-er We actually plan to release today, do you think you will fix the issues soon? LMK if you need my support.

Joluck:
> @JL-er We actually plan to release today, do you think you will fix the issues soon? LMK if you need my support.

give me 2 hours

Joluck:
@BenjaminBossan How can I test test_custom_models.py individually?

BenjaminBossan:
Personally, I would run `pytest tests/test_custom_models.py -k miss -v`. This results in the errors you can see on CI.

Joluck:
> Personally, I would run `pytest tests/test_custom_models.py -k miss -v`. This results in the errors you can see on CI.

image
last issue

BenjaminBossan:
> last issue

:fire: 

Just from that screenshot, it's hard to say what the problem is. If you want me to check, please push your changes and I can run those tests on my machine.

Joluck:
> > last issue
> 
> 🔥
> 
> Just from that screenshot, it's hard to say what the problem is. If you want me to check, please push your changes and I can run those tests on my machine.

I figured out the reason — it's because the initialization is zero in mini mode.

Joluck:
> > > last issue
> > 
> > 
> > 🔥
> > Just from that screenshot, it's hard to say what the problem is. If you want me to check, please push your changes and I can run those tests on my machine.
> 
> I figured out the reason — it's because the initialization is zero in mini mode.

    (
        "MiSS-Mini Same",
        "miss-mini",
        MissConfig,
        {"target_modules": ["lin0"], "init_weights": "mini", "r": 2},
        {"target_modules": ["lin0"], "init_weights": "mini", "r": 2},
    ),
    (
        "MiSS-Mini Different",
        "miss-mini",
        MissConfig,
        {"target_modules": ["lin0"], "init_weights": "mini", "r": 2},
        {"target_modules": ["lin1"], "init_weights": "mini", "r": 2},
    ),Is it possible to remove this option? @BenjaminBossan

BenjaminBossan:
> Is it possible to remove this option?

So IIUC, with "mini" initialization, the MiSS weights are always identical, there is no random component? If so, I'd say it's fine to remove the option and to put a comment there instead that mentions something like: 

> \# Not testing "mini" initialization targeting the same layer, because ...

Joluck:
> > Is it possible to remove this option?
> 
> So IIUC, with "mini" initialization, the MiSS weights are always identical, there is no random component? If so, I'd say it's fine to remove the option and to put a comment there instead that mentions something like:
> 
> > # Not testing "mini" initialization targeting the same layer, because ...

Yes, since MiSS is essentially initialized to zeros, I added random initialization to Bone (when init_weights=False) just for this test.

Joluck:
@BenjaminBossan I think we've resolved all the issues.

Joluck:
@BenjaminBossan Could we add MiSS evaluation to peft_comparison?

BenjaminBossan:
> Could we add MiSS evaluation to peft_comparison?

We'll try to add it soon, currently facing issues with NVIDIA drivers on our EC2. We forgot to add an experiment in this PR, but I think just using the default settings should be good. The expected score should be the same as for Bone, right?

Joluck:
> > Could we add MiSS evaluation to peft_comparison?
> 
> We'll try to add it soon, currently facing issues with NVIDIA drivers on our EC2. We forgot to add an experiment in this PR, but I think just using the default settings should be good. The expected score should be the same as for Bone, right?

Yes, the settings are the same as for Bone, and I also want to see the performance of Miss-Mini. I’d like to include this experiment in my paper.

BenjaminBossan:
> Yes, the settings are the same as for Bone, and I also want to see the performance of Miss-Mini. I’d like to include this experiment in my paper.

I see. Could you please create a PR to add those two experiments?

Joluck:
> > Yes, the settings are the same as for Bone, and I also want to see the performance of Miss-Mini. I’d like to include this experiment in my paper.
> 
> I see. Could you please create a PR to add those two experiments?

I don’t have any machines available for training recently, so I may need to trouble you to run the tests. Keep the settings for default and Bat unchanged. For the Miss-Mini test, simply set `r=8, mini_r=8` or `r=64, mini_r=64` to verify that good performance can still be maintained under such small parameter settings.

BenjaminBossan:
> I don’t have any machines available for training recently, so I may need to trouble you to run the tests.

No worries, just add the experiments (i.e. the PEFT configs), we will run the tests on our machine.

Joluck:
> > I don’t have any machines available for training recently, so I may need to trouble you to run the tests.
> 
> No worries, just add the experiments (i.e. the PEFT configs), we will run the tests on our machine.

Would it be feasible to directly provide the JSON configuration? The `default` and `bat` settings can follow those of BONE, so that when BONE is removed later, the experiments will not be missing.
miss-mini：
{
  "auto_mapping": null,
  "bias": "none",
  "exclude_modules": null,
  "inference_mode": false,
  "init_weights": "mini",
  "layers_pattern": null,
  "layers_to_transform": null,
  "modules_to_save": null,
  "peft_type": "MISS",
  "r": 64,
  "mini_r": 64,
  "revision": null,
  "target_modules": [
    "v_proj",
    "q_proj"
  ],
  "task_type": null
}

Joluck:
@BenjaminBossan  Do I need to submit a PR?

BenjaminBossan:
No, it's fine, I can create the experiments, hopefully I'll find time later today to finish this.

PR #2606, opened 2025-06-24
enable method comparison to device-agnostic, to enable it on other accelerators like xpu

yao-matrix:


githubnemo:
I'm in favor of supporting XPU.

Regarding official benchmark numbers, if there was a device that we could run the benchmarks on, I would even go as far as maintaining a second set of benchmarks. But there doesn't seem to be a way, currently, so this feature will only benefit those people who want to run the benchmarks locally.

BenjaminBossan:
Okay, then let's do it!

Before proceeding with this PR: It will probably conflict with #2602, right? How would you suggest to proceed? First merge #2602 and then update this PR? All the results files might need updating, but it should hopefully just be some simple string replacements.

PR #2611, opened 2025-06-25
FIX: Avoid possible CUDA Graph re-record when hotswapping LoRAs.

keepdying:
Motivation:

In some cases, recorded CUDA Graph gets silently invalided if `TORCH_LOGS="perf_hints"` not provided, due to data pointer changes.

Torch log:

```
static input data pointer changed.
...
input name: arg17_1. data pointer changed from x to y. input stack trace:
...
  File "/opt/conda/lib/python3.11/site-packages/diffusers/models/resnet.py", line 346, in forward
    temb = self.time_emb_proj(temb)[:, :, None, None]
  File "/opt/conda/lib/python3.11/site-packages/peft/tuners/lora/layer.py", line 727, in forward
...
```

This leads to an [unexpected CUDA Graph re-record.](https://github.com/pytorch/pytorch/blob/134179474539648ba7dee1317959529fbd0e7f89/torch/_inductor/cudagraph_trees.py#L2151)

We can avoid this by copying LoRA weights in place without changing underlying tensor itself. I didn't see any performance regressions with in place copy.

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -4631,7 +4631,9 @@ class TestHotSwapping:
     # it is important to check hotswapping small to large ranks and large to small ranks
     @pytest.mark.parametrize("ranks", [(11, 11), (7, 13), (13, 7)])
     def test_hotswapping_compiled_model_does_not_trigger_recompilation(self, ranks):
-        with torch._dynamo.config.patch(error_on_recompile=True):  # raise an error on recompilation
+        dynamo_config_ctx = torch._dynamo.config.patch(error_on_recompile=True)
+        inductor_config_ctx = torch._inductor.config.patch("triton.cudagraph_unexpected_rerecord_limit", 0)
+        with dynamo_config_ctx, inductor_config_ctx:
             self.check_hotswap(do_hotswap=True, ranks=ranks, alpha_scalings=ranks)
 
     def test_no_hotswapping_compiled_model_triggers_recompilation(self):
--- a/src/peft/utils/hotswap.py
+++ b/src/peft/utils/hotswap.py
@@ -480,7 +480,7 @@ def hotswap_adapter_from_state_dict(
             # either
             # - adapters had the same rank
             # - adapters were padded with prepare_model_for_compiled_hotswap and 2nd adapter was larger
-            old_val.data = new_val.data
+            old_val.data.copy_(new_val.data)
         else:
             # if 2nd adapter was smaller, ensure to fill up to adapter dimension and set the rest to zeros
             if old_val.dim() not in (2, 4):
@@ -492,10 +492,10 @@ def hotswap_adapter_from_state_dict(
             # Linear or Conv2d: the check for dim 0 or 1 works for both of these layer types
             if old_val.shape[0] > new_val.shape[0]:
                 old_val.data.fill_(0)
-                old_val.data[: new_val.shape[0]] = new_val.data
+                old_val.data[: new_val.shape[0]].copy_(new_val.data)
             elif old_val.shape[1] > new_val.shape[1]:
                 old_val.data.fill_(0)
-                old_val.data[:, : new_val.shape[1]] = new_val.data
+                old_val.data[:, : new_val.shape[1]].copy_(new_val.data)
             else:
                 raise ValueError(
                     f"Incompatible shapes found for LoRA weights {key}: {old_val.shape} vs {new_val.shape}. Please "


yemreck:
todo:
tests don’t cover this case yet. didnt have time to make an isolated repro.

tests/test_gpu_examples.py
```diff
@@ -4631,7 +4631,9 @@ def check_hotswap(self, do_hotswap, ranks, alpha_scalings):
     # it is important to check hotswapping small to large ranks and large to small ranks
     @pytest.mark.parametrize("ranks", [(11, 11), (7, 13), (13, 7)])
     def test_hotswapping_compiled_model_does_not_trigger_recompilation(self, ranks):
-        with torch._dynamo.config.patch(error_on_recompile=True):  # raise an error on recompilation
```
BenjaminBossan, line 4634: Could you please conserve the comment and add a similar one for `inductor_config_ctx`?

BenjaminBossan (commented):
Thanks for this PR. I'm not an expert in all the intricacies of `torch.compile`, so this is very welcome. If you have an example of where this helps, please share it and we can work on a unit test.

yemreck:
It should help all re-record cases related to weight change but even current tests have some weird re-record with current pr (cudagraphs so finicky to get it working properly). If you profile test case with torch.profiler, you should see "CUDAGraphs.record_function" only once but we see it twice. 

My case is in company code so ill update the pr once I manage to get a minimal working repro.

BenjaminBossan:
> My case is in company code so ill update the pr once I manage to get a minimal working repro.

Thanks for effort.

sayakpaul:
Thanks very much for the PR.

> If you profile test case with torch.profiler, you should see "CUDAGraphs.record_function" only once but we see it twice.

You meant with this PR, we see it only once? 👁️ The PR looks good to me. We should merge after a unit-test.

@anijain2305 if you have comments.

anijain2305:
@zou3519 on improving the ux for cudagraphs

yemreck:
Turns out, we just needed a couple of configs to raise error on static input changes. also updated tests to include a third forward pass since first forward pass always going to be `run_eager` to warmup.


Current main
```
================================================ short test summary info ================================================
FAILED tests/test_gpu_examples.py::TestHotSwapping::test_hotswapping_compiled_model_does_not_trigger_recompilation[ranks0] - RuntimeError: static input data pointer changed.
FAILED tests/test_gpu_examples.py::TestHotSwapping::test_hotswapping_compiled_model_does_not_trigger_recompilation[ranks2] - RuntimeError: static input data pointer changed.
```

pr
```
tests/test_gpu_examples.py::TestHotSwapping::test_hotswapping_compiled_model_does_not_trigger_recompilation[ranks0] PASSED [ 33%]
tests/test_gpu_examples.py::TestHotSwapping::test_hotswapping_compiled_model_does_not_trigger_recompilation[ranks1] PASSED [ 66%]
tests/test_gpu_examples.py::TestHotSwapping::test_hotswapping_compiled_model_does_not_trigger_recompilation[ranks2] PASSED [100%]
```

tests/test_gpu_examples.py
yemreck: added extensive comments to cover both dynamo error on recompile and new configs

yemreck:
> Thanks very much for the PR.
> 
> > If you profile test case with torch.profiler, you should see "CUDAGraphs.record_function" only once but we see it twice.
> 
> You meant with this PR, we see it only once? 👁️ The PR looks good to me. We should merge after a unit-test.
> 
> @anijain2305 if you have comments.

Yes, with this PR, CUDAGraph.record_function will be called only once. Third and later calls will be replayed.

This improves initial forward call after hotswap greatly in scenarios where LoRas swapped frequently

sayakpaul (approved):

BenjaminBossan (approved):
Thanks a lot @keepdying for identifying this `torch.compile` issue, fixing it, and extending the test case to cover it. I tested this locally (PR CI does not use GPU) and the test case indeed fails without the fix and works with it.

The failing CI is unrelated, so this PR is good to be merged.

yemreck:
Glad I could help. Thanks for this great library

sayakpaul:
@BenjaminBossan do we wanna propagate this to `diffusers` too?

BenjaminBossan:
@sayakpaul The fix should not need propagating, so I assume you mean the tests should be updated. Yes, indeed, I was starting with updating the diffusers test in PEFT: #2619. If that is good, we can apply the same logic to the diffusers tests.

PR #2614, opened 2025-06-25
Make Task Type validation case insensitive

### LINKED ISSUE #2613: Make Task Type validation case insensitive
### Feature request

This would be a simple QOL improvement.
From line 66 in ``config.py``:
```python
       if (self.task_type is not None) and (self.task_type not in list(TaskType)):
            raise ValueError(
                f"Invalid task type: '{self.task_type}'. Must be one of the following task types: {', '.join(TaskType)}."
            )
```
In this case, if ``task_type == 'causal_lm'`` the validation will simply fail. 

Mistakes like the one above can easily happen and are a bit annoying to run into.
As all the task type names defined in ``TaskType`` are very different, there does not seem to be a reason for this check to be case-sensitive.

### Your contribution

Submitted the one-line PR needed for the fix.

cartesian-plane:
Fixes #2613

### DIFF
--- a/src/peft/config.py
+++ b/src/peft/config.py
@@ -63,7 +63,7 @@ class PeftConfigMixin(PushToHubMixin):
 
     def __post_init__(self):
         # check for invalid task type
-        if (self.task_type is not None) and (self.task_type not in list(TaskType)):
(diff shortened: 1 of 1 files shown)


githubnemo:
See discussion in #2613.

PR #2617, opened 2025-06-26
ENH Method comparison: temporary and cancelled result files include timestamp

BenjaminBossan:
In #2593, the timestamp was removed from the file name of result files. This makes sense for the proper results, as those should have unique file names and are tracked in git. However, for temporary and cancelled results, this is not true. Therefore, the timestamp is added back in.

Moreover, I applied ruff to the `MetaMathQA/` directory (it's not applied automatically) and fixed some imports. Ruff seems to get confused about local modules, thus the `data` and `utils` import are treated differently, but IMO no big deal.

### DIFF
--- a/method_comparison/MetaMathQA/run.py
+++ b/method_comparison/MetaMathQA/run.py
@@ -25,14 +25,11 @@ import random
 import sys
 import textwrap
 import time
-from contextlib import nullcontext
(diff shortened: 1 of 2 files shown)


githubnemo (approved):
LGTM, thanks :)

PR #2618, opened 2025-06-27
FIX Update signature for resolve_lora_variant

BenjaminBossan:
The function signature was missing `**kwargs`, which results in a [failure](https://github.com/huggingface/peft/actions/runs/15916801267/job/44895818094) after merging #2571.


BenjaminBossan:
~~Note: CI cancelled for now as we're waiting for another PR to be merged, which makes tests pass again with transformers 4.53.0~~

githubnemo (approved):

PR #2619, opened 2025-06-27
TST Update diffusers hotswap tests

BenjaminBossan:
When the diffusers hotswap tests were added to PEFT in #2120, the diffusers test was marked as xfail because hotswapping was not yet implemented in diffusers. This has long been achieved but the test was not updated.

This PR now updates the diffusers test in PEFT and removes the xfail. The new test is basically a copy of the corresponding test in diffusers. Moreover, I enhanced the test according to #2611 to also ensure that there are no CUDA graph re-records.

Note:

- I ran the tests locally. Without the fix from #2611, they fail as expected, and with that fix, the tests pass.
- CI is not passing at the moment for unrelated reasons, so I canceled it, I will trigger it once the fix is merged to main.

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -22,6 +22,7 @@ from copy import deepcopy
 from dataclasses import dataclass
 from typing import Any, Union
 
+import numpy as np
 import pytest
 import torch
 from accelerate import infer_auto_device_map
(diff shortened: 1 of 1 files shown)


sayakpaul (approved):
It looks good. If you want to remove these tests since we test them in `diffusers` any way, we can do that too.

BenjaminBossan:
> If you want to remove these tests since we test them in `diffusers` any way, we can do that too.

I think it's better to keep the tests as otherwise, we might otherwise add changes to PEFT that break hot-swapping in diffusers.

PR #2630, opened 2025-07-03
FIX: Faulty OFT parameter device test

BenjaminBossan:
There is an error in an OFT test because `.cpu()` is called on a parameter instead of a module. Calling it on parameter is not an in-place operation, so it has no effect.

FYI @zqiu24


zqiu24:
@BenjaminBossan Thanks, can you point me to the test?

BenjaminBossan:
> Thanks, can you point me to the test?

https://github.com/huggingface/peft/blob/45996a1d6e00b0951424a210d24aa4058c7b0dcf/tests/test_common_gpu.py#L1882

PR #2633, opened 2025-07-04
FIX: Create mask function signature change in transformers 4.53.1

BenjaminBossan:
We use `create_mask_for_generate` from transformers. It was introduced in v4.53.0 but in v4.53.1, the function signature was changed to include `position_ids` as mandatory argument:

https://github.com/huggingface/transformers/pull/39194

This breaks our function call in PEFT. This PR fixes the function call by passing position_ids. This in turn would break the function call with transformers v4.53.0, thus a strict version check is being used for >= v4.53.1.


BenjaminBossan:
@githubnemo This fix is required to make our CI pass again.

githubnemo (approved):

PR #2635, opened 2025-07-07
Fix #2634: Allow peft_type to be a string

### LINKED ISSUE #2634: Auto-tagging does not support customized peft type
### System Info

`peft v0.16.0` includes the auto-tagging in 7995572 #2599 . However, I got this error when I am using customized `peft_type` like `config_dict["peft_type"] = "MYLORA"` and I believe this use case is valid as [config.py](https://github.com/huggingface/peft/blob/main/src/peft/config.py#L310) defines type to be either `Enum` or `str` (`peft_type: Optional[Union[str, PeftType]]`).

```
File ".../peft/peft_model.py", line 1446, in _get_peft_specific_model_tags
    peft_method = self.active_peft_config.peft_type.value
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'value'
```

### Who can help?

_No response_

### Reproduction

```python
    with open(os.path.join(model_name_or_path, "config.json"), "r") as f:
        config_dict = json.load(f)
        config_dict["peft_type"] = "MYLORA"
    config = PeftConfig(**config_dict)
```
And it will throw exception when save the model.

### Expected behavior

Save adapter should not throw exception when using a customized `peft_type`

githubnemo:
The auto-tagging code assumed that every `PeftConfig.peft_type` value is an Enum value but when adding custom types without modifying the enum it is possible to have strings as well (and the interface supports that).

This change allows for string values of `PeftConfig.peft_type` in the auto-tagging code.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -1443,7 +1443,9 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
         base model is important for enabling support for HF inference providers but it also makes models more
         searchable on the HF hub.
         """
-        peft_method = self.active_peft_config.peft_type.value
+        peft_method = self.active_peft_config.peft_type
+        if not isinstance(peft_method, str):
+            peft_method = peft_method.value
 
         tags = []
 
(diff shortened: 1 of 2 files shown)


BenjaminBossan (approved):
Thanks for the quick fix.

PR #2637, opened 2025-07-09
[WIP] ENH: Adapter injection based on state_dict

BenjaminBossan:
Make it possible to inject the PEFT adapters based on a state_dict instead of the PEFT config.

See https://github.com/huggingface/diffusers/issues/11874 for context.

## Description

Right now, when creating a PEFT adapter like LoRA, the adapter layers are injected based on the PEFT config, most notably the entries in `target_modules`, but other arguments also play into this. Generally, this is a good approach, but it breaks down in some situations. For instance, in diffusers, we often have the situation that the checkpoint was created without PEFT/diffusers, thus there is no PEFT config, only the `state_dict`. To load these checkpoints in diffusers, the current approach is to reverse-engineer a valid PEFT config based on the keys in the `state_dict`.

Unfortunately, this is error prone. Moreover, not every combination of `state_dict` keys can be easily expressed in a PEFT config through a combination of `target_modules`, `exclude_modules`, etc. Yes, in theory everything can be expressed by passing `target_module=<regex_pattern>`, but reverse-engineering such a regex correctly and efficiently is very hard (and thus currently not done).

This PR implements a completely different approach to inject adapters. Instead of relying on the PEFT config to determine which layers to target, it takes the `state_dict` directly as the source of truth. This should allow to exactly match what is desired.

## Implementation details

I took care to implement this change in a way that if no `state_dict` is passed, the exact same code path as previously is taken. The risk of breaking anything should thus be minimized.

Technically, it is not necessary to pass the `state_dict`, we are only interested in the keys. I still called the argument `state_dict`, since that is typically what we have at this point, but this can be easily changed.

I thought it might be a good idea, if the `state_dict` is used, to still check what modules would have been targeted if we had used the PEFT config. Then, the results are compared and a warning is given if they differ. This allows the user to see if the PEFT config is not correctly specified. While running some diffusers tests, I never encountered this warning, which is good. However, if we plan, for instance, to get rid of all the reverse engineering of the PEFT config in diffusers, it would make more sense to not give this warning.

## Caveats

When the original LoRA model was using `target_parameters`, injecting from `state_dict` will not work correctly. The problem is that the `state_dict` looks the same, whether the module or a parameter was targeted. Therefore, we cannot correctly determine the user's intent.

For now, what I decided to do is:

1. Always assume that `target_modules` is meant, as it's the far more common occurrence.
2. When we detect `target_parameters` while using `state_dict` for injection, we raise an error.
3. If we don't detect this, injection might just slip through, resulting in modules being targeted (if they are valid modules) instead of parameters.
4. Document that these two features don't work together.

I think overall, this is not too concerning, as both features are rather niche and thus unlikely to be used in conjunction.

## Related changes

While working on this PR, I made a couple of related, though not strictly necessary, changes:

- Refactor tests in `test_low_level_api.py` to use pytest instead of unittest
- Add default target modules for LoHa and LoKr (just copying LoRA)
- Most PEFT method's model classes like `LoraModel` had an `__init__` that effectively just called `super()` with the same arguments. I removed these `__init__` methods.

## Testing with Diffusers

To test this PR right now, install this branch directly. Then, inside of diffusers, change this line:

https://github.com/huggingface/diffusers/blob/754fe85cace17ae8615d53578d6d842c9e4a1bd9/src/diffusers/loaders/peft.py#L321

```diff
- inject_adapter_in_model(lora_config, self, adapter_name=adapter_name, **peft_kwargs)
+ inject_adapter_in_model(lora_config, self, adapter_name=adapter_name, state_dict=state_dict, **peft_kwargs)
```

That's it. I tested the change with SD:

`RUN_SLOW=1 pytest tests/lora/test_lora_layers_sd.py -v`

and all tests passed. A new test based on the original issue of https://github.com/huggingface/diffusers/issues/11874 should be added to confirm that this PR solves this difficult case.


src/peft/tuners/tuners_utils.py
```diff
@@ -517,18 +540,67 @@ def inject_adapter(
...
+                    unmatched_modules.append(key)
+                else:
+                    self.targeted_module_names.append(key)
+                    parent, target, target_name = _get_submodules(model, key)
+                    self._check_target_module_compatiblity(peft_config, model, target_name)
+                    ctx = init_empty_weights if low_cpu_mem_usage else nullcontext
+                    with ctx():
+                        self._create_and_replace(
+                            peft_config, adapter_name, target, target_name, parent, current_key=key
+                        )
```
BenjaminBossan, line 580: Note to reviewers: This is the exact same code as before, just indented by one level.

BenjaminBossan:
FYI @sayakpaul it would be great if you could help with testing.

sayakpaul:
I got the following error:

```
  File "/fsx/sayak/diffusers/check_flux_lora.py", line 15, in <module>
    pipeline.load_lora_weights("glif/l0w-r3z")
  File "/fsx/sayak/diffusers/src/diffusers/loaders/lora_pipeline.py", line 2198, in load_lora_weights
  File "/fsx/sayak/diffusers/src/diffusers/loaders/lora_pipeline.py", line 2274, in load_lora_into_transformer
    transformer.load_lora_adapter(
  File "/fsx/sayak/diffusers/src/diffusers/loaders/peft.py", line 321, in load_lora_adapter
    inject_adapter_in_model(lora_config, self, adapter_name=adapter_name, state_dict=state_dict, **peft_kwargs)
  File "/fsx/sayak/peft/src/peft/mapping.py", line 82, in inject_adapter_in_model
    peft_model = tuner_cls(
  File "/fsx/sayak/peft/src/peft/tuners/lora/model.py", line 144, in __init__
    super().__init__(model, config, adapter_name, low_cpu_mem_usage=low_cpu_mem_usage, **kwargs)
  File "/fsx/sayak/peft/src/peft/tuners/tuners_utils.py", line 206, in __init__
    self.inject_adapter(self.model, adapter_name, low_cpu_mem_usage=low_cpu_mem_usage, state_dict=state_dict)
  File "/fsx/sayak/peft/src/peft/tuners/tuners_utils.py", line 602, in inject_adapter
    1/0
ZeroDivisionError: division by zero
```

Code:

```py
from diffusers import DiffusionPipeline
import torch 

pipeline = DiffusionPipeline.from_pretrained(
    "black-forest-labs/FLUX.1-dev", torch_dtype=torch.bfloat16
)
pipe.enable_model_cpu_offload()
pipe_kwargs = {
    "prompt": "A heavy metal rock band made up of farm animals on stage at a concert",
    "height": 1024,
    "width": 1024,
    "guidance_scale": 3.5,
    "num_inference_steps": 50,
    "max_sequence_length": 512,
}
pipeline.load_lora_weights("glif/l0w-r3z")
image = pipeline(**pipe_kwargs).images[0]
image.save("chicken.png")
```

BenjaminBossan:
Thx for testing. This is a good result, as this is exactly the bug you reported in https://github.com/huggingface/diffusers/issues/11874, where layers are wrongly excluded because `exclude_modules` contains `proj_out`.

I removed the `1/0` and re-ran your example with 20 steps and 4bit bnb. This is what I got, does it look right?
output-foo

sayakpaul:
it does!

BenjaminBossan:
LMK if you have other examples I should test, including ones that already work, so that I can ensure that they keep on working. Also, what other diffusers tests should I run?

If you think this is a good direction to take, I will continue with the PR. For the time being, even if this works, I guess we keep the old approach in diffusers in case users have old PEFT versions installed. But do you think in the future, we can completely omit the reverse engineering of `target_modules`, `exclude_modules` etc. or are they needed for other purposes?

sayakpaul:
> LMK if you have other examples I should test, including ones that already work, so that I can ensure that they keep on working. Also, what other diffusers tests should I run?

There's currently no `diffusers` test that would catch it as mentioned in https://github.com/huggingface/diffusers/issues/11874. https://github.com/huggingface/diffusers/issues/11874#issuecomment-3044380033 is a good enough representation.

> If you think this is a good direction to take, I will continue with the PR. For the time being, even if this works, I guess we keep the old approach in diffusers in case users have old PEFT versions installed. 

We mandate `peft==0.15.0` already:
https://github.com/huggingface/diffusers/blob/2d3d376bc00a11afb9e3c3e51a27bb2ef2f28b11/setup.py#L119

So, in this case, I won't mind lifting that up to the latest release that will contain this change. It will help us get rid of nasty issues anyway.

> But do you think in the future, we can completely omit the reverse engineering of target_modules, exclude_modules etc. or are they needed for other purposes?

If we can derive it fully from the state_dict (i.e., PEFT takes care of it internally), then yes, why not! Happy to work with you to facilitate that from `diffusers`.

I will review the PR tomorrow before going for my time-off starting 11th July.

src/peft/tuners/tuners_utils.py
```diff
@@ -498,11 +514,18 @@ def inject_adapter(
...
+        # MATCHING & CREATING MODULES #
+        ###############################
+
         existing_adapter_map = {}
         for key, module in named_modules:
             if isinstance(module, BaseTunerLayer):
                 existing_adapter_map[key] = module
 
+        # TODO: check if this the most robust way
+        state_dict_keys = {k.rsplit(".", 2)[0] for k in state_dict} if state_dict is not None else set()
```
sayakpaul: Are these state dict keys or module keys? I think it's the latter no? For example, `foo.bar.weight`. Here the module and submodule are `foo` and `bar`. If we were to just obtain state dict keys, a simple `state_dict.keys()` would have sufficed. So, I think we should consider renaming it.

src/peft/tuners/tuners_utils.py
```diff
@@ -517,18 +540,66 @@ def inject_adapter(
...
+                        )
             else:
-                self.targeted_module_names.append(key)
-                parent, target, target_name = _get_submodules(model, key)
-                self._check_target_module_compatiblity(peft_config, model, target_name)
-                ctx = init_empty_weights if low_cpu_mem_usage else nullcontext
-                with ctx():
-                    self._create_and_replace(peft_config, adapter_name, target, target_name, parent, current_key=key)
+                # use the state_dict to match modules instead
+                if key not in state_dict_keys:
```
sayakpaul: In case, `state_dict` is none these keys will be empty. Just flagging.

sayakpaul (commented):
Thanks a lot for continuing to give your attention to the diffusers community and fixing the nasty stuff. PR looks very reasonable.

src/peft/tuners/tuners_utils.py
BenjaminBossan: We check above if `state_dict` is `None` or not, so this is safe here.

BenjaminBossan:
@sayakpaul I updated the PR to be in a pretty mature state. I believe as a next step, we could combine it with the necessary changes in diffusers and do some testing to confirm that the new approach works. If for some reason it is not viable, I don't think we need to proceed with the PR, as I'm not aware of any need for this outside of diffusers.

But there is no hurry, we can work on this after your return.

sayakpaul:
Thanks, @BenjaminBossan. I will introduce minimal changes on the `diffusers` end as a PoC to test this PR. I don't see any reason not to graduate this PR into a merge, honestly.

sayakpaul:
https://github.com/huggingface/diffusers/pull/11999

docs/source/developer_guides/low_level_api.md
```diff
@@ -87,6 +87,21 @@ DummyModel(
...
+```python
+from safetensors.torch import load_file
+
+model = ...
+state_dict = load_file(<path-to-safetensors-file>)
+lora_config = LoraConfig(...)
+model = inject_adapter_in_model(lora_config, model, state_dict=state_dict)
+```
+
+In this case, PEFT will use the `state_dict` as reference for which layers to target instead of using the PEFT config. As a user, you don't have to set the exact `target_modules` of the PEFT config for this to work. However, you should still pass a PEFT config of the right type, in this example `LoraConfig`, you can leave the `target_modules` as `None`. Be aware that this still only creates the uninitialized PEFT layers, the values from the `state_dict` are not used to populate the model weights.
```
githubnemo: A hint on how to proceed from here to load the weights from the state dict would surely be appreciated by the user.

src/peft/tuners/tuners_utils.py
```diff
@@ -510,11 +527,21 @@ def inject_adapter(
...
         for key, module in named_modules:
             if isinstance(module, BaseTunerLayer):
                 existing_adapter_map[key] = module
 
+        # TODO: check if this the most robust way
+        module_names: set[str] = set()
+        if state_dict is not None:
+            prefix = PEFT_TYPE_TO_PREFIX_MAPPING[peft_config.peft_type]
+            module_names = {k.rsplit("." + prefix, 1)[0] for k in state_dict}
+
```
githubnemo, line 550: This assumes that if I target one `o_proj`, I target all `o_proj`, right? If so, that's not correct if I am selective on layers or if I'm using a very specific regex to target that one specific `o_proj`. I think we'd have to memorize the full path of the state dict key and find that specific name in the module tree to be correct.

tests/test_low_level_api.py
```diff
@@ -49,25 +65,25 @@ def setUp(self):
...
-        peft_state_dict = get_peft_model_state_dict(self.model)
+    def test_get_peft_model_state_dict(self, model):
+        peft_state_dict = get_peft_model_state_dict(model)
 
         for key in peft_state_dict.keys():
             assert "lora" in key
 
-    def test_modules_to_save(self):
-        self.model = DummyModel()
+    def test_modules_to_save(selfv):
```
githubnemo:
```suggestion
    def test_modules_to_save(self):
```

:)

tests/test_low_level_api.py
```diff
@@ -78,20 +94,218 @@ def test_modules_to_save(self):
...
+
+        with hub_online_once(model_id):
+            # use config for injection
+            model = model_cls.from_pretrained(model_id)
+            model = inject_adapter_in_model(config, model)
+            sd_before = get_peft_model_state_dict(model)
+            del model
+
+            model = model_cls.from_pretrained(model_id)
+            # TODO check no warnings
```
githubnemo: still open?

githubnemo (commented):
LGTM in general, a few questions/comments

src/peft/tuners/tuners_utils.py
BenjaminBossan: I don't think so, we only cut off the last part of the name, the parameter name itself. Therefore, the module name is fully qualified and contains the layer index etc. I added a new unit test, `test_inject_from_state_dict_transformers_irregular_targets`, to check for this.

docs/source/developer_guides/low_level_api.md
BenjaminBossan: Done

tests/test_low_level_api.py
BenjaminBossan: Done

tests/test_low_level_api.py
BenjaminBossan: thx for the reminder, done

BenjaminBossan (commented):
Thanks for the feedback @githubnemo, your comments should be addressed.

While working on this, I noticed one flaw: When the original LoRA model was using `target_parameters`, injecting from `state_dict` will not work correctly. The problem is that the `state_dict` looks the same, whether the module or a parameter was targeted. Therefore, we cannot correctly determine the user's intent.

For now, what I decided to do is:

1. Always assume that `target_modules` is meant, as it's the far more common occurrence.
2. When we detect `target_parameters` while using `state_dict` for injection, we raise an error.
3. If we don't detect this, injection might just slip through, resulting in modules being targeted (if they are valid modules) instead of parameters.
4. Document that these two features don't work together.

I think overall, this is not too concerning, as both features are rather niche and thus unlikely to be used in conjunction.

src/peft/tuners/tuners_utils.py
githubnemo: You are correct, thanks for clarifying!

githubnemo (approved):
LGTM! The way we deal with target parameters is fine :)

PR #2638, opened 2025-07-09
FEAT Allow LoRA  to target nn.Parameter

BenjaminBossan:
Normally, `nn.Parameter` cannot be targeted with LoRA adapters. This can be problematic, e.g. when there are MoE layers that use `nn.Parameter` directly, or when there is `nn.Linear` but the weight is passed directly instead of calling forward (e.g. MHA).

It would be possible to craft a solution involving a special LoRA layer for each of the modules that use `nn.Parameter` directly (e.g. `lora.MHA`) but that doesn't scale. This PR is an attempt at implementing a direct way to target `nn.Parameter` making use of `torch.nn.utils.parametrize`.

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -591,6 +591,21 @@ class BaseTuner(nn.Module, ABC):
                 "Please check that exclude_modules was set correctly."
             )
 
+        elif not uses_dummy_target_modules:
+            # If we landed here, it means that at least one module or parameter was adapted, so let's not raise an
+            # error. However, let's warn the user if it seems like
+            # - they wanted to match a module but there was no match
+            # - they wanted to match a parameter but there was no match
+            if peft_config.target_modules and not self.targeted_module_names:
+                warnings.warn(
+                    f"target_modules={peft_config.target_modules} were set but no module was matched.", RuntimeWarning
+                )
+            elif getattr(peft_config, "target_parameters", []) and not self.targeted_parameter_names:
+                warnings.warn(
+                    f"target_parameters={peft_config.target_parameters} were set but no parameter was matched.",
+                    RuntimeWarning,
+                )
+
         tied_target_modules = self._get_tied_target_modules(model=model)
         if tied_target_modules:
             warnings.warn(
--- a/tests/test_initialization.py
+++ b/tests/test_initialization.py
@@ -1435,6 +1435,40 @@ class TestLoraInitialization:
         with pytest.raises(ValueError, match=msg):
             get_peft_model(model, config)
 
+    @pytest.mark.parametrize("target_parameters", [["linear"], ["foobar"], ["foobar.weight"], ["foo", "bar"]])
+    @pytest.mark.parametrize("target_modules", [None, [], ""])
+    def test_valid_no_target_module_nor_target_parameter_match_raises(self, target_parameters, target_modules):
+        model = self.get_model()
+        config = LoraConfig(target_modules=target_modules, target_parameters=target_parameters)
+        msg = re.escape(
+            "No `target_modules` passed but also no `target_parameters` found. Please check the values for "
+            "these arguments."
+        )
+        with pytest.raises(ValueError, match=msg):
+            get_peft_model(model, config)
+
+    def test_target_parameters_wrong_type_raises(self):
+        # Check that target_parameters being a string raises a useful error message -- this is an easy mistake to make
+        # because strings are allowed for target_modules
+        model = self.get_model()
+        msg = "`target_parameters` must be a list of strings or None."
+        with pytest.raises(TypeError, match=msg):
+            LoraConfig(target_parameters="linear.weight")
+
+    def test_valid_target_parameters_invalid_target_modules_warns(self):
+        model = self.get_model()
+        config = LoraConfig(target_modules=["foobar"], target_parameters=["linear.weight"])
+        msg = re.escape("target_modules={'foobar'} were set but no module was matched.")
+        with pytest.warns(RuntimeWarning, match=msg):
+            get_peft_model(model, config)
+
+    def test_valid_target_modules_invalid_target_parameters_warns(self):
+        model = self.get_model()
+        config = LoraConfig(target_modules=["linear"], target_parameters=["foobar.weight"])
+        msg = re.escape("target_parameters=['foobar.weight'] were set but no parameter was matched.")
+        with pytest.warns(RuntimeWarning, match=msg):
+            get_peft_model(model, config)
+
 
 class TestLokrInitialization:
     torch_device = infer_device()


src/peft/tuners/lora/layer.py
```diff
@@ -1759,10 +1761,307 @@ def __repr__(self) -> str:
         return "lora." + rep
 
 
+class _LoraProxy(nn.Module):
+    """This proxies an nn.Parameter that is targeted with LoRA.
```
qgallouedec:
```suggestion
    """This proxies an `nn.Parameter` that is targeted with LoRA.
```

docs/source/developer_guides/lora.md
```diff
@@ -269,6 +269,15 @@ Outer(
 
 The same logic applies to `alpha_pattern`. If you're in doubt, don't try to get fancy with regular expressions -- just pass the full name for each module with a different rank/alpha, preceded by the `^` prefix, and you should be good.
 
+### Targeting `nn.Parameter` directly
+
+> [!WARNING]
+> This feature is experimental and subject to change.
+
+Generally, you should use `target_modules` to target the module (e.g. `nn.Linear`). However, in some circumstances, this is not possible. E.g., many mixture of expert (MoE) layers in HF Transformers, instead of using `nn.Linear`, an `nn.Parameter` is used. PEFT normally overwrites the `forward` method for LoRA, but for `nn.Parameter`, there is none. Therefore, to apply LoRA to that parameter, it needs to be targeted with `target_parameters`. As an example, for [Llama4](https://huggingface.co/collections/meta-llama/llama-4-67f0c30d9fe03840bc9d0164), you can pass: `target_parameters=['feed_forward.experts.gate_up_proj', 'feed_forward.experts.down_proj]`.
```
githubnemo:
```suggestion
Generally, you should use `target_modules` to target the module (e.g. `nn.Linear`). However, in some circumstances, this is not possible. E.g., in many mixture of expert (MoE) layers in HF Transformers, instead of using `nn.Linear`, an `nn.Parameter` is used. PEFT normally overwrites the `forward` method for LoRA, but for `nn.Parameter`, there is none. Therefore, to apply LoRA to that parameter, it needs to be targeted with `target_parameters`. As an example, for [Llama4](https://huggingface.co/collections/meta-llama/llama-4-67f0c30d9fe03840bc9d0164), you can pass: `target_parameters=['feed_forward.experts.gate_up_proj', 'feed_forward.experts.down_proj]`.
```

src/peft/tuners/lora/layer.py
```diff
@@ -1759,10 +1761,307 @@ def __repr__(self) -> str:
         return "lora." + rep
 
 
+class _LoraProxy(nn.Module):
```
githubnemo: normally I'd suggest being more specific with the naming (i.e. `LoraParameterProxy`) but I can see that this could become a bit more complicated in the future (e.g., `LoraExpertParameterProxy` or something like that) so I think it's fine for now.

src/peft/tuners/lora/layer.py
```diff
@@ -1759,10 +1761,307 @@ def __repr__(self) -> str:
...
+        if lora_variant is not None:
+            raise ValueError(f"lora.{self.__class__.__name__} does not work with LoRA variants like DoRA.")
+
+        self.r[adapter_name] = r
+        self.lora_alpha[adapter_name] = lora_alpha
+        if lora_dropout > 0.0:
+            # It's not possible to factor out x from lora_B(lora_A(dropout(x))), so dropout can't be correctly implemented
+            lora_dropout_layer = nn.Dropout(p=lora_dropout)
+        else:
+            lora_dropout_layer = nn.Identity()
```
githubnemo: I think we should just raise an exception here in case `lora_dropout` > 0

src/peft/tuners/lora/layer.py
```diff
@@ -1759,10 +1761,307 @@ def __repr__(self) -> str:
...
+    def get_delta_weight(self, adapter_name, *args, **kwargs):
+        if self.num_experts == 1:
+            delta_weight = Linear.get_delta_weight(self, adapter_name, *args, **kwargs).T
+        else:
+            weight_A = self.lora_A[adapter_name].weight
+            weight_B = self.lora_B[adapter_name].weight
+            weight_A = weight_A.reshape(self.num_experts, -1, weight_A.shape[-1])
+            weight_B = weight_B.reshape(weight_B.shape[0], -1, self.num_experts)
+            # fan_in_fan_out must be False, so no transpose call here
+            delta_weight = torch.einsum("m n e, e n d -> e d m", weight_B, weight_A)
```
githubnemo: Wouldn't `s/n/r` in the einsum expression make more sense or am I misreading?

src/peft/tuners/lora/layer.py
```diff
@@ -1759,10 +1761,307 @@ def __repr__(self) -> str:
...
+                delta_weight = self.get_delta_weight(active_adapter)
+            else:
+                delta_weight = delta_weight + self.get_delta_weight(active_adapter)
+
+        base_layer = self.get_base_layer()
+        requires_grad_before = self.get_param().requires_grad
+        nn.utils.parametrize.register_parametrization(
+            base_layer, self.parameter_name, _LoraProxy(delta_weight, num_experts=self.num_experts)
+        )
+        base_layer.parametrizations[self.parameter_name].original.requires_grad_(requires_grad_before)
```
githubnemo, line 2003: I don't understand why it is necessary to reset `requires_grad` of the original parameter here. This warrants a comment, I think

src/peft/tuners/tuners_utils.py
```diff
@@ -596,6 +615,38 @@ def inject_adapter(
...
+                    with ctx():
+                        self._create_and_replace(
+                            peft_config,
+                            adapter_name,
+                            target,
+                            target_name,
+                            parent,
+                            current_key=key,
+                            parameter_name=param_name.rpartition(".")[-1],
+                        )
```
githubnemo: clarification: this also 'accidentally' allows for targeting multiple parameters of the same module by nesting lora layers? does it work at all? if it does then it should definitely be documented here.

(I understood the comment above as "We might need to update the adapter names dict on existing layers")

src/peft/tuners/tuners_utils.py
```diff
@@ -1049,6 +1101,10 @@ def check_target_module_exists(config, key: str) -> bool | re.Match[str] | None:
         if any(re.match(rf"(^|.*\.){m}($|\..*)", key) for m in modules_to_save):
             return _ExcludedModule()
 
+    if config.target_modules is None:
+        # this is allowed if config.target_parameters are specified
```
githubnemo, line 1130: is this checked somewhere?

src/peft/utils/save_and_load.py
```diff
@@ -213,10 +213,17 @@ def renamed_dora_weights(k):
     # DEAL WITH EMBEDDINGS
     # check the common embedding layers in `target_modules` to reset `save_embedding_layers` if necessary
     is_embedding_in_target_modules = False
+    embedding_is_targeted = False
+    if hasattr(config, "target_modules"):
+        if isinstance(config.target_modules, str):
+            # TODO: implement this; note: this change is not directly related to the PR, the bug already existed b4
```
githubnemo, line 239: I can take this in a separate PR

tests/test_target_parameters.py
```diff
@@ -0,0 +1,296 @@
...
+            "r": 8,
+            "lora_alpha": 32,
+            "target_modules": None,
+            "lora_dropout": 0.0,
+            "bias": "none",
+            "target_parameters": [
+                "feed_forward.experts.gate_up_proj",
+                "feed_forward.experts.down_proj",
+            ],
+        },
```
githubnemo: Maybe I missed it but there doesn't seem to be a test that checks whether all targeted parameters are getting updated.

src/peft/tuners/lora/layer.py
BenjaminBossan: You're right, I'm going with `_LoraParameterProxy`.

src/peft/tuners/lora/layer.py
BenjaminBossan: Yes, right

src/peft/tuners/lora/layer.py
BenjaminBossan: How about `"o r e, e r i -> e i o"` with

- o: out_features
- r: rank
- e: num experts
- i: in_fetures

?

src/peft/tuners/lora/layer.py
BenjaminBossan: It defaults to `False`, I add a comment.

src/peft/tuners/tuners_utils.py
BenjaminBossan: I'm not quite sure what you mean. If there is a layer with two `nn.Parameter`s, like `gate_down_proj` and `up_proj` in Llama4, the first time we encounter it, we, create a new `lora.ParamWrapper`, the second time we update the existing one. This is okay.

What we don't want is to target the same weight with both `target_modules` and `target_parameters`. For this to happen, the user would have to set something like `target_modules=["weight"]`. I guess this is not impossible, so I'll add a check to prevent the double targeting and a test as well.

src/peft/tuners/tuners_utils.py
```diff
@@ -902,6 +953,7 @@ def _move_adapter_to_device_of_base_layer(self, adapter_name: str, device: Optio
             if any(p.device == meta for p in adapter_layer.parameters()):
                 continue
 
+            # TODO: weight is not necessarily defined here, leading to a NameError, fix that
```
BenjaminBossan, line 981: Note: This is an existing bug and has nothing to do with the PR, just flagging it here.

src/peft/tuners/tuners_utils.py
BenjaminBossan: Right, it should be checked explicitly here, I add this.

src/peft/utils/save_and_load.py
BenjaminBossan: Thx!

tests/test_target_parameters.py
BenjaminBossan: I guess we don't have such a test at all, have we? The closest thing that we have is here:

https://github.com/huggingface/peft/blob/35000fda88fdf29db785783f8a40431cee2cab10/tests/testing_common.py#L1213-L1217

I agree that it could be explicitly tested but IMO that's out of scope of this PR and should be implemented generally. However, I do have a local script that explicitly tests that the targeted parameters are updated and it works.

BenjaminBossan (commented):
Thanks for the reviews @githubnemo and @qgallouedec. The comments should be addressed. I added another test config involving a mixture of normal LoRA and LoRA targeting nn.Parameter and then merging. This results in one test,

`tests/test_target_parameters.py::TestDecoderModelsTargetParameters::test_merge_layers[LoraConfig-config_kwargs2-trl-internal-testing/tiny-Llama4ForCausalLM]`

to fail for me locally with somewhat large numerical differences in the expected outputs. I'm not quite sure why that is. Possibly this is a bug or possibly this just requires higher tolerance, I'll investigate later.

src/peft/tuners/lora/layer.py
githubnemo: Yep, looks good

src/peft/tuners/tuners_utils.py
githubnemo: > I'm not quite sure what you mean. If there is a layer with two nn.Parameters, like gate_down_proj and up_proj in Llama4, the first time we encounter it, we, create a new lora.ParamWrapper, the second time we update the existing one. This is okay.

Yes but I don't see how this can work. We're updating `targeted_parameter_names` but that seems to be only for book-keeping. We only have a singular parameter name in `ParamWrapper`, so the only way to target multiple parameters as I see it now is to wrap the wrapper again. So, if I'm right with this then the comment above this code section should mention that we're wrapping wrappers.

BenjaminBossan: I think it works as expected:

```python

# note: a bit annoying that we need to set target_modules=[] or else q_proj is automatically targeted w/ lora.Linear
config = LoraConfig(target_modules=[], target_parameters=["q_proj.weight"])
```

This results in the following model with 2 LoRA weights, which looks correct to me:

```
PeftModel(
  (base_model): LoraModel(
    (model): OPTForCausalLM(
      (model): OPTModel(
        (decoder): OPTDecoder(
          (embed_tokens): Embedding(50272, 768, padding_idx=1)
          (embed_positions): OPTLearnedPositionalEmbedding(2050, 768)
          (final_layer_norm): LayerNorm((768,), eps=1e-05, elementwise_affine=True)
          (layers): ModuleList(
            (0-11): 12 x OPTDecoderLayer(
              (self_attn): OPTAttention(
                (k_proj): Linear(in_features=768, out_features=768, bias=True)
                (v_proj): Linear(in_features=768, out_features=768, bias=True)
                (q_proj): lora.ParamWrapper(
                  (base_layer): Linear(in_features=768, out_features=768, bias=True)
                  (lora_dropout): ModuleDict(
                    (default): Identity()
                    (other): Identity()
                  )
                  (lora_A): ModuleDict(
                    (default): Linear(in_features=768, out_features=8, bias=False)
                    (other): Linear(in_features=768, out_features=8, bias=False)
                  )
                  (lora_B): ModuleDict(
                    (default): Linear(in_features=8, out_features=768, bias=False)
                    (other): Linear(in_features=8, out_features=768, bias=False)
                  )
                  (lora_embedding_A): ParameterDict()
                  (lora_embedding_B): ParameterDict()
                  (lora_magnitude_vector): ModuleDict()
                )
                (out_proj): Linear(in_features=768, out_features=768, bias=True)
              )
              (activation_fn): ReLU()
              (self_attn_layer_norm): LayerNorm((768,), eps=1e-05, elementwise_affine=True)
              (fc1): Linear(in_features=768, out_features=3072, bias=True)
              (fc2): Linear(in_features=3072, out_features=768, bias=True)
              (final_layer_norm): LayerNorm((768,), eps=1e-05, elementwise_affine=True)
            )
          )
        )
      )
      (lm_head): Linear(in_features=768, out_features=50272, bias=False)
    )
  )
)
```

As we are summing up the `delta_weight`s of each active adapter when creating the proxy, all of them should be applied.

BenjaminBossan: Okay, thx for explaining this internally, I now get what you mean: targeting _different_ parameters on the _same_ module with the _same_ adapter does not work.

Indeed, this is true and I tried for a bit to make it possible. However, at the end of the day, this would require that on the _same_ LoRA wrapper layer, for the _same_ adapter, we would need to have _multiple_ `lora_A` and `lora_B`. This is simply not something we currently support and would make adapter handling more complex (e.g. `lora_A` would need to be a nested `nn.ModuleDict` with first level being adapter (status quo) and second level being parameter name (new)). Therefore, I decided not to support this for the time being. I ensured that users get a nice error message if they try this, though.

It is still possible to achieve a similar result, by using _multiple_ adapters, one for each parameter. I added `test_target_multiple_parameters_on_same_module` to `test_target_parameters.py` to demonstrate this, although I admit that from a user's point of view, this is not ideal.

PR #2641, opened 2025-07-11
add a asdict for dataclass model_config

xliu0105:
The `get_model_config` function in `tuners_utils.py` does not handle model_config of dataclass data type, because some models (such as the lerobot model) have `config` attributes, but the dataclass type cannot be converted into a dictionary in the `get_model_config function`, resulting in an error when using `.get` in the `_get_tied_target_modules` function.

I fixed this problem. Make peft adapt to a wider range of model types.

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -21,6 +21,7 @@ import warnings
 from abc import ABC, abstractmethod
 from contextlib import contextmanager, nullcontext
 from typing import Any, Optional, Union, overload
+from dataclasses import asdict, is_dataclass
 
 import torch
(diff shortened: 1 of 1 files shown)


githubnemo:
Yes, I planned to do this as a follow-up to a potential discussion of https://github.com/huggingface/lerobot/pull/1411 but I think it doesn't hurt to support dataclasses as model config objects, especially since the change is not that big.

Could you just add a unit test for this case in `tests/test_config.py`? We could simply test that `get_model_config` returns a dictionary given a model with a dataclass config.

githubnemo:
@xliu0105 are you still interested in implementing this?

PR #2440, opened 2025-03-19
Draft: Adding GPT2 support to Adaption prompts

efraimdahl:
[Llama-Adapters](https://arxiv.org/abs/2303.16199) unlike the name suggests are model agnostic. This contribution seeks to add an adjustment to the llama-adapter implementation to support GPT2 Models. Currently this is achieved through an additional class `AdaptedAttentionGPT` that wraps the attention layer of GPT2 - type models that handles the difference in attention calculation and the different input formats of the forward function between LLama and GPT transformers.

Currently I am testing that the learning behavior of this implementation is as expected, comparing similar LLama and GPT configurations on the same datasets. It passes initial tests for saving/loading/passing data. 

Llama adapter require that the the initialized adapter should not change the generation of the base model. 
I am having trouble testing for the non-evasiveness of the model as mentioned here, the following test would fail, with or without the adapter. I am looking for alternative ways to test this. 

```python
config=GPT2Config(
            vocab_size=16,
            hidden_size=8, #mapped to n_embd
            n_layers=8, #mapped to n_layers
            num_attention_heads=4, #mapped to n_head
            use_cache=True,
            attn_implementation="eager"
        )
input_ids = torch.LongTensor([[1, 1, 1], [2, 1, 2]]).to(device)
target_ids = torch.LongTensor([[0, 0, 0], [0, 0, 0]]).to(device)
attention_mask = torch.LongTensor([[1, 1, 1], [1, 0, 1]]).to(device)
torch.manual_seed(42)
np.random.seed(42)
random.seed(42)

# Create and compare gpt2 model outputs .
model_gpt2 = GPT2Model(create_test_gpt2_config())
model_gpt2 = model_gpt2.to(device)
a= model_gpt2(input_ids=input_ids, attention_mask=attention_mask)
b= model_gpt2(input_ids=input_ids, attention_mask=attention_mask)
assert_close(a.last_hidden_state, b.last_hidden_state, rtol=0, atol=0)
```

PR #2643, opened 2025-07-13
GPT2 compatible version of LLama-Adapters

efraimdahl:
Cheers, I opened a [similar PR](https://github.com/huggingface/peft/pull/2440) that was closed do to lengthy inactivity on my part. Now with more time on hand, I wanted to revisit this. 

[Llama-Adapters](https://arxiv.org/abs/2303.16199) unlike the name suggests are model agnostic. This contribution seeks to add an adjustment to the llama-adapter implementation to support GPT2 Models. Currently this is achieved through an additional class AdaptedAttentionGPT that wraps the attention layer of GPT2 - type models that handles the differences in attention calculation (fused vs seperate key, query,  value projections) and the different input formats of the forward function between LLama/Mistral and GPT transformers.

In this new PR I have integrated my testing into the refactored tests introduced in [6c8c3c3](https://github.com/huggingface/peft/commit/6c8c3c386edf8dfd74199b30485428153e26bc97), which all pass.

### DIFF
--- a/src/peft/tuners/adaption_prompt/layer.py
+++ b/src/peft/tuners/adaption_prompt/layer.py
@@ -73,9 +73,7 @@ class AdaptedAttentionGPT(_BaseAdaptedAttention):
 
     def __init__(self, model_type, adapter_len, model):
         target_dtype = (
-                model.c_proj.weight.dtype
-                if model.c_proj.weight.dtype not in [torch.int8, torch.uint8]
-                else torch.float32
+            model.c_proj.weight.dtype if model.c_proj.weight.dtype not in [torch.int8, torch.uint8] else torch.float32
         )
         super().__init__(model_type, adapter_len, model, target_dtype=target_dtype)
 
@@ -155,11 +153,9 @@ class AdaptedAttention(_BaseAdaptedAttention):
 
     def __init__(self, model_type, adapter_len, model):
         target_dtype = (
-                model.q_proj.weight.dtype
-                if model.q_proj.weight.dtype not in [torch.int8, torch.uint8]
-                else torch.float32
-            )
-        super().__init__(model_type, adapter_len, model,target_dtype=target_dtype)
+            model.q_proj.weight.dtype if model.q_proj.weight.dtype not in [torch.int8, torch.uint8] else torch.float32
+        )
+        super().__init__(model_type, adapter_len, model, target_dtype=target_dtype)
 
     def forward(self, **kwargs):
         """


src/peft/tuners/adaption_prompt/layer.py
```diff
@@ -17,9 +17,129 @@
...
+    def __init__(self, model_type: str, adapter_len: int, model):
+        """
+        Initialize object.
+        Args:
+            model_type: The transformer model type. This is used to retrieve the right method to
+                compute query states.
+            adapter_len: The length of the adaption prompt to insert.
+            model: The original transformer attention module that is being wrapped.
+        """
+        assert not isinstance(model, AdaptedAttention)
```
BenjaminBossan: Is this a concrete error that you encountered, how would it happen? If the check is indeed needed, it should raise a proper error, say, `ValueError`, with a message that helps users figuring out what went wrong.

src/peft/tuners/adaption_prompt/layer.py
```diff
@@ -17,9 +17,129 @@
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
+from typing import Optional, Tuple, Union
```
BenjaminBossan: No need for `Tuple`, we can use `tuple[...]`.

src/peft/tuners/adaption_prompt/model.py
```diff
@@ -65,7 +65,7 @@ def add_adapter(self, adapter_name: str, config: AdaptionPromptConfig) -> None:
 
         parents = []
         for name, _ in self.model.named_modules():
-            if name.endswith(config.target_modules):
+            if name.endswith(config.target_modules) and not name.endswith(f"_{config.target_modules}"): #accounts for attn vs c_attn
```
BenjaminBossan: Would this not be sufficient?

```suggestion
            if name.endswith(f".{config.target_modules}"):
```

BenjaminBossan (changes requested):
Thanks for reviving the PR. I only did a quick review so far, will do a more in-depth review once the PR has run. For this, however, please ensure to call `make style` on your changes before committing them.

src/peft/tuners/adaption_prompt/layer.py
efraimdahl: Good point! This would never happen because the types would never be mixed. However, when changing the type to match the original implementation `assert not isinstance(model, AdaptedAttentionGPT)`, there is a case to be made, you would not want to overwrite those AdaptionPrompts already in place? I replaced it  with a proper ValueError.

src/peft/tuners/adaption_prompt/layer.py
efraimdahl: This is also covered in the make style check! Thank you for checking that.

src/peft/tuners/adaption_prompt/model.py
efraimdahl: I'm not quite happy with this myself, maybe you have a suggestion to deal with the layer naming conventions. If we do as suggested in the GPT2 model, we would end up attempting to apply the AdaptionPrompt to the 1DConv Layer named c_attn, which is a sublayer of the self attention layer "attn" in GPT2.

src/peft/tuners/adaption_prompt/layer.py
efraimdahl: Ah, looks like `make style` also changed up some other files (making the import order alphabetic) hope that is not too messy.

BenjaminBossan: I'm not quite sure what happened there, but please undo all the unrelated changes.

src/peft/tuners/adaption_prompt/model.py
BenjaminBossan: Would my solution not solve the issue? If the module names are `foo.bar.attn` and `bar.baz.c_attn`, then if we check for `".attn"`, we would not accidentally match `c_attn`.

src/peft/tuners/adaption_prompt/layer.py
efraimdahl: Done. Fixed in [2b69c6a](https://github.com/huggingface/peft/pull/2643/commits/2b69c6ac7eed9e5808c6248f145559cf69af9977)

src/peft/tuners/adaption_prompt/model.py
efraimdahl: You are correct, I missread the suggestion. Fixed in [2b69c6a](https://github.com/huggingface/peft/pull/2643/commits/2b69c6ac7eed9e5808c6248f145559cf69af9977)

BenjaminBossan:
@efraimdahl Please always let me know when the PR is ready for another review, otherwise I'll assume you're still working on further changes.

efraimdahl:
> @efraimdahl Please always let me know when the PR is ready for another review, otherwise I'll assume you're still working on further changes.

Yes!  I think its ready for another review. Thank you for your patience @BenjaminBossan . I attempted to re-request a review through the interface.

tests/test_adaption_prompt.py
```diff
@@ -29,6 +29,7 @@
 
 
 MODELS_TO_TEST = [
+    "trl-internal-testing/tiny-GPT2LMHeadModel",
```
BenjaminBossan: Can we use `"hf-internal-testing/tiny-random-gpt2"` too?

src/peft/tuners/adaption_prompt/utils.py
```diff
@@ -127,6 +128,31 @@ def llama_compute_query_states(model: nn.Module, **kwargs) -> torch.Tensor:
...
+) -> torch.Tensor:
+    """
+    Compute query states for GPT2 models. They need to be recomputed as the forward() method of the GPT@ in the
+    transformers library does not return them. See the related discussion in the PR:
+    """
+    if encoder_hidden_states is not None:
+        if not hasattr(model, "q_attn"):
+            raise ValueError(
+                "If class is used as cross attention, the weights `q_attn` have to be defined. "
+                "Please make sure to instantiate class with `GPT2Attention(..., is_cross_attention=True)`."
```
BenjaminBossan: Not so happy with the wording "class" here, how about using an f-string and replacing class with `model.__class__.__name__`?

src/peft/tuners/adaption_prompt/layer.py
```diff
@@ -34,7 +150,8 @@ def __init__(self, model_type: str, adapter_len: int, model):
             adapter_len: The length of the adaption prompt to insert.
             model: The original transformer attention module that is being wrapped.
         """
-        assert not isinstance(model, AdaptedAttention)
+        if isinstance(model, AdaptedAttention):
+            raise ValueError("Unable to stack multiple adaption prompts")
```
BenjaminBossan: Thanks for adding a proper error.

src/peft/tuners/adaption_prompt/layer.py
```diff
@@ -21,6 +22,121 @@
 from .config import TRANSFORMERS_MODEL_CONFIG
 
 
+class AdaptedAttentionGPT(nn.Module):
+    """This module wraps a GPT2Attention module and injects adaption prompts"""
+
+    def __init__(self, model_type: str, adapter_len: int, model):
```
BenjaminBossan: If I'm not missing something, the `__init__` is the same for both `AdaptedAttention` and `AdaptedAttentionGPT`, only difference being the `isinstance` check at the start. I wonder: Could we create a base class, say, `_BaseAdaptedAttention`, move the `__init__` there, and inherit from that class for both? The type check can be against `_BaseAdaptedAttention`. WDYT?

efraimdahl: You are right, the code is largely identical, accept for one small section that i moved to the respective init functions, but that's a lot more concise.

tests/test_adaption_prompt.py
efraimdahl: For sure

efraimdahl:
Following your suggestion @BenjaminBossan I implemented a base class, changed the gpt2-test model and improved the cross attention error message in [8156d6b](https://github.com/huggingface/peft/pull/2643/commits/8156d6b9b847b3514879d4c49c5aedda7535870a).

BenjaminBossan:
Thanks @efraimdahl, could you please run `make style`?

efraimdahl:
> Thanks @efraimdahl, could you please run `make style`?

Done @BenjaminBossan, thanks again!

BenjaminBossan (approved):
Thanks for adding GPT2 support, the PR LGTM.

Failing MacOS tests are unrelated.

Issue #2642, opened 2025-07-11
Prompt_Tuning.ipynb example doesn't seem to train the model


ruixing76:
I am running Prompt-Tuning notebook example from PEFT lib examples [here](https://github.com/huggingface/peft/blob/main/examples/sequence_classification/Prompt_Tuning.ipynb).  I did **not** change any line of code and I ran the code block sequentially.

However, the performance under metrics remain exactly the **same** for each epoch, which is very weird. From the [orignal notebook](https://github.com/huggingface/peft/blob/main/examples/sequence_classification/Prompt_Tuning.ipynb), we can see accuracy fluctuates and can increase to 0.70. 

I checked the output logits for the training data is changing every epoch (set shuffle=False, and this is the only change for debugging). Now I am very confused, any suggestions would be very much welcome, please let me know if I am doing something very wrong, thanks in advance!

Here's the performance log:
```
100%|██████████| 115/115 [00:20<00:00,  5.74it/s]
100%|██████████| 13/13 [00:01<00:00, 10.36it/s]
epoch 0: {'accuracy': 0.6838235294117647, 'f1': 0.8122270742358079}
100%|██████████| 115/115 [00:20<00:00,  5.72it/s]
100%|██████████| 13/13 [00:01<00:00, 10.49it/s]
epoch 1: {'accuracy': 0.6838235294117647, 'f1': 0.8122270742358079}
100%|██████████| 115/115 [00:20<00:00,  5.74it/s]
100%|██████████| 13/13 [00:01<00:00, 10.34it/s]
epoch 2: {'accuracy': 0.6838235294117647, 'f1': 0.8122270742358079}
100%|██████████| 115/115 [00:20<00:00,  5.72it/s]
100%|██████████| 13/13 [00:01<00:00, 10.35it/s]
epoch 3: {'accuracy': 0.6838235294117647, 'f1': 0.8122270742358079}
100%|██████████| 115/115 [00:20<00:00,  5.74it/s]
100%|██████████| 13/13 [00:01<00:00, 10.47it/s]
epoch 4: {'accuracy': 0.6838235294117647, 'f1': 0.8122270742358079}
100%|██████████| 115/115 [00:20<00:00,  5.69it/s]
100%|██████████| 13/13 [00:01<00:00, 10.63it/s]
epoch 5: {'accuracy': 0.6838235294117647, 'f1': 0.8122270742358079}
100%|██████████| 115/115 [00:20<00:00,  5.75it/s]
[...]

PR #2220, opened 2024-11-18
FIX: Avoid needless copy from modules_to_save

BenjaminBossan:
Resolves #2206

~~NOT READY TO MERGE YET. Let's wait for the 0.14.0 release.~~

~~Tentative solution to that issue.~~

The problem is that we keep a "global" `modules_to_save` on the `model` which contains all possible `modules_to_save` for each adapter. When the first adapter targets layer "foo" with `modules_to_save` and the second adapter targets "bar", then "foo" will create a copy of the original module for the second adapter, even though it's not needed.

This does not change the result but is unnecessary and takes up memory. Thus it should be avoided.

PR #2646, opened 2025-07-14
FIX: Prompt learning methods modules_to_save issue

BenjaminBossan:
This PR relates to #2642 but is probably not the cause of the reported issue there.

When using prompt learning methods, `modules_to_save` is not correctly set automatically. This is really bad when using, for instance, sequence classification tasks, which require the classifier layer to be added to `modules_to_save`.

The issue was introduced in #2220 where it is wrongly assumed that the PEFT config always has a `modules_to_save` attribute, which is not true for prompt learning. In #2481, this was partly fixed by using `getattr` to avoid an error. However, this did not resolve the fundamental issue that for prompt learning, there is no such attribute, resulting in `module_to_save` not being applied.

This PR proposes to fix this by adding `modules_to_save` to the prompt learning configs.


githubnemo (approved):
Thanks for fixing, the change LGTM.

PR #2649, opened 2025-07-15
Fix not detecting regex-targeted embedding layer

githubnemo:
This issue was found in PR #2638 and is defined thusly:

> When calling `get_peft_model_state_dict(..., save_embedding_layers="auto")` we check if the
> embedding layer is targetted to determine if the embedding layers need saving. This is not
> done when `PeftConfig.target_modules` is a regex-string, potentially missing to save embeddings.

This is fixed by adding a check similar to the existing query of whether `EMBEDDING_LAYER_NAMES` is a subset of the defined target modules, only that the regex matching from `BaseTuner.inject_adapter` is used. To avoid code duplication, the matching was moved to its own utility function `match_target_against_key`.

The main complication was to define the test-cases as it was non-trivial to find what the meaning of `save_embedding_layers="auto"` entails. I've assembled a list of cases that I think are correct in the corresponding unit test.

### DIFF
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -28,6 +28,7 @@ from transformers.utils import http_user_agent
 
 from peft.mapping import PEFT_TYPE_TO_PREFIX_MAPPING
 
+from .constants import INCLUDE_LINEAR_LAYERS_SHORTHAND
 from .other import (
     EMBEDDING_LAYER_NAMES,
     SAFETENSORS_WEIGHTS_NAME,
@@ -35,6 +36,7 @@ from .other import (
     AuxiliaryTrainingWrapper,
     check_file_exists_on_hf_hub,
     infer_device,
+    match_target_against_key,
 )
 from .peft_types import PeftType
 
@@ -232,15 +234,24 @@ def get_peft_model_state_dict(
 
     # DEAL WITH EMBEDDINGS
     # check the common embedding layers in `target_modules` to reset `save_embedding_layers` if necessary
-    is_embedding_in_target_modules = False
-    if (
-        save_embedding_layers == "auto"
-        and hasattr(config, "target_modules")
-        and any(k in config.target_modules for k in EMBEDDING_LAYER_NAMES)
-        and config.peft_type != PeftType.TRAINABLE_TOKENS
-    ):
+    embedding_is_targeted = False
+    if hasattr(config, "target_modules"):
+        if isinstance(config.target_modules, str) and config.target_modules != INCLUDE_LINEAR_LAYERS_SHORTHAND:
+            embedding_is_targeted = any(
+                match_target_against_key(config.target_modules, k)
+                for k, _ in model.named_modules()
+                if any(e in k for e in EMBEDDING_LAYER_NAMES)
+            )
+        elif config.target_modules:
+            embedding_is_targeted = any(k in config.target_modules for k in EMBEDDING_LAYER_NAMES)
+
+    using_trainable_tokens = (
+        config.peft_type == PeftType.TRAINABLE_TOKENS or getattr(config, "trainable_token_indices", None) is not None
+    )
+
+    if save_embedding_layers == "auto" and embedding_is_targeted and not using_trainable_tokens:
         warnings.warn("Setting `save_embedding_layers` to `True` as embedding layers found in `target_modules`.")
-        save_embedding_layers = is_embedding_in_target_modules = True
+        save_embedding_layers = True
     elif save_embedding_layers == "auto":
         vocab_size = getattr(getattr(model, "config", None), "vocab_size", None)
         model_id = getattr(config, "base_model_name_or_path", None)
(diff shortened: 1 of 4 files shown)


githubnemo:
> I wonder about the trainable tokens case. If the vocab size has increased, I think we need to save the embeddings, even if trainable tokens are used. If the embedding matrix has new entries, those would otherwise not be loaded correctly. I'm not sure what valid use case there could be for that, so it might be more of a theoretical situation, but still.

I understand the problematic scenario as:
```python
base_model.resize_token_embeddings(n)
peft_model = get_peft_model(base_model, LoraConfig(..., trainable_token_indices=[n-2, n-1]))
```
We would expect the user to call `base_model.resize_token_embeddings(n)` at inference time as well, if they don't there will be an size mismatch error. If the user employs the `AutoPeft*` class this should work out of the box since that class is resizing embeddings, if they are smaller than the tokenizer vocab size. So, the only error case would be that the user first resizes the embedding, overwrites certain contents of the embedding with trainable tokens and then, when loading using a specific PEFT model class, misses the resizing. I'd argue that this error wouldn't be too surprising to that one specific user. But when the user doing inference is not the same user who is doing the training, it could become problematic because they might not know that the embeddings were resized. Therefore, by always saving the embedding layer when the embedding size was changed (regardless of trainable tokens used or not) we would limit the loading error to only that adapter group that has explicitly set `save_embedding_layers=False`. I'd be OK with that but the real problem is that we could be more helpful in explaining the resize error.

So in my ideal world we'd:
1. keep the save logic as is
2. catch embedding loading errors and give the user a helpful message that they should try to do `model.resize_token_embeddings(len(tokenizer))` or look up the model card

But (2) is probably full of caveats. If you think that (2) is doable then I'll give it a try. But if you also think that it is going to be messy and spotty, I'm fine with always saving the embedding matrix when it was resized (even when trainable tokens was used).

> Regarding the tests, I wonder if we should leave this test as it is, since this functionality is not really about the low level API. Instead, maybe we can change/extend these tests:

They look like good candidates for this, will do.

BenjaminBossan:
> So, the only error case would be that the user first resizes the embedding, overwrites certain contents of the embedding with trainable tokens and then, when loading using a specific PEFT model class, misses the resizing.

I was thinking of a different use case: The user resizes the embeddings to add some new tokens. Those are trained in one part of the training pipeline. In addition, the user also employs trainable tokens to modify some of the existing tokens (not necessarily the new ones) and trains them in a different part of the training pipeline. In that case, we would need to store the embeddings, otherwise the user would lose the training from the first part. It is admittedly unlikely to encounter this, but I'd rather save too much than too little.

tests/test_decoder_models.py
```diff
@@ -680,3 +681,36 @@ def process(samples):
...
+    @pytest.mark.parametrize("save_embedding_layers", ["auto", True, False])
+    @pytest.mark.parametrize(
+        "peft_config",
+        [
+            (LoraConfig(target_modules=["lin0", "embed_tokens"], init_lora_weights=False)),
+            (LoraConfig(target_modules=r".*\.embed_tokens", init_lora_weights=False)),
+        ],
+    )
+    def test_save_pretrained_targeting_lora_to_embedding_layer(self, save_embedding_layers, tmp_path, peft_config):
+        model_id = "trl-internal-testing/tiny-random-LlamaForCausalLM"
```
BenjaminBossan, line 694: Should `hub_online_once` be used here?

tests/test_custom_models.py
```diff
@@ -2392,52 +2392,62 @@ def test_non_existing_model_card(self):
         assert len(model_card) > 1000
 
     @pytest.mark.parametrize("save_embedding_layers", ["auto", True, False])
-    def test_targeting_lora_to_embedding_layer(self, save_embedding_layers):
+    @pytest.mark.parametrize(
```
BenjaminBossan, line 2395: Thanks for also cleaning up these tests.

src/peft/utils/save_and_load.py
```diff
@@ -231,23 +233,35 @@ def renamed_dora_weights(k):
...
     embedding_is_targeted = False
     if hasattr(config, "target_modules"):
-        if isinstance(config.target_modules, str):
-            # TODO: implement this; note: this change is not directly related to the PR, the bug already existed b4
-            pass
+        if isinstance(config.target_modules, str) and config.target_modules != INCLUDE_LINEAR_LAYERS_SHORTHAND:
+            embedding_is_targeted = any(
+                match_target_against_key(config.target_modules, k)
+                for k, _ in model.get_base_model().named_modules()
+                if any(e in k for e in EMBEDDING_LAYER_NAMES)
```
BenjaminBossan: Just wondering if `e in k` might be a bit too vague as a match, perhaps `e + "." in k`?

BenjaminBossan (changes requested):
Thanks for the updates. Only two minuscule comments, otherwise the PR is good.

src/peft/utils/save_and_load.py
githubnemo: That's also a bit tricky since we iterate over modules and not weights, therefore there will be no `embed_tokens.weight` in the list, just `embed_tokens`. The proper way would probably something like `(.*\.)?{e}$`, similar to `get_pattern_key`. I've implemented it as such.

PR #2661, opened 2025-07-21
[WIP] ENH Cache DoRA weight norm for inference

### LINKED ISSUE #2651: DoRA slow forward inference
### System Info

Please update the forward of DoRA so that during evaluation it doesn't wastefully recompute weight normalization during each forward pass when it doesn't need to. Hopefully switching between the current implementation and an implementation that precomputes the weight norm is as simple as checking `model.eval() # not model.training`.

I understand one could merge the weights but it tends to be more lossy than simple unmerged adapters. Additionally, this should speed up evals during finetuning larger models.

### Who can help?

@BenjaminBossan 

### Reproduction

https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/dora.py#L66-L101

### Expected behavior

Inference with unmerged adapters (`model.training==False`) does not compute weight norm during each forward pass.

BenjaminBossan:
Resolves #2651

This is an experimental branch that caches the weight norm from DoRA for faster inference. Since, during inference, the weights don't change, there is no need to recalculate the weight norm of a DoRA module each time.

During training, recalculation is needed, thus there is no caching when the module has `training=True`.

The cache does not prevent each and every possible duplicate calculation. For instance, the weight norm is calculated during module initialization and then again during the first forward pass when performing inference. Only starting from the second forward pass on will the weight norms be cached.

The reason why this is just a draft PR is that before finishing it, I want to ensure that the addition of cache is worth it. Caches can often be a tricky business and lead to subtle bugs. Just as an example here, we detect if users put the model into training mode when calling `model.train()` and clear the cache. However, we would not detect it if the user directly sets `module.training=True`.

Some preliminary testing with a small model, meta-llama/Llama-3.2-1B, and some dummy data didn't show huge improvements, with an average of 10 runs showing:

- with caching: 1.3555 sec
- w/o caching: 1.4071 sec

Thus, before continuing, I'd like to see some more real world measurements. If the change is considered to be worth it, tests should be added to the PR before it's ready.

### DIFF
--- a/src/peft/tuners/lora/dora.py
+++ b/src/peft/tuners/lora/dora.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 from copy import deepcopy
+from functools import wraps
+from typing import Any, Optional
 
 import torch
 import torch.nn.functional as F
@@ -22,11 +24,50 @@ from peft.utils.integrations import dequantize_module_weight, gather_params_ctx
 from peft.utils.other import transpose
 
 
+def cache_weight_norm(func):
+    # TODO
+    cache_key = "weight_norm"
+
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        if self.training:
+            return func(self, *args, **kwargs)
+
+        output = self._cache_get(cache_key, None)
+        if output is not None:
+            return output
+
+        output = func(self, *args, **kwargs)
+        self._cache_store(cache_key, output)
+        return output
+
+    return wrapper
+
+
 class DoraLinearLayer(nn.Module):
     def __init__(self, fan_in_fan_out):
         super().__init__()
         self.fan_in_fan_out = fan_in_fan_out
+        self._caches: dict[str, Any] = {}  # small ad hoc cache; values are not part of the state_dict
 
+    def _cache_store(self, key: str, value: Any) -> None:
+        # cache intermediate values, e.g. weight norm of DoRA
+        self._caches[key] = value
+
+    def _cache_get(self, key: str, default: Optional[Any]) -> Optional[Any]:
+        # retrieve from ad hoc cache
+        return self._caches.get(key, default)
+
+    def _cache_clear(self) -> None:
+        self._caches.clear()
+
+    def train(self, mode: bool = True):
+        if mode:
+            self._cache_clear()
+        super().train(mode=mode)
+        return self
+
+    @cache_weight_norm
     def get_weight_norm(self, weight, lora_weight, scaling) -> torch.Tensor:
         # calculate L2 norm of weight matrix, column-wise
         weight = transpose(weight, self.fan_in_fan_out)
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
@phemw The PR is ready from my side, if you want to give this a try, LMK what you find. Note that the memory overhead of caching is quite significant (41% in one test), so it's turned off by default and users need to opt-in.

```python
from peft.helpers import DoraCaching

model.eval()
with DoraCaching():
    output = model(inputs)
```

src/peft/helpers.py
```diff
@@ -249,3 +249,44 @@ def disable_input_dtype_casting(model: nn.Module, active: bool = True):
                 continue
             if name in original_values:
                 module.cast_input_dtype_enabled = original_values[name]
+
+
+class DoraCaching:
+    """Context manager to enable DoRA caching, which improves speed of DoRA inference at the expense of memory.
+
+    Even within the caching context, if the model is in training mode, caching is disabled. When the model switches to
+    training mode, the cache will be cleared.
```
githubnemo, line 261: I think this would benefit from a description of what is cached (assuming the reader knows what DoRA is doing). Something simple would suffice, e.g. "When activated, caches matrix norms of base and adapter weights as well as computed adapter weights.".

src/peft/tuners/lora/dora.py
```diff
@@ -22,18 +24,80 @@
...
+    @cache_decorator("weight-norm")
+    def get_weight_norm(self, weight, lora_weight, scaling, adapter_name: Optional[str] = None) -> torch.Tensor:
         # calculate L2 norm of weight matrix, column-wise
         weight = transpose(weight, self.fan_in_fan_out)
         weight = weight + scaling * lora_weight
         weight_norm = torch.linalg.norm(weight, dim=1).to(weight.dtype)
         return weight_norm
 
+    @cache_decorator("lora-weight")
+    def get_lora_weight(self, lora_A, lora_B, adapter_name: Optional[str] = None):
```
githubnemo: I'm not too excited about the 'magic' `adapter_name` parameter but I don't have a better way and going explicit is a lot more code and less readability, so I'm OK with this.

src/peft/tuners/lora/dora.py
```diff
@@ -132,25 +203,39 @@ def __repr__(self) -> str:
...
         return weight_norm
 
-    def forward(self, x, *, lora_A, lora_B, scaling, base_layer, base_result=None):
+    @cache_decorator("lora-weight")
+    def get_lora_weight(self, lora_A, lora_B, adapter_name: Optional[str] = None) -> torch.Tensor:
+        # Don't use `lora_weight = lora_B.weight @ lora_A.weight` because this causes errors with FSDP. Instead,
+        # calculate the same but using forward.
+        r = lora_A.weight.shape[0]
+        lora_weight = torch.mm(lora_B.weight.view([-1, r]), lora_A.weight.view([r, -1]))
+        lora_weight = lora_weight
```
githubnemo:
```suggestion
```

BenjaminBossan:
@githubnemo Comments should be addressed, please check again.

PR #2662, opened 2025-07-22
FIX Prefix tuning after transformers PR 38635

BenjaminBossan:
Due to https://github.com/huggingface/transformers/pull/38635, [several tests involving prefix tuning broke](https://github.com/huggingface/peft/actions/runs/16417140904/job/46385751329).

This PR fixes this by resolving two issues:

1. The `_supports_cache_class` attribute was removed, we can now assume that it is True if the attribute does not exist.

2. We had special handling of `past_key_values` for `GPTBigCodeForCausalLM` which is no longer required (nor valid) after that PR, so it is removed depending on the transformers version.


zucchini-nlp (approved):
Nicee

PR #2665, opened 2025-07-23
ENH: Targeting multiple parameters on the same module

BenjaminBossan:
When the `target_parameters` feature for LoRA was introduced in #2638, there was one gap, namely the possibility to target multiple `nn.Parameter`s on the same module (there was only a workaround involving multiple adapters, but that is not user friendly). With this PR, it is now possible to achieve this.

The mechanism to enable this is a bit crude, namely allowing to nest multiple `ParamWrapper`s. This should generally be fine as long as there are only a couple of `nn.Parameter`s being targeted on the same module. When there are dozens or hundreds, this approach could load to slow downs or other issues.

A side effect of this implementation is that the `ParamWrapper`, when it removes the parametrization, now only removes its own parametrization. When using `nn.utils.parametrize.remove_parametrization`, it removes all parametrizations, which is bad when we have nested parametrizations.

## Alternative approaches

Some alternative approaches were discussed internally but the chosen one was considered most practical.

1. Allow to have more than one adapted parameter per LoRA layer. This would require to have nested dicts for the LoRA parameters, something like `self.lora_A[adapter_name][parameter_name]`. We don't have this anywhere so far and it would probably break implicit assumptions about PEFT layers in many places (like, parsing of `state_dict` keys), requiring many adjustments.
2. Have an auxiliary module that contains the individual LoRA layers that target the individual parameters. This could be the cleanest solution and would probably be more efficient if there are a huge number of targeted parameters per module. However, this also brings extra complexity, as it requires implementing the logic of how to route the information to the right parameter, and it may be a solution to a problem that is irrelevant in practice (large number of targets per module).

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -1777,6 +1777,15 @@ class _LoraParameterProxy(nn.Module):
             return W + self.delta_weight
 
 
+# copied from:
+# https://github.com/pytorch/pytorch/blob/5e386eec9426f174eea130c0c012d9f65ebe65fb/torch/nn/utils/parametrize.py#L75-L79
+def _register_parameter_or_buffer(module, name, X):
+    if isinstance(X, nn.Parameter):
+        module.register_parameter(name, X)
+    else:
+        module.register_buffer(name, X)
+
+
 class ParamWrapper(nn.Module, LoraLayer):
     """A LoRA wrapper for `nn.Parameter`. This layer is dispatched if users target a parameter directly with
     `lora_config.target_parameters`
@@ -1807,8 +1816,8 @@ class ParamWrapper(nn.Module, LoraLayer):
     ) -> None:
         super().__init__()
         LoraLayer.__init__(self, base_layer, **kwargs)
-        param = getattr(base_layer, parameter_name)
         self.parameter_name = parameter_name
+        param = self.get_param()
         if param.ndim == 3:
             self.num_experts, self.in_features, self.out_features = param.shape
         else:
@@ -1867,15 +1876,6 @@ class ParamWrapper(nn.Module, LoraLayer):
         # This code works for linear layers, override for other layer types
         if r <= 0:
             raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
-        if adapter_name in self.lora_A:
-            # It is not allowed to target multiple parameters on the same module. Supporting this would complicate
-            # things quite a lot, since we would require multiple self.lora_A, self.lora_B, etc., one for each targeted
-            # parameter.
-            raise ValueError(
-                f"lora.{self.__class__.__name__} already has an adapter for parameter '{self.parameter_name}'. "
-                "It is currently not possible to apply the same adapter to multiple parameters, please add a "
-                "different adapter to target another parameter of the same module."
-            )
 
         lora_variant = self.resolve_lora_variant(
             use_dora=use_dora, use_qalora=use_qalora, qalora_group_size=qalora_group_size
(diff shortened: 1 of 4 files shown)


src/peft/tuners/lora/layer.py
```diff
@@ -1958,7 +1958,11 @@ def _move_adapter_to_device_of_base_layer(self, adapter_name: str, device: Optio
                 adapter_layer[adapter_name] = adapter_layer[adapter_name].to(device)
 
     def get_param(self):
-        return getattr(self.base_layer, self.parameter_name)
+        if isinstance(self.base_layer, ParamWrapper):
+            param = getattr(self.get_base_layer(), self.parameter_name)
+        else:
+            param = getattr(self.base_layer, self.parameter_name)
+        return param
```
githubnemo: Why the distinction, why not always use `self.get_base_layer()`? Let's also document the reason

src/peft/tuners/lora/layer.py
```diff
@@ -2004,10 +2008,34 @@ def _activate_lora(self, active_adapters: list[str]):
...
+            # see https://github.com/pytorch/pytorch/blob/7d296d5c19750cecd82e2b95f6fb0f8dd918282e/torch/nn/utils/parametrize.py#L731-L737
+
+            original = base_layer.parametrizations[parameter_name].original
+            # Delete the property that manages the parametrization
+            delattr(base_layer.__class__, parameter_name)
+            # Delete the ParametrizationList
+            del base_layer.parametrizations[parameter_name]
+            # Restore the parameter / buffer into the main class
+            _register_parameter_or_buffer(base_layer, parameter_name, original)
+
```
githubnemo: Maybe I'm misunderstanding the comment but AFAICS [`remove_parametrizations`](https://github.com/pytorch/pytorch/blob/v2.7.0/torch/nn/utils/parametrize.py#L652) only removes the parametrization for `tensor_name`, in this case `self.parameter_name`? Has this changed between pytorch versions?

tests/test_target_parameters.py
```diff
@@ -314,38 +330,75 @@ def test_targeting_module_and_targeting_param_equivalent(self):
...
-            assert not torch.allclose(out_lora_0, out_lora_1, atol=atol, rtol=rtol)
+            # CHECK FORWARD CALLS
+
+            # log the weights seen during the forward call
+            weights = []
+
+            def mock_forward(self, W):
+                weights.append(W)
+                with torch.nn.utils.parametrize.cached():
+                    return W + self.delta_weight
```
githubnemo: this leaks the implementation, wouldn't it be better to call the original implementation from the mock like a decorator?

githubnemo (commented):
Surprisingly small change :) 
Looks good, I appreciate the thorough description.
Some remarks below

src/peft/tuners/lora/layer.py
matthewdouglas: @githubnemo That sounds right. To expand on this:`len(base_layer.parametrizations) == 1` tells us that only one parameter has a parametrization, and based on the check prior to it, we know that it is `parameter_name`. However it doesn't tell us how many parametrizations are applied to that parameter; i.e. we just know we have only one `ParametrizationList` which may have length >= 1.

Since we're interested in just the parametrizations on `self.parameter_name` we should instead check `len(base_layer.parametrizations[parameter_name]) == 1`. If it is, assume the `remove_parametrizations` API can be safely used, even if we've got more than one target parameter on the module.

Otherwise, we shouldn't be deleting the entire ParametrizationList; just the specific `_LoraParameterProxy`.

src/peft/tuners/lora/layer.py
BenjaminBossan: It was just defensive programming, but yeah, I don't think always calling `self.get_base_layer()` has any practical disadvantage.

src/peft/tuners/lora/layer.py
BenjaminBossan: Yes, you are both right, this doesn't really make sense as is. For now, I "reverted" to `nn.utils.parametrize.remove_parametrizations`, which is enough for this PR, and added a `# TODO` for your case, Matthew. Let's fix that in a separate PR.

tests/test_target_parameters.py
BenjaminBossan: It is better, but I was lazy :-p

Fixed now.

PR #2669, opened 2025-07-24
Update extending vocab docs

githubnemo:
- Recommends trainable tokens as first measure
- Clarifies a few things about saving embeddings
- Adds full-finetuning as an option of last resort

### DIFF
--- a/docs/source/developer_guides/troubleshooting.md
+++ b/docs/source/developer_guides/troubleshooting.md
@@ -145,9 +145,37 @@ As an example, when loading a model that is using the DeBERTa architecture for s
 
 ### Extending the vocabulary
 
-For many language fine-tuning tasks, extending the model's vocabulary is necessary since new tokens are being introduced. This requires extending the embedding layer to account for the new tokens and also storing the embedding layer in addition to the adapter weights when saving the adapter.
+For many language fine-tuning tasks, extending the model's vocabulary is necessary since new tokens are being introduced. This requires extending the embedding layer to account for the new tokens and, depending on the fine-tuning method, also storing the embedding layer in addition to the adapter weights when saving the adapter. There are a few ways of achieving this ordered by parameter effectiveness:
 
-Save the embedding layer by adding it to the `target_modules` of the config. The embedding layer name must follow the standard naming scheme from Transformers. For example, the Mistral config could look like this:
+- [trainable tokens](../package_reference/trainable_tokens), train only the specified tokens, optionally store only the updated values
+- training an adapter on the embedding matrix, optionally store only the updated values
+- full-finetuning of the embedding layer
+
+#### Using trainable tokens
+
+Let's start with trainable tokens, in this case it's [LoRA integration](../lora#efficiently-train-tokens-alongside-lora).  If you're interested in only training the embeddings and nothing else, refer to the [standalone documentation](../package_reference/trainable_tokens).
+
+To enable selective token training of the embedding layer, you'll need to supply the token ids of your newly added tokens via the `trainable_token_indices` parameter.  Optionally you can specify which layer to target if there is more than one embedding layer. For a Mistral model this could look like this:
+
+```python
+new_tokens = ['<think>', '</think>']
+tokenizer.add_tokens(new_tokens)
+base_model.resize_token_embeddings(len(tokenizer))
+
+lora_config = LoraConfig(
+    ...,
+    trainable_token_indices={'embed_tokens': tokenizer.convert_tokens_to_ids(new_tokens)},
+)
+```
+
+If your model uses tied weights (such as the `lm_head`), trainable tokens will try to resolve those and keep them updated as well, so in that case there should be no need for adding `modules_to_save=["lm_head"]`. This only works if the model uses the Transformers convention for tying weights.
+
+Saving the model with `model.save_pretrained` may save the full embedding matrix instead of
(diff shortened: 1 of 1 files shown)


docs/source/developer_guides/troubleshooting.md
```diff
@@ -145,17 +145,45 @@ As an example, when loading a model that is using the DeBERTa architecture for s
...
+For many language fine-tuning tasks, extending the model's vocabulary is necessary since new tokens are being introduced. This requires extending the embedding layer to account for the new tokens and, depending on the fine-tuning method, also storing the embedding layer in addition to the adapter weights when saving the adapter. There are a few ways of achieving this ordered by parameter effectiveness:
 
-Save the embedding layer by adding it to the `target_modules` of the config. The embedding layer name must follow the standard naming scheme from Transformers. For example, the Mistral config could look like this:
+- [trainable tokens](../package_reference/trainable_tokens), train only the specified tokens, optionally store only the updated values
+- training an adapter on the embedding matrix, optionally store only the updated values
+- full-finetuning of the embedding layer
+
+#### Using trainable tokens
+
+Let's start with trainable tokens, in this case it's [LoRA integration](../lora#efficiently-train-tokens-alongside-lora).  If you're interested in only training the embeddings and nothing else, refer to the [standalone documentation](../package_reference/trainable_tokens).
```
BenjaminBossan:
```suggestion
Let's start with trainable tokens, in this case its [LoRA integration](../developer_guides/lora#efficiently-train-tokens-alongside-lora).  If you're interested in only training the new embeddings and nothing else, refer to the [standalone documentation](../package_reference/trainable_tokens).
```

docs/source/developer_guides/troubleshooting.md
```diff
@@ -145,17 +145,45 @@ As an example, when loading a model that is using the DeBERTa architecture for s
...
+
+To enable selective token training of the embedding layer, you'll need to supply the token ids of your newly added tokens via the `trainable_token_indices` parameter.  Optionally you can specify which layer to target if there is more than one embedding layer. For a Mistral model this could look like this:
```
BenjaminBossan, line 158: Regarding this whole example here, I wonder if we can just remove it because it is pretty much covered in the provided links. Here, I'd rather focus on explaining when to use which approach.

docs/source/developer_guides/troubleshooting.md
```diff
@@ -145,17 +145,45 @@ As an example, when loading a model that is using the DeBERTa architecture for s
...
+lora_config = LoraConfig(
+    ...,
+    trainable_token_indices={'embed_tokens': tokenizer.convert_tokens_to_ids(new_tokens)},
+)
+```
+
+If your model uses tied weights (such as the `lm_head`), trainable tokens will try to resolve those and keep them updated as well, so in that case there should be no need for adding `modules_to_save=["lm_head"]`. This only works if the model uses the Transformers convention for tying weights.
+
+Saving the model with `model.save_pretrained` may save the full embedding matrix instead of
+only the difference as a precaution because the embedding matrix was resized. To save space you can disable this behavior by setting `save_embedding_layers=False` when calling `save_pretrained`. This is safe to do as long as you don't modify the embedding matrix through other means as well, as such changes will be not tracked by trainable tokens.
```
BenjaminBossan, line 174: I wonder: Should we add this recommendation to the docs of trainable tokens too?

docs/source/developer_guides/troubleshooting.md
```diff
@@ -167,6 +195,14 @@ For inference, load the base model first and resize it the same way you did befo
 
 For a complete example, please check out [this notebook](https://github.com/huggingface/peft/blob/main/examples/causal_language_modeling/peft_lora_clm_with_additional_tokens.ipynb).
 
+#### Full fine-tuning
+
+Full fine-tuning doesn't save VRAM or storage space but if all fails you can fall back to this and see if it works for you. Achieve it by adding the name of the embedding layer to `modules_to_save`. Note that you need to add tied layers as well, e.g. `lm_head`. Example for a Mistral model with LoRA:
```
BenjaminBossan:
```suggestion
Full fine-tuning is more costly in terms of VRAM or storage space but if all else fails, you can fall back to this and see if it works for you. Achieve it by adding the name of the embedding layer to `modules_to_save`. Note that you need to add tied layers as well, e.g. `lm_head`. Example for a Mistral model with LoRA:
```

I found that "Full fine-tuning doesn't save VRAM or storage space" could be confusing, as it could be interpreted to mean that all the benefits of, say, LoRA are lost when fully finetuning the embedding.

BenjaminBossan (changes requested):
Thanks for updating this part of the docs and clarifying better what options the users have. I made a few small suggestions, please check.

docs/source/developer_guides/troubleshooting.md
githubnemo: As a user I'd appreciate an example if I can't get an image, even though I could click on the link and get a similar example. It runs the risk of being outdated eventually but its grounding nature is helping I think.

BenjaminBossan: Okay, I guess you're right, each subsection has an example so it fits.

docs/source/developer_guides/troubleshooting.md
githubnemo: Good idea, put it there as a TIP and extended it to `get_peft_model_state_dict`.

BenjaminBossan (approved):
Thanks for the updates, LGTM.

PR #2670, opened 2025-07-24
Method Comparison: Improve formatting/layout of table

githubnemo:
Quick improvement to reduce the dominance of columns like `{peft,train}_config` and make numbers a bit more readable through proper decimal/thousands formatting.


githubnemo:
Done. Please check again :)

PR #2675, opened 2025-07-28
FIX: Missing device map for facebook/opt-125m

BenjaminBossan:
Fixes the [failing EETQ test](https://github.com/huggingface/peft/actions/runs/16546442926/job/46794737509) in the nighly multi device CI.

In #2612, hard coded `device_map`s were added for multi-GPU training as we could not rely on `device_map="auto"`. While doing this change, one `device_map` was missing, namely for `facebook/opt-125m`, which is used in the EETQ multi device test. This `device_map` was now added. This makes the test pass locally.


BenjaminBossan:
Note: Failing tests are unrelated and fixed via #2676. This PR only affects GPU tests, so the CPU CI is irrelevant here.

PR #2677, opened 2025-07-28
FIX: Small fixes to target_parameters

BenjaminBossan:
1. Better error message when same layer targeted twice
2. Remove unused attribute `num_experts` from `_LoraParameterProxy`

No functional changes.

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -1767,10 +1767,9 @@ class _LoraParameterProxy(nn.Module):
     Intended to be used in conjunction with `nn.utils.parametrize`, see `ParamWrapper`.
     """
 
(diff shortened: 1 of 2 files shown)


githubnemo (commented):
LGTM, do we need to adjust the test in `test_initialization.py:1410`?

BenjaminBossan:
> LGTM, do we need to adjust the test in `test_initialization.py:1410`?

No, the message still matches. We could extend it in the test, but I think there is no advantage. For me, the main reasons to test the error message are:

1. Make sure that the right error is caught, not another one that happens to be the same type. For this, we don't need to match the whole message.
2. If there is string interpolation, ensure it works as intended. The newly added part of the error message doesn't have string interpolation.

githubnemo (approved):

Issue #2672, opened 2025-07-25
Support for RoAD: 2D Rotary Adaptation for Efficient Finetuning, Efficient Batching and Composability


ppetrushkov:
### Feature request

[RoAD](https://arxiv.org/pdf/2409.00119) is a Parameter-efficient Fine-Tuning technique that is especially well-suited for efficient inference with mixed adapters in a batch, while still providing high output quality with very small parameter count.

RoAD learns 2D rotation matrices that are applied to the layer output which can be written using only element-wise multiplication (rather than matrix multiplication), enabling very fast inference with adapters in unmerged state. It is somewhat related to Orthogonal Finetuning (OFT) method.

Paper: https://arxiv.org/pdf/2409.00119
PEFT implementation: https://github.com/ppetrushkov/peft/tree/road (not from paper authors)

### Your contribution

The PEFT implementation should be ready and tested. I created this feature request following the contributing guidelines, if the maintainers are not against including this method, I can proceed with creating a pull request and add the missing documentation and examples.

I also have a fork of VLLM where I implemented and tested the efficiency of this method which shows significantly better performance than LoRA https://github.com/ppetrushkov/vllm/tree/v0.9.1-road

BenjaminBossan:
Thanks for your proposal to integrate RoAD (RoAd?) into PEFT. From a first quick reading, the method seems interesting and novel, and I like the focus on mixed batch inference.

Given that you're not one of the paper authors, and that the [authors promise to release their code](https://github.com/BaohaoLiao/road) at a later point in time, I would first try to contact them before proceeding with your fork. Would you be willing to contact them (privately, or better, publicly, by opening an issue on their repo) and ask what they think about it? Perhaps, they already plan a PEFT integration, in which case we should wait. Or their code is hard to integrate and they are happy to continue with your fork. Either way, we should contact them first and give them a bit of time to respond.

ppetrushkov:
Thanks @BenjaminBossan , I'm not the author, but I do know them and I already checked with them privately, they are ok with this implementation.

BenjaminBossan:
I see, in this case please go ahead. If you can rope them in to help reviewing the code, even better :)

BaohaoLiao:
> I see, in this case please go ahead. If you can rope them in to help reviewing the code, even better :)

[...]

PR #2678, opened 2025-07-28
Support for RoAD: 2D Rotary Adaptation

ppetrushkov:
Discussed in issue https://github.com/huggingface/peft/issues/2672

[RoAD](https://arxiv.org/pdf/2409.00119) is a Parameter-efficient Fine-Tuning technique that is especially well-suited for efficient inference with mixed adapters in a batch, while still providing high output quality with very small parameter count.

RoAD learns 2D rotation matrices that are applied to the layer output which can be written using only element-wise multiplication (rather than matrix multiplication), enabling very fast inference with adapters in unmerged state. It is somewhat related to Orthogonal Finetuning (OFT) method.

Paper: https://arxiv.org/pdf/2409.00119

I also have a fork of VLLM where I implemented and tested the efficiency of this method which shows significantly better performance than LoRA https://github.com/ppetrushkov/vllm/tree/v0.9.1-road


docs/source/package_reference/road.md
```diff
@@ -0,0 +1,29 @@
...
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# RoAd
+
+[RoAd](https://arxiv.org/pdf/2409.00119) is a parameter‑efficient fine‑tuning technique that adapts large language models by learning a small set of 2×2 rotation matrices (and optional scaling factors) applied to pairs of hidden dimensions, achieving competitive or superior performance with under 0.1% trainable parameters. Unlike LoRA’s batched low‑rank updates, RoAd’s sparse rotations reformulate to simple element‑wise operations, yielding significantly higher serving throughput when handling heterogeneous requests in the same batch. Moreover, RoAd integrates seamlessly into a distributed interchange intervention framework, enabling interpretable, composable task‑specific adaptations by combining orthogonal subspaces learned for different tasks.
```
BenjaminBossan: > superior performance

Superior to what?

> heterogeneous requests in the same batch
> distributed interchange intervention framework

I would elaborate on that, as it might not be clear what is meant by that to some users.

> significantly higher serving throughput when handling heterogeneous requests in the same batch

It would be great if you could add a toy example on how to achieve this. For reference, check this section for LoRA: https://huggingface.co/docs/peft/developer_guides/lora#inference-with-different-lora-adapters-in-the-same-batch

docs/source/package_reference/road.md
```diff
@@ -0,0 +1,29 @@
...
+
+Finetuning with RoAd typically requires higher learning rate compared to LoRA or similar methods, around 1e-3.
```
BenjaminBossan: Thanks for documenting that.

examples/road_finetuning/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+
+## Introduction
+[RoAd](https://arxiv.org/pdf/2409.00119) a novel method
+which employs a straightforward 2D rotation to adapt LLMs which is 
+remarkably parameter-efficient, delivering good
+performance with < 0.1% trainable parameters; efficient
+in serving requests requiring different adapters within a batch, with an overhead
+comparable to element-wise multiplication instead of batch matrix multiplication;
+enhances LLM’s interpretability.
```
BenjaminBossan: This paragraph could use a bit of rework to make it more readable.

examples/road_finetuning/road_finetuning.py
```diff
@@ -0,0 +1,187 @@
+import os
```
BenjaminBossan, line 15: Let's add the copyright notice.

method_comparison/MetaMathQA/experiments/road/llama-3.2-3B-variant2/adapter_config.json
BenjaminBossan, line 1: Let's rename the directory to `road/llama-3.2-3B-lr_0.001`

src/peft/tuners/road/config.py
```diff
@@ -0,0 +1,123 @@
...
+
+from dataclasses import dataclass, field
+from enum import Enum
+from typing import Optional, Union
+
+from peft.config import PeftConfig
+from peft.utils import PeftType
+
+
+class RoadVariant(str, Enum):
```
BenjaminBossan: I think it would be easier from the user's point of view to remove the `Enum` and just use a string literal: `variant: Literal["ROAD_1", "ROAD_2", "ROAD_4"]` (personally, I'd even lower case it). The values can be validated in `__post_init__`. Please also adjust the docs and type annotations accordingly.

src/peft/tuners/road/config.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    This is the configuration class to store the configuration of a [`RoadModel`]. Road adapter is proposed in
+    https://arxiv.org/pdf/2409.00119 .
+
+    Args:
+        variant (Union[`RoadVariant`, `str`]):
+            The variant of the Road model to use. It can be one of 1, 2, or 4.
+            - 1: Road-1
+            - 2: Road-2
+            - 4: Road-4
```
BenjaminBossan: There should be more information on this: Let's refer to Table 1 of the paper, explain how the number of trainable parameters scales with the choice, and when users should use which one.

src/peft/tuners/road/config.py
```diff
@@ -0,0 +1,123 @@
...
+    ROAD_1 = "1"
+    ROAD_2 = "2"
+    ROAD_4 = "4"
+
+
+@dataclass
+class RoadConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`RoadModel`]. Road adapter is proposed in
+    https://arxiv.org/pdf/2409.00119 .
```
BenjaminBossan:
```suggestion
    https://arxiv.org/pdf/2409.00119.
```

src/peft/tuners/road/config.py
```diff
@@ -0,0 +1,123 @@
...
+class RoadVariant(str, Enum):
+    ROAD_1 = "1"
+    ROAD_2 = "2"
+    ROAD_4 = "4"
+
+
+@dataclass
+class RoadConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`RoadModel`]. Road adapter is proposed in
```
BenjaminBossan:
```suggestion
    This is the configuration class to store the configuration of a [`RoadModel`]. RoAd adapter is proposed in
```

src/peft/tuners/road/config.py
```diff
@@ -0,0 +1,123 @@
...
+            - 4: Road-4
+        group_size (`int`):
+            Group size defines how elements are grouped together into 2D vectors for rotation.
+            Within each group element 0 is paired with element group_size/2,
+            then element 1 is paired with element group_size/2+1 and so on.
+            This has no effect on the model performance, since elements are unordered,
+            however it has some effect on inference speed when used in e.g. VLLM.
+            For best speed group size of at least 64 is recommended.
+            Note that model hidden size (or hidden size per partition when used with tensor parallelism)
+            must be divisible by group_size, so for very small models you might need to reduce this parameter.
```
BenjaminBossan, line 50: Is there a disadvantage to choosing a group size that is too large, as long as it's <= hidden size?

src/peft/tuners/road/config.py
```diff
@@ -0,0 +1,123 @@
...
+        init_weights (`bool`):
+            Whether to perform initialization of Road weights.
```
BenjaminBossan:
```suggestion
            Whether to perform initialization of RoAd weights.
```

src/peft/tuners/road/config.py
```diff
@@ -0,0 +1,123 @@
...
+            the model is a PreTrainedModel, the output layer excluded). If this is not specified, modules will be
+            chosen according to the model architecture. If the architecture is not known, an error will be raised -- in
+            this case, you should specify the target modules manually.
+        modules_to_save (`List[str]`):
+            List of modules apart from Road layers to be set as trainable and saved in the final checkpoint.
+    """
+
+    variant: Union[str, RoadVariant] = field(
+        default=RoadVariant.ROAD_1,
+        metadata={"help": ("Variant of the Road model to use. ")},
```
BenjaminBossan:
```suggestion
        metadata={"help": ("Variant of the Road model to use.")},
```

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
+
+from .config import RoadConfig, RoadVariant
+
+
+class RoadLayer(BaseTunerLayer):
+    """
+    Road layer.
+
+    Generally the idea of RoAD is to split the input vector into many 2D vectors and rotate each 2D vector with its own
```
BenjaminBossan, line 30: Is it really a "2D vector" or rather 2 vectors from the matrix (which can be much higher dimension than 2D)?

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+
+    For merging the weights, we make use of the following formula: R @ (W @ x + b) = (R @ W) @ x + R @ b. The lhs part
+    is how it is used in unmerged state (using efficient elementwise implementation instead of matrix multiplication)
+    and the rhs part is how it is used in merged state where (R @ W) becomes the new weight matrix and R @ b becomes
+    the new bias.
+
+    """
+
+    adapter_layer_names: tuple[str, ...] = ("road_theta", "road_alpha")
+
```
BenjaminBossan, line 55:
```suggestion
    other_param_names: tuple[str, ...] = ("variant", "group_size")

```

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+        self.road_alpha = nn.ParameterDict({})
+
+        self._disable_adapters = False
+        self.merged_adapters = []
+
+        base_layer = self.get_base_layer()
+        if isinstance(base_layer, nn.Linear):
+            in_features, out_features = base_layer.in_features, base_layer.out_features
+        else:
+            in_features, out_features = None, None
```
BenjaminBossan: In this case, the code below would fail, right? So let's just raise an error here (you probably copied this from LoRA, but it doesn't really make sense there either).

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+    def update_layer(
+        self,
+        adapter_name,
+        variant,
+        group_size,
+        init_weights,
+    ):
+        # collect the kwargs
+        kwargs = locals().copy()
+        del kwargs["self"]
```
BenjaminBossan: Can be deleted

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+        if variant == RoadVariant.ROAD_1:
+            size = self.out_features // 2
+        elif variant == RoadVariant.ROAD_2:
+            size = self.out_features
+        elif variant == RoadVariant.ROAD_4:
+            size = self.out_features * 2
+        else:
+            raise ValueError(f"Unsupported variant {variant} for RoadLayer. Supported variants are 1, 2, and 4.")
+        self.road_theta[adapter_name] = nn.Parameter(torch.rand(size))
+        self.road_alpha[adapter_name] = nn.Parameter(torch.rand(size))
```
BenjaminBossan: WDYT about initializing this with `torch.empty` and then in `reset_road_parameters`, in case that `init_weights=False`, fill them with `rand`? That way, if `init_weights=True` (which is most common), we don't double-initialize the values, making the initialization more efficient.

docs/source/package_reference/road.md
```diff
@@ -0,0 +1,29 @@
...
+
```
BenjaminBossan, line 24: Let's also mention that only linear layers are supported for now, as well as bnb.

src/peft/tuners/road/bnb.py
BenjaminBossan, line 1: Note to self: I only skimmed this file for now.

src/peft/tuners/road/__init__.py
BenjaminBossan, line 1: I think it would be good to mention somewhere that this implementation is based on your fork, with a reference to it. Perhaps here in the `__init__.py` is the right place, but maybe you have a better idea.

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+                    if not torch.isfinite(orig_weight).all():
+                        raise ValueError(
+                            f"NaNs detected in the merged weights. The adapter {active_adapter} seems to be broken"
+                        )
+
+                    base_layer.weight.data = orig_weight.contiguous().to(orig_dtype)
+
+                    if base_layer.bias is not None:
+                        orig_bias = base_layer.bias.clone()
+                        orig_bias = torch.matmul(road_R.to(orig_dtype), orig_bias)
```
BenjaminBossan, line 272: Let's also check for finiteness of the bias.

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+                weight = self.get_base_layer().weight
+                orig_dtype = weight.dtype
+                road_R = _get_delta_weight(
+                    self.variant[active_adapter],
+                    self.group_size[active_adapter],
+                    self.road_theta[active_adapter].data,
+                    self.road_alpha[active_adapter].data,
+                )
+                # Since our matrix are not necessarily orthogonal we need inverse instead of transpose
+                inv_road_R = torch.linalg.inv(road_R.to(torch.float32)).to(orig_dtype)
```
BenjaminBossan, line 313: Just curious: Do you know if this can introduce a significant overhead?

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+        # collect the kwargs
+        kwargs = locals().copy()
+        del kwargs["self"]
+
+        self.variant[adapter_name] = variant
+        self.group_size[adapter_name] = group_size
+
+        # Actual trainable parameters
+        if variant == RoadVariant.ROAD_1:
+            size = self.out_features // 2
```
BenjaminBossan, line 95: Let's first check that `self.out_features % 2 == 0` and raise an error with a helpful message otherwise.

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+        variant,
+        group_size,
+        init_weights,
+    ):
+        # collect the kwargs
+        kwargs = locals().copy()
+        del kwargs["self"]
+
+        self.variant[adapter_name] = variant
+        self.group_size[adapter_name] = group_size
```
BenjaminBossan, line 86: Let's check if `hidden_size % group_size == 0` and raise an error with a helpful message otherwise.

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+                f"Unsupported layer type '{type(base_layer)}' encountered, proceed at your own risk.", UserWarning
+            )
+        self.in_features = in_features
+        self.out_features = out_features
+
+    @property
+    def _available_adapters(self) -> set[str]:
+        return {*self.road_theta}
+
+    def update_layer(
```
BenjaminBossan, line 78: I suggest to add some checks to this method, as well as to `RoadConfig.__post_init__`. Please add tests for these checks to `test_initialzation.py`.

src/peft/tuners/road/config.py
```diff
@@ -0,0 +1,123 @@
...
+            )
+        },
+    )
+
+    def __post_init__(self):
+        super().__post_init__()
+        self.peft_type = PeftType.ROAD
+        self.target_modules = (
+            set(self.target_modules) if isinstance(self.target_modules, list) else self.target_modules
+        )
```
BenjaminBossan, line 122: Let's also add a check that `self.group_size % 2 == 0`.

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+def _apply_road(
+    variant: RoadVariant, group_size: int, road_theta: torch.Tensor, road_alpha: torch.Tensor, x: torch.Tensor
+):
+    first_col, second_col = _prepare_cols(variant, group_size, road_theta, road_alpha)
+
+    # Split in half groups and join back
+    x_grouped = x.reshape(-1, 2, group_size // 2)
+    x1 = x_grouped[:, 0, :]
+    x2 = x_grouped[:, 1, :]
+    rotate_half_x = torch.stack((-x2, x1), dim=1).reshape(x.shape)
```
BenjaminBossan, line 396: It's not clear to me why we use `-x2`, could you please add a clarifying comment?

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+        road_theta = road_theta.reshape(-1, 2, group_size)
+        theta_cos = road_theta[:, 0, :].cos().flatten()
+        theta_sin = road_theta[:, 1, :].sin().flatten()
+        road_alpha = road_alpha.reshape(-1, 2, group_size)
+        alpha_1 = road_alpha[:, 0, :].flatten()
+        alpha_2 = road_alpha[:, 1, :].flatten()
+
+        first_col = alpha_1 * theta_cos
+        second_col = alpha_2 * theta_sin
+
```
BenjaminBossan, line 382: Just to be safe, let's add an `else` and raise an error.

src/peft/tuners/road/layer.py
```diff
@@ -0,0 +1,398 @@
...
+    # here it is similar, but done in groups.
+    # https://discuss.huggingface.co/t/is-llama-rotary-embedding-implementation-correct/44509/3
+
+    # First column is simply put on the main diagonal
+    output_tensor = torch.diag(first_col)
+    # For second column we need to swap each half groups and add minus sign
+    size = second_col.shape[0]
+    swapped_second_col = second_col.reshape(-1, 2, group_size // 2)[:, [1, 0], :].flatten()
+    rotated_diag_second_col = torch.diag(swapped_second_col).reshape(-1, 2, group_size // 2, size)[:, [1, 0], :, :]
+    rotated_diag_second_col[:, 0, :, :] *= -1
```
BenjaminBossan, line 339: It's not clear to me why we use `*= -1`, could you please add a clarifying comment?

src/peft/tuners/road/model.py
```diff
@@ -0,0 +1,299 @@
...
+        for dispatcher in dispatchers:
+            new_module = dispatcher(target, adapter_name, road_config=road_config, **kwargs)
+            if new_module is not None:  # first match wins
+                break
+
+        if new_module is None:
+            # no module could be matched
+            raise ValueError(
+                f"Target module {target} is not supported. Currently, only the following modules are supported: "
+                "`torch.nn.Linear` "
```
BenjaminBossan:
```suggestion
                "`torch.nn.Linear`."
```

tests/test_custom_models.py
BenjaminBossan, line 1: Since RoAd supports mixed adapter batches, please extend the `TestMixedAdapterBatches` class to also test RoAd:

https://github.com/huggingface/peft/blob/92d65cafa51c829484ad3d95cf71d09de57ff066/tests/test_custom_models.py#L4509

Probably the easiest way would be to use `@pytest.mark.parametrize` and extract the `LoraConfig`s that are currently hard-coded into the test body there, check if the tests still pass, then add `RoadConfig`s and check again. Some tests are for layer types that RoAd doesn't support, you can leave those tests as they are, or you add `pytest.skipif`s for RoAd.

src/peft/tuners/road/model.py
BenjaminBossan, line 1: Note to self: Only skimmed this file.

BenjaminBossan (changes requested):
Thank you a lot for this PR to add RoAd to PEFT. I have to say, this is one of the most mature PRs to add a new PEFT method so far, really great work on understanding the PEFT code base and following the conventions.

I did a first review pass and didn't find any big flaws. There are a few smaller issues that I found, but I think they should be easily fixed. Please check my comments.

Something I'd be interested in is if you have checked the efficiency of mixed batch inference. If not, I think this it would be great to have a small script to check this. What I would imagine is to load a base model, add multiple RoAd/LoRA adapters (need not be trained), perform single adapter/batch inference and multi adapter/batch inference, then compare the relative overhead. If you don't have such a script, I'd be happy to help working on this.

Once you finish all your changes, don't forget to run `make style` before pushing and then ping me.

src/peft/tuners/road/config.py
ppetrushkov: One issue with large group size that I found appears when using tensor parallelism. Let's say with TP=2, in column parallel case the hidden layer output is split in two halves and each GPU processes its own half. Since computing RoAd requires inputs from the entire group, if we set group_size=hidden_size, it means that both halves need to be available on each GPU, requiring extra synchronisation and code changes. As long as each group fits in a tensor parallel replica, no extra synchronisation is needed.

Smaller group size (like 2), has no such disadvantage in theory, however at least with my current inference implementation it works much faster if the group size is at least 32 or 64.

src/peft/tuners/road/layer.py
ppetrushkov: In this case I'm talking about the theoretical idea of the paper to take a pair of elements from the hidden output, and rotate them with a 2D matrix, so I think saying 2D vector is appropriate. In the actual implementation it is indeed computed as a sum of 2 higher dimensional vectors instead, but in my view this representation is just a mathematical trick.

src/peft/tuners/road/layer.py
ppetrushkov: I initially did it like that based on lora implementation, but I ran into issues with some unit tests. If I understood it correctly, the issue was that some test uses init_weights=False and taking a cosine of a randomly large value produced some unstable result. I'll have another look to see if I can make it work.

src/peft/tuners/road/layer.py
ppetrushkov: I don't know. I didn't think much about this, since I consider speed of merging to be not very crucial. Another thing here is I'm not sure this matrix will always be invertible, I need to think about this part and maybe enhance this somehow.

src/peft/tuners/road/layer.py
ppetrushkov: I added a reference to the equation from the paper, basically it comes from the minus sign in the rotation matrix in the top right element.

ppetrushkov:
@BenjaminBossan thank you for the thorough review! I addressed some of the comments, will look at the remaining things and get back to you.

One comment regarding benchmarking mixed adapters - my focus was mainly on the VLLM implementation, as this is how we want to use it. To do this efficiently one needs something like grouped gemm to multiply hidden outputs with different adapter weights based on the adapter mapping. I implemented this in VLLM with a triton kernel, somewhat like what punica is doing, but much simpler. Doing this in plain torch would be much slower, or at least I don't know of a good way to do that. So benchmarking the torch version of mixed road inference in peft doesn't make much sense to me. Although, RoAd itself is still much faster than LoRA, whether mixed or not mixed.

FYI attaching one plot I created with my fork of VLLM comparing mixed inference with LoRA vs RoAd vs just base model. 
lora_vs_road_tpot_512_128

src/peft/tuners/road/config.py
BenjaminBossan: Okay, got it. I don't think we need to go into quite as much detail in the docs. Let's just mention that 64 is the default:

```diff
For best speed group size of at least 64 is recommended.
For best speed group size of at least 32 or 64 (the default) is recommended.

File: src/peft/tuners/road/layer.py (hunk 16 above)
Comment (BenjaminBossan) [reply src/peft/tuners/road/layer.py]: When passing `init_weights=False`, the values should be randomly initialized in `reset_road_parameters` (which can be renamed to `reset_parameters`) to avoid this.

File: src/peft/tuners/road/layer.py (hunk 19 above)
Comment (BenjaminBossan) [reply src/peft/tuners/road/layer.py:313]: I tested it locally with `meta-llama/Llama-3.1-8B` and unmerging was 4x slower than merging, but still fast overall (0.67 sec). Regarding invertibility, do you think there can be an issue in practice?

**Review by BenjaminBossan — COMMENTED** [review]
Regarding benchmarks: It makes sense that users who would be searching for the best performance would look into vLLM etc. Still, I think it's good to check if the advantages also manifest in PEFT.

I created a small toy benchmark:

<details>

<summary>Show code</summary>

```python
import time

import torch
from peft import get_peft_model, RoadConfig, LoraConfig

num_adapters = 10
num_iters = 50
model_id = "meta-llama/Llama-3.2-3B"

config0 = LoraConfig(init_lora_weights=False, r=4)  # matching params of RoadConfig
config1 = RoadConfig(init_weights=False)
inputs = [torch.randint(0, 100, (num_adapters, 50)).to(0) for _ in range(num_iters)]

for config in [config0, config1]:
    model = AutoModelForCausalLM.from_pretrained(model_id, torch_dtype=torch.bfloat16).to(0)
    model = get_peft_model(model, config, adapter_name="adapter0")
    for i in range(1, num_adapters):
        model.add_adapter(f"adapter{i}", config)
    model.eval()

    adapter_names = [f"adapter{i}" for i in range(num_adapters)]

    # warmup
    model(inputs[0])

    # Run the model w/o mixed adapter batch
    durations = {"no mixed batch": [], "   mixed batch": []}

    for i in range(num_iters):
        x = inputs[i]
        tic = time.perf_counter()
        model(x)
        toc = time.perf_counter()
        durations["no mixed batch"].append(toc - tic)

    # Run the model with mixed adapter batch
    for i in range(num_iters):
        x = inputs[i]
        tic = time.perf_counter()
        model(x, adapter_names=adapter_names)
        toc = time.perf_counter()
        durations["   mixed batch"].append(toc - tic)

    print(f"{config.peft_type.value} summary (over 10 runs)")
    for name, dur in durations.items():
        print(f"{name}: avg {sum(dur)/num_iters:.4f} sec, min {min(dur):.4f} sec, max {max(dur):.4f} sec, total: {sum(dur):.4f} sec")

    del model
    torch.cuda.empty_cache()
```

</details>

The results look quite promising so far:

```
LORA summary (over 10 runs)
no mixed batch: avg 0.0262 sec, min 0.0159 sec, max 0.0272 sec, total: 1.3117 sec
   mixed batch: avg 0.0878 sec, min 0.0848 sec, max 0.1182 sec, total: 4.3920 sec
ROAD summary (over 10 runs)
no mixed batch: avg 0.0263 sec, min 0.0179 sec, max 0.0269 sec, total: 1.3160 sec
   mixed batch: avg 0.0266 sec, min 0.0263 sec, max 0.0269 sec, total: 1.3297 sec
```

src/peft/tuners/road/layer.py
ppetrushkov: I thought about invertibility, I think in practice we can basically always expect it to work, since we initialize it as a block-diagonal matrix with each block being a 2D rotation matrix (but the rows are permuted when group_size>2) and from what I've seen after training the values don't change very significantly.

ppetrushkov:
@BenjaminBossan I think I addressed all your comments so far, please have another look. Thanks!

tests/test_custom_models.py
```diff
@@ -4629,17 +4668,28 @@ def test_requires_grad_fourierft_same_targets(self):
         )
 
 
+MIXED_ADAPTER_TEST_CASES = [
```
BenjaminBossan, line 4708: Let's just add a small comment that this is for PEFT methods that support mixed adapter batches.

tests/test_custom_models.py
```diff
@@ -4678,32 +4728,68 @@ def run_checks(self, model, inputs):
         assert torch.allclose(output0[1::3], output_mixed[1::3])
         assert torch.allclose(output1[2::3], output_mixed[2::3])
 
-    def test_mixed_adapter_batches_lora_mlp(self, mlp_lora):
+    @pytest.mark.parametrize("test_name, config0, config1", MIXED_ADAPTER_TEST_CASES)
+    def test_mixed_adapter_batches_lora_mlp(self, test_name, config0, config1):
```
```suggestion
    def test_mixed_adapter_batches_mlp(self, test_name, config0, config1):
```

tests/test_custom_models.py
```diff
@@ -4678,32 +4728,68 @@ def run_checks(self, model, inputs):
...
+                LoraConfig(target_modules=["lin1"], init_lora_weights=False),
+            ),
+            (
+                "RoAd mixed adapter with different target layers",
+                RoadConfig(target_modules=["lin0"], group_size=2, init_weights=False),
+                RoadConfig(target_modules=["lin1"], group_size=2, init_weights=False),
+            ),
+        ],
+    )
+    def test_mixed_adapter_batches_lora_different_target_layers(self, test_name, config0, config1):
```
```suggestion
    def test_mixed_adapter_batches_different_target_layers(self, test_name, config0, config1):
```

tests/test_custom_models.py
```diff
@@ -4678,32 +4728,68 @@ def run_checks(self, model, inputs):
...
+                LoraConfig(target_modules=["lin0"], modules_to_save=["lin1"], init_lora_weights=False),
+            ),
+            (
+                "RoAd mixed adapter with modules to save",
+                RoadConfig(target_modules=["lin0"], modules_to_save=["lin1"], group_size=2, init_weights=False),
+                RoadConfig(target_modules=["lin0"], modules_to_save=["lin1"], group_size=2, init_weights=False),
+            ),
+        ],
+    )
+    def test_mixed_adapter_batches_lora_multiple_modules_to_save(self, test_name, config0, config1):
```
```suggestion
    def test_mixed_adapter_batches_multiple_modules_to_save(self, test_name, config0, config1):
```

tests/test_custom_models.py
```diff
@@ -4678,32 +4728,68 @@ def run_checks(self, model, inputs):
...
+        "test_name, config0, config1",
+        [
+            (
+                "LoRA mixed adapter with unsupported layer",
+                LoraConfig(target_modules=["lin0"], modules_to_save=["gru"], init_lora_weights=False),
+                LoraConfig(target_modules=["lin0"], modules_to_save=["gru"], init_lora_weights=False),
+            ),
+        ],
+    )
+    def test_mixed_adapter_batches_lora_unsupported_layer_raises(self, test_name, config0, config1):
```
```suggestion
    def test_mixed_adapter_batches_unsupported_layer_raises(self, test_name, config0, config1):
```

tests/test_custom_models.py
```diff
@@ -4714,50 +4800,95 @@ def test_mixed_adapter_batches_lora_unsupported_layer_raises(self, mlp_lora):
...
+                LoraConfig(target_modules=["lin0", "lin1"], init_lora_weights=False),
+            ),
+            (
+                "RoAd mixed adapter with overlapping layers",
+                RoadConfig(target_modules=["lin0"], group_size=2, init_weights=False),
+                RoadConfig(target_modules=["lin0", "lin1"], group_size=2, init_weights=False),
+            ),
+        ],
+    )
+    def test_mixed_adapter_batches_lora_partly_overlapping_target_layers(self, test_name, config0, config1):
```
```suggestion
    def test_mixed_adapter_batches_partly_overlapping_target_layers(self, test_name, config0, config1):
```

tests/test_custom_models.py
```diff
@@ -4766,56 +4897,76 @@ def test_mixed_adapter_batches_mha_raises(self):
         with pytest.raises(TypeError, match=msg):
             self.run_checks(peft_model, inputs)
 
-    def test_mixed_adapter_batches_lora_length_mismatch_raises(self, mlp_lora):
+    @pytest.mark.parametrize("test_name, config0, config1", MIXED_ADAPTER_TEST_CASES)
+    def test_mixed_adapter_batches_lora_length_mismatch_raises(self, test_name, config0, config1):
```
```suggestion
    def test_mixed_adapter_batches_length_mismatch_raises(self, test_name, config0, config1):
```

tests/test_custom_models.py
```diff
@@ -4766,56 +4897,76 @@ def test_mixed_adapter_batches_mha_raises(self):
...
             "adapter_names": ["__base__"] * 5,  # wrong length!
         }
         msg = r"Length of `adapter_names` should be the same as the number of inputs, but got "
         with pytest.raises(ValueError, match=msg):
-            mlp_lora.forward(**inputs)
+            mlp_peft.forward(**inputs)
 
-    def test_mixed_adapter_batches_lora_training_mode_raises(self, mlp_lora):
+    @pytest.mark.parametrize("test_name, config0, config1", MIXED_ADAPTER_TEST_CASES)
+    def test_mixed_adapter_batches_lora_training_mode_raises(self, test_name, config0, config1):
```
```suggestion
    def test_mixed_adapter_batches_training_mode_raises(self, test_name, config0, config1):
```

tests/test_custom_models.py
```diff
@@ -4766,56 +4897,76 @@ def test_mixed_adapter_batches_mha_raises(self):
...
-        mlp_lora = mlp_lora.train()
+        mlp_peft = mlp_peft.train()
         msg = r"Cannot pass `adapter_names` when the model is in training mode."
         with pytest.raises(ValueError, match=msg):
-            mlp_lora.forward(**inputs)
+            mlp_peft.forward(**inputs)
 
-    def test_mixed_adapter_batches_lora_disabled(self, mlp_lora):
+    @pytest.mark.parametrize("test_name, config0, config1", MIXED_ADAPTER_TEST_CASES)
+    def test_mixed_adapter_batches_lora_disabled(self, test_name, config0, config1):
```
```suggestion
    def test_mixed_adapter_batches_disabled(self, test_name, config0, config1):
```

tests/test_custom_models.py
```diff
@@ -4766,56 +4897,76 @@ def test_mixed_adapter_batches_mha_raises(self):
...
-        with mlp_lora.disable_adapter():
-            output_mixed = mlp_lora.forward(**inputs)
+        with mlp_peft.disable_adapter():
+            output_mixed = mlp_peft.forward(**inputs)
 
         assert torch.allclose(output_disabled, output_mixed)
 
-    def test_mixed_adapter_batches_lora_merged_raises(self, mlp_lora):
+    @pytest.mark.parametrize("test_name, config0, config1", MIXED_ADAPTER_TEST_CASES)
+    def test_mixed_adapter_batches_lora_merged_raises(self, test_name, config0, config1):
```
```suggestion
    def test_mixed_adapter_batches_merged_raises(self, test_name, config0, config1):
```

tests/test_custom_models.py
```diff
@@ -4766,56 +4897,76 @@ def test_mixed_adapter_batches_mha_raises(self):
...
-    def test_mixed_adapter_batches_lora_wrong_adapter_name_raises(self):
+    @pytest.mark.parametrize(
+        "test_name, config",
+        [
+            (
+                "LoRA mixed batch wrong adapter name",
+                LoraConfig(target_modules=["lin0"], init_lora_weights=False),
+            ),
+            (
+                "LoRA mixed batch wrong adapter name",
```
```suggestion
                "RoAD mixed batch wrong adapter name",
```

examples/road_finetuning/road_finetuning.py
```diff
@@ -0,0 +1,201 @@
...
+    parser.add_argument("--batch_size", type=int, default=1, help="Batch size")
+    parser.add_argument("--num_epochs", type=int, default=1, help="Number of training epochs")
+    parser.add_argument("--learning_rate", type=float, default=3e-3, help="Learning rate")
+    parser.add_argument("--cutoff_len", type=int, default=512, help="Cutoff length for tokenization")
+    parser.add_argument("--val_set_size", type=int, default=500, help="Validation set size")
+    parser.add_argument("--quantize", action="store_true", help="Use quantization")
+    parser.add_argument("--eval_step", type=int, default=10, help="Evaluation step interval")
+    parser.add_argument("--save_step", type=int, default=100, help="Save step interval")
+    parser.add_argument("--device", type=str, default="cuda:0", help="Device to use for training")
+    parser.add_argument("--variant", type=str, default="1", choices=["1", "2", "4"], help="RoAD variant")
```
BenjaminBossan: Variants should be `"road_1"`, etc. or you have to allow `"1"` etc. too.

examples/road_finetuning/road_finetuning.py
BenjaminBossan, line 1: Not a huge deal, but I got an OOM with 24 GB VRAM using the script like this:

`python road_finetuning.py --base_model "meta-llama/Llama-3.2-3B" --data_path timdettmers/openassistant-guanaco --variant road_1`

The same script runs when using LoRA (rank 8) instead of RoAD. I had to restrict the target layers to `--road_target_modules "q_proj,v_proj"` to avoid OOM.

src/peft/tuners/road/layer.py
BenjaminBossan: Sounds good, then let's just leave this as is.

BenjaminBossan (changes requested):
Thanks for addressing my comments, there is not much missing. There is an error in the included example, please check my comments. The of my comments are just smaller changes.

Before we can merge, since we have bnb support, let's also add some GPU tests for this. You can copy the RandLoRA examples and adjust them for RoAD:

https://github.com/huggingface/peft/blob/e0b2ca79771066162d6efec86764b047a323b98a/tests/test_gpu_examples.py#L1492-L1718

examples/road_finetuning/road_finetuning.py
BaohaoLiao: Hi @BenjaminBossan, this is expected. 

RoAd utilizes more memory for saving the activation due to its implementation:
- LoRA: y = (W + BA)x, need to save x and y for back-propagation
- RoAd: z = Ry = R(Wx), need to save x, y and z for back-propagation

This is one of his limitation, similar to other orthogonal finetuning methods, like OFT and BOFT.

ppetrushkov:
@BenjaminBossan done.

Regarding MetaMath, I tried a bit, but didn't improve much. Including all linear layers helped most. I think some tasks that require larger changes and more parameters are harder for RoAd.

examples/road_finetuning/road_finetuning.py
BenjaminBossan: Makes sense, thanks for explaining.

BenjaminBossan (changes requested):
Thanks for all the changes, this PR looks pretty much ready for me. Please run `make style` to ensure that the linter is happy.

BenjaminBossan:
@ppetrushkov We're missing an entry in the [toc tree](https://github.com/huggingface/peft/blob/04d41cbcd061bf1ab1185e111054bae012cb1894/docs/source/_toctree.yml#L135), which is why the doc builder complains. Regarding the failing CI tests, this is just flakiness and can be ignored.

When you're finished with your changes, please ping me.

ppetrushkov:
@BenjaminBossan added it to the toc, let me know if anything else is missing.

BenjaminBossan:
Thanks for updating the toc. A recent commit created a merge conflict with your PR, could you please merge with/rebase on the latest main and fix the conflict? For the resolution, just take what's on main, it'll work with RoAd.

PR #2681, opened 2025-07-30
Fix trainable tokens with fsdp

BenjaminBossan:
When using FSDP with trainable tokens, there was an error when retrieving the `state_dict` of the `TrainableTokensWrapper`. The reason is that for the `state_dict` that is passed to `get_peft_model_state_dict`, the FSDP wrapper was already unwrapped, which means the keys don't have the FSDP-specific prefix. However, in the PEFT code, when looking up keys from said `state_dict`, the prefix was not removed. Now it is removed, making the lookup succeed. The same logic applies to `set_peft_model_state_dict`.

I could successfully start training with FSDP and trainable tokens locally by adjusting the `examples/sft` script to include trainable tokens. Checkpoints could be successfully created and resumed from. The only change I needed to make was to configure `use_orig_params=True` for FSDP.

I updated the docs to mention this. Furthermore, I removed mention of potential issues with DeepSpeed (I tested ZeRO3 locally and it worked), as well as mention of weight-tying not working (outdated info).

### DIFF
--- a/src/peft/tuners/lora/config.py
+++ b/src/peft/tuners/lora/config.py
@@ -283,7 +283,7 @@ class LoraConfig(PeftConfig):
             Either you specify a list of indices which will then target the model's input embedding layer (or, if not
             found, `embed_tokens`). Alternatively, you can specify a dictionary where the key is the name of the
             embedding module and the values are the list of token indices, e.g. `{'embed_tokens': [0, 1, ...]}`. Note
-            that training with FSDP/DeepSpeed might not yet be fully supported with this option enabled.
+            that training with FSDP requires `use_orig_params=True`.
         loftq_config (`Optional[LoftQConfig]`):
             The configuration of LoftQ. If this is not None, then LoftQ will be used to quantize the backbone weights
             and initialize Lora layers. Also pass `init_lora_weights='loftq'`. Note that you should not pass a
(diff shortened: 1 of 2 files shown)


src/peft/utils/save_and_load.py
```diff
@@ -219,6 +219,10 @@ def renamed_dora_weights(k):
     # ADDITIONAL TRAINING MODULES / MODULES_TO_SAVE
     for name, module in model.named_modules():
         if isinstance(module, AuxiliaryTrainingWrapper):
+            if name.startswith("_fsdp_wrapped_module."):
```
BenjaminBossan, line 222: I know the check is not technically required, but to me, it adds clarity to the intent. LMK if I should remove it though.

src/peft/tuners/lora/config.py
```diff
@@ -283,7 +283,7 @@ class LoraConfig(PeftConfig):
             Either you specify a list of indices which will then target the model's input embedding layer (or, if not
             found, `embed_tokens`). Alternatively, you can specify a dictionary where the key is the name of the
             embedding module and the values are the list of token indices, e.g. `{'embed_tokens': [0, 1, ...]}`. Note
-            that training with FSDP/DeepSpeed might not yet be fully supported with this option enabled.
+            that training with FSDP requires `use_orig_params=True`.
```
githubnemo: I think it's worthwhile to add an info about what happens in case we don't add this parameter or a reason for why it is required.

PR #2686, opened 2025-07-31
Updated MetaMathQA results

githubnemo:
- Updated results for OFT, C3A and shira
- New results for trainable tokens (for completeness)

Trainable tokens wasn't tuned a lot, we could probably search for better tokens and increase the learning rate. We can do this later.


BenjaminBossan (approved):
Thanks for updating, LGTM.

PR #2693, opened 2025-08-01
[WIP] Add fast-forward training to MetaMathQA

githubnemo:
This is an experiment for implementing fast-forward training (https://aclanthology.org/2024.emnlp-main.535.pdf) in PEFT. If it performs well on the MetaMathQA set it can be explored further.

The implementation is solely derived from the paper, it is possible that the implementation still contains errors but for a fixed backprop budget (1000 steps) the fast-forward implementation reaches a slightly better test score after a negligible amount of added train time (12s). Since fast-forward steps are much faster than backprop steps it doesn't make sense to add update steps but rather to compare wall-clock training time, a much longer running experiment might make sense. The difference so far is, however, pretty slim.

While the implementation allows for up to 1000 fast-forward steps I didn't observe more than 10 consecutive steps during training. Swapping out training for validation loss increases the step count, so does allowing a certain error margin between losses but both seem to lead to overfitting.


githubnemo:
There were some doubts whether fast-forward is not just a fancy way of increasing the learning rate 10x. I think the recent [LoRA without regret](https://thinkingmachines.ai/blog/lora/#optimal-learning-rate-and-rank) blog post underlines what we found here as well. LoRA adapters can be trained with 10x the LR compared to full fine-tuning.

PR #2694, opened 2025-08-01
validated cpt, ephemeral_gpu_offloading and eva finetuning on XPU

yao-matrix:

### DIFF
--- a/examples/eva_finetuning/eva_finetuning.py
+++ b/examples/eva_finetuning/eva_finetuning.py
@@ -77,7 +77,6 @@ initialize_lora_eva_weights(peft_model, dataloader)
 
 # setup training arguments
 training_args = TrainingArguments(
-    report_to="none",
     per_device_train_batch_size=batch_size,
     learning_rate=learning_rate,
     gradient_accumulation_steps=gradient_accumulation_steps,
--- a/examples/eva_finetuning/eva_finetuning_multi_accelerator.py
+++ b/examples/eva_finetuning/eva_finetuning_multi_accelerator.py
@@ -113,7 +113,6 @@ initialize_lora_eva_weights(peft_model, eva_state_dict=eva_state_dict)
 
 # setup training arguments
 training_args = TrainingArguments(
-    report_to="none",
     per_device_train_batch_size=batch_size,
     learning_rate=learning_rate,
     gradient_accumulation_steps=gradient_accumulation_steps,


examples/eva_finetuning/README.md
```diff
@@ -103,9 +104,9 @@ Later you can load the state_dict into a `PeftModel` by using the `eva_state_dic
 initialize_lora_eva_weights(peft_model, eva_state_dict=eva_state_dict)
 ```
 
-## Leveraging multiple GPUs
+## Leveraging multiple accelerators
 
-EVA initialization can be parallelized across multiple GPUs. In this case inputs from multiple GPUs are gathered before computing the SVD for the batch. This requires that the model is wrapped in a `torch.nn.DataParallel` or `torch.nn.DistributedDataParallel` class. An example of how to use this can be found in [eva_finetuning_multi_gpu.py](https://github.com/huggingface/peft/blob/main/examples/eva_finetuning/eva_finetuning_multi_gpu.py).
+EVA initialization can be parallelized across multiple accelerators. In this case inputs from multiple accelerators are gathered before computing the SVD for the batch. This requires that the model is wrapped in a `torch.nn.DataParallel` or `torch.nn.DistributedDataParallel` class. An example of how to use this can be found in [eva_finetuning_multi_accelerator.py](https://github.com/huggingface/peft/blob/main/examples/eva_finetuning/eva_finetuning_multi_accelerator.py).
```
BenjaminBossan, line 109: Just wondering: Are Intel accelerators not still technically "GPUs"? Is it really needed to replace "GPU" in the docs and comments with "accelerator"?

yao-matrix: Yes, Intel XPU are still technical GPUs architecture, which is general-purpose and programmable. I am using the term `accelerator` by following PyTorch's vocabulary([here](https://docs.pytorch.org/docs/stable/torch.html#accelerators)), PyTorch are supporting more accelerators, so I am using `accelerator` here, so if some people validated another new device later, they don't need to change README anymore, and even not `eva_finetuning.py`. Pls let me know your opinions, thx.

BenjaminBossan: That makes sense, thanks for explaining.

BenjaminBossan (approved):
Thanks for checking that CPT, EVA, and GPU offloading work with XPU

PR #2698, opened 2025-08-02
[WIP] Update `LoraConfig` for KaSA implementation

nsbg:
cc @BenjaminBossan 

I was delayed in updating the code because I was focusing on company work, but now I'm planning to resume the project in earnest. If I have any questions about implementing the code, may I continue to ask you?

I apologize for opening a new pull request, as the previous one was closed 🥲 Thank you for your understanding.

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -113,6 +113,7 @@ class LoraLayer(BaseTunerLayer):
         self._disable_adapters = False
         self.merged_adapters = []
         self.use_dora: dict[str, bool] = {}  # not actively used anymore after #2443, keep it for BC
+        self.use_rslora: dict[str, bool] = {}
         self.lora_bias: dict[str, bool] = {}
         self.lora_magnitude_vector = torch.nn.ModuleDict()  # for DoRA
         self._caches: dict[str, Any] = {}
@@ -254,6 +255,8 @@ class LoraLayer(BaseTunerLayer):
             self.scaling[adapter_name] = lora_alpha / math.sqrt(r)
         else:
             self.scaling[adapter_name] = lora_alpha / r
+
+        self.use_rslora[adapter_name] = use_rslora
 
         self.use_dora[adapter_name] = use_dora
 
@@ -528,7 +531,10 @@ class LoraLayer(BaseTunerLayer):
         if adapter not in self.scaling:
             # Ignore the case where the adapter is not in the layer
             return
-        self.scaling[adapter] = scale * self.lora_alpha[adapter] / self.r[adapter]
+        if self.use_rslora.get(adapter, False):
+            self.scaling[adapter] = scale * self.lora_alpha[adapter] / math.sqrt(self.r[adapter])
+        else:
+            self.scaling[adapter] = scale * self.lora_alpha[adapter] / self.r[adapter]
 
     def scale_layer(self, scale: float | int) -> None:
         """Multiply the current scale of all active adapters by the provided factor"""
@@ -553,9 +559,12 @@ class LoraLayer(BaseTunerLayer):
                 continue
 
             if scale is None:
-                self.scaling[active_adapter] = self.lora_alpha[active_adapter] / self.r[active_adapter]
+                if self.use_rslora.get(active_adapter, False):
+                    self.scaling[active_adapter] = self.lora_alpha[active_adapter] / math.sqrt(self.r[active_adapter])
+                else:
+                    self.scaling[active_adapter] = self.lora_alpha[active_adapter] / self.r[active_adapter]
             else:
-                self.scaling[active_adapter] /= scale
+                self.scaling[active_adapter] = self.scaling[active_adapter] / scale
 
     def _check_forward_args(self, x, *args, **kwargs):
         """Check if the arguments are compatible with the configs and state of the model"""
@@ -960,6 +969,8 @@ class Embedding(nn.Module, LoraLayer):
         else:
             self.scaling[adapter_name] = lora_alpha / r
 
+        self.use_rslora[adapter_name] = use_rslora
+
         self.use_dora[adapter_name] = use_dora
 
         if init_lora_weights == "loftq":
@@ -1259,6 +1270,8 @@ class _ConvNd(nn.Module, LoraLayer):
             self.scaling[adapter_name] = lora_alpha / math.sqrt(r)
         else:
             self.scaling[adapter_name] = lora_alpha / r
+
+        self.use_rslora[adapter_name] = use_rslora
 
         self.use_dora[adapter_name] = use_dora
 
@@ -2032,6 +2045,8 @@ class ParamWrapper(nn.Module, LoraLayer):
             self.scaling[adapter_name] = lora_alpha / math.sqrt(r)
         else:
             self.scaling[adapter_name] = lora_alpha / r
+
+        self.use_rslora[adapter_name] = use_rslora
 
         self.use_dora[adapter_name] = use_dora
 
--- a/tests/test_initialization.py
+++ b/tests/test_initialization.py
@@ -14,6 +14,7 @@
 
 import copy
 import itertools
+import math
 import platform
 import re
 import warnings
@@ -3889,6 +3890,44 @@ class TestScaling:
         self.unscale_layer(model, 3)
         scalings = self.get_scalings(model)
         expected = [2.0] * n_layers
+        assert scalings == expected
+
+    def test_scaling_with_rslora(self, model):
+        n_layers = 5
+        rank, lora_alpha = 8, 16
+        config = LoraConfig(
+            r=rank,
+            lora_alpha=lora_alpha,
+            use_rslora=True,
+            target_modules=["k_proj"],
+        )
+        model = get_peft_model(model, config)
+        scalings = self.get_scalings(model)
+        expected = [lora_alpha / math.sqrt(rank)] * n_layers
+        assert scalings == expected
+
+        # double
+        self.scale_layer(model, 2)
+        scalings = self.get_scalings(model)
+        expected = [2 * lora_alpha / math.sqrt(rank)] * n_layers
+        assert scalings == expected
+
+        # back to original
+        self.unscale_layer(model, None)
+        scalings = self.get_scalings(model)
+        expected = [lora_alpha / math.sqrt(rank)] * n_layers
+        assert scalings == expected
+
+        # triple
+        self.set_scale(model, "default", 3)
+        scalings = self.get_scalings(model)
+        expected = [3 * lora_alpha / math.sqrt(rank)] * n_layers
+        assert scalings == expected
+
+        # back to original
+        self.unscale_layer(model, 3)
+        scalings = self.get_scalings(model)
+        expected = [lora_alpha / math.sqrt(rank)] * n_layers
         assert scalings == expected
 
     def test_scaling_rank_pattern_alpha_pattern(self, model):


BenjaminBossan (changes requested):
Thank you for resuming your work on KaSA.

Implementation-wise, we need to take a different approach. Right now, KaSA is just added to the normal LoRA code, but we only want to activate it if the user opts in. Therefore, it should be implemented in a separate class, something like `KasaVariant`, in `peft/tuners/lora/variants.py`. Please check how DoRA is implemented and use a similar approach, as I have detailed in [my previous comment](https://github.com/huggingface/peft/pull/2543#issuecomment-2930428128). If anything is unclear, feel free to ask.

BenjaminBossan:
gentle ping @nsbg

iambogeumkim:
Thank you for your alert!

I spent some time looking over the KaSA paper and code to get ready for more serious work, but it does seem pretty difficult 🥲 My goal is to upload code that's ready for review before the end of September, so I'm going to try even harder.

Right now, I'm stuck at the 'Extend LoRA variant resolution' stage you mentioned. Honestly, this seems like the most important part, but it's hard for me to figure out where to start—specifically, which file and class I should work on first. Could you help me with this?

BenjaminBossan:
That's great to see, thanks for picking this back up.

> Right now, I'm stuck at the 'Extend LoRA variant resolution' stage you mentioned. Honestly, this seems like the most important part, but it's hard for me to figure out where to start—specifically, which file and class I should work on first. Could you help me with this?

You're already on the right track, you added `KasaLinearVariant`, which is the most important step. There are definitely some changes required there, as there is some code that is only relevant for DoRA and can be removed for KaSA. But we can leave that as is for now. 

Next about resolving the variants. As a first step, let's revert the changes you made to `lora/layer.py` and start fresh. We don't need a `self.use_kasa` attribute, we only have `self.use_dora` for backwards compatibility, as we didn't have LoRA variants when we first implemented DoRA.

Then let's look at these lines in `lora.Linear`:

https://github.com/huggingface/peft/blob/a3197b1ec54baa107e2efc4a30c19209d9ea489a/src/peft/tuners/lora/layer.py#L636-L642

Here we need to extend the functionality to add KaSA. The updated method could be something like:

```python
    def resolve_lora_variant(self, *, use_dora: bool, use_kasa: bool, **kwargs) -> Optional[LoraVariant]:
        if use_dora and use_kasa:
            raise ValueError("Cannot use DoRA and KaSA at the same time, please choose only one.")

        variant = None
        if use_dora:
            from .variants import DoraLinearVariant

            variant = DoraLinearVariant()
        elif use_kasa:
            ...

        return variant
```

Does that make sense? Similarly, we'd have to update the `resolve_lora_variant` methods of other LoRA layers, depending on whether they work with KaSA or not (I'm not sure if KaSA works with `Conv2d` etc.).

I would suggest that you work on this as a next step, then we'll see what else needs to be done.

iambogeumkim:
wow I really appreciate your sincere feedback. I'll read your advice carefully and then move forward 🤗

iambogeumkim:
@BenjaminBossan I modified the code in the files below based on what you explained. Please give me feedback if there are parts that still need fixing, and then we can discuss the next steps.

### 1. variants.py

- Completed updates to methods in the `KasaLinearVariants` class

### 2. layer.py

- In the LoraLayer class, added `self.use_kasa[adapter_name] = use_kasa` inside the update_layer method

- In the Linear class, added KaSA handling logic inside the `get_delta_weight` method

src/peft/tuners/lora/layer.py
```diff
@@ -173,7 +176,18 @@ def resolve_lora_variant(self, *, use_dora: bool, **kwargs) -> Optional[LoraVari
         convention, and not here.
 
         """
-        return None
+        if use_dora and use_kasa:
```
BenjaminBossan: Let's undo the changes in this method body and return `None`. Instead, since this KaSA layer is implemented for `Linear` only, add the logic to `lora.Linear.resolve_lora_variant` instead.

Also, we should update the `resolve_lora_variant` methods of the other layer types like `lora.Embedding.resolve_lora_variant` to accept the `use_kasa` argument but raise an error if it's `True`. Otherwise, users may add it to non-supported layers and not notice that it doesn't actually do anything there.

src/peft/tuners/lora/layer.py
```diff
@@ -217,7 +232,20 @@ def update_layer(
...
+        
+        weight = self.get_base_layer().weight
+        dtype = weight.dtype
+        svd_rank = self.in_features - r
+        weight = weight.to(torch.float32)
+        U, S, Vh = torch.linalg.svd(weight.data, full_matrices=False)
+        U_principle, S_principle, Vh_principle = U[:, :svd_rank], S[:svd_rank], Vh[:svd_rank, :]
+        self.get_base_layer().weight.data = (U_principle @ torch.diag(S_principle) @ Vh_principle).to(dtype)
 
+        #########################
```
BenjaminBossan: All of this can be removed, since it's part of `KasaLinearVariant.init`, right?

src/peft/tuners/lora/variants.py
```diff
@@ -308,3 +309,56 @@ class DoraConv3dVariant(_DoraConvNdVariant):
...
+    @staticmethod
+    def init(module: Linear, adapter_name: str, **kwargs: Any) -> None:
+        if not module.lora_diag:
+            # first kasa layer being added, add lora_diag to the list of learnable parameters
+            module.adapter_layer_names = module.adapter_layer_names[:] + ("lora_diag",)
+
+        # initialize lora_diag
+        module.lora_diag[adapter_name] = nn.Parameter(torch.randn(module.r[adapter_name]), requires_grad=True)
+
+        # SVD
```
BenjaminBossan: Let's add a reference here, so that we know the origin:
`# see https://github.com/juyongjiang/KaSA/blob/f85e88c22d0fa4cb8ab2923d7c2bf1bbec152da3/peft/src/peft/tuners/lora/layer.py#L132`

src/peft/tuners/lora/variants.py
```diff
@@ -308,3 +309,56 @@ class DoraConv3dVariant(_DoraConvNdVariant):
...
+
+    @staticmethod
+    def merge_unsafe(module: Linear, active_adapter: str, orig_weight: torch.Tensor) -> None:
+        delta_weight = module.get_delta_weight(active_adapter)
+        orig_weight.data += delta_weight
+
+    @staticmethod
+    def unmerge(module: Linear, active_adapter: str, orig_weight: torch.Tensor) -> torch.Tensor:
+        delta_weight = module.get_delta_weight(active_adapter)
+        return orig_weight - delta_weight
```
BenjaminBossan, line 521: KaSA should have an influence on the merged weights, should it not?

src/peft/tuners/lora/variants.py
```diff
@@ -308,3 +309,56 @@ class DoraConv3dVariant(_DoraConvNdVariant):
...
+        dropout = module.lora_dropout[active_adapter]
+        scaling = module.scaling[active_adapter]
+        
+        diag = torch.diag(module.lora_diag[active_adapter])
+        x = module._cast_input_dtype(x, lora_A.weight.dtype)
+        if isinstance(dropout, nn.Identity) or not module.training:
+            x = dropout(x)
+    
+        # KaSA calculation
+        lora_output = lora_B(torch.einsum('ijk,kl->ijl', lora_A(x), diag)) * scaling
```
BenjaminBossan: Again, let's add a reference:
`# see https://github.com/juyongjiang/KaSA/blob/f85e88c22d0fa4cb8ab2923d7c2bf1bbec152da3/peft/src/peft/tuners/lora/layer.py#L602C21-L602C110`

BenjaminBossan (changes requested):
Thanks for integrating my feedback. I gave this another review and noted the next few changes that are necessary. Please check my comments.

Apart from this, the branch is now encountering merge conflicts. Could you please bring your fork up-to-date with the remote and then merge with, or rebase on, the latest main branch from PEFT? If you have questions on how to resolve the merge conflicts, don't hesitate to ask.

Furthermore, please always run `make style` on your changes before pushing to make our linter happy.

More of a note for myself: Since KaSA updates the base weights of the model, we will have to take extra care to ensure that it works correctly when saving and loading the adapter.

src/peft/tuners/lora/variants.py
iambogeumkim:
```
# initialize lora_diag
module.lora_diag[adapter_name] = nn.Parameter(torch.randn(module.r[adapter_name]), requires_grad=True)

# see https://github.com/juyongjiang/KaSA/blob/f85e88c22d0fa4cb8ab2923d7c2bf1bbec152da3/peft/src/peft/tuners/lora/layer.py#L132
        
# SVD
```
I put it in here, how is it?

src/peft/tuners/lora/variants.py
iambogeumkim:
```
# KaSA calculation
# see https://github.com/juyongjiang/KaSA/blob/f85e88c22d0fa4cb8ab2923d7c2bf1bbec152da3/peft/src/peft/tuners/lora/layer.py#L602C21-L602C110
lora_output = lora_B(torch.einsum('ijk,kl->ijl', lora_A(x), diag)) * scaling
return result + lora_output
```
I inserted this near where the actual calculation logic begins, rather than just in an empty space. I think this is a bit better.

iambogeumkim:
@BenjaminBossan oh I didn't mean to close the branch, but it seems to have closed while I was merging with the main branch. I guess I'll have to open a new PR, right? 😰

+) when I tried to sync with the main branch, I ended up discarding all my commits, so did that cause it to close?

src/peft/tuners/lora/variants.py
iambogeumkim: Although this PR is closed, it seems I've incorporated everything else except for this comment (of course, you'd have to look at the code). Could you explain this question in more detail?

BenjaminBossan:
> oh I didn't mean to close the branch, but it seems to have closed while I was merging with the main branch. I guess I'll have to open a new PR, right? 😰
> 
> +) when I tried to sync with the main branch, I ended up discarding all my commits, so did that cause it to close?

I don't know what happened, but I could re-open the PR and there are some changes visible. Can you double check that everything looks as expected? If for some reason it's not what it's expected, you can create a new PR and push your local branch.

iambogeumkim:
I usually handle merges in the terminal, and I suspect the pull request was closed because I accidentally wiped the commit history while using the 'Sync fork' feature on GitHub. I'll be more careful in the future. Thanks for reopening it.

I'll review the changes and open a new PR if needed. Sorry to keep bothering you with this.

BenjaminBossan:
> I'll review the changes and open a new PR if needed. Sorry to keep bothering you with this.

No worries. If the diff on this PR looks good, let me know and I'll do a review. Only open a new PR if for some reason, the code here does not correspond to what it should be.

iambogeumkim:
@BenjaminBossan I checked layer.py/variants.py and KasaLinearVariants class in variants.py was removed. I added it again and I updated file based on your minor feedback, so I think we can discuss in this PR continually.

BTW I ran `make style` command and got this error.
```
make style
process_begin: CreateProcess(NULL, ruff check --fix src tests examples docs scripts docker, ...) failed.
```

I ran `pip install -e .[test]` command in https://huggingface.co/docs/peft/install#source, but I got same error. Do I just run that command directly without needing to set up a virtual environment?

iambogeumkim:
image

maybe `make style` related error was fixed. After applying this command, quite a few files have changed. Is it okay to just push them? Also, what exactly does `make style` do?

BenjaminBossan:
> maybe `make style` related error was fixed. After applying this command, quite a few files have changed. Is it okay to just push them? Also, what exactly does `make style` do?

No, let's not push any changes to unrelated files. If `make style` changes unrelated files, it's often to one of these reasons:

1. Wrong `ruff` version: check that v0.12.12 is installed in your virtual environment.
2. Not picking up the config: There are some settings for ruff in the `pyproject.toml` file, ensure that it's there when you run `make style`.

iambogeumkim:
@BenjaminBossan 

> Also, we should update the resolve_lora_variant methods of the other layer types like lora.Embedding.resolve_lora_variant to accept the use_kasa argument but raise an error if it's True. Otherwise, users may add it to non-supported layers and not notice that it doesn't actually do anything there.

I referred to your explanation and added the use_kasa parameter to the `resolve_lora_variant` method in the classes below. 

- Linear (line 702)
- Embedding (line 934)
- _ConvNd (line 1263)
- Conv2d (line 1507)
- Conv1d (line 1524)
- Conv3d (line 1541)

The logic for raising errors in each layer hasn’t been applied yet, but I committed first to check whether adding the parameter in this way matches what you meant. Excluding the Linear class, it seems like an error should be raised when use_kasa is true in the other classes. However, I might be mistaken, so please feel free to give me feedback anytime. Also, I noticed there’s no part that calls `KasaLinearVariant`—should this be called inside the linear class? I’m a bit confused about this part.

src/peft/tuners/lora/layer.py
```diff
@@ -123,6 +124,9 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
         self.lora_variant: dict[str, LoraVariant] = {}
         self.kwargs = kwargs
 
+        # Diag value
+        self.lora_diag = nn.ParameterDict({})
```

src/peft/tuners/lora/layer.py
```diff
@@ -259,6 +266,7 @@ def update_layer(
         self.use_rslora[adapter_name] = use_rslora
 
         self.use_dora[adapter_name] = use_dora
+        self.use_kasa[adapter_name] = use_kasa
```
BenjaminBossan: Can be removed, `use_dora` is only here for legacy reasons.

src/peft/tuners/lora/layer.py
```diff
@@ -819,7 +829,12 @@ def get_delta_weight(self, adapter) -> torch.Tensor:
             weight_A = weight_A.float()
             weight_B = weight_B.float()
 
-        output_tensor = transpose(weight_B @ weight_A, self.fan_in_fan_out) * self.scaling[adapter]
+        # KaSA handling
+        if self.use_kasa.get(adapter, False):
+            diag = torch.diag(self.lora_diag[adapter])
+            output_tensor = transpose(weight_B @ diag @ weight_A, self.fan_in_fan_out) * self.scaling[adapter]
+        else:
+            output_tensor = transpose(weight_B @ weight_A, self.fan_in_fan_out) * self.scaling[adapter]
```
BenjaminBossan: Let's remove this. Instead, this logic should be implemented in the `KasaLinearVariant` (see comment below).

src/peft/tuners/lora/variants.py
```diff
@@ -438,6 +438,57 @@ def init(module: Conv3d, adapter_name: str, **kwargs: Any) -> None:
...
+        
+         # SVD
+        weight = module.get_base_layer().weight
+        dtype = weight.dtype
+        svd_rank = module.in_features - module.r[adapter_name]
+        weight = weight.to(torch.float32)
+        U, S, Vh = torch.linalg.svd(weight.data, full_matrices=False)
+        U_principle, S_principle, Vh_principle = U[:, :svd_rank], S[:svd_rank], Vh[:svd_rank, :]
+        module.get_base_layer().weight.data = (U_principle @ torch.diag(S_principle) @ Vh_principle).to(dtype)
+
```
BenjaminBossan: Let's add a new `@staticmethod` called `_get_delta_weight` here. This method should implement the KaSA delta weight logic from above:

```python
            diag = torch.diag(module.lora_diag[adapter])
            output_tensor = transpose(weight_B @ diag @ weight_A, module.fan_in_fan_out) * module.scaling[adapter]
```

Then, the `merge_safe`, `merge_unsafe`, and `unmerge` methods below can call `KasaLinearVariant._get_delta_weight(...)`.

src/peft/tuners/lora/variants.py
```diff
@@ -438,6 +438,57 @@ def init(module: Conv3d, adapter_name: str, **kwargs: Any) -> None:
...
+        lora_A = module.lora_A[active_adapter]
+        lora_B = module.lora_B[active_adapter]
+        dropout = module.lora_dropout[active_adapter]
+        scaling = module.scaling[active_adapter]
+        diag = torch.diag(module.lora_diag[active_adapter])
+        
+        x = module._cast_input_dtype(x, lora_A.weight.dtype)
+        if isinstance(dropout, nn.Identity) or not module.training:
+            x = dropout(x)
+
```
BenjaminBossan: These lines can be removed, they're already being called by the normal `forward` of the LoRA layer.

BenjaminBossan (changes requested):
Thanks for the updates, we're making good progress here.

> The logic for raising errors in each layer hasn’t been applied yet, but I committed first to check whether adding the parameter in this way matches what you meant. Excluding the Linear class, it seems like an error should be raised when use_kasa is true in the other classes.

Nice, this looks correct, please raise the error in the unsupported layers as indicated.

> Also, I noticed there’s no part that calls KasaLinearVariant—should this be called inside the linear class? I’m a bit confused about this part.

Don't worry, it is being called. E.g. in `lora.Linear.forward` we have this code:

https://github.com/huggingface/peft/blob/190f9873b15660d9092f70065c18e4993fe10d5b/src/peft/tuners/lora/layer.py#L806-L816

So if the KaSA variant is found, `KasaLinearVariant.forward` will be used here. Same for the other methods.

iambogeumkim:
@BenjaminBossan 
I removed unnecessary parts and added the `_get_delta_weight` static method specifically for the `KasaLinearVariants` class. If adding `_get_delta_weight` in this way is correct, I plan to update all occurrences of `module.get_delta_weight` within the current `KasaLinearVariants` class to use the modified method. Please review and share your thoughts.

src/peft/tuners/lora/variants.py
```diff
@@ -439,6 +439,64 @@ def init(module: Conv3d, adapter_name: str, **kwargs: Any) -> None:
...
+        return orig_weight + delta_weight
+
+    @staticmethod
+    def merge_unsafe(module: Linear, active_adapter: str, orig_weight: torch.Tensor) -> None:
+        delta_weight = module.get_delta_weight(active_adapter)
+        orig_weight.data += delta_weight
+
+    @staticmethod
+    def unmerge(module: Linear, active_adapter: str, orig_weight: torch.Tensor) -> torch.Tensor:
+        delta_weight = module.get_delta_weight(active_adapter)
```
BenjaminBossan: Instead of `module.get_delta_weight`, these methods should call the newly added `KasaLinearVariant._get_delta_weight`, right?

iambogeumkim:
I've made the changes you commented on and added the following to the test file.  I used the name 'Vanilla MLP LoRA with KaSA' for the lines added to the test file, which is different from what you suggested. Please let me know if this wasn't your intention.

Also, I have a  question: in forward method, 
```@staticmethod
def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.Tensor) -> torch.Tensor:
  lora_A = module.lora_A[active_adapter]
  lora_B = module.lora_B[active_adapter]
  dropout = module.lora_dropout[active_adapter]
  scaling = module.scaling[active_adapter]
  diag = torch.diag(module.lora_diag[active_adapter])

  # KASA calculation
  # see https://github.com/juyongjiang/KaSA/blob/f85e88c22d0fa4cb8ab2923d7c2bf1bbec152da3/peft/src/peft/tuners/lora/layer.py#L602C21-L602C110
  lora_output = lora_B(torch.einsum("ijk,kl->ijl", lora_A(x), diag)) * scaling
  return result + lora_output
```
`dropout` variable isn't being used. Should I remove it or just leave it as is?

BenjaminBossan:
> `dropout` variable isn't being used. Should I remove it or just leave it as is?

Good catch. I think the correct way is to apply it to `x`, like in the original code: https://github.com/juyongjiang/KaSA/blob/f85e88c22d0fa4cb8ab2923d7c2bf1bbec152da3/peft/src/peft/tuners/lora/layer.py#L602C21-L602C110

iambogeumkim:
I've added up to this point. Please share your thoughts if there are any further modifications needed in the work so far, and we can discuss the next steps for the testing phase.

tests/test_custom_models.py
```diff
@@ -159,6 +159,8 @@
         LoraConfig,
         {"target_modules": ["lin0"], "target_parameters": ["lin1.weight"]},
     ),
+    ("Vanilla MLP 7 LoRA with KaSA", "MLP", LoraConfig, {"target_modules": ["lin0"], "use_kasa": True}),
+    ("Vanilla MLP 8 LoRA with KaSA", "MLP", LoraConfig, {"target_modules": ["lin0", "lin1"], "use_kasa": True})
```
```suggestion
    ("Vanilla MLP 8 LoRA with KaSA", "MLP", LoraConfig, {"target_modules": ["lin0", "lin1"], "use_kasa": True}),
```

src/peft/tuners/lora/layer.py
```diff
@@ -819,8 +825,6 @@ def get_delta_weight(self, adapter) -> torch.Tensor:
             weight_A = weight_A.float()
             weight_B = weight_B.float()
 
-        output_tensor = transpose(weight_B @ weight_A, self.fan_in_fan_out) * self.scaling[adapter]
```
BenjaminBossan, line 780: Please restore this line.

src/peft/tuners/lora/layer.py
```diff
@@ -113,6 +113,7 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
         self._disable_adapters = False
         self.merged_adapters = []
         self.use_dora: dict[str, bool] = {}  # not actively used anymore after #2443, keep it for BC
+        self.use_kasa: dict[str, bool] = {}
```
BenjaminBossan: Can be removed

src/peft/tuners/lora/variants.py
```diff
@@ -439,6 +439,82 @@ def init(module: Conv3d, adapter_name: str, **kwargs: Any) -> None:
         _DoraConvNdVariant.init_convd_variant(module, adapter_name, dora_layer=dora_layer)
 
 
+class KasaLinearVariant(LoraVariant):
+    @staticmethod
+    def init(module: Linear, adapter_name: str, **kwargs: Any) -> None:
+        if not module.lora_diag:
```
BenjaminBossan:
```suggestion
        if not hasattr(module, "lora_diag"):
            module.lora_diag = nn.ParameterDict()
```

src/peft/tuners/lora/variants.py
```diff
@@ -439,6 +439,82 @@ def init(module: Conv3d, adapter_name: str, **kwargs: Any) -> None:
...
+            module.lora_A[active_adapter].weight,
+            module.lora_B[active_adapter].weight,
+            module.lora_diag[active_adapter],
+            module.scaling[active_adapter],
+            module.fan_in_fan_out,
+        )
+        return orig_weight - delta_weight
+
+    @staticmethod
+    def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.Tensor) -> torch.Tensor:
```
BenjaminBossan:
```suggestion
    def forward(module: Linear, active_adapter: str, x: torch.Tensor, result: torch.Tensor, **kwargs) -> torch.Tensor:
```

src/peft/tuners/lora/variants.py
```diff
@@ -439,6 +439,82 @@ def init(module: Conv3d, adapter_name: str, **kwargs: Any) -> None:
...
+        lora_A = module.lora_A[active_adapter]
+        lora_B = module.lora_B[active_adapter]
+        dropout = module.lora_dropout[active_adapter]
+        scaling = module.scaling[active_adapter]
+        diag = torch.diag(module.lora_diag[active_adapter])
+
+        # KaSA calculation
+        # see https://github.com/juyongjiang/KaSA/blob/f85e88c22d0fa4cb8ab2923d7c2bf1bbec152da3/peft/src/peft/tuners/lora/layer.py#L602C21-L602C110
+        lora_output = lora_B(torch.einsum("ijk,kl->ijl", lora_A(dropout(x)), diag)) * scaling
```
BenjaminBossan: I noticed that the current KaSA implementation assumes 3d input (e.g. for language modeling) and thus breaks for 2d input. This extension should fix that.

```suggestion
        if x.ndim == 3:
            lora_output = lora_B(torch.einsum("ijk,kl->ijl", lora_A(dropout(x)), diag)) * scaling
        elif x.ndim == 2:
            lora_output = lora_B(lora_A(dropout(x)) @ diag) * scaling
        else:
            raise ValueError(f"Using KaSA with inputs of shape {x.ndim} is not supported, only 2 or 3 dims.")
```

src/peft/tuners/lora/layer.py
iambogeumkim: done

iambogeumkim:
I have a question after local testing. I'll briefly share the test results and process.

### 1. 10 failures occurred
- 8 dtype mismatch issues, 2 AssertionError

### 2. dtype issues (all resolved ✅)
- When testing cases where the model's data type changes, like in the `test_forward_bfloat16` method, dtype mismatch issues occurred. 
- This was resolved by adding a part to match the dtype in a method within the `KasaLinearVariants` class.

### 3. AssertionError issues
- An assert False error occurred in the `test_disable_adapters` method.
  image

- So, I added the following conditional statement to the forward method so that `outputs_before == outputs_disabled` holds True in the `test_disable_adapters`, by skipping KaSA calculation and returning only the base model output when adapters are disabled.
  ```Python
  # Check if adapters are disabled
  if module.disable_adapters:
    return result
  ```
- However, assert False is still occurring.

But now that I think about it, looking at the current `KasaLinearVariants` class, it applies SVD to the weights and calculates from the init method. So I'm wondering if assert False is actually correct, or if I just tried to solve it in the wrong way.

iambogeumkim:
Sorry for the delayed update; I've been working on this in bits and pieces due to personal scheduling. Since some time has passed, I'll summarize the feedback you previously gave me and my related work:

## 1. `_skip_test_disable_adapters` function
- Added skip logic for test functions that do not work with KaSA adapters. (you can check in `tests/test_custom_models.py`)
## 2. `TestKasaInitialization` class
- `test_kasa_mixed_adapter_error` method: When existing adapters are not KaSA, adding a new KaSA adapter should raise a ValueError.
- `test_kasa_mixed_adapter_error_reverse` method: When existing adapters are KaSA, adding a non-KaSA adapter should raise a ValueError.

Both of these test functions passed because they successfully raised a ValueError. My understanding is that "PASSED" here means the ValueError was raised, which is the intended behavior for these functions. Please correct me if my understanding is wrong.
## 3. Logic to avoid repetitive SVD processes (Just an idea 💡)
I agree with your opinion that repeatedly modifying the base weights for the same KaSA adapter is not ideal. My current idea is to apply a cache, as shown below, and check if SVD has already been applied by checking the existence of this attribute. What do you think of this approach?
```python
if not hasattr(module, "_kasa_svd_cache"):
    # First KaSA adapter: perform SVD and cache the result
    weight = module.get_base_layer().weight
    dtype = weight.dtype
    weight = weight.to(torch.float32)
    U, S, Vh = torch.linalg.svd(weight.data, full_matrices=False)
    module._kasa_svd_cache = (U, S, Vh)
else:
    # Reuse cached SVD result
    U, S, Vh = module._kasa_svd_cache
    dtype = module.get_base_layer().weight.dtype
```

src/peft/tuners/lora/model.py
```diff
@@ -158,6 +158,43 @@ def _check_new_adapter_config(self, config: LoraConfig) -> None:
                 f"{self.__class__.__name__} supports only 1 adapter with bias. When using multiple adapters, "
                 "set bias to 'none' for all adapters."
             )
+        
+        # Check KaSA adapter compatibility (only when adding additional adapters)
```
BenjaminBossan, line 170: I think this logic can be simplified. In the end, what we want to do is count the number of KaSA vs non-KaSA configs and if they're both greater than 0, raise an error (one error message for both that says that they cannot be mixed is fine).

Also, I think there is no need for `getattr(config, 'use_kasa', False)`, using `config.use_kasa` should be fine, since the attribute is added to the `LoraConfig` and should always exist.

tests/test_initialization.py
```diff
@@ -4477,3 +4477,67 @@ def test_key_mapping_save_old_load_new_vblora(self, old_model, new_model, tmp_pa
     def test_key_mapping_save_new_load_old_vblora(self, old_model, new_model, tmp_path):
         # save the new model, load it into the old model, should work without issues (forwards compatibility)
         self.check_vblora_load_no_warning(new_model, old_model, tmp_path)
+
+
+class TestKasaInitialization:
```
BenjaminBossan: Could you please move the test functions to the bottom of `class TestLoraInitialization`? This is because KaSA is a LoRA variant, and it also means we can reuse the existing `get_model` method.

tests/test_initialization.py
```diff
@@ -4477,3 +4477,67 @@ def test_key_mapping_save_old_load_new_vblora(self, old_model, new_model, tmp_pa
...
+                # choose a large weight so that averages are close to expected values
+                self.linear = nn.Linear(1000, 1000)
+
+            def forward(self, x):
+                return self.linear(x)
+
+        return MyModule().eval().to(self.torch_device)
+
+    @pytest.fixture
+    def data(self):
```
BenjaminBossan: This fixture is not needed, as `data` is not used in the tests.

BenjaminBossan (changes requested):
Thanks for the updates, we're moving towards the goal. I commented on some of the changes, please check.

> My current idea is to apply a cache, as shown below, and check if SVD has already been applied by checking the existence of this attribute. What do you think of this approach?

I'm not sure it's needed, but I could be wrong. AFAICT, the SVD is only needed to update the base weight. If we add a second KaSA adapter, the base weight is already updated, so there is no need to cache `U, S, Vh`, we can just skip completely. Is my understanding correct?

Also, there are some merge conflicts on this PR now. I think they should be easy to resolve, but don't hesitate to ask if you have questions. Finally, before committing, don't forget to call `make style`.

iambogeumkim:
Check

iambogeumkim:
@BenjaminBossan

I've addressed the points you mentioned, applied `make style`, and resolved the conflicts. Let me know if anything else needs to be updated.

Regarding the SVD value caching, I gave it some thought and realized I was stuck on the idea that 'caching is always efficient.' Since the base weights are already updated in the first adapter even when using multiple KaSA adapters, I realized we can simply reuse those values subsequently. So, I modified the code to skip the calculation as you suggested.

tests/test_initialization.py
```diff
@@ -1482,6 +1482,47 @@ def test_loading_loading_adapters_with_target_parameters_raises(self, tmp_path):
...
 
+    def test_kasa_mixed_adapter_error(self):
+        """Test that KaSA adapters cannot be mixed with other adapter types."""
+        model = self.get_model()
+        
+        # Add regular LoRA adapter first
+        config1 = LoraConfig(
+            r=8,
+            target_modules=["linear"],
+            init_lora_weights=True,
```
BenjaminBossan: You can remove this line, as it's irrelevant.

tests/test_initialization.py
```diff
@@ -1482,6 +1482,47 @@ def test_loading_loading_adapters_with_target_parameters_raises(self, tmp_path):
...
+            target_modules=["linear"],
+            use_kasa=True,
+        )
+        model = get_peft_model(model, config1, adapter_name="kasa1")
+        
+        # Try to add regular LoRA adapter - should raise error
+        config2 = LoraConfig(
+            r=16,
+            target_modules=["linear"],
+            init_lora_weights=True,
```
BenjaminBossan: You can remove this line, as it's irrelevant.

iambogeumkim:
I applied what you mentioned and resolvd conflicts. Please take a look!

tests/test_initialization.py
iambogeumkim:
```python
# src/peft/tuners/lora/model.py
if len(self.peft_config) > 1:
  kasa_count = sum(1 for cfg in self.peft_config.values() if cfg.use_kasa)
  non_kasa_count = len(self.peft_config) - kasa_count
  
  if kasa_count > 0 and non_kasa_count > 0:
    raise ValueError("KaSA adapters cannot be mixed with other adapter types.")
```

I understood this to mean that since it's handled in this section, it's irrelevant elsewhere. Is my understanding correct?

BenjaminBossan: Oh, this was a misunderstanding. I meant that the single line I commented on (`init_lora_weights=True,`) can be removed, the test as a whole is good to keep :) Please restore these tests.

iambogeumkim: ah okay haha

iambogeumkim: I changed the tests back :) !

src/peft/tuners/lora/model.py
```diff
@@ -155,6 +155,29 @@ class LoraModel(BaseTuner):
...
+        Raise a ValueError if there is something wrong with the config or if it conflicts with existing adapters.
+
+        """
+        # TODO: there should be a check if any of the existing adapters actually has bias != "none", or else the check
+        # does not fully correspond to the error message.
+        if (len(self.peft_config) > 1) and (config.bias != "none"):
+            raise ValueError(
+                f"{self.__class__.__name__} supports only 1 adapter with bias. When using multiple adapters, "
+                "set bias to 'none' for all adapters."
+            )
```
BenjaminBossan: Let's remove this and call `super()._check_new_adapter_config(config)` instead.

iambogeumkim:
Is this the final step? Please let me know if there's anything else needed.

BenjaminBossan:
@iambogeumkim Could you please run `make style`?

iambogeumkim:
I did run `make style`, but it looks like one file was missed. I re-ran it and pushed the commit.

Also, thank you for your patience with all my questions, even the trivial ones. I know I might have been a bit of a bother 😅 Since this was my first code contribution, I learned so much thanks to your guidance. Wishing you a warm and happy holiday season!

BenjaminBossan:
> I did run `make style`, but it looks like one file was missed. I re-ran it and pushed the commit.

Something doesn't seem to work right, as the formatter is still complaining. These changes should resolve it:

```diff
modified   src/peft/tuners/lora/config.py
@@ -764,8 +764,9 @@ class LoraConfig(PeftConfig):
                 "singular value decomposition (SVD) with knowledge-aware singular values to dynamically "
                 "activate parametric knowledge according to its relevance to downstream tasks."
             )
-        }
+        },
     )
+
     def to_dict(self):
         """
         Returns the configuration for your adapter model as a dictionary. Removes runtime configurations.
modified   tests/test_custom_models.py
@@ -1265,10 +1265,12 @@ def _skip_tests_with_multiple_adapters_with_target_parameters(config_cls, config
     if (config_cls == LoraConfig) and config_kwargs.get("target_parameters"):
         pytest.skip("LoRA with multiple adapters with target_parameters is not supported")
 
+
 def _skip_test_disable_adapters(config_cls, config_kwargs):
     if (config_cls == LoraConfig) and config_kwargs.get("use_kasa"):
         pytest.skip("KaSA modifies base weights, so adapter disable test is skipped")
 
+
 class MLP(nn.Module):
     def __init__(self, bias=True):
         super().__init__()
```

> Also, thank you for your patience with all my questions, even the trivial ones. I know I might have been a bit of a bother 😅 Since this was my first code contribution, I learned so much thanks to your guidance. Wishing you a warm and happy holiday season!

Don't worry, it's always the first time for someone. Happy to hear that you learned a lot.

iambogeumkim:
I double-checked if there were any unpushed files related to KaSA. Aside from those two files, everything seems to be pushed, so it should be ready to be merged now.

BenjaminBossan:
@iambogeumkim There are a bunch of failing tests because `Embedding.update_layer` and `_ConvNd.update_layer` need to be passed the `use_kasa` argument. For this, you need to update their `__init__` methods. Could you please update those? Once you finish, you can run `pytest tests/ -k kasa` to check locally if the tests now pass.

iambogeumkim:
I’ve updated `layer.py` as you suggested, and I’ve confirmed that all local tests are passing. I’ve also run the `make style` command. I hope everything looks good now, but please let me know if there’s anything else I should address.

BenjaminBossan:
Thanks for the latest changes. There are still some errors, this time caused by X-LoRA. I checked and the issue there is that X-LoRA models can have PEFT configs that contain both normal LoRA and X-LoRA configs. Since X-LoRA configs don't have `.use_case`, this check fails:

`kasa_count = sum(1 for cfg in self.peft_config.values() if cfg.use_kasa)`

It's a bit of an edge case, but let's add `if isinstance(cfg, LoraConfig)` and it should work.

githubnemo:
*not stale*

PR #2701, opened 2025-08-04
Use `hub_online_once` in trainable token tests

githubnemo:
Also fix a minor import nit where `TrainableTokensWrapper` was not added to `utils/__init__.py`. Fixed the corresponding imports as well.

Another housekeeping job is to move `hub_online_once` to `testing_utils.py` since it has grown to be used in a lot of places and `testing_utils.py` is the better place to keep such utilities.


BenjaminBossan (approved):
Thanks for extending the usage and clean ups, LGTM and can be merged.

Just a nit: Do you want to document the potential issue with yield fixtures and `hub_online_once` directly in the `hub_online_once` docs?

Issue #2557, opened 2025-05-29
ValueError during evaluation when using label_names=['labels']


Cryptocho:
### System Info
peft==0.15.2
transformers==4.52.3

### Reproduction
~~~python
# init model
from transformers import AutoModelForCausalLM, AutoTokenizer, GenerationConfig, BitsAndBytesConfig
from peft import get_peft_model, PolyConfig, TaskType
import torch

peft_config = PolyConfig(
    r=32,
    target_modules=["q_proj", "k_proj", "v_proj", "o_proj"], 
    n_tasks=2,
    n_skills=2,
    n_splits=4,
    task_type="CAUSAL_LM"
)


generation_config = GenerationConfig(
    temperature=0.7,
    top_k=20,
    top_p=0.8,
    min_p=0,
    num_beams=4,
    do_sample=True
)

bnb_config = BitsAndBytesConfig(
    load_in_4bit=True,
    bnb_4bit_quant_type="nf4",
    bnb_4bit_use_double_quant=False,
    bnb_4bit_compute_dtype=torch.float16,
)

model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen3-4B", quantization_config=bnb_config)
model.generation_config = generation_config
tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen3-4B")

from peft import prepare_model_for_kbit_training, get_peft_model

model = prepare_model_for_kbit_training(model, use_gradient_checkpointing=False)
model = get_peft_model(model, peft_config) # peft.peft_model.PeftModelForCausalLM``

# proc data

from datasets import load_dataset

data = load_dataset("REILX/Modern-Chinese-to-Classical-Chinese")
data = data['train']
data = data.filter(lambda example: '《' not in example['input']) 

def apply_template(message, role):
    return [{"role": role, "content": message}]

def preprocess_cls2anc(example):
    result={}
    result['input'] = tokenizer.apply_chat_template(apply_template(example['input'][example['input'].index('：'):], 'user'), tokenize=False, add_generation_prompt=True, enable_thinking=False)
    result['labels'] = example['output'][example['output'].index('：'):] + "<|im_end|>\n"
    result['task_ids'] = 1
    return result

def tokenize(example):
    result = tokenizer(example['input'], return_tensors='pt', padding="max_length", max_length=768)
    result['labels'] = tokenizer(example['labels'], return_tensors='pt', padding="max_length", max_length=768)['input_ids']
    for key in result.keys():
[...]

PR #2702, opened 2025-08-04
validated POLY on XPU, pass

yao-matrix:
@BenjaminBossan, pls help review, thx very much

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -945,7 +945,7 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
         """
         Returns the base model.
         """
-        return self.base_model if (self.active_peft_config.is_prompt_learning) else self.base_model.model
+        return self.base_model if self.active_peft_config.is_prompt_learning else self.base_model.model
 
     def add_adapter(self, adapter_name: str, peft_config: PeftConfig, low_cpu_mem_usage: bool = False) -> None:
         """
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -945,11 +945,7 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
         """
         Returns the base model.
         """
-        return (
-            self.base_model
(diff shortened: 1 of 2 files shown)


src/peft/peft_model.py
```diff
@@ -945,11 +945,7 @@ def get_base_model(self) -> torch.nn.Module:
         """
         Returns the base model.
         """
-        return (
-            self.base_model
-            if (self.active_peft_config.is_prompt_learning or self.peft_type == PeftType.POLY)
-            else self.base_model.model
-        )
+        return self.base_model if (self.active_peft_config.is_prompt_learning) else self.base_model.model
```
yao-matrix: this is to fix `TypeError: T5ForConditionalGeneration.forward() got an unexpected keyword argument 'num_items_in_batch'` during `trainer.train()` of the notebook. 
The reason is the original logic will return `PolyModel` whose `accepts_loss_kwargs` is `True`, but real base model should be `T5ForConditionalGeneration` which is `self.base_model.model` and its `accepts_loss_kwargs` is `False`. So the trainer wrongly treated the base model can `accepts_loss_kwargs`(code as [here](https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py#L643)) and lead to the error.

But i don't know the background why just return `self.base_model` in `POLY` here, @BenjaminBossan so need your comments on whether it's a proper fix or not, thx very much.

BenjaminBossan: Thanks for investigating. Regarding your conclusion, see [my comment in this thread](https://github.com/huggingface/peft/issues/2557#issuecomment-2929985608). So I suspect this extra handling of Poly can indeed just be removed.

```suggestion
        return self.base_model if self.active_peft_config.is_prompt_learning else self.base_model.model
```

yao-matrix: OK, then let's do it.

BenjaminBossan (approved):
Thank you for updating the Poly example, validating it with XPU, and also fixing this small Poly issue we discussed.

PR #2703, opened 2025-08-04
validated PISSA on xpu, pass

yao-matrix:
@BenjaminBossan, pls help review, thx very much


BenjaminBossan (approved):
Thanks for fixing these issues in the PiSSA example and validating it for XPU.

PR #2706, opened 2025-08-05
fix bug for feature_extraction example

kaixuanliu:
Here we need to pin datasets to an old version to load the datasets in example. Otherwise it will crash in [L267](https://github.com/huggingface/peft/blob/main/examples/feature_extraction/peft_lora_embedding_semantic_search.py#L267)


kaixuanliu:
@BenjaminBossan The `peft_model_id`  in ipynb file `smangrul/peft_lora_e5_semantic_search` does not exist, maybe we can delete the file or use another model, WDYT?

BenjaminBossan (approved):
Thanks for fixing this bug in the example and updating the requirements.

PR #2709, opened 2025-08-05
Add XPU support for sft example

kaixuanliu:


kaixuanliu:
@BenjaminBossan Hi, right now DeepSpeed for XPU depends on IPEX which we will deprecate soon. And to make it work on XPU, we need to make small adjustment to DeepSpeed code, hence I suggest do not add this to the doc. And I can delete the `deepspeed_xpu_config.yaml` file. WDYT?

BenjaminBossan:
If it's deprecated, I agree that it's better to remove it for now.

BenjaminBossan:
@kaixuanliu Could you please run `make style`?

kaixuanliu:
@BenjaminBossan Done.

BenjaminBossan (approved):
Thanks for making the SFT example work with XPU.

PR #2712, opened 2025-08-06
verified mlp on xpu, pass

yao-matrix:


BenjaminBossan (approved):
Thanks for updating this notebook and making it work with XPU, LGTM.

PR #2727, opened 2025-08-07
validated token classification on xpu, pass

yao-matrix:


BenjaminBossan (approved):
Thanks for fixing the token classification/NER examples and adding support for XPU.

PR #2728, opened 2025-08-07
extend docs to cover more accelerators like intel XPU

yao-matrix:
1. extend docs to cover more accelerators like intel XPU
2. fix some places in examples where xpu enabling missed

@BenjaminBossan , pls help review, thx very much


BenjaminBossan (approved):
Thanks for helping to make the docs more device agnostic and for adjusting some more examples to work with XPU.

PR #2729, opened 2025-08-08
enable xpu for train_memory script

yao-matrix:


BenjaminBossan (approved):
Thanks for updating the `train_memory.py` script to work with XPU.

PR #2730, opened 2025-08-08
extend text-generation-benchmark to xpu, pass

yao-matrix:
@BenjaminBossan, pls help review, thx very much.

### DIFF
--- a/method_comparison/text_generation_benchmark/utils.py
+++ b/method_comparison/text_generation_benchmark/utils.py
@@ -344,7 +344,7 @@ def get_memory_usage() -> tuple[float, float, float]:
         accelerator_reserved = torch.cuda.memory_reserved()
         accelerator_allocated_mb = accelerator_allocated / (1024 * 1024)
         accelerator_reserved_mb = accelerator_reserved / (1024 * 1024)
-    if torch.xpu.is_available():
+    elif torch.xpu.is_available():
         accelerator_allocated = torch.xpu.memory_allocated()
         accelerator_reserved = torch.xpu.memory_reserved()
         accelerator_allocated_mb = accelerator_allocated / (1024 * 1024)
@@ -363,7 +363,7 @@ def init_accelerator() -> tuple[float, float]:
         torch.cuda.empty_cache()
         _, accelerator_allocated, accelerator_reserved = get_memory_usage()
         return accelerator_allocated, accelerator_reserved
-    if torch.xpu.is_available():
+    elif torch.xpu.is_available():
         torch.xpu.init()
         torch.xpu.empty_cache()
         _, accelerator_allocated, accelerator_reserved = get_memory_usage()
--- a/method_comparison/text_generation_benchmark/utils.py
+++ b/method_comparison/text_generation_benchmark/utils.py
@@ -363,10 +363,11 @@ def init_accelerator() -> tuple[float, float]:
         torch.cuda.empty_cache()
         _, accelerator_allocated, accelerator_reserved = get_memory_usage()
         return accelerator_allocated, accelerator_reserved
-    elif torch.xpu.is_available():
+    if torch.xpu.is_available():
         torch.xpu.init()
         torch.xpu.empty_cache()
         _, accelerator_allocated, accelerator_reserved = get_memory_usage()
+        return accelerator_allocated, accelerator_reserved
     return 0.0, 0.0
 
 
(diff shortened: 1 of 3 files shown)


method_comparison/text_generation_benchmark/utils.py
```diff
@@ -331,30 +334,39 @@ def validate_experiment_path(path: str) -> tuple[str, "BenchmarkConfig"]:
...
+        accelerator_reserved_mb = 0.0
 
-    return ram_usage_mb, gpu_allocated_mb, gpu_reserved_mb
+    return ram_usage_mb, accelerator_allocated_mb, accelerator_reserved_mb
 
 
-def init_cuda() -> tuple[float, float]:
-    """Initialize CUDA and return initial memory usage."""
+def init_accelerator() -> tuple[float, float]:
+    """Initialize acceleratot and return initial memory usage."""
```
BenjaminBossan:
```suggestion
    """Initialize accelerator and return initial memory usage."""
```

method_comparison/text_generation_benchmark/utils.py
```diff
@@ -331,30 +334,39 @@ def validate_experiment_path(path: str) -> tuple[str, "BenchmarkConfig"]:
...
+        accelerator_allocated_mb = accelerator_allocated / (1024 * 1024)
+        accelerator_reserved_mb = accelerator_reserved / (1024 * 1024)
     else:
-        gpu_allocated_mb = 0.0
-        gpu_reserved_mb = 0.0
+        accelerator_allocated_mb = 0.0
+        accelerator_reserved_mb = 0.0
 
-    return ram_usage_mb, gpu_allocated_mb, gpu_reserved_mb
+    return ram_usage_mb, accelerator_allocated_mb, accelerator_reserved_mb
```
BenjaminBossan, line 356: For some reason, I'm getting incorrect results. When I run `python run_base.py -v --force` on this branch on a 4090, I get 0 MB memory usage in the report. To debug this, you can set a breakpoint at this line and run the aforementioned command. The first time we get here, `accelerator_allocated_mb` and `accelerator_reserved_mb` are 0, which is expected. After this, the model is loaded to the accelerator. Thus, when we get here the second time, these values should be > 0, but I still get 0 here. When running the same on the main branch, I get correct reports.

I could not determine where this difference comes from, `nvidia-smi` shows the same output for main branch and this branch, `torch.cuda.is_available()` always returns `True`, and the `model_kwargs` are also identical. Can you replicate this and do you have an idea what the issue could be?

BenjaminBossan (changes requested):
Thanks for updating the text generation benchmark, this was definitely on our list to do soon. Unfortunately, I ran into an issue when trying it out. Could you please check?

method_comparison/text_generation_benchmark/utils.py
yao-matrix: @BenjaminBossan oh, my fault, the chain should be `if-elif-else`, but i break to 2(`if` and `if-else`), so CUDA GPU goes into first `if` and second `else`(which gives it 0), sorry and thx for checking

BenjaminBossan: Ah yes, makes sense.

method_comparison/text_generation_benchmark/utils.py
```diff
@@ -331,30 +334,39 @@ def validate_experiment_path(path: str) -> tuple[str, "BenchmarkConfig"]:
...
+def init_accelerator() -> tuple[float, float]:
+    """Initialize accelerator and return initial memory usage."""
     if torch.cuda.is_available():
         torch.cuda.init()
         torch.cuda.empty_cache()
-        _, gpu_allocated, gpu_reserved = get_memory_usage()
-        return gpu_allocated, gpu_reserved
+        _, accelerator_allocated, accelerator_reserved = get_memory_usage()
+        return accelerator_allocated, accelerator_reserved
+    elif torch.xpu.is_available():
```
BenjaminBossan, line 365: Here, `elif` is not necessary because the previous branch returns, but it also doesn't hurt.

method_comparison/text_generation_benchmark/utils.py
```diff
@@ -331,30 +334,39 @@ def validate_experiment_path(path: str) -> tuple[str, "BenchmarkConfig"]:
...
+        torch.xpu.init()
+        torch.xpu.empty_cache()
+        _, accelerator_allocated, accelerator_reserved = get_memory_usage()
```
BenjaminBossan, line 368: These values are not returned, so if XPU is used, the next line will return `0.0, 0.0`.

BenjaminBossan (changes requested):
Thanks for fixing the issue with CUDA. I think the script still has a small error for XPU, which I flagged, please check.

method_comparison/text_generation_benchmark/utils.py
yao-matrix: done, thx very much.

BenjaminBossan (approved):
Thanks for making the text generation benchmark XPU compatible.

PR #2732, opened 2025-08-08
add xpu support for sequence_classification example

kaixuanliu:


BenjaminBossan (approved):
Thanks for updating the sequence classification examples to be device agnostic.

PR #2735, opened 2025-08-12
FIX: DynamicCache max_cache_len attribute error

BenjaminBossan:
Resolves [current CI errors with prefix tuning](https://github.com/huggingface/peft/actions/runs/16904314494/job/47890407632).

Due to some recent changes in transformers (surfaced by https://github.com/huggingface/transformers/pull/39797), checking `hasattr(cache, "max_cache_len")` results in an error:

```python
>>> cache = DynamicCache()
>>> hasattr(cache, "max_cache_len")
  File "/home/name/work/forks/transformers/foo.py", line 9, in <module>
    hasattr(cache, "max_cache_len")
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/name/work/forks/transformers/src/transformers/cache_utils.py", line 916, in max_cache_len
    return max(values)
           ^^^^^^^^^^^
ValueError: max() iterable argument is empty
```

This has been reported and will be fixed in transformers. On the PEFT side, it is safest check the cache type and avoid accessing this attribute in the first place, which is what this PR does.

Moreover, that PR also changed the argument order to initialize `HybridCache`, which is also taken into account in this PR by only using keyword arguments.

**Update**: This PR also [takes care](https://github.com/huggingface/peft/pull/2735/commits/dc50a16faef4995e6dfa7de87aca795f40e782fa) of a [recent change in transformers](https://github.com/huggingface/transformers/pull/40135) that changed the cache type for Gemma to `DynamicCache` from `HybridCache`.

Since the PR CI uses the transformers release, it won't show the error, but I tested locally and the failing tests pass.


Cyrilvallez (approved):
Yup, should do the charm!

src/peft/peft_model.py
```diff
@@ -2068,15 +2076,18 @@ def prepare_inputs_for_generation(self, *args, task_ids: Optional[torch.Tensor]
                 )
                 kwargs["token_type_ids"] = None
 
+            cache: transformers.Cache | None = model_kwargs.get("past_key_values", None)
```
githubnemo, line 2083:
```suggestion
            cache: Optional[transformers.Cache] = model_kwargs.get("past_key_values", None)
```

since the `|` syntax is not compatible with python 3.9 (or we need a future import)

src/peft/peft_model.py
```diff
@@ -749,22 +749,30 @@ def get_prompt(
                 post_process_fn = TRANSFORMERS_MODELS_TO_PREFIX_TUNING_POSTPROCESS_MAPPING[self.config.model_type]
                 past_key_values = post_process_fn(past_key_values)
             elif ("gemma2" in model_type) or ("gemma3_text" in model_type):
+                # TODO: remove this logic once transformers < 4.56 is dropped
+                transformers_le_4_55 = packaging.version.parse(transformers.__version__) <= packaging.version.parse(
+                    "4.55.2"
+                )
```
githubnemo: I agree it would make more sense to make it `lt_4_60` except I'm overlooking something.

BenjaminBossan: Changed to `transformers_lt_4_56`.

src/peft/peft_model.py
BenjaminBossan: This should work, since this file has the `__future__` import. Also, CI includes Python 3.9, so we would get an error if it didn't.

PR #2736, opened 2025-08-13
CHORE: Clean up config kwargs in custom model tests

### LINKED ISSUE #2695: TST: Some SHiRA tests are failing when run with parallelism
### System Info

PEFT 0.17.0

### Who can help?

_No response_

### Reproduction

When running

`pytest tests/test_custom_models.py -k shira -n2`

(or with higher parallelism), some SHiRA tests fail locally. Which ones depends on the level of parallelism. This is true with CPU and GPU. I'm not sure why the CI doesn't pick it up.

@kkb-code can you reproduce that?

### Expected behavior

Tests should pass.

BenjaminBossan:
Resolves #2695

For some PEFT methods, there was a bit of a mess when it comes to how the `init_weights` argument was set in `test_custom_models.py`. The default `kwargs` for the tests should be that the PEFT method is initialized as an identity transform, and for specific tests we want to disable that. Note that most PEFT methods are initialized by default to be identity transforms, which is why the argument does not need to be set explicitly, but it's not true for all PEFT methods.

With this PR, SHiRA, C3A, and FourierFT are now initialized to be consistent with this. This made it possible to remove some extra handling of those methods which was intermingled with certain tests.

Moreover, in the appropriate places, I did the following:

- use the `set_init_weights_false` helper function
- cleaned up a bit the docs for the `init_weights` arguments of these PEFT methods
- added clarifying comments
- for `test_unload_adapter`, simplified a config type check and rewrote it to load the base model only once

### DIFF
--- a/src/peft/tuners/c3a/config.py
+++ b/src/peft/tuners/c3a/config.py
@@ -46,8 +46,8 @@ class C3AConfig(PeftConfig):
             The mapping from layer names or regexp expression to block_size which are different from the default
             specified. For example, `{"model.decoder.layers.0.encoder_attn.k_proj": 1280`}
         init_weights (`Union[bool, Literal["gaussian", "kaiming_uniform", "xavier_uniform"]]`):
-            The initialization of the C3A weights. Set this to False if the weights should be initialized to a commonly
-            used distribution. Set this to True if the weights should be initialized to zeros.
+            Defaults to 'xavier_uniform'. Setting this to `False` also usese 'xavier_uniform'. To set the weights to
+            zeros (thus making C3A a no-op), set the value to `False`.
     """
 
     block_size: int = field(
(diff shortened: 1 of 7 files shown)


src/peft/tuners/c3a/config.py
```diff
@@ -46,8 +46,8 @@ class C3AConfig(PeftConfig):
             The mapping from layer names or regexp expression to block_size which are different from the default
             specified. For example, `{"model.decoder.layers.0.encoder_attn.k_proj": 1280`}
         init_weights (`Union[bool, Literal["gaussian", "kaiming_uniform", "xavier_uniform"]]`):
-            The initialization of the C3A weights. Set this to False if the weights should be initialized to a commonly
-            used distribution. Set this to True if the weights should be initialized to zeros.
+            Defaults to 'xavier_uniform'. Setting this to `False` also uses 'xavier_uniform'. To set the weights to
+            zeros (thus making C3A a no-op), set the value to `False`.
```
githubnemo:
```suggestion
            Defaults to 'xavier_uniform'. Setting this to `False` also uses 'xavier_uniform'. To set the weights to
            zeros (thus making C3A a no-op), set the value to `True`.
```

src/peft/tuners/c3a/config.py
```diff
@@ -116,10 +116,8 @@ class C3AConfig(PeftConfig):
         default="xavier_uniform",
         metadata={
             "help": (
-                "The initialization of the C3A weights. Leave it as default or"
-                " set it to False if the weights should be initialized with Xavier uniform,"
-                " which is experimentally suitable for C3A."
-                " Set this to True if the weights should be initialized to zeros."
+                "Defaults to 'xavier_uniform'. Setting this to `False` also uses 'xavier_uniform'. To set the weights "
+                "to zeros (thus making C3A a no-op), set the value to `False`."
```
githubnemo:
```suggestion
                "to zeros (thus making C3A a no-op), set the value to `True`."
```

githubnemo (approved):

PR #2737, opened 2025-08-13
FIX: DynamicCache key_cache attribute deprecation

BenjaminBossan:
Resolves [failing CI with transformers source install](https://github.com/huggingface/peft/actions/runs/16925932392/job/47961440263).

The `key_cache` attribute on `DynamicCache` is deprecated and will be removed in the 4.56.0 transformers release. The deprecation message mentions that the `layers` attribute should be used instead, which is what this PR does.


src/peft/peft_model.py
```diff
@@ -785,7 +785,7 @@ def get_prompt(
                 past_key_values = EncoderDecoderCache.from_legacy_cache(past_key_values)
                 past_key_values.cross_attention_cache = DynamicCache()
                 past_key_values.is_updated = {
-                    layer_idx: False for layer_idx in range(len(past_key_values.cross_attention_cache.key_cache))
+                    layer_idx: False for layer_idx in range(len(past_key_values.cross_attention_cache.layers))
```
githubnemo: Won't we need to check both attributes for a certain time? AFAICS the old `DynamicCache` instance doesn't have a `layers` attribute.

BenjaminBossan: Yes, true, but while looking at this, I found another issue. Namely, we first initialize `past_key_values.cross_attention_cache = DynamicCache()` and then iterate over its `key_cache` / `layers`, but of course those are going to be empty. I confirmed this when running the tests. So, technically, this line is a no-op.

Now I wonder: can it be removed or should we save the `is_updated` from before we override `cross_attention_cache` with the empty `DynamicCache()`? When checking the attribute before, it is not empty but can have values like `{0: True, 1: True}`.

@zucchini-nlp I think something was messed up in #2096 to end up in this situation.

zucchini-nlp: We had a major refactor on cache recently, and prob the new format isn't 100% backward compatible. Can you remind me what was the desired behavior for prefix-tuning on encoder decoder models. We append virtual tokens to decoder inputs only or to encoder ids as well?

BenjaminBossan: I just checked the [prefix tuning paper](https://arxiv.org/abs/2101.00190), the idea is to have separate prefix tokens (or rather, embeddings) for both encoder and decoder.

zucchini-nlp: Ahh, I see, Overwriting `cross_attention_cache` actually won't change the values of `is_updated` afaik and thus we still have to set them to `False` in order to save cross-encoder cache

Maybe we can just iterate over `cache.is_updated` and change it in-place, so we don't have to infer number of layers from `cache.layers`, wdyt?

BenjaminBossan: Ah, good suggestion. Just so I'm sure I understand correctly, currently we have:

```python
                past_key_values = EncoderDecoderCache.from_legacy_cache(past_key_values)
                past_key_values.cross_attention_cache = DynamicCache()
                past_key_values.is_updated = {
                    layer_idx: False for layer_idx in range(len(past_key_values.cross_attention_cache.layers))
                }
```

which makes no sense because `past_key_values.is_updated` will always end up being an empty dict. Instead, we should do this, right?

```python
                past_key_values = EncoderDecoderCache.from_legacy_cache(past_key_values)
                past_key_values.cross_attention_cache = DynamicCache()
                for key in past_key_values.is_updated.keys():
                    past_key_values.is_updated[key] = False
```

And this should also work with older transformers versions.

zucchini-nlp: yep, an empty dict would throw errors so we need an explicit `False`

BenjaminBossan:
@githubnemo Ready for another review.

src/peft/peft_model.py
```diff
@@ -774,19 +774,17 @@ def get_prompt(
...
+                # Dont' apply this to encoder-decoder models that don't support new Cache format yet
                 # If we don't apply this, prefix-tuning fails to update cross-attn cache
                 # TODO: remove check for _supports_cache_class once transformers 4.53 is no longer supported
                 past_key_values = EncoderDecoderCache.from_legacy_cache(past_key_values)
                 past_key_values.cross_attention_cache = DynamicCache()
-                past_key_values.is_updated = {
-                    layer_idx: False for layer_idx in range(len(past_key_values.cross_attention_cache.key_cache))
-                }
+                for key in past_key_values.is_updated.keys():
+                    past_key_values.is_updated[key] = False
```
githubnemo, line 788: Let's add a comment why we do this and what the consequence is

githubnemo (commented):
LGTM otherwise

BenjaminBossan:
@githubnemo Could you please re-review?

githubnemo (approved):

PR #2741, opened 2025-08-14
enable TP with lora adapter

3outeille:
This PR + the fix of HF tensor parallel at (https://github.com/huggingface/transformers/pull/40162) will enable lora to be used along with TP. Example of scripts: https://github.com/huggingface/gpt-oss-recipes/pull/24

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -113,7 +113,11 @@ class LoraLayer(BaseTunerLayer):
 
         base_layer = self.get_base_layer()
         if isinstance(base_layer, nn.Linear):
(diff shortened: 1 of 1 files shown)


src/peft/tuners/lora/layer.py
```diff
@@ -16,6 +16,7 @@
 import math
 import warnings
 from contextlib import contextmanager
+from packaging import version
```
BenjaminBossan: Move to line 24 to make ruff happy :)

BenjaminBossan (approved):
Thanks for enabling tensor parallel for LoRA Linear layers, LGTM.

PR #2743, opened 2025-08-15
X cast

nyxkrage:


BenjaminBossan:
Hi, I only skimmed the PR, from my understanding the goal would have been to allow targeting MoE layers with LoRA. You may have seen the new `target_parameters` feature in PEFT (#2638) which tries to solve this problem in a general fashion, perhaps that's why you closed the PR. But if you have custom kernel code that allows to make this more efficient, feel free to contribute.

nyxkrage:
Hey, the actual kernel is "stolen" from https://github.com/foundation-model-stack/fms-acceleration/blob/main/plugins/accelerated-moe/src/fms_acceleration_moe/utils/scattermoe.py, and yeah I did end up seeing target_parameters halfway through implementing this and that would be the "ideal" way to handle it. The PR only happened because I forgot to change the base repo/branch to the fork.

I don't have any intention of upstreaming this because the kernel requires models to be converted and have the modeling code patched, so I don't think it super makes sense to even have upstream in the huggingface org plus it was mostly to quickly get the kernel working with existing solutions like axolotl.

BenjaminBossan:
Thanks for providing the context, @nyxkrage, it makes sense.

PR #2758, opened 2025-08-28
FIX: Multiple active adapters with auxiliary layers

BenjaminBossan:
## Description

This PR fixes a few issues with the handling of active adapters for auxiliary modules.

### 1. Calling `set_adapter` on the `model.base_model`

When calling `peft_model.set_adapter`, it is not possible to activate more than one adapter, as not all PEFT methods support that. However, many PEFT methods like LoRA do, in which case users should call `peft_model.base_model.set_adapter(['default', 'other'])`.

Now the issue was that the activation of auxiliary modules was only done on `PeftModel.set_adapter`. This means that if users are calling `peft_model.base_model.set_adapter` (i.e. `LoraModel.set_adapter` etc.), the auxiliary adapters were not activated.

This PR fixes this issue by ensuring that even if the user activates adapters like this, the auxiliary modules are activated. When users activate more than one adapter, additional checks are performed to ensure that they are not activating multiple auxiliary modules on the same module.

Note that some existing PEFT code could start raising errors now because of the change. However, this PEFT code is buggy right now so IMO it is fine to raise an error.

### 2. Adding multiple adapters with non-overlapping auxiliary modules

Furthermore, I found an activation issue that could occur when adding multiple adapters with non-overlapping auxiliary modules. Normally, when the second/third/... adapter are being added, they are not automatically activated. However, when these additional adapters target new auxiliary modules, those would be incorrectly activated (because they look like they're the _first_ adapter). This has also been fixed.

## Open question

Right now, we don't allow users to activate multiple auxiliary adapters on the same module. However, this limitation could be considered too strict:

1. For trainable tokens, as long as the indices don't overlap, there is no conflict.
2. For `modules_to_save`, we could theoretically determine the "`delta_weight`" as `new_weight - original_weight`, then add up all `delta_weight`s.

This is not implemented in the PR for now to prevent it becoming even more complex.

## Note to maintainers

This PR requires the addition of `self.set_auxiliary_adapters` to the `set_adapters` method of all non-prompt learning PEFT method model classes. If this PR is merged, all open PRs that add such PEFT methods must be updated to include this call.

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -842,16 +842,13 @@ class BaseTuner(nn.Module, ABC):
         """
         Sets the active adapter(s) on auxiliary modules.
 
-        The subclass should call this method.
+        If the subclass (e.g. `LoraModel`) supports auxiliary modules like `modules_to_save`, it should call this
+        method in `set_adapter` to ensure that those auxiliary modules are being set correctly.
 
         Args:
             adapter_name (`str` or `list[str]`):
                 The name(s) of the adapter to be set as active. The adapters must be loaded first.
         """
-        # Note: When the user calls peft_model.set_adapter, this already takes care of auxiliary modules, so this call
-        # is redundant (but that shouldn't hurt). However, users may also call `peft_model.base_model.set_adapter` when
-        # they want to activate multiple adapters at the same time. In that case, the PeftModel instance is skipped,
-        # thus auxiliary modules must be set explicitly here.
         _set_adapter(self, adapter_name)
 
     def _delete_auxiliary_adapter(self, adapter_name: str, new_active_adapters: Optional[list[str]]) -> None:
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -423,6 +423,13 @@ class AuxiliaryTrainingWrapper(torch.nn.Module):
         else:
             self._disable_adapters = True
 
+    def check_set_adapter(self, adapter_name: str | list[str]) -> str | None:
+        """Helper function to check if the given adapter(s) can be set.
+
+        Return the name of the adapter to be set or None if no adapter should be set.
+        """
+        raise NotImplementedError
+
     def set_adapter(self, adapter_names: Union[str, list[str]]):
         """Set the active adapter
 
@@ -539,6 +546,31 @@ class ModulesToSaveWrapper(AuxiliaryTrainingWrapper):
         else:
             self.original_module.requires_grad_(True)
             self.modules_to_save.requires_grad_(False)
+
+    def check_set_adapter(self, adapter_name: str | list[str]) -> str | None:
+        """Helper function to check if the given adapter(s) can be set.
+
+        Return the name of the adapter to be set or None if no adapter should be set.
+        """
+        if isinstance(adapter_name, str):
+            return adapter_name
+
+        # adapter_name is a list of str
+        if len(adapter_name) == 0:
+            raise ValueError("Please specify at least one adapter to set")
+
+        adapter_names_in_module = [n for n in adapter_name if n in self.modules_to_save]
+
+        if len(adapter_names_in_module) > 1:
+            raise ValueError(f"Only one adapter can be set at a time for {self}, got {len(adapter_names_in_module)}")
+
+        adapter_name_to_set: str | None
+        if not adapter_names_in_module:
+            adapter_name_to_set = None
+        else:
+            adapter_name_to_set = adapter_names_in_module[0]
+
+        return adapter_name_to_set
 
     def set_adapter(self, adapter_names: Union[str, list[str]]):
         """Set the active adapter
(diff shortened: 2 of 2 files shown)


src/peft/peft_model.py
```diff
@@ -1437,8 +1437,11 @@ def set_adapter(self, adapter_name: str) -> None:
             raise ValueError(f"Adapter {adapter_name} not found.")
         self.active_adapter = adapter_name
         if not self.peft_config[adapter_name].is_prompt_learning:
+            # _set_adapter does not need to be called, since it's called through the BaseTuner class.
             self.base_model.set_adapter(adapter_name)
-        _set_adapter(self, adapter_name)
+        else:
+            # handle auxiliary modules
+            _set_adapter(self, adapter_name)
```
githubnemo, line 1444: We've recently added `modules_to_save` to `PromptLearningConfig` so at least for `modules_to_save` we'd need handling here, no?

src/peft/tuners/boft/model.py
```diff
@@ -246,6 +246,7 @@ def disable_adapter_layers(self):
         self._set_adapter_layers(enabled=False)
 
     def set_adapter(self, adapter_name):
+        self.set_auxiliary_adapters(adapter_name)
```
githubnemo, line 249: Can we move `set_adapter` to `BaseTuner`? It seems to be very similar for most of the model classes.

src/peft/tuners/tuners_utils.py
```diff
@@ -832,6 +838,22 @@ def unmerge_adapter(self):
                 with onload_layer(module):
                     module.unmerge()
 
+    def set_auxiliary_adapters(self, adapter_name: str | list[str]) -> None:
+        """
+        Sets the active adapter(s) on auxiliary modules.
+
+        The subclass should call this method.
```
githubnemo: Let's document when and why we should call this method.

src/peft/tuners/tuners_utils.py
```diff
@@ -832,6 +838,22 @@ def unmerge_adapter(self):
...
+
+        Args:
+            adapter_name (`str` or `list[str]`):
+                The name(s) of the adapter to be set as active. The adapters must be loaded first.
+        """
+        # Note: When the user calls peft_model.set_adapter, this already takes care of auxiliary modules, so this call
+        # is redundant (but that shouldn't hurt). However, users may also call `peft_model.base_model.set_adapter` when
+        # they want to activate multiple adapters at the same time. In that case, the PeftModel instance is skipped,
+        # thus auxiliary modules must be set explicitly here.
```
githubnemo:
```suggestion
        # they want to activate multiple adapters at the same time. In that case, the PeftModel instance is skipped
        # and the adapter model (this instance) is called, thus auxiliary modules must be set explicitly here.
```

src/peft/utils/other.py
```diff
@@ -560,12 +561,18 @@ def set_adapter(self, adapter_names: Union[str, list[str]]):
...
+            return
+
         adapter_name = adapter_names[0]
 
         if adapter_name not in self._adapters:
             raise ValueError(f"Adapter {adapter_name} not found in {self._adapters}")
 
-        self.modules_to_save[self.active_adapters[0]].requires_grad_(False)
+        for currently_active_adapter_name in self.active_adapters:
+            self.modules_to_save[currently_active_adapter_name].requires_grad_(False)
```
githubnemo, line 607: In case of `adapter_names == []` this will not reset the `requires_grad` field. Wouldn't it be best to do it similar to `enable_adapters` and use `self.modules_to_save.requires_grad_(False)` and then enable the adapters that should be enabled?

src/peft/utils/other.py
```diff
@@ -876,30 +889,48 @@ def _set_trainable(
...
+
+        if len(adapter_names_in_module) > 1:
+            raise ValueError(f"Only one adapter can be set at a time for {module}, got {len(adapter_names_in_module)}")
+
+        if not adapter_names_in_module:
+            adapter_name_to_set = None
+        else:
+            adapter_name_to_set = adapter_names_in_module[0]
+
+        return adapter_name_to_set
```
githubnemo: I wonder if we should delegate the check (and potential checking of capabilities like merging given a method - out of scope but forseeable) to the wrapper classes themselves, e.g. `AuxiliaryTrainingWrapper.check_new_adapter()`. WDYT?

githubnemo (commented):
Oof, thanks.

> Right now, we don't allow users to activate multiple 
auxiliary adapters on the same module. However, this limitation could be considered too strict:

The merging behavior you describe (summation of delta weights) can be done for overlapping trainable tokens and modules to save alike. I think there are at least three cases here:

- user just wants to load adapters and if they're overlapping, receive an error because the assumed orthogonality of the adapters is not given
- user wants to override specific parts of a pre-trained adapter and wants to load another adapter to overwrite specific parts (without merging)
- user is aware of the overlap between adapters and wants to merge the deltas in some way

Ideally we would support all cases (defaulting to the first case) with additional arguments to `add_adapter` but I'm not yet sure what this means for other places that use `add_adapter` and how to propagate such an option through possible abstractions.

I would see these as out of scope for this PR and I don't see anything anything blocking a possible implementation.

src/peft/peft_model.py
BenjaminBossan: Not sure if I understand the comment. The `else` clause is for prompt learning, my change above is to _skip_ calling `_set_adapter` for non-prompt learning as it's already called through `self.base_model.set_adapter` (via `set_auxiliary_adapters`). thus I avoid calling it twice, which would actually not hurt but it's just redundant.

src/peft/tuners/boft/model.py
BenjaminBossan: I thought the same thing. It's not 100% the same, so a bit of extra work would be required. I left it out on purpose, as this PR is already quite complex. I think this would be a good candidate for the refactor we discussed of improving the `BaseTuner` abstraction to reduce duplication.

src/peft/tuners/tuners_utils.py
BenjaminBossan: I extended the docstring and also removed the comment below, as it is no longer correct (I had started with a different implementation).

src/peft/tuners/tuners_utils.py
BenjaminBossan: As mentioned above, I removed the comment.

src/peft/utils/other.py
BenjaminBossan: Good idea, I refactored it.

src/peft/utils/other.py
BenjaminBossan: Discussing internally.

src/peft/peft_model.py
githubnemo: Ah, I understand. Thanks.

Issue #1472, opened 2024-02-16
Integrating X-LoRA


EricLBuehler:
Hello all, and thank you for your great work!

Earlier this week, we announced [X-LoRA](https://arxiv.org/abs/2402.07148), a flexible MoE approach for LoRA adapters. We implement deep layer- and token-wise scalings for multiple LoRA adapters and provide an implementation (https://github.com/EricLBuehler/xlora) that enables straightforward application to any model to which `peft` LoRA adapters may be applied to. This offers the possibility to orchestrate at a much finer level, that is, to achieve new combinations of adapter layers. This results in never-before-seen per-token deep layer-wise combinations of parameters to solve specific tasks. Sample weights are provided at: https://huggingface.co/lamm-mit/x-lora with examples in protein science in the paper. 

Would you be interested to perhaps integrate X-LoRA into `peft`? I would be happy to work on this if there is interest from you and the community.

BenjaminBossan:
Thanks for bringing this to our attention, this could be an interesting addition. I just had a quick glance at the repo but it looks quite solid to me.

I think something we should pay attention to if we want to integrate it to PEFT is the API. Ideally, we would like to be able to initiate an X-LoRA model (or load a trained one) just based on a config using the usual PEFT/transformers methods. That way, we could ensure that users can make use of X-LoRA without running custom code. I'm not sure how feasible that is, but it would be great to have.

EricLBuehler:
I think this would be feasible. The primary entry point for converting to X-LoRA is a function `add_xlora_to_model`. It already takes a `PreTrainedModel` and we have a config class that could easily be ported. Specifically, would adding a new model type class where I would port the mechanics from `add_xlora_to_model` to the mapping make sense?

BenjaminBossan:
> would adding a new model type class where I would port the mechanics from `add_xlora_to_model` to the mapping make sense?

Do you mean a new model class on the level of `PeftModel`, or something on the level of `LoraModel`? If the latter, yes, I think that would be a good idea here. Maybe you could just start with a quick draft PR and we see from there? Often it's easier to discuss once we can see the code.

EricLBuehler:
I have an implementation that is nearly ready for a draft PR on my fork. However, after looking at how `PeftModel` initializes its base model, I think a modification to the instantiation call for the `PeftModel` base model may be necessary for `xLoRAModel` to be on the same level as `LoraModel`.

[...]

PR #2761, opened 2025-08-29
FIX X-LoRA forward hook issue during generate

BenjaminBossan:
There was an issue that forward hooks would accumulate during generation, since one hook per forward step was being registered and generate would call `forward` multiple times. This is already undesirable, but to make it worse, only the last hook was removed, resulting in hooks accumulating.

This PR fixes the issue.

See https://github.com/huggingface/peft/issues/1472#issuecomment-3235817807

### DIFF
--- a/src/peft/tuners/xlora/model.py
+++ b/src/peft/tuners/xlora/model.py
@@ -382,13 +382,13 @@ class XLoraModel(BaseTuner):
             forward_handle = self.lora_model.model.register_forward_pre_hook(_pre_forward, with_kwargs=True)
 
         # Run the forward pass: first the scaling pass in the hook, and then with the base model
-        yield
-
-        if not self.disabled:
-            # TODO(EricLBuehler): If we get a forward exception, we may have multiple forward hooks.
-            for handle in hook_handles:
-                handle.remove()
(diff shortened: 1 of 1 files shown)


EricLBuehler (approved):
LGTM!

src/peft/tuners/xlora/model.py
```diff
@@ -374,17 +378,15 @@ def pre_forward(module, *args, **kwargs):
...
-            forward_handle = self.lora_model.model.register_forward_pre_hook(pre_forward, with_kwargs=True)
+            forward_handle = self.lora_model.model.register_forward_pre_hook(_pre_forward, with_kwargs=True)
 
         # Run the forward pass: first the scaling pass in the hook, and then with the base model
         yield
 
         if not self.disabled:
             # TODO(EricLBuehler): If we get a forward exception, we may have multiple forward hooks.
-            for handle in handles_to_remove:
+            for handle in hook_handles:
```
githubnemo: Can we remove this todo by wrapping the `yield` in `try: yield; except: raise; finally: remove_hooks`?

BenjaminBossan:
@githubnemo I added `try ... finally`

Issue #2756, opened 2025-08-27
Loading multiple adapters with PeftMixedModel


NguyenRichard:
Hi,

I have came across the issue of loading 2 adapters on a peft model as the PeftMixedModel __set_adapter raises the **error ValueError: Only one adapter can be set at a time for modules_to_save**.

My plan was to train a network with two different adapters adapter_a and adapter_b with several modules_to_save and lora targets.
Then in a second step have a finetuning with two adapters adapter_ft_a and adapter_ft_b with only some lora targets for some light training. The PeftMixedModel  would in this situation **need to switch between [adapter_a, adapter_ft_a] and [adapter_b, adapter_ft_b]**

Am I doing something wrong or is it just not supported?

From my meager understanding of the subject, I can see a problem trying to load 2 adapters with modules_to_save but loading one adapter with modules_to_save and another without might be possible by first applying the modules_to_save weights before taking care of all the lora weights.

I'm also open to suggestion for any work around if there is any or I might fallback to simply retrain the adapter_a and adapter_b without adding new adapters.

BenjaminBossan:
First a clarification, don't use `PeftMixedModel` for your use case. This class is specifically intended for combining _different_ PEFT methods, e.g. LoRA and LoHa; moreover, it's only intended for inference, not for further training. Since you use two LoRA adapters, it's not necessary.

For the case you described, you can do something like this:

```python
# load the trained adapters
base_model = ...
peft_model = PeftModel.from_pretrained(base_model, <path-to-adater-a>, adapter_name="adapter_a")
# note: pass is_trainable=True to from_pretrained if you plan to train the model further
peft_model.load_adapter(<path-to-adapter-b>, adapter_name="adapter_b")

# add the new adapters
peft_model.add_adapter("adapter_ft_a", lora_config_ft_a)
peft_model.add_adapter("adapter_ft_b", lora_config_ft_b)

# right now, only adapter a is active, so to activate two adapters:
peft_model.base_model.set_adapter(["adapter_a", "adapter_ft_a"])

# start training
```

When you create the optimizer, ensure that it receives all the LoRA parameters that should be trained, not just those of the active adapters, otherwise when you later switch to `adapter_ft_b`, it will not be trained.

[...]

Issue #2749, opened 2025-08-21
Set multiple adapters actively when training


liaoyy-hub:
Hi! In incremental scenarios, I want to train a new adapter while keeping some old adapters actively. Notice that PeftModel can set active adapter by "model.set_adapter()". But every time can set only one adapter, where the type of args "adapter_name" is "str" rather than "List[str]". I also notice that class "PeftMixedModel" can set multiple adapters actively but only support for inference, and this class uses "model.base_model.set_adapter()" to achieve it. So I am not sure can I also set multiple adapters actively when training. My code is as following:

```python
model = AutoModelForCausalLM.from_pretrained()
peft_config = LoraConfig()
model = get_peft_model(model, peft_config, adapter_name="new")
model.load_adapter(adapter_path, adapter_name="old")
model.base_model.set_adapter(["new", "old"])
for name, param in model.named_parameters():
    if "lora_A.old" in name or "lora_B.old" in name:
        param.requires_grad = False
training_args = TrainingArguments()
trainer = Trainer()
trainer.train()
```

BenjaminBossan:
Indeed, on the `PeftModel` instance, you can only set a single adapter. The reason for that is that not all PEFT methods support activating multiple adapters at the same time, so we don't allow that. Regarding `PeftMixedModel`, this is intended for inference with multiple different PEFT methods, e.g. mixing LoRA and IA³. So this is not what you want.

The way you show this in your code, using `model.base_model.set_adapter(["new", "old"])`, is indeed the correct way to activate multiple LoRA adapters at the same time. Therefore, you can keep your code as is.

One thing to pay attention to: When you use `Trainer` from transfomers, when it initializes the optimizer, it does something like:

`optimizer = Optimizer(p for p in model.parameters() if p.requires_grad)`

What this means is that if you initialize the `Trainer` and later you switch the adapter using `set_adapter`, the new adapter's parameters are not updated by the optimizer. If this is a concern, I would suggest adding all the adapters right from the start, then manually initialize the optimizer with all LoRA parameters, and pass this optimizer instance to the `Trainer`.

liaoyy-hub:
> Indeed, on the `PeftModel` instance, you can only set a single adapter. The reason for that is that not all PEFT methods support activating multiple adapters at the same time, so we don't allow that. Regarding `PeftMixedModel`, this is intended for inference with multiple different PEFT methods, e.g. mixing LoRA and IA³. So this is not what you want.
> 
[...]

PR #2763, opened 2025-08-29
DOC: Explain how to use multiple adapters at the same time

BenjaminBossan:
Explain how to use multiple adapters (e.g. 2 LoRA adapters) at the same time, as the API is not quite intuitive and there are some footguns around trainable parameters.

This question has come up multiple times in the past (for recent examples, check #2749 and #2756). Thus it's a good idea to properly document this.

### DIFF
--- a/docs/source/developer_guides/troubleshooting.md
+++ b/docs/source/developer_guides/troubleshooting.md
@@ -422,11 +422,8 @@ model.load_adapter(lora_path_1, adapter_name="other")
 model.base_model.set_adapter(["default", "other"])
 ```
 
-<Tip>
-
-In the example above, you can see that we need to call `model.base_model.set_adapter(["default", "other"])`. Why can we not call `model.set_adapter(["default", "other"])`? This is unfortunately not possible because, as explained earlier, some PEFT methods don't support activating more than one adapter at a time.
-
-</Tip>
+> [!TIP]
+> In the example above, you can see that we need to call `model.base_model.set_adapter(["default", "other"])`. Why can we not call `model.set_adapter(["default", "other"])`? This is unfortunately not possible because, as explained earlier, some PEFT methods don't support activating more than one adapter at a time.
 
 It is also possible to train two adapters at the same time, but you should be careful to ensure that the weights of both adapters are known to the optimizer. Otherwise, only one adapter will receive updates.
 
@@ -465,9 +462,6 @@ model.base_model.set_adapter(["default", "other"])
 trainer = Trainer(model=model,  ...)
 trainer.train()
 ```
-<Tip>
 
-This section deals with using multiple adapters _of the same type_ on the same model, for example, using multiple LoRA adapters at the same time. It does not apply to using _different types_ of adapters on the same model, for example one LoRA adapter and one LoHa adapter. For this, please check [`PeftMixedModel`](https://huggingface.co/docs/peft/developer_guides/mixed_models).
-
-</Tip>
-
+> [!TIP]
+> This section deals with using multiple adapters _of the same type_ on the same model, for example, using multiple LoRA adapters at the same time. It does not apply to using _different types_ of adapters on the same model, for example one LoRA adapter and one LoHa adapter. For this, please check [`PeftMixedModel`](https://huggingface.co/docs/peft/developer_guides/mixed_models).


docs/source/developer_guides/troubleshooting.md
```diff
@@ -401,3 +401,73 @@ If it is not possible for you to upgrade PEFT, there is a workaround you can try
 Assume the error message says that the unknown keyword argument is named `foobar`. Search inside the `adapter_config.json` of this PEFT adapter for the `foobar` entry and delete it from the file. Then save the file and try loading the model again.
 
 This solution works most of the time. As long as it is the default value for `foobar`, it can be ignored. However, when it is set to some other value, you will get incorrect results. Upgrading PEFT is the recommended solution.
+
+## Adapter handling
+
+### Using multiple adapters at the same time
```
stevhliu, line 407:
```suggestion
## Using multiple adapters at the same time
```

docs/source/developer_guides/troubleshooting.md
```diff
@@ -401,3 +401,73 @@ If it is not possible for you to upgrade PEFT, there is a workaround you can try
...
+
+PEFT allows you to create more than one adapter on the same model. This can be useful in many situations. For example, for inference, you may want to serve two fine-tunes using the same base model instead of loading the base model once for each fine-tune, which would cost more memory. However, a less well known feature is that multiple adapters can be activated at the same time. This way, the model may leverage the learnings from all those adapters at the same time. As an example, if you have a diffusion model, you may want to use one LoRA adapter to change the style and a different one to change the subject.
```
stevhliu:
```suggestion
PEFT allows you to create more than one adapter on the same model. This can be useful in many situations. For example, for inference, you may want to serve two fine-tuned models from the same base model instead of loading the base model once for each fine-tuned model, which would cost more memory. However, multiple adapters can be activated at the same time. This way, the model may leverage the learnings from all those adapters at the same time. As an example, if you have a diffusion model, you may want to use one LoRA adapter to change the style and a different one to change the subject.
```

docs/source/developer_guides/troubleshooting.md
```diff
@@ -401,3 +401,73 @@ If it is not possible for you to upgrade PEFT, there is a workaround you can try
...
+
+Activating multiple adapters at the same time is generally possible on all PEFT methods except for prompt learning methds (p-tuning, prefix tuning, etc.). Therefore, it works for LoRA, LoHa, IA³ etc. The following example illustrates how to achieve this:
```
stevhliu:
```suggestion
Activating multiple adapters at the same time is generally possible on all PEFT methods (LoRA, LoHa, IA³, etc.) except for prompt learning methods (p-tuning, prefix tuning, etc.). The following example illustrates how to achieve this:
```

docs/source/developer_guides/troubleshooting.md
```diff
@@ -401,3 +401,73 @@ If it is not possible for you to upgrade PEFT, there is a workaround you can try
...
+model.base_model.set_adapter(["default", "other"])
+```
+
+<Tip>
+
+In the example above, you can see that we need to call `model.base_model.set_adapter(["default", "other"])`. Why can we not call `model.set_adapter(["default", "other"])`? This is unfortunately not possible because, as explained earlier, some PEFT methods don't support activating more than one adapter at a time.
+
+</Tip>
+
+Some of you may also want to train two adapters at the same time, e.g. training one LoRA adapter on one loss and the other one on another loss. This is also possible but you should be careful to ensure that the weights of both adapters are known to the optimizer. Otherwise, only one adapter will receive updates.
```
stevhliu:
```suggestion
It is also possible to train two adapters at the same time, but you should be careful to ensure that the weights of both adapters are known to the optimizer. Otherwise, only one adapter will receive updates.
```

docs/source/developer_guides/troubleshooting.md
```diff
@@ -401,3 +401,73 @@ If it is not possible for you to upgrade PEFT, there is a workaround you can try
...
+trainer.train()
+```
+
+or
+
+```python
+optimizer = torch.optim.AdamW([param for param in model.parameters() if param.requires_grad], ...)
+```
+
+then the second LoRA adapter (`"other"`) would not be trained. This is because it is inactive at this moment, thus the `requires_grad` attribute on its parameters is set to `False` and the optimizer will ignore it. Therefore, ensure to activate all adapters that should be trained _before_ initializing the optimizer:
```
stevhliu:
```suggestion
then the second LoRA adapter (`"other"`) would not be trained. This is because it is inactive at this moment, which means the `requires_grad` attribute on its parameters is set to `False` and the optimizer will ignore it. Therefore, make sure to activate all adapters that should be trained _before_ initializing the optimizer:
```

docs/source/developer_guides/troubleshooting.md
```diff
@@ -401,3 +401,73 @@ If it is not possible for you to upgrade PEFT, there is a workaround you can try
...
+
+model_id = ...
+base_model = AutoModelForCausalLM.from_pretrained(model_id)
+model = PeftModel-from_pretrained(base_model, lora_path_0)  # default adapter_name is 'default'
+model.load_adapter(lora_path_1, adapter_name="other")
+# the 'other' adapter was loaded but it's not active yet, so to activate both adapters:
+model.base_model.set_adapter(["default", "other"])
+```
+
+<Tip>
```
stevhliu: Minor nit, but it may be nice to reduce the number of lines to use this instead of `<Tip> ... </Tip>`

```md
> [!TIP]
> ...
```

docs/source/developer_guides/troubleshooting.md
```diff
@@ -401,3 +401,73 @@ If it is not possible for you to upgrade PEFT, there is a workaround you can try
...
+
+Activating multiple adapters at the same time is generally possible on all PEFT methods except for prompt learning methds (p-tuning, prefix tuning, etc.). Therefore, it works for LoRA, LoHa, IA³ etc. The following example illustrates how to achieve this:
+
+```python
+from transformers import AutoModelForCausalLM
+from peft import PeftModel
+
+model_id = ...
+base_model = AutoModelForCausalLM.from_pretrained(model_id)
+model = PeftModel-from_pretrained(base_model, lora_path_0)  # default adapter_name is 'default'
```
stevhliu:
```suggestion
model = PeftModel.from_pretrained(base_model, lora_path_0)  # default adapter_name is 'default'
```

stevhliu (approved):
Thanks for adding new docs!

docs/source/developer_guides/troubleshooting.md
BenjaminBossan: It's supposed to be a subsection of "Adapter handling", so `###`.

docs/source/developer_guides/troubleshooting.md
BenjaminBossan: Changed.

BenjaminBossan (commented):
Thanks for the review @stevhliu

Issue #2757, opened 2025-08-28
Remove usage of from_legacy_cache


BenjaminBossan:
We use the `from_legacy_cache` method of transformers `Cache` classes in PEFT. This method will be removed from transformers in v4.58.0. PEFT should remove this code and find a different solution before then.

SP1029:
Greetings @BenjaminBossan 

I tried going over this, and the following was what I was able to understand. Could you please provide suggestions?

**Existing Situation**

`from_legacy_cache` is being used in `peft` in the following, where it is being primarily used to convert key-values to a `cache` object for the `DynamicCache` and the `EncoderDecoderCache`.

https://github.com/huggingface/peft/blob/a3197b1ec54baa107e2efc4a30c19209d9ea489a/src/peft/peft_model.py#L789-L796


In the `transformers` repo, the following is the main file for our purposes:

https://github.com/huggingface/transformers/blob/1363fceeec955db125141c3d7768f8ef345ed288/src/transformers/cache_utils.py

The overall structure is as follows. `Cache` is an object consisting of one cache per layer, where each layer's cache stores the keys and corresponding values in two tensors. `DynamicCache` is a variant of `Cache`, allowing updates to the keys and values by appending new keys and values on calling `update`. `EncoderDecoderCache` consists of two `DynamicCache` objects. `from_legacy_cache` converts the cache in the form of tuples to these cache objects.

**Possible Solution**

For `DynamicCache` in `init`, there is an argument `ddp_cache_data` which takes the data of the cache, the `past_key_values` can be passed as this argument, since `from_legacy_cache` also seems to process in a similar manner. Similarly, for the `caches` in the `init` function of `EncoderDecoderCache`.
```diff
- 789 past_key_values = DynamicCache.from_legacy_cache(past_key_values)
+ 789 past_key_values = DynamicCache(past_key_values)
- 796 past_key_values = EncoderDecoderCache.from_legacy_cache(past_key_values)
+ 796 past_key_values = EncoderDecoderCache(past_key_values)
```

BenjaminBossan:
@SP1029 Thanks for taking a look at this. Indeed, the proposed change should work. Ideally, we can have a small script for this purpose that we can run once before and once after to ensure that the outputs don't change (a regression test). It's not absolutely needed, but it could give us some peace of mind. LMK if you would like to take the task.

SP1029:
Yes, thank you.

SP1029:
[...]

PR #2767, opened 2025-09-03
Replace from_legacy_cache method with constructors 

SP1029:
Issue #2757 is about replacing the `from_legacy_cache` method of `transformers` `Cache` classes in `peft` due to their removal from transformers in v4.58.0. The `from_legacy_cache` method is used to convert data in the form of iterables to Cache objects. 

The issue is addressed by replacing the method with the corresponding constructors of the Cache classes since they also deal with the data in a similar way.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -786,14 +786,30 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
                 past_key_values = new_cache
             elif peft_config.num_transformer_submodules == 1:
                 # Dont' apply this to encoder-decoder models and not to models requiring special processing.
-                past_key_values = DynamicCache(past_key_values)
+                # TODO: remove from_legacy_cache once transformers < 4.56 is dropped
+                transformers_lt_4_56 = packaging.version.parse(transformers.__version__) < packaging.version.parse(
+                    "4.56.0.dev0"
+                )
+                if transformers_lt_4_56:
+                    past_key_values = DynamicCache.from_legacy_cache(past_key_values)
+                else:
+                    past_key_values = DynamicCache(past_key_values)
+
             elif (peft_config.num_transformer_submodules == 2) and getattr(
                 self.base_model, "_supports_cache_class", True
             ):
                 # Dont' apply this to encoder-decoder models that don't support new Cache format yet
                 # If we don't apply this, prefix-tuning fails to update cross-attn cache
                 # TODO: remove check for _supports_cache_class once transformers 4.53 is no longer supported
-                past_key_values = EncoderDecoderCache(past_key_values)
+                # TODO: remove from_legacy_cache once transformers < 4.56 is dropped
+                transformers_lt_4_56 = packaging.version.parse(transformers.__version__) < packaging.version.parse(
+                    "4.56.0.dev0"
+                )
+                if transformers_lt_4_56:
+                    past_key_values = EncoderDecoderCache.from_legacy_cache(past_key_values)
+                else:
+                    past_key_values = EncoderDecoderCache(past_key_values)
+
                 past_key_values.cross_attention_cache = DynamicCache()
                 # invalidate the cross attention cache, since we add virtual tokens to the encoder
                 for key in past_key_values.is_updated.keys():
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -786,14 +786,14 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
                 past_key_values = new_cache
             elif peft_config.num_transformer_submodules == 1:
                 # Dont' apply this to encoder-decoder models and not to models requiring special processing.
-                past_key_values = DynamicCache.from_legacy_cache(past_key_values)
+                past_key_values = DynamicCache(past_key_values)
             elif (peft_config.num_transformer_submodules == 2) and getattr(
                 self.base_model, "_supports_cache_class", True
             ):
                 # Dont' apply this to encoder-decoder models that don't support new Cache format yet
                 # If we don't apply this, prefix-tuning fails to update cross-attn cache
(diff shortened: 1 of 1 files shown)


SP1029:
The testing script below passes locally for me. @BenjaminBossan, would this script be fine, or would we want to add anything else?

```py
import pickle
import torch
from peft import PrefixTuningConfig, get_peft_model
from transformers import AutoModelForCausalLM, AutoModelForSeq2SeqLM

torch.manual_seed(0)

# Constants
CODEBASE = "old" # new/old

# Test definitions

TEST = [
    "Decoder",
    "Encoder-Decoder",
    "Gemma"
]

MODEL = [
    AutoModelForCausalLM.from_pretrained("hf-internal-testing/tiny-random-MistralForCausalLM"),
    AutoModelForSeq2SeqLM.from_pretrained("hf-internal-testing/tiny-random-BartForConditionalGeneration"),
    AutoModelForCausalLM.from_pretrained("hf-internal-testing/tiny-random-Gemma3ForCausalLM"),
]

PEFT_CONFIG = [
    PrefixTuningConfig(num_virtual_tokens=10, task_type="CAUSAL_LM"),
    PrefixTuningConfig(num_virtual_tokens=10, task_type="SEQ_2_SEQ_LM"),
    PrefixTuningConfig(num_virtual_tokens=10, task_type="CAUSAL_LM")
]

# Supporting functions

def save_cache(model, test):
    batch_size = 1
    
    with open(f"{test}_{CODEBASE}.pkl", "wb") as f:
        pickle.dump(model.get_prompt(batch_size), f, pickle.HIGHEST_PROTOCOL)    
       
 # for cache data of a single layer
def is_equal_layer(layer1, layer2):
    
    for attr in layer1.__dict__:
        check = bool(float(torch.tensor(layer1.__dict__[attr] == layer2.__dict__[attr]).min()))
        if check == False:
            return False
        
    return True

# For DynamicCache objects
def is_equal_DynamicCache(cache1, cache2):
    
    if len(cache1.layers)!=len(cache2.layers):
        return False
    
    for i in range(len(cache1.layers)):
        if not is_equal_layer(cache1.layers[i], cache2.layers[i]):
            return False
        
    return True

# For DynamicCache objects of the EncoderDecoderCache
def is_equal_EncoderDecoderCache(cache1, cache2):
    
    if not is_equal_DynamicCache(cache1.self_attention_cache, cache2.self_attention_cache):
        return False
    
    if not is_equal_DynamicCache(cache1.cross_attention_cache, cache2.cross_attention_cache):
        return False
    
    return True

def is_equal_cache(cache1, cache2):
    
    cache_type = type(cache1).__name__
    
    if cache_type=="DynamicCache":
        return is_equal_DynamicCache(cache1, cache2)
    elif cache_type=="EncoderDecoderCache":
        return is_equal_EncoderDecoderCache(cache1, cache2)

# Executing tests

for i in range(3):
    model = get_peft_model(MODEL[i], PEFT_CONFIG[i])
    test = TEST[i]
    save_cache(model, test)

# Comparing

for i in range(3):
    test = TEST[i]
    with open(f"{test}_old.pkl", "rb") as f:
        cache1 = pickle.load(f)
    with open(f"{test}_new.pkl", "rb") as f:
        cache2 = pickle.load(f)
        
    print(test, is_equal_cache(cache1, cache2))
    
```

src/peft/peft_model.py
```diff
@@ -786,14 +786,14 @@ def get_prompt(
                 past_key_values = new_cache
             elif peft_config.num_transformer_submodules == 1:
                 # Dont' apply this to encoder-decoder models and not to models requiring special processing.
-                past_key_values = DynamicCache.from_legacy_cache(past_key_values)
+                past_key_values = DynamicCache(past_key_values)
```
BenjaminBossan: We have to support older transformers versions too. We should do a version check and then do either this or the old way with `from_legacy_cache`. Here is an example how we check versions in PEFT:

https://github.com/huggingface/peft/blob/2ea5377ee3069fa5aa49ed3930ac52bc706aed85/src/peft/peft_model.py#L754-L756

Let's also add a `TODO` comment to remove `from_legacy_cache` completely once we no longer support transformers < 4.58.0.

src/peft/peft_model.py
```diff
@@ -786,14 +786,14 @@ def get_prompt(
...
             elif (peft_config.num_transformer_submodules == 2) and getattr(
                 self.base_model, "_supports_cache_class", True
             ):
                 # Dont' apply this to encoder-decoder models that don't support new Cache format yet
                 # If we don't apply this, prefix-tuning fails to update cross-attn cache
                 # TODO: remove check for _supports_cache_class once transformers 4.53 is no longer supported
-                past_key_values = EncoderDecoderCache.from_legacy_cache(past_key_values)
+                past_key_values = EncoderDecoderCache(past_key_values)
```
BenjaminBossan: Same comment as above.

BenjaminBossan (changes requested):
Thanks for working on this PR and especially for checking the results. This check is more in-depth than what I had in mind (I would have simply compared model outputs) but it works too.

I have one comment concerning backwards compatibility, otherwise the PR looks good.

src/peft/peft_model.py
SP1029: Ok, the check and the `TODO` for 4.56.0 have been added.

@BenjaminBossan Just to confirm 4.5**8**.0 for `TODO` was a typo, right, since check is for 4.5**6**.0, or was it intended?

src/peft/peft_model.py
SP1029: Updated.

BenjaminBossan (approved):
Thanks for working on this task, LGTM.

> Just to confirm 4.58.0 for TODO was a typo, right, since check is for 4.56.0, or was it intended?

Indeed. The deprecation is schedules for 4.58.0 but the support for `from_legacy_cache` can still be dropped once we move beyond 4.56.0, good catch.

PR #2771, opened 2025-09-04
[WIP] The great deduplication

BenjaminBossan:


docs/source/package_reference/functional.md
```diff
@@ -0,0 +1,33 @@
+<!--⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+-->
+
+# Functions for PEFT integration
```
sayakpaul, line 5: Documentation stuff in a deduplication PR? 👁️

src/peft/utils/constants.py
```diff
@@ -364,47 +281,9 @@ def starcoder_model_postprocess_past_key_value(past_key_values):
     "qwen3": ["q_proj", "v_proj"],
 }
```
sayakpaul, line 283: How can this be deleted? A small comment in-line here would he helpful.

src/peft/tuners/adalora/model.py
```diff
@@ -221,25 +222,6 @@ def _create_new_module(lora_config, adapter_name, target, device_map=None, **kwa
 
         return new_module
 
-    @staticmethod
```
sayakpaul, line 224: Seems like these are already a part of the baseclass the said class is overriding?

src/peft/utils/constants.py
BenjaminBossan: As these mappings were all identical or almost identical to the one for LoRA, I just made them copies of the LoRA mappings, e.g. https://github.com/huggingface/peft/pull/2771/files#diff-b054446713795a79a89f2285d66068573d69f0b701ee31419bb223d1601e7844R108 for C3A.

I mentioned that in the PR description:

> While working on this PR, I also deduplicated almost identical TRANSFORMERS_MODULES_TO_XXX_TARGET_MODULES_MAPPING constants by copying them from LoRA and only overriding a few values that differ. Moreover, some PEFT methods didn't have their own TRANSFORMERS_MODULES_TO_XXX_TARGET_MODULES_MAPPING but used the one from LoRA instead. They now each have their own constant, which is a copy from the one from LoRA.

src/peft/tuners/adalora/model.py
BenjaminBossan: Sorry, I don't understand. Could you please elaborate?

docs/source/package_reference/functional.md
BenjaminBossan: I thought while I'm at it, I might as well go the extra mile.

sayakpaul:
> Btw. one unfortunate side effect of this PR is that the test coverage declines slightly (by 1%) because there are fewer duplicated lines being tested :D

Exactly why regular software engineering coverage often cannot be applied one-on-one in these libraries. I am totally fine with that.

src/peft/tuners/adalora/model.py
sayakpaul: Why are these methods be deleted.

BenjaminBossan: I think I still don't get it. These methods are deleted exactly because the base class already implements them, so there is no need to have them on the subclass.

sayakpaul: Yeah that is what I thought.

src/peft/tuners/ia3/model.py
```diff
@@ -294,109 +201,6 @@ def _prepare_adapter_config(peft_config, model_config):
...
-                NaNs. Defaults to `False`.
-            adapter_names (`List[str]`, *optional*):
-                The list of adapter names that should be merged. If None, all active adapters will be merged. Defaults
-                to `None`.
-        """
-        if getattr(self.model, "is_loaded_in_8bit", False):
-            raise ValueError("Cannot merge ia3 layers when the model is loaded in 8-bit mode")
-
-        if getattr(self.model, "is_loaded_in_4bit", False):
-            raise ValueError("Cannot merge ia3 layers when the model is loaded in 4-bit mode")
```
githubnemo, line 317: Is this not true anymore?

src/peft/tuners/ln_tuning/model.py
```diff
@@ -113,59 +93,13 @@ def _create_new_module(
...
-                    module.unmerge()
-                module.set_adapter(adapter_name, inference_mode=inference_mode)
-        self.active_adapter = adapter_name
+    def _unloading_checks(self, adapter_names: Optional[list[str]]):
+        adapters_to_consider = adapter_names or self.active_adapters
+        is_modules_to_save_available = any(
+            self.peft_config[adapter].modules_to_save for adapter in adapters_to_consider
+        )
+        if is_modules_to_save_available and len(adapters_to_consider) > 1:
+            raise ValueError("Cannot unload multiple adapters that specify `modules_to_save`.")
```
githubnemo, line 102: is this a remnant and can be removed? doesn't seem to be used.

src/peft/tuners/randlora/model.py
```diff
@@ -418,151 +354,3 @@ def _create_new_module(randlora_config, randlora_A, randlora_B, adapter_name, ta
...
-
-    def _unload_and_optionally_merge(
-        self,
-        merge=True,
-        progressbar: bool = False,
-        safe_merge: bool = False,
-        adapter_names: Optional[list[str]] = None,
-    ):
-        # we cannot use self.prefix as we want to include non-trainable randlora parameters
-        key_list = [key for key, _ in self.model.named_modules() if "randlora" not in key]
```
githubnemo, line 487: the comment indicates that we cannot use `self.prefix` (== `randlora_`) and instead uses `randlora` without the underscore. 

the oldest currently present mention of this is in fourierft (in `delete_adapter`) but it didn't seem to be relevant there either:

```
% git grep 'we cannot use self' | cut -d ":" -f 1 | uniq | while read f; do git blame "$f" | grep 'we cannot'; done | sort -k 3 -r
e72a96f7 (Aochuan Chen       2024-07-09 18:20:01 +0800 312)         # we cannot use self.prefix as we want to include non-trainable fourierft parameters
5a4b9cad (Benjamin Bossan 2024-04-19 10:55:58 +0200 455)         # we cannot use self.prefix as we want to include non-trainable vera parameters
5a4b9cad (Benjamin Bossan 2024-04-19 10:55:58 +0200 424)         # we cannot use self.prefix as we want to include non-trainable vera parameters
d3ff1334 (Paul Albert     2025-04-25 22:56:24 +1000 517)         # we cannot use self.prefix as we want to include non-trainable randlora parameters
d3ff1334 (Paul Albert     2025-04-25 22:56:24 +1000 486)         # we cannot use self.prefix as we want to include non-trainable randlora parameters
a4f9334f (kkb-code        2025-07-14 02:16:10 -0700 287)         # we cannot use self.prefix as we want to include non-trainable shira parameters
a4f9334f (kkb-code        2025-07-14 02:16:10 -0700 256)         # we cannot use self.prefix as we want to include non-trainable shira parameters
```

not sure if this is relevant at all but i thought i'll mention it.

src/peft/tuners/randlora/model.py
```diff
@@ -418,151 +354,3 @@ def _create_new_module(randlora_config, randlora_A, randlora_B, adapter_name, ta
...
-
-        Args:
-            adapter_name (str): Name of the adapter to be deleted.
-        """
-        if adapter_name not in list(self.peft_config.keys()):
-            raise ValueError(f"Adapter {adapter_name} does not exist")
-        del self.peft_config[adapter_name]
-
-        # we cannot use self.prefix as we want to include non-trainable randlora parameters
-        key_list = [key for key, _ in self.model.named_modules() if "randlora" not in key]
```
githubnemo, line 518: same as above

src/peft/utils/save_and_load.py
```diff
@@ -58,7 +58,15 @@ def get_peft_model_state_dict(
     model, state_dict=None, adapter_name="default", unwrap_compiled=False, save_embedding_layers="auto"
 ):
     """
-    Get the state dict of the Peft model.
+    Get the state dict of the given adapter of the PEFT model.
+
+    This only includes the PEFT parameters, not the parameters of the base model. Thus the returned `state_dict` is
+    generally small compared to the full model size. To retrieve the full `state_dict`, just call `model.state_dict`().
```
githubnemo:
```suggestion
    generally small compared to the full model size. To retrieve the full `state_dict`, just call `model.state_dict()`.
```

src/peft/utils/save_and_load.py
```diff
@@ -58,7 +58,15 @@ def get_peft_model_state_dict(
...
+
+    Note that the adapter name is removed from the `state_dict`, is this is just an arbitrary name that can be changed
```
githubnemo:
```suggestion
    Note that the adapter name is removed from the `state_dict`, as this is just an arbitrary name that can be changed
```

src/peft/tuners/ia3/model.py
BenjaminBossan: Good catch, I re-introduced these checks and then call `super()._unload_and_optionally_merge(...)`.

src/peft/tuners/ln_tuning/model.py
BenjaminBossan: Hmm, good question. So I agree that this check doesn't make too much sense, but the general idea is right: If we have multiple `modules_to_save`, then `merge_and_unload` doesn't really work. Right now, we just let it pass:

```python
from copy import deepcopy

config = LoraConfig(modules_to_save=["lm_head"])
model.add_adapter("adapter2", config)
model.base_model.model.lm_head.modules_to_save.default.weight.data.fill_(111)
model.base_model.model.lm_head.modules_to_save.adapter2.weight.data.fill_(222)
unloaded = deepcopy(model).merge_and_unload(adapter_names=["default", "adapter2"])
print(unloaded.lm_head.weight)  # is [[111, 111, ...]]
```

I'd suggest to leave this situation as is for now and handle it in a separate PR. WDYT?

src/peft/tuners/randlora/model.py
BenjaminBossan: Yeah, I don't think it's necessary. I added an assert that the `key_list` is the same whether `randlora` or `randlora_` is being used, then ran the whole test suite of randlora on main. The assert never triggers, so I think it's safe to remove.

src/peft/utils/save_and_load.py
BenjaminBossan: done

src/peft/utils/save_and_load.py
BenjaminBossan: done

src/peft/tuners/ln_tuning/model.py
githubnemo: Agreed, let's leave it as-is for now.

Issue #2733, opened 2025-08-08
LoraLayer.set_scale not working correctly with rslora


dongmingli-Ben:
### System Info

peft: 0.15.2
diffusers: 0.34.0

### Reproduction

https://github.com/huggingface/peft/blob/a90003f0edd6353f489f48bd2c35080d27bb6974/src/peft/tuners/lora/layer.py#L496

Here the scale is set assuming that the LoRA is not trained with rslora. For rslora LoRAs, the scale is incorrectly set comparing to this:

https://github.com/huggingface/peft/blob/a90003f0edd6353f489f48bd2c35080d27bb6974/src/peft/tuners/lora/layer.py#L229

This matters when trying to load lora adapters thourgh diffusers' `PeftAdapterMixin.load_lora_adapter` and then setting the active adapters with lora scales. For example:

```python
class DiT(nn.Module, PeftAdapterMixin):
    ...


def _load_lora_adapter(model: DiT, lora_paths):
    lora_paths = lora_paths.split(',')
    adapters = []
    # the lora is trained like this: https://huggingface.co/docs/peft/task_guides/lora_based_methods#peft-configuration-and-model and rslora=True, and saved with `.save_pretrained`
    for i, path in enumerate(lora_paths):
        lora_name = path.split('/')[-1]
        print(f"Load LoRA {lora_name} from {path}")
        state_dict = load_file(os.path.join(path, 'adapter_model.safetensors'))
        ready_state_dict = {}
        for k, v in state_dict.items():
            # state_dict.pop(k)
            ready_state_dict[k.removeprefix("base_model.model.")] = v.to('cuda')
        with open(os.path.join(path, 'adapter_config.json'), 'r') as f:
            lora_config = json.load(f)
        model.load_lora_adapter(ready_state_dict, prefix=None, adapter_name=lora_name, metadata=lora_config)
        adapters.append(lora_name)
    # With the following .set_adapters uncommented, the output is different when running the model with the same input
    # model.set_adapters(adapters, [1.] * len(lora_paths))
```

### Expected behavior

The model output should be the same with `model.set_adapters(adapters, [1.] * len(lora_paths))`

BenjaminBossan:
Thanks for reporting this error @dongmingli-Ben. Your conclusion looks correct to me. Are you interested in providing a PR to fix this?

tanuj-rai:
[...]

PR #2775, opened 2025-09-07
Fix RS-LoRA scaling in set_scale

tanuj-rai:
This PR fixes a scaling mismatch for **RS-LoRA** adapters when calling **set_scale**

Fixes issue #2733 

Who can review:
@BenjaminBossan

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -113,6 +113,7 @@ class LoraLayer(BaseTunerLayer):
         self._disable_adapters = False
         self.merged_adapters = []
         self.use_dora: dict[str, bool] = {}  # not actively used anymore after #2443, keep it for BC
+        self.use_rslora: dict[str, bool] = {}
         self.lora_bias: dict[str, bool] = {}
         self.lora_magnitude_vector = torch.nn.ModuleDict()  # for DoRA
         self._caches: dict[str, Any] = {}
@@ -248,6 +249,8 @@ class LoraLayer(BaseTunerLayer):
             self.scaling[adapter_name] = lora_alpha / math.sqrt(r)
         else:
             self.scaling[adapter_name] = lora_alpha / r
+
+        self.use_rslora[adapter_name] = use_rslora
 
         self.use_dora[adapter_name] = use_dora
 
@@ -511,8 +514,11 @@ class LoraLayer(BaseTunerLayer):
         if adapter not in self.scaling:
             # Ignore the case where the adapter is not in the layer
             return
-        self.scaling[adapter] = scale * self.lora_alpha[adapter] / self.r[adapter]
-
+        if self.use_rslora.get(adapter, False):
+            self.scaling[adapter] = scale * self.lora_alpha[adapter] / math.sqrt(self.r[adapter])
+        else:
+            self.scaling[adapter] = scale * self.lora_alpha[adapter] / self.r[adapter]
+    
     def scale_layer(self, scale: float | int) -> None:
         """Multiply the current scale of all active adapters by the provided factor"""
         if scale == 1:
@@ -920,6 +926,8 @@ class Embedding(nn.Module, LoraLayer):
         else:
             self.scaling[adapter_name] = lora_alpha / r
 
+        self.use_rslora[adapter_name] = use_rslora
+
         self.use_dora[adapter_name] = use_dora
 
         if init_lora_weights == "loftq":
@@ -1206,6 +1214,8 @@ class _ConvNd(nn.Module, LoraLayer):
             self.scaling[adapter_name] = lora_alpha / math.sqrt(r)
         else:
             self.scaling[adapter_name] = lora_alpha / r
+
+        self.use_rslora[adapter_name] = use_rslora
 
         self.use_dora[adapter_name] = use_dora
 
@@ -1980,6 +1990,8 @@ class ParamWrapper(nn.Module, LoraLayer):
             self.scaling[adapter_name] = lora_alpha / math.sqrt(r)
         else:
             self.scaling[adapter_name] = lora_alpha / r
+
+        self.use_rslora[adapter_name] = use_rslora
 
         self.use_dora[adapter_name] = use_dora
 


tanuj-rai:
Hey @BenjaminBossan! I have raised a PR to fix this issue. Please take a look when you have a moment and let me know if anything needs to be updated further.

tanuj-rai:
@

> Thanks for working on this PR. Before we proceed, let's add some unit tests to ensure that everything works as expected. As a starting point, check these tests:
> 
> (
> 
> https://github.com/huggingface/peft/blob/5d97453235d3c3073b059995f56260d364d44e83/tests/test_initialization.py#L3857-L3892
> 
> I would suggest that you create a copy of these tests with a new name, e.g. `test_scaling_rslora`, and then change the testing code to use rslora. Change the expected values accordingly. LMK if you have questions.

Thank you @BenjaminBossan, for reviewing it. I will add the test file as soon as possible.

BenjaminBossan:
@tanuj-rai When the PR is ready for another review, please ping me.

tanuj-rai:
@BenjaminBossan I'm trying to run the test `test_scaling_simple` in `tests/test_scaling_rslora.py`, but I get the error: `fixture 'model' not found`. I'm unsure how to define or locate the `model` fixture required for this test. 

```
 pytest tests/test_scaling_rslora.py
============================= test session starts =============================
platform win32 -- Python 3.10.11, pytest-8.4.1, pluggy-1.6.0
rootdir: C:\Users\1234\peft
configfile: pyproject.toml
plugins: anyio-4.4.0, cov-7.0.0
collected 1 item

tests\test_scaling_rslora.py E                                           [100%]

=================================== ERRORS ====================================
______________ ERROR at setup of TestScaling.test_scaling_simple ______________
file C:\Users\1234\peft\tests\test_scaling_rslora.py, line 5
      def test_scaling_simple(self, model):
E       fixture 'model' not found
>       available fixtures: anyio_backend, anyio_backend_name, anyio_backend_options, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, capteesys, cov, doctest_namespace, monkeypatch, no_cover, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.

C:\Users\1234\peft\tests\test_scaling_rslora.py:5
=============================== tests coverage ================================
______________ coverage: platform win32, python 3.10.11-final-0 ___

`

**Comment by BenjaminBossan** [comment]
@tanuj-rai The reason why you cannot run your test is because you put it in a separate file. If you place it below the other tests I referenced above, it will work. Why? Because in that test class, a so called [`pytest.fixture`](https://docs.pytest.org/en/6.2.x/fixture.html#what-fixtures-are) with the name `model` is defined ([here](https://github.com/huggingface/peft/blob/5d97453235d3c3073b059995f56260d364d44e83/tests/test_initialization.py#L3832-L3836)), which will be automatically added to the test when it's called.

When you move your tests over there, you should give the function a unique name, e.g. `test_scaling_with_rslora`. Then, you want to run only this specific test, call `pytest tests/test_initialization.py -k test_scaling_with_rslora`.

**Comment by tanuj-rai** [comment]
> @tanuj-rai The reason why you cannot run your test is because you put it in a separate file. If you place it below the other tests I referenced above, it will work. Why? Because in that test class, a so called [`pytest.fixture`](https://docs.pytest.org/en/6.2.x/fixture.html#what-fixtures-are) with the name `model` is defined ([here](https://github.com/huggingface/peft/blob/5d97453235d3c3073b059995f56260d364d44e83/tests/test_initialization.py#L3832-L3836)), which will be automatically added to the test when it's called.
> 
> When you move your tests over there, you should give the function a unique name, e.g. `test_scaling_with_rslora`. Then, you want to run only this specific test, call `pytest tests/test_initialization.py -k test_scaling_with_rslora`.

Ahh, I see. Thank you @BenjaminBossan. That was my mistake. I’ve removed it now and added the new test function 'test_scaling_with_rslora' into test_initialization.py.

**Comment by tanuj-rai** [comment]
When I am running the pytest. It got deselected.

```
$ pytest tests/test_initialization.py -k test_scaling_with_rslora
============================= test session starts =============================
platform win32 -- Python 3.10.11, pytest-8.4.1, pluggy-1.6.0
rootdir: C:\Users\1234\peft
configfile: pyproject.toml
plugins: anyio-4.4.0, cov-7.0.0
collected 241 items / 241 deselected / 0 selected

=============================== tests coverage ================================
______________ coverage: platform win32, python 3.10.11-final-0 __
`

BenjaminBossan:
> When I am running the pytest. It got deselected.

Strange, when I run the exact same command on your branch, the test runs as expected. It does, however, raise an error because `math` needs to be imported:

```
>       expected = [lora_alpha / math.sqrt(rank)] * n_layers
E       NameError: name 'math' is not defined. Did you forget to import 'math'
```

tanuj-rai:
> pytest tests/test_initialization.py -k test_scaling_with_rslora

@BenjaminBossan! As you said, it is running fine for you. But I don't know why it skips it when I run pytest.  What should I do now to fix it? I have added `import math`.

tests/test_initialization.py
```diff
@@ -3891,6 +3892,43 @@ def test_scaling_simple(self, model):
         expected = [2.0] * n_layers
         assert scalings == expected
 
+    def test_scaling_with_rslora(self, model):
+        n_layers = 5
+        rank, lora_alpha = 8, 16
+        config = LoraConfig(
+            r=rank,
+            lora_alpha=lora_alpha,
+            target_modules=["k_proj"],
```
BenjaminBossan, line 3902: We need to pass `use_rslora=True`, right?

tests/test_initialization.py
```diff
@@ -3891,6 +3892,43 @@ def test_scaling_simple(self, model):
...
+        )
+        model = get_peft_model(model, config)
+        scalings = self.get_scalings(model)
+        expected = [lora_alpha / math.sqrt(rank)] * n_layers
+        assert scalings == expected
+    
+        # double
+        self.scale_layer(model, 2)
+        scalings = self.get_scalings(model)
+        expected = [4.0] * n_layers
```
BenjaminBossan: Do we really expect 4 here? Or not rather `2 * lora_alpha / math.sqrt(rank)`? Similarly for the values below, I think they need to be adjusted.

tanuj-rai: Thank you, @BenjaminBossan, for suggesting improvements. I have updated the code. Please take a look when you have a moment.

tests/test_initialization.py
```diff
@@ -3891,6 +3892,44 @@ def test_scaling_simple(self, model):
...
+        self.scale_layer(model, 2)
+        scalings = self.get_scalings(model)
+        expected = [2 * lora_alpha / math.sqrt(rank)] * n_layers
+        assert scalings == expected
+
+        # back to original
+        self.unscale_layer(model, None)
+        scalings = self.get_scalings(model)
+        expected = [lora_alpha / math.sqrt(rank)] * n_layers
+        assert scalings == expected
```
BenjaminBossan, line 3919: When I run these tests, I get an error here:

> assert [2.0, 2.0, 2.0, 2.0, 2.0] == [5.6568542494...5685424949238]

However, I think this time, the expected values are correct and the actual scalings are wrong. This is because the `unscale_layer` method does not take rslora into account when passing `scale=None`. Let's fix that one too, in a similar way to `set_scale`.

tanuj-rai: Sure @BenjaminBossan. I will fix the `unscale_layer` as soon as possible. Apologies for the delayed response.

tanuj-rai:
Hi @BenjaminBossan, I’ve updated unscale_layer to correctly handle rslora by resetting to lora_alpha / sqrt(r) when scale=None. Please take a look and let me know if anything needs to be updated. 

```
def unscale_layer(self, scale: Optional[float | int] = None) -> None:
        if active_adapter not in self.lora_A:
            continue

        if scale is None:
            if self.use_rslora.get(active_adapter, False):
                self.scaling[active_adapter] = self.lora_alpha[active_adapter] / math.sqrt(self.r[active_adapter])
            else:
                self.scaling[active_adapter] = self.lora_alpha[active_adapter] / self.r[active_adapter]
        else:
            self.scaling[active_adapter] /= scale
`

BenjaminBossan:
Thanks for the update @tanuj-rai. The tests are passing now. If the PR is ready from your side, please convert it from draft to "ready for review".

tanuj-rai:
> Thanks for the update @tanuj-rai. The tests are passing now. If the PR is ready from your side, please convert it from draft to "ready for review".

Thanks, @BenjaminBossan for reviewing it. I have converted it from draft to "ready for review".

PR #2778, opened 2025-09-08
Support dataclass model configs

githubnemo:
LeRobot uses dataclasses to manage policy configs. If we want to support LeRobot policy fine-tuning it'd be easiest to support these configs in `get_model_config`.

While it is possible to fix this on LeRobot's side (add a `to_dict` implementation to the config classes) I think it'd be cleaner to support it on our side since the cost is relatively low and dataclasses are getting more popular anyway.

Thanks @xliu0105 for raising this issue and proposing a fix (supersedes #2641).

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -14,6 +14,7 @@
 from __future__ import annotations
 
 import copy
+import dataclasses
 import os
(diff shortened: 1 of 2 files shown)


BenjaminBossan (approved):
LGTM, can be merged if CI is green.

Issue #2760, opened 2025-08-28
Integrate DeLoRA


mwbini:
This request proposes integrating DeLoRA (Decoupled Low-rank Adaptation), as described in the ICLR25 accepted paper

paper: https://arxiv.org/abs/2503.18225
code: https://github.com/ExplainableML/DeLoRA

### Motivation
DeLoRA tackles finetuning in a Frobenius-norm bounded setup: this allows to prevent divergence from the pretrained model, effectively decoupling the learning of angles and magnitudes.

This is done by 
- normalization of the BA low-rank matrices, which bound the updates' Frobenius norm
- (learnable) scaling λ, which controls the update's boundary/magnitude
- layer-wise scaling of ||W||, to adapt each update's norm to the original weights' norm (mimicking multiplicative finetuning).

<img width="400" height="235" alt="Image" src="https://github.com/user-attachments/assets/07750438-5855-4319-b94a-d802393b42db" />

The method might feel quite similar to DoRA (given the similar target of decoupling angular from magnitude learning), however it presents key differences: 
- DoRA applies normalization and scaling operations on the fully finetuned weights ($W + \Delta W$) 
- the normalization operation is performed on the column space of the weight matrices

Conversely DeLoRA 
- introduces the normalization and scaling operations directly on the weight updates $\Delta W$ (more effectively preventing divergence from the pretrained model) 
- normalizes the inner low-dimensional space, which implicitly enforces a Frobenius-norm boundary. While, in theory, setting the scaling parameter λ as learnable does not prevent divergence, this does not happen in practice (see figure below, showing performance and norms at the varying of the learning rate)

<img width="4167" height="1160" alt="Image" src="https://github.com/user-attachments/assets/02d81c57-e947-487e-b83c-0f16bac48ba5" />

As a result, DeLoRA is able to achieve better decoupling and, consequently, robustness. In addition, one can arbitrarily initialize the parameter λ to achieve countless norm-bounded variations

### Your Contribution
The implementation in https://github.com/ExplainableML/DeLoRA is based on peft, and we would be pleased submit a pull request, welcoming any suggestions or guidance on this.

BenjaminBossan:
Thank you for suggesting to add DeLoRA to PEFT. The method looks interesting and we'd be happy to have it integrated. When you work on your PR, you can check past recent additions (e.g. #2577) as a template. Also, don't hesitate to create a quick, incomplete draft PR (e.g. without tests) for early feedback.

[...]

PR #2780, opened 2025-09-10
FEAT add DeLoRA

mwbini:
I implemented a first version, based on #2760


docs/source/package_reference/delora.md
```diff
@@ -0,0 +1,36 @@
...
+rendered properly in your Markdown viewer.
+
+-->
+
+# DeLoRA: Decoupled Low-rank Adaptation
+
+[DeLoRA](https://huggingface.co/papers/2503.18225) is a parameter-efficient fine-tuning technique that leverages effectively decouples the learning of angles and magnitudes.
+
+Note:
+- use larger learning rate than standard LoRA variants, no weight decay on the BA matrices
```
BenjaminBossan: Maybe I missed this in the paper, but does this mean that I can use a higher learning rate and training converges faster?

docs/source/package_reference/delora.md
```diff
@@ -0,0 +1,36 @@
...
+- we utilize alpha as the initial value of the boundary parameter. Using different alphas for different weights is possible
```
BenjaminBossan: This sentence needs more explanation. What is alpha, what are boundary parameters?

src/peft/tuners/delora/config.py
```diff
@@ -0,0 +1,147 @@
...
+@dataclass
+class DeLoRAConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`DeLoRAModel`].
+
+    Args:
+        r (`int`):
+            DeLoRA rank.
+        alpha (`int`):
+            The alpha parameter for DeLoRA boundary.
```
BenjaminBossan: Just curious: IIUC, this corresponds to lambda in the paper, right? Why was it renamed? Because the keyword is protected in Python or to make it more similar to `lora_alpha`? But this alpha has a different role than alpha has for LoRA.

src/peft/tuners/delora/config.py
```diff
@@ -0,0 +1,147 @@
...
+        exclude_modules (`Optional[Union[List[str], str]]`):
+            The names of the modules to not apply the adapter. When passing a string, a regex match will be performed.
+            When passing a list of strings, either an exact match will be performed or it is checked if the name of the
+            module ends with any of the passed strings.
+        bias (`str`):
+            Bias type for DeLoRA. Can be 'none' or 'all'. If 'all', the corresponding biases will be updated during
+            training. Be aware that this means that, even when disabling the adapters, the model will not produce the
+            same output as the base model would have without adaptation.
+        init_weights (`bool`):
+                Whether to perform initialization of adapter weights. If `True` (default): adapters are initialized
```
BenjaminBossan: Let's indent this correctly.

src/peft/tuners/delora/config.py
```diff
@@ -0,0 +1,147 @@
...
+        default=None,
+        metadata={"help": "List of module names or regex expression of the module names to exclude from DeLoRA."},
+    )
+    bias: str = field(default="none", metadata={"help": "Bias type for DeLoRA. Can be 'none' or 'all'"})
+    init_weights: bool = field(
+        default=True,
+        metadata={
+            "help": (
+                "Whether to initialize the weights of the DeLoRA layers with their default initialization. Don't change "
+                "this setting, except if you know exactly what you're doing."
```
BenjaminBossan: I think the text in the docstring explains this better. It's fine to just copy-paste the argument docstrings here.

src/peft/tuners/delora/config.py
```diff
@@ -0,0 +1,147 @@
...
+        self.target_modules = (
+            set(self.target_modules) if isinstance(self.target_modules, list) else self.target_modules
+        )
+        # if target_modules is a regex expression, then layers_to_transform should be None
+        if isinstance(self.target_modules, str) and self.layers_to_transform is not None:
+            raise ValueError("`layers_to_transform` cannot be used when `target_modules` is a str.")
+
+        # check for layers_to_transform and layers_pattern
+        if self.layers_pattern and not self.layers_to_transform:
+            raise ValueError("When `layers_pattern` is specified, `layers_to_transform` must also be specified. ")
```
BenjaminBossan, line 154: Would it make sense to check if `init_weights` and `use_residual_init` are compatible?

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+        "r",
+        "delora_dropout",
+        # Persisted initial copies and norms (use delora_ prefix so they are saved)
+        "delora_initial_A",
+        "delora_initial_B",
+        "delora_initial_alpha",
+        "delora_w_norm",
+        "delora_initial_w_norm",
+        # Misc flag
+        "use_residual_init",
```
BenjaminBossan: I think there is no need for `"use_residual_init"` here.

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+
+    def __init__(self, base_layer: nn.Module, **kwargs) -> None:
+        self.base_layer = base_layer
+        self.r = {}
+        self.delora_dropout = nn.ModuleDict({})
+        self.delora_A = nn.ParameterDict({})
+        self.delora_B = nn.ParameterDict({})
+        self.delora_alpha = nn.ParameterDict({})
+        # Store flag from kwargs (defaults to True if not provided)
+        self.use_residual_init = bool(kwargs.get("use_residual_init", True))
```
BenjaminBossan: Let's not do `kwargs.get(...)`, just put `use_residual_init: bool = True` into the `__init__` arguments.

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+        alpha_ = alpha.to(comp_dtype)
+        w_norm_ = w_norm.to(comp_dtype)
+        An = torch.clamp(A_.norm(dim=1), min=1e-12)
+        Bn = torch.clamp(B_.norm(dim=0), min=1e-12)
+        diag = torch.diag_embed((alpha_ / r) / (An * Bn))
+        delta = B_ @ diag @ A_
+        delta = delta * w_norm_.unsqueeze(0)
+        return delta.to(A.dtype, copy=False)
+
+    def get_delta_weight(self, adapter: str) -> torch.Tensor:
```
BenjaminBossan, line 73: There is a lot of similarity between this method and `_compute_delta_from_params`, I wonder if, we a little refactoring of `_compute_delta_from_params`, it could be reused here. WDYT?

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+
+        A = self.delora_A[adapter].to(dtype=comp_dtype, device=device)
+        B = self.delora_B[adapter].to(dtype=comp_dtype, device=device)
+        alpha = self.delora_alpha[adapter].to(dtype=comp_dtype, device=device)
+        An = torch.clamp(A.norm(dim=1), min=1e-12)
+        Bn = torch.clamp(B.norm(dim=0), min=1e-12)
+        diag = torch.diag_embed((alpha / self.r[adapter]) / (An * Bn))  # (r,r)
+        delta = B @ diag @ A  # (out,in)
+
+        if (not self.use_residual_init) and (adapter in self.delora_initial_A and adapter in self.delora_initial_B):
```
BenjaminBossan: Could you please add a comment about why this conditional is needed?

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+            alpha0 = self.delora_initial_alpha[adapter].to(dtype=comp_dtype, device=device).squeeze(0)
+            A0n = torch.clamp(A0.norm(dim=1), min=1e-12)
+            B0n = torch.clamp(B0.norm(dim=0), min=1e-12)
+            diag0 = torch.diag_embed((alpha0 / self.r[adapter]) / (A0n * B0n))
+            delta = delta - (B0 @ diag0 @ A0)
+        # scale by a fixed per-adapter reference base weight norm when available (avoids caching for exact unmerge)
+        if adapter in self.delora_w_norm:
+            w_norm = self.delora_w_norm[adapter].to(dtype=comp_dtype, device=device)
+        else:
+            # fallback: compute current base weight norm on-the-fly
```
BenjaminBossan: Under what conditions could `self.delora_w_norm` not contain the adapter?

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+        init_weights: bool | Literal["kaiming_uniform", "gaussian", "xavier_uniform"] = True,
+        use_residual_init: Optional[bool] = None,
+    ) -> None:
+        if r <= 0:
+            raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
+
+        self.r[adapter_name] = r
+        if use_residual_init is not None:
+            # allow toggling via update for already-wrapped layers
+            self.use_residual_init = bool(use_residual_init)
```
BenjaminBossan: I don't think we need casting to `bool`, the argument is documented as being of `bool` type and we generally should assume that users pass the correct type (otherwise, we'd have to carefully check each possible argument).

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+
+        weight = self.get_base_layer().weight
+        self.delora_A[adapter_name] = nn.Parameter(torch.empty(r, self.in_features, device=weight.device, dtype=torch.float32))
+        self.delora_B[adapter_name] = nn.Parameter(torch.empty(self.out_features, r, device=weight.device, dtype=torch.float32))
+        self.delora_alpha[adapter_name] = nn.Parameter(torch.full((1,), float(alpha), device=weight.device, dtype=torch.float32))
+
+        # initialize and store initial copies
+        self.reset_delora_parameters(adapter_name, init_weights, alpha)
+        self._move_adapter_to_device_of_base_layer(adapter_name)
+        self.set_adapter(self.active_adapters)
```
BenjaminBossan: There was a recent change to `set_adapter` in #2765, with a new argument `inference_mode` being added. Could you please do that too? For example, check the change here: https://github.com/huggingface/peft/pull/2765/files#diff-4a806218d334935f1dcc647b35501e73dbe3bc7804d4c8d1296d260fff060cfe

src/peft/tuners/delora/config.py
```diff
@@ -0,0 +1,147 @@
...
+            the target modules manually.
+        exclude_modules (`Optional[Union[List[str], str]]`):
+            The names of the modules to not apply the adapter. When passing a string, a regex match will be performed.
+            When passing a list of strings, either an exact match will be performed or it is checked if the name of the
+            module ends with any of the passed strings.
+        bias (`str`):
+            Bias type for DeLoRA. Can be 'none' or 'all'. If 'all', the corresponding biases will be updated during
+            training. Be aware that this means that, even when disabling the adapters, the model will not produce the
+            same output as the base model would have without adaptation.
+        init_weights (`bool`):
```
BenjaminBossan, line 52: There are more options than just `True` or `False` here, right? Let's document all of them.

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+                # subtract initial delta using the reference norm (only if not on meta device)
+                if self.delora_initial_A[adapter_name].device.type != "meta" and w.device.type != "meta":
+                    with torch.no_grad():
+                        # subtract exact initial delta so effective initial effect is zero
+                        A0 = self.delora_initial_A[adapter_name].to(dtype=w.dtype, device=w.device)
+                        B0 = self.delora_initial_B[adapter_name].to(dtype=w.dtype, device=w.device)
+                        alpha0 = self.delora_initial_alpha[adapter_name].to(dtype=w.dtype, device=w.device).squeeze(0)
+                        wnorm0 = self.delora_initial_w_norm[adapter_name].to(dtype=w.dtype, device=w.device)
+                        delta0 = self._compute_delta_from_params(A0, B0, alpha0, self.r[adapter_name], wnorm0)
+                        self.get_base_layer().weight.data.sub_(delta0)
```
BenjaminBossan: If we add two DeLoRA adapters with `use_residual_init=True` to the model, would the 2nd one not subtract its delta again from the base weight? Would this not result in the base weight being incorrect for the first adapter?

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+            warnings.warn("Already unmerged. Nothing to do.")
+            return
+        while len(self.merged_adapters) > 0:
+            active_adapter = self.merged_adapters.pop()
+            if active_adapter in self.delora_A.keys():
+                with torch.no_grad():
+                    delta_weight = self.get_delta_weight(active_adapter)
+                    self.get_base_layer().weight.data.sub_(delta_weight)
+        # If we had subtracted initial delta(s) at setup, restore them now to get back the original base weights
+        # This should not be done for a model that has just been loaded from a checkpoint, as the state is already correct.
```
BenjaminBossan: I don't understand why this is needed, how is the modification of the base weights, which happens at layer initialization, related to unmerging?

Let's imagine a use case where a user does the following:

1. Create a DeLoRA model
2. Base weights are being modified under the hood
3. Train for a few steps
4. To perform some evals, they merge the weights
5. Once they're finished, they unmerge again
6. Now the base weights are reset to their original value under the hood
7. Then the user wants to continue training for more steps

Would this not lead to a mismatch between the DeLoRA weights and the base weights?

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+        if self.use_residual_init and len(self._initials_applied) > 0:
+            w = self.get_base_layer().weight
+            for adapter in list(self._initials_applied):
```
BenjaminBossan: No need to cast to list.

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+                    and (adapter in self.delora_initial_B)
+                    and (adapter in self.delora_initial_w_norm)
+                ):
+                    A0 = self.delora_initial_A[adapter].to(dtype=w.dtype, device=w.device)
+                    B0 = self.delora_initial_B[adapter].to(dtype=w.dtype, device=w.device)
+                    alpha0 = self.delora_initial_alpha[adapter].to(dtype=w.dtype, device=w.device).squeeze(0)
+                    wnorm0 = self.delora_initial_w_norm[adapter].to(dtype=w.dtype, device=w.device)
+                    delta0 = self._compute_delta_from_params(A0, B0, alpha0, self.r[adapter], wnorm0)
+                    self.get_base_layer().weight.data = self.get_base_layer().weight.data + delta0
+                self._initials_applied.discard(adapter)
```
BenjaminBossan: Since we checked earlier that `adapter` is in `_initials_applied`, there is no need to use `discard`, you can use `remove`.

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+        elif self.merged:
+            result = self.base_layer(x, *args, **kwargs)
+        else:
+            if not self.active_adapters:
+                return self.base_layer(x, *args, **kwargs).to(previous_dtype)
+
+            base_out = self.base_layer(x, *args, **kwargs)
+            add_out = torch.zeros_like(base_out)
+
+            for a in self.active_adapters:
```
BenjaminBossan: Let's not use `a` but `adapter` or `adapter_name`.

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+                comp_dtype = torch.float32 if out_dtype in (torch.float16, torch.bfloat16) else out_dtype
+
+                A = self.delora_A[a].to(dtype=comp_dtype, device=device)
+                B = self.delora_B[a].to(dtype=comp_dtype, device=device)
+                alpha = self.delora_alpha[a].to(dtype=comp_dtype, device=device)
+
+                if a in self.delora_w_norm:
+                    w_norm = self.delora_w_norm[a].to(dtype=comp_dtype, device=device)
+                else:
+                    w = base_w
```
BenjaminBossan: Why assign a new variable `w` here?

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,630 @@
...
+                # 2. h @ diag
+                An = torch.clamp(A.norm(dim=1), min=1e-12)
+                Bn = torch.clamp(B.norm(dim=0), min=1e-12)
+                scaling = (alpha / self.r[a]) / (An * Bn)
+                h = h * scaling
+
+                # 3. h @ B.T
+                h = nn.functional.linear(h, B)
+
+                if (not self.use_residual_init) and (a in self.delora_initial_A and a in self.delora_initial_B):
```
BenjaminBossan: Same comment as above for `get_delta_weight`.

BenjaminBossan (changes requested):
Thanks a lot for this PR to add DeLoRA to PEFT. The PR is already quite advanced and in a nice state, good job with that.

I did a review of the PR and added a few comments, please check those. But while working on this, I also noticed some more general areas for discussion, which should be addressed first:

## Naming

In general, I would suggest to rename variables with `DeLoRA` to `Delora` or `DeLora`. Not only is that easier to type, it's also consistent with the other names, e.g. `LoraConfig`.

## Initialization

I think a lot of complexity of the implementation comes from the initialization with `use_residual_init=True`. This is the reason why we need the `delora_initial_*` parameters and update methods like `_load_from_state_dict`, right? I wonder if we can think of a way to significantly reduce this complexity.

First of all, I still need to get a good understanding why this is required in the first place. Do I get it right that normally, initializing DeLoRA cannot be done in a way that makes it an identity transform? So there is no choice of `delora_A` and `delora_B` that would make their contribution zero? For me, this would be the ideal solution.

Second, how important is it to have the `use_residual_init=True` option? Is it better for training or could we just drop it entirely? I know that this would require some adjustments to the unit tests, as some tests assume that the adapter is an identity transform without training. But these adjustments can be done and it is not a requirement for PEFT methods to have an initialization option that comes down to identity.

If this option is not strictly needed for performance reasons but mainly to satisfy tests, I would rather remove it, eliminate a lot of the complexity associated with it, and then update the tests.

## Device and dtype handling

There is a lot of code that deals with moving the DeLoRA parameters to the correct device and dtype, such as:

```python
A = self.delora_A[a].to(dtype=comp_dtype, device=device)
B = self.delora_B[a].to(dtype=comp_dtype, device=device)
alpha = self.delora_alpha[a].to(dtype=comp_dtype, device=device)
```

Would it not make sense to initialize them with the correct device and dtype right from the start (as is normally done for other PEFT methods) instead of casting and moving them each time, which can be costly? Or is this some sort of memory optimization?

## Other

Those are the big points I found while reviewing and they should be the focus for the next iteration. For the final PR, I would also like to see an example (it could be a copy of an existing example with modifications to use DeLoRA) and setting for the [PEFT method comparison suite](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA).

Finally, please always call `make style` before pushing your changes.

mwbini:
Hi @BenjaminBossan, thank you for thorough review! I made a few changes following your suggestions. I will address below your main concerns

**Naming**: I agree, I now modified everything to `Delora*`

**Initialization**: the method requires to be identity-initialized to work properly. To do this while avoiding divisions by zeros in the normalization (which could lead to instabilities), the idea is to add and subtract the same initial weights W + BA - BA and keep one of the copies frozen (`delora_initial_A`, `delora_initial_B`). With `use_residual_init=False`, the frozen copies are directly used in the forward pass, and merging/unmerging and others should work similarly to other peft methods (on the negative, these initial weights need to be kept on device, and require extra computations). With `use_residual_init=True` instead, W - BA are merged once at the beginning, and the forward pass simply uses the trainable `delora_A` `delora_B`. This saves some computations, and avoids keeping on device `delora_initial_A`, `delora_initial_B`, but indeed adds several complications when merging/unmerging etc, requiring before each operation to add back the initial BA (offloaded to cpu) from the base weights W (so to correctly go back to the base model). I tried to describe the role of `use_residual_init` a bit more in the config file

**Device and dtype handling**: a lot of these were indeed redundant/unnecessary. I simplified and tried to remove most of them now

docs/source/package_reference/delora.md
mwbini: From my experiments convergence speed seemed on-par with lora. However the learning rate needs to be quite high or convergence speed might be significantly lower.

src/peft/tuners/delora/config.py
mwbini: Yes I didn't use lambda because it was protected. But indeed using alpha could be confusing. For consistency with the paper I changed it to `delora_lambda` and `lambda_` now

src/peft/tuners/delora/config.py
mwbini: To be consistent with most methods, I now removed other alternatives and only left the default one, selected with `init_weights=True`

BenjaminBossan: I'd say consistency is not so critical here, some PEFT methods like LoRA have many init options. I'd rather approach this from the point of view whether those other initialization schemes are useful or not. If you have no evidence that they help, I'd agree that removing them is the best option.

src/peft/tuners/delora/config.py
```diff
@@ -65,15 +69,20 @@ class DeLoRAConfig(PeftConfig):
         rank_pattern (`dict`):
             The mapping from layer names or regexp expression to ranks which are different from the default rank
             specified by `r`. For example, `{'^model.decoder.layers.0.encoder_attn.k_proj': 16}`.
-        alpha_pattern (`dict`):
-            The mapping from layer names or regexp expression to alphas which are different from the default alpha
-            specified by `alpha`. For example, `{'^model.decoder.layers.0.encoder_attn.k_proj': 16}`.
+        lambda_pattern (`dict`):
```
BenjaminBossan, line 66: There is no actual argument `lambda_pattern` (or `alpha_pattern`) on the config. I'm fine with not having it, up to you if you think it's useful to add it.

docs/source/package_reference/delora.md
BenjaminBossan: Got it. If you have concrete numbers, this would be helpful to users. E.g. "A good start for the lr is 1e-3" or "Use roughly 10x higher lr than with LoRA".

docs/source/package_reference/delora.md
BenjaminBossan: Since you renamed the argument, let's use "lambda" here. Also, whether different lambdas can be used for different weights depends on whether `lambda_pattern` is implemented (see my other comment).

src/peft/tuners/delora/layer.py
BenjaminBossan: I still think we don't need `w` here, in the line below we can just do: `w_norm = torch.norm(base_w.data, dim=0).detach()`

src/peft/tuners/delora/config.py
mwbini: I added this check in the new version

src/peft/tuners/delora/layer.py
mwbini: Thanks for the suggestion, I refactored the code to reuse the same `_compute_delta` function

src/peft/tuners/delora/layer.py
mwbini: Indeed this was mistakenly still in the code. I now removed it

src/peft/tuners/delora/layer.py
mwbini: I see. I first added this in case one wanted to unmerge and revert back to base weights. 

Since this might be a secondary usage, I now added an additional argument `unapply_initials` to unmerge (which by default is False) that can further revert to base weights. I am not sure if this is necessary, and I am okay with removing it if it's not

src/peft/tuners/delora/config.py
mwbini: Thanks for pointing this out, I think it would be useful to have it. I now added it

mwbini:
Thank you for the feedbacks. Now I should have addressed most of the comments, please let me know if something is missing.

Regarding the initialization I am not sure what could be the best approach. Maybe B=0 could still be used by setting a large enough min value in torch.clamp... As a next step, I will try to setup some experiments and check this zero initialization, the overhead of using use_residual_init=False, etc.

Regarding using multiple adapters, it would be indeed not possible with use_residual_init=True. For now, I added a check in _check_new_adapter_config that should check for this

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,610 @@
...
+
+        else:
+            if adapter_name in self.delora_initial_A:
+                del self.delora_initial_A[adapter_name]
+            if adapter_name in self.delora_initial_B:
+                del self.delora_initial_B[adapter_name]
+            if adapter_name in self.delora_initial_lambda:
+                del self.delora_initial_lambda[adapter_name]
+
+    def _move_adapter_to_device_of_base_layer(self, adapter_name: str, device: Optional[torch.device] = None) -> None:
```
BenjaminBossan: It seems that we cannot use the `_move_adapter_to_device_of_base_layer` method from the parent class (`BaseTunerLayer`). Could you please add a short comment on what is different here?

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,610 @@
...
+        """
+        Move the adapter of the given name to the device of the base layer.
+        Override to handle meta tensors properly.
+        """
+        if device is None:
+            base_layer = self.get_base_layer()
+            if isinstance(base_layer, nn.MultiheadAttention):
+                base_layer = base_layer.out_proj
```
BenjaminBossan: MHA is not supported by DeLoRA so this can be removed.

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,610 @@
...
+                                self.delora_initial_w_norm[adapter_name].to(w.device),
+                            )
+                            w.data += fresh_delta.to(w.dtype)
+
+        # Create a filtered state dict with only keys that belong to this module
+        module_state_dict = {}
+        for key, value in state_dict.items():
+            if key.startswith(prefix):
+                # Remove the prefix to get the relative key
+                relative_key = key[len(prefix) :]
```
BenjaminBossan: `relative_key = key.removeprefix(prefix)` should do the job.

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,610 @@
...
+                module_state_dict[relative_key] = value
+
+        # Manually load ParameterDict and BufferDict parameters
+        for param_name in ["delora_A", "delora_B", "delora_lambda"]:
```
BenjaminBossan: We can use `self.adapter_layer_names` here, right?

src/peft/tuners/delora/model.py
```diff
@@ -0,0 +1,300 @@
...
+        Raise a ValueError if there is something wrong with the config or if it conflicts with existing adapters.
+
+        """
+        # TODO: there should be a check if any of the existing adapters actually has bias != "none", or else the check
+        # does not fully correspond to the error message.
+        if (len(self.peft_config) > 1) and (config.bias != "none"):
+            raise ValueError(
+                f"{self.__class__.__name__} supports only 1 adapter with bias. When using multiple adapters, "
+                "set bias to 'none' for all adapters."
+            )
```
BenjaminBossan: Let's replace this with:
```suggestion
        super()._check_new_adapter_config(config)
```

src/peft/tuners/delora/model.py
```diff
@@ -0,0 +1,300 @@
...
+        # if there are multiple adapters, residual init cannot be used
+        if (len(self.peft_config) > 1) and (config.use_residual_init):
+            raise ValueError(
+                f"{self.__class__.__name__} supports only 1 adapter with residual init. When using multiple adapters, "
+                "set use_residual_init to False for all adapters."
+            )
+
+    @staticmethod
```
BenjaminBossan: We recently had a PR (#2771) that allows us to greatly reduce the number of methods required to implement here. You can simply remove the following methods:

- `_check_target_module_exists`
- `_mark_only_adapters_as_trainable`
- `__getattr__`
- `get_peft_config_as_dict`
- `_set_adapter_layers`
- `enable_adapter_layers`
- `disable_adapter_layers`
- `set_adapter`
- `_prepare_adapter_config`
- `_unload_and_optionally_merge`
- `merge_and_unload`
- `unload`

On the flip side, you need to add:

```python
    tuner_layer_cls = DeloraLayer
    target_module_mapping = TRANSFORMERS_MODELS_TO_DELORA_TARGET_MODULES_MAPPING
```

as class attributes, i.e. directly below `prefix: str = "delora_"`.

For this to work, you also need to merge with/rebase on the latest main branch.

tests/test_initialization.py
```diff
@@ -1846,6 +1847,98 @@ def test_road_with_conv2d_layer(self):
...
+    def test_merge_and_unload_same_outputs(self, data):
+        base = self.get_model()
+        cfg = DeloraConfig(target_modules=["lin0"], r=8, lambda_=8)
+        model = get_peft_model(base, cfg)
+        y_before = model(data)
+
+        merged = model.merge_and_unload()
+        merged.eval()
+        y_after = merged(data)
+        assert torch.allclose(y_before, y_after, atol=1e-6, rtol=1e-6)
```
BenjaminBossan: I think this should already be well covered by the existing tests and can be removed here.

src/peft/tuners/delora/model.py
```diff
@@ -0,0 +1,300 @@
...
+        """
+        # TODO: there should be a check if any of the existing adapters actually has bias != "none", or else the check
+        # does not fully correspond to the error message.
+        if (len(self.peft_config) > 1) and (config.bias != "none"):
+            raise ValueError(
+                f"{self.__class__.__name__} supports only 1 adapter with bias. When using multiple adapters, "
+                "set bias to 'none' for all adapters."
+            )
+        # if there are multiple adapters, residual init cannot be used
+        if (len(self.peft_config) > 1) and (config.use_residual_init):
```
BenjaminBossan: Hmm, would it work if I first add a DeLoRA adapter with `use_residual_init=False` and then a DeLoRA adapter with `use_residual_init=True`? I think it should work but this check would result in an error. IIUC, the correct check is that there cannot be more than one config with `use_residual_init=True`.

tests/test_initialization.py
```diff
@@ -1846,6 +1847,98 @@ def test_road_with_conv2d_layer(self):
...
+
+    def test_no_use_residual_init_keeps_output(self, data):
+        # With use_residual_init=False, the implementation compensates by subtracting initial delta in get_delta
+        base = self.get_model()
+        y_base = base(data)
+        cfg = DeloraConfig(target_modules=["lin0"], r=8, lambda_=8, use_residual_init=False)
+        model = get_peft_model(base, cfg)
+        y_peft = model(data)
+        assert torch.allclose(y_base, y_peft, atol=1e-6, rtol=1e-6)
+
```
BenjaminBossan, line 2157: Let's add a test adding a second adapter (via `model.add_adapter`) that has `use_residual_init=True` raises the "supports only 1 adapter with residual init" error.

BenjaminBossan (changes requested):
Thanks for the updates. I found a couple of new issues, but nothing big. Also, please merge with/rebase on the latest PEFT main branch and resolve any merge conflicts that may occur.

> Regarding the initialization I am not sure what could be the best approach. Maybe B=0 could still be used by setting a large enough min value in torch.clamp... As a next step, I will try to setup some experiments and check this zero initialization, the overhead of using use_residual_init=False, etc.

Great, LMK if you find something.

mwbini:
Hi @BenjaminBossan, thank you for your review. I now pushed several changes (last two commits) and rebased to the latest version of the code. The main modification is that I removed the residual initialization in favor of the zero initialization of B (making DeLoRA init similar to LoRA).
I tested if setting a high enough minimal value in torch.clamp in the A/B normalization would avoid creating issues (by testing on both GLUE and MetaMathQA), and it seems that indeed issues do not appear, and performance are very similar to using the residual initialization. For this min value I tried different choices, and since nans can still appear in float16 for values below 1e-4, while leading to no significant performance differences in full precision, I set 1e-4 as default value.

Because of this, given all the complications that it was creating (requiring special version of functions, or limiting its usability with multiple adapters), I removed the residual initialization and all related modifications (so use_residual_init, delora_initial_A/B, etc). So now the method should be much simpler to handle. Please let me know what do you think about this, if maybe it was better to keep a residual initialization version, or if it's ok to keep it simple.

I also removed unnecessary methods and tests as you suggested. Since I removed use_residual_init and related methods, some modifications were not needed anymore. I also added some configs for experiments on MetaMathQA.

src/peft/tuners/delora/layer.py
```diff
@@ -0,0 +1,270 @@
...
+            return
+
+        if init_weights is True:
+            nn.init.kaiming_uniform_(self.delora_A[adapter_name], a=math.sqrt(5))
+            nn.init.zeros_(self.delora_B[adapter_name])
+        else:
+            nn.init.kaiming_uniform_(self.delora_A[adapter_name], a=math.sqrt(5))
+            nn.init.kaiming_uniform_(self.delora_B[adapter_name], a=math.sqrt(5))
+
+        self.delora_lambda[adapter_name].data.fill_(float(delora_lambda))
```
BenjaminBossan, line 141: This line is redundant with https://github.com/huggingface/peft/pull/2780/files#diff-1acd6f7d805393f2f46612371a37e1d6511b12b40cc3a5817c2e61524e334a8cR120, right?

mwbini:
Thanks for the review. I now pushed the necessary changes

> Do you foresee different use cases requiring different values here? If yes, you could make it configurable through DeloraConfig with the default being 1e-4.

Not really, I would probably leave it like this

> Apart from that, I only found a very small issue which I commented on. Once you're finished, remember to call make style before pushing.

Thanks, I corrected it now. I should have run `make style` correctly now (I noticed I was missing one package for this). In this regard, when running`make style`, I noticed that it also made a small change to `examples/boft_controlnet/test_controlnet.py`. Given it was simply removing an unused import, I pushed with this change as well

mwbini:
Perfect, thank you for the feedbacks and reviews! I now added an example based on existing ones, and improved the method description

examples/delora_finetuning/delora_finetuning.py
```diff
@@ -0,0 +1,190 @@
...
+    # Compute type
+    device_type = device.type
+    device_module = getattr(torch, device_type, torch.cuda)
+    bf16_supported = device_module.is_available() and device_module.is_bf16_supported()
+    torch_dtype = torch.bfloat16 if bf16_supported else torch.float32
+
+    # Load the base model
+    model = AutoModelForCausalLM.from_pretrained(
+        base_model,
+        torch_dtype=torch_dtype,
```
BenjaminBossan: Let's use `dtype` instead of `torch_dtype`, since the latter is deprecated.

examples/delora_finetuning/delora_finetuning.py
```diff
@@ -0,0 +1,190 @@
...
+        # token=hf_token,
+    )
+
+    # DeLoRA config for the PEFT model
+    peft_config = DeloraConfig(
+        r=rank,
+        delora_lambda=delora_lambda,
+        target_modules=(target_modules.split(",") if target_modules else ["q_proj", "v_proj"]),
```
BenjaminBossan: I think it makes more sense to default to `None` instead of `["q_proj", "v_proj"]`, since `None` will use `TRANSFORMERS_MODELS_TO_DELORA_TARGET_MODULES_MAPPING`.

examples/delora_finetuning/README.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+To utilize the fine-tuned DeLoRA modules, simply run the following command:
+```python
+import torch
+from peft import PeftModel
+from transformers import AutoModelForCausalLM
+
+model = AutoModelForCausalLM.from_pretrained(
+    "meta-llama/Meta-Llama-3-8B", torch_dtype=torch.bfloat16, device_map="auto"
```
BenjaminBossan: Let's use `dtype` instead of the deprecated `torch_dtype`.

BenjaminBossan:
@mwbini We will probably release a new PEFT version sometime next week, if the last few issues can be fixed by then, it would ensure that DeLoRA is included ;-)

mwbini:
That would be great! I pushed the last few changes, please let me know if there are any issues left

BenjaminBossan (approved):
Thanks a lot for the last few changes and for contributing this method to PEFT. The PR is now good to go.

PR #2782, opened 2025-09-12
ENH: Store PEFT version in PEFT config file

BenjaminBossan:
This PR adds the PEFT version to the `adapter_config.json`. This can be useful in the future -- for instance when we change the state dict format of a PEFT method, we can convert it in a backwards compatible way based on the PEFT version being used. It can also be useful for debugging by providing an easy way to see the PEFT version that was used to train a PEFT adapter.

## Notes

In #2038, we made a change to PEFT configs to make it so that even if new arguments are added to a config, it can still be loaded with older PEFT versions (forward compatibility). Before that change, adding the PEFT version would have been quite disruptive, as it would make all PEFT configs incompatible with older PEFT versions. Said PR was included in the 0.14.0 release from Dec 2024, so we can expect the vast majority of PEFT users to use this version or a more recent one.

If the PEFT version is a dev version, the version tag is ambiguous. Therefore, I added some code to try to determine the commit hash. This works if users installed PEFT with `git+...@<HASH>`. Unit testing that the function to determine the hash works with these types of installs is not trivial. Therefore, I just patched the function to return a fixed hash. I did, however, test it locally and it works:

```sh
python -m pip install git+https://github.com/huggingface/diffusers.git@5e181eddfe7e44c1444a2511b0d8e21d177850a0
python -c "from peft.config import _get_commit_hash; print(_get_commit_hash('diffusers'))"
```

Also note that I tried to make the retrieval of the hash super robust by adding a broad `try ... except Exception`. If there is an error there, e.g. due to a busted install path, we never want this to fail, but rather just accept that the hash cannot be determined.

If users installed a dev version of PEFT in different way, e.g. using `git clone && pip install .`, the commit hash will not be detected. I think this is fine, I really don't want to start shelling out with git just for this purpose.

### DIFF
--- a/src/peft/config.py
+++ b/src/peft/config.py
@@ -21,6 +21,7 @@ import warnings
 from dataclasses import asdict, dataclass, field
 from typing import Optional, Union
 
+import packaging.version
 from huggingface_hub import hf_hub_download
 from transformers.utils import PushToHubMixin, http_user_agent
 
@@ -50,8 +51,7 @@ def _check_and_remove_unused_kwargs(cls, kwargs):
 
 def _is_dev_version(version: str) -> bool:
     # check if the given version is a dev version
-    _, _, patch = version.rpartition(".")
-    return patch.startswith("dev")
+    return packaging.version.Version(version).dev is not None
 
 
 def _get_commit_hash(pkg_name: str) -> str | None:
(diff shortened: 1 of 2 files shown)


src/peft/config.py
```diff
@@ -43,6 +48,31 @@ def _check_and_remove_unused_kwargs(cls, kwargs):
     return kwargs, unexpected_kwargs
 
 
+def _is_dev_version(version: str) -> bool:
+    # check if the given version is a dev version
+    _, _, patch = version.rpartition(".")
+    return patch.startswith("dev")
```
githubnemo: I'd vote for using `packaging.version.Version(version).dev is not None` here since version parsing is full of eels (or something like that). Example: `1.2.3dev0` is a valid python version but would fail to be detected here as dev version.

src/peft/config.py
```diff
@@ -60,13 +90,37 @@ class PeftConfigMixin(PushToHubMixin):
...
+
+        try:
+            git_hash = _get_commit_hash("peft")
+        except Exception:
+            # Broad exception: We never want to break user code just because the git_hash could not be determined
+            warnings.warn(
+                "A dev version of PEFT is used but there was an error while trying to determine the commit hash. "
+                "Please open an issue: https://github.com/huggingface/peft/issues"
+            )
+            git_hash = None
```
githubnemo: does it make sense to specifically mark this, e.g. `{version}@UNKNOWN`? i can imagine this behavior being unhelpful when something goes wrong in a dev install but we report a non-dev version number.

PR #2790, opened 2025-09-19
Drop Python 3.9, add 3.13

cyyever:


BenjaminBossan:
Right now, we don't test PEFT with Python 3.13:

https://github.com/huggingface/peft/blob/20a9829f76419149f5e447b856bc0abe865c28a7/.github/workflows/tests.yml#L44

Our test infrastructure is already strained, so we don't add a 5th Python version. Once Python 3.9 is EOL, which is in October, we'll remove it and add Python 3.13 to the test matrix. That said, I'm pretty sure that PEFT will work with 3.13, but I would only add it as a classifier once it's tested in CI.

cyyever:
@BenjaminBossan Could I add 3.13 and drop 3.9 in this PR so that it can be merged once 3.9 is out.

BenjaminBossan:
@cyyever We can make this the "Drop Python 3.9, add 3.13" PR. I turned it into a draft, as we only want to merge it in Oct. 

There are probably more things that could be added for this transition, e.g. removing a lot of `from __future__ import annotations` imports, but this can be done separately.

BenjaminBossan:
Okay, so we get the green tick for Ubuntu, but CI fails for MacOS and Windows. Honestly, MacOS is a lost cause at this point and I think it's fine to just [exclude it from the test matrix](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#excluding-matrix-configurations).

For Windows, the issue appears to be the low numpy version, which is installed due to this step:

https://github.com/huggingface/peft/blob/20a9829f76419149f5e447b856bc0abe865c28a7/.github/workflows/tests.yml#L82-L87

Based on some testing (#2791), it can probably be removed completely. If not, we can condition it on the Python version.

BenjaminBossan:
@cyyever Would you like to add the changes I proposed above?

cyyever:
@BenjaminBossan Do you mean removing the section in `peft/.github/workflows/tests.yml`?

BenjaminBossan:
Yes, removing it and also adding an exclusion rule for MacOS + Python 3.13.

cyyever:
@BenjaminBossan Let's try removing it.

BenjaminBossan:
Generally, this seems to work. The CI is very flaky, merging with/rebasing on the latest main will most likely fix that.

Also, let's add an [exclusion rule](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#excluding-matrix-configurations) after [this line](https://github.com/huggingface/peft/blob/190f9873b15660d9092f70065c18e4993fe10d5b/.github/workflows/tests.yml#L44) for MacOS + Python 3.13:

```yaml
        exclude:
          - os: macos-latest
            python-version: "3.13"
```

cyyever:
@BenjaminBossan Done

.github/workflows/tests.yml
```diff
@@ -40,8 +40,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
         os: ["ubuntu-latest", "macos-13", "windows-latest"]
+        exclude:
+          - os: macos-latest
```
BenjaminBossan: Ah, sorry, wrong MacOS
```suggestion
          - os: macos-13
```

PR #2792, opened 2025-09-20
Proposal: make trainable tokens more flexible to support LMHead tuning

busycalibrating:
In the case of untied Embed/LM heads, it seems natural to let the Trainable Tokens implementation also support LM Heads (especially in the case of reserved special tokens). Currently, my understanding is that it only allows you to fine tune the input embeddings (unless I'm horribly mistaken?). Is there any reason to restrict it to just input embeddings? I hacked my local PEFT install to lift this restriction, and it could be broadly useful for others as well. 

I didn't do any rigorous testing, so let me know if I'm missing anything obvious. I'd be happy to help out with this if there's interest.

### DIFF
--- a/src/peft/tuners/trainable_tokens/layer.py
+++ b/src/peft/tuners/trainable_tokens/layer.py
@@ -116,13 +116,18 @@ class TrainableTokensLayer(nn.Module, BaseTunerLayer):
         # onto the new values, we would get undefined behavior. By replacing the specific token values we always
         # get defined behavior.
         weight = self.get_base_layer().weight
-        embed_dim = self.get_base_layer().embedding_dim
+        base = self.get_base_layer()
+        embed_dim = getattr(base, "embedding_dim", None)
+        if embed_dim is None:
+            embed_dim = getattr(base, "in_features", None)
+        if embed_dim is None:
+            embed_dim = weight.shape[-1]
 
         if init_weights:
             if check_deepspeed_zero3_enabled():
                 values = self._collect_token_weights(weight, self.token_indices[adapter_name], embed_dim)
             else:
-                values = self.weight[self.token_indices[adapter_name]]
+                values = weight[self.token_indices[adapter_name]]
         else:
             # random init with matching dtype/device
             values = torch.randn(
@@ -230,9 +235,11 @@ class TrainableTokensLayer(nn.Module, BaseTunerLayer):
                 )
             elif isinstance(self.base_layer, torch.nn.Linear):
                 # Probably a tied adapter that wraps an LM head.
+                bias = getattr(self.base_layer, "bias", None)
                 result = F.linear(
                     input=x,
                     weight=W,
+                    bias=bias,
                 )
             else:
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


githubnemo:
Thanks for the suggestion!

I think that this is already supported by passing `target_modules=["embed_tokens", "lm_head"]` in `TrainableTokensConfig` (or `trainable_token_indices={"embed_tokens": [...], "lm_head": [...]}`) unless I'm missing something. Can you confirm that?

busycalibrating:
No there's a tiny issue (which is what my PR addresses) - passing the `lm_head` triggers an error when it tries to access the embedding dim; the base implementation does this:

```
embed_dim = self.get_base_layer().embedding_dim
```

Since the LMHead (at least for some models like Llama3.2) is `Linear`, this fails. I changed it to this in my local install:

```
        base = self.get_base_layer()
        embed_dim = getattr(base, "embedding_dim", None)
        if embed_dim is None:
            embed_dim = getattr(base, "in_features", None)
        if embed_dim is None:
            embed_dim = weight.shape[-1]
`````

File: src/peft/tuners/trainable_tokens/layer.py [hunk 1]
```diff
@@ -230,9 +235,11 @@ def forward_adapters(self, x: torch.Tensor, active_adapters, *args, **kwargs) ->
                 )
             elif isinstance(self.base_layer, torch.nn.Linear):
                 # Probably a tied adapter that wraps an LM head.
+                bias = getattr(self.base_layer, "bias", None)
                 result = F.linear(
                     input=x,
                     weight=W,
+                    bias=bias,
```
githubnemo, line 242: I think we should use `self.get_base_layer()` instead to be consistent with `update_layer()`. If you want, you can update the lines above for the `F.embedding` call and instance check as well.

src/peft/tuners/trainable_tokens/layer.py
```diff
@@ -116,13 +116,18 @@ def update_layer(self, adapter_name, **kwargs):
         # onto the new values, we would get undefined behavior. By replacing the specific token values we always
         # get defined behavior.
         weight = self.get_base_layer().weight
-        embed_dim = self.get_base_layer().embedding_dim
+        base = self.get_base_layer()
+        embed_dim = getattr(base, "embedding_dim", None)
+        if embed_dim is None:
+            embed_dim = getattr(base, "in_features", None)
+        if embed_dim is None:
+            embed_dim = weight.shape[-1]
```
githubnemo, line 124: Can you elaborate the purpose of the last case?

src/peft/tuners/trainable_tokens/model.py
```diff
@@ -41,7 +41,27 @@ def __getattr__(self, name: str):
...
+
+            if not are_tied and hasattr(self.model, "get_output_embeddings"):
+                out_emb = self.model.get_output_embeddings()
+                if out_emb is not None:
+                    for name, module in self.model.named_modules():
+                        if module is out_emb:
+                            targets.append(name)
+                            break
+
+            peft_config.target_modules = list(dict.fromkeys(targets))
```
githubnemo, line 64: What's the idea behind targeting the output embedding automatically in case of untied weights? I don't see the benefit and I think this is also breaking backward compatibility with existing checkpoints.

busycalibrating:
Thanks for the feedback, I'll follow up by the end of the week/next week!

githubnemo:
*not stale*

githubnemo:
I think that there have been quite some changes to the way trainable tokens handles tied/untied LM heads.

Can you confirm that your proposed changes are integrated in #2863 and #2870?

busycalibrating:
Yes this looks good - thanks for following up on this!

githubnemo:
Thanks for confirming and taking the initiative! I'll close this PR then.

Issue #2786, opened 2025-09-17
Two issues with [X-LoRA] Implementation: Scalings Logging and Top-K Softmax Normalization


Che-Xu:
Hello PEFT team,
I am writing to report two issues I encountered while using X-LoRA with the Qwen2-VL-7B model.

**Issue 1: `internal_xlora_scalings` Not Properly Initialized `in _enable_peft_forward_hooks()`**

**Location:** `_enable_peft_forward_hooks()` in `src/peft/tuners/xlora/model.py`

**Problem:** After calling `enable_scalings_logging()`, the subsequent call to `get_latest_scalings()` returns `None` for `self.internal_xlora_scalings`.

**Suggested Fix:**

I suspect the issue is that in the `_enable_peft_forward_hooks()` function, after the first forward pass computes the xlora_scalings via:
```
xlora_scalings = self.internal_xlora_classifier(result=base_output, *args_real, **kwargs_real)
```
the value of xlora_scalings is not assigned to `self.internal_xlora_scalings`. I suggest adding:
```
self.internal_xlora_scalings = xlora_scalings
``` 
immediately after the above line.

**Issue 2: Incorrect Probability Normalization in Token-Level Top-K Selection**
**Location:** `get_maybe_topk_scalings()` in `src/peft/tuners/xlora/layer.py`
**Problem:** When using `enable_softmax=False, enable_softmax_topk=True, and top_k_lora=1`, the current implementation normalizes the expert probabilities such that the sum over all tokens is 1, rather than summing to 1 per token. Here is the current code:
```
if self.config.enable_softmax_topk:
    nonzero_mask = xlora_scalings != 0
    softmax_res_nonzero = torch.softmax(xlora_scalings[nonzero_mask], dim=-1)
    xlora_scalings[nonzero_mask] = softmax_res_nonzero
```
**Suggested Fix:**
I believe the intended behavior is that each token's top-k expert probabilities should sum to 1. A more appropriate implementation may be:
```
if self.config.enable_softmax_topk:
    full_scalings = torch.full_like(xlora_scalings, float('-inf'))
    full_scalings.scatter_(-1, topk_indices, scalings.gather(-1, topk_indices))
    xlora_scalings = torch.softmax(full_scalings, dim=-1)
```
After applying this change in my experiments, the behavior matched expectations.

Could you please confirm whether these are indeed issues and whether the proposed changes are appropriate? Thank you for your time and effort.

BenjaminBossan:
[...]

PR #2793, opened 2025-09-21
Fix issue #2786: Store xlora scaling and fix per token normalization

Che-Xu:
Resolves [#2786](https://github.com/huggingface/peft/issues/2786)

## Description

This PR addresses two issues identified while using X-LoRA with the Qwen2-VL-7B model.

### Issue 1: Internal Scaling Storage Problem

**Location:** `_enable_peft_forward_hooks()` in `src/peft/tuners/xlora/model.py`
**Problem:** After calling `enable_scalings_logging()`, the subsequent call to `get_latest_scalings()` returned `None` because computed `xlora_scalings` were not properly stored for later retrieval.

### Issue 2: Incorrect Probability Normalization
**Location:** `get_maybe_topk_scalings()` in `src/peft/tuners/xlora/layer.py`
**Problem:** The current implementation incorrectly normalized expert probabilities such that the sum over all tokens was 1, rather than summing to 1 per token.

## Implementation

- src/peft/tuners/xlora/model.py: Added storage of computed scalings in `_enable_peft_forward_hooks()`
```
xlora_scalings = self.internal_xlora_classifier(result=base_output, *args_real, **kwargs_real)
# Store computed scalings to fix get_latest_scalings() returning None
self.internal_xlora_scalings = xlora_scalings
```
- src/peft/tuners/xlora/layer.py: Fixed normalization logic in `get_maybe_topk_scalings()`
```
# Apply per-token normalization to the xLoRA scaling factors using a softmax
if self.config.enable_softmax_topk:
    nonzero_mask = xlora_scalings != 0
    full = xlora_scalings.masked_fill(~nonzero_mask, float("-inf"))
    new_scalings = torch.softmax(full, dim=-1)
    xlora_scalings = new_scalings.masked_fill(~nonzero_mask, 0.0)
```

### DIFF
--- a/src/peft/tuners/xlora/layer.py
+++ b/src/peft/tuners/xlora/layer.py
@@ -73,11 +73,6 @@ class XLoraLayer:
 
             xlora_scalings = xlora_scalings * mask.to(xlora_scalings.dtype)
 
-        if self.config.enable_softmax_topk:
-            nonzero_mask = xlora_scalings != 0
-            softmax_res_nonzero = torch.softmax(xlora_scalings[nonzero_mask], dim=-1)
-            xlora_scalings[nonzero_mask] = softmax_res_nonzero
-
         # Apply per-token normalization to the xLoRA scaling factors using a softmax
         if self.config.enable_softmax_topk:
             nonzero_mask = xlora_scalings != 0
--- a/src/peft/tuners/xlora/layer.py
+++ b/src/peft/tuners/xlora/layer.py
@@ -78,6 +78,13 @@ class XLoraLayer:
             softmax_res_nonzero = torch.softmax(xlora_scalings[nonzero_mask], dim=-1)
             xlora_scalings[nonzero_mask] = softmax_res_nonzero
 
+        # Apply per-token normalization to the xLoRA scaling factors using a softmax
+        if self.config.enable_softmax_topk:
+            nonzero_mask = xlora_scalings != 0
+            full = xlora_scalings.masked_fill(~nonzero_mask, float("-inf"))
+            new_scalings = torch.softmax(full, dim=-1)
+            xlora_scalings = new_scalings.masked_fill(~nonzero_mask, 0.0)
+
         return xlora_scalings
 
 
--- a/src/peft/tuners/xlora/model.py
+++ b/src/peft/tuners/xlora/model.py
@@ -368,6 +368,8 @@ class XLoraModel(BaseTuner):
                     self.lora_model.enable_adapter_layers()
 
             xlora_scalings = self.internal_xlora_classifier(result=base_output, *args_real, **kwargs_real)
+            # Store computed scalings to fix get_latest_scalings() returning None
+            self.internal_xlora_scalings = xlora_scalings
 
             # =========================== Real forward pass with calculated scalings ==================
 


BenjaminBossan:
Thanks a lot for creating this PR to fix the issues you identified. At a glance, the changes look good. Ideally, we should also have unit tests to check for these bugs. I think it shouldn't be too hard to add them by extending [`test_xlora.py`](https://github.com/huggingface/peft/blob/main/tests/test_xlora.py). Would you be interested in adding these tests @Che-Xu?

PS: The failing CI is unrelated and can be ignored.

BenjaminBossan:
@Che-Xu Do you still plan to work on this?

Che-Xu:
Hi @BenjaminBossan,

Thank you for the reminder and my sincere apologies for the delayed response. I've been occupied with other projects over the past two weeks.

I'm still committed to this and will submit the unit tests within the next two days. I understand the importance of completing this and appreciate you checking in.

Again, sorry for the delay and thank you for your patience. I'll prioritize this task and keep you updated.

BenjaminBossan:
@Che-Xu No worries, take the time you need. My ping was just a reminder, as sometimes people forget or miss notifications. Feel free to let me know if you have any questions.

Che-Xu:
@BenjaminBossan 

Thanks for the feedback! I've extended `test_xlora.py` with two new unit tests:

1. **`test_scalings_storage()`**
- Verifies that scaling values are properly stored after generation
- Validates that `get_latest_scalings()` returns non-None torch.Tensors  
- Ensures all scaling values are finite and numerically stable

2. **`test_per_token_normalization_with_softmax_topk()`**
- Implements a testing approach using monkey-patching to hook into the forward pass of `XLoraLinear` layers
- Captures scaling values during the second forward pass and validates the `get_maybe_topk_scalings()` function behavior
- Validates both the existence and mathematical correctness of normalized scalings
- Includes shape validation, batch processing checks, and numerical stability assertions

The tests have been added to the existing PR. Please let me know if you'd like me to adjust anything in the test coverage!

tests/test_xlora.py
```diff
@@ -381,3 +382,126 @@ def test_xlora_loading_valid(self):
...
+                max_new_tokens=1,
+            )
+
+            assert len(captured_data) > 0, "No scaling data was captured during forward pass."
+
+            for data in captured_data:
+                normalized_scalings = data.get("normalized_scalings")
+                if normalized_scalings is None:
+                    assert normalized_scalings is not None, (
+                        f"Missing normalized_scalings in layer {data['layer']} {data['projection']}"
```
BenjaminBossan: The `if` check can be removed, right?

tests/test_xlora.py
```diff
@@ -381,3 +382,126 @@ def test_xlora_loading_valid(self):
...
+                    )
+                    continue
+
+                if hasattr(normalized_scalings, "cpu"):
+                    scalings_np = normalized_scalings.cpu().detach().numpy()
```
BenjaminBossan: Why is it necessary to move the array to numpy?

tests/test_xlora.py
```diff
@@ -381,3 +382,126 @@ def test_xlora_loading_valid(self):
...
+                assert scalings_np.ndim == 3, (
+                    f"Unexpected scalings shape {scalings_np.shape} in layer {data['layer']} {data['projection']}"
+                )
+
+                batch_size, seq_len, num_experts = scalings_np.shape
+                for b in range(batch_size):
+                    for t in range(seq_len):
+                        weights = scalings_np[b, t, :]
+                        weight_sum = weights.sum()
+                        assert np.isclose(weight_sum, 1.0, atol=1e-5), (
```
BenjaminBossan: I think this is the essential part of the test. I would focus on this assert, no need to check the other stuff and also no need to report the layer, batch, and token in detail.

tests/test_xlora.py
```diff
@@ -381,3 +382,126 @@ def test_xlora_loading_valid(self):
...
+            max_new_tokens=10,
+        )
+
+        latest_scalings = model.get_latest_scalings()
+        assert latest_scalings is not None, "get_latest_scalings() should not return None after generation"
+        assert isinstance(latest_scalings, torch.Tensor)
+        assert torch.isfinite(latest_scalings).all(), "Scalings should contain finite values"
+
+    def test_per_token_normalization_with_softmax_topk(self, tokenizer, model):
+        captured_data = []
```
BenjaminBossan: I think the whole test can be greatly simplified if we are content with only logging the scalings. I think that should be enough, as the other logged data is just needed for a nicer error message and I believe we can do without that.

Here is my proposal:

```python
from peft.tuners.xlora.layer import XLoraLayer

...
    def test_per_token_normalization_with_softmax_topk(self, tokenizer, model, monkeypatch):
        orig_get_maybe_topk_scalings = XLoraLayer.get_maybe_topk_scalings

        captured_data = []
        def mock_get_maybe_topk_scalings(*args, **kwargs):
            result = orig_get_maybe_topk_scalings(*args, **kwargs)
            captured_data.append(result)
            return result

        monkeypatch.setattr(XLoraLayer, "get_maybe_topk_scalings", mock_get_maybe_topk_scalings)

        model.enable_scalings_logging()
        inputs = tokenizer.encode("Test per token normalization", add_special_tokens=False, return_tensors="pt")
        outputs = model.generate(
            input_ids=inputs.to(self.torch_device),
            max_new_tokens=1,
        )
        for scaling in captured_data:
            assert ...
```

Che-Xu:
@BenjaminBossan,

Thank you for your suggestion! The revised version is much cleaner and clearer, and I have learned a lot from your approach.

One small addition I made is that, since XLoRA performs two forward passes (a dummy pass and a real pass), and we only want to capture the scalings from the real pass, I included the check `if getattr(model, "internal_xlora_scalings", None) is not None:` in the mock function. This ensures we only record the normalized scalings that are actually used in the real forward pass.

I have already made these changes in the existing PR. Thank you again for your guidance!

PR #2797, opened 2025-09-22
TST: fix `to` issue for 8-bit model

yao-matrix:
## Issue
4 cases fail 
```
pytest -rA tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit
pytest -rA tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_dora
pytest -rA tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_iter_5
pytest -rA tests/test_gpu_examples.py::TestLoftQ::test_t5_loftq_8bit
```
w/ log

> >               raise ValueError(
>                     "`.to` is not supported for `8-bit` bitsandbytes models. Please use the model as it is, since the"
>                     " model has already been set to the correct devices and casted to the correct `dtype`."
> E                   ValueError: `.to` is not supported for `8-bit` bitsandbytes models. Please use the model as it is, since the model has already been set to the correct devices and casted to the correct `dtype`.

## Root Cause
8-bit models don't support `to` and `cuda` according [transformers logic](https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L4353)

## Fix
check whether model is loaded in 8 bit, and skip `to` if so.

## Results:
Pass

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -2838,7 +2838,8 @@ class TestLoftQ:
     def get_base_model(self, model_id, device, **kwargs):
         cls = AutoModelForSeq2SeqLM if "t5" in str(model_id) else AutoModelForCausalLM
         model = cls.from_pretrained(model_id, **kwargs).eval()
-        model = model.to(self.device)
+        if not getattr(model, "is_loaded_in_8bit", False):
+            model = model.to(self.device)
         return model
 
     def get_logits(self, model, inputs):


yao-matrix:
@BenjaminBossan , could you pls take a review? Thx very much.

githubnemo:
Thanks for the suggestion! I wonder if we should skip the test instead since we're "requesting" to use `self.device` but that model doesn't support using that device. WDYT?

src/peft/utils/integrations.py
```diff
@@ -101,6 +104,7 @@ def dequantize_bnb_weight(weight: torch.nn.Parameter, state=None):
             weight = weight.to(torch.device("cuda"))
         elif is_xpu_available():
             weight = weight.to(torch.device("xpu"))
+            state.SCB = state.SCB.to(torch.device("xpu"))
```
yao-matrix, line 108: `weight` is moved to accelerator, but `SCB` not, will make later `int8_vectorwise_dequant` cry to say `weight in xpu, but SCB in cpu` and make CPU CI, fail, so move it to accelerator too.

src/peft/utils/integrations.py
```diff
@@ -121,6 +122,7 @@ def dequantize_bnb_weight(weight: torch.nn.Parameter, state=None):
 
     if is_cpu:
         dequantized = dequantized.to(device)
+        state.SCB = state.SCB.to(device)
```
yao-matrix, line 126: both should be move back

yao-matrix:
> Thanks for the suggestion! I wonder if we should skip the test instead since we're "requesting" to use `self.device` but that model doesn't support using that device. WDYT?

@githubnemo, I took some time to dive into the issue and then cleaned the test cases, then the device cases passed, we have `torch_device`, `device` of class `TestLoftQ`, `device` through method arguments, kind of make the test case confused, so I cleaned it, then accelerator(xpu, cuda) case can pass.

for `cpu` failures, its because incomplete tensor transfer btw cpu and accelerator, i commented inline.

Thx very much for review and sorry for before's lazy workaround.

yao-matrix:
@githubnemo , could you pls take a review again, thx very much

tests/test_gpu_examples.py
```diff
@@ -2823,22 +2823,22 @@ class TestLoftQ:
...
-    # The error factor indicates by how much the quantization error should be decreased when using LoftQ compared to
-    # quantization without LoftQ. Thus 1.03 means that the error should be decreased by 3% at least. This is a very
-    # conservative value to prevent flakiness, in practice most gains are > 1.5
-    device = infer_device()
-    error_factor = 1.005 if device in ("xpu", "cpu") else 1.03
+    def get_error_factor(self, device):
+        # The error factor indicates by how much the quantization error should be decreased when using LoftQ compared to
+        # quantization without LoftQ. Thus 1.03 means that the error should be decreased by 3% at least. This is a very
+        # conservative value to prevent flakiness, in practice most gains are > 1.5
+        error_factor = 1.005 if device in ("xpu", "cpu") else 1.03
```
BenjaminBossan, line 2831: Did you find it necessary to reduce the factor to 1.005 for XPU and CPU?

BenjaminBossan (commented):
Thanks for working on this with 8bit bnb and device placement. Unfortunately, when I run this locally, I still get errors:

```
$ CUDA_VISIBLE_DEVICES=0 pytest tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_dora tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_iter_5 tests/test_gpu_examples.py::TestLoftQ::test_t5_loftq_8bit

FAILED tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit[cpu] - assert tensor(nan, grad_fn=<MeanBackward0>) > 0.0
FAILED tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_dora[cpu] - RuntimeError: invalid argument to getCurrentStream
FAILED tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_iter_5[cuda] - torch.AcceleratorError: CUDA error: an illegal memory access was encountered
FAILED tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_iter_5[cpu] - torch.AcceleratorError: CUDA error: an illegal memory access was encountered
FAILED tests/test_gpu_examples.py::TestLoftQ::test_t5_loftq_8bit[cuda] - torch.AcceleratorError: CUDA error: an illegal memory access was encountered
FAILED tests/test_gpu_examples.py::TestLoftQ::test_t5_loftq_8bit[cpu] - torch.AcceleratorError: CUDA error: an illegal memory access was encountered
```

I cannot test on XPU, but for CPU and CUDA, I get these errors. Do these tests pass for you?

Furthermore, I was wondering why these tests are not running on CI and I think it's a simple oversight, there is no test marker for `TestLoftQ`. Could you please add the `@pytest.mark.single_gpu_tests` decorator?

yao-matrix:
> Thanks for working on this with 8bit bnb and device placement. Unfortunately, when I run this locally, I still get errors:
> 
> ```
> $ CUDA_VISIBLE_DEVICES=0 pytest tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_dora tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_iter_5 tests/test_gpu_examples.py::TestLoftQ::test_t5_loftq_8bit
> 
> FAILED tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit[cpu] - assert tensor(nan, grad_fn=<MeanBackward0>) > 0.0
> FAILED tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_dora[cpu] - RuntimeError: invalid argument to getCurrentStream
> FAILED tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_iter_5[cuda] - torch.AcceleratorError: CUDA error: an illegal memory access was encountered
> FAILED tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_iter_5[cpu] - torch.AcceleratorError: CUDA error: an illegal memory access was encountered
> FAILED tests/test_gpu_examples.py::TestLoftQ::test_t5_loftq_8bit[cuda] - torch.AcceleratorError: CUDA error: an illegal memory access was encountered
> FAILED tests/test_gpu_examples.py::TestLoftQ::test_t5_loftq_8bit[cpu] - torch.AcceleratorError: CUDA error: an illegal memory access was encountered
> ```
> 
> I cannot test on XPU, but for CPU and CUDA, I get these errors. Do these tests pass for you?
> 
> Furthermore, I was wondering why these tests are not running on CI and I think it's a simple oversight, there is no test marker for `TestLoftQ`. Could you please add the `@pytest.mark.single_gpu_tests` decorator?

@BenjaminBossan Could you try installing latest bnb with `pip install --force-reinstall https://github.com/bitsandbytes-foundation/bitsandbytes/releases/download/continuous-release_main/bitsandbytes-1.33.7.preview-py3-none-manylinux_2_24_x86_64.whl`, since there is a bug in bnb and the fixing PR just merged https://github.com/bitsandbytes-foundation/bitsandbytes/pull/1769. In my env, with this bnb fix, the test cases pass on XPU, CPU and CUDA(A100). Thx

tests/test_gpu_examples.py
yao-matrix: currently, it's needed. We are continuing enhance bnb support, when it works, we can remove it.

BenjaminBossan: I see. I checked again with CUDA/CPU and the improvement is much larger than 1.03 (which, as indicated, is a very conservative value to avoid flakiness). So if XPU is below that, I think it's an indicator that something is missing.

PR #2800, opened 2025-09-24
Use technical user for CI runs

githubnemo:
Makes it easier to track rate limiting issues.


BenjaminBossan (approved):
Fantastic. Merge it!!

Issue #2777, opened 2025-09-08
Update the model config to tie_word_embeddings = False in case of lm_head or embedding layer update


romitjain:
### Feature request

I wanted to describe an unexpected (or unusual) behaviour:

If we pass either of the embedding layers or the LM head in modules to save for a model whose embeddings are tied (tie_word_embeddings = True), peft does the following:

1. It adds adaptors to both layers (in case both layers are added, then separately), which essentially results in an untied layer (divergence from the original model architecture)
2. Trains them separately, and the resultant model has different weights for both the embedding layer and the LM head. While storing the state dict, it stores different weight parameters
3. But most critically, it does not update the model config to set tie_word_embeddings to be False. It leads to unexpected downstream behaviour in case the model is loaded and merged. 
  3.1. If we load the model again via PeftModel.from_pretrained, it works as expected 
  3.2. If we merge it and load again using AutoModel.from_pretrained, it ties the embedding layer and the LM head. This results in nonsensical generations.

There have been several reported issues and solutions in the past:

1. https://github.com/huggingface/peft/issues/2018
2. https://github.com/huggingface/peft/issues/2244
3. https://github.com/huggingface/peft/issues/1750

One of the solves (that I am aware of) 
1. https://github.com/huggingface/peft/pull/2025 - But this just adds the warning

### Your contribution

There are 2 possible solutions I see (applicable only for models with tie_word_embeddings = True

1. If either of embedding layer or the lm_head is added as a target module, mark them both as target modules to respect the original model's config
2. Update the model config and mark tie_word_embeddings = False. This will solve the issue but all LoRA finetunes will diverge in the model config from the base model

If we can agree on the right solution, I can raise a PR to solve this.

BenjaminBossan:
Thanks for reporting this issue and giving a detailed description and thinking about possible solutions. I agree that how this is handled can be messy in PEFT.

First, let's establish a common basis:

> If we pass either of the embedding layers or the LM head in modules to save for a model whose embeddings are tied (tie_word_embeddings = True), peft does the following:
> 
> 1. It adds adaptors to both layers (in case both layers are added, then separately), which essentially results in an untied layer (divergence from the original model architecture)

Hmm, is this really what happens?

```python
from transformers import AutoModelForCausalLM
from peft import LoraConfig, get_peft_model

model_id = "google/gemma-2-2b-it"
model = AutoModelForCausalLM.from_pretrained(model_id)
assert model.config.tie_word_embeddings

[...]

PR #2803, opened 2025-09-29
Ensure weight tying is maintained for embed_tokens and lm_head

romitjain:
This PR solves: #2777 

Essentially, the PR:
1. Adds an argument in `LoraConfig` for users to maintain weight tying in the model
2. Checks if the passed model has weight tying enabled, gets the layers that are tied tothe  embedding layer
3. Add a `ModulesToSaveWrapper` on modules returned in (2) and a reference to the original layer

This ensures that the weight tying is enabled if the user passes either `embed_tokens` or `lm_head`.

Limitations
1. Only added for `LoraConfig`
2. Only added for `modules_to_save` parameter. This can be extended to `target_modules` as well once the approach is approved

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -1194,13 +1194,6 @@ class BaseTuner(nn.Module, ABC):
         )
         warnings.warn(msg)
 
-    def _maybe_warn_ensure_weight_tying(self, is_embedding_to_save, tied_weight_keys):
-        """
-        This method adds a warning if `ensure_weight_tying` is not enabled. Each tuner class extending `BaseTuner` and
-        implementing `ensure_weight_tying` in PeftConfig should implement this.
-        """
-        pass
-
     def _check_tied_modules(self, model: nn.Module, peft_config):
         """
         Checks if any of the tied layers are targetted via `modules_to_save`. Updates the `peft_config.modules_to_tie`
@@ -1213,7 +1206,7 @@ class BaseTuner(nn.Module, ABC):
 
         if getattr(peft_config, "ensure_weight_tying", False):
             if is_embedding_to_save and tied_weight_keys:
-                self._maybe_add_modules_to_tie(peft_config, tied_weight_keys)
+                self._add_modules_to_tie(peft_config, tied_weight_keys)
 
             elif not is_embedding_to_save and tied_weight_keys:
                 warnings.warn(
@@ -1223,8 +1216,25 @@ class BaseTuner(nn.Module, ABC):
             elif not tied_weight_keys:
                 warnings.warn("You have requested `ensure_weight_tying`, but no tied modules were found in the model")
 
-        else:
-            self._maybe_warn_ensure_weight_tying(is_embedding_to_save, tied_weight_keys)
+        elif is_embedding_to_save and tied_weight_keys:
+            if hasattr(peft_config, "ensure_weight_tying"):
+                msg = (
+                    "Model has `tie_word_embeddings=True` and a tied layer is part of the adapter, "
+                    "but `ensure_weight_tying` is not set to True. "
+                    "This can lead to complications, for example when merging the adapter "
+                    "or converting your model to formats other than safetensors. "
+                    "Check the discussion here: https://github.com/huggingface/peft/issues/2777"
+                )
+                warnings.warn(msg)
+            else:
+                msg = (
+                    "Model has `tie_word_embeddings=True` and a tied layer is part of the adapter, "
+                    "but no implementation exists to tie the adapters. "
+                    "This can lead to complications, for example when merging the adapter "
+                    "or converting your model to formats other than safetensors. "
+                    "Check the discussion here: https://github.com/huggingface/peft/issues/2777"
+                )
+                warnings.warn(msg)
 
     def __getattr__(self, name: str):
         """Forward missing attributes to the wrapped module."""
--- a/tests/test_initialization.py
+++ b/tests/test_initialization.py
@@ -111,39 +111,6 @@ class TestLoraInitialization:
 
         return MyModule().eval().to(self.torch_device)
 
-    def get_lm_model(self, bias=True, tie_weights=True):
-        # Mimicking a LM with embed_tokens and lm_head layers
-        # to test weight tying of adapters
-        class MyModule(nn.Module):
-            def __init__(self):
-                super().__init__()
-
-                self.embed_tokens = nn.Embedding(1000, 1000)
-                self.linear = nn.Linear(1000, 1000, bias=bias)
-
-        class CausalLM(nn.Module):
-            if tie_weights:
-                _tied_weights_keys = ["lm_head.weight"]
-
-            def __init__(self):
-                super().__init__()
-                self.model = MyModule()
-                self.config = {"tie_word_embeddings": tie_weights}
-
-                if tie_weights:
-                    self.lm_head = nn.Linear(1000, 1000, bias=False)
-                    self.lm_head.weight = self.model.embed_tokens.weight
-                else:
-                    self.lm_head = nn.Linear(1000, 1000, bias=bias)
-
-            def prepare_inputs_for_generation(self):
-                return
-
-            def get_input_embeddings(self):
-                return self.model.embed_tokens
-
-        return CausalLM().eval().to(self.torch_device)
-
     @pytest.fixture
     def data(self):
         return torch.rand(10, 1000).to(self.torch_device)
@@ -1599,89 +1566,6 @@ class TestLoraInitialization:
         # it's okay to add a config with bias="none" (the default)
         config2 = LoraConfig(target_modules=["linear"], bias="none")
         model.add_adapter("other", config2)  # does not raise
-
-    def test_weight_tying_tied_model(self):
-        # If weight tying is enabled and `embed_tokens`
-        # is passed as a `modules_to_save`, it needs to be ensured
-        # that lm_head is tied to the adapter added to `embed_tokens`
-
-        model = self.get_lm_model()
-        embed_token_config = LoraConfig(
-            modules_to_save=["embed_tokens"],
-            target_modules=["linear"],
-            ensure_weight_tying=True,
-        )
-        model = get_peft_model(model, embed_token_config)
-
-        assert isinstance(model.base_model.model.model.embed_tokens, ModulesToSaveWrapper), (
-            "Embed tokens is not added in Modules to Save"
-        )
-        assert type(model.base_model.model.model.embed_tokens) is type(model.base_model.model.lm_head), (
-            "Embed tokens and LM head types are not same"
-        )
-
-        # Validating that all model parameters are same
-        embed_np = dict(model.base_model.model.model.embed_tokens.named_parameters())
-        lm_head_np = dict(model.base_model.model.lm_head.named_parameters())
-
-        for k in embed_np.keys():
-            assert torch.allclose(embed_np[k], lm_head_np[k])
-            assert embed_np[k] is lm_head_np[k]
-
-    def test_weight_tying_non_tied_model(self):
-        model = self.get_lm_model(tie_weights=False)
-        embed_token_config = LoraConfig(
-            modules_to_save=["embed_tokens"],
-            target_modules=["linear"],
-            ensure_weight_tying=True,
-        )
-        with pytest.warns(UserWarning, match="no tied modules were found in the model"):
-            model = get_peft_model(model, embed_token_config)
-
-        assert isinstance(model.base_model.model.model.embed_tokens, ModulesToSaveWrapper), (
-            "Embed tokens is not added in Modules to Save"
-        )
-        assert isinstance(model.base_model.model.lm_head, torch.nn.modules.linear.Linear), (
-            "LM head is not of type nn.linear"
-        )
-
-    def test_not_weight_tying_tied_model(self):
-        model = self.get_lm_model()
-        embed_token_config = LoraConfig(
-            modules_to_save=["embed_tokens"],
-            target_modules=["linear"],
-            ensure_weight_tying=False,
-        )
-        with pytest.warns(UserWarning, match="`ensure_weight_tying` is not set to True"):
-            model = get_peft_model(model, embed_token_config)
-
-        assert isinstance(model.base_model.model.model.embed_tokens, ModulesToSaveWrapper), (
-            "Embed tokens is not added in Modules to Save"
-        )
-        assert isinstance(model.base_model.model.lm_head, torch.nn.modules.linear.Linear), (
-            "LM head is not of type nn.linear"
-        )
-
-    def test_weight_tying_tied_model_no_embed(self):
-        model = self.get_lm_model()
-        embed_token_config = LoraConfig(
-            target_modules=["linear"],
-            ensure_weight_tying=True,
-        )
-
-        with pytest.warns(UserWarning, match="no tied modules are added in `modules_to_save`"):
-            model = get_peft_model(model, embed_token_config)
-
-        assert isinstance(model.base_model.model.model.embed_tokens, torch.nn.modules.Embedding)
-        assert isinstance(model.base_model.model.lm_head, torch.nn.modules.linear.Linear)
-
-        # Validating that all model parameters are same
-        embed_np = dict(model.base_model.model.model.embed_tokens.named_parameters())
-        lm_head_np = dict(model.base_model.model.lm_head.named_parameters())
-
-        for k in embed_np.keys():
-            assert torch.allclose(embed_np[k], lm_head_np[k])
-            assert embed_np[k] is lm_head_np[k]
 
 
 class TestLokrInitialization:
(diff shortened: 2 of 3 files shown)


romitjain:
@BenjaminBossan Request your review on the approach. If this looks fine, I can handle the other case (target_modules) and add tests.

Thanks

.gitignore
```diff
@@ -143,3 +143,6 @@ wandb
 # method_comparison logs
 method_comparison/MetaMathQA/cancelled_results/
 method_comparison/MetaMathQA/temporary_results/
+
+.peft/
+test.py
```
BenjaminBossan: Let's remove these. You can set up a [global .gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c) for these cases.

BenjaminBossan (commented):
Thanks for working on the PR to deal with targeting tied weights.

I have a suspicion that parts of this PR can be implemented more easily by extending the `peft_config.modules_to_save` to include the tied module instead of calling `_set_trainable` explicitly.

However, before checking that, it makes most sense to add one or a few unit tests to establish if everything works as expected. I think the best place for those would be in [this test class](https://github.com/huggingface/peft/blob/4469af57a08b0a751b7b1359d431c03a22b33116/tests/test_initialization.py#L76). LMK if you need help with implementing the tests.

Once we have the tests and confirm that everything works as expected, we can check if my proposal above works.

romitjain:
@BenjaminBossan I have addressed your comments. You mentioned that I can extend `peft_config.modules_to_save` to enable the functionality. Can you elaborate on that?

Essentially, we would need to mark the tied modules as modules to save and then update the weight pointer to point to the source adapter.

PTAL.
Thanks

src/peft/peft_model.py
```diff
@@ -1845,6 +1846,31 @@ def __init__(
         super().__init__(model, peft_config, adapter_name, **kwargs)
         self.base_model_prepare_inputs_for_generation = self.base_model.prepare_inputs_for_generation
 
+        # Condition to check if embedding layer (`embed_tokens`) is added
+        # in `modules_to_save` and we want to ensure the `lm_head`
+        # does not diverge from the `embed_tokens` layer
+        if (
+            peft_config.task_type == "CAUSAL_LM"
```
BenjaminBossan: Why is it important?

tests/test_initialization.py
```diff
@@ -1566,6 +1607,75 @@ def test_multiple_configs_with_bias_raises(self, tmp_path):
...
+        from peft.utils.other import ModulesToSaveWrapper
+
+        model = self.get_lm_model(tie_weights=False)
+        embed_token_config = LoraConfig(
+            task_type="CAUSAL_LM",
+            modules_to_save=["embed_tokens"],
+            target_modules=["linear"],
+            ensure_weight_tieing=True,
+        )
+        model = get_peft_model(model, embed_token_config)
```
BenjaminBossan: I think it would make sense to have a warning here, since it would seem that the user tried to tie the weights but there was nothing to tie, WDYT?

.gitignore
```diff
@@ -142,4 +142,4 @@ wandb
 
 # method_comparison logs
 method_comparison/MetaMathQA/cancelled_results/
-method_comparison/MetaMathQA/temporary_results/
+method_comparison/MetaMathQA/temporary_results/
```
BenjaminBossan: Please add newline :)

src/peft/peft_model.py
romitjain: Updating `lm_head` makes sense only for `CAUSAL_LM` tasks. We can extend it to Seq2Seq.
I might be wrong here though.

tests/test_initialization.py
romitjain: Yes, I have added a warning here:

https://github.com/huggingface/peft/pull/2803/files#diff-95787fcc5356568929e383449ade2c1bac5da43deccab17318652b62ed171ae7R1860

romitjain:
@BenjaminBossan I think what you suggested is a cleaner approach. I am aligned with that.
~The only concern I have is that we would first register the redundant weights (extra memory) for the tied modules and then replace the pointer post hoc.
However, the extra memory might not be too big, so it should be okay? What do you think?~

I realized that the extra weights are being created in both situations, I will go ahead with the solution you proposed.

src/peft/peft_model.py
BenjaminBossan: The proposed code is not really requiring there to be an `lm_head`, right? In theory, it should also work when there are other tied weights. I'm not sure how relevant that is in practice, but if there are no strict reasons for this check, I'd say let's remove it.

tests/test_initialization.py
BenjaminBossan: Let's check the warning here then, e.g. with `recwarn` or `with pytest.warns`.

src/peft/peft_model.py
romitjain: Yes, I'll make the changes such that we only worry about weight tying in case one of the tied layers is added in the modules to save

romitjain:
@BenjaminBossan 

I have made slight tweaks to your proposed approach. Instead of modifying `_set_trainable`, which has to check for tied modules and a bunch of other checks, I am doing that once in `BaseTuners.inject_adapter` and adding a new runtime field in LoraConfig for `modules_to_tie` that can be used later on to tie the modules.

On tests, I have covered the following cases

```
- tied weights
    - `embed_tokens` not added as modules_to_save
        - `ensure_weight_tying` == True : Warning test
        - `ensure_weight_tying` == False : No need to test
    - `embed_tokens` added as modules_to_save
        - `ensure_weight_tying` == True : Test
        - `ensure_weight_tying` == False : Warning test
- non tied weights
    - `embed_tokens` added/not added as modules_to_save
        - `ensure_weight_tying` == True : Warning test
        - `ensure_weight_tying` == False : No need to test
```

BenjaminBossan:
Thanks for your continued work on this PR.

The idea I proposed was to create the `ModulesToSaveWrapper`s as normal in a first pass and then apply the weight tying in a second pass. If I understand your proposal correctly, you would apply the weight tying on the fly (while iterating through the modules). The reason why I wanted to avoid that is because it requires a specific order, right? If the module that should be tied is wrapped _before_ the module that it's being tied to, your proposed approach would fail. Is my understanding correct here?

I guess one could argue that this would be a user error and the user needs to take care to define `modules_to_save` correctly. On the other hand, the very idea of this feature is to make it easier for the user to avoid errors. So if my understanding is correct, I would propose to go with an approach closer to my proposal with two passes, even if it results in a bit of extra code.

src/peft/utils/other.py
```diff
@@ -1402,6 +1410,17 @@ def set_additional_trainable_modules(model, peft_config, model_config, adapter_n
             activate_adapter=activate_adapter,
         )
 
+    if getattr(peft_config, "modules_to_tie", None) is not None:
```
romitjain, line 1411: @BenjaminBossan This is what I am referring to.

`_set_trainable` for `modules_to_tie` will always be called after `_set_trainable` for `modules_to_save`

romitjain:
@BenjaminBossan 

>  If I understand your proposal correctly, you would apply the weight tying on the fly (while iterating through the modules).

I would add them to a separate list called `modules_to_tie`. The modules will not be tied yet.

>  If the module that should be tied is wrapped before the module that it's being tied to, your proposed approach would fail. Is my understanding correct here?

In my proposed flow, the modules in `modules_to_save` will always be wrapped first, and only then, weight tying will be called. Please see the flow of `set_additional_trainable_modules` function.

src/peft/tuners/lora/config.py
```diff
@@ -663,6 +663,16 @@ class LoraConfig(PeftConfig):
     arrow_config: Optional[ArrowConfig] = field(
         default=None, metadata={"help": "The necessary config to apply arrow routing on the model."}
     )
+    ensure_weight_tying: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Whether to tie weights or not after peft initialization. "
+                "This will ensure that the adapters added to the tied layers "
+                "are also tied."
```
BenjaminBossan: Let's mention that right now, this only applied to `modules_to_save`, not the LoRA weights.

src/peft/tuners/tuners_utils.py
```diff
@@ -690,6 +690,36 @@ def inject_adapter(
         # in a bad (half-initialized) state.
         self._check_new_adapter_config(peft_config)
 
+        modules_to_save = (
```
BenjaminBossan: Okay, so let's move this whole block into a sub-method on `BaseTuner`, e.g. `def _check_tied_modules`. On the `BaseTuner`, if the `peft_config` has `modules_to_save`, you can do a basic check if `modules_to_save` targets a tied module. This way, all PEFT methods with `modules_to_save` benefit from having the warnings.

Then, on `LoraModel`, override `_check_tied_modules` with the logic below. There, you can remove the `PeftType.LORA` check.

src/peft/tuners/tuners_utils.py
```diff
@@ -690,6 +690,36 @@ def inject_adapter(
...
+            set(getattr(peft_config, "modules_to_save", [])) if getattr(peft_config, "modules_to_save", []) else set()
+        )
+        is_embedding_to_save = [m for m in modules_to_save if m in EMBEDDING_LAYER_NAMES]
```
BenjaminBossan: Makes more sense like this IMO:

```suggestion
        is_embedding_to_save = any(m in EMBEDDING_LAYER_NAMES for m in modules_to_save)
```

src/peft/utils/other.py
```diff
@@ -548,9 +548,17 @@ def update(self, adapter_name, **kwargs):
                 context_manager = deepspeed.zero.GatheredParameters(self.original_module.parameters(), modifier_rank=0)
                 break
 
+        tied_module = kwargs.get("tied_module", None)
```
BenjaminBossan: Let's add `tied_module` to the signature of `init_modules` with default `None`, instead of getting it from `kwargs` (but you can leave `kwargs` in the signature too, shouldn't hurt).

src/peft/utils/other.py
BenjaminBossan: Okay. Let's add a short comment here to further explain what is happening.

tests/test_initialization.py
```diff
@@ -110,6 +110,47 @@ def forward(self, x):
...
         return MyModule().eval().to(self.torch_device)
 
+    def get_lm_model(self, bias=True, tie_weights=True):
+        class MyModule(nn.Module):
+            def __init__(self):
+                super().__init__()
+
+                self.linear = nn.Linear(1000, 1000, bias=bias)
+                self.embed_tokens = nn.Embedding(1000, 1000)
+                self.conv2d = nn.Conv2d(100, 100, 3, bias=bias)
```
```suggestion
                self.embed_tokens = nn.Embedding(1000, 1000)
```

Not needed, right?

tests/test_initialization.py
```diff
@@ -110,6 +110,47 @@ def forward(self, x):
...
+
+            def forward(self, x):
+                x_int = (x * 100).int()
+                x_4d = x.reshape(1, 100, 10, 10)
+
+                return self.linear(x), self.embed(x_int), self.conv2d(x_4d)
```
BenjaminBossan: Since we don't need to call `forward`, let's remove this to focus on what matters for these tests.

tests/test_initialization.py
```diff
@@ -110,6 +110,47 @@ def forward(self, x):
...
+                self.config = {"tie_word_embeddings": tie_weights}
+
+                if tie_weights:
+                    self.lm_head = nn.Linear(1000, 1000, bias=False)
+                    self.lm_head.weight = self.model.embed_tokens.weight
+                else:
+                    self.lm_head = nn.Linear(1000, 1000, bias=bias)
+
+            def forward(self, x):
+                return self.model(x)
```
BenjaminBossan: Also remove this.

tests/test_initialization.py
```diff
@@ -110,6 +110,47 @@ def forward(self, x):
 
         return MyModule().eval().to(self.torch_device)
 
+    def get_lm_model(self, bias=True, tie_weights=True):
```
BenjaminBossan: Let's add a small explanation of what this model is used for.

src/peft/tuners/tuners_utils.py
```diff
@@ -690,6 +690,36 @@ def inject_adapter(
         # in a bad (half-initialized) state.
         self._check_new_adapter_config(peft_config)
 
+        modules_to_save = (
+            set(getattr(peft_config, "modules_to_save", [])) if getattr(peft_config, "modules_to_save", []) else set()
```
BenjaminBossan: This is enough, right?
```suggestion
            set(getattr(peft_config, "modules_to_save", []))
```

tests/test_initialization.py
```diff
@@ -1566,6 +1607,98 @@ def test_multiple_configs_with_bias_raises(self, tmp_path):
         config2 = LoraConfig(target_modules=["linear"], bias="none")
         model.add_adapter("other", config2)  # does not raise
 
+    def test_weight_tying_tied_model(self):
+        # If weight tying is enabled and `embed_tokens`
+        # is passed as a `modules_to_save`, it needs to be ensured
+        # that lm_head is tied to the adapter added to `embed_tokens`
+
+        from peft.utils.other import ModulesToSaveWrapper
```
BenjaminBossan: Can be imported at root level instead of locally on each test.

tests/test_initialization.py
```diff
@@ -1566,6 +1607,98 @@ def test_multiple_configs_with_bias_raises(self, tmp_path):
...
+        # is passed as a `modules_to_save`, it needs to be ensured
+        # that lm_head is tied to the adapter added to `embed_tokens`
+
+        model = self.get_lm_model()
+        embed_token_config = LoraConfig(
+            target_modules=["linear"],
+            ensure_weight_tying=True,
+        )
+
+        model = get_peft_model(model, embed_token_config)
```
BenjaminBossan: How about checking that there is no warning related to weight tying here?

src/peft/tuners/tuners_utils.py
```diff
@@ -690,6 +690,36 @@ def inject_adapter(
...
+
+            if getattr(peft_config, "ensure_weight_tying", False):
+                peft_config.modules_to_tie = missing_keys
+            elif not getattr(peft_config, "ensure_weight_tying", False):
+                msg = (
+                    "Model has `tie_word_embeddings=True` and the tied layer is part of the adapter, "
+                    "but `ensure_weight_tying` is not set to True. "
+                    "This can lead to complications, for example when merging the adapter "
+                    "or converting your model to formats other than safetensors. "
+                    "See for discussion: https://github.com/huggingface/peft/issues/2777"
```
```suggestion
                    "Check the discussion here: https://github.com/huggingface/peft/issues/2777"
```

src/peft/tuners/tuners_utils.py
romitjain: `modules_to_save` is set as `None` in `src/peft/mixed_model.py` which results in an error

src/peft/utils/other.py
romitjain: Did you mean `update`? Since this peice of code lies in the `update` method

romitjain: I have added in `update`

romitjain:
@BenjaminBossan I have resolved the comments. Let me know if I missed something?

src/peft/tuners/tuners_utils.py
```diff
@@ -1154,6 +1156,64 @@ def _get_tied_target_modules(self, model: nn.Module) -> list[str]:
...
+
+        return tied_weight_keys
+
+    def _check_tied_modules(self, model: nn.Module, peft_config):
+        """
+        Checks if any of the tied layers are targetted via `modules_to_save`
+        Updates the `peft_config.modules_to_tie` with any layers that needs to be tied
+        """
+        modules_to_save = (
+            set(getattr(peft_config, "modules_to_save", [])) if getattr(peft_config, "modules_to_save", []) else set()
```
```suggestion
            set(getattr(peft_config, "modules_to_save", []) or [])
```

This should work with `None`.

src/peft/tuners/tuners_utils.py
```diff
@@ -1154,6 +1156,64 @@ def _get_tied_target_modules(self, model: nn.Module) -> list[str]:
...
+            tied_weight_keys.extend(names)
+        for name, submodule in model.named_children():
+            local_prefix = f"{prefix}.{name}" if prefix else name
+            tied_weight_keys.extend(self._get_tied_weight_keys(submodule, prefix=local_prefix))
+
+        tied_weight_keys = [".".join(n.split(".")[:-1]) for n in tied_weight_keys]
+
+        return tied_weight_keys
+
+    def _check_tied_modules(self, model: nn.Module, peft_config):
```
BenjaminBossan, line 1197: IMO, it makes more sense to just do some basic checks here in this generic method of the `BaseTuner`, and add a more LoRA-specific implementation to `LoraModel`. That way, there is no need to check for `peft_config.peft_type == PeftType.LORA`.

tests/test_initialization.py
```diff
@@ -110,6 +111,45 @@ def forward(self, x):
...
+        # to test weight tying of adapters
+        class MyModule(nn.Module):
+            def __init__(self):
+                super().__init__()
+
+                self.embed_tokens = nn.Embedding(1000, 1000)
+                self.linear = nn.Linear(1000, 1000, bias=bias)
+
+            def forward(self, x):
+                return
```
BenjaminBossan: It's safe to completely remove the `forward` method here and below, as they're not used.

src/peft/tuners/tuners_utils.py
romitjain: I don't follow.

The `_check_tied_modules` function is called inside `inject_adaptors` which the `LoraModel` does not inherit. 

What you suggested makes sense, I can remove Lora specific changes from this function and handle that condition outside. I don't know where to place the call `self._check_tied_modules(model, peft_config)`. Do you have any suggestions?

PS: I have addressed rest of your comments

BenjaminBossan: This should work. `LoraModel` inherits from `BaseTuner`, thus it does have `LoraModel.inject_adapter`. So if we have:

```python
class BaseTuner:
    def inject_adapter(...):
        self.self._check_tied_modules(...)
        ...
    def _check_tied_modules(...): ...
```

and

```python
class LoraModel(BaseTuner):
    def _check_tied_modules(...): ...
```

then when calling `LoraModel.inject_adapter`, it calls `BaseTuner.inject_adapter`, which in turn will call `LoraModel._check_tied_modules`.

romitjain: Got it, so I should initialize `_check_tied_modules` as an abstract method and implement in `LoraModel`.

BenjaminBossan: Hmm, I was about to say that we can have some basic checks for existence of tied weights and `modules_to_save` here and give a warning for non-LoRA; for LoRA, we would override the method and give more specific warnings. But now I wonder: Why only implement this for LoRA in the first place? It should work for other PEFT methods too, at least I don't see why it wouldn't.

romitjain: Yes, that was my original thought as well, but since I have not worked with other PEFT methods apart from LoRA, I am not sure if I can comment if it would work or not just OOB.

Alternatively, we can expose this for LoRA (or other similar PEFT wherever you are confident) and open it for all PEFT based on feedback?

BenjaminBossan: Your proposal makes sense, let's keep the scope small for now, and later, it can be extended to other PEFT methods if there is a need.

This means, let's keep just a generic check here in `BaseTuner._check_tied_modules` and then in `LoraModel._check_tied_modules`, do the more specific check.

romitjain: I have updated my implementation by adding an internal method `_maybe_add_modules_to_tie`. This is the cleanest approach I could find. Let me know your thoughts!

src/peft/tuners/tuners_utils.py
```diff
@@ -1154,6 +1156,77 @@ def _get_tied_target_modules(self, model: nn.Module) -> list[str]:
...
+        else:
+            if is_embedding_to_save and tied_weight_keys:
+                msg = (
+                    "Model has `tie_word_embeddings=True` and a tied layer is part of the adapter, "
+                    "but `ensure_weight_tying` is not set to True. "
+                    "This can lead to complications, for example when merging the adapter "
+                    "or converting your model to formats other than safetensors. "
+                    "Check the discussion here: https://github.com/huggingface/peft/issues/2777"
+                )
+                warnings.warn(msg)
```
BenjaminBossan: I think this message has the potential to confuse users. `ensure_weight_tying` is only implemented for LoRA right now but this message can also show up for non-LoRA methods, right?

romitjain: Ahh yes, I missed this.

Fixed it now

src/peft/tuners/tuners_utils.py
```diff
@@ -1154,6 +1156,76 @@ def _get_tied_target_modules(self, model: nn.Module) -> list[str]:
...
+            elif not is_embedding_to_save and tied_weight_keys:
+                warnings.warn(
+                    "You have requested `ensure_weight_tying`, but no tied modules are added in `modules_to_save`"
+                )
+
+            elif not tied_weight_keys:
+                warnings.warn("You have requested `ensure_weight_tying`, but no tied modules were found in the model")
+
+        else:
+            self._maybe_warn_ensure_weight_tying(is_embedding_to_save, tied_weight_keys)
```
BenjaminBossan: I think this is a bit too specific as a method. How about removing `_maybe_warn_ensure_weight_tying` and just adding this:

```python
elif is_embedding_to_save and tied_weight_keys:
    if hasattr(peft_config, "ensure_weight_tying"):
        # warning from https://github.com/huggingface/peft/pull/2803/files#diff-52158f12774d14e2d4f7f5236439c310680c3c1ed6f87229827bc05bd38e01d4R1189
        warnings.warn("Model has `tie_word_embeddings=True` ...")
    else:  # no ensure_weight_tying option, warn user about potential issues
        warnings.warn(<slightly different warning>)
```

src/peft/tuners/tuners_utils.py
```diff
@@ -1154,6 +1156,76 @@ def _get_tied_target_modules(self, model: nn.Module) -> list[str]:
...
+            tied_weight_keys.extend(names)
+        for name, submodule in model.named_children():
+            local_prefix = f"{prefix}.{name}" if prefix else name
+            tied_weight_keys.extend(self._get_tied_weight_keys(submodule, prefix=local_prefix))
+
+        tied_weight_keys = [".".join(n.split(".")[:-1]) for n in tied_weight_keys]
+
+        return tied_weight_keys
+
+    def _maybe_add_modules_to_tie(self, peft_config, tied_weight_keys):
```
BenjaminBossan: I have a slight preference for removing the `maybe_` here

```suggestion
    def _add_modules_to_tie(self, peft_config, tied_weight_keys):
```

tests/test_initialization.py
```diff
@@ -1566,6 +1600,93 @@ def test_multiple_configs_with_bias_raises(self, tmp_path):
...
+        embed_np = dict(model.base_model.model.model.embed_tokens.named_parameters())
+        lm_head_np = dict(model.base_model.model.lm_head.named_parameters())
+
+        for k in embed_np.keys():
+            assert torch.allclose(embed_np[k], lm_head_np[k])
+            assert embed_np[k] is lm_head_np[k]
+
+    def test_weight_tying_non_tied_model(self):
+        from peft.utils.other import ModulesToSaveWrapper
+
```
```suggestion
```

tests/test_initialization.py
```diff
@@ -1566,6 +1600,93 @@ def test_multiple_configs_with_bias_raises(self, tmp_path):
...
+        assert isinstance(model.base_model.model.model.embed_tokens, ModulesToSaveWrapper), (
+            "Embed tokens is not added in Modules to Save"
+        )
+        assert isinstance(model.base_model.model.lm_head, torch.nn.modules.linear.Linear), (
+            "LM head is not of type nn.linear"
+        )
+
+    def test_not_weight_tying_tied_model(self):
+        from peft.utils.other import ModulesToSaveWrapper
+
```
```suggestion
```

tests/test_initialization.py
```diff
@@ -110,6 +111,39 @@ def forward(self, x):
 
         return MyModule().eval().to(self.torch_device)
 
+    def get_lm_model(self, bias=True, tie_weights=True):
```
BenjaminBossan: Since the `bias` argument is not required for the weight tying tests, let's remove it.

tests/test_initialization.py
```diff
@@ -1566,6 +1600,93 @@ def test_multiple_configs_with_bias_raises(self, tmp_path):
         config2 = LoraConfig(target_modules=["linear"], bias="none")
         model.add_adapter("other", config2)  # does not raise
 
+    def test_weight_tying_tied_model(self):
```
BenjaminBossan: I'm rethinking how to best deal with these tests. Right now, they are tied to LoRA, but with the recent changes, other PEFT methods will also benefit (even if it's just a warning about potential issues). Therefore, I propose the following: Let's create a new test class, say `TestWeightTying`. Move `get_lm_model` and the test methods (`test_weight_tying_tied_model` etc.) into that class. Add `_lora` to the name of these test methods.

Then, let's add tests for a non-LoRA PEFT method that supports `modules_to_save`, e.g. LoKr. The test can be very similar to the LoRA tests, just that there is no `ensure_weight_tying` option and that the warning may look different. Add a comment that LoKr is just one example of many PEFT methods that could be chosen for this test.

romitjain: Done @BenjaminBossan

src/peft/tuners/tuners_utils.py
```diff
@@ -1154,6 +1156,86 @@ def _get_tied_target_modules(self, model: nn.Module) -> list[str]:
...
+        )
+        warnings.warn(msg)
+
+    def _check_tied_modules(self, model: nn.Module, peft_config):
+        """
+        Checks if any of the tied layers are targetted via `modules_to_save`. Updates the `peft_config.modules_to_tie`
+        with any layers that needs to be tied
+        """
+        modules_to_save = set(getattr(peft_config, "modules_to_save", []) or [])
+        is_embedding_to_save = any(m in EMBEDDING_LAYER_NAMES for m in modules_to_save)
```
githubnemo, line 1203: Should we use `model.get_input_embeddings` if it is available instead of guessing by name?

githubnemo (commented):
This change looks cool :)

One question, otherwise LGTM.

src/peft/tuners/tuners_utils.py
romitjain: @githubnemo 
There was some precedence in using this [here](https://github.com/huggingface/peft/blob/main/src/peft/tuners/tuners_utils.py#L1167) and [here](https://github.com/huggingface/peft/blob/main/src/peft/peft_model.py#L200)

githubnemo: makes sense. I think in the future we shouldn't rely on the embedding names if possible (e.g., custom models wouldn't be possible). but it's ok for me to leave it as is and do this change separately.

romitjain:
Thanks for merging @BenjaminBossan 

yes, I can work on a similar PR for tied weights. I think now with the base work done, it should be easier to integrate. I am away this week, so I’ll come back to it next week.

PR #2805, opened 2025-09-29
minor changes to OFT to make it faster

zqiu24:
Dear All,

I have noticed some minor issues in the forward of OFT, the most important is that the indexing is on cpu (instead of on the device), resulting in slower training.

I have made some minor modifications, I tested on 8B finetuning and it is approximately 30% faster.

Hope for a quick review :) @BenjaminBossan 

Thank you so much.

Best,

### DIFF
--- a/src/peft/tuners/oft/config.py
+++ b/src/peft/tuners/oft/config.py
@@ -14,8 +14,11 @@
 
 from __future__ import annotations
 
+import warnings
 from dataclasses import dataclass, field
 from typing import Literal, Optional, Union
+
+import packaging.version
 
 from peft.config import PeftConfig
 from peft.utils import PeftType
@@ -193,4 +196,18 @@ class OFTConfig(PeftConfig):
                 "with the latest version of OFT. Please retrain your adapter weights with newer PEFT versions. "
                 "Alternatively, downgrade PEFT to version 0.13.0 to use the old adapter weights."
             )
+        if kwargs.get("use_cayley_neumann", False):
+            peft_version = kwargs.get("peft_version", "0.0.0")  # if not present, set a low dummy version
+            # remove commit hash, if present
+            peft_version = peft_version.partition("@")[0]
+            parsed_version = packaging.version.Version(peft_version)
+            min_version = packaging.version.Version("0.18.0")
+            # note: config.peft_version was added in 0.18.0, so if it's missing, it means we're below min version
+            if parsed_version < min_version:
+                msg = (
+                    "The cayley-neumann parameterization has been slightly changed to be more numerically stable in "
+                    "PEFT 0.18.0. Please retrain your adapter weights with newer PEFT versions. Alternatively, "
+                    "downgrade PEFT to version 0.17.0 to use the old parameterization."
+                )
+                warnings.warn(msg)
         return super().check_kwargs(**kwargs)
--- a/src/peft/tuners/oft/layer.py
+++ b/src/peft/tuners/oft/layer.py
@@ -97,7 +97,9 @@ class OFTRotationModule(nn.Module):
         self.use_cayley_neumann = use_cayley_neumann
         self.num_cayley_neumann_terms = num_cayley_neumann_terms
         # Create indices for upper triangle (excluding diagonal)
-        self.rows, self.cols = torch.triu_indices(block_size, block_size, 1)
+        rows, cols = torch.triu_indices(block_size, block_size, 1)
+        self.register_buffer("rows", rows, persistent=False)
+        self.register_buffer("cols", cols, persistent=False)
 
     def _pytorch_skew_symmetric(self, vec, block_size):
         batch_size = vec.shape[0]
@@ -139,9 +141,11 @@ class OFTRotationModule(nn.Module):
                     R.add_(Q_squared, alpha=2.0)
 
                     Q_power = Q_squared
-                    for i in range(3, num_neumann_terms):
+                    for _ in range(3, num_neumann_terms - 1):
                         Q_power = torch.bmm(Q_power, Q_skew)
                         R.add_(Q_power, alpha=2.0)
+                    Q_power = torch.bmm(Q_power, Q_skew)
+                    R.add_(Q_power)
         else:
             id_mat = (
                 torch.eye(Q_skew.shape[-1], device=Q_skew.device)
@@ -621,9 +625,13 @@ class Linear(nn.Module, OFTLayer):
             if active_adapter in self.oft_R.keys():
                 oft_mat = self.get_delta_weight(active_adapter)
 
+                previous_dtype = oft_mat.dtype
+                if previous_dtype != torch.float32:
+                    oft_mat = oft_mat.to(torch.float32)
+
                 orig_weights = self.get_base_layer().weight.data
                 orig_weights = torch.transpose(orig_weights, 0, 1)
-                orig_weights = torch.mm(oft_mat.t(), orig_weights.to(oft_mat.dtype))
+                orig_weights = torch.mm(torch.linalg.inv(oft_mat).to(previous_dtype), orig_weights.to(previous_dtype))
                 orig_weights = torch.transpose(orig_weights, 0, 1)
 
                 base_layer.weight.data = orig_weights.to(orig_dtype)
@@ -855,13 +863,17 @@ class Conv2d(nn.Module, OFTLayer):
             if active_adapter in self.oft_R.keys():
                 oft_mat = self.get_delta_weight(active_adapter)
 
+                previous_dtype = oft_mat.dtype
+                if previous_dtype != torch.float32:
+                    oft_mat = oft_mat.to(torch.float32)
+
                 orig_weights = self.get_base_layer().weight.data.clone()
                 orig_weights = orig_weights.view(
                     self.out_features,
                     self.in_features * self.get_base_layer().kernel_size[0] * self.get_base_layer().kernel_size[0],
                 )
                 orig_weights = torch.transpose(orig_weights, 0, 1)
-                orig_weights = torch.mm(oft_mat.t(), orig_weights.to(oft_mat.dtype))
+                orig_weights = torch.mm(torch.linalg.inv(oft_mat).to(previous_dtype), orig_weights.to(previous_dtype))
                 orig_weights = torch.transpose(orig_weights, 0, 1)
                 orig_weights = orig_weights.view(
                     self.out_features,


src/peft/tuners/oft/layer.py
```diff
@@ -114,6 +114,7 @@ def _pytorch_skew_symmetric_inv(self, matrix, block_size):
         vec = matrix[:, self.rows, self.cols]
         return vec
 
+    @torch.compile
```
BenjaminBossan: Right now, we don't `torch.compile` any code but instead leave it to the user to choose if they want to compile or not. Testing on my machine with a 4090, I also don't see a big difference (6582 vs 6532 tokens / sec). Did you find this to help a lot in your setting? If not, I'd suggest to remove it.

src/peft/tuners/oft/layer.py
```diff
@@ -139,9 +140,11 @@ def _cayley_batch(
                     R.add_(Q_squared, alpha=2.0)
 
                     Q_power = Q_squared
-                    for i in range(3, num_neumann_terms):
+                    for _ in range(3, num_neumann_terms - 1):
                         Q_power = torch.bmm(Q_power, Q_skew)
                         R.add_(Q_power, alpha=2.0)
+                    Q_power = torch.bmm(Q_power, Q_skew)
+                    R.add_(Q_power)
```
BenjaminBossan, line 148: Okay, so IIUC, this makes it so that the `.add_(..., power=2.0)` is not called on the last iteration. Why is that so? 

This will change the results of using OFT, i.e. existing checkpoints would stop giving the same results. Therefore, I would really like to avoid this change. If the benefit is too large to ignore, we need to add a warning when a user loads an old checkpoint with `use_cayley_neumann` to tell them that the outputs are changing (meaning the user either needs to pin the PEFT version or train a new OFT model).

src/peft/tuners/oft/layer.py
```diff
@@ -248,6 +251,10 @@ def forward(self, x):
         if required_dtype != self.weight.dtype:
             x = x.to(self.weight.dtype)
 
+        if self.rows.device != self.weight.device:
```
BenjaminBossan: IIRC, `tensor.to(device)` is a no-op if the device is already the same, i.e. we cna remove the device check.

src/peft/tuners/oft/layer.py
```diff
@@ -248,6 +251,10 @@ def forward(self, x):
...
+            self.rows = self.rows.to(self.weight.device)
+            self.cols = self.cols.to(self.weight.device)
```
BenjaminBossan: I wonder: Should we also move `rows` and `cols` to the correct device directly when they're initialized?

https://github.com/huggingface/peft/blob/4469af57a08b0a751b7b1359d431c03a22b33116/src/peft/tuners/oft/layer.py#L463-L473

As these tensors are quite small, I think the memory impact is negligible but we avoid moving them during each `forward` call. I would still keep this code here to be safe (as it's a no-op if they're already on the right device).

BenjaminBossan (changes requested):
Thanks for this PR to make OFT faster, I found a few issues, please check my comments.

In my local testing, this PR increased the speed from 5726 tok/sec to 6582 tok/sec using the default MetaMath experiment for OFT (~15% faster).

src/peft/tuners/oft/layer.py
zqiu24: Thanks for the quick reply! The thing is that I noticed it will speed it up a bit (I deliberately do not add the specific configurations like dynamic etc. to avoid error), but we can leave it out.

src/peft/tuners/oft/layer.py
zqiu24: Thanks for the remark. This is a small error I made, I looked at the Cayley-Neumann parameterization (which OFT uses to orthogonalize matrices), it actually needs the last term to not have this 2.0. But change should be minimal, since it is close to identity, but it should be more stable, since it reduces the orthogonalization error. I ablationed locally that the less the orthogonality, the better the training is. So scientifically it is more correct to have this way. So I would definitely want to incorporate this. Best,

src/peft/tuners/oft/layer.py
zqiu24: Yes, we could also do this, but when it is initialized I think it is on cpu? (I checked the device of the base layer weight, it is on cpu), so how should I know the correct device? Best,

src/peft/tuners/oft/layer.py
zqiu24: General question: if we have custom kernels (for example triton kernel, is it allowed to add?)

BenjaminBossan: > The thing is that I noticed it will speed it up a bit

How much difference does it make for you? If it's not significantly more than I observed, I'd say, let's remove it for now.

> General question: if we have custom kernels (for example triton kernel, is it allowed to add?)

Good question. So at HF, there is a general push to make it easier to optionally make use of kernels: https://huggingface.co/blog/hello-hf-kernels. In transformers, there is already an integration, though the API is still early stage (I think there is no option for fine-grained control of what kernels to use).

For PEFT, we plan to add the same functionality but we're still working out how to exactly implement it. Once it's there, we will happily accept kernels for specific PEFT methods (and migrate existing kernels there). Until then, feel free to add your OFT kernel to the HF kernel hub, we could use it for testing the kernels integration.

src/peft/tuners/oft/layer.py
BenjaminBossan: Okay, so in this case, we need to add a check and give the user an appropriate warning. The best place is to extend this method:

https://github.com/huggingface/peft/blob/815956b9b85f893e5acd2dd1d8391954b0f93e07/src/peft/tuners/oft/config.py#L182

Here, we should add another check, namely if `use_cayley_neumann is True` and if the PEFT version is lower than 0.18 (which would be the next release with this PR included). We added a feature to store the PEFT version in the config recently (#2782) but need to consider that older PEFT configs don't have this attribute. Thus, to check the version, please use something like this:

```python
# oft/config.py
...

import packaging.version

...

    def check_kwargs(cls, **kwargs):
        ...

        # TODO add comment to explain why
        if kwargs["use_caylay_neumann"]:
            peft_version = kwargs.get("peft_version", "unknown")
            parsed_version = packaging.version.Version(peft_version)
            min_version = packaging.version.Version("0.18.0")
            # note: config.peft_version was added in 0.18.0, so if it's missing, it means we're below min version
            if (peft_version == "unknown") or (parsed_version < min_version):
                msg = "warning message that explains what is happening"
                warnings.warn(msg)

        return super().check_kwargs(**kwargs)
```

Ideally, we should also add a test for this.

src/peft/tuners/oft/layer.py
BenjaminBossan: I think at this point, the weights of the base model should generally already be on the accelerator. E.g. when I run these tests with a GPU:

```
pytest tests/test_custom_models.py -k "test_forward_float16 and oft and not boft"
```

and I add this check before `OFTRotationModule` is assigned:

```python
base_layer_device = self.get_base_layer().device
assert base_layer_device.type == "cuda"
```

it passes. So we can add the device as input argument to `OFTRotationModule` and then ensure that it's moved to the right device. WDYT?

src/peft/tuners/oft/layer.py
zqiu24: Thanks for the information :) In another project I am building kernels for this orthgonalization, which makes it faster and more memory-efficient than the current torch.compile, but sure, we can add it later :)
Let's just remove the torch.compile for now.

zqiu24:
@BenjaminBossan I just pushed a newer version. Best

src/peft/tuners/oft/config.py
```diff
@@ -193,4 +196,12 @@ def check_kwargs(cls, **kwargs):
                 "with the latest version of OFT. Please retrain your adapter weights with newer PEFT versions. "
                 "Alternatively, downgrade PEFT to version 0.13.0 to use the old adapter weights."
             )
+        if kwargs["use_caylay_neumann"]:
+            peft_version = kwargs.get("peft_version", "unknown")
+            parsed_version = packaging.version.Version(peft_version)
+            min_version = packaging.version.Version("0.18.0")
+            # note: config.peft_version was added in 0.18.0, so if it's missing, it means we're below min version
+            if (peft_version == "unknown") or (parsed_version < min_version):
+                msg = "warning message that explains what is happening"
```
BenjaminBossan: Well, this message was just a placeholder :-D Could you please add a short explanation here?

src/peft/tuners/oft/layer.py
```diff
@@ -98,6 +100,8 @@ def __init__(
         self.num_cayley_neumann_terms = num_cayley_neumann_terms
         # Create indices for upper triangle (excluding diagonal)
         self.rows, self.cols = torch.triu_indices(block_size, block_size, 1)
+        self.rows.to(device)
+        self.cols.to(device)
```
BenjaminBossan: As `rows` and `cols` are tensors, calling `.to()` is not in-place operations, you have to re-assign the values.

src/peft/tuners/oft/layer.py
```diff
@@ -470,6 +476,7 @@ def update_layer(
             block_share=block_share,
             use_cayley_neumann=use_cayley_neumann,
             num_cayley_neumann_terms=num_cayley_neumann_terms,
+            device=self.weight.device,
```
BenjaminBossan: Slightly more robust:

```suggestion
            device=self.get_base_layer().weight.device,
```

src/peft/tuners/oft/layer.py
```diff
@@ -770,6 +777,7 @@ def update_layer(
             kernel_size=base_layer.kernel_size,
             use_cayley_neumann=use_cayley_neumann,
             num_cayley_neumann_terms=num_cayley_neumann_terms,
+            device=self.weight.device,
```
BenjaminBossan: Slightly more robust:

```suggestion
            device=self.get_base_layer().weight.device,
```

src/peft/tuners/oft/config.py
```diff
@@ -193,4 +196,12 @@ def check_kwargs(cls, **kwargs):
                 "with the latest version of OFT. Please retrain your adapter weights with newer PEFT versions. "
                 "Alternatively, downgrade PEFT to version 0.13.0 to use the old adapter weights."
             )
+        if kwargs["use_caylay_neumann"]:
```
BenjaminBossan: Let's make this more backwards compatible, also, there was a typo:

```suggestion
        if kwargs.get("use_cayley_neumann", False):
```

src/peft/tuners/oft/config.py
```diff
@@ -193,4 +196,12 @@ def check_kwargs(cls, **kwargs):
...
+            peft_version = kwargs.get("peft_version", "unknown")
+            parsed_version = packaging.version.Version(peft_version)
+            min_version = packaging.version.Version("0.18.0")
+            # note: config.peft_version was added in 0.18.0, so if it's missing, it means we're below min version
+            if (peft_version == "unknown") or (parsed_version < min_version):
```
BenjaminBossan:
```suggestion
            peft_version = kwargs.get("peft_version", "0.0.0")  # if not present, set a low dummy version
            parsed_version = packaging.version.Version(peft_version)
            min_version = packaging.version.Version("0.18.0")
            # note: config.peft_version was added in 0.18.0, so if it's missing, it means we're below min version
            if parsed_version < min_version:
```

I just tested and parsing with `"unknown"` throws an error, so let's put a dummy value here.

src/peft/tuners/oft/layer.py
zqiu24: Done.

src/peft/tuners/oft/layer.py
zqiu24: Done.

src/peft/tuners/oft/config.py
zqiu24: Sure:)

src/peft/tuners/oft/config.py
zqiu24: Done.

src/peft/tuners/oft/config.py
zqiu24: Done.

src/peft/tuners/oft/layer.py
zqiu24: Done.

zqiu24:
@BenjaminBossan Just updated the comments. Best,

zqiu24:
@BenjaminBossan What do these error messages mean? Best,

src/peft/tuners/oft/config.py
```diff
@@ -193,4 +196,12 @@ def check_kwargs(cls, **kwargs):
                 "with the latest version of OFT. Please retrain your adapter weights with newer PEFT versions. "
                 "Alternatively, downgrade PEFT to version 0.13.0 to use the old adapter weights."
             )
+        if kwargs.get("use_cayley_neumann", False):
+            peft_version = kwargs.get("peft_version", "0.0.0")  # if not present, set a low dummy version
+            parsed_version = packaging.version.Version(peft_version)
```
BenjaminBossan, line 203:
```suggestion
            # remove commit hash, if present
            peft_version = peft_version.partition("@")[0]
            parsed_version = packaging.version.Version(peft_version)
```

zqiu24:
@BenjaminBossan Thanks. Just updated.

src/peft/tuners/oft/layer.py
```diff
@@ -470,6 +476,7 @@ def update_layer(
             block_share=block_share,
             use_cayley_neumann=use_cayley_neumann,
             num_cayley_neumann_terms=num_cayley_neumann_terms,
+            device=self.get_base_layer().weight.device,
```
BenjaminBossan: Ah, this doesn't work if the device is `meta`, as we'll get:

>       self.weight = self.weight.to(device)
>                       ^^^^^^^^^^^^^^^^^^^^^^
> E       NotImplementedError: Cannot copy out of meta tensor; no data!

Let's do something like:

```python
device = self.get_base_layer().weight.device
[...]
self.oft_R[adapter_name] = OFTRotationModule(
    ...
    device=device if device.type != "meta" else None,

File: src/peft/tuners/oft/layer.py [hunk 13]
```diff
@@ -770,6 +777,7 @@ def update_layer(
             kernel_size=base_layer.kernel_size,
             use_cayley_neumann=use_cayley_neumann,
             num_cayley_neumann_terms=num_cayley_neumann_terms,
+            device=self.get_base_layer().weight.device,
```
BenjaminBossan: Same comment as above

src/peft/tuners/oft/layer.py
zqiu24: @BenjaminBossan How to reproduce this with meta?

zqiu24: @BenjaminBossan but then it is not correct right? During initialization it is meta, so the variables val / cols will never get shifted to gpu (this is why the code is currently slow)?

zqiu24: @BenjaminBossan I think I can leave out this self.weight = self.weight.to(device), since it had worked correctly before, I just want to make sure that the self.cols and self.rows get shifted correctly

BenjaminBossan: To reproduce the error, just run the failing tests:

`pytest tests/test_custom_models.py::TestPeftCustomModel::test_load_model_low_cpu_mem_usage -k "oft and not boft"`

> During initialization it is meta, so the variables val / cols will never get shifted to gpu (this is why the code is currently slow)?

So the situation is the following: A user loads a trained PEFT adapter with `low_cpu_mem_usage=True`, the weights are at first created on meta device, but then when the actual weights are loaded, they are moved to the correct device. This should happen at the end of `update_layer`, when `self._move_adapter_to_device_of_base_layer(adapter_name)` is called. I think this method won't work on `rows` and `cols` though, as they are just tensors. I think what can be done:

1. Make them non-persistent buffers, I believe then they would be automatically be moved when `oft_R` is moved.
2. Override `_move_adapter_to_device_of_base_layer`, call `super()._move_adapter_to_device_of_base_layer`, then handle `rows` and `cols` explicitly.

zqiu24:
@BenjaminBossan Ok, the newest should be working. I used the non-persistent buffer. Best,

tests/test_initialization.py
```diff
@@ -1756,6 +1758,49 @@ def test_vblora_with_incompatible_vector_length_with_out_features(self):
...
+
+        if peft_version is None:
+            del config_json["peft_version"]
+        else:
+            config_json["peft_version"] = peft_version
+
+        with open(tmp_path / "adapter_config.json", "w") as f:
+            json.dump(config_json, f)
+
+        msg = "TODO"  # <= replace with final warning message
```
BenjaminBossan: Please replace it with the correct error message (start of the message is fine).

src/peft/tuners/oft/config.py
```diff
@@ -193,4 +196,14 @@ def check_kwargs(cls, **kwargs):
...
             )
+        if kwargs.get("use_cayley_neumann", False):
+            peft_version = kwargs.get("peft_version", "0.0.0")  # if not present, set a low dummy version
+            # remove commit hash, if present
+            peft_version = peft_version.partition("@")[0]
+            parsed_version = packaging.version.Version(peft_version)
+            min_version = packaging.version.Version("0.18.0")
+            # note: config.peft_version was added in 0.18.0, so if it's missing, it means we're below min version
+            if parsed_version < min_version:
+                msg = "The cayley-neumann parameterization has been slightly changed to be more numerically stable in PEFT 0.18.0. Please retrain your adapter weights with newer PEFT versions. Alternatively, downgrade PEFT to version 0.17.0 to use the old parameterization."
```
BenjaminBossan:
```suggestion
                msg = (
                    "The cayley-neumann parameterization has been slightly changed to be more numerically stable in "
                    "PEFT 0.18.0. Please retrain your adapter weights with newer PEFT versions. Alternatively, "
                    "downgrade PEFT to version 0.17.0 to use the old parameterization."
                )
```

Let's add some line breaks to keep <= 120 chars.

tests/test_initialization.py
```diff
@@ -1756,6 +1758,49 @@ def test_vblora_with_incompatible_vector_length_with_out_features(self):
...
+        model = get_peft_model(self.get_model(), config)
+        model.save_pretrained(tmp_path)
+        del model
+
+        # overwrite the peft_version
+        with open(tmp_path / "adapter_config.json") as f:
+            config_json = json.load(f)
+
+        if peft_version is None:
+            del config_json["peft_version"]
```
```suggestion
            config_json.pop("peft_version", None)
```

In case the key does not exist.

BenjaminBossan (changes requested):
Thanks for the latest updates, going with buffers is okay for me. I found a few small remaining TODOs, after that the PR should be good. Please also run `make style` before committing.

src/peft/tuners/oft/config.py
zqiu24: OK, done

tests/test_initialization.py
zqiu24: Done.

zqiu24:
@BenjaminBossan Thanks, just updated again. Best.

BenjaminBossan:
I checked the failing CI and most failing tests are unrelated (anything with "TypeError: CLIPTextModel.__init__() got an unexpected keyword argument 'offload_state_dict'" can be ignored). However, there are 2 actual failures:

```
FAILED tests/test_custom_models.py::TestPeftCustomModel::test_disable_adapters_with_merging[Conv2d 1 OFT-Conv2d-OFTConfig-config_kwargs107] - assert False
 +  where False = <built-in method allclose of type object at 0x00007FFE3FBAB860>(tensor([[-1.1555e+01, -9.5367e-06],\n        [-4.0945e+01,  0.0000e+00]], grad_fn=<LogSoftmaxBackward0>), tensor([[-1.1542e+01, -9.7751e-06],\n        [-4.0898e+01,  0.0000e+00]]), atol=0.0001, rtol=0.0001)
 +    where <built-in method allclose of type object at 0x00007FFE3FBAB860> = torch.allclose
FAILED tests/test_custom_models.py::TestPeftCustomModel::test_disable_adapters_with_merging[Conv2d 4 OFT-Conv2d-OFTConfig-config_kwargs109] - assert False
 +  where False = <built-in method allclose of type object at 0x00007FFE3FBAB860>(tensor([[-1.1555e+01, -9.5367e-06],\n        [-4.0945e+01,  0.0000e+00]], grad_fn=<LogSoftmaxBackward0>), tensor([[-1.1547e+01, -9.6559e-06],\n        [-4.0910e+01,  0.0000e+00]]), atol=0.0001, rtol=0.0001)
 +    where <built-in method allclose of type object at 0x00007FFE3FBAB860> = torch.allclose
```

I could reproduce these errors locally. What surprises me is that the model output is quite difference compared to the main branch:

```python
# main
>>> torch.testing.assert_close(outputs_before, outputs_disabled)
*** AssertionError: Tensor-likes are not close!

Mismatched elements: 2 / 4 (50.0%)
Greatest absolute difference: 0.002262115478515625 at index (1, 0) (up to 1e-05 allowed)
Greatest relative difference: 5.868273365194909e-05 at index (0, 0) (up to 1.3e-06 allowed)

# this branch
>>> torch.testing.assert_close(outputs_before, outputs_disabled)
*** AssertionError: Tensor-likes are not close!

Mismatched elements: 2 / 4 (50.0%)
Greatest absolute difference: 0.047229766845703125 at index (1, 0) (up to 1e-05 allowed)
Greatest relative difference: 0.0011932472698390484 at index (0, 0) (up to 1.3e-06 allowed)
```

So the greatest absolute and relative errors increased by a factor of 20. Of course, we could just increase the tolerance further and the tests would pass, but I wonder if something strange could be going on with the new `_cayley_batch` calculation and conv layers.

zqiu24:
@BenjaminBossan Thanks for pointing this out. I will look into it. Best,

BenjaminBossan:
@zqiu24 Did you make any progress? If not, I think we can still merge with higher tolerance. It could just be that this example is degenerate but there is no issue in general. It would be good to have this PR merged before the next release (current plan: next week).

zqiu24:
@BenjaminBossan Thanks for this update. I will come back to that soon, I am a bit occupied with some other stuff. I will come back to you end of this week. Best,

zqiu24:
@BenjaminBossan Hi, Benjamin, I locally trained some models with the new formulation, I did not notice major differences (so no divergence or similar thing), so I think it should be fine. Should I change the tests? Best,

BenjaminBossan:
Thanks for checking, then this is just a test artifact. Please increase the tolerance for OFT and add a comment that this works well in practice, it's just the test having issues.

Also, could you please merge with/rebase on main? Then the PR should be good to go.

BenjaminBossan:
@zqiu24 I came across an OFT issue with unmerging, maybe we can check this and possibly provide a fix in this PR too.

Here is the reproducer first:

```python
import torch
from peft import OFTConfig, get_peft_model

config = OFTConfig(init_weights=False)

dummy_input = torch.randint(0, model.config.vocab_size, (1, 16))
model.eval()

logits = model(dummy_input)[0]
model.merge_adapter()
logits_merged = model(dummy_input)[0]
model.unmerge_adapter()
logits_unmerged = model(dummy_input)[0]

atol, rtol = 1e-4, 1e-4
assert torch.allclose(logits, logits_merged, atol=atol, rtol=rtol)
assert torch.allclose(logits, logits_unmerged, atol=atol, rtol=rtol)
```

This fails for me and I think the issue is [this line](https://github.com/huggingface/peft/blob/d0fa97413f9b642e5bb8ea4d290440b4316393b2/src/peft/tuners/oft/layer.py#L626) in the `unmerge` method.

To me, it's not obvious why taking the transpose of `oft_mat` would be the inverse operation when using `torch.mm`. Shouldn't the actual inverse be used, i.e.:

```python
orig_weights = torch.mm(torch.linalg.inv(oft_mat), orig_weights.to(oft_mat.dtype))
```

I know it's not efficient, but at least with this change, the code runs successfully. WDYT @zqiu24?

If we want to make this change, we need to update the other OFT layers too.

zqiu24:
> @zqiu24 I came across an OFT issue with unmerging, maybe we can check this and possibly provide a fix in this PR too.
> 
> Here is the reproducer first:
> 
> ```python
> import torch
> from peft import OFTConfig, get_peft_model
> from transformers import AutoModelForCausalLM
> 
> model_id = "facebook/opt-125m"
> model = AutoModelForCausalLM.from_pretrained(model_id)
> config = OFTConfig(init_weights=False)
> 
> model = get_peft_model(model, config)
> dummy_input = torch.randint(0, model.config.vocab_size, (1, 16))
> model.eval()
> 
> logits = model(dummy_input)[0]
> model.merge_adapter()
> logits_merged = model(dummy_input)[0]
> model.unmerge_adapter()
> logits_unmerged = model(dummy_input)[0]
> 
> atol, rtol = 1e-4, 1e-4
> assert torch.allclose(logits, logits_merged, atol=atol, rtol=rtol)
> assert torch.allclose(logits, logits_unmerged, atol=atol, rtol=rtol)
> ```
> 
> This fails for me and I think the issue is [this line](https://github.com/huggingface/peft/blob/d0fa97413f9b642e5bb8ea4d290440b4316393b2/src/peft/tuners/oft/layer.py#L626) in the `unmerge` method.
> 
> To me, it's not obvious why taking the transpose of `oft_mat` would be the inverse operation when using `torch.mm`. Shouldn't the actual inverse be used, i.e.:
> 
> ```python
> orig_weights = torch.mm(torch.linalg.inv(oft_mat), orig_weights.to(oft_mat.dtype))
> ```
> 
> I know it's not efficient, but at least with this change, the code runs successfully. WDYT @zqiu24?
> 
> If we want to make this change, we need to update the other OFT layers too.

Hi, thanks for this comment. Actually I think you are right, the reason why I do this transpose is because orthogonal matrix's inverse is its transpose, but apperently because of the approximation, it has more approximation error to orthogonality, this is why this unmerge has now bigger error. I agree to change it, since it is not in the training and will only be used quite rare.

zqiu24:
@BenjaminBossan I think the current version is able to pass the test without changing the tolerance? And it also fixes this unmerge issue. Best.

BenjaminBossan:
@zqiu24 Unfortunately, GH shows 54 commits and 127 changed files. Could you please try merging with/rebasing on main again, so that I can see the actual diff?

zqiu24:
@BenjaminBossan Thanks for the comment, does this work? Best,

BenjaminBossan (approved):
Thanks for the updates, the PR LGTM.

zqiu24:
@BenjaminBossan Thank you so much for the support:)

PR #2806, opened 2025-09-30
Some more TIP migration

githubnemo:
Either re-introduced or overlooked `<Tip>`s converted to new syntax.


BenjaminBossan (approved):
Thanks, these were probably missed because they were in the docstrings, not in the docs folder.

PR #2810, opened 2025-10-02
Add RWKV LoRA defaults and opt-in test

nirbo:
### Summary

  - Add rwkv and rwkv7 entries to the LoRA target-module mapping so Transformers-based RWKV checkpoints automatically receive adapters.
  - Document the new behaviour in the README with a short RWKV snippet.
  - Introduce tests/test_rwkv_lora.py, a tiny forward/backward regression guarded by PEFT_RUN_RWKV_TESTS=1 so the Triton compile remains opt-in.

  ### Testing
```
  make quality

  python -m pytest tests/test_rwkv_lora.py   # (skipped unless PEFT_RUN_RWKV_TESTS=1)

  # To exercise the RWKV-specific test:
  PEFT_RUN_RWKV_TESTS=1 python -m pytest tests/test_rwkv_lora.py
```


README.md
```diff
@@ -78,6 +78,22 @@ print(tokenizer.decode(outputs[0], skip_special_tokens=True))
 # prints something like: Preheat the oven to 350 degrees and place the cookie dough in a baking dish [...]
 ```
 
+> [!NOTE]
```
BenjaminBossan: IMO this README entry can be removed, as we generally don't highlight individual models here.

tests/test_rwkv_lora.py
```diff
@@ -0,0 +1,49 @@
+# Copyright 2025-present the HuggingFace Inc. team.
```
BenjaminBossan: There isn't really a need to add this test here, we don't have tests just to ensure that the default target modules are being set. I did, however, confirm that the test passes locally.

I could see an argument to add it to the general test suite, since RWKV has a different architecture:

https://github.com/huggingface/peft/blob/815956b9b85f893e5acd2dd1d8391954b0f93e07/tests/test_decoder_models.py#L56-L66

However, at the moment, the PEFT CI is already stressing the rate limit of HF Hub, so adding yet another model would not be a good idea. I think that if this situation relaxes and if we find that there is a big demand for RWKV finetuning with PEFT, we can consider that option.

BenjaminBossan (changes requested):
Thanks for adding default target modules for RWKV. I have some comments, please take a look.

BenjaminBossan:
gentle ping @nirbo

nirbo:
Sorry for the delay, been busy and this slipped my mind.
I've pushed the changes you requested.

@BenjaminBossan

PR #2811, opened 2025-10-03
Support Negative Weights When Merging LoRA Adapters #2796

### LINKED ISSUE #2796: Support negative weights when merging LoRA adapters
### Feature request

Currently, PEFT’s adapter merging only considers positive weights (except for `svd`). However, several works show that allowing negative weights can be useful:

- The [LoRA Hub](https://arxiv.org/abs/2307.13269) paper uses negative weights in its implementations ([Repo](https://github.com/sail-sg/lorahub/blob/main/lorahub/algorithm.py#L277))
- The [Task Arithmetic](https://arxiv.org/abs/2212.04089) paper also demonstrates the use of negative coefficients when combining adapters.

Adding support for negative weights in PEFT’s LoRA adapter merging would make the library more flexible and consistent with practices from recent literature.

### Your contribution

I have a locally tested version of PEFT that supports negative weights in adapter merging. I would be happy to submit this as a PR and contribute the feature to the library.

sambhavnoobcoder:
## Summary

This PR implements support for negative weights in LoRA adapter merging, resolving issue #2796. This feature enables task arithmetic and adapter subtraction capabilities as demonstrated in the LoRA Hub and Task Arithmetic papers.

---

### Current Issue

When attempting to merge LoRA adapters with negative weights using `add_weighted_adapter()`, the operation fails with a `ValueError: math domain error`. This error occurs because the current implementation computes `math.sqrt(weight * target.scaling[adapter])`, which is undefined for negative values.

### Why Negative Weights Matter

Several influential papers demonstrate the utility of negative weights in adapter merging:

1. **LoRA Hub Paper**: Uses negative weights in their implementation for combining multiple task-specific adapters
2. **Task Arithmetic Paper**: Demonstrates using negative coefficients to subtract unwanted task knowledge from models

Without negative weight support, users cannot:
- Perform task subtraction (removing unwanted behaviors)
- Implement LoRA Hub's weight optimization strategies
- Apply task arithmetic techniques to their models

---

### Mathematical Foundation

The implementation uses the following approach:

For each adapter `i` with weight `w_i` and scaling factor `s_i`:
- Compute `r_i = sqrt(|w_i * s_i|)` (magnitude using absolute value)
- Compute `sign_i = sign(w_i * s_i)` (preserve the sign)
- Apply `sign_i * r_i` to both A and B matrices

When merging:
- `A_merged = Σ (sign_i * r_i * A_i)`
- `B_merged = Σ (sign_i * r_i * B_i)`

This approach:
- Maintains the mathematical property: `(sign * sqrt(|w*s|) * A) @ (sign * sqrt(|w*s|) * B) = w*s*(A@B)`
- Works correctly for multi-adapter merging scenarios
- Is fully backward compatible with positive weights

---

### Code Changes

**Modified File**: `src/peft/tuners/lora/model.py`

**Change 1: Implementation** (lines 747-750 in `_generalized_task_arithmetic_weighted_adapter`)

The fix replaces the single line that was causing the error with three lines:
1. Compute `weight_with_scaling = weight * target.scaling[adapter]`
2. Determine the sign: `sign = 1 if weight_with_scaling >= 0 else -1`
3. Apply to weight calculation: `valid_weights.append(sign * math.sqrt(abs(weight_with_scaling)))`

This change affects the following combination types:
- `linear`
- `ties`
- `dare_linear`
- `dare_ties`
- `magnitude_prune`

**Change 2: Documentation** (lines 542-545)

Updated the docstring for `add_weighted_adapter` to document that weights can now be negative, with references to the LoRA Hub and Task Arithmetic papers.

### Methods Verified (No Changes Needed)

**SVD-based methods** (`_svd_generalized_task_arithmetic_weighted_adapter`):
- Already compatible with negative weights
- Uses `weight * target.scaling[adapter]` directly without square root
- Tested to ensure correct behavior with negative weights

**Cat combination**:
- Already compatible with negative weights
- Directly multiplies weights without square root operation
- Tested to ensure correct behavior with negative weights

### Strict Adherence to Proposed Solution

This implementation follows @valteu's approved solution exactly with **zero deviations**. We did not introduce any alternative approaches or optimizations beyond what was discussed and validated in the issue thread.

---

### Comprehensive Test Suite

Created `tests/test_negative_weights.py` with 21 comprehensive tests covering:

**1. Basic Negative Weight Functionality (4 tests)**
- Single negative weight with linear combination
- Multiple adapters with mixed positive/negative weights
- All negative weights
- Zero and negative weight combinations

**2. All Combination Types (11 tests)**
- Non-SVD methods: `linear`, `ties`, `dare_linear`, `dare_ties`, `magnitude_prune`
- SVD methods: `svd`, `ties_svd`, `dare_linear_svd`, `dare_ties_svd`, `magnitude_prune_svd`
- Cat combination with negative weights

**3. Mathematical Correctness (2 tests)**
- Verification that weight=-1.0 properly negates an adapter (both A and B matrices negated)
- Verification that weights [1.0, -1.0] produce correct difference between adapters
- Validation that merged matrices match expected formulas with scaling factors

**4. Backward Compatibility (1 test)**
- Verification that positive weights produce identical results to previous implementation
- Mathematical validation: `sqrt(0.6*s)*A + sqrt(0.4*s)*B` matches exactly

**5. Edge Cases (3 tests)**
- Very small negative weights (1e-6)
- Very large negative weights (100.0)
- Negative weights with different scaling factors across adapters

### Verification Strategy

**Mathematical Verification**:
- Manually computed expected values for test cases
- Verified matrix operations match theoretical formulas
- Confirmed sign is applied to both A and B matrices

**Backward Compatibility Verification**:
- For positive weights, verified behavior is identical to pre-fix implementation
- When weight ≥ 0, sign = 1, producing: `1 * sqrt(w*s) = sqrt(w*s)` (unchanged)

**Integration Testing**:
- Tested with simple models to ensure end-to-end functionality
- Verified adapter merging works across all supported combination types
- Confirmed no breaking changes to existing functionality

---

## Quick Demo

Here's a minimal example demonstrating the fix:

```python
import torch
from torch import nn

class SimpleModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(10, 5)
    def forward(self, x):
        return self.linear(x)

torch.manual_seed(42)
model = SimpleModel()
config = LoraConfig(r=8, lora_alpha=16, target_modules=['linear'], lora_dropout=0.0)
model = get_peft_model(model, config, adapter_name='adapter1')
model.add_adapter('adapter2', config)

# THIS WOULD HAVE FAILED BEFORE - NOW WORKS!
model.add_weighted_adapter(['adapter1', 'adapter2'], [0.7, -0.3], 'merged', combination_type='linear')
print('✅ SUCCESS: Negative weights work!')
print(f'Merged adapter created: {list(model.peft_config.keys())}')
```

**Before this PR**: The above code raises `ValueError: math domain error`
**After this PR**: Successfully creates merged adapter with negative weights

---

## Use Cases Enabled

This implementation enables several important use cases:

**1. Task Arithmetic**
```python
# Subtract unwanted task knowledge
model.add_weighted_adapter(
    adapters=["helpful_assistant", "sycophantic_behavior"],
    weights=[1.0, -0.5],  # Keep helpful, remove sycophantic
    adapter_name="improved_assistant",
    combination_type="linear"
)
```

**2. LoRA Hub Weight Optimization**
```python
# Use negative weights in optimization (as in LoRA Hub paper)
model.add_weighted_adapter(
    adapters=["task_a", "task_b", "task_c"],
    weights=[2.0, 0.3, -0.7],  # Optimized weights including negative
    adapter_name="multi_task_optimized",
    combination_type="ties",
    density=0.5
)
```

**3. Adapter Negation**
```python
# Create inverse of an adapter effect
model.add_weighted_adapter(
    adapters=["formal_style"],
    weights=[-1.0],  # Negate to get informal style
    adapter_name="informal_style",
    combination_type="linear"
)
```

---

## Test Results

All 21 tests pass successfully:

```
tests/test_negative_weights.py::TestNegativeWeightsLinear - 4 tests PASSED
tests/test_negative_weights.py::TestNegativeWeightsAllCombinations - 11 tests PASSED
tests/test_negative_weights.py::TestNegativeWeightsMathematicalCorrectness - 2 tests PASSED
tests/test_negative_weights.py::TestBackwardCompatibility - 1 test PASSED
tests/test_negative_weights.py::TestEdgeCases - 3 tests PASSED
```

---

### Test Suite Results
image

### Quick Demo Verification
image

---

## Breaking Changes

**None**. This PR is fully backward compatible:
- Positive weights behave identically to the previous implementation
- All existing functionality is preserved
- Only adds support for previously unsupported negative weight values

---

## Related Issues

Closes #2796

---

## Credits

- **Solution design**: @valteu (tested and validated against LoRA Hub)
- **Review and approval**: @BenjaminBossan
- **Implementation and testing**: Based on the discussion and approved solution in #2796 , by @sambhavnoobcoder 

---

## Checklist

- [x] Implementation follows the approved solution from issue discussion
- [x] Comprehensive test suite added (21 tests)
- [x] All tests passing
- [x] Backward compatibility verified
- [x] Documentation updated
- [x] Mathematical correctness validated
- [x] All combination types tested
- [x] Edge cases covered
- [x] No breaking changes

### DIFF
--- a/tests/testing_common.py
+++ b/tests/testing_common.py
@@ -1775,6 +1775,8 @@ class PeftCommonTester:
                     if "single" in adapter_name:
                         new_delta_weight = target.get_delta_weight(adapter_name)
                         weighted_original_delta_weights = target.get_delta_weight(adapter_list[0]) * weight_list[0]
+                        sign = 1 if weight_list[0] > 0 else -1
+                        weighted_original_delta_weights = sign * weighted_original_delta_weights
                         assert torch.allclose(new_delta_weight, weighted_original_delta_weights, atol=1e-4, rtol=1e-4)
                     elif "svd" in adapter_name:
                         assert target.r[adapter_name] == 20


src/peft/tuners/lora/model.py
```diff
@@ -540,7 +540,9 @@ def add_weighted_adapter(
             adapters (`list`):
                 List of adapter names to be merged.
             weights (`list`):
-                List of weights for each adapter.
+                List of weights for each adapter. Weights can be positive or negative, allowing for both addition
+                and subtraction of adapter effects. Negative weights enable task arithmetic and adapter negation as
+                described in the LoRA Hub and Task Arithmetic papers.
```
valteu: I would cut this, e.g. for `_svd_generalized_task_arithmetic_weighted_adapter` in [SVD-based merging](https://github.com/huggingface/peft/blob/ce5c2044f1430aa0bb3d2bccda73e242efde9e72/src/peft/tuners/lora/model.py#L775), this was already available and I don't think the hint to LoraHub or Task Arithemetic neccessary here

```suggestion
                List of weights for each adapter.
```

tests/test_negative_weights.py
```diff
@@ -0,0 +1,379 @@
...
+
+        model.add_weighted_adapter(
+            adapters=["adapter1"],
+            weights=[-100.0],
+            adapter_name="merged_large_negative",
+            combination_type="linear",
+        )
+
+        assert "merged_large_negative" in model.peft_config
+
```
valteu: If I'm not missing sth, I don't see a benefit in this. Nothing regarding weight magnitude was updated and we are not even testing for any mathematical correctness, just to get no error.

```suggestion

```

valteu (commented):
I took a look into the proposed changes and the logic update aligns with my solution. I'm however unsure about the effectiveness of some test cases and don't find the docstring update necessary (reasons provided in the comments added).
Also I think the PR comment should be updated, I don't see a value in the LoraHub code example, this has nothing todo with LoraHub imo.
Otherwise Lgtm!

tests/test_negative_weights.py
BenjaminBossan, line 1: Thanks for also including unit tests. However, we already have a couple of unit tests for `add_weighted_adapter`, and I would prefer to extend those instead of having this set of completely separate tests.

I think the best way to do that is to check this test function:

https://github.com/huggingface/peft/blob/b0954e0daa9b263449cef6d6b4b31f31e862e041/tests/testing_common.py#L1822-L1853

It is already parameterized over PEFT methods (right now, only LoRA and IA³), different model architectures, and merging methods. Right now, it is called with a list of weights `weight_list = [0.5, 1.5, 1.5]`. The easiest extension would be to define another set of weights which includes negative weights and then run the tests with those as well. So e.g.:

```diff
-            if isinstance(config, LoraConfig):
-                self._test_weighted_combination_of_adapters_lora(model, config, adapter_list, weight_list)
-            elif isinstance(config, IA3Config):
-                self._test_weighted_combination_of_adapters_ia3(model, config, adapter_list, weight_list)
+            if isinstance(config, LoraConfig):
+                self._test_weighted_combination_of_adapters_lora(model, config, adapter_list, weight_list)
+                self._test_weighted_combination_of_adapters_lora(model, config, adapter_list, negative_weight_list)
+            elif isinstance(config, IA3Config):
+                self._test_weighted_combination_of_adapters_ia3(model, config, adapter_list, weight_list)
+                self._test_weighted_combination_of_adapters_ia3(model, config, adapter_list, negative_weight_list)
```

Since those tests are generic in nature, I believe it would be useful to keep the `test_negative_weight_negates_adapter` test you added. Instead of putting it into a separate file, I'd rather add it to the custom tests after this test:

https://github.com/huggingface/peft/blob/b0954e0daa9b263449cef6d6b4b31f31e862e041/tests/test_custom_models.py#L2908

Then you can re-use one of the models defined there, e.g. `MLP`, instead of defining a new `SimpleModel`.

Similarly, I think we can transfer the `test_subtraction_with_negative_weights`, which looks useful to me. I wonder, however, if we can make that test even more strict. If we have two identical LoRA adapters and then merge them together with weight `[1.0, -1.0]`, should we not expect that they cancel out exactly? Then we can check that the new merged adapter has weights of approximately 0.

src/peft/tuners/lora/model.py
sambhavnoobcoder: okay , fixed in [c190977](https://github.com/huggingface/peft/pull/2811/commits/c1909774ff0b190501bfafe117d3a4597e0c39cb)

tests/test_negative_weights.py
sambhavnoobcoder: resolved in [c190977](https://github.com/huggingface/peft/pull/2811/commits/c1909774ff0b190501bfafe117d3a4597e0c39cb)

tests/test_negative_weights.py
sambhavnoobcoder: Got it — the goal was to extend the generic tests instead of creating new files. I’ve refactored accordingly, gone through the existing tests, and only added the cases that I felt weren’t already covered, directly within the generic test suite. This should now be resolved in [6a1c92d](https://github.com/huggingface/peft/pull/2811/commits/6a1c92da84709ba98c0092adf76c8cd96bb5118e) .
Please let me know if I’ve missed anything or if further changes are needed.

sambhavnoobcoder: just added another small test for testing negative weights with differential scaling in [bea394f](https://github.com/huggingface/peft/pull/2811/commits/bea394f421aabdbda279baed9ff5937fd5aaf952) . i think this covers all the edge cases for this now .

sambhavnoobcoder:
Thank you for your reviews @BenjaminBossan and @valteu . i have made the requested changes here . kindly look into this and let me know if this PR requires any more changes .

tests/test_custom_models.py
```diff
@@ -2800,6 +2800,120 @@ def test_add_weighted_adapter_cat_with_rank_pattern(self, config_cls):
             ["default", "other"], weights=[1.0, 1.0], adapter_name="merged", combination_type="cat"
         )
 
+    def test_negative_weight_negates_adapter(self):
```
BenjaminBossan: For clarity, let's extend the test name like this:

```suggestion
    def test_add_weighted_adapter_negative_weight_negates_adapter(self):
```

Same idea for the other tests.

tests/test_custom_models.py
```diff
@@ -2800,6 +2800,120 @@ def test_add_weighted_adapter_cat_with_rank_pattern(self, config_cls):
...
+
+    def test_subtraction_with_negative_weights(self):
+        # Test that merging two identical adapters with weights [1.0, -1.0] results in approximately zero weights
+        torch.manual_seed(42)
+        model = MLP()
+        config = LoraConfig(target_modules=["lin0"], init_lora_weights=False)
+        model = get_peft_model(model, config, adapter_name="adapter1")
+
+        # Create a second identical adapter by copying the first
+        model.add_adapter("adapter2", config)
```
BenjaminBossan, line 2850: I think what should work is if you set a seed before the `get_peft_model` call, then set the same seed before the `add_adapter`. Then the weights should be identical, no need to manually clone them.

tests/test_custom_models.py
```diff
@@ -2800,6 +2800,120 @@ def test_add_weighted_adapter_cat_with_rank_pattern(self, config_cls):
...
+        # Merge with negative weight - should handle different scalings correctly
+        model.add_weighted_adapter(
+            adapters=["adapter1", "adapter2"],
+            weights=[0.5, -0.3],
+            adapter_name="merged_diff_scaling",
+            combination_type="linear",
+        )
+
+        # Verify the merged adapter was created successfully
+        assert "merged_diff_scaling" in model.peft_config
```
BenjaminBossan: Can be removed

tests/test_custom_models.py
```diff
@@ -2800,6 +2800,120 @@ def test_add_weighted_adapter_cat_with_rank_pattern(self, config_cls):
...
+
+        # Verify the merged adapter can run forward pass
```
BenjaminBossan, line 2909: You need to call `model.set_adapter("merged_diff_scaling")`, otherwise we don't actually test it.

tests/testing_common.py
```diff
@@ -1831,6 +1831,7 @@ def _test_weighted_combination_of_adapters(self, model_id, config_cls, config_kw
 
         adapter_list = ["adapter1", "adapter_2", "adapter_3"]
         weight_list = [0.5, 1.5, 1.5]
+        negative_weight_list = [0.5, -0.8, 1.2]
```
BenjaminBossan:
```suggestion
        negative_weight_list = [-0.5, -0.8, -1.2]
```

Let's do all negative numbers, just for good measure :)

tests/test_custom_models.py
sambhavnoobcoder: resolved in [41409af](https://github.com/huggingface/peft/pull/2811/commits/41409afba935e5896f92dd5785b78f3518e60505)

tests/test_custom_models.py
sambhavnoobcoder: resolved in [41409af](https://github.com/huggingface/peft/pull/2811/commits/41409afba935e5896f92dd5785b78f3518e60505)

tests/test_custom_models.py
sambhavnoobcoder: resolved in [41409af](https://github.com/huggingface/peft/pull/2811/commits/41409afba935e5896f92dd5785b78f3518e60505)

tests/test_custom_models.py
sambhavnoobcoder: resolved in [41409af](https://github.com/huggingface/peft/pull/2811/commits/41409afba935e5896f92dd5785b78f3518e60505)

tests/testing_common.py
sambhavnoobcoder: resolved in [41409af](https://github.com/huggingface/peft/pull/2811/commits/41409afba935e5896f92dd5785b78f3518e60505)

sambhavnoobcoder:
@BenjaminBossan i went through the suggestions you made , implemented them in [41409af](https://github.com/huggingface/peft/pull/2811/commits/41409afba935e5896f92dd5785b78f3518e60505) . do tell me if any other changes are needed , i will implement those as well .

BenjaminBossan:
@sambhavnoobcoder Could you please run `make style` to satisfy the linter? Ignore the error about the doc build failure.

sambhavnoobcoder:
@BenjaminBossan ran the `make style` , style changes in 2 files , committed in [c9f3230](https://github.com/huggingface/peft/pull/2811/commits/c9f3230a10a9bea4954ae9df234ab60975c18bdc) . do check it out and tell me if any other changes are needed .

BenjaminBossan:
@sambhavnoobcoder We have a bunch of failing tests because the unit test does not consider that weights can be negative:

https://github.com/huggingface/peft/blob/b0954e0daa9b263449cef6d6b4b31f31e862e041/tests/testing_common.py#L1777-L1778

This should be easily fixed like so:

```diff
                         weighted_original_delta_weights = target.get_delta_weight(adapter_list[0]) * weight_list[0]
+                        sign = 1 if weight_list[0] > 0 else -1
+                        weighted_original_delta_weights = sign * weighted_original_delta_weights
```

Could you please update the test?

sambhavnoobcoder:
@BenjaminBossan added that in [8f9a625](https://github.com/huggingface/peft/pull/2811/commits/8f9a6256bc97e1a867fb0f7118ccf08abc0e5234) . can you check now ?

valteu:
Hi, thanks @sambhavnoobcoder and @BenjaminBossan for the great work!
Would it be possible to add the following line to the commit to credit my co-authorship?
```
Co-authored-by: Valentin Teutschbein <valentin.teutschbein@student.hpi.uni-potsdam.de>
```

BenjaminBossan:
Great, done @valteu. Greetings from Berlin.

valteu:
Thanks, greetings from Potsdam :)

sambhavnoobcoder:
Thank you both @BenjaminBossan  , @valteu  and greetings from Bengaluru as well :)

PR #2815, opened 2025-10-07
Added new initialization option for PromptEmbedding

macmacmacmac:
**TLDR:** Implemented a new initialization option for PromptEmbedding that fixes issue with original `RANDOM` initialization causing initialized embeddings to fall outside of the model's vocabulary manifold.

Currently, PromptEmbedding's random initialization uses `torch.nn.Embedding` to initialize its embeddings. However, because different models have different embedding spaces each with its own manifold of well-defined vocab embeddings, this naive initialization is highly unlikely to produce embeddings that land on the manifold, which leads to really poor learning empirically. From my testing naive random initialization reduces accuracy by almost a factor of 3.

To help visualize, here's a PCA of Llama 3.1 8b's vocab embeddings vs naively randomly initializing embeddings

Image

VS. if we initialized embeddings by say randomly sampling tokens instead (suggested fix/feature). 

Image

I implemented this as a new initialization option called `RANDOM_DISCRETE.`  The changes should be backward compatible. My first time contributing to a big project, I apologize if I neglected something.


src/peft/tuners/prompt_tuning/model.py
```diff
@@ -64,7 +64,20 @@ def __init__(self, config, word_embeddings):
...
+            vocab_size = word_embeddings.num_embeddings
+            init_token_ids = np.random.randint(0, vocab_size, total_virtual_tokens)
+            init_token_ids = torch.LongTensor(init_token_ids).to(word_embeddings.weight.device)
+            with gather_params_ctx(word_embeddings.parameters()):
+                word_embedding_weights = word_embeddings(init_token_ids).detach().clone()
+            word_embedding_weights = word_embedding_weights.to(torch.float32)
+            self.embedding.weight = torch.nn.Parameter(word_embedding_weights)
+
+        elif config.prompt_tuning_init == PromptTuningInit.TEXT and not config.inference_mode:
+            import numpy as np
```
BenjaminBossan: Can be removed
```suggestion
```

src/peft/tuners/prompt_tuning/model.py
```diff
@@ -64,7 +64,20 @@ def __init__(self, config, word_embeddings):
 
         total_virtual_tokens = config.num_virtual_tokens * config.num_transformer_submodules
         self.embedding = torch.nn.Embedding(total_virtual_tokens, config.token_dim)
-        if config.prompt_tuning_init == PromptTuningInit.TEXT and not config.inference_mode:
+        if config.prompt_tuning_init == PromptTuningInit.RANDOM_DISCRETE and not config.inference_mode:
+            import numpy as np
```
BenjaminBossan: How about using torch directly, instead of creating numpy arrays and later converting them to torch?

src/peft/tuners/prompt_tuning/config.py
```diff
@@ -31,7 +32,10 @@ class PromptTuningConfig(PromptLearningConfig):
     This is the configuration class to store the configuration of a [`PromptEmbedding`].
 
     Args:
-        prompt_tuning_init (Union[[`PromptTuningInit`], `str`]): The initialization of the prompt embedding.
+        prompt_tuning_init (Union[[`PromptTuningInit`], `str`]):
+            The initialization of the prompt embedding. `TEXT` will initialize with your text. `RANDOM_DISCRETE` will
```
BenjaminBossan: Thanks for extending the docstring. Since RANDOM_DISCRETE is not one of the options described in the paper, let's mention that. Otherwise, users might start searching the paper for that option.

BenjaminBossan (changes requested):
Thanks for coming up with, and implementing, this idea for better initialization of prompt tuning. I have a few small comments, please check.

Moreover, let's add a unit test for this new option. There are some existing tests here:

https://github.com/huggingface/peft/blob/b0954e0daa9b263449cef6d6b4b31f31e862e041/tests/test_decoder_models.py#L346-L381

Could you please create a new one below those? You can use `test_prompt_tuning_text_prepare_for_training` as a template.

Regarding the implementation itself, my understanding is that it will use random embedding vectors from the existing embedding matrix. This looks like a good idea for better initialization, but I'm curious if you have tested it in an experiment to compare the performance to random initialization?

Also, I wonder if, instead of completely copying a vector, we could copy random elements from the embedding size axis, so that the resulting vectors are unique. Maybe it makes no difference, just something that I was thinking about.

src/peft/tuners/prompt_tuning/config.py
macmacmacmac: Hi, I believe the paper does actually mention this initialization scheme. I italicized the part `RANDOM_DISCRETE` corresponds to. `RANDOM` corresponds to the first option. 

Though...yeah...`RANDOM_DISCRETE` maybe isn't the most intuitive name. I'm not familiar with naming schemes, so I would appreciate suggestions

> 2.1 Design Decisions
There are many possible ways to initialize the
prompt representations. The simplest is to train
from scratch, using random initialization. **_A more
sophisticated option is to initialize each prompt
token to an embedding drawn from the model’s
vocabulary._** Conceptually, our soft-prompt mod-
ulates the frozen network’s behavior in the same
way as text preceding the input, so it follows that
a word-like representation might serve as a good
initialization spot.

macmacmacmac:
@BenjaminBossan

> Moreover, let's add a unit test for this new option. There are some existing tests here:

I'll work on that! Thank you for the other feedback!

> Regarding the implementation itself, my understanding is that it will use random embedding vectors from the existing embedding matrix. This looks like a good idea for better initialization, but I'm curious if you have tested it in an experiment to compare the performance to random initialization?

I did actually test this out. I ran a couple trials with my task experimenting between random, discrete random, some sensible prompt, and some unrelated prompt using Llama as my base model on a regression task. I was getting as low as 0.3 Pearson R for random initialization while other methods (including random discrete) were consistently >0.9 Pearson R. 

From what I gathered, random initialization is producing vectors that fall outside of the part of the embedding space that is well-defined semantically to the model, so the model doesn't know how to interpret these embeddings, leading to worse performance. You can check this for yourself if you'd like by checking the average L2 norm of the random embeddings vs the model's vocab embeddings. For LLama, I found that it was around ~64x larger.

src/peft/tuners/prompt_tuning/config.py
macmacmacmac: The paper actually ran some experiments comparing the different initialization schemes--it seems like it performs consistently better than random initialization up to a certain model size (They called this 'sampled vocab' so I will change the name to reflect that). I would suggest making this the default initialization method, but I guess that might cause backward compatability issues.  

image

BenjaminBossan:
Thanks for your feedback. Indeed, your method corresponds to the "sampled vocab" option. I agree that it should be renamed to make this connection clear. Also thanks for reporting your results. Intuitively, I find the idea of taking random embedding vectors a bit strange, but if it works empirically, I won't argue with that ;)

PR #2824, opened 2025-10-09
Method comparison: Add prompt tuning experiment with sample vocab

BenjaminBossan:
A new initialization method was added to prompt tuning in #2815. This PR adds an experiment config for this method to the MetaMathQA benchmark.

Testing locally, this got a test accuracy of 36%, compared to 25% with random initialization.


BenjaminBossan:
ping @macmacmacmac

Issue #2830, opened 2025-10-13
Embedding scaling for TrainableTokensModel in XLoraEmbeddingLayer


sambhavnoobcoder:
### System Info

peft: 0.17.1
transformers: 4.56.2

### Who can help?

@benjaminbossian

This issue tracks the fix needed for XLoraEmbeddingLayer, corresponding to the changes in #2809 for LORA.
The parent issue addresses LORA updates, but a similar adjustment is required for XLoRA to ensure consistent behavior.

Creating this issue for tracking purposes — will push a PR for the same shortly.

### Reproduction

#2809 

### Expected behavior

Adapter should distinguish between vanilla torch.nn.Embedding layer and custom variants.

BenjaminBossan:
Resolved via #2831.

PR #2825, opened 2025-10-09
Handling embeddings scaling for TrainableTokensModel #2809

sambhavnoobcoder:
# Fix: Add support for scaled embeddings (Gemma3TextScaledWordEmbedding) in TrainableTokens and LoRA

Fixes #2809
 
## Original Issue
 
TrainableTokensModel and LoRA don't handle embedding scaling correctly for models like Gemma3 that use scaled embeddings.
 
**The Problem:**
- Gemma3 uses `Gemma3TextScaledWordEmbedding` which extends `nn.Embedding` and applies scaling (`embed_scale`) in its forward method
- TrainableTokensLayer bypasses the custom forward logic by calling `F.embedding()` directly, skipping the scaling step
- LoRA applies adapter modifications but doesn't scale them to match the base embeddings
- This results in magnitude mismatches, high gradient norms, and loss issues during training
 
---
 
### Root Cause Analysis
 
1. **TrainableTokens Issue:**
- TrainableTokensLayer uses `isinstance(layer, nn.Embedding)` check which passes for `Gemma3TextScaledWordEmbedding` (it's a subclass)
- However, it calls `F.embedding()` directly instead of `base_layer.forward()`, bypassing Gemma3's custom scaling logic
- Result: Embeddings are unscaled
 
2. **LoRA Issue:**
- LoRA calls `base_layer(x)` which correctly applies Gemma3's scaling to base embeddings
- However, LoRA adapter contributions `(A @ B) * scaling` are NOT multiplied by `embed_scale`
- Result: Base embeddings are scaled (~33x larger for Gemma3), but adapter deltas are not, creating magnitude mismatch
 
 
---
 
#### 1. TrainableTokensLayer (`src/peft/tuners/trainable_tokens/layer.py`)
 
Added helper method `_get_embed_scale()`:
- Detects `embed_scale` attribute on base layer
- Validates it's a scalar or 1-element tensor
- Logs warning if shape is invalid
- Returns `None` if not found or invalid
 
Modified `forward_adapters()` method:
- After calling `F.embedding()`, checks for `embed_scale`
- Applies scaling: `result = result * embed_scale.to(result.dtype)`
- Only applies when `embed_scale` is detected and valid
 
#### 2. LoRA Embedding (`src/peft/tuners/lora/layer.py`)
 
Added helper method `_get_embed_scale()`:
- Same logic as TrainableTokensLayer for consistency
 
Modified `forward()` method:
- Detects `embed_scale` once before processing adapters
[...]

PR #2831, opened 2025-10-13
XLoRA embed_scale Support #2830

sambhavnoobcoder:
## Summary

This PR adds embed_scale support to XLoraEmbeddingLayer, ensuring X-LoRA correctly handles models with scaled embeddings (e.g., Gemma3TextScaledWordEmbedding). This is a companion PR to the LoRA/TrainableTokens embed_scale fix, following the same approach but adapted for X-LoRA's architecture.

### Code
- **src/peft/tuners/xlora/layer.py**: Modified `XLoraEmbeddingLayer.forward()` to apply `embed_scale` to X-LoRA adapter contributions
  - Retrieves `embed_scale` via `self.target._get_embed_scale()` (inherits from BaseTunerLayer)
  - Applies scaling to adapter output before adding to result
  - Follows same pattern as LoRA implementation with comments explaining the approach

### Tests
- **tests/test_xlora.py**: Added `test_xlora_embed_scale_is_applied`
  - Uses real Gemma3 model with multiple LoRA adapters (X-LoRA's primary use case)
  - Varies embed_scale to extreme values (10000.0, 0) to verify scaling is applied correctly
  - Follows same test pattern as LoRA/TrainableTokens tests

## Key Differences from LoRA Implementation

**Implementation:**
- Uses `self.target._get_embed_scale()` instead of `self._get_embed_scale()` because X-LoRA wraps a LoRA Embedding layer (via `self.target`)
- No mixed batch forward variant needed (X-LoRA handles multiple adapters through its `scalings` parameter, not `adapter_names`)

**Testing:**
- Creates Gemma3-compatible LoRA adapters on-the-fly (X-LoRA requires saved LoRA adapters as input)
- Sets `use_cache=False` (X-LoRA requirement)
- Converts adapter list to dict format (X-LoRA config requirement)

## Test Results
- ✅ New test passes (2.69s)
- ✅ All 16 existing X-LoRA tests pass (no regression)
- ✅ `make style` passed

**Fixes :** #2830 
**cc:** @BenjaminBossan

### DIFF
--- a/tests/test_xlora.py
+++ b/tests/test_xlora.py
@@ -445,7 +445,7 @@ class TestXlora:
                 adapters[str(i)] = adapter_path
 
             # Load base model and test X-LoRA with embed_scale
-            base_model = AutoModelForCausalLM.from_pretrained(model_id)
+            base_model = AutoModelForCausalLM.from_pretrained(model_id).to(self.torch_device)
             base_model.config.use_cache = False
             orig_embedding = base_model.get_input_embeddings()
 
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Just FYI, I'll review the PR once #2825 is merged.

sambhavnoobcoder:
yes sure , i would prefer it that way . This PR will stay more aligned that way .

BenjaminBossan:
The other PR is merged. If you could please merge with/rebase on the current main and resolve the merge conflicts, then this PR should be ready for review.

sambhavnoobcoder:
@BenjaminBossan i've taken a pull form main , resolved all merge conflicts and made sure this is consistent with any of your comments from the last PR as well. looking forward to your review .

BenjaminBossan:
Hey @sambhavnoobcoder I did some tests with embed scales in DoRA, notably whether they should be applied before or after the weight norm is determined. To check that, I simply took the existing LoRA test, enabled DoRA, then checked if it would pass. Especially the merged vs unmerged check should reveal the correct behavior. What I found was that the scale should _not_ be applied during weight norm calculation. So the required change is simply:

```diff
modified   src/peft/tuners/lora/variants.py
@@ -326,6 +326,10 @@ class DoraEmbeddingVariant(DoraLinearVariant):
             base_layer=module.get_base_layer(),
             embed_fn=module._embed,
         )
+        # TODO comment to explain
+        embed_scale = module._get_embed_scale()
+        if embed_scale is not None:
+            dora_result = dora_result * embed_scale
         result = mag_norm_scale * result + dora_result
         return result
```

As for testing, it should be as easy as parametrizing the LoRA test you added with `use_dora = True/False`. Since you worked on this issue so far, do you want to complete the DoRA support too?

sambhavnoobcoder:
Thank you for your review on this PR  . I hadn't looked at DoRA so far , but i'll try this out , and raise a PR for the same . will also create an issue similar to xlora to keep track of the same .

Issue #2834, opened 2025-10-14
FSDP QLORA doesn't work with multiple adapters


ojh31:
### System Info

- `Accelerate` version: 1.10.1
- Platform: Linux-5.15.0-157-generic-x86_64-with-glibc2.39
- `accelerate` bash location: /opt/uv/venv/bin/accelerate
- Python version: 3.12.11
- Numpy version: 2.2.6
- PyTorch version: 2.9.0.dev20250825+cu128
- PyTorch accelerator: CUDA
- System RAM: 1003.13 GB
- GPU type: NVIDIA H100 80GB HBM3
- `Accelerate` default config:
        Not found
- peft version: 0.15.2
- transformers version: 4.56.1

### Who can help?

@benjaminbossan @githubnemo

### Reproduction

```
"""Based on peft/examples/sft/run_peft_qlora_fsdp.sh

Launch command:
{
    "name": "Accelerate Launch - Minimal FSDP QLoRA Training",
    "type": "debugpy",
    "request": "launch",
    "module": "accelerate.commands.launch",
    "args": [
        "--config_file",
        "scripts/fsdp_config_qlora.yaml",
        "--num_processes",
        "2",
        "scripts/20251008_fsdp_qlora_sft_custom.py",
        "--seed",
        "100",
        "--model_name_or_path",
        "meta-llama/Llama-3.1-8B-Instruct",
        "--dataset_name",
        "smangrul/ultrachat-10k-chatml",
        "--add_special_tokens",
        "False",
        "--append_concat_token",
        "False",
        "--splits",
        "train,test",
        "--max_seq_len",
        "2048",
        "--num_train_epochs",
        "1",
        "--logging_steps",
        "5",
        "--log_level",
        "info",
        "--logging_strategy",
        "steps",
        "--learning_rate",
        "1e-4",
        "--lr_scheduler_type",
        "cosine",
        "--weight_decay",
        "1e-4",
        "--warmup_ratio",
        "0.0",
        "--max_grad_norm",
        "1.0",
        "--output_dir",
        "llama-sft-qlora-fsdp",
        "--per_device_train_batch_size",
        "2",
        "--per_device_eval_batch_size",
        "2",
        "--gradient_accumulation_steps",
        "2",
        "--gradient_checkpointing",
        "True",
        "--lora_r",
        "8",
        "--lora_alpha",
        "16",
        "--lora_dropout",
        "0.1",
        "--lora_target_modules",
        "all-linear",
        "--max_steps",
[...]

PR #2836, opened 2025-10-14
bugfix: meta device handling in _move_adapter_to_device_of_base_layer

ojh31:
Fixes https://github.com/huggingface/peft/issues/2834


BenjaminBossan:
As [discussed on the issue](https://github.com/huggingface/peft/issues/2834#issuecomment-3406773822), this would break existing PEFT functionality. The recommended solution for now is to set `fsdp_cpu_ram_efficient_loading: false`.

Issue #2587, opened 2025-06-12
Support for ABBA: Highly Expressive Hadamard Product Adaptation for LLMs


RaghavSinghal10:
### Feature request

We just released ABBA, a new architecture for Parameter-Efficient Fine-Tuning (PEFT) that significantly outperforms LoRA and its major variants (e.g., HiRA, DoRA, LoRA-Pro), under the same parameter budget.

Unlike LoRA, which adds a low-rank delta to frozen weights, ABBA models the update as a Hadamard product of two independently learned low-rank matrices. This gives it higher expressivity and flexibility while remaining efficient.

ABBA consistently beats SoTA LoRA variants on commonsense and arithmetic reasoning across 4 open-source LLMs (Mistral-7B, Gemma-2 9B, LLaMA-3.2 1B/3B). In some cases, it even outperforms full fine-tuning.

Paper: https://arxiv.org/abs/2505.14238
Code: https://github.com/CERT-Lab/abba

Would love to get this integrated into peft. Happy to help with this as well!

### Motivation

ABBA is the new state-of-the-art in PEFT. It outperforms previous methods (LoRA, DoRA, HiRA. LoRA-Pro, etc) across a wide range of models and tasks. We have open-sourced the code and paper and provided links above.

### Your contribution

I am the lead author of ABBA. I would love to contribute in any way I can (for instance, submitting a PR)

BenjaminBossan:
Thanks for the offer to integrate ABBA into PEFT. Before starting, could you please explain how ABBA compares to LoHa (aka [FedPara](https://arxiv.org/abs/2108.06098))? I see that you reference that paper but the relation is not clear to me. Is ABBA a special case or generalization of LoHa or is it a completely different method that only looks similar on the surface?

monatis:
Hi @BenjaminBossan, I don't have any link to the authors of the paper, just experimenting with the idea of ABBA and trying to optimize it. AFAIK, FedPara applies a convolution operation over the original weights with the newly initialized learnable convolution kernel. On the other hand, ABBA initializes two separate pairs of LoRA weights with `lora_r = r/2` instead of a single pair of LoRA weights with `lora_r = r`. They calculate the Hadamard product of these two sets of LoRA weights before adding it to the original weights. So it seems to me that ABBA is beyond a specialization / generalization of FedPara and it's something different.

BenjaminBossan:
[...]

PR #2842, opened 2025-10-16
Integrate ABBA to LoHA

monatis:
This PR integrates ABBA to LoHA, based on the discussion [here](https://github.com/huggingface/peft/issues/2587#issuecomment-2995542253).

Raising it to track the progress more easily.


monatis:
@Hyperakan In addition to type hints and docstrings we talked earlier, could you also test it with the method comparison suite like in this [comment](https://github.com/huggingface/peft/issues/2587#issuecomment-2988544594).
Beside testing with the [current experiment for LoHA](https://github.com/altaidevorg/peft/tree/abba_feature/method_comparison/MetaMathQA/experiments/loha/) we can also add another for ABBA there.

src/peft/tuners/loha/config.py
```diff
@@ -69,7 +72,25 @@ class LoHaConfig(LycorisConfig):
...
+    )
+    r2: Optional[int] = field(
+        default=None,
+        metadata={
+            "help": (
+                "Rank for the second Hadamard component (w2a @ w2b). "
+                "If not specified, defaults to r/2 for ABBA-style initialization, or r otherwise."
+            )
+        },
+    )
```
BenjaminBossan: I see that these two parameters are described in the paper, but I think it's too much to have 3 parameters to set the rank. I would suggest either:

1. Drop r1 and r2, always use r1=r2=r.
2. If it's important to have r1!=r2 as an option, I would drop r1, use r and r2 only, where r takes the role of r1 and r2 defaults to r.

Unless I'm missing something, I wouldn't use r/2 as default, as I think it's confusing that changing from LoHa to ABBA halves the effective rank.

src/peft/tuners/loha/config.py
```diff
@@ -86,6 +107,18 @@ class LoHaConfig(LycorisConfig):
             )
         },
     )
+    use_khatri_rao: bool = field(
+        default=False,
```
BenjaminBossan: Note for the future: If we can observe that the numerical difference of using Khatri-Rao is negligible and it is strictly better memory-wise, I wouldn't be opposed to make it the default.

If we want to keep the default to be `False`, I would, however, change it to `"auto"` and then, when ABBA is used, `"auto" -> True` and for LoHa `"auto" -> False`. If the user sets `True` or `False`, it should be respected. As is, it's impossible for a user to use ABBA without `use_khatri_rao`.

src/peft/tuners/loha/layer.py
```diff
@@ -38,35 +38,43 @@ def __init__(self, base_layer: nn.Module):
         self.hada_w2_b = nn.ParameterDict({})
         self.hada_t1 = nn.ParameterDict({})
         self.hada_t2 = nn.ParameterDict({})
+        
+        # Khatri-Rao optimization flag
+        self.use_khatri_rao = {}
+        
+        # Store separate ranks for ABBA (r1 for first component, r2 for second)
+        self.r1 = {}
+        self.r2 = {}
```
BenjaminBossan, line 57: These new parameters have to be added to `other_param_names` (see line 28).

src/peft/tuners/loha/layer.py
```diff
@@ -98,38 +106,158 @@ def reset_adapter_parameters_random(self, adapter_name: str):
...
+                fourth_root_S1 = torch.pow(S_r1, 0.25)
+                self.hada_w1_a[adapter_name].data.copy_(U_r1 * fourth_root_S1)
+                self.hada_w1_b[adapter_name].data.copy_(fourth_root_S1.unsqueeze(1) * Vh_r1)
+                
+                # Initialize second component: w2a @ w2b
+                fourth_root_S2 = torch.pow(S_r2, 0.25)
+                self.hada_w2_a[adapter_name].data.copy_(U_r2 * fourth_root_S2)
+                self.hada_w2_b[adapter_name].data.copy_(fourth_root_S2.unsqueeze(1) * Vh_r2)
+                
+            except Exception as e:
```
BenjaminBossan: IMO, there is no need to catch the exception. As a user, if I explicitly indicated that I want ABBA and it fails, I would rather get an error than a warning.

src/peft/tuners/loha/layer.py
```diff
@@ -98,38 +106,158 @@ def reset_adapter_parameters_random(self, adapter_name: str):
...
+
         self.r[adapter_name] = r
+        self.r1[adapter_name] = r1
+        self.r2[adapter_name] = r2
         self.alpha[adapter_name] = alpha
-        self.scaling[adapter_name] = alpha / r
+        self.scaling[adapter_name] = alpha / r  # Original scaling (for backward compatibility)
+        
+        # ABBA paper: separate scaling factors α/√r₁ and α/√r₂
+        import math
```
BenjaminBossan: Move the import to root level.

src/peft/tuners/loha/layer.py
```diff
@@ -98,38 +106,158 @@ def reset_adapter_parameters_random(self, adapter_name: str):
...
+        self.scaling1 = getattr(self, 'scaling1', {})
+        self.scaling2 = getattr(self, 'scaling2', {})
```
BenjaminBossan: These attributes should be set during `__init__` and added to `other_param_names`.

BenjaminBossan (changes requested):
Thanks a lot for working on the ABBA method. This contribution already looks quite clean and well integrated, well done.

I found a couple of smaller issues, please check my comments. Also, as a next step, could you please:

- Add documentation (either extend LoHa or add a separate section for ABBA).
- Extend the tests: For a good start, add some ABBA settings [here](https://github.com/huggingface/peft/blob/87b90f045e84356cbd9bb839f3882b675f362711/tests/test_custom_models.py#L248-L286).

Pinging @RaghavSinghal10 if you have the opportunity, it would be great if you could review, with the focus on the algorithm itself :pray:.

Hyperakan:
I’ve implemented the requested changes. All tests are passing now, and the test coverage can be extended further.

BenjaminBossan (changes requested):
Thanks for the updates, the PR looks quite good. Could you please run `make style` to satisfy ruff?

Also, let's add the documentation for ABBA. One suggestion is that @RaghavSinghal10 writes the docs and I would later add him as co-author, if you're both fine with that suggestion.

> All tests are passing now, and the test coverage can be extended further.

Let's extend the `test_stablediffusion.py` and `test_vision_models.py`. They already have a LoHa config, so just copying it and enabling ABBA should be enough.

Hyperakan:
> This works for me - I can go ahead and write the documentation if it works for @Hyperakan

That works for me as well, thanks!

Hyperakan:
Now only the following test is failing:
FAILED tests/test_stablediffusion.py::TestStableDiffusionModel::test_merge_layers_safe_merge[LoHaConfig-config_kwargs2-hf-internal-testing/tiny-sd-pipe] - AssertionError: False, 1.0

I couldn’t identify the root cause of the mismatch. It fails for the regular LoHa configuration (i.e. init_weights = False), but it doesn’t fail in the original codebase.

src/peft/tuners/loha/layer.py
```diff
@@ -365,6 +577,113 @@ def __repr__(self) -> str:
...
+    Mathematical equivalence: result[i,j] = scale1 * (sum_k w1a[i,k]*w1b[k,j]) * scale2 * (sum_k w2a[i,k]*w2b[k,j])
+
+    This can be computed without materializing full matrices by processing one row at a time or using einsum with no
+    intermediate storage.
+
+    Memory savings: O(m*n) -> O(n) for forward pass (processing row by row)
+    """
+
+    @staticmethod
+    def forward(ctx, w1a, w1b, w2a, w2b, scale1=torch.tensor(1), scale2=torch.tensor(1)):
```
BenjaminBossan:
```suggestion
    def forward(ctx, w1a, w1b, w2a, w2b, scale1, scale2):
```

Not necessary, right?

tests/test_stablediffusion.py
```diff
@@ -288,11 +323,21 @@ def test_merge_layers(self, model_id, config_cls, config_kwargs):
             merged_output = np.array(model(**dummy_input).images[0]).astype(np.float32)
 
         # Images are in uint8 drange, so use large atol
-        assert np.allclose(peft_output, merged_output, atol=1.0)
+        # Increase tolerance for LoHa when init_weights is set to "abba"
+        if config_cls == LoHaConfig and original_config_kwargs.get("text_encoder", {}).get("init_weights") == "abba":
+            atol = 15.0
```
BenjaminBossan: This seems a bit concerning, any idea why ABBA is much more susceptible here?

BenjaminBossan (changes requested):
Thanks for the updates @Hyperakan. When I run the tests locally with GPU, some ABBA tests with `Conv1D` layers are failing: `pytest tests/test_custom_models.py -k "abba and conv1d"`:

```
FAILED tests/test_custom_models.py::TestPeftCustomModel::test_merge_layers[Conv1d ABBA 3-Conv1dBigger-LoHaConfig-config_kwargs84] - AssertionError
FAILED tests/test_custom_models.py::TestPeftCustomModel::test_safe_merge[Conv1d ABBA 3-Conv1dBigger-LoHaConfig-config_kwargs84] - AssertionError
FAILED tests/test_custom_models.py::TestPeftCustomModel::test_disable_adapters_with_merging[Conv1d ABBA 3-Conv1dBigger-LoHaConfig-config_kwargs84] - AssertionError: assert False
```

This just appears to be a matter of tolerances, so e.g. loosening from 1e-4 to 1e-3 for ABBA should be enough. Could you please check? On CPU, they pass for me.

> Now only the following test is failing: FAILED tests/test_stablediffusion.py::TestStableDiffusionModel::test_merge_layers_safe_merge[LoHaConfig-config_kwargs2-hf-internal-testing/tiny-sd-pipe] - AssertionError: False, 1.0
> 
> I couldn’t identify the root cause of the mismatch. It fails for the regular LoHa configuration (i.e. init_weights = False), but it doesn’t fail in the original codebase.

This was quite a rabbit hole. Indeed, the addition of the ABBA tests results in another LoHa test to fail. That test passes when run in isolation. Therefore, the ABBA test must have some kind of side-effect that makes the LoHa test fail. After some investigation, it's actually caused by this line in `HadaWeightKR.forward`:

`diff_weight = torch.empty(m, n, dtype=w1a.dtype, device=w1a.device)`

For some reason unclear to me, if we replace this with `torch.zeros`, the test passes. At first I thought that the loop could be wrong and some empty data is being returned, which *somehow* stays allocated and thus affects the subsequent tests (LoHa uses `torch.empty` for weight initialization) but AFAICT, there is no issue with the loop. Also, calling `torch.cuda.empty_cache()` between tests also doesn't solve the issue. Therefore, I have no idea what the underlying cause is. Probably it's somewhere else in the LoHa code (or even a PyTorch bug?) but I can't see anything.

For now, I would probably change the `torch.empty` call to `torch.zeros` to be safe, even though that's of course more expensive. LMK if you have any ideas. **Update**: Another idea: collect the chunks in a list and concat it at the end. **Update2**: I tried the idea of collecting and concatenating the chunks, thus avoiding `torch.empty` but the old test still fails, what the heck?? It's got to be some kind of CUDA side effect, as running this on CPU works.

Apart from that, just two small comments.

BenjaminBossan:
@Hyperakan I found a better solution to the failing stable diffusion test:

```python
# test_stablediffusion.py
import torch

[...]

class TestStableDiffusionModel(PeftCommonTester):
    [...]

    @pytest.fixture(autouse=True)
    def autofixture(self):
        # Disable TF32 in cudnn
        prev_value = torch.backends.cudnn.allow_tf32
        try:
            torch.backends.cudnn.allow_tf32 = False
            yield
        finally:
            torch.backends.cudnn.allow_tf32 = prev_value
        torch.cuda.empty_cache()
```

For some reason, `allow_tf32=True` causes the issue, by setting it to `False`, all tests pass. No need to change the `torch.empty` call. I also added `torch.cuda.empty_cache()` for good measure but it's not required.

BenjaminBossan:
> where all do I need to update the documentation?

> I was planning to fill in this page (https://github.com/huggingface/peft/blob/main/docs/source/developer_guides/lora.md)

This wouldn't be a good fit, as it only talks about LoRA.

> and add an abba.md file under https://github.com/huggingface/peft/tree/main/docs/source/package_reference.

Yes, please put a file for ABBA there, check the other files to see what content is expected. Also mention that it's implemented in terms of LoHa. Don't forget to add an entry in the `_toctree.yml`. You could also consider adding an example, for instance by copying an existing one in `examples/` and adjusting it.

Hyperakan:
> @Hyperakan I found a better solution to the failing stable diffusion test:
> 
> ```python
> # test_stablediffusion.py
> import torch
> 
> [...]
> 
> class TestStableDiffusionModel(PeftCommonTester):
>     [...]
> 
>     @pytest.fixture(autouse=True)
>     def autofixture(self):
>         # Disable TF32 in cudnn
>         prev_value = torch.backends.cudnn.allow_tf32
>         try:
>             torch.backends.cudnn.allow_tf32 = False
>             yield
>         finally:
>             torch.backends.cudnn.allow_tf32 = prev_value
>         torch.cuda.empty_cache()
> ```
> 
> For some reason, `allow_tf32=True` causes the issue, by setting it to `False`, all tests pass. No need to change the `torch.empty` call. I also added `torch.cuda.empty_cache()` for good measure but it's not required.

After implementing your suggested solution, I didn’t need to set a special atol value for abba. It's passing all the tests now.

BenjaminBossan:
> After implementing your suggested solution, I didn’t need to set a special atol value for abba. It's passing all the tests now.

Great, all the better.

It seems like there are some merge conflicts with the latest main branch, could you please merge it @Hyperakan?

Hyperakan:
I've resolved the merge conflicts, but I'm still encountering errors in test_custom_models.py , even when running on the upstream/main branch. I'm not sure why this is happening.

BenjaminBossan:
> but I'm still encountering errors in test_custom_models.py , even when running on the upstream/main branch. I'm not sure why this is happening.

To me, it looks like the precision of ABBA with `Conv1D` is not so high, resulting in the corresponding tests failing. If you check these tests, you can see that we already lower the required tolerance for some PEFT methods, you could add a similar logic for ABBA.

Hyperakan:
> > but I'm still encountering errors in test_custom_models.py , even when running on the upstream/main branch. I'm not sure why this is happening.
> 
> To me, it looks like the precision of ABBA with `Conv1D` is not so high, resulting in the corresponding tests failing. If you check these tests, you can see that we already lower the required tolerance for some PEFT methods, you could add a similar logic for ABBA.

If I’m not doing something wrong, the errors don’t seem to be limited to ABBA-related ones. When I run the tests on upstream/main, I get similar failures too. In my branch, the only additional ones appear to come from the LoHa tests that use ABBA initialization (as shown in the screenshot). The rest of the failures are already present in upstream/main.
Screenshot 2025-10-27 at 15 07 52

BenjaminBossan:
@Hyperakan Those failures could be GPU related, please check if the tests pass on CPU. For this PR, let's focus on fixing any ABBA-related failures for now.

Hyperakan:
I needed to add an if loss.requires_grad check around loss.backward() and optimizer.step() because module_dropout=0.1 in the Vanilla MLP 3 ABBA setup can produce adapter parameters that don't contribute to the computational graph.
With this change, the tests now pass on CPU.

BenjaminBossan:
> I needed to add an if loss.requires_grad check around loss.backward() and optimizer.step() because module_dropout=0.1 in the Vanilla MLP 3 ABBA setup can produce adapter parameters that don't contribute to the computational graph.
> With this change, the tests now pass on CPU.

Ah, nice find, could you please add a comment on each instance on why the check needs to be performed?

@RaghavSinghal10 just a friendly ping if you already had time to work on the docs.

BenjaminBossan:
Gentle ping @RaghavSinghal10

BenjaminBossan:
Another ping @RaghavSinghal10

githubnemo:
Ping @RaghavSinghal10 :)

BenjaminBossan:
It's been a very long time since we last heard from @RaghavSinghal10 and it would be a pity to let this PR go to waste. @monatis @Hyperakan would one of you be up to providing the missing docs?

monatis:
Hi @BenjaminBossan, thanks for the ping. Agree that this PR should not be wasted. We talked offline and @Hyperakan will update this pR with the missing docs and other changes required for merging because it's a long time and the Peft library naturally evolved.

BenjaminBossan:
Hey, I wanted to ask if we can revive this PR, as not much was missing. Is anyone of the participants willing to pick this up again?

BenjaminBossan:
I revived the idea of adding Khatri-Rao to LoHa in #3560. It's not full on ABBA, but still contains a lot of changes in the same spirit as this PR.

The new PR does not directly build on this PR, but since there was still a lot of ground work laid out here, I would like @monatis and @Hyperakan to be added as co-authors of that PR should be merged. If you two want to be added, please LMK how I should add you. Also, feel free to give feedback on said PR.

monatis:
Hi @BenjaminBossan much of the work in this PR belongs to @Hyperakan so you may kindly consider adding him as a co-author. I reviewed #3560 and LGTM. Thank you

PR #2846, opened 2025-10-16
TST: Clean up testing

BenjaminBossan:
There is still some chaos in our test suite, despite recent efforts to refactor it. This PR tries to improve the situation a bit.

Some of the changes:

- for some PEFT methods, adjust some tolerances where necessary instead of skipping the whole test
- don't add return to `pytest.skip`, it's not necessary
- avoid `self.skipTest`, always use `pytest.skip`
- unused `CONFIG_TESTING_KWARGS` is removed
- factor out common skipping logic into dedicated functions
- many tests had code like: skip unless it's LoRA or IA³ or ... often, new PEFT methdos would actually work but were not added to this list, resulting in unnecessary skips
- make it so that LN Tuning is allowed to run with 0 active adapters (ping @DTennant)
- fix a bug in BOFT resulting in `cayley_batch` always returning float32 dtype (ping @zqiu24)
- transpose `delta_weight` for WaveFT merging if base layer is Conv1D (ping @Bilican)
- fix `UnboundLocalError` in C3A (ping @Phoveran)

For those that I pinged, this is just to let you know I made some fixes on PEFT methods that may concern you and ask you, if you have time, to double check them.

Errors not fixed in this PR:
- Layer indexing test is flaky with VBLoRA and VeRA, even with looser tolerances. Suspiciously, both methods have in common that they share weights across layers and layer indexing skips some layers. There is probably something there, but it's not immediately obvious to me. I checked that this test would already fail on main if not skipped, so I think it's fine to keep skipping it for these PEFT methods.

### DIFF
--- a/tests/testing_common.py
+++ b/tests/testing_common.py
@@ -1024,6 +1024,9 @@ class PeftCommonTester:
         if (config_cls == AdaLoraConfig) and ("roberta" in model_id.lower()):
             # TODO: no gradients on the "dense" layer, other layers work, not sure why
             pytest.skip("AdaLora with RoBERTa does not work correctly")
+        if (config_cls == PrefixTuningConfig) and ("bert" in model_id.lower()):
+            # TODO: prefix tuning fails with roberta, deberta, etc., requires investigation
+            pytest.skip("Prefix tuning fails with Bert* type models.")
 
         with hub_online_once(model_id):
             model = self.transformers_class.from_pretrained(model_id)
@@ -1041,7 +1044,7 @@ class PeftCommonTester:
             loss = output.sum()
             loss.backward()
 
-            if not issubclass(config_cls, PromptLearningConfig):
+            if issubclass(config_cls, PromptLearningConfig):
                 # we cannot reliably identify the trainable part of the prompt learning method, thus skipping this check
                 return
 


BenjaminBossan:
@Phoveran I also noticed that we missed a TODO:

https://github.com/huggingface/peft/blob/2813b9c4bfd240408de398028da722ed52528cdb/src/peft/tuners/c3a/model.py#L33

Phoveran:
Hi,

Thanks for the notice! I have added the paper link in my fork.
However, for the error, I tried to replicate it by `python -m pytest tests/ -k c3a` but failed. The test cases are all passed or skipped. Could you show me more details on it?

BenjaminBossan:
> I have added the paper link in my fork.

Nice, could you please create a PR?

> However, for the error, I tried to replicate it by `python -m pytest tests/ -k c3a` but failed. The test cases are all passed or skipped. Could you show me more details on it?

You would have to use my branch from this PR. It removes a lot of skipping logic from the tests that I found unnecessary. But there is no need for you to get active, I already fixed the issue in C3A, my ping is just to let you know and for you to [double check my change](https://github.com/huggingface/peft/pull/2846/files#diff-e8f36d9a5bb7c62fe17cbf3da36403dcf33a8a2cb5cb7ddbbc08df0c4ac8fc26).

Phoveran:
Got you. The [PR](https://github.com/huggingface/peft/pull/2852) is ready. Thanks!

BenjaminBossan:
Note on some of the failing tests:

- OFT unmerging: see https://github.com/huggingface/peft/pull/2805#issuecomment-3421595323
- DeLoRA adapter deletion and safe merge error message: #2853

tests/testing_common.py
```diff
@@ -1211,10 +1022,10 @@ def _test_prefix_tuning_half_prec_conversion(self, model_id, config_cls, config_
 
     def _test_training(self, model_id, config_cls, config_kwargs):
         if issubclass(config_cls, PromptLearningConfig):
-            return pytest.skip(f"Test not applicable for {config_cls}")
+            pytest.skip("Prompt learning does not support merging, skipping this test.")
```
githubnemo: This message seems unrelated to the name `_test_training`?

tests/testing_common.py
```diff
@@ -1350,29 +1169,22 @@ def _test_training_gradient_checkpointing(self, model_id, config_cls, config_kwa
...
-            )
-
-        if (config_cls == OFTConfig) and ("deberta" in model_id.lower()):
-            # TODO: no gradients on the "dense" layer, other layers work, not sure why
-            self.skipTest("OFT with Deberta does not work correctly")
+            pytest.skip("AdaLora with RoBERTa does not work correctly")
 
         if "gptbigcode" in model_id.lower():
-            self.skipTest("GPTBigCode currently doesn't implement gradient checkpointing correctly.")
+            pytest.skip("GPTBigCode currently doesn't implement gradient checkpointing correctly.")
```
githubnemo: This check can be removed, gptbigcode supports gradient checkpointing since https://github.com/huggingface/transformers/pull/41818 and is present in 5.0rc0

tests/testing_common.py
```diff
@@ -1350,29 +1169,22 @@ def _test_training_gradient_checkpointing(self, model_id, config_cls, config_kwa
...
+        if "bart" in model_id.lower() and version.parse(transformers.__version__) <= version.parse("5.0"):
+            # TODO: remove once torch < 5.0 no longer supported
```
githubnemo:
```suggestion
            # TODO: remove once transformers < 5.0 no longer supported
```

Issue #2636, opened 2025-07-09
Support for GraLoRA


yeonjoon-jung01:
### Feature request

This request proposes integrating [GraLoRA](https://arxiv.org/pdf/2505.20355) (Granular Low-Rank Adaptation for Parameter-Efficient Fine-Tuning), an effective PEFT method that overcomes LoRA’s structural bottleneck. Traditional LoRA introduces gradient entanglement to unrelated input channels, which distorts gradient propagation.

GraLoRA addresses this limitation by dividing the adaptation space into k^2  independently trained low-rank adapters, enabling spatially localized and context-aware updates with negligible computational and storage overhead.

### Your contribution

We implement GraLoRA in https://github.com/SqueezeBits/GraLoRA

yeonjoon-jung01:
We've received requests for GraLoRA support both publicly (https://github.com/SqueezeBits/GraLoRA/issues/2, https://github.com/huggingface/peft/issues/2558) and privately. As a result, we're planning to submit a PR to integrate GraLoRA.

Feedback is welcome! :)

BenjaminBossan:
Thanks for working on this. Feel free to create an early draft PR so that we can give quick feedback and iterate.

yeonjoon-jung01:
Sorry for the delay! After being selected as a NeurIPS 2025 Spotlight, we’ve opened the initial PR https://github.com/huggingface/peft/pull/2851. Can you please reopen this issue?

PR #2851, opened 2025-10-20
FEAT add GraLoRA

yeonjoon-jung01:
### Summary
We opened the initial PR for the GraLoRA method (a granular low-rank adaptation that improves expression power and outlier handling, selected as a NeurIPS 2025 Spotlight), based on https://github.com/huggingface/peft/issues/2636

### DIFF
--- a/examples/gralora_finetuning/gralora_finetuning.py
+++ b/examples/gralora_finetuning/gralora_finetuning.py
@@ -52,7 +52,7 @@ def train_model(
     # GraLoRA config for the PEFT model
     gralora_config = GraloraConfig(
         r=gralora_r,  # Rank of matrix
-        gralora_alpha=gralora_alpha,
+        alpha=gralora_alpha,
         target_modules=(
             gralora_target_modules.split(",")
             if gralora_target_modules
--- a/src/peft/tuners/gralora/config.py
+++ b/src/peft/tuners/gralora/config.py
@@ -39,8 +39,8 @@ class GraloraConfig(PeftConfig):
             specified, modules will be chosen according to the model architecture, If the architecture is " not known,
             an error will be raised -- in this case, you should specify the target modules manually. " To avoid
             targeting any modules (because you want to apply `target_parameters`), set " `target_modules=[]`.
-        gralora_alpha (`int`): GraLoRA alpha.
-            GraLoRA alpha is the scaling factor for the GraLoRA adapter. Scale becomes gralora_alpha / (r + hybrid_r).
+        alpha (`int`): GraLoRA alpha.
+            GraLoRA alpha is the scaling factor for the GraLoRA adapter. Scale becomes alpha / (r + hybrid_r).
         gralora_dropout (`float`):
             GraLoRA dropout is the dropout probability for the GraLoRA adapter. It is used to prevent overfitting and
             improve the generalization of the GraLoRA adapter.
@@ -104,12 +104,11 @@ class GraloraConfig(PeftConfig):
             )
         },
     )
-    gralora_alpha: int = field(
+    alpha: int = field(
         default=64,
         metadata={
             "help": (
-                "gralora alpha is the scaling factor for the GraLoRA adapter. "
-                "Scale becomes gralora_alpha / (r + hybrid_r). "
+                "gralora alpha is the scaling factor for the GraLoRA adapter. Scale becomes alpha / (r + hybrid_r). "
             )
         },
     )
--- a/src/peft/tuners/gralora/layer.py
+++ b/src/peft/tuners/gralora/layer.py
@@ -27,12 +27,12 @@ from peft.utils.other import transpose
 class GraloraLayer(BaseTunerLayer):
     # List all names of layers that may contain adapter weight
     adapter_layer_names = ("gralora_A", "gralora_B", "gralora_A_general", "gralora_B_general")
-    other_param_names = ("r", "hybrid_r", "gralora_alpha", "scaling", "gralora_dropout")
+    other_param_names = ("r", "hybrid_r", "alpha", "scaling", "gralora_dropout")
 
     def __init__(self, base_layer: nn.Module, **kwargs):
         self.base_layer = base_layer
         self.r = {}
-        self.gralora_alpha = {}
+        self.alpha = {}
         self.gralora_k = {}
         self.hybrid_r = {}
         self.scaling = {}
@@ -66,7 +66,7 @@ class GraloraLayer(BaseTunerLayer):
         adapter_name,
         module_name,
         r,
-        gralora_alpha,
+        alpha,
         gralora_dropout,
         gralora_k: int = 2,
         hybrid_r: int = 0,
@@ -78,7 +78,7 @@ class GraloraLayer(BaseTunerLayer):
             raise ValueError(f"`hybrid_r` should be a non-negative integer value but the value passed is {hybrid_r}")
 
         self.r[adapter_name] = r
-        self.gralora_alpha[adapter_name] = gralora_alpha
+        self.alpha[adapter_name] = alpha
         self.gralora_k[adapter_name] = gralora_k
         self.hybrid_r[adapter_name] = hybrid_r
 
@@ -148,7 +148,7 @@ class GraloraLayer(BaseTunerLayer):
 
         self.module_name = module_name
 
-        self.scaling[adapter_name] = gralora_alpha / (gralora_r + hybrid_r)
+        self.scaling[adapter_name] = alpha / (gralora_r + hybrid_r)
         self._move_adapter_to_device_of_base_layer(adapter_name)
         self.set_adapter(self.active_adapters)
 
@@ -161,7 +161,7 @@ class Linear(nn.Linear, GraloraLayer):
         adapter_name: str,
         module_name,
         r: int = 0,
-        gralora_alpha: int = 1,
+        alpha: int = 1,
         gralora_dropout: float = 0.0,
         gralora_k: int = 2,
         hybrid_r: int = 0,
@@ -175,9 +175,7 @@ class Linear(nn.Linear, GraloraLayer):
         self.fan_in_fan_out = fan_in_fan_out
 
         self._active_adapter = adapter_name
-        self.update_layer(
-            adapter_name, module_name, r, gralora_alpha, gralora_dropout, gralora_k, hybrid_r, init_weights
-        )
+        self.update_layer(adapter_name, module_name, r, alpha, gralora_dropout, gralora_k, hybrid_r, init_weights)
 
     def merge(self, safe_merge: bool = False, adapter_names: Optional[list[str]] = None) -> None:
         """
--- a/src/peft/tuners/gralora/model.py
+++ b/src/peft/tuners/gralora/model.py
@@ -79,7 +79,7 @@ class GraloraModel(BaseTuner):
         bias = hasattr(target, "bias") and target.bias is not None
         kwargs = {
             "r": r,
-            "gralora_alpha": gralora_config.gralora_alpha,
+            "alpha": gralora_config.alpha,
             "gralora_dropout": gralora_config.gralora_dropout,
             "gralora_k": gralora_config.gralora_k,
             "fan_in_fan_out": gralora_config.fan_in_fan_out,
@@ -93,7 +93,7 @@ class GraloraModel(BaseTuner):
                 adapter_name,
                 current_key,
                 r,
-                gralora_config.gralora_alpha,
+                gralora_config.alpha,
                 gralora_config.gralora_dropout,
                 gralora_config.gralora_k,
                 gralora_config.hybrid_r,
--- a/method_comparison/MetaMathQA/experiments/gralora/llama-3.2-3B-rank32/adapter_config.json
+++ b/method_comparison/MetaMathQA/experiments/gralora/llama-3.2-3B-rank32/adapter_config.json
@@ -3,7 +3,7 @@
   "base_model_name_or_path": null,
   "bias": "none",
   "fan_in_fan_out": false,
-  "gralora_alpha": 64,
+  "alpha": 64,
   "gralora_dropout": 0.0,
   "gralora_k": 2,
   "hybrid_r": 0,
--- a/tests/test_decoder_models.py
+++ b/tests/test_decoder_models.py
@@ -143,7 +143,7 @@ ALL_CONFIGS = [
         {
             "task_type": "CAUSAL_LM",
             "r": 8,
-            "gralora_alpha": 16,
+            "alpha": 16,
             "target_modules": None,
             "gralora_dropout": 0.05,
             "gralora_k": 2,
@@ -155,7 +155,7 @@ ALL_CONFIGS = [
         {
             "task_type": "CAUSAL_LM",
             "r": 16,
-            "gralora_alpha": 32,
+            "alpha": 32,
             "target_modules": None,
             "gralora_dropout": 0.05,
             "gralora_k": 4,


src/peft/tuners/gralora/config.py
```diff
@@ -0,0 +1,92 @@
...
+from dataclasses import dataclass, field
+from typing import Optional, Union
+
+from peft.config import PeftConfig
+from peft.utils import PeftType
+
+
+@dataclass
+class GraloraConfig(PeftConfig):
+    r: int = field(default=8, metadata={"help": "gralora attention dimension"})
```
BenjaminBossan: The description is not helpful IMO, please give a more complete description.

src/peft/tuners/gralora/config.py
```diff
@@ -0,0 +1,92 @@
...
+        default=None,
+        metadata={
+            "help": (
+                "List of module names or regex expression of the module names to replace with gralora."
+                "For example, ['q', 'v'] or '.*decoder.*(SelfAttention|EncDecAttention).*(q|v)$'. "
+                "Only linear layers are supported."
+            )
+        },
+    )
+    gralora_alpha: int = field(default=8, metadata={"help": "gralora alpha"})
```
BenjaminBossan: Same, let's extend the description

src/peft/tuners/gralora/config.py
```diff
@@ -0,0 +1,92 @@
...
+    gralora_dropout: float = field(default=0.0, metadata={"help": "gralora dropout"})
+    gralora_k: int = field(default=2, metadata={"help": "gralora k"})
```
BenjaminBossan: Same, let's extend the description

src/peft/tuners/gralora/config.py
```diff
@@ -0,0 +1,92 @@
...
+
+from peft.config import PeftConfig
+from peft.utils import PeftType
+
+
+@dataclass
+class GraloraConfig(PeftConfig):
+    r: int = field(default=8, metadata={"help": "gralora attention dimension"})
+    hybrid_r: int = field(
+        default=0, metadata={"help": "hybrid_r is the rank allocated to vanilla LoRA method when using Hybrid GraLoRA"}
```
BenjaminBossan: IIUC, by passing a value > 0, hybrid GraLoRA is enabled. Let's make that clear.

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,427 @@
...
+    def __init__(self, base_layer: nn.Module, **kwargs):
+        self.base_layer = base_layer
+        self.r = {}
+        self.gralora_alpha = {}
+        self.gralora_k = {}
+        self.hybrid_r = {}
+        self.scaling = {}
+        self.gralora_dropout = nn.ModuleDict({})
+
+        # Set to `None` otherwise to avoid computation with random weight
```
BenjaminBossan: Remove

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,427 @@
...
+        self._disable_adapters = False
+        self.merged_adapters = []
+
+        base_layer = self.get_base_layer()
+        if isinstance(base_layer, nn.Linear):
+            in_features, out_features = base_layer.in_features, base_layer.out_features
+        elif isinstance(base_layer, Conv1D):
+            in_features, out_features = (
+                base_layer.weight.ds_shape if hasattr(base_layer.weight, "ds_shape") else base_layer.weight.shape
+            )
```
BenjaminBossan, line 56: Let's raise an error in the `else` case.

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,427 @@
...
+        for adapter_layer_name in self.adapter_layer_names + self.other_param_names:
+            adapter_layer = getattr(self, adapter_layer_name, None)
+            if not isinstance(adapter_layer, (nn.ModuleDict, nn.ParameterDict, BufferDict)):
+                continue
+            if adapter_name not in adapter_layer:
+                continue
+            if weight.dtype.is_floating_point or weight.dtype.is_complex:
+                adapter_layer[adapter_name] = adapter_layer[adapter_name].to(device, dtype=dtype)
+            else:
+                adapter_layer[adapter_name] = adapter_layer[adapter_name].to(device)
```
BenjaminBossan: AFAICT, this is the same as on the parent class, so it can be removed.

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,427 @@
...
+
+    @property
+    def bias(self) -> torch.Tensor:
+        base_layer = self.get_base_layer()
+        if isinstance(base_layer, nn.Linear):
+            return base_layer.bias
+        elif isinstance(base_layer, Conv1D):
+            return base_layer.bias
+        else:
+            return None
```
BenjaminBossan: Also same as the parent class and thus can be removed.

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,427 @@
...
+        module_name,
+        r,
+        gralora_alpha,
+        gralora_dropout,
+        gralora_k: int = 2,
+        hybrid_r: int = 0,
+        init_weights: bool = True,
+    ):
+        if r <= 0:
+            raise ValueError(f"`r` should be a positive integer value but the value passed is {r}")
```
BenjaminBossan, line 76: Let's also check `hybrid_r`.

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,427 @@
...
+        else:
+            gralora_dropout_layer = nn.Identity()
+
+        self.gralora_dropout.update(nn.ModuleDict({adapter_name: gralora_dropout_layer}))
+
+        # Actual trainable parameters
+        subblock_in_features = self.in_features // gralora_k
+        subblock_out_features = self.out_features // gralora_k
+
+        gralora_r = r - hybrid_r  # gralora_r is the rank allocated to gralora method
```
BenjaminBossan: Just wondering if it's a good idea to subtract `hybrid_r` from `r`. Would it not be easier to handle if we just leave `r` as it is? I know it would be different from how `hybrid_r` is handled in the original repo, but I find it more intuitive.

In case we agree to change this, the line `self.scaling[adapter_name] = gralora_alpha / r` below also needs to be changed: `self.scaling[adapter_name] = gralora_alpha / (r + hybrid_r)`

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,427 @@
...
+        assert gralora_r % gralora_k == 0, f"r should be divisible by gralora_k, but got {r} and {gralora_k}"
```
BenjaminBossan: Let's raise a proper `ValueError` here. Let's move the check into `GraloraConfig.__post_init__`.

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,427 @@
...
+
+        gralora_A = nn.ParameterList()
+        gralora_B = nn.ParameterList()
+        for _ in range(gralora_k):
+            new_A = nn.Parameter(torch.zeros(gralora_r, subblock_in_features))
+            new_B = nn.Parameter(torch.zeros(subblock_out_features, gralora_r))
```
BenjaminBossan: A small optimization would be to use `torch.empty` instead of `torch.zeros`. Would require to add init with zeros in case of `init_weights=True`.

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,427 @@
...
+
+        # Actual trainable parameters
+        subblock_in_features = self.in_features // gralora_k
+        subblock_out_features = self.out_features // gralora_k
+
+        gralora_r = r - hybrid_r  # gralora_r is the rank allocated to gralora method
+        assert gralora_r % gralora_k == 0, f"r should be divisible by gralora_k, but got {r} and {gralora_k}"
+
+        gralora_A = nn.ParameterList()
+        gralora_B = nn.ParameterList()
```
BenjaminBossan: Normal `[]` instead of `ParameterList` would be enough here, right?

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,427 @@
...
+        if gralora_dropout > 0.0:
+            gralora_dropout_layer = nn.Dropout(p=gralora_dropout)
+        else:
+            gralora_dropout_layer = nn.Identity()
+
+        self.gralora_dropout.update(nn.ModuleDict({adapter_name: gralora_dropout_layer}))
+
+        # Actual trainable parameters
+        subblock_in_features = self.in_features // gralora_k
+        subblock_out_features = self.out_features // gralora_k
```
BenjaminBossan, line 102: Is it necessary that the mod is zero?

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,427 @@
...
+                gralora_B = self.gralora_B[active_adapter]
+
+                gralora_A_general = self.gralora_A_general[active_adapter]
+                gralora_B_general = self.gralora_B_general[active_adapter]
+
+                r = self.r[active_adapter]
+                gralora_k = self.gralora_k[active_adapter]
+                hybrid_r = self.hybrid_r[active_adapter]
+
+                assert len(gralora_A) == len(gralora_B)
```
BenjaminBossan: I don't think this check is necessary. If you think it's better to have it, let's move it to `update_layer`.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+        - **model** ([`~transformers.PreTrainedModel`]) -- The model to be adapted.
+        - **peft_config** ([`GraloraConfig`]): The configuration of the Gralora model.
+    """
+
+    prefix: str = "gralora_"
+
+    def __init__(self, model, config, adapter_name) -> None:
+        super().__init__(model, config, adapter_name)
+
+    def _check_new_adapter_config(self, config: GraloraConfig) -> None:
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+        # TODO: there should be a check if any of the existing adapters actually has bias != "none", or else the check
+        # does not fully correspond to the error message.
+        if (len(self.peft_config) > 1) and (config.bias != "none"):
+            raise ValueError(
+                f"{self.__class__.__name__} supports only 1 adapter with bias. When using multiple adapters, "
+                "set bias to 'none' for all adapters."
+            )
+
+    @staticmethod
+    def _check_target_module_exists(gralora_config, key):
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+            )
+        else:
+            new_module = self._create_new_module(gralora_config, adapter_name, target, current_key, **kwargs)
+            if adapter_name not in self.active_adapters:
+                # adding an additional adapter: it is not automatically trainable
+                new_module.requires_grad_(False)
+            self._replace_module(parent, target_name, new_module, target)
+
+    @staticmethod
+    def _replace_module(parent, child_name, new_module, child):
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+            else:
+                new_module.state = child.state
+            new_module.to(child.weight.device)
+
+        # dispatch to correct device
+        for name, module in new_module.named_modules():
+            if "gralora_" in name:
+                module.to(child.weight.device)
+
+    def _mark_only_adapters_as_trainable(self, model: nn.Module) -> None:
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
BenjaminBossan, line 1: We had a recent refactor in `BaseTuner` that makes most of the methods you have to implement here obsolete. I commented on those methods, you can simply remove them.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+        new_module = Linear(
+            target,
+            adapter_name,
+            module_name,
+            **kwargs,
+        )
+
+        return new_module
+
+    def __getattr__(self, name: str):
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+        """Forward missing attributes to the wrapped module."""
+        try:
+            return super().__getattr__(name)  # defer to nn.Module's logic
+        except AttributeError:
+            return getattr(self.model, name)
+
+    def get_peft_config_as_dict(self, inference: bool = False):
```
BenjaminBossan: No longer needed, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+        self._set_adapter_layers(enabled=False)
+
+    def set_adapter(self, adapter_name, inference_mode: bool = False):
+        for module in self.model.modules():
+            if isinstance(module, GraloraLayer):
+                if module.merged:
+                    warnings.warn("Adapter cannot be set when the model is merged. Unmerging the model first.")
+                    module.unmerge()
+                module.set_adapter(adapter_name, inference_mode=inference_mode)
+        self.active_adapter = adapter_name
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+
+    @staticmethod
+    def _prepare_adapter_config(peft_config, model_config):
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+        if peft_config.target_modules is None:
+            if model_config["model_type"] not in TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING:
+                raise ValueError("Please specify `target_modules` in `peft_config`")
+            peft_config.target_modules = set(
+                TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING[model_config["model_type"]]
+            )
+        return peft_config
+
+    def _unload_and_optionally_merge(
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+                    target.merge(safe_merge=safe_merge, adapter_names=adapter_names)
+
+                self._replace_module(parent, target_name, target.get_base_layer(), target)
+            elif isinstance(target, ModulesToSaveWrapper):
+                # save any additional trainable modules part of `modules_to_save`
+                setattr(parent, target_name, target.modules_to_save[target.active_adapter])
+
+        return self.model
+
+    def delete_adapter(self, adapter_name: str):
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+        return self._unload_and_optionally_merge(
+            progressbar=progressbar, safe_merge=safe_merge, adapter_names=adapter_names
+        )
+
+    def unload(self):
+        """
+        Gets back the base model by removing all the Gralora modules without merging. This gives back the original base
+        model.
+        """
+        return self._unload_and_optionally_merge(merge=False)
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+        >>> config = GraloraConfig(r=128)
+        >>> model = get_peft_model(base_model, config)
+        ```
+
+    **Attributes**:
+        - **model** ([`~transformers.PreTrainedModel`]) -- The model to be adapted.
+        - **peft_config** ([`GraloraConfig`]): The configuration of the Gralora model.
+    """
+
+    prefix: str = "gralora_"
```
BenjaminBossan, line 59: Let's add two attributes here:

    tuner_layer_cls = GraloraLayer
    target_module_mapping = TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+    def get_peft_config_as_dict(self, inference: bool = False):
+        config_dict = {}
+        for key, value in self.peft_config.items():
+            config = {k: v.value if isinstance(v, Enum) else v for k, v in asdict(value).items()}
+            if inference:
+                config["inference_mode"] = True
+            config_dict[key] = config
+        return config_dict
+
+    def _set_adapter_layers(self, enabled=True):
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+        for module in self.model.modules():
+            if isinstance(module, (BaseTunerLayer, ModulesToSaveWrapper)):
+                module.enable_adapters(enabled)
+
+    def enable_adapter_layers(self):
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+        self._set_adapter_layers(enabled=True)
+
+    def disable_adapter_layers(self):
```
BenjaminBossan: Same as parent class, can be removed.

src/peft/tuners/gralora/model.py
```diff
@@ -0,0 +1,360 @@
...
+        for key in key_list:
+            _, target, _ = _get_submodules(self.model, key)
+            if isinstance(target, GraloraLayer):
+                target.delete_adapter(adapter_name)
+                if new_adapter is None:
+                    new_adapter = target.active_adapter[:]
+
+        self.active_adapter = new_adapter or []
+
+    def merge_and_unload(
```
BenjaminBossan: Same as parent class, can be removed.

tests/test_gralora.py
BenjaminBossan, line 1: Thanks a lot for very diligently adding unit tests for GraLoRA. However, I'm afraid that we can delete the majority of them :)

Most of what is tested here is already covered in existing tests (merging, disabling adapters, different dtypes, ...). This is especially true for the tests in `test_custom_models.py`, so please add the GraLoRA settings there (once without and once with hybrid) and delete them here.

There are some tests that check for errors or warnings at initialization time. Those should be moved to `test_initialization.py`. Check the tests there for inspiration

Some other tests, like checking the shape of parameters, are IMHO too granular and can be deleted without replacement. If the shapes are wrong, we would notice in the forward pass for instance.

There are a few tests that are not covered by existing tests and are not necessarily too granular, like 
`test_gralora_hybrid_forward_computation`, `test_gralora_information_exchange_via_permutation`, and `test_gralora_scaling_factor`. Personally, I'd be fine with removing those too but keeping them would be fine as well.

BenjaminBossan:
@yeonjoon-jung01 Please ping me when you're finished so that I know that I can give this another review. Also, if possible, please avoid force pushes or rebases, as those make reviews harder.

src/peft/tuners/gralora/layer.py
yeonjoon-jung01: Added `NotImplementedError` for other classes

src/peft/tuners/gralora/layer.py
yeonjoon-jung01: Added code for checking `hybrid_r`

src/peft/tuners/gralora/layer.py
yeonjoon-jung01: Yes, it is required to preserve the in_features and out_features value. Added code fore checking ValueError.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: Added the attributes for GraLoRA

src/peft/tuners/gralora/model.py
yeonjoon-jung01: Thank you for your kind review. I’ve removed the inherited methods from the parent class as suggested.

tests/test_gralora.py
yeonjoon-jung01: Again, thank you for your kind review. I’ve added the GraLoRA test code to test_custom_models.py.
You can run it using:

`pytest tests/test_custom_models.py -k "Gralora"`

Just in case, I’ve kept the `test_gralora.py `file to test GraLoRA’s unique features.

src/peft/tuners/gralora/config.py
yeonjoon-jung01: I’ve extended the description for better clarity.

src/peft/tuners/gralora/config.py
yeonjoon-jung01: I’ve extended the description.

src/peft/tuners/gralora/config.py
yeonjoon-jung01: I’ve extended the description.

src/peft/tuners/gralora/config.py
yeonjoon-jung01: I’ve extended the description, making it more clear.

src/peft/tuners/gralora/layer.py
yeonjoon-jung01: Removed the wrong code comment.

src/peft/tuners/gralora/layer.py
yeonjoon-jung01: Removed the inherited functions from the parent class.

src/peft/tuners/gralora/layer.py
yeonjoon-jung01: Removed the inherited functions from the parent class.

src/peft/tuners/gralora/layer.py
yeonjoon-jung01: Thank you for your insightful comment. I agree that having `r` and `hybrid_r` handled independently makes the design more intuitive. I’ve updated my code accordingly based on your suggestion. :)

src/peft/tuners/gralora/layer.py
yeonjoon-jung01: Yes, it is. I have updated the code.

src/peft/tuners/gralora/layer.py
yeonjoon-jung01: Added ValueError to check rank `r` and `gralora_k`

src/peft/tuners/gralora/layer.py
yeonjoon-jung01: Applied your suggestion :)

src/peft/tuners/gralora/layer.py
yeonjoon-jung01: I agree, I have removed the unnecessary check.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I have removed the deprecated code.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

src/peft/tuners/gralora/model.py
yeonjoon-jung01: I’ve removed the inherited methods from the parent class.

yeonjoon-jung01:
> @yeonjoon-jung01 Please ping me when you're finished so that I know that I can give this another review. Also, if possible, please avoid force pushes or rebases, as those make reviews harder. 

@BenjaminBossan  I’ve finished updating the code 🙂. I saw your message a bit late — I had already rebased the branch to sync with the main stream, just in case there might be any conflicts. I’ll make sure to avoid force pushes or rebases from now on. Sorry about that!

yeonjoon-jung01:
@BenjaminBossan I have also resolved the previously missed features.

I’ve extended the test coverage to include `test_custom_models.py`, `test_encoder_decoder_models.py`, `test_feature_extraction_models.py`, and `test_seq_classifier.py`.
Additionally, I’ve added corresponding documentation and example code.

docs/source/package_reference/gralora.md
```diff
@@ -0,0 +1,32 @@
+# GraLoRA
```
BenjaminBossan, line 1: Please also add an entry to the toctree.

src/peft/tuners/gralora/config.py
```diff
@@ -0,0 +1,125 @@
...
+            )
+        },
+    )
+    gralora_dropout: float = field(default=0.0, metadata={"help": "gralora dropout"})
+    gralora_k: int = field(
+        default=2,
+        metadata={
+            "help": (
+                "gralora_k determines the number of subblocks in the GraLoRA adapter."
+                "The total parameter count is preserved regardles of gralora_k, while the expressivitiy is multiplied by gralora_k."
```
BenjaminBossan: This sounds like the higher `gralora_k`, the better. Probably there is a tradeoff here. Could you please document it? Also, please mention the `self.r % self.gralora_k != 0` constraint.

src/peft/tuners/gralora/config.py
```diff
@@ -0,0 +1,125 @@
...
+
+from dataclasses import dataclass, field
+from typing import Optional, Union
+
+from peft.config import PeftConfig
+from peft.utils import PeftType
+
+
+@dataclass
+class GraloraConfig(PeftConfig):
```
BenjaminBossan, line 23: Please also add a docstring here. You can just copy the help text for the parameter description.

src/peft/tuners/gralora/layer.py
BenjaminBossan: Nice. Could you please add tests for these errors to `test_initialization.py`? Check the other tests there for inspiration. Please also add a test for the `self.r % self.gralora_k != 0` error.

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,398 @@
...
+        # stack A and B and transpose to get the final shape
+        gralora_A = torch.stack(tuple(gralora_A), dim=0)  # [N, gralora_r, in_features//N]
+        gralora_A = gralora_A.transpose(1, 2).contiguous()  # [N, in_features//N, gralora_r]
+
+        gralora_B = torch.stack(tuple(gralora_B), dim=0)  # [N, out_features//N, gralora_r]
+        gralora_B = gralora_B.transpose(1, 2).contiguous()  # [N, gralora_r, out_features//N]
+
+        if hybrid_r > 0:
+            general_gralora_A = nn.Linear(self.in_features, hybrid_r, bias=False)
+            general_gralora_B = nn.Linear(hybrid_r, self.out_features, bias=False)
```
BenjaminBossan, line 131: As a small optimization, you could use the `torch.utils.integration.init_empty_weights` context manager to initialize these parameters with empty tensors too.

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,398 @@
...
+            gralora_B = gralora_B.float()
+
+        # Get dimensions
+        in_features = self.in_features
+        out_features = self.out_features
+        gralora_rank = r
+        if in_features % gralora_k != 0:
+            raise ValueError(f"in_features should be divisible by gralora_k, but got {in_features} and {gralora_k}")
+        elif out_features % gralora_k != 0:
+            raise ValueError(f"out_features should be divisible by gralora_k, but got {out_features} and {gralora_k}")
```
BenjaminBossan: We already check this at initialization, so there is no need for this, right? The shape can't change later on.

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,398 @@
...
+        elif gralora_rank % gralora_k != 0:
+            raise ValueError(f"rank should be divisible by gralora_k, but got {gralora_rank} and {gralora_k}")
```
BenjaminBossan: Should this be checked at initialization time?

tests/test_gralora.py
```diff
@@ -0,0 +1,1051 @@
...
+        mlp_hybrid = MLP()
+        config_hybrid = GraloraConfig(
+            target_modules=["lin1", "lin2"],
+            r=12,
+            gralora_k=4,
+            hybrid_r=4,
+        )
+        model_hybrid = get_peft_model(mlp_hybrid, config_hybrid)
+
+        def count_trainable_params(model):
```
BenjaminBossan: `model.get_nb_trainable_parameters()` should work.

tests/test_gralora.py
```diff
@@ -0,0 +1,1051 @@
...
+        loaded_model.eval()
+
+        # Get output after load
+        with torch.no_grad():
+            output_after = loaded_model(x)
+
+        # Outputs should be very close
+        assert torch.allclose(output_before, output_after, atol=1e-5, rtol=1e-5)
+
+    def test_gralora_state_dict_structure(self, mlp_gralora_hybrid, tmp_path):
```
BenjaminBossan: IMO, this test is redundant, if this test failed, then saving and loading GraLoRA would also fail.

tests/test_gralora.py
```diff
@@ -0,0 +1,1051 @@
...
+
+        # Output with adapter disabled
+        with model.disable_adapter():
+            with torch.no_grad():
+                output_disabled = model(x)
+
+        # Outputs should be different
+        assert not torch.allclose(output_enabled, output_disabled, atol=1e-6, rtol=1e-6)
+
+    def test_gralora_different_k_values(self):
```
BenjaminBossan: IMO, this test is unnecessary. If we have reason to believe that it could cover some bugs, I'd rather add one more setting to `test_custom_models.py` using a different `r` and `gralora_k`.

tests/test_gralora.py
```diff
@@ -0,0 +1,1051 @@
...
+            )
+            model = get_peft_model(mlp, config)
+
+            x = torch.randn(5, 10)
+            output = model(x)
+
+            assert output.shape == (5, 2)
+            assert not torch.isnan(output).any()
+
+    def test_gralora_rank_divisibility_check(self):
```
BenjaminBossan: This should be in `test_initialization.py`.

tests/test_custom_models.py
```diff
@@ -666,6 +667,25 @@
...
+        "MLP",
+        GraloraConfig,
+        {"target_modules": ["lin0"], "modules_to_save": ["lin1"]},
+    ),
+    (
+        "Embedding + transformers Conv1D 1 GraLoRA",
+        "EmbConv1D",
+        GraloraConfig,
+        {"target_modules": ["conv1d"], "gralora_k": 1},
+    ),
```
BenjaminBossan, line 700: How about adding a setting with hybrid GraLoRA?

tests/test_gralora.py
```diff
@@ -0,0 +1,1051 @@
...
+        torch.manual_seed(0)
+        mlp = MLP()
+        config = GraloraConfig(target_modules=["lin1"], r=8, gralora_k=2)
+        model = get_peft_model(mlp, config)
+
+        # Try to unmerge without merging first
+        with pytest.warns(UserWarning, match="Already unmerged"):
+            model.base_model.model.lin1.unmerge()
+
+    def test_gralora_hybrid_forward_computation(self):
```
BenjaminBossan: If you add a setting to `test_custom_models.py` with hybrid GraLoRA, this test would become redundant.

tests/test_gralora.py
```diff
@@ -0,0 +1,1051 @@
...
+        x = torch.randn(5, 10)
+
+        with torch.no_grad():
+            output_hybrid = model_hybrid(x)
+            output_pure = model_pure(x)
+
+        # Outputs should be different due to hybrid component
+        assert not torch.allclose(output_hybrid, output_pure, atol=1e-3)
+
+    def test_gralora_invalid_rank_zero(self):
```
BenjaminBossan: Move to `test_initialization.py`

tests/test_gralora.py
```diff
@@ -0,0 +1,1051 @@
...
+        """Test that r=0 raises error"""
+        mlp = MLP()
+        config = GraloraConfig(target_modules=["lin1"], r=0, gralora_k=2)
+
+        with pytest.raises(ValueError, match="`r` should be a positive integer"):
+            get_peft_model(mlp, config)
+
+    def test_gralora_invalid_rank_negative(self):
```
BenjaminBossan: Move to `test_initialization.py`

tests/test_gralora.py
```diff
@@ -0,0 +1,1051 @@
...
+            gralora_k=2,
+            bias="gralora_only",
+        )
+        model = get_peft_model(mlp, config)
+
+        # Only GraLoRA layer biases should be trainable
+        assert model.base_model.model.lin1.bias.requires_grad
+        assert not model.base_model.model.lin0.bias.requires_grad
+
+    def test_gralora_multiple_adapters_with_bias_raises(self):
```
BenjaminBossan: Move to `test_initialization.py`

tests/test_gralora.py
```diff
@@ -0,0 +1,1051 @@
...
+        # Unload without merge
+        unloaded_model = model.unload()
+
+        with torch.no_grad():
+            unloaded_output = unloaded_model(x)
+
+        # Should match base model output (no merge)
+        assert torch.allclose(base_output, unloaded_output, atol=1e-5)
+
+    def test_gralora_merge_with_hybrid_component(self):
```
BenjaminBossan: If you add a setting to test_custom_models.py with hybrid GraLoRA, this test would become redundant.

tests/test_gralora.py
```diff
@@ -0,0 +1,1051 @@
...
+        X = self.lin1(X)
+        X = self.relu(X)
+        X = self.lin2(X)
+        X = self.relu(X)
+        X = self.lin3(X)
+        X = self.sm(X)
+        return X
+
+
+class TestGralora:
```
BenjaminBossan: There are still a lot of tests here that are unnecessary because they're already covered by existing tests. Please remove them:

- test_gralora_forward_pass
- test_gralora_backward_pass
- test_gralora_save_load_roundtrip
- test_gralora_merge_and_unload
- test_gralora_merge_unmerge
- test_gralora_multiple_adapters
- test_gralora_dtype_compatibility
- test_gralora_disable_adapters
- test_gralora_trainable_parameters_only
- test_gralora_save_pretrained_files
- test_gralora_safe_merge_success
- test_gralora_safe_merge_detects_nan
- test_gralora_cpu_fp16_merge
- test_gralora_cpu_bf16_merge
- test_gralora_delete_adapter
- test_gralora_delete_nonexistent_adapter_raises
- test_gralora_unload_without_merge
- test_gralora_merge_with_adapter_names
- test_gralora_enable_disable_adapter_layers
- test_gralora_forward_with_merged_adapter
- test_gralora_forward_with_disable_adapters_and_merged
- test_gralora_add_non_active_adapter

BenjaminBossan (changes requested):
Thanks for the updates to the PR. I did another review round, please check.

Also, before committing your changes, please call `make style`. Ensure that you have the correct version of ruff installed (`0.12.12`).

yeonjoon-jung01:
@BenjaminBossan  I’ve resolved all of your comments and applied the suggested changes. The main update is that I removed tests/test_gralora.py and integrated the related test cases into the existing test_initialization and test_custom_models files, including additional scenarios for Hybrid GraLoRA.

tests/test_gralora.py
yeonjoon-jung01: I have simply removed `test_gralora.py`, and integrated into existing files

BenjaminBossan:
@yeonjoon-jung01 Could you please run `make style`?

yeonjoon-jung01:
@BenjaminBossan I have run `make style` updated the documentations

src/peft/tuners/gralora/layer.py
```diff
@@ -0,0 +1,392 @@
...
+        in_features = self.in_features
+        out_features = self.out_features
+        gralora_rank = r
+        subblock_gralora_rank = gralora_rank // gralora_k
+
+        # scatter gralora_A to get the scattered weight matrix
+        l_indices = torch.arange(in_features, device=device)
+        n_indices = l_indices // (in_features // gralora_k)
+        i_indices = l_indices % (in_features // gralora_k)
+        gralora_A_scattered = torch.zeros(in_features, gralora_k, gralora_rank, device=device, dtype=dtype)
```
BenjaminBossan: float16 tests are failing when run on CPU. The reason is because `if cast_to_fp32`, `gralora_A` is cast to float32 but `gralora_A_scattered` is float16. I see two options:

1. If `cast_to_fp32`, cast `gralora_A_scattered` to float32 instead.
2. Get rid of `cast_to_fp32` completely

The reason why `cast_to_fp32` exists is explained in #1977. To be honest, I don't know how big of a factor this still is (perhaps new torch versions no longer suffer from this?) and I don't have a means to test it.

yeonjoon-jung01: @BenjaminBossan I have followed option 1 and updated the code :)

yeonjoon-jung01:
@BenjaminBossan Do you think there’s any additional code I should test or update?

examples/gralora_finetuning/README.md
```diff
@@ -0,0 +1,71 @@
...
+from peft import GraloraConfig, get_peft_model
+from transformers import AutoTokenizer, AutoModelForCausalLM, Trainer
+from datasets import load_dataset
+
+model = AutoModelForCausalLM.from_pretrained("huggyllama/llama-7b", device_map="auto")
+tokenizer = AutoTokenizer.from_pretrained("huggyllama/llama-7b")
+dataset = load_dataset("timdettmers/openassistant-guanaco", split="train")
+gralora_config = GraloraConfig()
+peft_model = get_peft_model(model, gralora_config)
+trainer = transformers.Trainer(
```
BenjaminBossan:
```suggestion
trainer = Trainer(
```

examples/gralora_finetuning/README.md
```diff
@@ -0,0 +1,71 @@
...
+    dataset_text_field="text",
+    max_seq_length=2048,
+    tokenizer=tokenizer,
+)
+trainer.train()
+peft_model.save_pretrained("gralora-llama-3-8b")
+```
+
+Run the finetuning script simply by running:
+```python
```
BenjaminBossan:
```suggestion
```sh
```

examples/gralora_finetuning/README.md
```diff
@@ -0,0 +1,71 @@
...
+dataset = load_dataset("timdettmers/openassistant-guanaco", split="train")
+gralora_config = GraloraConfig()
+peft_model = get_peft_model(model, gralora_config)
+trainer = transformers.Trainer(
+    model=peft_model,
+    train_dataset=dataset,
+    dataset_text_field="text",
+    max_seq_length=2048,
+    tokenizer=tokenizer,
+)
```
BenjaminBossan: The example didn't run for me because of unknown arguments. I know you copied an existing one, but either that one also didn't work, or it's because of some recent changes in `Trainer`. Anyway, here is a suggestion that works for me:

```python
#peft_model = get_peft_model(model, gralora_config)  <= remove this
trainer = SFTTrainer(
    model=model,
    train_dataset=dataset,
    processing_class=tokenizer,
    peft_config=gralora_config,
    args=SFTConfig(
        max_length=2048,
        dataset_text_field="text",
        per_device_train_batch_size=2,
    ),
)
trainer.train()
trainer.model.save_pretrained("gralora-llama-3-8b")
```

SFTTrainer and SFTConfig need to imported from trl. Alternatively, you could also rewrite it to use

examples/gralora_finetuning/README.md
```diff
@@ -0,0 +1,71 @@
...
+peft_model = PeftModel.from_pretrained(model, "gralora-llama-3-8b")
+```
+
+## Additonal Notes
+While `gralora_k` is set to 2 for default, you can increase this value to create more fine-grained adapters. `gralora_k` of 4 is recommended when the total rank (`r + hybrid_r`) is 64 or higher.
+
+
+
+
```
BenjaminBossan: Delete 3 empty lines

examples/gralora_finetuning/gralora_finetuning.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # Define training arguments
+    training_args = TrainingArguments(
+        output_dir=output_dir,
+        num_train_epochs=num_epochs,
+        per_device_train_batch_size=batch_size,
+        per_device_eval_batch_size=batch_size,
+        warmup_steps=100,
+        weight_decay=0.01,
+        logging_dir="./logs",
```
BenjaminBossan: This argument errors for me, does it work for you?

examples/gralora_finetuning/gralora_finetuning.py
```diff
@@ -0,0 +1,213 @@
...
+    )
+    parser.add_argument(
+        "--output_dir", type=str, default="path/to/output", help="Output directory for the fine-tuned model"
+    )
+    parser.add_argument("--batch_size", type=int, default=1, help="Batch size")
+    parser.add_argument("--num_epochs", type=int, default=1, help="Number of training epochs")
+    parser.add_argument("--learning_rate", type=float, default=1e-4, help="Learning rate")
+    parser.add_argument("--cutoff_len", type=int, default=512, help="Cutoff length for tokenization")
+    parser.add_argument("--val_set_size", type=int, default=500, help="Validation set size")
+    parser.add_argument("--quantize", action="store_true", help="Use quantization")
```
BenjaminBossan: Quantizing wouldn't work, right?

BenjaminBossan (changes requested):
Thanks for the updates, the change looks good.

I focused on the examples this time and found a few issues. Some are possibly due to some recent changes in transformers, not sure, but we should update them so that the examples run out of the box.

Moreover, I ran an experiment with GraLoRA on the [PEFT MetaMath benchmark](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA). I used the default settings for the config and in general, the results compare to LoRA rank 32, with similar memory usage and training time. However, the final test accuracy fell slightly short, attaining 46.2% compared to 48.2% with LoRA. If you have any suggestion for better GraLoRA hyper-parameters for this experiment, feel free to check them in as an experiment. Otherwise, we can also work the defaults.

yeonjoon-jung01:
> Thanks for the updates, the change looks good.
> 
> I focused on the examples this time and found a few issues. Some are possibly due to some recent changes in transformers, not sure, but we should update them so that the examples run out of the box.
> 
> Moreover, I ran an experiment with GraLoRA on the [PEFT MetaMath benchmark](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA). I used the default settings for the config and in general, the results compare to LoRA rank 32, with similar memory usage and training time. However, the final test accuracy fell slightly short, attaining 46.2% compared to 48.2% with LoRA. If you have any suggestion for better GraLoRA hyper-parameters for this experiment, feel free to check them in as an experiment. Otherwise, we can also work the defaults.

@BenjaminBossan Could you please try with learning rate 2e-4 instead of the default 1e-4 for GraLoRA?

examples/gralora_finetuning/gralora_finetuning.py
yeonjoon-jung01: It worked for me! I guess it’s a version dependency issue. I used `transformers==4.57.1`, and the error did not occur.

examples/gralora_finetuning/README.md
yeonjoon-jung01: I have updated the code to use SFTTrainer and SFTConfig :)

examples/gralora_finetuning/gralora_finetuning.py
yeonjoon-jung01: I have checked that quantization works fine with the GraLoRA method.

yeonjoon-jung01:
> > Thanks for the updates, the change looks good.
> > I focused on the examples this time and found a few issues. Some are possibly due to some recent changes in transformers, not sure, but we should update them so that the examples run out of the box.
> > Moreover, I ran an experiment with GraLoRA on the [PEFT MetaMath benchmark](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA). I used the default settings for the config and in general, the results compare to LoRA rank 32, with similar memory usage and training time. However, the final test accuracy fell slightly short, attaining 46.2% compared to 48.2% with LoRA. If you have any suggestion for better GraLoRA hyper-parameters for this experiment, feel free to check them in as an experiment. Otherwise, we can also work the defaults.
> 
> @BenjaminBossan Could you please try with learning rate 2e-4 instead of the default 1e-4 for GraLoRA?

@BenjaminBossan I’ve tested the method on my side with both rank 32 and 64, and in both cases, it achieved higher performance than LoRA. If you’d like, I can also commit the configuration and result JSON files, or you’re welcome to try reproducing it on your end.
I used a learning rate of 2e-4 for both cases, as GraLoRA is more robust to outliers, making a higher learning rate manageable. For the other settings, I used rank 32 with alpha 64 and rank 64 with alpha 128. All other parameters were kept at their default values.

examples/gralora_finetuning/gralora_finetuning.py
BenjaminBossan: I checked and it's an upcoming change for transformers v5. Thus it's better to update the example to make it work so that it's not out of date in a few weeks. You could install transformers from source for testing.

I raised my concern with the colleagues about removing this argument without deprecation (transformers PR 41375), let's see if they change their mind, but still let' remove it here.

examples/gralora_finetuning/gralora_finetuning.py
BenjaminBossan: Ah okay, it works because `Linear4bit` is a subclass of `nn.Linear`, but it still won't work correctly, e.g. merging weights is not supported for quantized weights. Therefore, I'd rather remove the option. Otherwise users may spend time and resources to train on a quantized model only to later notice that they can't merge when it comes to inference.

BenjaminBossan (changes requested):
> Could you please try with learning rate 2e-4 instead of the default 1e-4 for GraLoRA?

This does indeed help. With rank 32, I get higher accuracy now, 48.6% compared to LoRA rank 32 getting 48.2%. For rank 64 (alpha 128), I get 52.7% with GraLoRA and 53.0% with LoRA. In both cases, the memory usage is very close between the two methods.

We can check in those experiments, but if you have other suggestions that may work better, e.g. different `hybrid_r` or `gralora_k`, LMK.

yeonjoon-jung01:
> > Could you please try with learning rate 2e-4 instead of the default 1e-4 for GraLoRA?
> 
> This does indeed help. With rank 32, I get higher accuracy now, 48.6% compared to LoRA rank 32 getting 48.2%. For rank 64 (alpha 128), I get 52.7% with GraLoRA and 53.0% with LoRA. In both cases, the memory usage is very close between the two methods.
> 
> We can check in those experiments, but if you have other suggestions that may work better, e.g. different `hybrid_r` or `gralora_k`, LMK.

@BenjaminBossan I also tried using a learning rate of 2e-4 for rank 32 and alpha 64, which achieved a test accuracy of 50.7%. The experiment was run on a single NVIDIA RTX A6000. I believe the difference may have resulted from variations in hardware or library versions.

I’m just curious if there might be any other differing settings. I’ve attached the following as adapter_config.json.
```
{
  "r": 32,
  "hybrid_r": 0,
  "target_modules": null,
  "gralora_alpha": 64,
  "gralora_dropout": 0.0,
  "gralora_k": 2,
  "fan_in_fan_out": false,
  "bias": "none",
  "init_weights": true,
  "layers_to_transform": null,
  "layers_pattern": null,
  "modules_to_save": null,
  "peft_type": "GRALORA",
  "megatron_config": null,
  "megatron_core": "megatron.core",
  "revision": null,
  "task_type": "CAUSAL_LM"
}
```
with following for the training_params.json
```
{
  "optimizer_kwargs": {
    "lr": 2e-4
  }
}
```
the result and package info is as follow
 ```
"step": 5000,
"test accuracy": 0.5072024260803639,
"train loss": 0.5921622431278228,
"train samples": 20000,
"train total tokens": 4198051
```

```
"transformers-version": "4.57.1",
"transformers-commit-hash": null,
"peft-version": "0.17.2.dev0",
"peft-commit-hash": "a1c944a4c3aceffb5d9dc3ee4e37aaad76d55f5c",
"datasets-version": "4.3.0",
"datasets-commit-hash": null,
"bitsandbytes-version": "0.48.1",
"bitsandbytes-commit-hash": null,
"torch-version": "2.7.1+cu126",
"torch-commit-hash": null
```

**Comment by BenjaminBossan** [comment]
My `adapter_config.json` for rank 32 is:

```json
{
  "auto_mapping": null,
  "bias": "none",
  "fan_in_fan_out": false,
  "gralora_alpha": 64,
  "gralora_dropout": 0.0,
  "gralora_k": 2,
  "hybrid_r": 0,
  "inference_mode": false,
  "init_weights": true,
  "layers_pattern": null,
  "layers_to_transform": null,
  "modules_to_save": null,
  "peft_type": "GRALORA",
  "peft_version": "0.17.2.dev0@UNKNOWN",
  "r": 32,
  "revision": null,
  "target_modules": null,
  "task_type": null
}
```

Training params are the same.

> I also tried using a learning rate of 2e-4 for rank 32 and alpha 64, which achieved a test accuracy of 50.7%

The same accuracy for both ranks?

> I believe the difference may have resulted from variations in hardware or library versions.

Possibly, just as an example, I use torch 2.8.0. The final run will be on an AWS instance we use for all models, so the score may differ yet from what I reported, as I ran the experiment locally.

yeonjoon-jung01:
> > Could you please try with learning rate 2e-4 instead of the default 1e-4 for GraLoRA?
> 
> This does indeed help. With rank 32, I get higher accuracy now, 48.6% compared to LoRA rank 32 getting 48.2%. For rank 64 (alpha 128), I get 52.7% with GraLoRA and 53.0% with LoRA. In both cases, the memory usage is very close between the two methods.
> 
> We can check in those experiments, but if you have other suggestions that may work better, e.g. different `hybrid_r` or `gralora_k`, LMK.



> My `adapter_config.json` for rank 32 is:
> 
> 
> Training params are the same.
> 
> > I also tried using a learning rate of 2e-4 for rank 32 and alpha 64, which achieved a test accuracy of 50.7%
> 
> The same accuracy for both ranks?
> 
> > I believe the difference may have resulted from variations in hardware or library versions.
> 
> Possibly, just as an example, I use torch 2.8.0. The final run will be on an AWS instance we use for all models, so the score may differ yet from what I reported, as I ran the experiment locally.

@BenjaminBossan For rank 64, the accuracy ranged between 52.5 and 53%, depending on the hardware (A6000 and H100), which aligns with your reported results. I haven't tested yet, but I think we could consider setting `gralora_k=4` for rank 64.

Additionally, I believe the accuracy for rank 64 LoRA is documented as 48.9% in the `lora--llama-3.2-3B-rank64.json` file. I wonder if any configuration has changed.

Finally, I think a batch size of 4 might be too small for stable training. How about adding an accumulation step option and increasing the effective batch size to a more common value, such as 128 or 192?

yeonjoon-jung01:
@BenjaminBossan  I guess you could add the GraLoRA rank-32 example with a learning rate of 2e-4 for now.

I believe the accuracy results vary significantly across different settings (hardware and library versions) due to the instability caused by the small batch size. If you’re planning to scale up the batch size, I might try other configurations then.

Otherwise, please let me know if there’s anything else I should take care of.

BenjaminBossan:
> I guess you could add the GraLoRA rank-32 example with a learning rate of 2e-4 for now.

Could you please push the experiments to this PR (only the configs, not the results)? Since the learning rate also needs a different value, please include the `training_params.json` (like [here](https://github.com/huggingface/peft/blob/d43b3151611e14b220371f169cc73cf5727e8b51/method_comparison/MetaMathQA/experiments/prefixtuning/llama-3.2-3B-lr_0.001/training_params.json)).

> Additionally, I believe the accuracy for rank 64 LoRA is documented as 48.9% in the `lora--llama-3.2-3B-rank64.json` file. I wonder if any configuration has changed.

I compared to LoRA with rank 64 and rslora enabled for better alpha values:

https://github.com/huggingface/peft/blob/d43b3151611e14b220371f169cc73cf5727e8b51/method_comparison/MetaMathQA/results/lora--llama-3.2-3B-rank64-rslora.json#L321

> I think a batch size of 4 might be too small for stable training. How about adding an accumulation step option and increasing the effective batch size to a more common value, such as 128 or 192?

We could think about adding gradient accumulation to the script, but we wanted to keep it simple on purpose, and gradient accumulation can be tricky to get right. For this PR, let's keep the setting as they are. Since the other methods use the same batch size, I think the comparison is still fair.

yeonjoon-jung01:
@BenjaminBossan I was just wondering if increasing the batch size could help stabilize the training process and make the final results more consistent across different settings (hardware and library versions). However, I also agree that it’s fine to keep the current settings as they are in this PR.

I’ve added the experiment configs for GraLoRA :)

BenjaminBossan:
@yeonjoon-jung01 Could you please run `make style` :)

yeonjoon-jung01:
> @yeonjoon-jung01 Could you please run `make style` :)

@BenjaminBossan  I have updated the code 👍

yeonjoon-jung01:
> Thanks a lot @yeonjoon-jung01 for the last update and for your great work on this PR overall. It's now in a finished state as everything LGTM.
> 
> We will not merge it right now as PEFT is currently in feature freeze. As soon as the next release is out, which shouldn't be too long in the future, this PR will be merged.

@BenjaminBossan Thanks a lot for your guidance during this PR! Really appreciate the helpful feedbacks. Looking forward to it being merged after the next release.

src/peft/tuners/gralora/config.py
```diff
@@ -0,0 +1,183 @@
...
+            Hybrid GraLoRA, a combination of GraLoRA and vanilla LoRA, becomes available when hybrid_r > 0. The
+            parameter count of the GraLoRA adapter is r + hybrid_r.
+        target_modules (`Union[List[str], str]`):
+            List of module names or regex expression of the module names to replace with GraLoRA. " For example, ['q',
+            'v'] or '.*decoder.*(SelfAttention|EncDecAttention).*(q|v)$'. " This can also be a wildcard 'all-linear'
+            which matches all linear/Conv1D " "(if the model is a PreTrainedModel, the output layer excluded). " If not
+            specified, modules will be chosen according to the model architecture, If the architecture is " not known,
+            an error will be raised -- in this case, you should specify the target modules manually. " To avoid
+            targeting any modules (because you want to apply `target_parameters`), set " `target_modules=[]`.
+        gralora_alpha (`int`): GraLoRA alpha.
```
BenjaminBossan: Let's rename this argument to `alpha`.

yeonjoon-jung01: @BenjaminBossan I have update the argument name from `gralora_alpha` to `alpha`.

yeonjoon-jung01:
@BenjaminBossan Do you have an estimated timeline for when this PR might be merged?

BenjaminBossan (approved):
Thanks for the last update, now the PR is ready to be merged. Great work and thanks for your patience.

yeonjoon-jung01:
Thank you again for the review and for your time guiding this through :)

PR #2853, opened 2025-10-20
FIX: DeLoRA adapter deletion issue

BenjaminBossan:
Currently, adapter deletion raises an error with DeLoRA. The reason is that the dropout module is called `module_dropout`, i.e. the prefix `"delora"` is not part of the name, which is required for proper working. This PR renames the attribute to `delora_dropout`.

The reason why this was not caught is because the corresponding test was not updated to include DeLoRA. The test was thus also changed.

I also noticed that the error message for save merging is different from other PEFT methods. In general, that would be okay, but it means we would have to add a special case for unit testing this, so I consolidated the error message.

Note: I came across these issues in #2846 but I wanted to fix it in a separate PR, as #2846 is probably not going to make it into the next PEFT release.

### DIFF
--- a/src/peft/tuners/delora/layer.py
+++ b/src/peft/tuners/delora/layer.py
@@ -200,7 +200,7 @@ class DeloraLinear(nn.Module, DeloraLayer):
 
                         if not torch.isfinite(orig_weights).all():
                             raise ValueError(
-                                f"NaNs detected in merged weights for adapter {active_adapter}; aborting merge"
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
FYI @mwbini

mwbini:
Thanks for fixing these. I also noticed the methods constraints in `delora.md` were missing (looking at other methods' docs) and updated this in PR #2854

githubnemo (approved):

PR #2854, opened 2025-10-20
DOC Update DeLoRA docs

mwbini:
I added the constraints for DeLoRA and rephrased a few parts in the package reference docs


BenjaminBossan (approved):
Good catch, thanks.

PR #2860, opened 2025-10-22
Fix #2826: implement gradient checkpoint callbacks

### LINKED ISSUE #2826: Gradient checkpointing and Activated LoRA models
### System Info

NA

### Who can help?

I can potentially implement changes in a PR, but would like guidance on how maintainers would prefer this fixed. Previously, I worked with @githubnemo on building the Activated LoRA feature for PEFT.

### Reproduction

Activated LoRA in PEFT works by having forward hooks for each layer that allow an argument "alora_offsets" to be passed from the PeftModelforCausalLM.forward into the appropriate lora variant forward. 

I've discovered that with gradient checkpointing (at least in the simple DPO example below), the .backward operation does not save this alora_offsets input, does not re-run the forward hooks to recreate it, and hence uses alora_offsets = None in practice for the .backward computation. To be clear, no errors are thrown, just the computed gradients are zero. Also turning gradient_checkpointing=False fixes the issue (as you might expect).

There might be a lot of ways to try to fix this and keep support for gradient checkpointing. For instance, the simplest fix (but also pretty unsafe) would be to edit ALoraLinearVariant.forward in tuners/lora/variants.py by adding the following lines: 
```
alora_offsets = kwargs.get("alora_offsets", None) 
if alora_offsets is not None:
         module.alora_offsets = alora_offsets
 else:
         alora_offsets = module.alora_offsets
```
Here, the initial forward pass in the training loop will use the specified alora_offsets coming from the forward hooks and save it. In the gradient-checkpointed backward pass, alora_offsets will be None, but the original value would be recovered and used. While this does seem to work (I tried it), it feels unsafe to me.

Perhaps there could be a safer way of doing this, e.g. by hashing the layer input x and active adapter, and using that to store alora_offsets more safely, i.e. the saved alora_offsets will only be used if the hashed value matches a key in some offsets table. As a result, this should only trigger in a .backward scenario where the true alora_offsets got removed for some reason. 

What do you think (e.g. @githubnemo )? Happy to clarify anything and/or discuss further.

Example train script where you can see that the gradients are all 0 (fixed by turning off gradient checkpointing).
```
from peft import LoraConfig
import os
from trl import DPOConfig, DPOTrainer


model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen2-0.5B-Instruct",device_map="cuda", trust_remote_code=True)
tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen2-0.5B-Instruct",padding_side='left', trust_remote_code=True)
model.pad_token_id = tokenizer.pad_token_id
train_dataset = load_dataset("trl-lib/ultrafeedback_binarized", split="train")
test_dataset = load_dataset("trl-lib/ultrafeedback_binarized", split="test")
peft_config = LoraConfig(
            task_type="CAUSAL_LM",
                r=32,
                lora_alpha=32,
                lora_dropout=0.05,
                bias="none",
                alora_invocation_tokens = [151644], #this is from the chat template, i.e. the generation prompt
                target_modules=["q_proj", "k_proj", "v_proj","o_proj",
                                                    "gate_proj", "up_proj", "down_proj",
                                                        ],
)
training_args = DPOConfig(output_dir="Qwen2-0.5B-DPO", report_to="wandb",per_device_train_batch_size=4,
            per_device_eval_batch_size=4,
            learning_rate=5e-6,
            gradient_checkpointing=True, 
            gradient_accumulation_steps=2,
            eval_strategy="steps",   
            eval_steps=500,                
            logging_steps=50,
            max_length=2048,
            save_strategy="steps",
            save_steps=1000,
            save_total_limit=2,)
trainer = DPOTrainer(model=model, args=training_args, processing_class=tokenizer, train_dataset=train_dataset,peft_config=peft_config, eval_dataset=test_dataset)
trainer.train()
model.save_pretrained("./models/dpo/alora")
```

### Expected behavior

See above

githubnemo:
Relevant issue: #2826

Gradient checkpointing is different to the normal forward/backward process in that it may execute forward
steps in the backward process. It is therefore decoupled from the normal forward process. Since we have
methods, such as activated LoRA, that depend on `peft_forward_hooks` in normal forward we have to make sure
that these hooks are applied in the independently executed forward by gradient checkpointing as well.

To this end we had several options:
- Don't use hooks to communicate the alora offsets, write them to the module directly. We can do that but then
  we'd need a mechanism to clean up these values afterwards (probably involving hooks) and we would be introducing
  a non-general way which might biting us in the future since more methods need parameter injection.
- Don't support `use_reentrant=True` (default for transformers) and use `context_fn` parameter to inject parameters
  when using `use_reentrant=False`. `torch.utils.checkpoint` supports adding a `context_fn` parameter which returns
  two context managers (one for the normal, one for the checkpoint forward). In theory this could be a way to
  inject the variables into the module. In practice we would need to modify the `keywords` argument of every
  `._gradient_checkpointing_func` attribute of every module to inject the `context_fn` callback and update those
  accordingly every forward call. This is far less reliable than forward hooks.
- Register forward hooks on the gradient checkpointing layers that apply the same hooks that `enable_peft_forward_hooks`
  does - but decoupled from `enable_peft_forward_hooks`. These hooks are removed once a full backward hook on the
  gradient checkpoint layer is called. We'd still need to use shared storage to store the hook handles so that
  we don't rack up forward and backward hooks but this storage is a general way of implementing forward hooks
  in gradient checkpointing. It also let's us control the flow without using private methods/variables. Since this
  adds forward hooks that are only removed when backward is called, we therefore disallow multiple forward calls
  in succession before a backward call (the user can do that with gradient checkpointing disabled).

In this change I'm implementing the last option, forward hooks on gradient checkpointing layers.

We already had tests but there were some issues that are improved in this PR:

- parametrize `use_reentrant` so we check both, even though `use_reentrant=True` is the default;
  since `use_reentrant=False` has a consistency checker it might detect errors that we don't cover
- check consistency between normal and checkpointed model runs, both runs must have the same set
  of non-zero gradients. it is not sufficient to check that there is a gradient, it must be non-zero
  as well (gradient checkpointing can fail with zero grads)
- set `model.train` to enable gradient checkpointing
- disable `lora_dropout` if set to make gradients (a bit more) deterministic

While testing I found that GPTBigCode doesn't support gradient checkpointing even though it says so,
skipping for now until fixed. BART doesn't work with the newest changes and fails at `loss.backward()`
with `RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn`).
**Update:** The BART issue is caused by the model not sharing the embed tokens module but just the weights. This in turn leads to `get_input_embeddings` returning one specific module which may not be invoked in the forward call at all (`model.shared` in this case - it has the same weights but it is a different `nn.Module`). Since `enable_input_require_grads` depends on the module returned by `get_input_embeddings` to have working forward hooks, the preparation for gradient checkpointing fails. This needs to be fixed in transformers either by targeting all tied weights or sharing the embedding *modules* instead of just the weights (like in T5).

Also, I emoved gradient requirement in CPT's embedding.
This module is never called in a gradient path so it is safe to set it
to `requires_grad=False`. This helps in upholding the assumption that
all parameters that require a gradient need to receive one.

Pinging @kgreenewald as well.

### DIFF
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -363,8 +363,9 @@ class LoraModel(BaseTuner):
             for n, layer in self.named_modules():
                 # gradient checkpointing layer are executed concurrently to the 'normal' forward call
                 # (in the backward step the gradient checkpointing layer's forward will be executed again).
-                # to be consistent with the normal forward we need to enable the pre hooks for this concurrent
-                # forward call as well.
+                # this means that when the gradient checkpointing layer is called, the _enable_peft_forward_hooks
+                # context manager is long gone. to be consistent with the normal forward we need to register the pre
+                # hooks for this concurrent forward call as well.
                 #
                 # Note that this will lead to double application of whatever the callbacks do in normal forward.
                 # Make sure that whatever change is done, can be applied more than once without harm (idempotency).
@@ -380,7 +381,7 @@ class LoraModel(BaseTuner):
                                 module._peft_gradient_checkpointing_forward_hooks.append(handle)
 
                     def backward_hook(name, module, *grad_output, **kwargs):
-                        for _ in range(len(module._peft_gradient_checkpointing_forward_hooks)):
+                        while module._peft_gradient_checkpointing_forward_hooks:
                             module._peft_gradient_checkpointing_forward_hooks.pop().remove()
 
                     if getattr(layer, "_peft_gradient_checkpointing_forward_hooks", []):
--- a/tests/testing_common.py
+++ b/tests/testing_common.py
@@ -25,6 +25,7 @@ from operator import attrgetter
 
 import pytest
 import torch
+import transformers
 import yaml
 from diffusers import StableDiffusionPipeline
 from packaging import version
@@ -1327,9 +1328,10 @@ class PeftCommonTester:
             # TODO: no gradients on the "dense" layer, other layers work, not sure why
             self.skipTest("AdaLora with RoBERTa does not work correctly")
 
-        if "bart" in model_id.lower():
-            # TODO: no backprop possible with Bart, not sure why
-            self.skipTest("Bart does not work correctly")
+        if "bart" in model_id.lower() and version.parse(transformers.__version__) <= version.parse("5.0"):
+            self.skipTest(
+                "Bart in transformers < 5.0 doesn't handle input sharing well enough. See transformers#41821"
+            )
 
         if (config_cls == OFTConfig) and ("deberta" in model_id.lower()):
             # TODO: no gradients on the "dense" layer, other layers work, not sure why
@@ -1344,9 +1346,10 @@ class PeftCommonTester:
             if not getattr(model, "supports_gradient_checkpointing", False):
                 return pytest.skip(f"Model {model_id} does not support gradient checkpointing")
 
-            # Disable lora_dropout to remove non-determinism in gradient creation
-            if "lora_dropout" in config_kwargs:
-                del config_kwargs["lora_dropout"]
+            # Disable lora_dropout and friends to remove non-determinism in gradient creation
+            for key in list(config_kwargs.keys()):
+                if key.endswith("dropout"):
+                    del config_kwargs[key]
 
             config = config_cls(
                 base_model_name_or_path=model_id,


src/peft/tuners/lora/model.py
```diff
@@ -357,7 +358,39 @@ def _enable_peft_forward_hooks(self, *args, **kwargs):
...
+                    def backward_hook(name, module, *grad_output, **kwargs):
+                        for _ in range(len(module._peft_gradient_checkpointing_forward_hooks)):
+                            module._peft_gradient_checkpointing_forward_hooks.pop().remove()
+
+                    if getattr(layer, "_peft_gradient_checkpointing_forward_hooks", []):
+                        raise ValueError(
+                            "Multiple invocations of PEFT forward hooks before .backward() with enabled gradient "
+                            "checkpointing. Disable gradient checkpointing or only call forward once per backward."
+                        )
+                    layer._peft_gradient_checkpointing_forward_hooks = []
```
BenjaminBossan, line 392: Should we check for pre-existing `_peft_gradient_checkpointing_forward_hooks`? Right now, we know they don't exist, but since this is supposed to be a general solution (not aLoRA-specific), I'd say it's safer to check.

src/peft/tuners/lora/model.py
```diff
@@ -357,7 +358,39 @@ def _enable_peft_forward_hooks(self, *args, **kwargs):
...
+                        for submodule in module.modules():
+                            if isinstance(submodule, LoraLayer):
+                                handle = submodule.register_forward_pre_hook(
+                                    partial(_alora_offsets_pre_forward_hook, alora_offsets=kwargs["alora_offsets"]),
+                                    with_kwargs=True,
+                                )
+                                module._peft_gradient_checkpointing_forward_hooks.append(handle)
+
+                    def backward_hook(name, module, *grad_output, **kwargs):
+                        for _ in range(len(module._peft_gradient_checkpointing_forward_hooks)):
```
BenjaminBossan: Would `while module._peft_gradient_checkpointing_forward_hooks:` be a bit easier to understand?

src/peft/tuners/lora/model.py
```diff
@@ -357,7 +358,39 @@ def _enable_peft_forward_hooks(self, *args, **kwargs):
...
+                # forward call as well.
+                #
+                # Note that this will lead to double application of whatever the callbacks do in normal forward.
+                # Make sure that whatever change is done, can be applied more than once without harm (idempotency).
+                if isinstance(layer, GradientCheckpointingLayer) and layer.gradient_checkpointing:
+
+                    def forward_pre_hook(name, module, inputs, **kwargs):
+                        for submodule in module.modules():
+                            if isinstance(submodule, LoraLayer):
+                                handle = submodule.register_forward_pre_hook(
```
BenjaminBossan, line 377: This part is not obvious to me. We register `forward_pre_hook` as a forward pre hook and this function itself registers more forward pre hooks. Is this just a convenient way to recurse? Could we not make this explicit by looping over the submodules in https://github.com/huggingface/peft/pull/2860/files#diff-4331e1b00e557c2af6e682bc221d1517f554d7ca24b9149520237b431d8d797fR390 and directly register the final hooks?

src/peft/tuners/lora/model.py
BenjaminBossan, line 1: The `_enable_peft_forward_hooks` is becoming quite complex at this point. It could be worth it to refactor it into the mixed batch part and into the aLoRA part. Not necessarily in this PR, but could be done later.

tests/testing_common.py
```diff
@@ -1315,41 +1315,76 @@ def _test_training_layer_indexing(self, model_id, config_cls, config_kwargs):
...
+
         if config_cls == PrefixTuningConfig:
             return pytest.skip(f"Test not applicable for {config_cls}")
 
         if (config_cls == AdaLoraConfig) and ("roberta" in model_id.lower()):
             # TODO: no gradients on the "dense" layer, other layers work, not sure why
             self.skipTest("AdaLora with RoBERTa does not work correctly")
 
+        if "bart" in model_id.lower():
+            # TODO: no backprop possible with Bart, not sure why
```
BenjaminBossan: Comment can be updated, right?

tests/testing_common.py
```diff
@@ -1315,41 +1315,76 @@ def _test_training_layer_indexing(self, model_id, config_cls, config_kwargs):
...
+
         with hub_online_once(model_id):
             model = self.transformers_class.from_pretrained(model_id)
 
             if not getattr(model, "supports_gradient_checkpointing", False):
                 return pytest.skip(f"Model {model_id} does not support gradient checkpointing")
 
-            model.gradient_checkpointing_enable()
+            # Disable lora_dropout to remove non-determinism in gradient creation
+            if "lora_dropout" in config_kwargs:
```
BenjaminBossan: Wondering about non-LoRA PEFT methods.

BenjaminBossan (changes requested):
Thanks for working on this complex issue and coming up with a solution. Generally, it LGTM, I just have a few comments, please check.

It would also be great if @@kgreenewald could try if this branch resolves the original issue.

kgreenewald:
Awesome, thanks so much @githubnemo ! Sorry for the delay here - one question I have is the original DPO use case that led me to discover this issue. I'm still in the midst of a job change so I don't have access to test right now, but DPO has 2 forward passes (one the reference model, one the model being trained) prior to the backward. In particular, if you look at https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L919 you can see that the reference model is just the PEFT model with .disable_adapter(). I see you mention this implementation "disallows multiple forward calls in succession" - I don't see this explicitly being enforced & I think the implementation you have would work here since the alora_offsets are the same for these 2 forward calls (and not used in the reference model at all). Just want to make sure I didn't miss some logic that explicitly forbids this type of pattern.

src/peft/tuners/lora/model.py
githubnemo: I think we can keep it as is and once there are more methods that require access we can move to method-specific entries in a dictionary instead of a global list.

tests/testing_common.py
githubnemo: Yeah, me too but I've not encountered any issues/inconsistencies so far so I'd opt for updating as we go.

githubnemo:
> Awesome, thanks so much @githubnemo ! Sorry for the delay here - one question I have is the original DPO use case that led me to discover this issue. I'm still in the midst of a job change so I don't have access to test right now, but DPO has 2 forward passes (one the reference model, one the model being trained) prior to the backward. In particular, if you look at https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L919 you can see that the reference model is just the PEFT model with .disable_adapter(). I see you mention this implementation "disallows multiple forward calls in succession" - I don't see this explicitly being enforced & I think the implementation you have would work here since the alora_offsets are the same for these 2 forward calls (and not used in the reference model at all). Just want to make sure I didn't miss some logic that explicitly forbids this type of pattern.

Yeah, that seems to be a problem. We've pushed a change that disables the registration of forward hooks when no adapter is active. Can you give it a test to see if that works?

kgreenewald:
> > Awesome, thanks so much @githubnemo ! Sorry for the delay here - one question I have is the original DPO use case that led me to discover this issue. I'm still in the midst of a job change so I don't have access to test right now, but DPO has 2 forward passes (one the reference model, one the model being trained) prior to the backward. In particular, if you look at https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L919 you can see that the reference model is just the PEFT model with .disable_adapter(). I see you mention this implementation "disallows multiple forward calls in succession" - I don't see this explicitly being enforced & I think the implementation you have would work here since the alora_offsets are the same for these 2 forward calls (and not used in the reference model at all). Just want to make sure I didn't miss some logic that explicitly forbids this type of pattern.
> 
> Yeah, that seems to be a problem. We've pushed a change that disables the registration of forward hooks when no adapter is active. Can you give it a test to see if that works?

Great! I verified that the DPO gradients are nonzero, so should be good to go.

githubnemo:
@kgreenewald great, thanks for testing!

PR #2862, opened 2025-10-23
use `max_length` to replace `max_seq_length`; correct README for

kaixuanliu:
delora_finetuning

### DIFF
--- a/examples/sft/train.py
+++ b/examples/sft/train.py
@@ -18,7 +18,7 @@ class ModelArguments:
     model_name_or_path: str = field(
         metadata={"help": "Path to pretrained model or model identifier from huggingface.co/models"}
     )
-    max_length: Optional[int] = field(
+    max_seq_length: Optional[int] = field(
         default=512,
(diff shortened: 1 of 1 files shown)


kaixuanliu:
@BenjaminBossan recent version trl/transformers use `max_length` instead if `max_seq_length` param, pls review this PR, thx!

examples/sft/train.py
```diff
@@ -18,7 +18,7 @@ class ModelArguments:
     model_name_or_path: str = field(
         metadata={"help": "Path to pretrained model or model identifier from huggingface.co/models"}
     )
-    max_seq_length: Optional[int] = field(
+    max_length: Optional[int] = field(
```
BenjaminBossan: This change isn't enough, because there is a conflicting argument with the same name. I already took care of that in #2857, so you can revert this change here.

kaixuanliu: Oops, I changed the code here by mistake. Yes, for sft case, we should take more actions. Have reverted.

PR #2605, opened 2025-06-23
FIX: Trainable tokens error with DeepSpeed ZeRO3

BenjaminBossan:
Resolves #2603

Trainable tokens are erroring when using DS Z3 because the embedding weights are not available on all ranks. This solution fixes this in an efficient way that collects these weights on a single rank, initializes them, and then broadcasts only the slice that is affected.

PR #2863, opened 2025-10-23
Fix `trainable_token_indices` for `lm_head`

aflueckiger:
It looks like that `trainable_token_indices` has been broken since #2605 for the `lm_head` in case the weights are not tied. The `lm_head` is an instance of `Linear` rather than `Embedding`, and, thus, it doesn't have an attribute `embedding_dim`.

The embedding dimension is not even needed further down unless when training with `deepspeed` or re-initializing the weights.

Asking @BenjaminBossan for a review due to the changes in #2605. I couldn't find any reports concerning that issue. Let me know if you would like to fix it differently.

Minimal config for reproduction:
```
model_name = "utter-project/EuroLLM-9B-Instruct"

trainable_tokens_indices = [5, 6, 7, 8, 9]

trainable_tokens = {"lm_head": trainable_tokens_indices, "embed_tokens": trainable_tokens_indices}

# Before this PR: Only training the embedding layer was possible
# trainable_tokens = {"embed_tokens": trainable_tokens_indices}
# trainable_tokens = trainable_tokens_indices

LoraConfig(
    lora_alpha=16,
    lora_dropout=0.05,
    r=16,
    bias="none",
    task_type="CAUSAL_LM",
    target_modules=[
        "q_proj",
        "k_proj",
        "v_proj",
        "o_proj",
        "gate_proj",
        "up_proj",
        "down_proj",
    ],
    modules_to_save=None,
    trainable_token_indices=trainable_tokens,
)

```

### DIFF
--- a/tests/test_initialization.py
+++ b/tests/test_initialization.py
@@ -71,7 +71,7 @@ from peft.utils import infer_device
 from peft.utils.hotswap import hotswap_adapter, prepare_model_for_compiled_hotswap
 from peft.utils.other import ModulesToSaveWrapper
 
-from .testing_utils import load_dataset_english_quotes, require_deterministic_for_xpu
+from .testing_utils import hub_online_once, load_dataset_english_quotes, require_deterministic_for_xpu
 
 
 try:
@@ -1144,6 +1144,14 @@ class TestLoraInitialization:
         )
         with pytest.raises(ValueError, match=msg):
             get_peft_model(model, config, adapter_name="foobar")
+
+    def test_trainable_token_indices_targets_head_and_embedding(self):
+        # targeting embedding and LM head explicitly, see #2863
+        model_id = "hf-internal-testing/tiny-random-OPTForCausalLM"
+        with hub_online_once(model_id):
+            model = AutoModelForCausalLM.from_pretrained(model_id)
+            config = LoraConfig(trainable_token_indices={"lm_head": [0], "embed_tokens": [0]})
+            get_peft_model(model, config)  # does not raise
 
     @require_deterministic_for_xpu
     def test_lora_use_dora_linear(self, data):


src/peft/tuners/trainable_tokens/layer.py
```diff
@@ -120,7 +120,12 @@ def update_layer(self, adapter_name, **kwargs):
         # onto the new values, we would get undefined behavior. By replacing the specific token values we always
         # get defined behavior.
         weight = self.get_base_layer().weight
-        embed_dim = self.get_base_layer().embedding_dim
+
+        try:
```
BenjaminBossan: Let's use `if hasattr` (or nested `getattr`) instead of `try ... except`.

BenjaminBossan (changes requested):
Thanks for providing this fix.

For the purpose of testing, could you please give a small example where this is broken? The model ID and PEFT config should be enough.

> The embedding dimension is not even needed further down unless when training with deepspeed or re-initializing the weights.

It's also used when `init_weights=False`. I guess we could move it into the corresponding branches to avoid retrieving this argument when it's not needed, but I think it's better to make the code robust enough that it does not break.

aflueckiger:
Thanks for the quick response. I added more information to reproduce the error in the PR description and replaced the try-expect block. Let me know about the preferred position for handling the attribute.

BenjaminBossan:
Thanks for providing the reproducer. I can confirm that it raises the error and it helped me narrow down why the existing tests didn't catch this. We have a gap where we don't test specifying `embed_tokens` and `lm_head` both. To fill the gap, could you please add a unit test? Here is a suggestion:

```python
    def test_trainable_token_indices_targets_head_and_embedding(self):
        # targeting embedding and LM head explicitly, see #2863
            config = LoraConfig(trainable_token_indices={"lm_head": [0], "embed_tokens": [0]})
            get_peft_model(model, config)  # does not raise
```

A good spot for this test would be here: https://github.com/huggingface/peft/blob/a18ba67f242ab2eb74cdabab76ea2fd836b5cd83/tests/test_initialization.py#L1147

aflueckiger:
Thanks for providing the right spot for the test. It is added now.

PR #2871, opened 2025-10-26
Update LoRA developer guides: non-in-place operations

DargorAbraxas:
The documentation of the [merge_and_unload()](https://huggingface.co/docs/peft/main/en/package_reference/tuners#peft.tuners.tuners_utils.BaseTuner.merge_and_unload) function clarifies that this is not an in-place operation. However, the [developer guides code snippets](https://huggingface.co/docs/peft/main/en/developer_guides/lora#merge-lora-weights-into-the-base-model) shows
``` py
from peft import PeftModel

base_model = AutoModelForCausalLM.from_pretrained("mistralai/Mistral-7B-v0.1")
peft_model_id = "alignment-handbook/zephyr-7b-sft-lora"
model = PeftModel.from_pretrained(base_model, peft_model_id)
model.merge_and_unload()
```
which do not match the function documentation. The same happens with the [unload()](https://huggingface.co/docs/peft/main/en/package_reference/tuners#peft.tuners.tuners_utils.BaseTuner.unload) function.

This PR just changes it to be consistent and minimize silent mistakes when using the guides!

### DIFF
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -511,7 +511,7 @@ from peft import PeftModel
 base_model = AutoModelForCausalLM.from_pretrained("mistralai/Mistral-7B-v0.1")
 peft_model_id = "alignment-handbook/zephyr-7b-sft-lora"
 model = PeftModel.from_pretrained(base_model, peft_model_id)
-merged_model = model.merge_and_unload()
+model = model.merge_and_unload()
 ```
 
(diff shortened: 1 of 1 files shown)


docs/source/developer_guides/lora.md
```diff
@@ -511,10 +511,10 @@ from peft import PeftModel
 base_model = AutoModelForCausalLM.from_pretrained("mistralai/Mistral-7B-v0.1")
 peft_model_id = "alignment-handbook/zephyr-7b-sft-lora"
 model = PeftModel.from_pretrained(base_model, peft_model_id)
-model.merge_and_unload()
+merged_model = model.merge_and_unload()
```
BenjaminBossan: Let's assign to the same variable. This is because after calling `merge_and_unload`, the original `model` shouldn't  be used anymore.

```suggestion
model = model.merge_and_unload()
```

DargorAbraxas: Thanks for the comment, @BenjaminBossan, I made the change you suggested and uploaded it. Let me know if you think there is anything else that should change!

PR #2875, opened 2025-10-28
Post rc-release version bump

githubnemo:
We'll just add a dev0 to the rc release since the full release is on its way anyway.


BenjaminBossan (approved):
Thanks.

PR #2879, opened 2025-10-29
ENH: Tie weights for target_modules in Lora (#2864)

romitjain:
Solves #2864 for `target_modules`

Enables `ensure_weight_tying` flag in `LoraConfig` for `target_modules`.

For LoRA, if any of the tied layers are added to `target_modules` and `ensure_weight_tying == True`, the adapters added to the layer are shared with all the tied layers.

For example, if a model has tied weights and `target_modules=['embed_tokens']` then, LoRA adapters are added to both `embed_tokens` and `lm_head`. The adapters in `lm_head` share the weights with the adapters added to `embed_tokens`

### DIFF
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -887,11 +887,8 @@ class LoraModel(BaseTuner):
 
         modules_to_save = getattr(peft_config, "modules_to_save", []) or []
         embed_layer_name = find_parameter_name_by_module(self.model, self.model.get_input_embeddings())
-        prefix, sep, suffix = embed_layer_name.partition(".")
-        if sep and "model" in prefix:
-            embed_layer_name = suffix
 
-        # the name suffix may already be included but we assume that adding the same module twice is not a problem
+        # the layer may already be included but we assume that adding the same module twice is not a problem
         if embed_layer_name not in modules_to_save:
             modules_to_save.append(embed_layer_name)
 
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -1233,7 +1233,9 @@ class BaseTuner(nn.Module, ABC):
         `peft_config` in place with any layers/adapters that needs to be tied
         """
         modules_to_save = set(getattr(peft_config, "modules_to_save", []) or [])
-        is_embedding_to_save = any(m in EMBEDDING_LAYER_NAMES for m in modules_to_save)
+        # `EMBEDDING_LAYER_NAMES` contains only the stripped name of the module
+        # eg: To get a match for model.embed_tokens, we need to extract `embed_tokens`
+        is_embedding_to_save = any(m.split(".")[-1] in EMBEDDING_LAYER_NAMES for m in modules_to_save)
 
         raw_target_modules = getattr(peft_config, "target_modules", None)
         if isinstance(raw_target_modules, str):
@@ -1242,7 +1244,9 @@ class BaseTuner(nn.Module, ABC):
             )
         else:
             target_modules = set(raw_target_modules or [])
-            is_embedding_in_target = any(m in EMBEDDING_LAYER_NAMES for m in target_modules)
+            # `EMBEDDING_LAYER_NAMES` contains only the stripped name of the module
+            # eg: To get a match for model.embed_tokens, we need to extract `embed_tokens`
+            is_embedding_in_target = any(m.split(".")[-1] in EMBEDDING_LAYER_NAMES for m in target_modules)
 
         tied_weight_keys = self._get_module_names_tied_with_embedding()
 


romitjain:
@BenjaminBossan 
I have added the relevant test cases and implemented the `ensure_weight_tying` flag for `target_modules`. The current implementation works only if `embed_tokens` is added and not if `lm_head` is added. I will implement that fix and update the PR, but meanwhile would appreciate your views on the logic and implementation.

At a high level
1. I have updated `BaseTuner._check_tied_modules` to check for tied modules in `target_modules`
2. I have added a private method `BaseTuner._add_targets_to_tie` that needs to be implemented by the inheriting classes
3. I have added a loop in `BaseTuner.inject_adapter` to tie the adapters. I have implemented this extra loop to ensure that the order in which adapters are added to the target modules do not matter.

Thank you

src/peft/tuners/lora/config.py
```diff
@@ -382,6 +382,11 @@ class LoraConfig(PeftConfig):
             `target_parameters`. As an example, for Llama4, you can pass:
             `target_parameters=['feed_forward.experts.gate_up_proj', 'feed_forward.experts.down_proj]`. Passing a
             string for regex matching is not implemented yet.
+        ensure_weight_tying (`bool`, *optional*)
```
BenjaminBossan, line 458: Thanks for adding.

src/peft/tuners/lora/model.py
```diff
@@ -811,3 +826,11 @@ def _add_modules_to_tie(self, peft_config, tied_weight_keys):
         missing_keys = set(tied_weight_keys) - modules_to_save
 
         peft_config.modules_to_tie = missing_keys
+
+    def _add_targets_to_tie(self, peft_config, tied_weight_keys):
+        target_modules = set(getattr(peft_config, "target_modules", []) or [])
```
BenjaminBossan: We need to consider the case that `target_modules` is a string and not a list of strings. If it's a string, we perform a regex match. Honestly, I'm not sure if there is a good solution. So far, I have 3 ideas:

1. We could try to use the `model.targeted_module_names` attribute, which lists all targeted modules after the targets have been resolved. But that would mean that we need to first apply all LoRA layers and only then can we check for tied layers, which is the opposite order of how things are implemented right now.
2. We could try using the string directly and then for example do something like: `config.target_modules += f"|{missing_key}"` but this is very brittle and won't work with all regexes, so I would like to avoid this.
3. We could forbid using `ensure_weight_tying=True` and `target_modules = <str>`. Then we'd raise an error and tell users they have to pass a list of str if they want `ensure_weight_tying`.

src/peft/tuners/tuners_utils.py
```diff
@@ -824,6 +827,47 @@ def inject_adapter(
                 peft_config=peft_config, model=model, adapter_name=adapter_name, low_cpu_mem_usage=low_cpu_mem_usage
             )
 
+        # Another loop for tying target modules
+        for key, module in named_modules:
```
BenjaminBossan: This loop is almost identical to the one above and it would add overhead. At the very least, we could first check if there is any `target_module_to_tie` and if not, skip the whole loop. Better yet though, could this not be integrated in the loop above?

BenjaminBossan (commented):
Thanks for this draft PR to extend the feature to `target_modules`. I haven't done a full review yet, as some implementation details have yet to be figured out, but I gave some early feedback. This feature could be a bit more difficult to implement than for `modules_to_save`, I added some comments on why, please check.

src/peft/tuners/lora/model.py
romitjain: Yes, after going through the code a few more times, I realized this would not work for all the cases. I would go with the 1st approach and move the call to this function after `model.targeted_module_names` is updated

src/peft/tuners/tuners_utils.py
romitjain: Since the user can add either `lm_head` or `embed_tokens`, it would necessitate another loop to tie the adapters. But yes, this loop can be skipped if there are no adapters to tie

src/peft/tuners/tuners_utils.py
```diff
@@ -824,6 +855,15 @@ def inject_adapter(
                 peft_config=peft_config, model=model, adapter_name=adapter_name, low_cpu_mem_usage=low_cpu_mem_usage
             )
 
+        # Another loop for tying target modules
```
romitjain: The loop is much shorter and only used if needed

src/peft/tuners/tuners_utils.py
```diff
@@ -910,15 +950,6 @@ def inject_adapter(
                     RuntimeWarning,
                 )
 
-        tied_target_modules = self._get_tied_target_modules(model=model)
```
romitjain, line 921: This is removed since this is no longer needed. We have fixed this warning

romitjain:
@BenjaminBossan This is now ready for review. I have also updated the logic for tied layers in `modules_to_save` so that `lm_head` and `[embed_tokens, lm_head]` cases are supported. Earlier, they would not have worked. The high level implementation remains the same but according to me it's much better placed then my earlier commits.

I have also added a few tests for the above case, and all of the tests pass.

The only thing remaining is how to check for `target_modules` in case it's a string. I will come back to it, but you can go ahead and review the core logic.

src/peft/tuners/lora/layer.py
```diff
@@ -156,6 +156,8 @@ def update_layer(
         arrow_config: ArrowConfig = None,
         qalora_group_size: int = 32,
         inference_mode: bool = False,
+        is_tied: bool = False,
+        tied_adapters: dict = {},
```
```suggestion
        tied_adapters: Optional[dict[str, nn.Parameter]] = None,
```

Let's not use mutable default values.

src/peft/tuners/lora/layer.py
```diff
@@ -195,6 +197,16 @@ def update_layer(
         # Actual trainable parameters
         self.lora_A[adapter_name] = nn.Linear(self.in_features, r, bias=False)
         self.lora_B[adapter_name] = nn.Linear(r, self.out_features, bias=lora_bias)
+        if is_tied:
```
BenjaminBossan: Do we really need the `is_tied` argument? We could just check `if tied_adapters` instead.

src/peft/tuners/lora/layer.py
```diff
@@ -631,6 +643,8 @@ def __init__(
             use_alora=use_alora,
             lora_bias=lora_bias,
             arrow_config=arrow_config,
+            is_tied=kwargs.get("is_tied", False),
+            tied_adapters=kwargs.get("tied_adapters"),
```
BenjaminBossan: Instead of `get`ting these from `kwargs`, let's add the arguments to the function signature explicitly (although for `is_tied`, as mentioned earlier, I think it can be removed).

src/peft/tuners/lora/layer.py
```diff
@@ -195,6 +197,16 @@ def update_layer(
...
         self.lora_B[adapter_name] = nn.Linear(r, self.out_features, bias=lora_bias)
+        if is_tied:
+            if not tied_adapters:
+                raise RuntimeError("Layer is marked as tied, but tied adapters are not provided")
+
+            lora_A_params = tied_adapters["lora_A"]
+            lora_B_params = tied_adapters["lora_B"]
+
+            self.lora_A[adapter_name].weight = torch.nn.Parameter(lora_A_params)
+            self.lora_B[adapter_name].weight = torch.nn.Parameter(lora_B_params)
```
BenjaminBossan, line 203: Let's add a comment here that weight tying is only implemented for `Embedding` => `Linear`, which is why this code is not necessary to be added to the other layer types (say, the conv layers).

src/peft/tuners/lora/model.py
```diff
@@ -187,6 +187,15 @@ def _create_and_replace(
         r = lora_config.rank_pattern.get(r_key, lora_config.r)
         alpha = lora_config.alpha_pattern.get(alpha_key, lora_config.lora_alpha)
 
+        is_tied = target_name in (getattr(lora_config, "target_modules_to_tie", []) or [])
```
BenjaminBossan, line 210: Let's add a short comment here explaining what happens.

src/peft/tuners/tuners_utils.py
```diff
@@ -792,6 +817,9 @@ def inject_adapter(
                 elif not result:
                     unmatched_modules.append(key)
                 else:
+                    if self._check_tied_module_exists(peft_config, key):
+                        targets_to_tie.append(key)
+                        continue
```
BenjaminBossan: Let's add a comment here why this is skipped for now.

src/peft/tuners/tuners_utils.py
```diff
@@ -805,6 +833,9 @@ def inject_adapter(
                 if key not in module_names:
                     unmatched_modules.append(key)
                 else:
+                    if self._check_tied_module_exists(peft_config, key):
```
BenjaminBossan, line 855: Let's add a comment here why this is skipped for now.

src/peft/tuners/tuners_utils.py
BenjaminBossan: Let's expand on the comment by explaining why these modules are only being added _after_ the other modules were targeted.

tests/test_initialization.py
```diff
@@ -4839,69 +4839,71 @@ def get_input_embeddings(self):
 
         return CausalLM().eval().to(self.torch_device)
 
-    def test_weight_tying_tied_model_lora(self):
+    @pytest.mark.parametrize("layer", ["lm_head", "embed_tokens", ["lm_head", "embed_tokens"]])
```
```suggestion
    @pytest.mark.parametrize("modules_to_save", [["lm_head"], ["embed_tokens"], ["lm_head", "embed_tokens"]])
```

Let's call this `modules_to_save` to make it immediately clear what is meant and also put everything into a list, so that we don't need to check `if isinstance(layer, list)` below. Same for the other tests.

BenjaminBossan (changes requested):
Thanks for the latest updates. This looks much cleaner now, I think we're approaching the finish line. There were still some issues I had though, so please check my comments.

src/peft/tuners/lora/layer.py
romitjain: Added

romitjain:
@BenjaminBossan I have addressed your comments. PTAL

src/peft/peft_model.py
```diff
@@ -315,6 +315,11 @@ def save_mutated_as_lora(peft_config, path_initial_model_for_weight_conversion,
                     output_state_dict = save_mutated_as_lora(
                         peft_config, path_initial_model_for_weight_conversion, output_state_dict, kwargs
                     )
+
+                for k, v in output_state_dict.items():
```
BenjaminBossan, line 337: Why was this added?

src/peft/tuners/lora/model.py
```diff
@@ -249,7 +261,8 @@ def _create_and_replace(
             if adapter_name not in self.active_adapters:
                 # adding an additional adapter: it is not automatically trainable
                 new_module.requires_grad_(False)
-            self._replace_module(parent, target_name, new_module, target)
+
+            self._replace_module(parent=parent, child_name=target_name, new_module=new_module, child=target)
```
BenjaminBossan: Although I'm not against using keywords here, this method is called numerous times in PEFT with just positional args, so let's leave it at that for consistency.

src/peft/tuners/lora/model.py
```diff
@@ -806,8 +819,45 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+        """
+        Tied weight keys contains the layers tied to the embedding layer. Add embedding layer and remove rest of the
+        tied layers from `module_to_save`. Maintain a separate set for layers to be tied
+
+        Args:
+            peft_config (LoraConfig)
+            tied_weight_keys (list[str])
+        """
+        tied_weight_keys = set(tied_weight_keys)
+        setattr(peft_config, "modules_to_tie", tied_weight_keys)
```
BenjaminBossan:
```suggestion
        peft_config.modules_to_tie = tied_weight_keys
```

src/peft/tuners/lora/model.py
```diff
@@ -806,8 +819,45 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+
+        modules_to_save = getattr(peft_config, "modules_to_save", []) or []
+        if "embed_tokens" not in modules_to_save:
+            modules_to_save.append("embed_tokens")
+
+        for m in tied_weight_keys:
+            if m in modules_to_save:
+                modules_to_save.remove(m)
+
+        setattr(peft_config, "modules_to_save", modules_to_save)
```
BenjaminBossan:
```suggestion
        peft_config.modules_to_save = modules_to_save
```

src/peft/tuners/lora/model.py
```diff
@@ -806,8 +819,45 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+        """
+        Tied weight keys contains the layers tied to the embedding layer. Add embedding layer and remove rest of the
+        tied layers from `target_modules`. Maintain a separate set for layers to be tied
+
+        Args:
+            peft_config (LoraConfig)
+            tied_weight_keys (list[str])
+        """
+        tied_weight_keys = set(tied_weight_keys)
+        setattr(peft_config, "target_modules_to_tie", tied_weight_keys)
```
BenjaminBossan:
```suggestion
        peft_config.target_modules_to_tie = tied_weight_keys
```

src/peft/tuners/lora/model.py
BenjaminBossan: Sounds good. This has yet to be updated, right?

src/peft/tuners/lora/model.py
```diff
@@ -806,8 +819,45 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+
+        target_modules = set(getattr(peft_config, "target_modules", []) or [])
+        target_modules.add("embed_tokens")
+
+        for m in tied_weight_keys:
+            if m in target_modules:
+                target_modules.remove(m)
 
-        peft_config.modules_to_tie = missing_keys
+        setattr(peft_config, "target_modules", target_modules)
```
BenjaminBossan:
```suggestion
        peft_config.target_modules = target_modules
```

src/peft/tuners/lora/model.py
romitjain: Marking this as resolved since I have added the comments above

src/peft/tuners/lora/model.py
romitjain: Sure, I think I updated it when I was adding an argument here, but makes sense to keep it untouched

src/peft/tuners/lora/model.py
romitjain: Yes, I will do this in the next commit

src/peft/peft_model.py
romitjain: When we add the reference from Linear Lora to Embedding Lora adapter (`lora_A = lora_embedding_B.T`), the transpose operation creates a view of the same tensor. That tensor is not contiguous in memory, and state dict export is not supported for non-contiguous tensors.

Hence, this applied only during the save operation of the model.

src/peft/tuners/lora/model.py
romitjain: @BenjaminBossan 

Moving this after `model.targeted_modules_name` is populated is tough, as the loop which populates this (https://github.com/huggingface/peft/blob/main/src/peft/tuners/tuners_utils.py#L773-L819) also needs to check and skip if the layers are tied. 

Reversing the order would mean that we may end up adding adapters where they're not required. The subsequent code would become more involved, but essentially, we would have to remove adapters from all tied layers, re-add in `embed_tokens`, and proceed to tie remaining adapters to this. This is an opinionated solve which has the least complexity, according to me.

We can go with (1) in your original comment and redo a few things, or keep the current flow and go with (3).

I think the above might have become tough to follow 😅, so let me know and I can share some schematics. Will wait for your input.

src/peft/tuners/lora/model.py
```diff
@@ -806,8 +819,56 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+        tied layers from `target_modules`. Maintain a separate set for layers to be tied
+
+        Args:
+            peft_config (LoraConfig)
+            tied_weight_keys (list[str])
+        """
+        tied_weight_keys = set(tied_weight_keys)
+        peft_config.target_modules_to_tie = tied_weight_keys
+
+        raw_target_modules = getattr(peft_config, "target_modules", None)
```
romitjain, line 923: @BenjaminBossan Please review this logic. I know this is a bit hacky! I am open to suggestions

tests/test_initialization.py
```diff
@@ -4935,6 +4937,130 @@ def test_weight_tying_tied_model_lokr(self):
...
+
+        lm_lora_A = model.base_model.model.lm_head.lora_A[adapter_name].weight
+        lm_lora_B = model.base_model.model.lm_head.lora_B[adapter_name].weight
+
+        assert torch.allclose(embed_lora_A, lm_lora_B.T)
+        assert torch.allclose(embed_lora_B, lm_lora_A.T)
+        assert embed_lora_A.data_ptr() == lm_lora_B.data_ptr()
+        assert embed_lora_B.data_ptr() == lm_lora_A.data_ptr()
+
+    @pytest.mark.parametrize("layer", [".*embed_tokens$", ".*lm_head$", ".*(embed_tokens|lm_head)$"])
```
romitjain: I have added a new test case for testing target modules as string inputs

romitjain:
@BenjaminBossan please review the latest changes now. I believe I have addressed all your comments, but let me know if I missed something.

I have added test cases where we are passing `target_modules` as `str` and added a (slightly) hacky solve for that. It's opinionated to keep the flow simple. 

Regarding the transformers v5 update, since we would be having a version locked in peft, I believe if this PR advances faster than that, we can merge this. I can take up changes too whenever they're needed. However, you are much closer to this, so you can decide and let me know.

src/peft/peft_model.py
BenjaminBossan: Could we make the tensor contiguous right there and remove the call here?

src/peft/tuners/lora/model.py
```diff
@@ -806,8 +819,56 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+
+        Args:
+            peft_config (LoraConfig)
+            tied_weight_keys (list[str])
+        """
+        tied_weight_keys = set(tied_weight_keys)
+        peft_config.modules_to_tie = tied_weight_keys
+
+        modules_to_save = getattr(peft_config, "modules_to_save", []) or []
+        if "embed_tokens" not in modules_to_save:
```
BenjaminBossan: If the embedding layer has a different name, this won't be correct, right? It's probably still fine for now.

src/peft/tuners/tuners_utils.py
```diff
@@ -1210,27 +1268,39 @@ def _add_modules_to_tie(self, peft_config, tied_weight_keys):
...
+        else:
+            target_modules = set(raw_target_modules or [])
+            is_embedding_in_target = any(m in EMBEDDING_LAYER_NAMES for m in target_modules)
+
         tied_weight_keys = self._get_tied_weight_keys(model)
 
         if getattr(peft_config, "ensure_weight_tying", False):
-            if is_embedding_to_save and tied_weight_keys:
-                self._add_modules_to_tie(peft_config, tied_weight_keys)
+            if (is_embedding_to_save or is_embedding_in_target) and tied_weight_keys:
```
BenjaminBossan: I think this whole block, line 1288-1298, can be replaced with:

```python
        if getattr(peft_config, "ensure_weight_tying", False):
            if tied_weight_keys:
                if is_embedding_to_save:
                    self._add_modules_to_tie(peft_config, tied_weight_keys)
                elif is_embedding_in_target:
                    self._add_targets_to_tie(peft_config, tied_weight_keys)
                else:
                    warnings.warn(
                        "You have requested `ensure_weight_tying`, but no tied modules are added in either "
                        "`modules_to_save` or `target_modules`"
                    )
            else:
                warnings.warn("You have requested `ensure_weight_tying`, but no tied modules were found in the model")
```

I think this is cleaner.

src/peft/tuners/tuners_utils.py
```diff
@@ -363,6 +363,30 @@ def _prepare_model(self, peft_config: PeftConfig, model: nn.Module):
...
+                True if key matches any tied modules from config, False if no match found.
+        """
+        _target_modules_to_tie = getattr(peft_config, "target_modules_to_tie", {}) or {}
+
+        if key in _target_modules_to_tie or any(
+            key.endswith(f".{target_key}") for target_key in _target_modules_to_tie
+        ):
+            return True
+
+        return False
```
```suggestion
        target_modules_to_tie = getattr(peft_config, "target_modules_to_tie", []) or []
        return key in target_modules_to_tie or any(
            key.endswith(f".{target_key}") for target_key in target_modules_to_tie
        )
```

Reason:
1. More concise
2. Doesn't really make sense to use dicts (`{}`) as default. If you meant to use the empty set, I don't think it amortizes here.

src/peft/tuners/tuners_utils.py
```diff
@@ -1198,6 +1238,24 @@ def _add_modules_to_tie(self, peft_config, tied_weight_keys):
         """
         This method adds modules to tie to `peft_config` so that those modules can be tied downstream. By default this
         method raises a warning, and each tuner class extending `BaseTuner` can choose to implement this.
+
+        Check `peft.tuners.lora.LoraModel._add_modules_to_tie` for an example.
+        """
+        msg = (
+            "Model has `tie_word_embeddings=True` and a tied layer is part of the adapter, "
```
BenjaminBossan: Since this message is now duplicated, let's make it a constant and re-use it.

src/peft/tuners/lora/model.py
BenjaminBossan: Hmm yeah, this is rough. We can't really operate on the string like this, as there are too many possible ways that the regex could be formed. I wonder if we should just leave it be and deal with the tied module edge case in `inject_adapter` directly. I haven't fully thought this through, perhaps you already tried that and there is a caveat that I'm missing?

src/peft/peft_model.py
romitjain: Converting the transpose view of the tensor to contiguous would create a new copy of the tensor, thereby breaking parameter sharing.

src/peft/tuners/lora/model.py
romitjain: https://github.com/huggingface/peft/pull/2879#discussion_r2502221448

It should be possible, it would just make the flow very convoluted.

src/peft/tuners/lora/model.py
romitjain: Yes, that is true. I am not able to find a way to get the embedding layer name from the model

src/peft/tuners/tuners_utils.py
romitjain: I think it should be fine? Otherwise will have to make in constant in `_check_tied_modules` and pass it along the function call. Or would you suggest some other to save the consant?

src/peft/tuners/lora/model.py
romitjain: I redid this a bit. We just need to make sure that `embed_tokens` is present in the `target_modules`

romitjain:
@BenjaminBossan I have made the updates. As of now, we have 3 outstanding updates we need to resolve

1. What to do in case `target_modules` is a str: https://github.com/huggingface/peft/pull/2879#discussion_r2503550670: I have replied to the comment with a small modification in the logic
2. Release of this PR: Can this PR be merged to main but not in release? Or do we have an ETA for v5 arrival? Some of our internal PRs depend on this, hence asking.
3. (Minor) Duplication of a warning message: https://github.com/huggingface/peft/pull/2879#discussion_r2510738148L I have replied to the comment

src/peft/peft_model.py
BenjaminBossan: I see. Then let's just add a comment here why this is needed.

src/peft/tuners/lora/model.py
BenjaminBossan: We could try to find the layer name whose parameter corresponds to `model.get_input_embedding()` but I'm fine with assuming the name here. Let's just add a comment.

src/peft/tuners/lora/model.py
```diff
@@ -806,8 +819,55 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+        tied_weight_keys = set(tied_weight_keys)
+        peft_config.target_modules_to_tie = tied_weight_keys
+
+        raw_target_modules = getattr(peft_config, "target_modules", None)
+
+        if isinstance(raw_target_modules, str):
+            # The way weight tying is handled for adapters, we always want to add
+            # lora adapters to the input embedding layer (embed_tokens)
+            # instead of output embedding lauyer.
+            raw_target_modules = f"(?:{raw_target_modules}|.*embed_tokens$)"
```
BenjaminBossan:
```suggestion
            raw_target_modules = rf"(?:{raw_target_modules}|.*embed_tokens$)"
```

Doesn't really matter as is, but to be on the safe side.

src/peft/tuners/tuners_utils.py
BenjaminBossan: We can just make it a global, I don't see an issue with that.

romitjain:
1. https://github.com/huggingface/peft/pull/2879#discussion_r2518948684: Good suggestion. I have added a new function to find a layer by reference tensors
2. I have resolved the merge conflict and addressed your remaining comments
3. Here's a psudo-code of the complete flow that is currently implemented
```
def inject_adapter():
    ...

    # finds tied modules and adds to a set - target_modules_to_tie
    self._check_tied_modules(model, peft_config)

    tied_targets = []

    # loop 1
    for k in keys:
        result = k in target_modules
        to_tie = k in target_modules_to_tie

        if to_tie:
            tied_targets.append(k)
            continue
        if result:
            add_lora(k)

    # loop 2
    for t in tied_targets:
        add_tied_lora(t)

    ...
```
4. Here is the flow of the alternate that you suggested

```
def inject_adapter():
    ...
    # loop 1
    for k in keys:
        result = k in target_modules
        if result:
            add_lora(k)


    # finds tied modules and adds to a set - target_modules_to_tie
    self._check_tied_modules(model, peft_config)

    # this is needed since all the tied adapters reference embedding
    # layer's lora as source
    add_lora(embed_tokens)

    # loop 2
    for t in target_modules_to_tie:
        remove_lora(t) # will remove lora if it exists, this might be a bit complex
        add_tied_lora(t)

    ...
```

BenjaminBossan (changes requested):
Thanks for the latest improvements to the PR. As you can see, there is one test failing because of a warning not matching, could you please take a look?

> Here's a psudo-code of the complete flow that is currently implemented

Thanks for this. I think the second approach wouldn't actually differ that much when it comes to complexity, as such I'm fine with keeping the current approach.

romitjain:
@BenjaminBossan, I have fixed the test. I have removed 3 redundant tests which are no longer required.

tests/test_initialization.py
```diff
@@ -4935,6 +4937,130 @@ def test_weight_tying_tied_model_lokr(self):
...
+            lm_lora_A = model.base_model.model.lm_head.lora_A[adapter_name].weight
+            lm_lora_B = model.base_model.model.lm_head.lora_B[adapter_name].weight
+
+            assert embed_lora_A.data_ptr() != lm_lora_B.data_ptr()
+            assert embed_lora_B.data_ptr() != lm_lora_A.data_ptr()
+        else:
+            raise NotImplementedError("Layer type {layer} is not supported for this test")
+
+    def test_weight_tying_tied_model_target_modules_lokr(self):
+        from peft.tuners.lokr.layer import LoKrLayer
```
BenjaminBossan: The import can be moved to the root level.

romitjain:
@BenjaminBossan Resolved your comment

romitjain:
@BenjaminBossan I made a small commit - in one of my earlier commits, I had made a change where the target modules were saved as `model.embed_tokens` instead of `embed_tokens`. This was causing some downstream issues. 
Apologies for the late commit, I know that you will have to trigger the runs again 😅

BenjaminBossan:
Thanks for the update @romitjain. There is a new merge conflict, could you please check?

romitjain:
@BenjaminBossan Done

romitjain:
@BenjaminBossan Let me know if any steps are remaining from my side for final push?

BenjaminBossan:
@romitjain No, thank you, let's wait for @githubnemo's review.

romitjain:
Hi @githubnemo, it would be very helpful if you could review the PR. One of our internal features depends on this :)

romitjain:
Hi @githubnemo, gentle reminder. Would really appreciate an update. Thanks

src/peft/tuners/lora/model.py
```diff
@@ -861,8 +870,63 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+        Args:
+            peft_config (LoraConfig)
+            tied_weight_keys (list[str])
+        """
+        tied_weight_keys = set(tied_weight_keys)
+        peft_config.modules_to_tie = tied_weight_keys
+
+        modules_to_save = getattr(peft_config, "modules_to_save", []) or []
+
+        embed_layer_name = find_parameter_name_by_tensor(self.model, self.model.get_input_embeddings().weight)
```
githubnemo: I think there is no guarantee that this will return the name of the embedding layer. It could also return the name of a layer tied to the embedding layer. It is probably safer to compare module identity instead (even though for transformers <5 this will also be flaky for models like T5).

src/peft/tuners/lora/model.py
```diff
@@ -861,8 +870,63 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+        # find_parameter_name_by_tensor returns the parameter name, so we need to strip the weight from the name
+        embed_layer_name = embed_layer_name.replace(".weight", "").replace("model.", "")
```
githubnemo: Not sure if replacing these strings is a good idea. `encoder_model.embed_tokens` would be turned into `encoder_embed_tokens`. Maybe using a more restricted approach (only one replacement, only if the key is found) would be better? `.weight` for example could be dropped by using `.removesuffix`.

src/peft/tuners/lora/model.py
```diff
@@ -861,8 +870,63 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
-        missing_keys = set(tied_weight_keys) - modules_to_save
+    def _add_modules_to_tie(self, peft_config: LoraConfig, tied_weight_keys: list[str]):
+        """
+        Tied weight keys contains the layers tied to the embedding layer. Add embedding layer and remove rest of the
+        tied layers from `module_to_save`. Maintain a separate set for layers to be tied
+
+        Args:
+            peft_config (LoraConfig)
+            tied_weight_keys (list[str])
+        """
```
githubnemo, line 884: Let's remove or document the parameters from the docstring, simply listing them doesn't add value.

Something like this:

```suggestion
        """
        Add embedding layer to `modules_to_save` and remove rest of the tied layers from `module_to_save`. 
        Maintain a separate set for layers to be tied in `peft_config.tied_weights_keys`.

        Args:
            peft_config (LoraConfig)
            tied_weight_keys (list[str])
                Contains the layers tied to the embedding layer.
        """
```

But what I'm still missing is an explanation to some of this. Especially: 
> "Maintain a separate set for layers to be tied [in `peft_config.tied_weights_keys`.]" 

Can you add to the docstring to what end this is done?

src/peft/tuners/lora/model.py
```diff
@@ -861,8 +870,63 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+        embed_layer_name = find_parameter_name_by_tensor(self.model, self.model.get_input_embeddings().weight)
+        # find_parameter_name_by_tensor returns the parameter name, so we need to strip the weight from the name
+        embed_layer_name = embed_layer_name.replace(".weight", "").replace("model.", "")
+
+        if embed_layer_name not in modules_to_save:
+            modules_to_save.append(embed_layer_name)
+
+        for m in tied_weight_keys:
+            if m in modules_to_save:
+                modules_to_save.remove(m)
```
githubnemo: I'm not sure how often this will generate a match. If I understand correctly, `tied_weight_keys` are fully-qualified keys. So this check will only match if the keys in `modules_to_save` are also fully-qualified. I don't think this happens often. cc @BenjaminBossan

src/peft/tuners/lora/config.py
```diff
@@ -670,7 +675,7 @@ class LoraConfig(PeftConfig):
                 "Whether to tie weights or not after peft initialization. "
                 "This will ensure that the adapters added to the tied layers "
                 "are also tied. This is only applicable for layers passed via "
-                "`modules_to_save`."
+                "`modules_to_save` and and `target_modules`."
```
githubnemo:
```suggestion
                "`modules_to_save` and `target_modules`."
```

src/peft/peft_model.py
```diff
@@ -328,6 +328,15 @@ def save_mutated_as_lora(peft_config, path_initial_model_for_weight_conversion,
                     output_state_dict = save_mutated_as_lora(
                         peft_config, path_initial_model_for_weight_conversion, output_state_dict, kwargs
                     )
+
+                # Before exporting the parameters we need to make sure
+                # all the tensors are contigious. Tensors can become non contigiuous
```
githubnemo:
```suggestion
                # Before exporting the parameters we need to make sure all the tensors are contigious as saving 
                # non-contiguous parameters is not supported. Tensors can become non contigiuous
```

src/peft/tuners/lora/layer.py
```diff
@@ -156,6 +156,7 @@ def update_layer(
         arrow_config: ArrowConfig = None,
         qalora_group_size: int = 32,
         inference_mode: bool = False,
+        tied_adapters: Optional[dict[str, nn.Parameter]] = None,
```
githubnemo: Is this a misnomer? IIUC it only contains LoRA parameters for one adapter?

src/peft/tuners/lora/model.py
```diff
@@ -861,8 +870,63 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
 
         return tensors_lora
 
-    def _add_modules_to_tie(self, peft_config, tied_weight_keys):
-        modules_to_save = set(getattr(peft_config, "modules_to_save", []) or [])
-        missing_keys = set(tied_weight_keys) - modules_to_save
+    def _add_modules_to_tie(self, peft_config: LoraConfig, tied_weight_keys: list[str]):
```
githubnemo: Now that we have `_add_target_modules` as well I'm wondering if we should refactor this to `_add_modules_to_save_to_tie` for clarity (it is verbose, yes).

Same goes for the config key `modules_to_tie`.

src/peft/tuners/lora/model.py
```diff
@@ -861,8 +870,63 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+            raw_target_modules = rf"(?:{raw_target_modules}|.*{embed_layer_name}$)"
+            peft_config.target_modules = raw_target_modules
+            return
+
+        target_modules = set(raw_target_modules or [])
+        target_modules.add(embed_layer_name)
+
+        for m in tied_weight_keys:
+            if m in target_modules:
+                target_modules.remove(m)
```
githubnemo: This will also only occasionally match, right? Only if users supply the fully-qualified module names.

src/peft/tuners/lora/layer.py
romitjain: Yes, this should be `tied_adapter`

src/peft/tuners/lora/model.py
romitjain: I have updated the docstring

romitjain:
@githubnemo Thanks for the review, I'll redo the PR a bit based on your suggestions

src/peft/tuners/lora/model.py
romitjain: Updated the logic

src/peft/tuners/lora/model.py
romitjain: @githubnemo In the flow, I am unable to find any place where we are converting keys in `modules_to_save` to fully qualified keys.

The only two relevant checks I see are:
1. https://github.com/huggingface/peft/blob/main/src/peft/tuners/tuners_utils.py#L1655
2. https://github.com/huggingface/peft/blob/main/src/peft/utils/other.py#L1016


Here, we want to make sure that fully qualified keys from `tied_weight_keys` match the ones in `modules_to_save`. I propose that I do the following:

For every key in `model.named_parameters`, I perform a check similar to what is given in (1) and match it with `tied_weight_keys`. If both of them give a match, I remove the key from `modules_to_save`

romitjain:
@githubnemo I redid a few parts. Let me explain:

The major concern, according to your comments was that we are doing a check for tied keys (which are fully qualified names (fqn) of the layers) with `modules_to_save` and `target_modules` which might not be fqn during the comparison.

For `modules_to_save`, I have added a comment here: https://github.com/huggingface/peft/pull/2879#discussion_r2621826763

For `target_modules`, I think the alternate flow would be too complex to follow. I explain my thought process in point 3 here: https://github.com/huggingface/peft/pull/2879#issuecomment-3526514139

But essentially, we are converting `target_modules` to fqns while also adding adapters to it. 

For your suggestion on collecting the fqns first, we will have to break this loop - https://github.com/huggingface/peft/blob/main/src/peft/tuners/tuners_utils.py#L774-L820 where we first generate fqns and then in a second loop add adapters which are aware of the weight tying. I can implement this as well, if you feel this is the best solve.

My latest changes (67a71d63a18d3f28d2333eca23325087666d2326) incorporated all of your suggestions and an opinionated solution on `target_modules` matching with tied keys (different from what you proposed).

Let me know your thoughts.

src/peft/tuners/lora/model.py
romitjain: Updated to check for module identity instead

romitjain:
Hi @githubnemo, gentle reminder for your review :)

BenjaminBossan:
@romitjain Sorry for the delay, we will hopefully get this finishes soon. As the corresponding PR for trainable tokens was merged, we now have some merge conflicts, but that should be easily resolved. Could you please take care?

romitjain:
@BenjaminBossan Resolved the merge conflict

romitjain:
Hi @BenjaminBossan/ @githubnemo 
Gentle reminder for the review. Let me know if the explanation I added to https://github.com/huggingface/peft/pull/2879/#issuecomment-3659635894 makes sense.

Thank you

BenjaminBossan:
Thanks for updating. From my POV, it's good to go, but since it's a complex topic, let's wait for the 2nd review.

romitjain:
@githubnemo Request you to please review this.
I have addressed your comments in the comment: https://github.com/huggingface/peft/pull/2879#issuecomment-3659635894

Thank you

src/peft/tuners/lora/layer.py
```diff
@@ -150,6 +150,19 @@ def update_layer(
         adapter_name: str,
         r: int,
         lora_alpha: int,
+        lora_dropout,
```
romitjain: I think this was added by merge commit, let me take a look and remove this if not required

githubnemo: Yeah this looks like a wrong merge. https://github.com/huggingface/peft/pull/2960 refactored this code to only include `lora_alpha` and `config` instead of all those params + kwargs. `lora_dropout` to `use_bdlora` should be removed.

tests/test_initialization.py
githubnemo: Part 1 of the comment is not resolved. I think all these tests would benefit from `s/layer/modules_to_save/` here.

tests/test_initialization.py
```diff
@@ -5064,6 +5067,128 @@ def test_weight_tying_tied_model_lokr(self):
...
+
+        lm_lora_A = model.base_model.model.lm_head.lora_A[adapter_name].weight
+        lm_lora_B = model.base_model.model.lm_head.lora_B[adapter_name].weight
+
+        assert torch.allclose(embed_lora_A, lm_lora_B.T)
+        assert torch.allclose(embed_lora_B, lm_lora_A.T)
+        assert embed_lora_A.data_ptr() == lm_lora_B.data_ptr()
+        assert embed_lora_B.data_ptr() == lm_lora_A.data_ptr()
+
+    @pytest.mark.parametrize("layer", [".*embed_tokens$", ".*lm_head$", ".*(embed_tokens|lm_head)$"])
```
githubnemo: Let's rename `layer` to `target_modules` in accordance with https://github.com/huggingface/peft/pull/2879/files#r2486161404

src/peft/tuners/lora/layer.py
```diff
@@ -743,6 +767,16 @@ def __init__(
...
+            lora_dropout=lora_dropout,
+            init_lora_weights=init_lora_weights,
+            use_rslora=use_rslora,
+            use_dora=use_dora,
+            use_alora=use_alora,
+            lora_bias=lora_bias,
+            arrow_config=arrow_config,
+            tied_adapter=kwargs.pop("tied_adapter", None),
+            lora_ga_config=lora_ga_config,
+            use_bdlora=use_bdlora,
```
githubnemo: Remove, merge artifact

src/peft/tuners/tuners_utils.py
```diff
@@ -1919,6 +1977,24 @@ def replicate_layers(model: nn.Module, layer_map: list[tuple[int, int]]):
...
+        reference_tensor (torch.Tensor): The reference tensor to find
+
+    Returns:
+        str: Name of the layer
+    """
+    for n, m in model.named_modules():
+        if m is reference_tensor:
+            return n
+
+    return ""
```
githubnemo: I think this is a misnomer now. `reference_tensor` should actually be `reference_module`. Let's change the function name as well to match the new behavior.

src/peft/tuners/lora/model.py
```diff
@@ -857,8 +867,86 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+        """
+        tied_weight_keys = set(tied_weight_keys)
+        peft_config.modules_to_tie = tied_weight_keys
+
+        modules_to_save = getattr(peft_config, "modules_to_save", []) or []
+
+        embed_layer_name = find_parameter_name_by_tensor(self.model, self.model.get_input_embeddings())
+        # find_parameter_name_by_tensor returns the parameter name, so we need to strip the weight from the name
+        if embed_layer_name.endswith(".weight"):
+            embed_layer_name = embed_layer_name.removesuffix(".weight")
```
githubnemo: I'm not sure if it is ever the case that `embed_layer_name` ends with `.weight`. When does that happen? Do we have a test for that?

src/peft/tuners/lora/model.py
```diff
@@ -857,8 +867,86 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+        prefix, sep, suffix = embed_layer_name.partition(".")
+        if sep and "model" in prefix:
+            embed_layer_name = suffix
+
+        if embed_layer_name not in modules_to_save:
+            modules_to_save.append(embed_layer_name)
```
githubnemo, line 893: I think that it's not a problem but a reader might wonder

```suggestion
        # the name suffix may already be included but we assume that adding the same module twice is not a problem
        if embed_layer_name not in modules_to_save:
            modules_to_save.append(embed_layer_name)
```

src/peft/tuners/lora/model.py
```diff
@@ -857,8 +867,86 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+
+        # Iterate over `tied_weight_keys` which are
+        # fully qualified keys and remove matching keys from
+        # `modules_to_save`. It will only remove first encounter
+        # in `module_to_save`, which should be safe, because `tied_weight_keys`
+        # is a unique set of keys
```
githubnemo: I like this comment but it would be vastly improved if would explain to what end we're doing this. Could you add the reason to this comment to inform the reader about why we're doing this?

src/peft/tuners/lora/model.py
```diff
@@ -857,8 +867,86 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+        """
+        tied_weight_keys = set(tied_weight_keys)
+        peft_config.target_modules_to_tie = tied_weight_keys
+
+        raw_target_modules = getattr(peft_config, "target_modules", None)
+
+        embed_layer_name = find_parameter_name_by_tensor(self.model, self.model.get_input_embeddings())
+        # find_parameter_name_by_tensor returns the parameter name, so we need to strip the weight from the name
+        if embed_layer_name.endswith(".weight"):
+            embed_layer_name = embed_layer_name.removesuffix(".weight")
```
githubnemo: Same comment as above

src/peft/tuners/lora/model.py
```diff
@@ -857,8 +867,86 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+        prefix, sep, suffix = embed_layer_name.partition(".")
+        if sep and "model" in prefix:
+            embed_layer_name = suffix
```
githubnemo: Why do we split the fully-qualified name and make it a less precise name only to add it to the `raw_target_modules` where it could have been a fully-qualified name? If there's a good reason then it should be documented in a comment. If it is just escaping, use `re.escape(embed_layer_name)` and use that result in the new `raw_target_modules` string.

src/peft/tuners/lora/model.py
```diff
@@ -857,8 +867,86 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+
+        if isinstance(raw_target_modules, str):
```
githubnemo, line 927: I think that `target_modules="all-linear"` qualifies for this case (technically) but it probably doesn't reach this code path since it doesn't match the embedding layer. Still, I think we should have a test that makes sure that `LoraConfig(target_modules="all-linear", ensure_weight_tying=True)` doesn't break (if it doesn't exist already, I may have missed it!)

src/peft/tuners/lora/model.py
```diff
@@ -857,8 +867,86 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
...
+            return
 
-        peft_config.modules_to_tie = missing_keys
+        target_modules = set(raw_target_modules or [])
+        target_modules.add(embed_layer_name)
+
+        # Iterate over `tied_weight_keys` which are
+        # fully qualified keys and remove matching keys from
+        # `target_modules`. It will only remove first encounter
+        # in `target_modules`, which should be safe, because `tied_weight_keys`
```
githubnemo, line 941: Same as above, misses the 'why'

tests/test_initialization.py
```diff
@@ -5064,6 +5067,128 @@ def test_weight_tying_tied_model_lokr(self):
         assert isinstance(model.base_model.model.model.embed_tokens, ModulesToSaveWrapper), (
             "Embed tokens is not added in Modules to Save"
         )
-        assert isinstance(model.base_model.model.lm_head, torch.nn.modules.linear.Linear), (
-            "LM head is not of type nn.linear"
+        assert isinstance(model.base_model.model.lm_head, torch.nn.modules.linear.Linear)
+
+    @pytest.mark.parametrize("layer", ["lm_head", "embed_tokens", ["lm_head", "embed_tokens"]])
+    def test_weight_tying_tied_model_target_modules_lora(self, layer):
```
githubnemo: I think it would be worthwhile to have a similar test for encoder-decoder models that attempts to target `embed_tokens` in T5 or something like that.

githubnemo (commented):
Hey @romitjain,

thanks for the updates and keeping up-to-date with main!

I have a few medium sized comments but generally this looks good.

One note regarding the review process: I think it's best to let the reviewer mark the issues as resolved to minimize the chances of missing a marked but still incomplete comment. It also helps me keep track of what I need to revise next time. It would help me a lot if we did it that way. Thanks :)

romitjain:
Thank you @githubnemo for the review. I am slightly held up in some tasks and will finish updates on this PR next week.

> One note regarding the review process: I think it's best to let the reviewer mark the issues as resolved to minimize the chances of missing a marked but still incomplete comment. It also helps me keep track of what I need to revise next time. It would help me a lot if we did it that way. 

I apologize, would keep this in mind 🙂

src/peft/tuners/lora/model.py
romitjain: I will remove this, this is an artifact of older logic where I was matching reference tensor by parameters instead of modules

src/peft/tuners/lora/model.py
romitjain: yes, makes sense. I have updated it

src/peft/tuners/lora/model.py
romitjain: Yes, this path does not execute for `target_modules="all-linear"`.
I have also added a test to make sure we are not breaking anything.

romitjain:
@githubnemo I have updated the PR to address your latest comments.

This is the commit - https://github.com/huggingface/peft/pull/2879/commits/19929f72f0eb7c07d39043f0ecb37a4f9b6f86a3 that addresses your comments.

Majorly:

1. Added tests for enc-dec (3 new tests), 1 test for all-linear. There are a total of 34 tests for weight tying across `modules_to_save` and `target_modules` (all passing locally)
2. I have done some small cleanup, expanded comments and removed artifacts of other merges or my older implmentation.

I know this PR has gone a few iterations over a long time. Thank you for reviewing it. I will be responisve on this PR, so appreciate your review.

romitjain:
@githubnemo gentle reminder

src/peft/tuners/lora/model.py
githubnemo: Looking at the diff this seems to be still open?

romitjain:
@githubnemo I missed changing that, please see my latest commit: https://github.com/huggingface/peft/pull/2879/changes/d3e954fa26e499709c441f798cdb30366e7f5bd9

The change in `src/peft/tuners/tuners_utils.py` is required because the saved config of the peft model contains the key

`adapter.json`
```
"modules_to_save": [
    "model.embed_tokens"
  ]
```

which would not have matched the embedding layer names while loading the model, and hence would have missed tying `lm_head`.

githubnemo (approved):
LGTM :)

Thanks for sticking with the PR and your continued effort!

Failing CI is unrelated.

romitjain:
Thanks @githubnemo and @BenjaminBossan for the continued review 🙂

PR #2887, opened 2025-11-03
CI: Install fbgemm package needed by torchao, update test

BenjaminBossan:
Newer torchao versions [require](https://github.com/huggingface/peft/actions/runs/19022204234/job/54319145351#step:6:648) `fbgemm-gpu-genai>=1.2.0`. This PR installs it into the GPU Dockerfile used for nightly testing.

I had initially hoped that this would also allow us to train `int4_weight_only` quantized torchao models, as explored in #2848. However, there is a new error when trying to use this. The test has been updated to reflect this.

### DIFF
--- a/docker/peft-gpu/Dockerfile
+++ b/docker/peft-gpu/Dockerfile
@@ -53,6 +53,7 @@ RUN source activate peft && \
     "soundfile>=0.12.1" \
     scipy \
     torchao \
+    fbgemm-gpu-genai>=1.2.0 \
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Note that the workflow for the test build of the Docker images failed:

https://github.com/huggingface/peft/actions/runs/19033675634

However, this is not due to any change in this PR but rather seems to be caused by an error when trying to determine the list of changed files (JSON parsing error), so I'd say this PR is still good to go.

PR #2893, opened 2025-11-04
FIX: Correct adapter dtype with bnb weights

### LINKED ISSUE #2889: `autocast_adapter_dtype=False` doesn't work when the model is quantized
### System Info

- peft @ db7fb1a
- transformers @ 307c523854
- bitsandbytes==0.48.2

### Who can help?

@BenjaminBossan 

### Reproduction

When the model is quantized, `autocast_adapter_dtype=False` doesn't work.

```python
from transformers import AutoModelForCausalLM, BitsAndBytesConfig
import torch

model_id = "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5"

    model_id,
    quantization_config=BitsAndBytesConfig(
        load_in_4bit=True,
        bnb_4bit_compute_dtype=torch.float16,
        bnb_4bit_quant_type="nf4",
    ),
    dtype=torch.float16,
)

model = get_peft_model(model, LoraConfig(), autocast_adapter_dtype=False)

for n, p in model.named_parameters():
    if "lora" in n:
        print(n, "\t", p.dtype)
```

```
base_model.model.model.layers.0.self_attn.q_proj.lora_A.default.weight   torch.float32
base_model.model.model.layers.0.self_attn.q_proj.lora_B.default.weight   torch.float32
base_model.model.model.layers.0.self_attn.v_proj.lora_A.default.weight   torch.float32
base_model.model.model.layers.0.self_attn.v_proj.lora_B.default.weight   torch.float32
base_model.model.model.layers.1.self_attn.q_proj.lora_A.default.weight   torch.float32
base_model.model.model.layers.1.self_attn.q_proj.lora_B.default.weight   torch.float32
base_model.model.model.layers.1.self_attn.v_proj.lora_A.default.weight   torch.float32
base_model.model.model.layers.1.self_attn.v_proj.lora_B.default.weight   torch.float32
```

when not quantized, it works fine:

```python
import torch

model_id = "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5"

    model_id,
    dtype=torch.float16,
)

model = get_peft_model(model, LoraConfig(), autocast_adapter_dtype=False)

for n, p in model.named_parameters():
    if "lora" in n:
        print(n, "\t", p.dtype)
```

```
base_model.model.model.layers.0.self_attn.q_proj.lora_A.default.weight   torch.float16
base_model.model.model.layers.0.self_attn.q_proj.lora_B.default.weight   torch.float16
base_model.model.model.layers.0.self_attn.v_proj.lora_A.default.weight   torch.float16
base_model.model.model.layers.0.self_attn.v_proj.lora_B.default.weight   torch.float16
base_model.model.model.layers.1.self_attn.q_proj.lora_A.default.weight   torch.float16
base_model.model.model.layers.1.self_attn.q_proj.lora_B.default.weight   torch.float16
base_model.model.model.layers.1.self_attn.v_proj.lora_A.default.weight   torch.float16
base_model.model.model.layers.1.self_attn.v_proj.lora_B.default.weight   torch.float16
```

### Expected behavior

I'd expect `autocast_adapter_dtype=False` to work in both cases (quantized and not quantized)

BenjaminBossan:
Resolves #2889

## Description

The reported bug is this: When the base model is quantized with 4bit bitsandbytes, the adapter weights would be cast to float32, even if `autocast_adapter_dtype=False` was passed. This is because the dtype of the base layer was not correctly determined in that case. This PR now correctly determines the dtype.

While working on this, I noticed that the `peft_model.add_adapter` method was lacking the option to disable autocasting. This was added now and the tests cover it as well. I also updated some of the corresponding docstrings.

## Tangential changes

OSF implemented its own `_cast_adapter_dtype`. This would basically bypass upcasting to float32 of the OSF weights if the base model is loaded in lower precision. However, unless the user explicitly passes `autocast_adapter_dtype=False`, the default in PEFT is to upcast the adapter weights to float32. With the changes to this PR, upcasting is now done. To make this work with the `forward` pass, the `x` is cast to the dtype of the `weight`. We assume that the output dtype should be the same as the original dtype of `x`.

## TODOs

There is still an issue left with 8bit bnb weights. They don't have a compute dtype, so at a layer level, it is not possible to determine what the dtype of the PEFT adapter should be (of course, it cannot be int8). Therefore, the corresponding tests for 8bit bnb are x-failing for now. One possible solution could be to pass down the dtype of the base model (if any) and use that as a fallback. This could be implemented in a later PR.

### DIFF
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -2101,7 +2101,7 @@ class TestPeftCustomModel(PeftCommonTester):
         if autocast_adapter_dtype and (config_cls == LNTuningConfig):
             # LN Tuning basically copies the base weight and makes it trainable, hence it makes sense to keep the dtype
             # of the base model weight.
-            pytest.skip("LNTuning and OSF are exempted from casting the adapter weights to float32")
+            pytest.skip("LNTuning is exempted from casting the adapter weights to float32")
 
         if autocast_adapter_dtype:
             expected_dtype = torch.float32


BenjaminBossan:
@NikhilNayak-debug this PR contains some updates to OSF. Could you please check if they make sense? Check the PR description above for the reason of the changes.

BenjaminBossan:
@githubnemo We haven't heard back from Nikhil for two weeks, let's just proceed. PR is ready for review.

src/peft/utils/other.py
```diff
@@ -555,6 +555,8 @@ def _forward_wrapped_passthrough(self, x, *args, **kwargs):
         return self.original_module(x, *args, **kwargs)
 
     def _hasattr_wrapped(self, name, modules):
+        if not self.active_adapters:
```
BenjaminBossan: TODO: Better fix in #2936, remove this.

BenjaminBossan:
@githubnemo I removed the change to `_hasattr_wrapped`, PR should be good to review.

tests/test_custom_models.py
```diff
@@ -2041,6 +2041,71 @@ def test_forward_bfloat16_no_autocast(self, test_name, model_id, config_cls, con
...
+
+        Checks:
+        - get_peft_model
+        - add_adapter
+        - PeftModel.from_pretrained
+        - load_adapter
+        - with and without autocasting adapter dtype
+        - with and without low_cpu_mem_usage (which only makes sense for loading adapters)
+        """
+        if autocast_adapter_dtype and (config_cls == LNTuningConfig):
```
NikhilNayak-debug, line 2101: The comment says this test should be skipped for OSF as well, but the if condition only checks for `LNTuningConfig`, OSF isn't included.

NikhilNayak-debug:
@BenjaminBossan apologies for the delay in response. Has this change been tested with OSF for both quantized and higher precision model training? Conceptually it looks good to me, but want to make sure OSF's SVD components (both high rank params and low rank frozen buffers) handle dtype transitions correctly in both scenarios.

BenjaminBossan:
Thanks for your feedback @NikhilNayak-debug. I wanted to add a test to ensure that OSF keeps working. This is what I came up with:

```python
class TestOsf:
    device = infer_device()
    causal_lm_model_id = "peft-internal-testing/opt-125m"

    @pytest.mark.parametrize("dtype", [torch.float32, torch.float16])
    @pytest.mark.parametrize("autocast", [True, False])
    def test_osf_training(self, dtype, autocast, tmp_path):
        # test is analog to PeftBnbGPUExampleTests.test_causal_lm_training
            self.causal_lm_model_id,
            # quantization_config=BitsAndBytesConfig(load_in_8bit=True),
            device_map="auto",
            dtype=dtype,
        )

        tokenizer = AutoTokenizer.from_pretrained(self.causal_lm_model_id)
        config = OSFConfig()
        model = get_peft_model(model, config, autocast_adapter_dtype=autocast)

        data = load_dataset_english_quotes()
        data = data.map(lambda samples: tokenizer(samples["quote"]), batched=True)

        trainer = Trainer(
            model=model,
            train_dataset=data["train"],
            args=TrainingArguments(
                per_device_train_batch_size=4,
                warmup_steps=2,
                max_steps=3,
                learning_rate=2e-4,
                logging_steps=1,
                output_dir=tmp_path,
            ),
            data_collator=DataCollatorForLanguageModeling(tokenizer, mlm=False),
        )
        trainer.train()
        # assert loss is not None
```

(adapted from other tests in `tests/test_gpu_examples.py`)

This test passes both on main and on this branch.

However, with quantized models, there is an error (uncomment the `quantization_config` to check). The error basically stems from this line:

https://github.com/huggingface/peft/blob/f2c066889543778ea3eaf43798d041bd9acda5a7/src/peft/tuners/osf/utils.py#L45

The dtype will be something like `int8`, which means we cannot create a trainable `nn.Parameter` with that dtype. From what I can tell, OSF has thus never worked with quantized models to begin with. Support for quantized models would furthermore require dedicated layer types to handle the merging/unmerging, which are not there. So I think we don't need to check with quantized models.

NikhilNayak-debug:
@BenjaminBossan you are right. OSF does not have support for quantized models yet. Your test without `quantization_config` should be fine as long as it covers the scenarios outside of quantized models.

tests/test_custom_models.py
BenjaminBossan: Good catch, I removed the mention of OSF it in the latest commit.

PR #2895, opened 2025-11-04
[FEAT] Integrate BD-LoRA into PEFT

Conzel:
[Block-Diagonal LoRA (BD-LoRA)](https://openreview.net/forum?id=1cjLvtFOmL) (NeurIPS 2025) is a LoRA-variant in which some LoRA factors are constrained to be block-diagonal. This allows faster serving (e.g. 10-40% with Llama 3.1-8B) by eliminating communication overheads when running inference on multiple GPUs. Despite the block-diagonal constraint, BD-LoRA is as performant as vanilla LoRA at similar parameter counts. For performance and speed examples, see the plots below. 

BD-LoRA is designed to be used with tensor parallelism, which means sharding the weights of a model among multiple GPUs. I am currently also working on a pull-request to [vLLM](https://github.com/vllm-project/vllm) which will enable vLLM to use the adapters from this BD-LoRA implementation for GPU speed-ups. I will release the pull-request to vLLM in the next 1-2 days and post a link to it. 

Image

Image

I have included in this PR:

- Implementation of BD-LoRA under `src/peft/tuners/bdlora`, following other LoRA variants
- Performance experiments under `method_comparison`
- Usage example in the form of an ipynb notebook under `examples/bdlora_finetuning`
- Tests under `tests/test_bdlora.py`

### DIFF
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -1802,21 +1802,21 @@ class TestPeftCustomModel(PeftCommonTester):
 
     @pytest.mark.parametrize("test_name, model_id, config_cls, config_kwargs", TEST_CASES)
     def test_merge_layers(self, test_name, model_id, config_cls, config_kwargs):
-        _skip_if_merging_not_supported(model_id, config_cls)
+        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
 
         config_kwargs = set_init_weights_false(config_cls, config_kwargs)
         self._test_merge_layers(model_id, config_cls, config_kwargs)
 
     @pytest.mark.parametrize("test_name, model_id, config_cls, config_kwargs", TEST_CASES)
     def test_merge_layers_fp16(self, test_name, model_id, config_cls, config_kwargs):
-        _skip_if_merging_not_supported(model_id, config_cls)
+        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
 
         config_kwargs = set_init_weights_false(config_cls, config_kwargs)
         self._test_merge_layers_fp16(model_id, config_cls, config_kwargs)
 
     @pytest.mark.parametrize("test_name, model_id, config_cls, config_kwargs", TEST_CASES)
     def test_merge_layers_is_idempotent(self, test_name, model_id, config_cls, config_kwargs):
-        _skip_if_merging_not_supported(model_id, config_cls)
+        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
 
         # calling merge twice with the same arguments should not change the output
         config_kwargs = set_init_weights_false(config_cls, config_kwargs)
@@ -1824,7 +1824,7 @@ class TestPeftCustomModel(PeftCommonTester):
 
     @pytest.mark.parametrize("test_name, model_id, config_cls, config_kwargs", TEST_CASES)
     def test_safe_merge(self, test_name, model_id, config_cls, config_kwargs):
-        _skip_if_merging_not_supported(model_id, config_cls)
+        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
 
         # calling merge twice with the same arguments should not change the output
         config_kwargs = set_init_weights_false(config_cls, config_kwargs)
@@ -1942,7 +1942,7 @@ class TestPeftCustomModel(PeftCommonTester):
         # check that none of this raises an error
         model(**X)
 
-        _skip_if_merging_not_supported(model_id, config_cls)
+        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
 
         model.merge_adapter(safe_merge=False)
         model(**X)
@@ -1982,7 +1982,7 @@ class TestPeftCustomModel(PeftCommonTester):
         # check that none of this raises an error
         model(**X)
 
-        _skip_if_merging_not_supported(model_id, config_cls)
+        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
 
         model.merge_adapter(safe_merge=False)
         model(**X)
@@ -2021,7 +2021,7 @@ class TestPeftCustomModel(PeftCommonTester):
         # check that none of this raises an error
         model(**X)
 
-        _skip_if_merging_not_supported(model_id, config_cls)
+        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
 
         model.merge_adapter(safe_merge=False)
         model(**X)
@@ -2060,7 +2060,7 @@ class TestPeftCustomModel(PeftCommonTester):
         # check that none of this raises an error
         model(**X)
 
-        _skip_if_merging_not_supported(model_id, config_cls)
+        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
 
         model.merge_adapter(safe_merge=False)
         model(**X)
@@ -2239,7 +2239,7 @@ class TestPeftCustomModel(PeftCommonTester):
 
     @pytest.mark.parametrize("test_name, model_id, config_cls, config_kwargs", TEST_CASES)
     def test_disable_adapters_with_merging(self, test_name, model_id, config_cls, config_kwargs):
-        _skip_if_merging_not_supported(model_id, config_cls)
+        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
 
         # same as test_disable_adapters, but with merging
         X = self.prepare_inputs_for_testing()
@@ -3884,13 +3884,9 @@ class TestRequiresGrad:
         config = LoraConfig(target_modules=["lin0"], modules_to_save=["lin1"])
         peft_model = get_peft_model(MLP(), config)
 
-        # when disabling the adapter, the original module's grad should be enabled and vice versa
+        # no layer should have requires_grad
         peft_model.disable_adapter_layers()
-        self.check_requires_grad(
-            peft_model,
-            "base_model.model.lin1.original_module.weight",
-            "base_model.model.lin1.original_module.bias",
-        )
+        self.check_requires_grad(peft_model)
 
         # when re-enabling the adapter, the original module's grad should be disabled and vice versa
         peft_model.enable_adapter_layers()
@@ -3902,13 +3898,9 @@ class TestRequiresGrad:
             "base_model.model.lin0.lora_B.default.weight",
         )
 
-        # when using the disable_adapter context, the original module's grad should be enabled and vice versa
+        # when using the disable_adapter context, no layer should have requires_grad
         with peft_model.disable_adapter():
-            self.check_requires_grad(
-                peft_model,
-                "base_model.model.lin1.original_module.weight",
-                "base_model.model.lin1.original_module.bias",
-            )
+            self.check_requires_grad(peft_model)
 
         # after context is exited, return to the previous state
         self.check_requires_grad(
--- a/docker/peft-gpu/Dockerfile
+++ b/docker/peft-gpu/Dockerfile
@@ -38,32 +38,33 @@ RUN apt-get update && \
 
 RUN chsh -s /bin/bash
 SHELL ["/bin/bash", "-c"]
-RUN source activate peft && \ 
-    python3 -m pip install --no-cache-dir bitsandbytes optimum auto-gptq && \
+
+RUN conda run -n peft pip install --no-cache-dir bitsandbytes optimum auto-gptq
+
+RUN \
     # Add autoawq for quantization testing
-    python3 -m pip install --no-cache-dir https://github.com/casper-hansen/AutoAWQ/releases/download/v0.2.7.post2/autoawq-0.2.7.post2-py3-none-any.whl && \
-    python3 -m pip install --no-cache-dir https://github.com/casper-hansen/AutoAWQ_kernels/releases/download/v0.0.9/autoawq_kernels-0.0.9-cp311-cp311-linux_x86_64.whl && \
-    # Add eetq for quantization testing
-    python3 -m pip install git+https://github.com/NetEase-FuXi/EETQ.git
+    conda run -n peft pip install --no-cache-dir https://github.com/casper-hansen/AutoAWQ/releases/download/v0.2.7.post2/autoawq-0.2.7.post2-py3-none-any.whl && \
+    conda run -n peft pip install --no-cache-dir https://github.com/casper-hansen/AutoAWQ_kernels/releases/download/v0.0.9/autoawq_kernels-0.0.9-cp311-cp311-linux_x86_64.whl && \
+    # Add eetq for quantization testing; needs to run without build isolation since the setup
+    # script directly imports torch from the environment which would fail with isolation.
+    conda run -n peft pip install --no-build-isolation git+https://github.com/NetEase-FuXi/EETQ.git
 
 # Activate the conda env and install transformers + accelerate from source
-RUN source activate peft && \
-    python3 -m pip install -U --no-cache-dir \
-    librosa \
-    "soundfile>=0.12.1" \
-    scipy \
-    torchao \
-    fbgemm-gpu-genai>=1.2.0 \
-    git+https://github.com/huggingface/transformers \
-    git+https://github.com/huggingface/accelerate \
-    peft[test]@git+https://github.com/huggingface/peft \
-    # Add aqlm for quantization testing
-    aqlm[gpu]>=1.0.2 \
-    # Add HQQ for quantization testing
-    hqq
+RUN conda run -n peft pip install -U --no-cache-dir \
+        librosa \
+        "soundfile>=0.12.1" \
+        scipy \
+        torchao \
+        fbgemm-gpu-genai>=1.2.0 \
+        git+https://github.com/huggingface/transformers \
+        git+https://github.com/huggingface/accelerate \
+        peft[test]@git+https://github.com/huggingface/peft \
+        # Add aqlm for quantization testing
+        aqlm[gpu]>=1.0.2 \
+        # Add HQQ for quantization testing
+        hqq
 
-RUN source activate peft && \
-    pip freeze | grep transformers
+RUN conda run -n peft pip freeze | grep transformers
 
 RUN echo "source activate peft" >> ~/.profile
 
--- a/method_comparison/MetaMathQA/README.md
+++ b/method_comparison/MetaMathQA/README.md
@@ -64,6 +64,12 @@ There is a default file for the non-PEFT parameters: `default_training_params.js
 
 For an overview of all possible arguments, you can also check the `TrainConfig` `dataclass` in `utils.py`.
 
+#### About `torch.compile`
+
+Right now, compilation is a simple on/off switch in `training_params.json`. There is probably room for optimization here.
+
+Due to the model being switched to `eval` mode for the validation metrics and then back to `train` mode, we will incur a re-compilation. This is acceptable to ensure that validation runs correctly. However, this prevents us from using `torch._dynamo.config.patch(error_on_recompile=True, inline_inbuilt_nn_modules=False)` to detect frequent recompilations. It should be noticeable from the duration of training steps, though, so we're fine with that.
+
 ### Runtime performance
 
 Several factors should be considered to achieve a fast runtime performance. Besides the obvious factors like `max_steps` or the base model size, we found the following factors to have a significant impact:
@@ -235,7 +241,6 @@ Python 3.12+ is required.
 - consider adding `weight` argument to cross entropy calculation to downweight the EOS token, but it would require calculating the loss manually instead of relying on transformers (see https://github.com/huggingface/transformers/blob/6a876462c308bd7cd7d3ca8e93abaa7d5b02e90e/src/transformers/loss/loss_utils.py#L24-L48)
 - do a sanity check against/comparison with transformers Trainer
 - consider using vLLM to potentially speed up generations, at least for the test set
-- using `torch.compile` leads to a huge slowdown, investigate (maybe recompiles), although it does save memory
 - AMP does not appear to help, investigate
 - packing of sequences (but this probably requires adjusting the attention matrix)
 - clean up what gets printed and where (stdout, stderr)
--- a/method_comparison/MetaMathQA/data.py
+++ b/method_comparison/MetaMathQA/data.py
@@ -16,8 +16,8 @@
 All utilities related to data handling.
 """
 
+from collections.abc import Callable
 from functools import partial
-from typing import Callable
 
 import datasets
 import numpy as np
@@ -107,3 +107,11 @@ def tokenize_wo_answer(samples, tokenizer, template):
         input_ids[: tokenizer.model_max_length] for input_ids in tokenized["attention_mask"]
     ]
     return tokenized
+
+
+def get_wiki_small(num_samples: int = 100) -> list[str]:
+    # This way of loading the dataset avoid having to download whole shards
+    ds = load_dataset("HuggingFaceFW/finewiki", split="train", streaming=True)
+    dataset_head = ds.take(num_samples)
+    rows = [row["text"] for row in dataset_head]
+    return rows
--- a/method_comparison/MetaMathQA/run.py
+++ b/method_comparison/MetaMathQA/run.py
@@ -24,12 +24,12 @@ import random
 import sys
 import textwrap
 import time
-from contextlib import nullcontext
+from collections.abc import Callable
+from contextlib import AbstractContextManager, nullcontext
 from functools import partial
-from typing import Any, Callable, Literal, Optional
+from typing import Any, Literal, Optional
 
 import torch
-from data import get_train_valid_test_datasets
 from torch import nn
 from torch.amp import GradScaler, autocast
 from tqdm import tqdm
@@ -53,6 +53,7 @@ from utils import (
     validate_experiment_path,
 )
 
+from data import get_train_valid_test_datasets, get_wiki_small
 from peft import AdaLoraConfig, PeftConfig
 from peft.utils import CONFIG_NAME, infer_device
 
@@ -61,6 +62,9 @@ from peft.utils import CONFIG_NAME, infer_device
 BUCKET_FACTOR = 20
 # empty device cache every N steps; 10 is a good compromise between keeping memory down while lowering runtime overhead
 ACCELERATOR_EMPTY_CACHE_SCHEDULE = 10
+
+# disable torch inductor caching to keep total runtime numbers comparable when torch.compile is used
+os.environ["TORCHINDUCTOR_FORCE_DISABLE_CACHES"] = "1"
 
 
 def get_generation_config(*, seq_len, generate_kwargs) -> GenerationConfig:
@@ -92,6 +96,30 @@ def evaluate(model, tokenizer, ds, batch_size, generate_kwargs, use_tqdm: bool =
             outputs = model.generate(**batch, generation_config=generation_config, pad_token_id=tokenizer.eos_token_id)
             predictions += tokenizer.batch_decode(outputs, skip_special_tokens=True)
     return predictions, responses
+
+
+@torch.inference_mode  # type: ignore
+def calculate_mean_per_token_loss(model, tokenizer, rows: list[str], batch_size: int, max_length: int) -> float:
+    """Calculate the mean loss per token on the given dataset.
+
+    Useful to determine general model performance before and after training to get an estimate of the magnitude of
+    'forgetting'. Note that for Wikipedia data, since the information density is quite high, the loss can be
+    surprisingly large.
+
+    """
+    losses: list[float] = []
+    for j in range(0, len(rows), batch_size):
+        sliced = rows[j : j + batch_size]
+        batch = tokenizer(sliced, truncation=True, max_length=max_length)
+        batch = tokenizer.pad(batch, return_tensors="pt", padding_side="left").to(model.device)
+        outputs = model(**batch, pad_token_id=tokenizer.eos_token_id)
+        logits = outputs.logits
+        for logit, target, mask in zip(logits, batch["input_ids"], batch["attention_mask"]):
+            # calculate loss per token so that the mean is not skewed by sequence length of sample
+            num_tokens = mask.sum()
+            token_losses = torch.nn.functional.cross_entropy(logit[:num_tokens], target[:num_tokens], reduction="none")
+            losses.extend(loss.item() for loss in token_losses)
+    return torch.tensor(losses).mean().item()
 
 
 class DummyGradScaler:
@@ -140,7 +168,7 @@ def train(
     torch_accelerator_module = getattr(torch, device_type, torch.cuda)
     if use_amp:
         grad_scaler: GradScaler | DummyGradScaler = GradScaler(device=device_type)
-        autocast_ctx: Callable[[], ContextManager[Any]] = partial(autocast, device_type=device_type)
+        autocast_ctx: Callable[[], AbstractContextManager[Any]] = partial(autocast, device_type=device_type)
     else:
         grad_scaler = DummyGradScaler()
         autocast_ctx = nullcontext
@@ -167,6 +195,14 @@ def train(
     tic_train = time.perf_counter()
     eval_time = 0.0
     error_msg = ""
+
+    rows_wiki = get_wiki_small()
+    model.eval()
+    # use small batch_size, not batch_size_eval, to prevent this from taking too much memory and affecting the max memory metric
+    wiki_loss_before = calculate_mean_per_token_loss(
+        model=model, tokenizer=tokenizer, rows=rows_wiki, batch_size=batch_size, max_length=768
+    )
+    model.train()
 
     ds_train, ds_valid, ds_test = get_train_valid_test_datasets(
         tokenizer=tokenizer, query_template=query_template, print_fn=print_verbose
@@ -310,6 +346,11 @@ def train(
             use_tqdm=len(ds_test) > 100,
         )
         accuracy = get_accuracy(predictions=predictions, responses=responses)
+        # use small batch_size, not batch_size_eval, to prevent this from taking too much memory and affecting the max memory metric
+        wiki_loss_after = calculate_mean_per_token_loss(
+            model=model, tokenizer=tokenizer, rows=rows_wiki, batch_size=batch_size, max_length=768
+        )
+        forgetting = wiki_loss_after - wiki_loss_before
         metrics.append(
             {
                 "step": step,
@@ -317,6 +358,7 @@ def train(
                 "train loss": sum(losses[-eval_steps:]) / eval_steps,
                 "train samples": total_samples,
                 "train total tokens": sum(total_tokens),
+                "forgetting": forgetting,
             }
         )
         print_verbose(f"Test accuracy: {accuracy:.3f}")
--- a/method_comparison/MetaMathQA/utils.py
+++ b/method_comparison/MetaMathQA/utils.py
@@ -23,9 +23,10 @@ import platform
 import subprocess
 import tempfile
 import warnings
+from collections.abc import Callable
 from dataclasses import asdict, dataclass
 from decimal import Decimal, DivisionByZero, InvalidOperation
-from typing import Any, Callable, Literal, Optional
+from typing import Any, Literal, Optional
 
 import bitsandbytes
 import datasets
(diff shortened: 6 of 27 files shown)


BenjaminBossan:
Thanks for your work on integrating block-diagonal LoRA into PEFT. I haven't checked the details yet, but wanted to give a first, high level feedback.

1\. Parallelism

The main advantage of BD-LoRA compared to LoRA would be the better parallelism story. Did you verify that it works with the proposed implementation? I think having a small example would be really helpful. AFAICT, the included notebook doesn't really cover that.

2\. Implementation approach

You implemented BD-LoRA as a subclass of normal LoRA. The idea is probably to leverage the existing code as much as possible. However, it can also be problematic, as BD-LoRA may or may not always be a substitute for LoRA, so some checks like `isinstance(layer, LoraLayer)` may falsely return `True` for BD-LoRA. I haven't checked the details, but historically, we had such issues with AdaLoRA, which also subclasses. This approach would also require us to keep `BdLoraModel` in sync with `LoraModel` each time there is a change there.

I wonder if you have considered implementing BD-LoRA as a[ LoRA variant](https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/variants.py) instead. E.g. DoRA is implemented as a LoRA variant. The advantage would be that we have a [defined interface](https://github.com/huggingface/peft/blob/e82e72a110b41ae8a62b4af83fab2b9aab614193/src/peft/tuners/lora/layer.py#L44-L91) for BD-LoRA.

From a user's point of view, it could look like this:

```python
from peft import LoraConfig, BdLoraConfig

bd_lora_config = BdLoraConfig(
    lora_a_is_blockdiagonal=["o_proj", "down_proj"],
    lora_b_is_blockdiagonal=["q_proj", "v_proj", "k_proj", "up_proj", "gate_proj"],
    nblocks=2,
)

lora_config = LoraConfig(
    r=16,
    target_modules=["q_proj", "v_proj", "k_proj", "up_proj", "gate_proj", "o_proj", "down_proj"],
    use_bdlora=bd_lora_config,
)
```

Maybe this idea won't quite work because LoRA variants are not the right abstraction, but I wanted to discuss this first before proceeding.

Conzel:
Hey @BenjaminBossan, thank you very much for your thoughtful comments. 

1. Using this implementation and the vLLM version at https://github.com/vllm-project/vllm/pull/28136, I achieved the following speed-up for Llama-3.1-8B-Instruct using LLMPerf for evaluation on an 8xA100 GPU compute node: 

If all adapters can be kept in GPU memory:
image

If no adapter can be kept in GPU memory:
image

I agree that adding some information on how to integrate BD-LoRA vLLM would be a good idea. I would wait until the PR in vLLM is accepted (or at least ready to merge), and then add further information to the notebook. 

2. Thanks for the suggestion! My latest commit contains a draft on how it would look like using the variant. This would involve adding the target_name parameter in the create_and_replace method in lora/model.py to record the current key, as BD-LoRA needs this parameter to determine whether a module is row- or column-sharded.

Conzel:
@BenjaminBossan Is there something else that we need to move forward / that I can work on?

method_comparison/MetaMathQA/results/bdlora--llama-3.2-3B-rank46.json
BenjaminBossan, line 1: Thanks for providing the results. Note that for the final PR, we need to remove this file, as we will run the experiment on our own hardware to keep results comparable.

Regarding the settings, do you have an idea why the end result, the test accuracy, is relatively low at 36.8%? For comparison [vanilla LoRA has 48.2% test accuracy](https://github.com/huggingface/peft/blob/9d3e0e574348babb84cff3a3f49a68738ca7b7f4/method_comparison/MetaMathQA/results/lora--llama-3.2-3B-rank32.json#L321). And that is with 9M trainable parameters, compared to 49M trainable parameters used here. From the paper results, I would have expected a very close match in performance.

src/peft/tuners/lora/config.py
```diff
@@ -119,6 +119,37 @@ def __post_init__(self):
...
+    """
+    Configuration for BD-LoRA (Block-Diagonal LoRA).
+
+    Args:
+        lora_a_is_blockdiagonal: Modules where LoRA-A is block-diagonal
+        lora_b_is_blockdiagonal: Modules where LoRA-B is block-diagonal
+        nblocks: Number of blocks in block-diagonal matrices
+    """
+
+    lora_a_is_blockdiagonal: Optional[list[str]] = field(
```
BenjaminBossan: The name `lora_a_is_blockdiagonal` would suggest the argument to be a bool. How about something like `target_modules_blockdiagonal_a` or even `target_modules_bd_a` (same for `b`)?

As to the meaning of these arguments: How would I, as a user, make the correct choice for which of A or B should be block-diagonal? I think it would be great if this could be automated (at least, as an option), for instance based on the base weight. Do you think that would be possible? If not, then at the very least the doc needs to be improved so that the users know what they should set here.

src/peft/tuners/lora/config.py
```diff
@@ -119,6 +119,37 @@ def __post_init__(self):
...
+            "help": "Modules where the LoRA-B is block-diagonal. Matches each pattern in the list against the module name via `pattern is in target_name`. Example: ['out_proj', 'down_proj']"
+        },
+    )
+    nblocks: int = field(
+        default=1,
+        metadata={
+            "help": "Number of blocks each block-diagonal matrix has. If using BD-LoRA to speed up inference, set it to be equal to the desired sharding degree during serving."
+        },
+    )
+
```
BenjaminBossan, line 184: Let's add a `__post_init__` to check that A and B targets don't overlap.

src/peft/tuners/lora/variants.py
```diff
@@ -771,3 +772,85 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
+class BlockDiagonalLinear(nn.Module):
+    def __init__(self, in_features: int, out_features: int, nblocks: int, bias: bool = True):
+        super().__init__()
+        self.in_features = in_features
+        self.out_features = out_features
+        self.nblocks = nblocks
+        self.weight = nn.Parameter(torch.empty(out_features, in_features // nblocks))
+        torch.nn.init.kaiming_uniform_(self.weight, a=math.sqrt(5))
+        if bias:
+            raise NotImplementedError
```
BenjaminBossan: Let's raise with a proper error message.

src/peft/tuners/lora/variants.py
```diff
@@ -771,3 +772,85 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        first_dims = x.shape[:-1]
+        if not len(x.shape) == 2:
```
BenjaminBossan: A bit clearer IMO:

```suggestion
        if x.dim() != 2:
```

src/peft/tuners/lora/variants.py
```diff
@@ -771,3 +772,85 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
             kwargs["alora_offsets"] = None
     return kwargs
+
+
+class BlockDiagonalLinear(nn.Module):
+    def __init__(self, in_features: int, out_features: int, nblocks: int, bias: bool = True):
+        super().__init__()
+        self.in_features = in_features
+        self.out_features = out_features
+        self.nblocks = nblocks
```
BenjaminBossan, line 790: Would it make sense to check if the in/out features are divisible by `nblocks`?

src/peft/tuners/lora/variants.py
```diff
@@ -771,3 +772,85 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
+            raise ValueError(f"Target {target_name} matches both A and B block-diagonal patterns")
+
+        if has_lora_a_blockdiagonal:
+            r = module.lora_A[adapter_name].out_features
+            layer = BlockDiagonalLinear(module.in_features, r, nblocks=nblocks, bias=False)
+            module.lora_A[adapter_name] = layer
+        elif has_lora_b_blockdiagonal:
+            r = module.lora_B[adapter_name].in_features
+            layer = BlockDiagonalLinear(r, module.out_features, nblocks=nblocks, bias=False)
+            module.lora_B[adapter_name] = layer
```
BenjaminBossan, line 870: What if neither matches? Is it fine or should we raise an error?

src/peft/tuners/lora/variants.py
```diff
@@ -771,3 +772,85 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
+        lora_B = module.lora_B[active_adapter]
+        dropout = module.lora_dropout[active_adapter]
+        scaling = module.scaling[active_adapter]
+        x = dropout(x)
+        result += lora_B(lora_A(x)) * scaling
+        return result
+
+    @staticmethod
+    def merge_safe(module: Linear, active_adapter: str, orig_weight: torch.Tensor) -> torch.Tensor:
+        return orig_weight + module.get_delta_weight(active_adapter)
```
BenjaminBossan: Hmm, is it expected that the use of block diagonal LoRA does not affect how merging should be performed?

src/peft/utils/peft_types.py
```diff
@@ -27,6 +27,7 @@ class PeftType(str, enum.Enum):
     - PREFIX_TUNING
     - LORA
     - ADALORA
+    - BDLORA
```
BenjaminBossan: Can be removed, since we're going with LoRA variants.

src/peft/utils/peft_types.py
```diff
@@ -56,6 +57,7 @@ class PeftType(str, enum.Enum):
     PREFIX_TUNING = "PREFIX_TUNING"
     LORA = "LORA"
     ADALORA = "ADALORA"
+    BDLORA = "BDLORA"
```
BenjaminBossan: Can be removed, since we're going with LoRA variants.

tests/test_bdlora.py
```diff
@@ -0,0 +1,249 @@
...
+        X = self.lin1(X)
+        X = self.relu(X)
+        X = self.lin2(X)
+        X = self.relu(X)
+        X = self.lin3(X)
+        X = self.sm(X)
+        return X
+
+
+class TestBdLora:
```
BenjaminBossan: Thanks for thinking about adding tests. From skimming those, I think most if not all can be covered by extending existing tests. To do so, check e.g. how the test matrix is expanded for DoRA:

https://github.com/huggingface/peft/blob/9d3e0e574348babb84cff3a3f49a68738ca7b7f4/tests/test_custom_models.py#L91-L134

If you add similar settings for BD-LoRA there (of course only for linear layers), it should automatically cover lots of use cases of BD-LoRA. Then we can check the test coverage for gaps and add more BD-LoRA specific tests if needed.

BenjaminBossan (changes requested):
Thanks a lot for refactoring the PR to implement BD-LoRA as a LoRA variant. I think this is indeed the much nicer approach, as it integrates better and is more future proof.

> This would involve adding the target_name parameter in the create_and_replace method in lora/model.py to record the current key, as BD-LoRA needs this parameter to determine whether a module is row- or column-sharded.

We can do that but I hope that we can find a way to avoid it. I started a discussion about the choice of LoRA A vs B, depending on the outcome there, we may be able to streamline the implementation.

> I agree that adding some information on how to integrate BD-LoRA vLLM would be a good idea. I would wait until the PR in vLLM is accepted (or at least ready to merge), and then add further information to the notebook.

Sounds good. Apart from vLLM, could we also expect potential speedups from transformers TP with this technique (at least for those models that support it)? If yes, it would be great if we could include an example.

> Is there something else that we need to move forward / that I can work on?

Sorry for the delay, we had our hands full with the last PEFT release and updates required for the upcoming transformers v5.

src/peft/tuners/lora/config.py
Conzel: Sounds good, I have updated the attribute.

src/peft/tuners/lora/config.py
Conzel: Done!

src/peft/tuners/lora/variants.py
Conzel: Done!

src/peft/tuners/lora/variants.py
Conzel: Done!

src/peft/tuners/lora/variants.py
Conzel: Good question, on one hand, raising an error can prevent the user from encountering errors from typos. On the other hand, nothing in the method prevents you from executing some layers as BD-LoRA, and some not. This might be especially helpful if you have a model architecture in which some layers are not compatible with BD-LoRA, and others are. If we don't raise an error, we can allow the user this flexibility.

We could also add an attribute to the config along the lines of 'match_strict', which is set to true as default, and raises an error if a matches neither a_blockdiagonal or b_blockdiagonal.

src/peft/tuners/lora/variants.py
Conzel: This was an oversight, thank you. I have added the appropriate method for getting the delta weight in the case of BD-LoRA adapters.

src/peft/utils/peft_types.py
Conzel: Done!

src/peft/utils/peft_types.py
Conzel: Done!

tests/test_bdlora.py
Conzel: That is true, thank you! I have moved the tests to the test_custom_model file. The only test that could not be converted is the block-diagonal structure test, but this is maybe a bit too implementation dependent anyways, I have removed it.

method_comparison/MetaMathQA/results/bdlora--llama-3.2-3B-rank46.json
Conzel: > Thanks for providing the results. Note that for the final PR, we need to remove this file, as we will run the experiment on our own hardware to keep results comparable.

Okay, that makes a lot of sense. 

> Regarding the settings, do you have an idea why the end result, the test accuracy, is relatively low at 36.8%? For comparison [vanilla LoRA has 48.2% test accuracy](https://github.com/huggingface/peft/blob/9d3e0e574348babb84cff3a3f49a68738ca7b7f4/method_comparison/MetaMathQA/results/lora--llama-3.2-3B-rank32.json#L321). And that is with 9M trainable parameters, compared to 49M trainable parameters used here. From the paper results, I would have expected a very close match in performance.

At the moment, BD-LoRA uses many more parameters as I set the LoRA adapters to every linear module. I just noticed that in the LoRA result we only apply the adapters to v and q projections. Applying them to all layers might already cause overfitting/require different learning rates, which explains BD-LoRAs comparatively worse performance.

Only adding LoRA adapters to v_proj and q_proj for BD-LoRA might not be the best test scenario for BD-LoRA, as then we only test performance for modules where the LoRA B adapter is block-diagonal. Do you have any suggestions?

src/peft/tuners/lora/variants.py
Conzel: Done!

src/peft/tuners/lora/config.py
Conzel: Regarding the meaning: Yes, it would be great to automate it. However, I am not sure if this can be done in a robust manner. 

As a general guideline: To use BD-LoRA, we want to set QKV and up projections as B-block-diagonal, and down,out projections as A-block-diagonal. The _exact_ sharding scheme however might depend on the implementation in the downstream inference engine. As an example, vLLM re-implements each supported model architecture and explicitly encodes which layers are ColumnParallel (needs LoRA B block-diagonal) and which ones are RowParallel (needs LoRA A block-diagonal): Here is Llama as an example: https://github.com/vllm-project/vllm/blob/0af3d4f0df360decc2115f43f5e4bc732342e7e4/vllm/model_executor/models/llama.py

If you have some ideas on how to determine this automatically, I would be very glad.

Conzel:
@BenjaminBossan Thank you very much for your thorough review!

We can do that but I hope that we can find a way to avoid it. I started a discussion about the choice of LoRA A vs B, depending on the outcome there, we may be able to streamline the implementation.

> Sounds good. Apart from vLLM, could we also expect potential speedups from transformers TP with this technique (at least for those models that support it)? If yes, it would be great if we could include an example.

Transformers TP would also need a PR that adds BD-LoRA compatibility, similar to vLLM. I am not too familiar with Transformers TP, but if Multi-LoRA serving with Megatron-style sharding is implemented already, it should not be much work. 

> Sorry for the delay, we had our hands full with the last PEFT release and updates required for the upcoming transformers v5.

No worries, thanks for your time.

I hope to have addressed all your concerns. By moving the tests to the LoRA variant test matrix, I have found some more minor issues (related to dtype and multi-adapter settings) which I have also fixed.
From my side, two points now remain open:

- How to organize the performance tests? Should we just run BD-LoRA with adapters on the q and v projections?
- Is passing the target_name to the LoRA adapters okay or is there a more clean approach we can do?

method_comparison/MetaMathQA/results/bdlora--llama-3.2-3B-rank46.json
BenjaminBossan: We have a vanilla LoRA experiments that [targets the 'MLP' part of the model](https://github.com/huggingface/peft/blob/7068a7ed51758eec01bcc5154a6afde168b1dfd3/method_comparison/MetaMathQA/results/lora--llama-3.2-3B-rank10-target-mlp.json#L42-L46). You could do the same, which would allow a mix of A and B being BD. Let's try to match the rank to achieve a similar number of trainable parameters (9-10M). The test accuracy for this experiment is 0.526.

src/peft/tuners/lora/config.py
```diff
@@ -119,6 +119,58 @@ def __post_init__(self):
...
+    def __post_init__(self):
+        # check for overlap
+        if self.target_modules_bd_a is None or self.target_modules_bd_b is None:
+            return  # one is empty so no overlap possible
+        for module in self.target_modules_bd_a:
```
BenjaminBossan: How about:

```python
overlap = set(self.target_modules_bd_a or []) & set(self.target_modules_bd_b or [])
if overlap:
    raise ValueError(f"Found overlapping ...")
```

src/peft/tuners/lora/variants.py
```diff
@@ -771,3 +772,163 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
+            )
+
+        # Create weight with specified dtype and device
+        self.weight = nn.Parameter(torch.empty(out_features, in_features // nblocks, dtype=dtype, device=device))
+
+        if init_zero:
+            torch.nn.init.zeros_(self.weight)
+        else:
+            torch.nn.init.kaiming_uniform_(self.weight, a=math.sqrt(5))
+        if bias:
```
BenjaminBossan: Let's move this check up, directly below the check for `% nblocks`. Also, let's raise a `ValueError` instead.

src/peft/tuners/lora/variants.py
```diff
@@ -771,3 +772,163 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
+        """Returns weight in a format similar to a vanilla LoRA adapter. For this, we stack the blocks on the diagonal,
+        leaving the off-diagonals padded with zero."""
+        return torch.block_diag(*torch.chunk(self.weight, self.nblocks, dim=0))
+
+
+class BdLoraLinearVariant(LoraVariant):
+    @staticmethod
+    def init(module: Linear, adapter_name: str, **kwargs) -> None:
+        use_bdlora = kwargs.get("use_bdlora")
+        if not use_bdlora:
```
BenjaminBossan: Can this ever happen? I don't think so, since the LoRA variant is only added if `use_bdlora` is given.

src/peft/tuners/lora/variants.py
BenjaminBossan: > nothing in the method prevents you from executing some layers as BD-LoRA, and some not. This might be especially helpful if you have a model architecture in which some layers are not compatible with BD-LoRA, and others are. If we don't raise an error, we can allow the user this flexibility.

This is helpful to know. Let's add this info to the config. It would be kinda bad if none of the passed names match, in which case BD-LoRA would not be applied anywhere. But I think it's a worthwhile tradeoff for the possibility to mix BD-LoRA with vanilla LoRA.

> We could also add an attribute to the config along the lines of 'match_strict', which is set to true as default, and raises an error if a matches neither a_blockdiagonal or b_blockdiagonal.

This sounds good.

BenjaminBossan (changes requested):
Thanks for the updates, the PR is getting close to the finish line. I did another review, please check.

Regarding testing: There are a few checks being performed at initialization time, e.g. for BD target module overlap or for `% nblocks`. Could you please add tests for those to `test_initialization.py`? I think the existing tests there should give you a good idea how to implement the tests, if not LMK.

> Transformers TP would also need a PR that adds BD-LoRA compatibility, similar to vLLM. I am not too familiar with Transformers TP, but if Multi-LoRA serving with Megatron-style sharding is implemented already, it should not be much work.

Okay, let's leave that aside then. It would be nice to have but is not required for this PR.

> As a general guideline: To use BD-LoRA, we want to set QKV and up projections as B-block-diagonal, and down,out projections as A-block-diagonal. The exact sharding scheme however might depend on the implementation in the downstream inference engine. As an example, vLLM re-implements each supported model architecture and explicitly encodes which layers are ColumnParallel (needs LoRA B block-diagonal) and which ones are RowParallel (needs LoRA A block-diagonal): Here is Llama as an example: https://github.com/vllm-project/vllm/blob/0af3d4f0df360decc2115f43f5e4bc732342e7e4/vllm/model_executor/models/llama.py
> If you have some ideas on how to determine this automatically, I would be very glad.

I see, thanks for explaining. In this case, I think it's hard to impossible to automate this step. Could you put this additional information into the docstring of `BdLoraConfig` too, e.g. by extending it [here](https://github.com/huggingface/peft/pull/2895/files#diff-13380145401d203d5935c5189dd09879f990b81aa63e8e3aaff8ce9110333f0eR138)?

One thing we could do is to determine some defaults for popular models with support in vLLM and, if the user leaves `target_modules_bd_a` and `target_modules_bd_b` as `None`, check if the used model matches one of the defaults. This is, however, a bit of extra work, especially as these defaults can quickly get out of date. Up to you if you think it's worth the effort.

> Is passing the target_name to the LoRA adapters okay or is there a more clean approach we can do?

I thought a bit about that and I think it's okay as is. Another idea that I had would be to provide two variants, one for BD-A and one for BD-B, and resolve which one to use based on the target name, but I think overall that would be equally complex, so we can leave the logic is as.

One more thing, could you try to limit the characters to 120 per line, there are some very long strings for instance in `config.py`. Calling `make style` should help with some of those.

Conzel:
Hey @BenjaminBossan, thanks again for your time. 

> This is helpful to know. Let's add this info to the config. It would be kinda bad if none of the passed names match, in which case BD-LoRA would not be applied anywhere. But I think it's a worthwhile tradeoff for the possibility to mix BD-LoRA with vanilla LoRA.

Great, I added a match_strict argument that is set to True per default and raises an error if it is set and we encounter an overlap. I think this is a good solution. 

> We have a vanilla LoRA experiments that [targets the 'MLP' part of the model](https://github.com/huggingface/peft/blob/7068a7ed51758eec01bcc5154a6afde168b1dfd3/method_comparison/MetaMathQA/results/lora--llama-3.2-3B-rank10-target-mlp.json#L42-L46). You could do the same, which would allow a mix of A and B being BD. Let's try to match the rank to achieve a similar number of trainable parameters (9-10M). The test accuracy for this experiment is 0.526.

This is great. I have added a similar experiment for BD-LoRA at rank 14 (to match the number of parameters) and have 0.51857 test accuracy on my device. Additionally, I have repeated the rank 44 experiment with the correct target modules (before, I had erroneously targeted all linear layers), this time achieving  0.479 accuracy. 

Additionally, I have added an example for vLLM-integration. I had to add 2 more script files for querying the server and starting it. At the moment, I have not added a real demo for the speed-up (just a demo for how to start the vLLM server with a BD-LoRA adapter and how to send requests to it). I think a speedup-demo would involve using larger models and a more sophisticated platform for sending the requests (I used a modified LLMPerf-fork for my experiments in the initial comments). If you think this is a priority, I can try to add this information however. 

> I see, thanks for explaining. In this case, I think it's hard to impossible to automate this step. Could you put this additional information into the docstring of BdLoraConfig too, e.g. by extending it [here](https://github.com/huggingface/peft/pull/2895/files#diff-13380145401d203d5935c5189dd09879f990b81aa63e8e3aaff8ce9110333f0eR138)?

I added information into the docstring in a previous commit, just above the line you linked. Is that enough or should I add more? 

> One thing we could do is to determine some defaults for popular models with support in vLLM and, if the user leaves target_modules_bd_a and target_modules_bd_b as None, check if the used model matches one of the defaults. This is, however, a bit of extra work, especially as these defaults can quickly get out of date. Up to you if you think it's worth the effort.

I think this could be a good idea further down the road. The vLLM maintainers have requested to keep the pull request in a draft state at the moment, and want to include it into the codebase if the community demand is high enough, so maybe keeping it mainly documented for LLama is okay at the moment.

I think I am only missing the remaining initialization tests now, I will add them today or tomorrow morning. Thanks for your support! **EDIT: Done!**

src/peft/tuners/lora/variants.py
Conzel: Yeah that is true, I have converted it to an assert as a sanity-check (also to make the type-checker aware that this is not None)

examples/bdlora_finetuning/chat.py
```diff
@@ -0,0 +1,51 @@
+import asyncio
```
BenjaminBossan, line 14: Let's add a copyright notice here. Also, could you please add a short README.md so that users know what happens here and how to use the scripts?

src/peft/tuners/lora/config.py
```diff
@@ -119,6 +119,78 @@ def __post_init__(self):
...
+                "help": "Number of blocks each block-diagonal matrix has. If using BD-LoRA to speed up inference, "
+                "set it to be equal to the desired sharding degree during serving."
+            },
+        ),
+    )
+    match_strict: bool = field(
+        default=True,
+        metadata={
+            "help": "If set to true, requires each target_module to have either a block-diagonal LoRA-A or LoRA-B, "
+            "and raises an error otherwise. You can set this to False this to mix LoRA and BD-LoRA training, "
```
BenjaminBossan:
```suggestion
            "and raises an error otherwise. You can set this to False to mix LoRA and BD-LoRA training, "
```

src/peft/tuners/lora/variants.py
```diff
@@ -771,3 +771,166 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
+        leaving the off-diagonals padded with zero."""
+        return torch.block_diag(*torch.chunk(self.weight, self.nblocks, dim=0))
+
+
+class BdLoraLinearVariant(LoraVariant):
+    @staticmethod
+    def init(module: Linear, adapter_name: str, **kwargs) -> None:
+        use_bdlora = kwargs.get("use_bdlora")
+        target_name = kwargs.get("target_name", "")
+        assert use_bdlora is not None
```
BenjaminBossan: IMO, this can be fully removed. If you really want to keep it, please raise a proper error (in this case I'd say `RuntimeError("Something went wrong with BD-LoRA initialization, please open an issue at https://github.com/huggingface/peft/issues")`).

src/peft/tuners/lora/variants.py
```diff
@@ -771,3 +771,166 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
+
+        # Handle case where use_bdlora is a dict (from saved config) instead of BdLoraConfig object
+        if isinstance(use_bdlora, dict):
+            from peft.tuners.lora.config import BdLoraConfig
```
BenjaminBossan: Can the import be moved to the root? Also, it is a bit strange to go from `BdLoraConfig` to dict back to `BdLoraConfig`. Can we not keep it as a `BdLoraConfig`?

tests/test_custom_models.py
```diff
@@ -868,6 +869,46 @@
...
+    (
+        "BD-LoRA with modules_to_save",
+        "MLP",
+        LoraConfig,
+        {
+            "target_modules": ["lin0"],
+            "modules_to_save": ["lin1"],
+            "use_bdlora": BdLoraConfig(target_modules_bd_a=["lin0"], nblocks=2),
+        },
+    ),
```
BenjaminBossan: Can we get an example with `match_strict=False` and one vanilla LoRA layer? On the other hand, no need for the `modules_to_save` example, as I don't think there is any interaction there with BD-LoRA.

tests/test_custom_models.py
```diff
@@ -868,6 +869,46 @@
         DeloraConfig,
         {"target_modules": ["lin0"], "module_dropout": 0.1},
     ),
+    ##########
+    # BD-LoRA #
+    ##########
```
```suggestion
    ###########
    # BD-LoRA #
    ###########
```

;-)

tests/test_initialization.py
```diff
@@ -1194,6 +1194,73 @@ def test_lora_use_dora_with_megatron_core_raises(self):
         with pytest.raises(ValueError, match="DoRA does not support megatron_core"):
             LoraConfig(target_modules=["linear"], use_dora=True, megatron_config=megatron_config)
 
+    def test_bdlora_lora_a_block_diagonal(self, data):
```
BenjaminBossan: I think this test and the next one can be scrapped, this should be covered pretty much by the `test_custom_models.py` tests. Also, they are not really about initialization.

tests/test_initialization.py
```diff
@@ -1194,6 +1194,73 @@ def test_lora_use_dora_with_megatron_core_raises(self):
...
+            "nblocks": 2,
+            "match_strict": True,
+        }
+
+        config = LoraConfig(target_modules=["linear"], use_bdlora=bdlora_config)
+
+        with pytest.raises(ValueError, match="matches neither A nor B block-diagonal patterns"):
+            get_peft_model(model, config)
+
+    def test_bdlora_non_strict_matching_works(self, data):
```
BenjaminBossan: Can be removed, see my suggestion above to add a non-strict BD-LoRA example to `test_custom_models.py`.

src/peft/tuners/lora/variants.py
```diff
@@ -771,3 +771,166 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
+        super().__init__()
+        self.in_features = in_features
+        self.out_features = out_features
+        self.nblocks = nblocks
+        if self.in_features % nblocks != 0 or self.out_features % nblocks != 0:
+            raise ValueError(
+                f"self.in_features={self.in_features} or self.out_features={self.out_features} not divisble by {self.nblocks}"
+            )
+        if bias:
+            raise ValueError("Bias is not implemented for BlockDiagonalLinear layer.")
```
BenjaminBossan: Let's add init tests for these two errors as well.

tests/test_custom_models.py
Conzel: This should be covered by the two tests above, right? 
```
    (
        "BD-LoRA A only",
        "MLP",
        LoraConfig,
        {
            "target_modules": ["lin0", "lin1"],
            "use_bdlora": BdLoraConfig(target_modules_bd_a=["lin0"], nblocks=2, match_strict=False),
        },
    ),
    (
        "BD-LoRA B only",
        "MLP",
        LoraConfig,
        {
            "target_modules": ["lin0", "lin1"],
            "use_bdlora": BdLoraConfig(target_modules_bd_b=["lin1"], nblocks=2, match_strict=False),
        },
    ),
```

I removed the modules_to_save test :)

src/peft/tuners/lora/variants.py
Conzel: I have added a test for the divisibility error, but I actually cannot get the bias error to raise. For example, I had a test like this:

```

 def test_bdlora_bias_raises(self):
        model = self.get_model()

        bdlora_config = {
            "target_modules_bd_a": [],
            "target_modules_bd_b": ["linear"],
            "nblocks": 2,
            "match_strict": True,
        }

        config = LoraConfig(target_modules=["linear"], use_bdlora=bdlora_config, bias='lora_only')

        with pytest.raises(ValueError, match="Bias is not implemented for BdLora."):
            get_peft_model(model, config)
```

And a check at the BdLoraLinearVariant like this:

```
 elif has_lora_b_blockdiagonal:
            r = module.lora_B[adapter_name].in_features
            print(module.lora_B[adapter_name].bias)
            if module.lora_B[adapter_name].bias:
                raise ValueError("Bias is not implemented for BdLora.")
            base_layer = module.get_base_layer()
```

I thought that this should raise the ValueError for the bias. At the moment, I have just removed bias from the BlockDiagonalLinear (it is probably not the right place to catch the error anyways).

Conzel:
Hey Benjamin, thanks again for your thorough review. I hope I have caught everything this time. Please let me know especially if the example could use more work. 

Additionally, I ran into some unexpected behaviour with the bias, would be glad if you could check it out.

examples/bdlora_finetuning/chat.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan:
```suggestion
# Copyright 2025-present the HuggingFace Inc. team.
```

method_comparison/MetaMathQA/experiments/bdlora/llama-3.2-3B-rank46/adapter_config.json
BenjaminBossan, line 1: As mentioned earlier, let's keep only the rank 14 experiment.

src/peft/tuners/lora/model.py
```diff
@@ -236,7 +238,9 @@ def _create_and_replace(
                 use_rslora=lora_config.use_rslora,
                 use_dora=lora_config.use_dora,
                 lora_bias=lora_config.lora_bias,
+                target_name=current_key,
```
BenjaminBossan: Let's put the new arguments to the end.

BenjaminBossan (changes requested):
Thanks for the updates, especially the description of the example.

> This should be covered by the two tests above, right?

Yes!

> I actually cannot get the bias error to raise.

I think there is a misunderstanding here. `LoraConfig(..., bias='lora_only')` means that for this LoRA adapter, the bias of the targeted modules should be made trainable. I don't think that this has a bearing on BD-LoRA, does it?

There is a small merge conflict in the test file, could you please take care?

Conzel:
> I think there is a misunderstanding here. LoraConfig(..., bias='lora_only') means that for this LoRA adapter, the bias of the targeted modules should be made trainable. I don't think that this has a bearing on BD-LoRA, does it?

Ah perfect, then this does not interfere with BD-LoRA. I guess it is fine to just initialize BlockDiagonalLinear with bias=False and not perform any checks.

> There is a small merge conflict in the test file, could you please take care?

Done!

BenjaminBossan:
@Conzel Some tests are failing, could you please take a look? I checked one of them and it passed with GPU but failed with CPU.

Conzel:
@BenjaminBossan Sorry for that oversight, seems that test was skipped on my device, as I also used a GPU. I now unfortunately only have a Mac available for further testing, but it seems that the issue was re-writing weights with wrong (dense) shapes back to the LoRA adapter weights in `_get_bdlora_delta_weight`, which only occured in the case of using a CPU with float16:

```python
[...]
if cast_to_fp32:
     output_tensor = output_tensor.to(dtype=dtype)
     # weight_A/weight_B are set to dense matrices in the lines before
     module.lora_A[adapter].weight.data = weight_A.to(dtype) 
     module.lora_B[adapter].weight.data = weight_B.to(dtype)
```

I hope that fixes it – I cannot directly test that it does not introduce other errors, as Mac seems to set the dtype to float32 even if I force the corresponding float16 tests to run. If an error persists, I can see to it that I timely organize an ubuntu machine with a GPU for further testing.

BenjaminBossan:
At least locally, this seems to have done the trick. Could you please merge with/rebase on latest main? Then I'll rerun the CI.

Conzel:
Done!

Conzel:
@BenjaminBossan  Thanks a lot for your support, it was a joy working with you. Will let you know when the vLLM PR is merged 👍🏻

PR #2902, opened 2025-11-06
Add forward compat. for tied_weights_keys dicts

githubnemo:
In the future, `_tied_weights_keys` will be mappings (mapping destination as key and source as value). This will also mean that the semantic of `_tied_weights_keys` will change from "the keys in this list are tied to the input embedding" to "this mapping defines tying between layers of any type". To this end we'll limit the scope and provide methods to retrieve input embedding ties.

If the need arises to retrieve more complicated mappings, we can do so at a later point.

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -1584,7 +1584,6 @@ def _get_module_names_tied_with_embedding(model) -> list[str]:
         model = model.model
 
     if not hasattr(model, "_tied_weights_keys"):
-        print("NO TIED WEIGHTS KEYS")
         return []
 
     base_layer_pattern = re.compile(r"[^.]+\.base_layer\.")
@@ -1622,4 +1621,4 @@ def _get_module_names_tied_with_embedding(model) -> list[str]:
         # TODO remove this when transformers <v5 is no longer supported
         tied_weights.extend(model._tied_weights_keys)
 
-    return sorted({".".join(name.split(".")[:-1]) for name in tied_weights})
+    return sorted({name.rpartition(".")[0] for name in tied_weights})
--- a/tests/test_other.py
+++ b/tests/test_other.py
@@ -574,7 +574,7 @@ class TestGetModuleNamesTiedWithEmbedding:
                 model = AutoModelForCausalLM.from_pretrained(model_id)
 
             tied_weights_keys = list(self.model_tied_weights_mapping[model_id].keys())
-            expected_module_names = sorted({".".join(k.split(".")[:-1]) for k in tied_weights_keys})
+            expected_module_names = sorted({k.rpartition(".")[0] for k in tied_weights_keys})
 
             if tied_weights_type == "list":
                 # for transformers >=5 this tests compatibility with transformers <5


src/peft/utils/other.py
```diff
@@ -1556,3 +1554,45 @@ def create_attention_mask(
             position_ids=position_ids,
         )
     return attention_mask
+
+
+def _get_module_names_tied_with_embedding(model):
+    """
+    Get the list of the fully qualified names of the modules that are tied to the input embeddings. In case of a
+    source-target-mapping `_tied_weights_keys`, it will attempt to identify the input embedding weights.
```
BenjaminBossan: I find the second sentence confusing ("In case of"), it makes it sound like if it's a dict, the input embedding will be returned, but it means it's still the target that is returned, but the function tries to identify only those targets that are actually tied to the input embedding.

Also, let's mention that this function is for compatibility with <v5 and >= v5. Let's add a TODO for when <v5 is no longer supported.

src/peft/utils/other.py
```diff
@@ -1556,3 +1554,45 @@ def create_attention_mask(
             position_ids=position_ids,
         )
     return attention_mask
+
+
+def _get_module_names_tied_with_embedding(model):
```
BenjaminBossan: Let's add `-> list[str]`.

src/peft/utils/other.py
```diff
@@ -1556,3 +1554,45 @@ def create_attention_mask(
...
+    For example: For models which have `embed_tokens` and `lm_head` as the tied keys this function will return
+    [`lm_head`].
+
+    Note that this function will not check if weight tying is disabled by the model's config. There can be the case
+    that the weight tying definition is present but the tying is disabled via `model_config.tie_word_embeddings=False`.
+    You have to check that yourself.
+    """
+    tied_weights = []
+
+    for name, module in model.named_modules():
```
BenjaminBossan: Just for my understanding: We can't just check on the model itself because there can be submodules with their own weight tying?

src/peft/utils/other.py
```diff
@@ -1556,3 +1554,45 @@ def create_attention_mask(
...
+            # doing that is because we need to filter out the original embedding name since we promise to just
+            # return the keys of the tying targets.
+            #
+            # the reason why we don't compute `candidates` once is that there might be a few levels of nesting
+            # so that the keys have various prefixes (e.g., `model.`). eventually we'll find the module that
+            # defines both input embedding and weight tying mapping, then the keys will match.
+            input_embedding_params = set(module.get_input_embeddings().parameters())
+            candidates = [n for n, p in module.named_parameters(remove_duplicate=False) if p in input_embedding_params]
+
+            tied_weights.extend(k for k, v in module._tied_weights_keys.items() if k in candidates)
```
BenjaminBossan: Do we need to do `f"{name}.{k}" if name else k for k` here too? Also, should we check `if v in candidates` instead of `k`?

BenjaminBossan (changes requested):
Thanks for handling the forward compatibility of `_tied_weights_keys`. It's unfortunate that it's a more complicated issue.

I tried to understand the comment but struggled a bit. Probably you have some examples in mind where these extra steps are necessary. Would it make sense to write a small suite of unit tests for `_get_module_names_tied_with_embedding` with these edge cases? This would also be helpful as v5 is not released yet, so the part of the code dealing with dicts is not covered yet.

src/peft/utils/other.py
githubnemo: That's a good question. I assumed so because the previous code walked through submodules recursively. Skimming transformers models, I haven't found any model that recursively defines tied weights keys. 

So, if this isn't the case we may be able to get away with checking `get_input_embeddigns` and `_tied_weights_keys` top-level. Not sure if we can align those values consistently though (`_tied_weights_keys` will always return base-model-relative values while `get_input_embeddings` resolution will be peft-model-based unless we do it the way it is currently solved).

BenjaminBossan: As discussed internally, let's keep the check on sub-modules, e.g. for the case where the transformers model is wrapped inside another `nn.Module`.

githubnemo:
I've taken some time to think about this and I want to propose a new variant which doesn't loop over all the modules but unpacks the model to look at the `_tied_weights_keys` attribute. This solves potential duplicates and prefix issues at the cost of function-local imports to avoid circular imports. Let me know what you think.

I've also added a few more unit tests.

src/peft/utils/other.py
```diff
@@ -1556,3 +1554,57 @@ def create_attention_mask(
...
         )
     return attention_mask
+
+
+def _get_module_names_tied_with_embedding(model) -> list[str]:
+    """
+    Get the list of the fully qualified names of the modules that are tied to the input embeddings. In case of a
+    source-target-mapping `_tied_weights_keys`, it will attempt to identify the input embedding weights from the
+    mapping and return the list of tied modules accordingly. This gives a unified interface to both transformers v4
+    tied weights and v5 mapped tied weights.
```
BenjaminBossan, line 1564: Perhaps worth mentioning:

- If the model is a PEFT model, it will be unpacked
- If the model is not a transformers model (but e.g. a custom model that wraps a transformers model), it should expose the `_tied_weights_keys` attribute.

src/peft/utils/other.py
```diff
@@ -1556,3 +1554,57 @@ def create_attention_mask(
...
+    from peft import PeftModel
+    from peft.tuners.tuners_utils import BaseTuner
+
+    tied_weights = []
+
+    if isinstance(model, PeftModel):
+        model = model.base_model
+
+    if isinstance(model, BaseTuner):
+        model = model.model
```
BenjaminBossan: How about:

```python
if hasattr(model, "get_base_model"):
    model = get_base_model()
```

That's more succinct and requires no local imports. The false positive rate on this should be pretty small and users with custom models can implement that API.

src/peft/utils/other.py
```diff
@@ -1556,3 +1554,57 @@ def create_attention_mask(
...
+            )
+
+        # technically it would be sufficient to just return candidates since that contains all the keys of
+        # all models that are tied (not just equal!) to the input embeddings. the only reason why we aren't
+        # doing that is because we need to filter out the original embedding name since we promise to just
+        # return the keys of the tying targets.
+        #
+        # the reason why we don't compute `candidates` once is that there might be a few levels of nesting
+        # so that the keys have various prefixes (e.g., `model.`). eventually we'll find the model that
+        # defines both input embedding and weight tying mapping, then the keys will match.
```
BenjaminBossan: This comment is no longer needed, right?

src/peft/utils/other.py
```diff
@@ -1556,3 +1554,57 @@ def create_attention_mask(
...
+        input_embedding_params = set(model.get_input_embeddings().parameters())
+        candidates = [n for n, p in model.named_parameters(remove_duplicate=False) if p in input_embedding_params]
+
+        tied_weights.extend(k for k, v in model._tied_weights_keys.items() if k in candidates)
```
BenjaminBossan: So `candidates` contains the names of the parameters of the source (the embedding). `_tied_weights_keys` is a mapping from target to source. Therefore, in the loop, `k` is the target weight name and `v` is the source (embedding) weight name. Shouldn't we thus check `if v in candidates`?

src/peft/utils/other.py
```diff
@@ -1556,3 +1554,57 @@ def create_attention_mask(
...
+    elif model._tied_weights_keys is not None:
+        # TODO remove this when transformers <v5 is no longer supported
+        tied_weights.extend(k for k in model._tied_weights_keys)
```
BenjaminBossan:
```suggestion
        tied_weights.extend(model._tied_weights_keys)
```

src/peft/utils/other.py
```diff
@@ -1556,3 +1554,57 @@ def create_attention_mask(
...
+
+    return list({".".join(name.split(".")[:-1]) for name in tied_weights})
```
BenjaminBossan: How about `sorted` instead of `list` for a predictable output?

tests/test_other.py
```diff
@@ -530,3 +538,67 @@ def test_get_no_split_modules_recursive(self):
 
         no_split_modules = _get_no_split_modules(model)
         assert no_split_modules == {"CLIPEncoderLayer", "LlamaDecoderLayer"}
+
+
+class TestGetTiedWithEmbedding:
```
BenjaminBossan:
```suggestion
class TestGetModulesTiedWithEmbedding:
```

tests/test_other.py
```diff
@@ -530,3 +538,67 @@ def test_get_no_split_modules_recursive(self):
...
+            if "t5" in model_id:
+                model = AutoModelForSeq2SeqLM.from_pretrained(model_id)
+            else:
+                model = AutoModelForCausalLM.from_pretrained(model_id)
+
+            expected_keys = list(
+                {".".join(k.split(".")[:-1]) for k in self.model_tied_weights_mapping[model_id].keys()}
+            )
+
+            if tied_weights_type == "linear":
```
BenjaminBossan: Let's add a comment that this is for transformers <v5 and below for \>= v5.

tests/test_other.py
```diff
@@ -530,3 +538,67 @@ def test_get_no_split_modules_recursive(self):
...
+
+            else:
+                raise RuntimeError("Invalid fixture request")
+
+    @pytest.mark.parametrize("tied_weights_type", ["linear", "mapping"])
+    @pytest.mark.parametrize("model_id", model_ids)
+    def test_get_modules_tied_to_embedding(self, model_id, tied_weights_type):
+        model, expected = self.get_model(model_id, tied_weights_type)
+
+        modules = _get_module_names_tied_with_embedding(model)
```
BenjaminBossan: You could do: `modules = model._get_module_names_tied_with_embedding()` and then there is no need to import `_get_module_names_tied_with_embedding`. It's also closer to what is executed in practice.

BenjaminBossan (changes requested):
Thanks for the updates. I still have a few comments. Most are just minor things, but there is one about the logic in checking embedding candidates where I'm confused.

src/peft/utils/other.py
githubnemo: Technically it is irrelevant if tying works. If tying doesn't work it is better to use `v` I think, going to change this.

src/peft/utils/other.py
```diff
@@ -1556,3 +1554,72 @@ def create_attention_mask(
...
+    if hasattr(model, "get_base_model"):
+        # unpack PeftModel
+        model = model.get_base_model()
+
+    if hasattr(model, "tuner_layer_cls"):
+        # unpack BaseTuner
+        model = model.model
+
+    if not hasattr(model, "_tied_weights_keys"):
+        print("NO TIED WEIGHTS KEYS")
```
BenjaminBossan:
```suggestion
        print("NO TIED WEIGHTS KEYS!!!!!!!")
```

src/peft/utils/other.py
```diff
@@ -1556,3 +1554,72 @@ def create_attention_mask(
...
+            peft_reverse_mapping.get(k, k)
+            for k, v in model._tied_weights_keys.items()
+            if peft_reverse_mapping.get(v, v) in candidates
+        )
+
+    elif model._tied_weights_keys is not None:
+        # TODO remove this when transformers <v5 is no longer supported
+        tied_weights.extend(model._tied_weights_keys)
+
+    return sorted({".".join(name.split(".")[:-1]) for name in tied_weights})
```
BenjaminBossan: Nit: 
```suggestion
    return sorted({name.rpartition(".")[0] for name in tied_weights})
```

BenjaminBossan (changes requested):
Thanks, what a mess. Just an issue with a debug print and a nit, otherwise LGTM.

BenjaminBossan (approved):
Thanks, LGTM. Can be merged once CI is green.

PR #2905, opened 2025-11-12
FIX: Patch mask for prefix tuning in transformers

BenjaminBossan:
**DON'T MERGE**

In transformers, the bidirecional mask creation (as e.g. used by Bert) does not take into account possible virtual tokens inserted by prefix tuning. This results in an attention mask of the wrong shape.

This PR monkey patches the `_preprocess_mask_arguments` function from transformers to take the virtual tokens into account. This is far from ideal but there are currently no plans to fix this on the transformers side (see https://github.com/huggingface/transformers/pull/41696).


BenjaminBossan:
Not needed, as the transformers fix was merged after all.

PR #2906, opened 2025-11-12
Resolve #2431: Remove macos-13 from tests

### LINKED ISSUE #2431: Feb 2026: Update MacOS GitHub runners or remove them completely.
Right now, the MacOS GitHub runners use `macos-13`, which uses Intel x86. This has the disadvantage that the latest torch wheels still provided for this architecture are torch v2.2.2. Later torch versions only provide wheels for MacOS arm64. Therefore, our GH MacOS runners can only install torch v2.2.2. We cannot update to arm64 because at this time, the [free GH runners for arm64](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) only have 7GB of RAM, which is not enough to run the whole test suite (see #2428).

Since we still support torch v2.2 for now, this is fine, but at one point, we'll stop supporting it. Since it was [released](https://github.com/pytorch/pytorch/releases/tag/v2.2.0) in Jan 2024, a good stop would be Feb 2026, i.e. after 2 years. When this time has come, do one of the following:

1. Remove macos from the test matrix completely.
2. If GH provides free arm64 runners with more RAM, make the switch.
3. Consider using macos large runners, which may require payment.
4. Build torch ourselves, but this might be time consuming and error prone.

githubnemo:
As discussed in #2431 since there are no better macos runners available and the intel macos runners are failing for quite some time now we now have little choice other than to stop using macos in the CI.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -84,10 +84,7 @@ jobs:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
           TRANSFORMERS_IS_CI: 1
         run: |
-          set +e
           make test
-          status=$?
-          exit $status
       - name: Dump cache content and diff
         # This is just debug info so that we can monitor if the model cache diverges substantially
         # over time and what the diverging model is.


.github/workflows/tests.yml
```diff
@@ -93,18 +87,6 @@ jobs:
           set +e
           make test
           status=$?
-          # Post a notice only if this is macOS AND tests failed
```
BenjaminBossan, line 96: I think we can also remove `set +e`, `status=$?`, and `exit $status` right? So go back to the status before #2715.

BenjaminBossan (approved):
Thanks for taking care, LGTM.

PR #2912, opened 2025-11-16
Fix caching for LoRA parametrizations on nn.Parameter

jonnyli1125:
TL;DR: Simple fix for caching can save up to ~32x memory for gpt-oss-20b and ~128x for gpt-oss-120b training.

**Explanation of bug:**

In src/peft/tuners/lora/layer.py, we inject parametrizations for nn.Parameter objects using the _LoraParameterProxy class.

https://github.com/huggingface/peft/blob/9d3e0e574348babb84cff3a3f49a68738ca7b7f4/src/peft/tuners/lora/layer.py#L1874-L1885

However, the line `with nn.utils.parametrize.cached()` doesn't work as intended since the cache is released immediately as soon as you reach the end of the context manager. This means if you access that parameter multiple times, it will still re-compute the result every time, and in training mode autograd will also store all of those tensors separately.

This causes gpt-oss LoRA training to run into CUDA OOM if the experts are included in target_parameters. There is a loop in GptOssExperts that references self.gate_up_proj and self.down_proj (nn.Parameter objects) up to num_experts times.

https://github.com/huggingface/transformers/blob/66d57110f089789ae285cc9d54d3bf051123246b/src/transformers/models/gpt_oss/modeling_gpt_oss.py#L109-L126
```py
            for expert_idx in expert_hit[:]:
                ...
                gate_up = current_state @ self.gate_up_proj[expert_idx] + self.gate_up_proj_bias[expert_idx]
                ...
                out = gated_output @ self.down_proj[expert_idx] + self.down_proj_bias[expert_idx]
```

For example, in gpt-oss-20b training, if all 32 experts were activated, this would call _LoraParameterProxy.forward 32 times for every reference to self.gate_up_proj and self.down_proj, and autograd would save each result separately, causing an eventual CUDA OOM.

The fix in this PR will make the caching work as intended.

### DIFF
--- a/tests/test_target_parameters.py
+++ b/tests/test_target_parameters.py
@@ -406,8 +406,8 @@ class TestTargetParameters:
             # step. This may be a bit wasteful but it's not clear how to prevent this and overall is probably negligible
             num_forward_per_step = 2
             # Since https://github.com/huggingface/transformers/pull/39501, one of the parameters is accessed twice per
-            # forward call, so add +1.
-            expected_call_count = num_steps * num_layers * (1 + num_params * num_forward_per_step)
+            # forward call, but we cache all calls after the first.
+            expected_call_count = num_steps * num_layers * num_params * num_forward_per_step
             assert actual_call_count == expected_call_count
 
             actual_shapes = {W.shape for W in weights}


jonnyli1125:
Tagging @BenjaminBossan since he implemented this originally in https://github.com/huggingface/peft/pull/2638

BenjaminBossan:
Nice, thanks @jonnyli1125. I already had the suspicion that the caching was not working but hadn't investigated yet. 

> autograd would save each result separately

I wanted to add a unit test to ensure that we don't have duplicates. Could you please share how you measured this? My attempt was:

```python
    def test_foobar(self):
        # TODO
        torch.manual_seed(0)
        model_id = "trl-internal-testing/tiny-GptOssForCausalLM"

        tensor_storage = []

        def pack(x):
            tensor_storage.append(x)
            return len(tensor_storage) - 1

        def unpack(x):
            return tensor_storage[x]

            config = LoraConfig(target_modules=[], target_parameters=["mlp.experts.gate_up_proj"], init_lora_weights=False)
            x = torch.arange(10).view(2, 5)
            with torch.autograd.graph.saved_tensors_hooks(pack, unpack):
                output = model(x, output_hidden_states=True)
                loss = (output.logits ** 2).sum()
                loss.backward()
```

But with this method, it seems like there is no difference between the current implementation and your suggestion.

jonnyli1125:
Sorry, I'm not actually 100% sure autograd is the culprit actually. My training setup was complicated since it also involved activation checkpointing and other stuff. I didn't get any OOM on the forward pass, but when it re-computed activations for the backward pass it ran into OOM within _LoraParamWrapper.forward. I identified _LoraParamWrapper as the problem because I injected print statements to show torch.cuda.memory_allocated() before/after every module's forward(), and I saw it accumulating memory within the GptOssExperts loop.

However, we can still see that caching doesn't work, regardless of what the source of the memory leak is:
```py
    ids = []

    def record_output_id(func):
        @functools.wraps(func)
        def wrapper(*args, **kwargs):
            res = func(*args, **kwargs)
            ids.append(id(res))
            return res

        return wrapper

    torch.manual_seed(0)
    model = AutoModelForCausalLM.from_pretrained("trl-internal-testing/tiny-GptOssForCausalLM")
    config = LoraConfig(target_modules=[], target_parameters=["gate_up_proj"], init_lora_weights=False)
    peft.tuners.lora.layer._LoraParameterProxy.forward = record_output_id(
        peft.tuners.lora.layer._LoraParameterProxy.forward
    )
    x = torch.arange(100).view(2, 50)
    model(x, output_hidden_states=True)
    print(ids)
    assert all(i == ids[0] for i in ids)
```

Without the change it should print a bunch of different memory ids, and with the change the id should always be the same.

Also side note, for some reason target_parameters = `mlp.experts.gate_up_proj` doesn't work, it has to be `gate_up_proj`.

jonnyli1125:
Sorry, I'm not actually 100% sure autograd is the culprit actually. My training setup was complicated since it also involved activation checkpointing, DeepSpeed, trl.SFTTrainer, etc. I didn't get any OOM on the forward pass, but when it re-computed activations for the backward pass it ran into OOM within _LoraParamWrapper.forward. I identified _LoraParamWrapper as the problem because I injected print statements to show torch.cuda.memory_allocated() before/after every module's forward(), and I saw it accumulating memory within the GptOssExperts loop. So something was holding onto those recomputed tensors but I couldn't tell what it was, and I just assumed it was autograd.

However, we can still see that caching doesn't work, regardless of what the source of the memory leak is:
```py
    ids = set()

    def record_output_id(func):
        @functools.wraps(func)
        def wrapper(*args, **kwargs):
            res = func(*args, **kwargs)
            ids.add(id(res))
            return res

        return wrapper

    torch.manual_seed(0)
    model = AutoModelForCausalLM.from_pretrained("trl-internal-testing/tiny-GptOssForCausalLM")
    config = LoraConfig(target_modules=[], target_parameters=["gate_up_proj"], init_lora_weights=False)
    peft.tuners.lora.layer._LoraParameterProxy.forward = record_output_id(
        peft.tuners.lora.layer._LoraParameterProxy.forward
    )
    x = torch.arange(100).view(2, 50)
    model(x, output_hidden_states=True)
    print(ids)
    assert len(ids) == model.config.num_hidden_layers
```

Without the change it should print a bunch of different memory ids, and with the change the id should always be the same.

Also side notes:
- For some reason target_parameters = `mlp.experts.gate_up_proj` doesn't work, it has to be `gate_up_proj`
- I changed input tensor size to 2x50 to guarantee a lot of experts are hit

BenjaminBossan:
Thanks for the further info. With your example, I came up with the following test:

```python
    def test_target_parameter_result_caching_works(self, monkeypatch):
        # See 2912
        # There was an issue with the caching of _LoraParameterProxy not working correctly. This test checks that the
        # results returned from the forward call are all identical to ensure they're not recomputed each time.
        torch.manual_seed(0)
        model_id = "trl-internal-testing/tiny-GptOssForCausalLM"

        tensor_storage = []

        def store_tensors_deco(fn):
            def wrapper(*args, **kwargs):
                result = fn(*args, **kwargs)
                tensor_storage.append(result)
                return result

            return wrapper

        monkeypatch.setattr(
            peft.tuners.lora.layer._LoraParameterProxy,
            "forward",
            store_tensors_deco(peft.tuners.lora.layer._LoraParameterProxy.forward),
        )

            config = LoraConfig(
                target_modules=[],
                # for simplicity, only target a single layer
                target_parameters=["0.mlp.experts.gate_up_proj"],
            )
            x = torch.arange(100).view(2, 50)  # larger input to hit many experts
            output = model(x, output_hidden_states=True)
            assert len(set(map(id, tensor_storage))) == 1

            # sanity check: a second forward call _does_ trigger a new forward
            output = model(x, output_hidden_states=True)
            assert len(set(map(id, tensor_storage))) == 2
```

If you agree, could you please add it to the bottom of `tests/test_target_parameters.py`? This should ensure that the feature works as intended.

Regarding the cache implementation itself, I wondered if the builtin PyTorch caching can still be used but needs to be applied elsewhere. Therefore, I tried adding it here:

https://github.com/huggingface/peft/blob/5fbdd672f5591f69472b46f84b12ae443f9579e6/src/peft/tuners/lora/layer.py#L2115-L2118

```python
        try:
            with nn.utils.parametrize.cached():
                yield
        finally:
            self._remove_parametrizations()
```

Interestingly, this led to two distinct tensors instead of one (without any caching: 25 distinct tensors). So it seems to work, just not quite 100%. Do you have an idea why that could be? The reason why I would prefer the PyTorch caching solution is that it should automatically take care of keeping the cache clean, whereas your implementation only works because elsewhere, we clean up the parametrization after each forward call, i.e. we introduce coupling. Perhaps, you could test my alternative solution on your initial issue to check if it would work equally well.

jonnyli1125:
The test looks good to me. I'm not sure why there are two tensors in the torch result though. Could there somehow be two distinct calls to ParamWrapper.forward or another reference to gate_up_proj somewhere within the model forward? Would that actually be the correct result somehow?

tests/test_target_parameters.py
```diff
@@ -505,3 +505,42 @@ def forward(self, x):
...
+        def store_tensors_deco(fn):
+            def wrapper(*args, **kwargs):
+                result = fn(*args, **kwargs)
+                tensor_storage.append(result)
+                return result
+
+            return wrapper
+
+        monkeypatch.setattr(
+            peft.tuners.lora.layer._LoraParameterProxy,
```
BenjaminBossan, line 528: `peft` needs to be imported at the start of the file. Could you please add it?

tests/test_target_parameters.py
```diff
@@ -505,3 +505,42 @@ def forward(self, x):
...
+            model = AutoModelForCausalLM.from_pretrained(model_id)
+            config = LoraConfig(
+                target_modules=[],
+                # for simplicity, only target a single layer
+                target_parameters=["0.mlp.experts.gate_up_proj"],
+            )
+            model = get_peft_model(model, config)
+            x = torch.arange(100).view(2, 50)  # larger input to hit many experts
+            output = model(x, output_hidden_states=True)
+            assert len(set(map(id, tensor_storage))) == 1
```
BenjaminBossan: If the caching is changed as I suggested, we will have 2 calls here and 4 below. Also, let's add a comment to make it clear what we expect here, e.g.:

```python
            # forward is called twice, once at initialization of the parametrization and once during the forward pass,
            # after which it is cached; without caching, it would be called 25 times.
```

jonnyli1125:
I just tested the change with torch.nn.utils.parametrize.cached() and the memory usage was good enough for both gpt-oss-20b and gpt-oss-120b training. I have added in all of your changes, please take a look.

BenjaminBossan (changes requested):
Thanks for the updates, we're almost good to go.

Due to the changes in this PR, another test is now failing. This is due to these lines:

https://github.com/huggingface/peft/blob/7068a7ed51758eec01bcc5154a6afde168b1dfd3/tests/test_target_parameters.py#L407-L410

As you can see, the comment says:

```
            # Since https://github.com/huggingface/transformers/pull/39501, one of the parameters is accessed twice per
            # forward call, so add +1.
```

However, now we cache the second access. Therefore, the +1 is no longer needed. Let's change that comment to:

```
            # Since https://github.com/huggingface/transformers/pull/39501, one of the parameters is accessed twice per
            # forward call, but we cache all calls after the first.
```

Then, also change this line:

```diff
-           expected_call_count = num_steps * num_layers * (1 + num_params * num_forward_per_step)
+           expected_call_count = num_steps * num_layers * (num_params * num_forward_per_step)
```

PR #2913, opened 2025-11-17
TST FIX: Issue with pickle models and caching

BenjaminBossan:
To avoid rate limits, we use a caching context, which sets offline mode after the first call, during tests to avoid calling Hugging Face Hub too frequently. However, we also use a lot of old pickle model formats (`pytorch_model.bin`) during testing. These are converted on the fly on the hub and then loaded using the safetensors format.

This used to work even in offline context but it no longer does (which is arguably the correct way to handle this). As a consequence, these models can no longer loaded within the caching context.

This PR switches these models to their safetensors variants, which are stored in `peft-internal-testing` (tokenizers also had to be moved there to be able to use the same `model_id`). If the model is already using safetensors in `hf-internal-testing` or `trl-internal-testing`, they were left as they are.

While working on this, we also remove a few, rarely used model architectures from the decoder model test suite.

Should resolve the [current failures on transformers main](https://github.com/huggingface/peft/actions/runs/19416523163).

### DIFF
--- a/tests/test_vision_models.py
+++ b/tests/test_vision_models.py
@@ -40,7 +40,7 @@ from peft import (
     get_peft_model,
 )
 
-from .testing_utils import hub_online_once, load_cat_image
+from .testing_utils import load_cat_image
 
 
 CONFIGS = {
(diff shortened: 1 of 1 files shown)


tests/test_hub_features.py
```diff
@@ -100,8 +100,6 @@ def test_save_and_load_base_model_revision(self, tmp_path):
         output_revision_loaded = peft_model_revision_loaded(test_inputs).logits
         assert torch.allclose(output_revision, output_revision_loaded)
 
-    # TODO remove when/if Hub is more stable
-    @pytest.mark.xfail(reason="Test is flaky on CI", raises=ValueError)
```
BenjaminBossan, line 104: Not strictly related to this PR, but I saw the comment and tried removing the xfail and it passed.

tests/test_decoder_models.py
```diff
@@ -59,21 +59,19 @@
...
 
 
+# Note: some models from peft-internal-testing are just the safetensors versions of hf-internal-testing
 PEFT_DECODER_MODELS_TO_TEST = [
-    "hf-internal-testing/tiny-random-OPTForCausalLM",
-    "hf-internal-testing/tiny-random-GPT2LMHeadModel",
-    "hf-internal-testing/tiny-random-BloomForCausalLM",
-    "hf-internal-testing/tiny-random-gpt_neo",
-    "hf-internal-testing/tiny-random-GPTJForCausalLM",
-    "hf-internal-testing/tiny-random-GPTBigCodeForCausalLM",
```
BenjaminBossan, line 68: As discussed internally, dropping a few, rarely used models.

Issue #2914, opened 2025-11-18
Loading AWQ weight PeftMixedModel crashes


Tfloow:
### System Info

Google colab T4 GPU
python 3.12

<details>
<summary>Pip packages</summary>
<pre>$ absl-py==1.4.0
absolufy-imports==0.3.1
accelerate==1.11.0
aiofiles==24.1.0
aiohappyeyeballs==2.6.1
aiohttp==3.13.2
aiosignal==1.4.0
alabaster==1.0.0
albucore==0.0.24
albumentations==2.0.8
ale-py==0.11.2
alembic==1.17.1
altair==5.5.0
annotated-doc==0.0.4
annotated-types==0.7.0
antlr4-python3-runtime==4.9.3
anyio==4.11.0
anywidget==0.9.19
argon2-cffi==25.1.0
argon2-cffi-bindings==25.1.0
array_record==0.8.2
arrow==1.4.0
arviz==0.22.0
astropy==7.1.1
astropy-iers-data==0.2025.11.10.0.38.31
astunparse==1.6.3
atpublic==5.1
attrs==25.4.0
audioread==3.1.0
Authlib==1.6.5
autoawq==0.2.9
autograd==1.8.0
babel==2.17.0
backcall==0.2.0
beartype==0.22.5
beautifulsoup4==4.13.5
betterproto==2.0.0b6
bigframes==2.28.0
bigquery-magics==0.10.3
bitsandbytes==0.48.2
bleach==6.3.0
blinker==1.9.0
blis==1.3.0
blobfile==3.1.0
blosc2==3.11.0
bokeh==3.7.3
Bottleneck==1.4.2
bqplot==0.12.45
branca==0.8.2
brotli==1.2.0
build==1.3.0
CacheControl==0.14.3
cachetools==5.5.2
catalogue==2.0.10
certifi==2025.10.5
cffi==2.0.0
chardet==5.2.0
charset-normalizer==3.4.4
chex==0.1.90
clarabel==0.11.1
click==8.3.0
cloudpathlib==0.23.0
cloudpickle==3.1.2
cmake==3.31.6
cmdstanpy==1.3.0
colorcet==3.1.0
colorlover==0.3.0
colour==0.1.5
community==1.0.0b1
confection==0.1.5
cons==0.4.7
contourpy==1.3.3
cramjam==2.11.0
cryptography==43.0.3
cuda-bindings==12.9.4
cuda-core==0.3.2
cuda-pathfinder==1.3.2
[...]

PR #2919, opened 2025-11-19
Added an experimental notebook demonstrating usage of `PeftMixedModel`

Tfloow:
In reference to my https://github.com/huggingface/peft/issues/2914#issuecomment-3547777863 . Here is a simple proof of concept of a research I am pursuing. To simulate possible Hardware gains from this approach, I used the  `PeftMixedModel`. 

I realized there was no notebook using it so I am suggesting here a quick code snippet coming from my work. This shows a simple way of using various LoRA adapters for different tasks.

### DIFF
--- a/examples/lora_mixed/lora_mixed_peft.ipynb
+++ b/examples/lora_mixed/lora_mixed_peft.ipynb
@@ -11,11 +11,13 @@
         "\n",
         "This notebook showcases an experimental form of extreme bit quantization. It uses right-bit shifting simulation which yields lots of potential for Hardware-Centric TPU and other embedded AI accelerator.\n",
         "\n",
-        "![Current bit-shifted quantization](https://github.com/Tfloow/peft/blob/main/examples/lora_mixed/fig/image.png?raw=1)\n",
+        "![Current bit-shifted quantization](https://github.com/huggingface/peft/blob/main/examples/lora_mixed/fig/image.png?raw=1)\n",
         "\n",
         "This model could reduce the RAM usage of the model but is designed first to reduce Memory traffic from cache to DRAM. Bit-shifting is inexpensive in hardware and the major bottleneck during inference is the data traffic. To reduce the memory trafic, we only transfer part of the weight and smaller LoRA weights that get reconstructed on the fly.\n",
         "\n",
-        "For this research, the module `PeftMixedModel` was used to apply various LoRA weights to different shifted groups. This is based on [AWQ](https://arxiv.org/abs/2306.00978) [Lin et al 2023] quantized LLama-3.1 8B, thanks to  [#2914](https://github.com/huggingface/peft/issues/2914#issuecomment-3547905030).\n",
+        "For this research, the module `PeftMixedModel` was used to apply various LoRA weights to different shifted groups. This is based on [AWQ](https://arxiv.org/abs/2306.00978) [Lin et al 2023] quantized LLama-3.1 8B, thanks to [#2914](https://github.com/huggingface/peft/issues/2914#issuecomment-3547905030).\n",
+        "\n",
+        "For efficient training, [EVA](https://huggingface.co/papers/2410.07170) [Paischer et al 2024] was used. EVA is based on Singular Value Decomposition (SVD) which helps the optimization algorithm when initiating the LoRA weights.\n",
         "\n",
         "## How to run the notebook?\n",
         "\n",
@@ -24,6 +26,10 @@
     },
     {
       "cell_type": "markdown",
+      "id": "WGcNznX8JCUh",
+      "metadata": {
+        "id": "WGcNznX8JCUh"
+      },
       "source": [
         "### Last tested requirements\n",
         "<details>\n",
@@ -711,11 +717,7 @@
         "zstandard==0.25.0\n",
         "</pre>\n",
         "</details>"
-      ],
-      "metadata": {
-        "id": "WGcNznX8JCUh"
-      },
-      "id": "WGcNznX8JCUh"
+      ]
     },
     {
       "cell_type": "code",
@@ -732,7 +734,8 @@
         "    !pip install huggingface_hub\n",
         "\n",
         "# Do this only in Colab notebooks\n",
-        "!pip install --no-deps bitsandbytes peft\n",
+        "# Install the latest version of PEFT from the main branch\n",
+        "!pip install git+https://github.com/huggingface/peft.git\n",
         "!pip install sentencepiece protobuf \"datasets>=3.4.1\" huggingface_hub hf_transfer\n",
         "!pip install --no-deps unsloth\n",
         "!pip install evaluate\n",
@@ -742,17 +745,15 @@
     },
     {
       "cell_type": "code",
+      "execution_count": 2,
+      "id": "d82e6d5b",
+      "metadata": {},
+      "outputs": [],
       "source": [
         "if \"COLAB_\" not in \"\".join(os.environ.keys()):\n",
-        "  from huggingface_hub import login\n",
-        "  login() # To access gated repo"
-      ],
-      "metadata": {
-        "id": "s-E9mVHQIIYu"
-      },
-      "id": "s-E9mVHQIIYu",
-      "execution_count": null,
-      "outputs": []
+        "    from huggingface_hub import login\n",
+        "    login() # To access gated repo"
+      ]
     },
     {
       "cell_type": "code",
@@ -766,7 +767,6 @@
         "import torch\n",
         "from datasets import load_dataset\n",
         "from peft import PeftMixedModel\n",
-        "import bitsandbytes as bnb\n",
         "from tqdm.notebook import tqdm\n",
         "import math\n",
         "from transformers import AutoModelForCausalLM, AutoTokenizer, AwqConfig"
@@ -774,7 +774,7 @@
     },
     {
       "cell_type": "code",
-      "execution_count": null,
+      "execution_count": 4,
       "id": "d0d0a101",
       "metadata": {
         "id": "d0d0a101"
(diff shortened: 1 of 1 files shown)


examples/lora_mixed/lora_mixed_peft.ipynb
```diff
@@ -0,0 +1,1174 @@
...
+    {
+      "cell_type": "markdown",
+      "id": "450eeb9c",
+      "metadata": {
+        "id": "450eeb9c"
+      },
+      "source": [
+        "# Example of loading various of LoRA PEFT models\n",
+        "\n",
+        "This notebook showcases an experimental form of extreme bit quantization. It uses right-bit shifting simulation which yields lots of potential for Hardware-Centric TPU and other embedded AI accelerator.\n",
```
BenjaminBossan, line 12: I can't say I understand what is happening here :D Is there perhaps a paper that I could check?

examples/lora_mixed/lora_mixed_peft.ipynb
```diff
@@ -0,0 +1,1174 @@
...
+        "\n",
+        "![Current bit-shifted quantization](https://github.com/Tfloow/peft/blob/main/examples/lora_mixed/fig/image.png?raw=1)\n",
```
BenjaminBossan: Let's link to the image on the PEFT repo (even though it doesn't exist there yet).

examples/lora_mixed/lora_mixed_peft.ipynb
```diff
@@ -0,0 +1,1174 @@
...
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "id": "af62ca32",
+      "metadata": {
+        "id": "af62ca32"
+      },
+      "source": [
+        "![Current bit-shifted quantization](https://github.com/Tfloow/peft/blob/main/examples/lora_mixed/fig/image.png?raw=1)\n",
```
BenjaminBossan: Let's link to the image on the PEFT repo (even though it doesn't exist there yet).

examples/lora_mixed/lora_mixed_peft.ipynb
```diff
@@ -0,0 +1,1174 @@
...
+        "    )\n",
+        "  except Exception as e:\n",
+        "    print(\"Trying to load again\")\n",
+        "    model = AutoModelForCausalLM.from_pretrained(\n",
+        "        model_name,\n",
+        "        device_map=\"cuda:0\",   # Put optimized layers on GPU\n",
+        "        torch_dtype=torch.float16,\n",
+        "        trust_remote_code=True,\n",
+        "        dtype=torch.float16,\n",
+        "        #quantization_config=quantization_config\n",
```
BenjaminBossan: Is it intended that `quantization_config` is commented out? If yes, how about just removing it completely?

examples/lora_mixed/lora_mixed_peft.ipynb
```diff
@@ -0,0 +1,1174 @@
...
+      "execution_count": null,
+      "id": "2cb680e8",
+      "metadata": {
+        "id": "2cb680e8"
+      },
+      "outputs": [],
+      "source": [
+        "# @title Apply a bit mask on weight to simulate 3 bits, 2 bits and 1 bit weight\n",
+        "\n",
+        "def apply_quantization(model, simulated_quantization, target_modules, quant_type='awq'):\n",
```
BenjaminBossan, line 915: This feels like magic ;-)

examples/lora_mixed/lora_mixed_peft.ipynb
```diff
@@ -0,0 +1,1174 @@
...
+        "    instructions = examples[\"instruction\"]\n",
+        "    inputs       = examples[\"input\"]\n",
+        "    outputs      = examples[\"output\"]\n",
+        "    texts = []\n",
+        "    for instruction, input, output in zip(instructions, inputs, outputs):\n",
+        "        # Must add EOS_TOKEN, otherwise your generation will go on forever!\n",
+        "        text = alpaca_prompt.format(instruction, input, output) + EOS_TOKEN\n",
+        "        texts.append(text)\n",
+        "    return { \"text\" : texts, }\n",
+        "pass\n",
```
BenjaminBossan, line 1078: not needed

BenjaminBossan (changes requested):
Thanks for providing this notebook. I haven't run it yet, but I did a review with what's there. My main issue right now is that I don't yet understand what's going on, do you have further resources so that I can read up?

When you check in the notebook, it would also be nice if you could execute it first, so that users can see the input right there without the need to run it themselves.

Regarding the use of `PeftMixedModel`, I still don't understand why it's needed. The two adapters you use are both LoRA, so I would have thought they should work together with a normal `PeftModel`. What is the error?

Also, out of curiosity, did you use EVA for these adapters?

More of a nit, but the notebook has a bit of an ugly mix of two and four spaces, could you please check that? I think [ruff](https://docs.astral.sh/ruff/configuration/#jupyter-notebook-discovery) should be able to fix it for you.

examples/lora_mixed/lora_mixed_peft.ipynb
Tfloow: Yes sorry it's a relica of previous code snippet

examples/lora_mixed/lora_mixed_peft.ipynb
Tfloow: Not yet, it's part of my work for my Master's thesis! I used the PEFT and SFT libraries to simulate on a high level a possible hardware "quantization mechanism".

Tfloow:
Thank you for your code review.

> Thanks for providing this notebook. I haven't run it yet, but I did a review with what's there. My main issue right now is that I don't yet understand what's going on, do you have further resources so that I can read up?

Not yet, it is simply an experimental research project for my Master Thesis where I used PEFT library to produce some high level simulation.


> When you check in the notebook, it would also be nice if you could execute it first, so that users can see the input right there without the need to run it themselves.

Yes I will try that, I will need to use a patched version of PEFT which I have yet to see how I can import it properly. 

> Regarding the use of `PeftMixedModel`, I still don't understand why it's needed. The two adapters you use are both LoRA, so I would have thought they should work together with a normal `PeftModel`. What is the error?

I need to apply various LoRA adapter on different weight. Those LoRA adapters for each groups are trained separately which is why I need to apply them one after the other. But they must be activated at the same time which is not supported by 'PeftModel`. If I understand correctly, in `PeftModel` I can load multiple adapters but only apply one at a given time. 

> Also, out of curiosity, did you use EVA for these adapters?

Yes, I did use EVA

```py
lora_config = LoraConfig(
    task_type=TaskType.CAUSAL_LM, # type of task to train on
    init_lora_weights="eva", 
    target_modules = target_modules, # Specific module type to target
    inference_mode=False, # set to False for training
    r=32, # dimension of the smaller matrices
    lora_alpha=32, # scaling factor
    lora_dropout=0.1 # dropout of LoRA layers
)
```

> More of a nit, but the notebook has a bit of an ugly mix of two and four spaces, could you please check that? I think [ruff](https://docs.astral.sh/ruff/configuration/#jupyter-notebook-discovery) should be able to fix it for you.

Yes I can check this out, I agree the 2 or 4 space indent is not clean. IDK why colab gave me automatic 2/4 lines not consistently.

BenjaminBossan:
> Not yet, it is simply an experimental research project for my Master Thesis where I used PEFT library to produce some high level simulation.

Exciting. Anything you can share, like past related research, would be helpful.

> Yes I will try that, I will need to use a patched version of PEFT which I have yet to see how I can import it properly.

You can install PEFT with `python -m pip install git+https://github.com/huggingface/peft.git` to get the latest version from main. But if we can make this work with `PeftModel`, that should not be required, right?

> I need to apply various LoRA adapter on different weight. Those LoRA adapters for each groups are trained separately which is why I need to apply them one after the other. But they must be activated at the same time which is not supported by 'PeftModel`. If I understand correctly, in `PeftModel` I can load multiple adapters but only apply one at a given time.

It is possible to activate multiple adapters at once by calling `model.base_model.set_adapter(["adapter1", "adapter2"])`.

> Yes, I did use EVA

Nice, could be worth a short mention.

> Yes I can check this out, I agree the 2 or 4 space indent is not clean. IDK why colab gave me automatic 2/4 lines not consistently.

Google being idiosyncratic :)

Tfloow:
> Exciting. Anything you can share, like past related research, would be helpful.

You can check out [ANDA](https://arxiv.org/abs/2411.15982) which is a bitplane accelerator that also quantize the Activation for efficient Hardware accelerator. I am trying to improve this as part of my studies. It's more EE focused than purely ML but an interesting read for sure.


> You can install PEFT with `python -m pip install git+https://github.com/huggingface/peft.git` to get the latest version from main. But if we can make this work with `PeftModel`, that should not be required, right?
>
> It is possible to activate multiple adapters at once by calling `model.base_model.set_adapter(["adapter1", "adapter2"])`.

I remember getting a `list is not hashable type` and from the [doc](https://huggingface.co/docs/peft/main/en/package_reference/peft_model#peft.PeftModel.set_adapter), I can only pass a string. It also mentions that "*Only one adapter can be active at a time.*"

BenjaminBossan:
> You can check out [ANDA](https://arxiv.org/abs/2411.15982) which is a bitplane accelerator that also quantize the Activation for efficient Hardware accelerator.

I'll check it out, thanks.

> I remember getting a `list is not hashable type` and from the [doc](https://huggingface.co/docs/peft/main/en/package_reference/peft_model#peft.PeftModel.set_adapter), I can only pass a string. It also mentions that "_Only one adapter can be active at a time._"

Give it a try ;-) **Edit**: More context [here](https://huggingface.co/docs/peft/developer_guides/troubleshooting#using-multiple-adapters-at-the-same-time).

Tfloow:
> Give it a try ;-) **Edit**: More context [here](https://huggingface.co/docs/peft/developer_guides/troubleshooting#using-multiple-adapters-at-the-same-time).

Seems to be working too.

```
TunerModelStatus(base_model_type='LlamaForCausalLM', adapter_model_type='LoraModel', peft_types={'adapter_qk_up': 'LORA', 'adapter_v_proj': 'LORA'}, trainable_params=56623104, total_params=1107562496, num_adapter_layers=160, enabled=True, active_adapters=['adapter_qk_up', 'adapter_v_proj'], merged_adapters=[], requires_grad={'adapter_qk_up': True, 'adapter_v_proj': True}, available_adapters=['adapter_qk_up', 'adapter_v_proj'], devices={'adapter_qk_up': ['cuda'], 'adapter_v_proj': ['cuda']})
```

mmmh, then I could perhaps use the PeftModel.

BenjaminBossan:
> mmmh, then I could perhaps use the PeftModel.

Yes, could you please switch to `PeftModel`, the usage of `PeftMixedModel` is not needed here and could be confusing for some readers.

Tfloow:
Mmmh, then I think adding this notebook wouldn't provide any significant benefits to `Peft` users, I will close this pull request.

PR #2920, opened 2025-11-19
FIX Move further models to safetensors

BenjaminBossan:
This is a continuation of #2913. Check that PR for explanation. Some models were missed, this should now hopefully cover all of them.


tests/test_tuners_utils.py
```diff
@@ -178,7 +179,7 @@
 BNB_TEST_CASES = [(x + y) for x in MAYBE_INCLUDE_ALL_LINEAR_LAYERS_TEST_CASES for y in BNB_QUANTIZATIONS]
 
 
-class PeftCustomKwargsTester:
```
BenjaminBossan, line 181: Unrelated change: These tests were not executed at all because pytest requires the test class name to start with `Test`.

PR #2937, opened 2025-12-04
FIX: Some GPU tests failing due to transformers v5

BenjaminBossan:
This PR works on some of the GPU tests that are now failing, but others are still open.

1. Mistral fine-tuning: Fails with `DataParallel` on two GPUs. This is actually not something we generally support, so just skipping this test.
2. Offload tests: Offloading now requires an `offload_folder` argument, previously the weights were just stored in the HF cache. This resolves one error in those tests but they still fail at a later step.
3. Offloat tests: Refactored for pytest style. Just boyscout rule.

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -2375,18 +2375,19 @@ class PeftGPTQGPUTests(unittest.TestCase):
 
 
 @require_non_cpu
-class OffloadSaveTests(unittest.TestCase):
-    def setUp(self):
-        self.causal_lm_model_id = "gpt2"
+class TestOffloadSave:
+    causal_lm_model_id = "gpt2"
 
-    def tearDown(self):
+    @pytest.fixture(scope="class")
+    def tear_down(self):
         r"""
         Efficient mechanism to free GPU memory after each test. Based on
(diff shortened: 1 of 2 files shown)


tests/test_gpu_examples.py
```diff
@@ -2403,22 +2404,27 @@ def test_offload_load(self):
...
-            input_tokens = tokenizer.encode("Four score and seven years ago", return_tensors="pt")
-            output = lora_model(input_tokens)[0]
-
-            # load the model with device_map
-            offloaded_model = AutoModelForCausalLM.from_pretrained(self.causal_lm_model_id, device_map=device_map)
-            assert len({p.device for p in offloaded_model.parameters()}) == 2  # 'cpu' and 'meta'
-            offloaded_lora_model = PeftModel.from_pretrained(offloaded_model, tmp_dir, max_memory=memory_limits).eval()
-            offloaded_output = offloaded_lora_model(input_tokens)[0]
+        model.save_pretrained(tmp_path)
+        del model
```
BenjaminBossan, line 2408: Not strictly necessary, but if the model is reloaded below, to me it's more obvious when the original model is deleted first.

PR #2938, opened 2025-12-04
Test cleaning pytest caches

githubnemo:
Currently we run out of disk space when packing the model cache update on ubuntu 3.10 runs so in an effort to make space, we'll delete the temp dir pytest uses after the testing has finished.

Also set the CI flag so that test summaries are not truncated which might be better for inspection of errors.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,6 +48,12 @@ jobs:
         with:
           persist-credentials: false
       - name: Make space for cache + models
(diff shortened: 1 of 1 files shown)


BenjaminBossan (approved):
Thanks for this. If CI is green, it can be merged from my point of view.

githubnemo:
@BenjaminBossan fixed the space issue as it seems, please review again.

PR #2939, opened 2025-12-05
FEAT Add function to convert non-LoRA PEFT adapters to LoRA

BenjaminBossan:
This adds the possibility to convert a non-LoRA adapter into a LoRA adapter. Not all LoRA adapters will support this, but many will.

Conversion is not precise, there will be a loss of performance. The higher the rank, the lower the loss, but also the less efficient the adapter. Also, for now, this only supports linear layers. Still, this has some advantages:

- In PEFT, LoRA supports more features than most other methods, e.g. mixed adapter batches. Thus the converted adapter can be used with those features.
- Some downstream packages support LoRA adapters, but not other PEFT methods, e.g. Diffusers. The conversion allows to use a non-LoRA adapter with those packages.

Users can pass a fixed rank for the LoRA adapter or a float that will use a dynamic rank based on the threshold of the contribution of the singular values. I ran an experiment with converting a LoHa model trained on MetaMath to LoRA (see the script included in the PR). The LoHa model's test accuracy was 41.85%, running at 12.25 GB. This is the test accuracy of the converted LoRAs:

| rank | trainable parameters | test accuracy (%) | memory reserved (max, GB) |
|------|----------------------|-------------------|---------------------------|
| 8    | 2293760              | 37.60             | 13.80                     |
| 16   | 4587520              | 38.89             | 13.54                     |
| 32   | 9175040              | 40.11             | 13.80                     |
| 64   | 18350080             | 39.20             | 13.57                     |
| 0.4  | 2428928              | 37.60             | 13.80                     |
| 0.5  | 4761600              | 40.18             | 13.80                     |
| 0.6  | 8857600              | 39.42             | 13.80                     |
| 0.7  | 16230400             | 39.04             | 13.54                     |

## For the future

1. Convert PEFT methods that are non-trivial, like IA³; in most cases, that shouldn't be too hard
2. Convert layers other than linear layers
3. Option to push the converted LoRA directly to the hub

Regarding 1., for PEFT methods that do not work according to `W' = W_0 + dW`, we have to find some workarounds, as we cannot simply use `module.get_delta_weight`. So let's say the method uses `W' = W_0 * dW` instead. We can reformulate this as `W' = W_0 * dW = W_0 + dW2`. As we're actually interested in `dW2`, we can isolate this as `dW2 = W_0 * dW - W_0`. I would suggest that we implement a new method on the tuner layer to implement this:

```python
# on the base class, `get_additive_delta_weight` would just call `get_delta_weight`
def get_additive_delta_weight(self, adapter_name):
    delta_weight = self.get_delta_weight(adapter_name)
    base_weight = self.get_base_layer().weight
    return base_weight * delta_weight - base_weight
```

Then, in the LoRA conversion function:

```diff
- delta_weight = module.get_delta_weight(adapter_name)
+ delta_weight = module.get_additive_delta_weigh(adapter_name)
```

This way, we can continue using `get_delta_weight` where applicable while leaving the door open for PEFT methods to implement a custom method specifically for LoRA conversion. LMK if you have a better idea for how to implement that.

Of course, there can be PEFT methods that cannot be expressed at all in terms of `dW`, which will be impossible to convert (most prominently the prompt learning methods).

## Unrelated changes

- I noticed that the VB-LoRA layer had no `__repr__`, so it was added.
- The return type of `set_peft_model_state_dict` was incorrect


src/peft/tuners/lora/conversion.py
```diff
@@ -36,7 +37,10 @@ def _find_cutoff_index(S: torch.Tensor, threshold: float) -> int:
     energy = S**2
     cs = torch.cumsum(energy, dim=0)
     total = cs[-1]
-    k = (cs / total > threshold).nonzero()[0].item()
+
+    cutoff = threshold * total
+    # smallest index i with cs[i] >= cutoff
+    k = torch.searchsorted(cs, cutoff).item()
```
sayakpaul: If this is a performance-related change, then note that `.item()` would still be a blocking call.

src/peft/tuners/lora/conversion.py
```diff
@@ -0,0 +1,348 @@
...
+        model.named_modules(), disable=not progressbar, desc="Converting to LoRA", total=num_modules_total
+    ):
+        if not isinstance(module, BaseTunerLayer):
+            continue
+
+        lora_A, lora_B, effective_rank = _convert_module_to_lora(module, rank=rank, adapter_name=adapter_name)
+        if effective_rank == 0:
+            # This shouldn't really happen, as we ensure that the rank is greater than 0 (int) or, for tresholds
+            # (float), at least one SV is included. But better be safe than sorry, as, in principle, it is fine to
+            # exclude some layers.
```
sayakpaul: Perhaps we log with a `logging.debug()`?

src/peft/tuners/lora/conversion.py
```diff
@@ -0,0 +1,348 @@
...
+            "Did not convert a single layer, this means that something went wrong. Please open an issue: "
+            "Please open an issue: https://github.com/huggingface/peft/issues"
+        )
+
+    ##################
+    # NON-LORA PARTS #
+    ##################
+
+    if (peft_config is not None) and getattr(peft_config, "modules_to_save", None):
+        # logic to take care of modules_to_save; might not cover all edge cases, like sharded model
```
sayakpaul, line 283: If we can detect it's a sharded model, would it make sense to error out?

src/peft/tuners/lora/conversion.py
```diff
@@ -0,0 +1,348 @@
...
+    """Convert a single BaseTunerLayer's adapter weight to a LoRA weight, return A, B, and the effective rank."""
+    if not hasattr(module, "get_delta_weight"):
+        # if we arrive here, it means that the layer actually does not support LoRA conversion, which should not happen
+        raise TypeError(
+            f"Module of type {type(module)} does not have a get_delta_weight method, which is required for conversion. "
+            "Please open an issue: https://github.com/huggingface/peft/issues"
+        )
+
+    delta_weight = module.get_delta_weight(adapter_name)
+    U, S, V = torch.linalg.svd(delta_weight, full_matrices=False)
```
sayakpaul, line 58: Thinking out loud. 

SVD is one choice (and also a time-consuming one). But there is some potential to also let the users experiment with others. We could have truncated SVD and randomized SVD as alternatives.

I did some experimentation here:
https://huggingface.co/sayakpaul/flux-lora-resizing

src/peft/tuners/lora/conversion.py
```diff
@@ -0,0 +1,348 @@
...
+
+    cutoff = threshold * total
+    # smallest index i with cs[i] >= cutoff
+    k = torch.searchsorted(cs, cutoff).item()
+    # k is the index of the first item that surpasses the threshold; since we want to include it, add + 1
+    return k + 1
+
+
+@torch.no_grad()
+def _convert_module_to_lora(
```
sayakpaul, line 49: Performance note:

We could let the users pass a `torch.compile` config to potentially speed up the process.

I would by default use the `"max-autotune-no=cudagraphs"` mode in `torch.compile` for cases like this.

src/peft/tuners/lora/conversion.py
```diff
@@ -0,0 +1,348 @@
...
+    if S.dim() != 1:
+        raise ValueError("Input vector must be 1d.")
+
+    energy = S**2
+    cs = torch.cumsum(energy, dim=0)
+    total = cs[-1]
+
+    cutoff = threshold * total
+    # smallest index i with cs[i] >= cutoff
+    k = torch.searchsorted(cs, cutoff).item()
```
sayakpaul, line 43: Performance note: the `item()` call is data-dependent and can be blocking. If the process works without `item()` we can remove it.

src/peft/tuners/lora/conversion.py
```diff
@@ -0,0 +1,348 @@
...
+def _convert_module_to_lora(
+    module: BaseTunerLayer, rank: int | float, adapter_name: str = "default"
+) -> tuple[torch.Tensor, torch.Tensor, int]:
+    """Convert a single BaseTunerLayer's adapter weight to a LoRA weight, return A, B, and the effective rank."""
+    if not hasattr(module, "get_delta_weight"):
+        # if we arrive here, it means that the layer actually does not support LoRA conversion, which should not happen
+        raise TypeError(
+            f"Module of type {type(module)} does not have a get_delta_weight method, which is required for conversion. "
+            "Please open an issue: https://github.com/huggingface/peft/issues"
+        )
```
sayakpaul: Would prefer to raise it from `convert_to_lora()`.

tests/test_lora_conversion.py
```diff
@@ -0,0 +1,513 @@
...
+        msg = re.escape("The chosen rank 123 is larger than the weight shape (16), please choose a lower rank")
+        with pytest.raises(ValueError, match=msg):
+            convert_to_lora(lokr_model, rank=123)
+
+    def test_fixed_rank_0_raises(self, lokr_model):
+        msg = "Passing a rank of 0 doesn't make sense"
+        with pytest.raises(ValueError, match=msg):
+            convert_to_lora(lokr_model, rank=0)
+
+    def test_converting_transformers_model_works(self, lokr_model, tmp_path):
```
sayakpaul, line 213: Check for a `diffusers` model too?

sayakpaul (commented):
Very useful feature. I left some performance-related comments which I think will be beneficial. 

Additionally,

> Users can pass a fixed rank for the LoRA adapter or a float that will use a dynamic rank based on the threshold of the contribution of the singular values. I ran an experiment with converting a LoKr model trained on MetaMath to LoRA. The LoKr model's test accuracy was 37.5%. This is the test accuracy of the converted LoRAs:

This is super helpful! Could we also log the original LoKr param count and possibly memory consumption for each rank? This will provide additional insights that user might want to take into consideration.

Let me know if anything is unclear.

sayakpaul:
I wish I could test it on diffusers but I don't have pre-trained non-LoRA adapters handy :/

src/peft/tuners/lora/conversion.py
BenjaminBossan: We need it as int as we store it as the rank in the LoRA config.

src/peft/tuners/lora/conversion.py
BenjaminBossan: I'm not sure I fully understand what you mean. Something like:

```python
save_as_lora(path, model, compile_kwargs={...})`
```

and then in `convert_to_lora`:

```python
if compile_kwargs:
    _convert_module_to_lora = torch.compile(_convert_module_to_lora, **compile_kwargs)
```

?

src/peft/tuners/lora/conversion.py
BenjaminBossan: Good point, moving it there

src/peft/tuners/lora/conversion.py
BenjaminBossan: Yes, for sure there are options that could work better, or at least faster, here. I wanted to get this out of the door first to see if there is any traction, then it shouldn't be too hard to add more options like you suggested.

src/peft/tuners/lora/conversion.py
BenjaminBossan: We don't have any logging like this in PEFT anywhere. I think introducing it just for this is overkill.

src/peft/tuners/lora/conversion.py
BenjaminBossan: For now, I opted to just document that it won't work with sharded models, I think we can work on better support if this feature gains traction.

tests/test_lora_conversion.py
BenjaminBossan: Yes, can do. Do you have a suggestion for a small model that can be tested? As I haven't add conv support yet, it would have to be solely linear layers for now. Or we can add it later, once conv layer support is there?

sayakpaul: Solely relying on linear is fine: https://huggingface.co/hf-internal-testing/tiny-flux2

```
from diffusers import AutoModel

model = AutoModel.from_pretrained("hf-internal-testing/tiny-flux2", subfolder="transformer")
```

src/peft/tuners/lora/conversion.py
sayakpaul: Hmm, it only gets to that point when we are saving, right? If so, I wonder if that logic (of converting from a torch scalar to float) could be moved there.

src/peft/tuners/lora/conversion.py
sayakpaul: Yup!

src/peft/tuners/lora/conversion.py
sayakpaul: Thanks for addressing it. Then I would at least leave a comment about it so that contributors could potentially explore it.

src/peft/tuners/lora/conversion.py
BenjaminBossan: done

src/peft/tuners/lora/conversion.py
BenjaminBossan: I wanted to change it but it's quite annoying to handle this as the effective rank type is then either torch tensor or int. I think the gain from omitting `.item()` won't be very big here.

src/peft/tuners/lora/conversion.py
BenjaminBossan: done

tests/test_lora_conversion.py
BenjaminBossan: Added a test for flux2

docs/source/package_reference/lora_conversion.md
```diff
@@ -0,0 +1,108 @@
...
+
+To give an example, here are some numbers that were derived on the [PEFT MetaMathQA benchmark](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA). For this, a [LoHa](https://huggingface.co/docs/peft/package_reference/loha) was used to fine-tune `meta-llama/Llama-3.2-3B` on MetaMathQA and evaluated on GSM8K. The initial LoKr adapter had rank 32, resulting in 18,350,080 trainable parameters, and a test accuracy of 41.85%. Evaluation required 12.25 GB of memory. The checkpoint was converted into LoRA with different values for the `rank`. The resulting outcome is:
+
+| rank | trainable parameters | test accuracy (%) | memory reserved (max, GB) |
```
sayakpaul: Not a necessity, but an indicator column to denote the "%" increase in the memory and "%" decrease in accuracy would even be better.

tests/test_lora_conversion.py
```diff
@@ -0,0 +1,517 @@
...
+    def get_base_model(self):
+        if self.base_model is None:
+            with hub_online_once(self.model_id):
+                self.base_model = AutoModelForCausalLM.from_pretrained(self.model_id).to(self.torch_device)
+        return copy.deepcopy(self.base_model)
+
+    @pytest.fixture
+    def lokr_model(self):
+        torch.manual_seed(0)
+        return get_peft_model(self.get_base_model(), LoKrConfig(init_weights=False))
```
sayakpaul, line 71: Should we test for other popular non-LoRA methods as well or do you think just LoKR is fine?

tests/test_lora_conversion.py
```diff
@@ -0,0 +1,517 @@
...
+        # load the converted LoRA weights
+        load_result = set_peft_model_state_dict(lora_model, state_dict)
+        assert not load_result.unexpected_keys
+
+        # sanity check the number of trainable parameters
+        num_train_params, total_params = lora_model.get_nb_trainable_parameters()
+        assert 100 < num_train_params < 0.1 * total_params
+
+        with torch.inference_mode():
+            output_converted = lora_model(inputs, output_hidden_states=True)
```
sayakpaul, line 277: Should it be similar to `output_lora`? If so, let's assert for equality?

tests/test_lora_conversion.py
```diff
@@ -0,0 +1,517 @@
...
+
+        mse_lora = self.get_mse(output_lora, output_lokr)
+        mse_converted = self.get_mse(output_converted, output_lokr)
+        assert mse_lora > 0.5
+        assert 0.0 < mse_converted < 0.1
+
```
sayakpaul, line 283: This difference seems quite tolerable

tests/test_lora_conversion.py
```diff
@@ -0,0 +1,517 @@
...
+        load_result = set_peft_model_state_dict(lora_model, state_dict)
+        assert not load_result.unexpected_keys
+
+        with torch.inference_mode():
+            output_before = lora_model(inputs).logits
+
+        # test that save_as_lora works as expected
+        save_as_lora(tmp_path, lokr_model, rank=8)
+        base_model = self.get_base_model()
+        loaded_model = PeftModel.from_pretrained(base_model, tmp_path).to(self.torch_device)
```
sayakpaul, line 326: Should we run any sanity check on the `peft_config` of `loaded_model` here? We kind of check in an earlier test. So, okay for me if we don't want to.

tests/test_lora_conversion.py
```diff
@@ -0,0 +1,517 @@
...
+            del unwrapped.peft_config
+            return unwrapped
+
+        inputs = torch.arange(10).view(1, -1).to(self.torch_device)
+        with torch.inference_mode():
+            output_lokr = lokr_model(inputs, output_hidden_states=True)
+
+        # remove the PeftModel wrapper and the peft_config attribute -- this should still work
+        unwrapped_lokr_model = unwrap(lokr_model)
+        lora_config, state_dict = convert_to_lora(unwrapped_lokr_model, rank=8)
```
sayakpaul, line 347: What would it convert if there's no existing adapter?

tests/test_stablediffusion.py
```diff
@@ -385,3 +387,76 @@ def test_load_model_low_cpu_mem_usage(self, model_id, config_cls, config_kwargs)
...
+
+            with torch.inference_mode():
+                output_converted = model_lora(**inputs)
+
+            # calculate MSE
+            mse_lora = torch.nn.functional.mse_loss(output_loha.sample, output_lora.sample)
+            mse_converted = torch.nn.functional.mse_loss(output_loha.sample, output_converted.sample)
+
+            # converted model should be significantly closer to the LoHa model than the base model
+            assert mse_lora / mse_converted > 2
```
sayakpaul, line 462: Good for me!

sayakpaul (commented):
Thanks for the further updates. Left some more comments, let me know if anything is unclear.

docs/source/package_reference/lora_conversion.md
BenjaminBossan: Good point, added

tests/test_lora_conversion.py
BenjaminBossan: The tests here are for deep testing, for broad testing across models and PEFT methods, we have the tests in `testing_common.py` (via `test_decoder_models.py`).

tests/test_lora_conversion.py
BenjaminBossan: No, it's different, `output_lora` is the LoRA output _before_ loading the converted LoRA weights.

tests/test_lora_conversion.py
BenjaminBossan: As this uses only a small dummy model, not too much can be concluded of course, but it's clear that this cannot be random, since the converted LoRA has much lower difference.

tests/test_lora_conversion.py
BenjaminBossan: Yes, my approach is not to do too many redundant checks to not distract from the main thing that is being tested.

tests/test_lora_conversion.py
BenjaminBossan: We would get an error: "Could not detect any layer that supports LoRA conversion", which is tested above.

scripts/evaluate-lora-conversion.py
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env python3
+# Copyright (c) 2025 Your Organization/Project. All rights reserved.
```
sayakpaul: Maybe needs to be updated to Hugging Face?

src/peft/tuners/c3a/layer.py
```diff
@@ -197,6 +197,9 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
         result = result.to(previous_dtype)
         return result
 
+    def supports_lora_conversion(self, adapter_name: str = "default") -> bool:
+        return True
```
sayakpaul, line 201: Not merge-blocking but is there a generic-enough way to add this method to the base tuner layer class so that we don't have to manually add it? Or do we think it's better to add it manually?

sayakpaul (approved):
Thanks for addressing my comments anf for shipping this cool utility!

scripts/evaluate-lora-conversion.py
BenjaminBossan: Good catch, copy paste error on my side, fixed.

src/peft/tuners/c3a/layer.py
BenjaminBossan: I intentionally return `False` on the `BaseTunerLayer` class as I want each method to consciously opt in. If a PEFT method returns `True` here even though `get_delta_weight` doesn't actually return a value such that `W' = W_0 + dW`, the conversion would seemingly pass but the results would be non-sense. Therefore, I wanted each decision to be made carefully.

docs/source/package_reference/lora_conversion.md
```diff
@@ -0,0 +1,129 @@
...
+Functions that allow to convert non-LoRA PEFT models to LoRA models.
+
+## Description
+
+PEFT supports dozens of different parameter effficient fine-tuning techniques. The most popular one by far is LoRA. This means that many other packages support LoRA too. For example, [Diffusers](https://huggingface.co/docs/diffusers/main/en/api/loaders/lora) allows to load LoRA adapters to change the capabilities of diffusion models. [vLLM](https://docs.vllm.ai/en/stable/features/lora/) allows serving models with LoRA adapters. This is nice but unfortunately, all the other, non-LoRA PEFT methods are rarely supported. Therefore, even if another PEFT method would work better for your specific use case, you may be prevented from using it because downstream packages offer no support.
+
+Here we present a potential solution. PEFT offers two functions, [`save_as_lora`] and [`convert_to_lora`], which allow to convert a PEFT adapter into a LoRA adapter. Not all PEFT methods support this for now, but if they do, it means you can start with the PEFT method that works best for you and then later use it as if it were a LoRA adapter.
+
+## Example
+
```
BenjaminBossan, line 16: Add a leading sentence here that there fixed and adaptive ranks and adaptive could be the best choice. Below, show a code example of adaptive rank.

tests/test_lora_conversion.py
BenjaminBossan, line 1: Add a unit test to check if the base model weights can be loaded on meta device.

docs/source/package_reference/lora_conversion.md
```diff
@@ -0,0 +1,129 @@
...
+| rank | trainable parameters | test accuracy (%) | accuracy change | memory reserved (max, GB) | memory increase |
+|------|----------------------|-------------------|-----------------|---------------------------|-----------------|
+| 8    | 2293760              | 37.60             | -4.25           | 12.41                     | 0.16            |
+| 16   | 4587520              | 38.89             | -2.96           | 12.15                     | -0.10           |
+| 32   | 9175040              | 40.11             | -1.74           | 12.41                     | 0.16            |
+| 64   | 18350080             | 39.20             | -2.65           | 12.18                     | -0.07           |
+| 0.4  | 2428928              | 37.60             | -4.25           | 12.41                     | 0.16            |
```
githubnemo: Let's add a separating line in the table, I confused it to be continuous far too often to admit :)

docs/source/package_reference/lora_conversion.md
```diff
@@ -0,0 +1,129 @@
...
+### LoRA to LoRA conversion
+
+Similar to the experiment above, we can also evaluate LoRA to LoRA conversion (i.e. LoRA compression). Here, we start with a LoRA adapter of rank 64 trained on the same setup as above with RS-LoRA. The initial adapter has 18,350,080 trainable parameters, a test accuracy of 52.92%, and requires 12.58 GB of memory for evaluation. The following table shows the results of converting this adapter to LoRA adapters of smaller rank:
+
+| rank | trainable parameters | test accuracy (%) | accuracy change | memory reserved (max, GB) | memory increase |
+|------|----------------------|-------------------|-----------------|---------------------------|-----------------|
+| 8    | 2293760              | 43.37             | -9.55           | 12.38                     | -0.20           |
+| 16   | 4587520              | 48.90             | -4.02           | 12.38                     | -0.20           |
+| 32   | 9175040              | 51.48             | -1.44           | 12.49                     | -0.09           |
+| 48   | 13762560             | 52.01             | -0.91           | 12.38                     | -0.20           |
```
githubnemo: same as above

src/peft/tuners/lora/conversion.py
```diff
@@ -0,0 +1,375 @@
...
+
+from .config import LoraConfig
+
+
+def _find_cutoff_index(S: torch.Tensor, threshold: float) -> int:
+    # assumes that the singular values are sorted
+    if S.dim() != 1:
+        raise ValueError("Input vector must be 1d.")
+
+    energy = S**2
```
githubnemo, line 37: Maybe test `S**2` vs. `S` directly.

src/peft/tuners/lora/conversion.py
```diff
@@ -0,0 +1,375 @@
...
+
+        # don't include adapter_name in key
+        state_dict[f"{name}.lora_A.weight"] = lora_A
+        state_dict[f"{name}.lora_B.weight"] = lora_B
+
+    if not state_dict:
+        # no layer was converted, which should not happen
+        raise ValueError(
+            "Did not convert a single layer, this means that something went wrong. Please open an issue: "
+            "Please open an issue: https://github.com/huggingface/peft/issues"
```
BenjaminBossan:
```suggestion
            "https://github.com/huggingface/peft/issues"
```

docs/source/package_reference/lora_conversion.md
BenjaminBossan: Done

docs/source/package_reference/lora_conversion.md
BenjaminBossan: Done

docs/source/package_reference/lora_conversion.md
BenjaminBossan: Done

src/peft/tuners/lora/conversion.py
BenjaminBossan: So I ran the experiments with S instead of S². Considering the typical fluctuation in accuracy, the resulting params vs accuracy tradeoff is basically indistinguishable:

image

The main difference between the two would be that for S, the "usable range" for the threshold is a bit squished. On the figure above, for S², the thresholds were 0.4, 0.5, 0.6, and 0.7. For S, the thresholds were 0.1, 0.15, 0.2, 0.25. Therefore, I'd say S² is a bit nicer from a user perspective, but I really don't mind either way.

src/peft/tuners/lora/conversion.py
BenjaminBossan: Done

tests/test_lora_conversion.py
BenjaminBossan: Doesn't really work right now but the reason is that we don't offer an option (yet) in PEFT to load the base model on meta while loading the PEFT adapter on non-meta. We always move the adapter to the base model device (at which point the data is lost, so moving the PEFT weights back to the real device after loading is not an option).

I think it would make sense to offer this as an option, but this is out of scope for this PR.

PR #2943, opened 2025-12-09
Intergrate MonteCLoRA (TMLR 2025 accepted) into PEFT

victor7246:
We are integrating Monte-Carlo-based robust LoRA (MonteCLoRA), accepted at TMLR 2025 - https://openreview.net/forum?id=2HFmicB8kh. This method uses probabilistic priors and mixture-of-LoRA for robust estimation of low-rank adapters. 

* Added monteclora under tuner
* Update tuner_utils with monteclora changes
* Added fine-tuning examples on GLUE classification task with RoBERTa-base model
* All quality checks are passed.


examples/monteclora_finetuning/sequence_classification_finetune_monteclora.py
```diff
@@ -0,0 +1,229 @@
...
+# Assuming MonteCLoRA is available in your local installed PEFT version
+from peft import MonteCLoraConfig, TaskType, get_peft_model
+
+
+# ----------------------------------------------------------------------------
+# 1. Custom Trainer Definition
+# ----------------------------------------------------------------------------
+# We subclass Trainer to inject the Variational Loss (KLD + Entropy)
+# into the training loop.
+class MonteCLoRATrainer(Trainer):
```
BenjaminBossan: It should be fine to add this to the proper PEFT package, so that users don't need to copy & paste this code. A good place would be the `helpers.py` file. Also, let's define a separate mixin class with `compute_loss` and `MonteCLoRATrainer` which inherits from `Trainer` and this mixin. The reason for that is so that users who use a different `Trainer` class, say, `SFTTrainer`, can use the mixin.

src/peft/tuners/monteclora/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+
+__all__ = ["MonteCLoraConfig", "MonteCLoraLinear", "MonteCLoraModel"]
+
+register_peft_method(
+    name="monteclora", prefix="lora_", config_cls=MonteCLoraConfig, model_cls=MonteCLoraModel, is_mixed_compatible=True
+)
+
+
+def __getattr__(name):
+    if (name == "Linear8bitLt") and is_bnb_available():
```
BenjaminBossan, line 37: This is all unnecessary, right? Let's remove this.

BenjaminBossan (changes requested):
Thanks for this PR to add MoneCLoRA to PEFT.

I went through the paper but must confess that I don't understand all of the math in there. I will trust you on the correctness of the implementation when it comes to this.

Regarding the PEFT integration, I think we should take a bit of a different approach. Right now, you inherit from `LoraLayer`, `LoraModel`, etc. The reason is that you can reuse most of the existing code. However, this approach can be problematic, as other code may do checks like `if isinstance(layer, LoraLayer)` assuming it gets a normal LoRA layer.

Checking the actual changes you added, I have thus a different proposal: Let's implement MonteCLoRA as a "LoRA variant". This way, users can use a normal LoRA model but with MonteCLoRA under the hood. Check the existing LoRA variants to get an idea of what is required for this:

https://github.com/huggingface/peft/blob/f2c066889543778ea3eaf43798d041bd9acda5a7/src/peft/tuners/lora/variants.py

From a user's perspective, it would look something like this:

```python
monte_clora_config = MonteCLoraConfig(...)
lora_config = LoraConfig(..., use_monteclora=monta_clora_config)  # by default use_monteclora = False
model = get_peft_model(model, lora_config)  # <= uses normal LoraModel
```

Please check if that makes sense to you.

Also, while working on this, please ensure that the copyright notice is present in all new files and with the current year. Once you finish, don't forget to call `make style`.

There are some other areas that need work, but let's get these out of the way first and then we can discuss the rest.

BenjaminBossan:
Gentle ping @victor7246

iamsecretlyflash:
Hi @BenjaminBossan , 
I am a co-author on the paper. Firstly, @victor7246 and I would like to apologise for the delay in communication.
As for the changes, we are okay with all the suggested changes and are working to incorporate them ASAP.

BenjaminBossan:
@iamsecretlyflash Please let me know when the PR is ready for review.

victor7246 (commented):
Updated PR.

src/peft/tuners/lora/config.py
```diff
@@ -693,6 +693,26 @@ class LoraConfig(PeftConfig):
...
+        metadata={
+            "help": (
+                "Enable MonteCLoRA (Monte Carlo Low-Rank Adaptation). This technique introduces variational "
+                "inference into LoRA by adding Monte Carlo sampling to the adapter weights during training. "
+                "This can improve model performance and uncertainty estimation. When enabled, you should also "
+                "provide `monteclora_config` with the MonteCLoRA hyperparameters."
+            )
+        },
+    )
+    monteclora_config: Optional[MonteCLoraConfig] = field(  # noqa: F821
```
BenjaminBossan: It never makes sense to only have `use_monteclora` and no `monteclora_config` and vice versa, right? So let's remove one of the two arguments, so e.g. only keeping `monteclora_config: Optional[MontaCloraConfig]` (default `None`).

src/peft/tuners/monteclora/config.py
```diff
@@ -0,0 +1,88 @@
...
+# limitations under the License.
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Optional
+
+
+@dataclass
+class MonteCLoraConfig:
```
BenjaminBossan:
```suggestion
class MonteCloraConfig:
```

IMO, this is easier to type and more consistent.

src/peft/tuners/monteclora/config.py
BenjaminBossan, line 1: Please move this to `tuners/lora/config.py`.

src/peft/tuners/monteclora/variant.py
BenjaminBossan, line 1: Please move this to `tuners/lora/variant.py`.

src/peft/tuners/monteclora_old/__init__.py
BenjaminBossan, line 1: `monteclora_old/` can be removed completely, right?

src/peft/tuners/monteclora/config.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    This is the sub-configuration class to store the configuration for MonteCLoRA (Monte Carlo Low-Rank Adaptation).
+    MonteCLoRA introduces variational inference into LoRA by adding Monte Carlo sampling to the adapter weights.
+
+    Args:
```
BenjaminBossan: Please explain the arguments better so that potential users don't need to be intimately familiar with the paper to use it. E.g. give a short idea when they should change the parameter and now (e.g. "Increasing this results in more trainable parameters").

src/peft/tuners/monteclora/variant.py
```diff
@@ -0,0 +1,188 @@
...
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from peft.tuners.lora.layer import Linear, LoraVariant
+from peft.tuners.tuners_utils import MonteCLoRASampler
+
+
+class MonteCLoraLinearVariant(LoraVariant):
```
BenjaminBossan:
```suggestion
class MonteCloraLinearVariant(LoraVariant):
```

src/peft/tuners/tuners_utils.py
```diff
@@ -58,8 +59,7 @@
 from ._buffer_dict import BufferDict
 
 
-_torch_supports_dtensor = version.parse(torch.__version__) >= version.parse("2.5.0")
```
BenjaminBossan, line 71: Please don't make unrelated changes like this. Probably this is the result of an incorrect merge conflict resolution.

src/peft/tuners/tuners_utils.py
```diff
@@ -2059,15 +2059,238 @@ def set_requires_grad(model, adapter_names: str | Sequence[str], requires_grad:
             module.set_requires_grad(adapter_names=adapter_names, requires_grad=requires_grad)
 
 
-def get_device_map(model) -> dict:
```
BenjaminBossan, line 2210: Please don't make unrelated changes like this. Probably this is the result of an incorrect merge conflict resolution.

src/peft/tuners/tuners_utils.py
```diff
@@ -2059,15 +2059,238 @@ def set_requires_grad(model, adapter_names: str | Sequence[str], requires_grad:
...
-        device_map = model.hf_device_map
-    else:
-        # Single-device case:
-        # Recent Transformers versions intentionally skip accelerate hooks when the
-        # device_map resolves to a single device (e.g. "cpu" or one GPU), so
-        # hf_device_map is not set. All parameters are guaranteed to be on the
-        # same device, which can be inferred from the first parameter.
-        device_map = {"": next(model.parameters()).device}
-    return device_map
+class BufferedMonteCLoRASampler:
```
BenjaminBossan: Please move these MonteCLoRA-specific classes to a dedicated sub-module, e.g. `tuners/lora/monteclora.py`. Also, for consistency:

```suggestion
class BufferedMonteCloraSampler:
```

src/peft/tuners/tuners_utils.py
```diff
@@ -2059,15 +2059,238 @@ def set_requires_grad(model, adapter_names: str | Sequence[str], requires_grad:
...
+
+        sample = self.buffer[self.index]
+        self.index += 1
+        return sample
+
+    def stop(self):
+        pass
+
+
+class MonteCLoRASampler(nn.Module):
```
```suggestion
class MonteCloraSampler(nn.Module):
```

src/peft/tuners/tuners_utils.py
```diff
@@ -2059,15 +2059,238 @@ def set_requires_grad(model, adapter_names: str | Sequence[str], requires_grad:
...
+                # FIX: Explicitly move samples from buffer to the current GPU
+                z_mvn = sample["z_mvn"].to(current_device)
+                z_wishart = sample["z_wishart"].to(current_device)
+                z_dirichlet = sample["z_dirichlet"].to(current_device)
+            else:
+                # Fallback if sampler fails
+                z_mvn = torch.randn((self.monteclora_n, self.in_features, self.out_features), device=current_device)
+
+                # FIX: Explicitly move Bartlett sampling result to current device
+                z_wishart = self.sampler.wish_sampler._bartlett_sampling(torch.Size()).to(current_device)
```
BenjaminBossan: Hmm, if `sample is None`, it means that `self.sampler` is also `None`, right? How can we use it here in that case?

src/peft/tuners/tuners_utils.py
```diff
@@ -2059,15 +2059,238 @@ def set_requires_grad(model, adapter_names: str | Sequence[str], requires_grad:
...
+        """
+        if self.training and self.mc_training:
+            # DYNAMIC DEVICE DETECTION
+            # Use std_prior (nn.Parameter) as the source of truth for the current device
+            current_device = self.std_prior.device
+
+            sample = self.sampler.get() if self.sampler else None
+
+            if sample is not None:
+                # FIX: Explicitly move samples from buffer to the current GPU
```
BenjaminBossan: Please take care of this.

src/peft/tuners/tuners_utils.py
```diff
@@ -2059,15 +2059,238 @@ def set_requires_grad(model, adapter_names: str | Sequence[str], requires_grad:
...
+                z_mvn = sample["z_mvn"].to(current_device)
+                z_wishart = sample["z_wishart"].to(current_device)
+                z_dirichlet = sample["z_dirichlet"].to(current_device)
+            else:
+                # Fallback if sampler fails
+                z_mvn = torch.randn((self.monteclora_n, self.in_features, self.out_features), device=current_device)
+
+                # FIX: Explicitly move Bartlett sampling result to current device
```
BenjaminBossan: Please take care of this.

src/peft/tuners/monteclora/config.py
```diff
@@ -0,0 +1,88 @@
...
+        default=1e-4,
+        metadata={
+            "help": "Scaling factor for the Monte Carlo samples. Controls the magnitude of variational perturbations."
+        },
+    )
+    kl_loss_weight: float = field(
+        default=1e-5,
+        metadata={"help": "Weight for the KL divergence loss component in the variational objective."},
+    )
+    mc_training: bool = field(
```
BenjaminBossan: For my understanding, when would it make sense to have `mc_training=False`?

src/peft/tuners/tuners_utils.py
```diff
@@ -2059,15 +2059,238 @@ def set_requires_grad(model, adapter_names: str | Sequence[str], requires_grad:
...
+            # Calculate Weights
+            expert_weights_logits = self.dirichlet_reparameterization(
+                torch.exp(self.expert_weights_prior), z_dirichlet
+            )
+            expert_weights = torch.softmax(expert_weights_logits, dim=-1)
+            self.expert_weights = expert_weights
+
+            return var, self.expert_weights
+        else:
+            return -1, -1
```
BenjaminBossan: Let's add a comment that these are dummy values. Also, let's turn the `if` check around and early return this result, I think it's more readable that way.

src/peft/tuners/monteclora/variant.py
```diff
@@ -0,0 +1,188 @@
...
+        if isinstance(dropout, nn.Identity) or not module.training:
+            x_dropped = x
+        else:
+            x_dropped = dropout(x)
+
+        current_weight_A = lora_A.weight
+
+        if module.training and hasattr(sampler, "mc_training") and sampler.mc_training:
+            lora_A_vars, lora_A_wts = sampler()
+            if not isinstance(lora_A_vars, int):
```
BenjaminBossan: IMO, it's better to check if the value `!= -1` to make it clear it's a dummy value (and add a comment to explain it). However, I wonder if we can ever get -1 here? This code path is only reached when we're training with MC, in which case the sampler should never return -1, right?

src/peft/tuners/monteclora/variant.py
```diff
@@ -0,0 +1,188 @@
...
+                if torch.isnan(lora_A_vars).any() or torch.isnan(lora_A_wts).any():
+                    warnings.warn("MonteCLoRA sampling produced NaNs, using base weights.")
```
BenjaminBossan: It would be great to explain if this is something to be concerned about and if yes, what the user should do to change it.

src/peft/utils/peft_types.py
```diff
@@ -82,7 +83,7 @@ class PeftType(str, enum.Enum):
     OSF = "OSF"
     DELORA = "DELORA"
     GRALORA = "GRALORA"
-    CARTRIDGE = "CARTRIDGE"
```
BenjaminBossan, line 91: Undo removal of CARTRIDGE, remove MONTECLORA.

src/peft/helpers.py
```diff
@@ -251,6 +251,95 @@ def disable_input_dtype_casting(model: nn.Module, active: bool = True):
...
+            outputs = None
+
+        # 2. Calculate Variational Loss (KLD + Entropy) from MonteCLoRA samplers
+        var_loss_sum = 0.0
+        num_monte_layers = 0
+
+        # Iterate through modules to find MonteCLoRA samplers
+        for name, module in model.named_modules():
+            # Check if this is a MonteCLoRASampler by checking for the get_variational_loss method
+            if hasattr(module, "get_variational_loss") and module.__class__.__name__ == "MonteCLoRASampler":
```
BenjaminBossan: Why not check `isinstance(module, MonteCLoRASampler)`?

src/peft/helpers.py
```diff
@@ -251,6 +251,95 @@ def disable_input_dtype_casting(model: nn.Module, active: bool = True):
...
+                try:
+                    kl_loss, entropy_loss = module.get_variational_loss()
+                    var_loss_sum += kl_loss + entropy_loss
+                    num_monte_layers += 1
+                except Exception:
+                    # Silently ignore if get_variational_loss fails
```
BenjaminBossan: Does this happen often? Should something be done in that case?

src/peft/helpers.py
```diff
@@ -251,6 +251,95 @@ def disable_input_dtype_casting(model: nn.Module, active: bool = True):
...
+                    pass
+
+        # 3. Normalize the Variational Loss
+        regularization_loss = 0.0
```
BenjaminBossan: IMO, "regularization" is confusing here, it's just normalization, right?

tests/test_monteclora.py
BenjaminBossan, line 1: I think all these tests can be removed, as they're redundant with existing tests, please check `test_custom_models.py`. You only need to extend the test matrix to include MonteCLoRA, similar to what we have for DoRA there:

https://github.com/huggingface/peft/blob/8f73327fb967ab47b83aac604bbc2e193207d0f4/tests/test_custom_models.py#L94

Then you can run `pytest tests/test_custom_models.py -k "montaclora" -v` and check if these tests pass.

The only test we should keep is `test_variational_loss_computation`, which can also be moved to `test_custom_models.py` (please rename it to make it clear that it's a MonteCLoRA specific test).

tests/test_monteclora_old.py
BenjaminBossan, line 1: Please delete.

.gitignore
BenjaminBossan, line 1: Please remove.

tmp.md
```diff
@@ -0,0 +1 @@
+accesscheck
```
BenjaminBossan: Please remove.

examples/monteclora_finetuning/sequence_classification_finetune_monteclora.py
BenjaminBossan: Please update the example to use the `MonteCLoRATrainerMixin`.

BenjaminBossan (changes requested):
Thanks a lot for rewriting the PR to use LoRA variants. IMO this looks much nicer. I re-reviewed the code and focused on the PEFT integration and usability. Please check my comments.

src/peft/tuners/monteclora/config.py
iamsecretlyflash: Hi
I think changing this to just MontecloraConfig and Monteclora everywhere would be the best.

src/peft/tuners/lora/config.py
iamsecretlyflash: Updated

src/peft/tuners/monteclora/variant.py
iamsecretlyflash: Done

src/peft/tuners/monteclora/config.py
iamsecretlyflash: Done

src/peft/tuners/monteclora_old/__init__.py
iamsecretlyflash: Yup. Had kept it to compare and ensure we don't miss out on monteclora specific things. 
Have removed it

src/peft/tuners/monteclora/variant.py
iamsecretlyflash: Done

src/peft/tuners/tuners_utils.py
iamsecretlyflash: Hi!
Yes, you are correct. I think the original plan with this was to handle exceptions when fetching value from the sampler.
Updating it do that

src/peft/tuners/tuners_utils.py
iamsecretlyflash: Acknowledged and updated.

src/peft/tuners/tuners_utils.py
iamsecretlyflash: Acknowledged and updated.

src/peft/tuners/tuners_utils.py
iamsecretlyflash: Done

src/peft/tuners/monteclora/config.py
iamsecretlyflash: We added this in case someone wants to train for a few epochs and then train for more epochs without using monteclora

src/peft/tuners/monteclora/variant.py
iamsecretlyflash: So, this was added for us to debug the code. NaNs were not produced the last time we tested the code. We didn't remove this part yet so that in the slight change that someone gets NaNs they can reach out to us. There is nothing that user can do to fix this without going into depth of the code

src/peft/utils/peft_types.py
iamsecretlyflash: Done

src/peft/helpers.py
iamsecretlyflash: Hi!
No this does not happen. Removed this

src/peft/helpers.py
iamsecretlyflash: Agreed. Updated to "normalised_loss"

src/peft/helpers.py
iamsecretlyflash: Yup, this makes a lot of sense. Thanks for suggesting. Will update

src/peft/tuners/tuners_utils.py
iamsecretlyflash: Have removed this entirely

src/peft/tuners/monteclora/variant.py
iamsecretlyflash: So I decided to go ahead and remove mc_training. From the sampler and the variant as well. So now the code does not return -1,-1 and hence no need to add this check

tests/test_monteclora_old.py
iamsecretlyflash: Done

src/peft/tuners/monteclora/config.py
iamsecretlyflash: Done

iamsecretlyflash:
Hi @BenjaminBossan 
I rebased the codebase on the latest hugginface peft main branch because there were a lot of differences between the branch we were working on and the main branch. After doing that I added the relevant Monteclora commits and ran the tests

I have incorporated all the changes as discussed.

BenjaminBossan:
@iamsecretlyflash The PR now contains 263 changed files, the majority of which being *.pyc files. Could you please up all of those (add `__pycache__` to your `.gitignore`). Otherwise, the PR is impossible to review. Also, please avoid force pushes if possible, as that means I cannot review the difference compared to the last time I reviewed.

iamsecretlyflash:
Hi!
@BenjaminBossan 
I have cleaned up the repo.
Apologies for the mess and the force push.

BenjaminBossan:
Thanks for cleaning up, but there are still unrelated files:

- 2 `.gitignore`s
- the `.coverage`
- the checkpoints
- change to `int8_correction.py`

Could you please clean those up too?

iamsecretlyflash:
Hi @BenjaminBossan 

Remaining files cleand up as well

src/peft/tuners/lora/config.py
```diff
@@ -931,3 +943,87 @@ class LoraGAConfig:
...
+            Size of the internal buffer used by the Monte Carlo sampler (e.g. for storing recent statistics). Larger
+            values can stabilize the estimated variational parameters at the cost of additional memory; reduce this if
+            you are memory constrained.
+    """
+
+    monteclora_n: int = field(
+        default=8,
+        metadata={"help": "Number of Monte Carlo samples to use."},
+    )
+    monteclora_m: Optional[int] = field(
```
BenjaminBossan: AFAICT, this argument isn't used anyhwere.

src/peft/tuners/lora/config.py
```diff
@@ -931,3 +943,87 @@ class LoraGAConfig:
...
+            emphasis on matching the prior (stronger regularization, potentially underfitting); smaller values rely
+            more on the data likelihood (weaker regularization, potentially overfitting). Tune this if you find
+            Monteclora over- or under-regularizing the adapter.
+        buffer_size (`int`):
+            Size of the internal buffer used by the Monte Carlo sampler (e.g. for storing recent statistics). Larger
+            values can stabilize the estimated variational parameters at the cost of additional memory; reduce this if
+            you are memory constrained.
+    """
+
+    monteclora_n: int = field(
```
BenjaminBossan: Nit: How about calling this `n_samples` or `num_samples`? This tells the user directly what it means. The `monteclora_` prefix isn't necessary, as we're already inside `MontecloraConfig`.

src/peft/tuners/lora/config.py
```diff
@@ -931,3 +943,87 @@ class LoraGAConfig:
         default="stable", metadata={"help": "Scaling strategy for initialization"}
     )
     stable_gamma: int = field(default=16, metadata={"help": "Gamma parameter for stable scaling"})
+
+
+@dataclass
+class MontecloraConfig:
+    """
+    This is the sub-configuration class to store the configuration for Monteclora (Monte Carlo Low-Rank Adaptation).
```
BenjaminBossan, line 1020: Much better docs now, thanks for that. I wouldn't mind if you copy the argument description from the docstring to the help field below, as the docstring contains more details.

src/peft/tuners/lora/layer.py
```diff
@@ -241,6 +245,8 @@ def update_layer(
 
         if adapter_name in self.lora_variant:
             self.lora_variant[adapter_name].init(self, adapter_name=adapter_name, config=config, **kwargs)
+            # Some variants create adapter modules during init; align them with base-layer device/dtype too.
+            self._move_adapter_to_device_of_base_layer(adapter_name)
```
BenjaminBossan: Right above, we already call `_move_adapter_to_device_of_base_layer` and the comment explicitly states:  "call this before init of the lora variants", so let's remove the call here. If needed, ensure that the `MontecloraLinearVariant` handles device placement autonomously.

src/peft/tuners/lora/monteclora.py
```diff
@@ -0,0 +1,233 @@
+from __future__ import annotations
```
BenjaminBossan, line 15: Let's add the copyright notice.

src/peft/tuners/lora/monteclora.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+class MontecloraSampler(nn.Module):
+    """
+    The main module responsible for maintaining the variational parameters.
+    """
+
+    def __init__(
+        self,
+        in_features,
```
BenjaminBossan: Not required, but having type annotations would be nice to have.

src/peft/tuners/lora/monteclora.py
```diff
@@ -0,0 +1,233 @@
...
+        out_features,
+        monteclora_n,
+        use_entropy=True,
+        dirichlet_prior=1.0,
+        sample_scaler=3e-4,
+        kl_loss_weight=1e-5,
+        buffer_size=10,
+        device="cuda" if torch.cuda.is_available() else "cpu",
```
BenjaminBossan: Does this only work with CUDA or CPU and not with other devices? Otherwise, consider using `peft.utils.infer_device`.

src/peft/tuners/lora/monteclora.py
```diff
@@ -0,0 +1,233 @@
...
+        eps = 1e-3
+        # z_wishart and std must be on the same device
+        updated_var = std @ z_wishart @ std.T
+        updated_var = torch.diag(torch.clip(updated_var.diag(), min=eps))
+        return updated_var
+
+    def multivariate_reparameterization(self, z_mvn, cov_matrix):
+        eps = 1e-3
+        # Cholesky decomposition for reparameterization
+        with torch.amp.autocast("cuda", enabled=False):
```
BenjaminBossan, line 144: Do we need this context to explicitly _disable_ autocast for this call? If yes, could you please add a short comment on why?

src/peft/tuners/lora/monteclora.py
```diff
@@ -0,0 +1,233 @@
...
+        alpha = torch.clamp(alpha, min=eps)
+
+        mu = torch.log(alpha) - torch.log(alpha).mean()
+        # Dirichlet approximation via Normal distribution in log-space
+        diag_term = 1 / alpha * (1 - 2 / self.monteclora_n) + 1 / (self.monteclora_n**2) * (1 / alpha).sum()
+        diag_term = torch.clamp(diag_term, min=eps)
+        sigma = torch.diag(diag_term)
+        sigma = sigma + eps * torch.eye(len(alpha), device=self.expert_weights_prior.device)
+
+        with torch.amp.autocast("cuda", enabled=False):
```
BenjaminBossan, line 165: Same comment about disabling autocast.

src/peft/tuners/lora/layer.py
```diff
@@ -158,8 +158,12 @@ def update_layer(
         init_lora_weights = config.init_lora_weights
         use_rslora = config.use_rslora
         lora_bias = config.lora_bias
+        monteclora_config = config.monteclora_config
         inference_mode = config.inference_mode
 
+        if monteclora_config is not None:
+            kwargs["monteclora_config"] = monteclora_config
+
```
BenjaminBossan: This shouldn't be required and can be removed. Below we call `self.lora_variant[adapter_name].init(self, adapter_name=adapter_name, config=config, **kwargs)` and `config` contains the `monteclora_config` so no need to put it into `kwargs`.

src/peft/tuners/lora/variants.py
```diff
@@ -921,3 +923,180 @@ def merge_unsafe(module: Linear, active_adapter: str, orig_weight: torch.Tensor)
...
+        Returns:
+            result + LoRA update with Monteclora sampling applied
+        """
+        lora_A = module.lora_A[active_adapter]
+        lora_B = module.lora_B[
+            active_adapter
+        ]  # lora_B is zero in the beginning. test for stochasticity of outputs might fail because of this
+        dropout = module.lora_dropout[active_adapter]
+        scaling = module.scaling[active_adapter]
+        if active_adapter not in module.lora_monteclora_sampler:
```
BenjaminBossan, line 1188: Please add a short comment that this is because of meta device during init.

src/peft/helpers.py
```diff
@@ -251,6 +252,92 @@ def disable_input_dtype_casting(model: nn.Module, active: bool = True):
...
+            # Check if this is a MontecloraSampler by checking for the get_variational_loss method
+            if isinstance(module, MontecloraSampler):
+                kl_loss, entropy_loss = module.get_variational_loss()
+                var_loss_sum += kl_loss + entropy_loss
+                num_monte_layers += 1
+
+        # 3. Normalize the Variational Loss
+        normalised_loss = 0.0
+        if num_monte_layers > 0:
+            normalised_loss = var_loss_sum / num_monte_layers
```
BenjaminBossan: A suggestion: How about we add a new method on `LoraModel` called `_get_monteclora_loss` which does exactly this calculation (and returns 0 if no MonteCLoRA is used). Then in here and in the tests, we can simple call:

```python
monteclora_loss = model._get_monteclora_loss()
total_loss = task_loss + monteclora_loss
```

tests/test_custom_models.py
```diff
@@ -2238,16 +2275,26 @@ def test_only_params_are_updated(self, test_name, model_id, config_cls, config_k
...
+                # add variational loss
+                for name, module in model.named_modules():
+                    if isinstance(module, MontecloraSampler):
+                        kl_loss, entropy_loss = module.get_variational_loss()
+                        loss += kl_loss + entropy_loss
             loss.backward()
             optimizer.step()
 
         tol = 1e-4
+        ## higher tol for monteclora to account for randomness from sampling
```
BenjaminBossan: What does this comment refer to?

tests/test_custom_models.py
```diff
@@ -2278,6 +2326,9 @@ def test_parameters_after_loading_model(self, test_name, model_id, config_cls, c
             config_cls, PveraConfig
         ):  # needs a very small lr to not get nan in pvera_lambda_b due to high input values in this test (up to 90)
             lr = 1e-6
+
```
BenjaminBossan: Remove empty line.

BenjaminBossan (changes requested):
Thanks for updating the PR and cleaning it up. We're making good progress. I still found a couple of potential improvements in the integration, please check.

Once that is done, we should add a section to the docs (`developer_guides/lora.md`) and explain a little when and why users may want to consider using MonteCLoRA.

iamsecretlyflash:
Hi @BenjaminBossan 
Have integrated the above changes.

I have added the infer_device section as you recommended, but we have tested the code only on cuda and cpu. I will try experimenting on Apple Silicon GPUs and will update you

iamsecretlyflash:
@BenjaminBossan the code works with Apple Silicon. Have added an example as well

src/peft/tuners/lora/monteclora.py
BenjaminBossan: Still open.

src/peft/tuners/lora/monteclora.py
BenjaminBossan: Still open.

src/peft/tuners/lora/monteclora.py
```diff
@@ -0,0 +1,263 @@
...
+    def get_variational_loss(self) -> tuple[torch.Tensor | float, torch.Tensor | float]:
+        """Calculates the KL divergence and entropy loss components."""
+        if self.training:
+            kl1 = self.dirichlet_kl(torch.exp(self.expert_weights_prior))
+            kl2 = self.wishart_kl(torch.diag(torch.exp(self.std_prior)))
+            kl3 = self.multivariate_kl(self.gaussian_var_prior)
+            entropy = (self.expert_weights**2).sum() if self.use_entropy else 0
+
+            return self.kl_loss_weight * (kl1 + kl2 + kl3), entropy
+        return 0, 0
```
BenjaminBossan: Or else they're ints:

```suggestion
        return 0.0, 0.0
```

src/peft/tuners/lora/model.py
```diff
@@ -875,6 +875,40 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
 
         return tensors_lora
 
+    def _get_monteclora_loss(self) -> torch.Tensor | float:
```
BenjaminBossan, line 878: I'm wondering now: If we have multiple LoRA adapters at the same time, but only a subset of them is active, we should only return the loss associated with these adapters, right?

So maybe we should add `adapter_names` as to the function; it can default to `None`, in which case we take the `active_adapters`. Then we loop through the modules searching for `LoraLayer` instances. Next, if the module has `lora_monteclora_sampler`, we go through the `lora_monteclora_sampler`s of the active adapters of that module and collect only the loss of those. WDYT?

tests/test_custom_models.py
```diff
@@ -3657,6 +3705,26 @@ def test_mha_gradients_set_correctly(self, with_forward_call):
         assert model.base_model.model.mha.base_layer.out_proj.base_layer.weight.requires_grad is True
         assert model.base_model.model.mha.base_layer.in_proj_weight.requires_grad is True
 
+    def test_monteclora_variational_loss_computation(self):
```
BenjaminBossan, line 3708: Let's add a test for when the model is in eval mode.

docs/source/developer_guides/lora.md
```diff
@@ -234,6 +234,58 @@ Caching can thus make inference with DoRA significantly faster but it also requi
 - DoRA introduces a bigger overhead than pure LoRA, so it is recommended to merge weights for inference, see [`LoraModel.merge_and_unload`].
 - DoRA should work with weights quantized with bitsandbytes ("QDoRA"). However, issues have been reported when using QDoRA with DeepSpeed Zero2.
 
+### MonteCLoRA (Monte Carlo Low-Rank Adaptation)
+
+MonteCLoRA wraps a standard LoRA adapter with a small variational module that draws Monte Carlo samples of stochastic perturbations on top of the LoRA `A` matrix during training. Concretely, it learns variational parameters (a Wishart-based covariance, a per-sample multivariate-normal noise term, and a Dirichlet weighting over the samples) and adds the resulting averaged perturbation to `lora_A` at every forward pass. A KL-divergence + entropy term is added to the training loss to keep these variational parameters anchored to a sensible prior. At inference time the sampler is disabled and MonteCLoRA behaves exactly like a regular LoRA adapter, so there is **no extra inference cost or extra parameters to merge**. For the full method see https://huggingface.co/papers/2410.05710.
```
BenjaminBossan, line 239: Nice addition.

examples/monteclora_finetuning/test_monteclora_mps.py
```diff
@@ -0,0 +1,120 @@
+from __future__ import annotations
```
BenjaminBossan: Not quite sure why we need a separate example for MPS. Could the other example be modified to allow MonteCLoRA to train successfully with MPS?

BenjaminBossan (changes requested):
Thanks for the updates, the PR is taking a shape that's close to allow merging. I still found a couple of smaller issues and unaddressed comments. Please check them out.

src/peft/tuners/lora/monteclora.py
iamsecretlyflash: Hi @BenjaminBossan 
Had added the explanation for autocast in the code. Forgot to add them in the reply to the comments.

So we do the autocast because, during experiments, we found that Cholesky decomposition is numerically unstable in low precision (fp16/bf16) and can produce NaNs negative-definite errors when autocast downcasts the inputs. We force the decomposition to run in fp32 and cast back to the original dtype afterwards.

iamsecretlyflash: Updated return values to float

examples/monteclora_finetuning/test_monteclora_mps.py
iamsecretlyflash: Yup, makes sense. I updated examples/monteclora_finetuning/sequence_classification_finetune_monteclora.py to use infer_device. Works fine with mps

docs/source/developer_guides/lora.md
iamsecretlyflash: Thanks!

src/peft/tuners/lora/model.py
iamsecretlyflash: Hey, I agree with the suggestion, but I am a bit unclear on the implementation details. What do you mean by "module has lora_monteclora_sampler"? Can't we just look for instances of MontecloraSampler?

BenjaminBossan: > Can't we just look for instances of MontecloraSampler?

I don't think we can. IIUC, we have one `MontecloraSampler` per adapter, so the `MontecloraSampler` itself doesn't "know" if it's active or not. Only the surrounding `LoraLayer`, which contains the `MontecloraLinearVariant`, knows which adapter is active, hence my suggestion.

iamsecretlyflash: Hi @BenjaminBossan 
I made this change and updated the codebase.

iamsecretlyflash:
Hi @BenjaminBossan 
Have added the requested changes

src/peft/tuners/lora/model.py
```diff
@@ -944,6 +944,59 @@ def subtract_mutated_init(self, output_state_dict: dict[str, torch.Tensor], adap
 
         return tensors_lora
 
+    def _get_monteclora_loss(self, adapter_names: Optional[Union[str, list[str]]] = None) -> torch.Tensor | float:
+        """
+        Compute the MonteCLoRA variational regularization loss.
+
+        Iterates over all `LoraLayer` modules in the model and, for each layer, collects the KL-divergence and entropy
+        components from the `MontecloraSampler`s that correspond to the requested adapters. The aggregated loss is
+        normalized by the number of contributing samplers. Returns ``0.0`` if no matching MonteCLoRA samplers are
```
BenjaminBossan: Let's use single backticks everywhere.

src/peft/tuners/lora/variants.py
```diff
@@ -1038,3 +1040,186 @@ def merge_unsafe(module: Linear, active_adapter: str, orig_weight: torch.Tensor)
...
+
+        This adds a MontecloraSampler to the layer that will be used during forward passes to sample variational
+        perturbations for the LoRA A matrix.
+
+        Args:
+            module: The Linear LoRA layer to add Monteclora to
+            adapter_name: Name of the adapter
+            **kwargs: Must contain 'config' (the `LoraConfig`) which carries the `monteclora_config`.
+        """
+        config = kwargs.get("config")
```
BenjaminBossan: Let's add `config` to the signature, the argument should always be there, so no need for `kwargs.get`.

tests/test_custom_models.py
```diff
@@ -192,6 +194,37 @@
         LoraConfig,
         {"target_modules": ["lin0"], "target_parameters": ["lin1.weight"]},
     ),
+    ###########
+    # MonteLoRA #
+    ###########
```
```suggestion
    #############
    # MonteLoRA #
    #############
```

tests/test_custom_models.py
```diff
@@ -192,6 +194,37 @@
...
+    (
+        "Vanilla MLP 3 LoRA with Monteclora",
+        "MLP",
+        LoraConfig,
+        {
+            "target_modules": "lin1",
+            "lora_alpha": 32,
+            "monteclora_config": MontecloraConfig(num_samples=2),
+        },
+    ),
```
BenjaminBossan: I think there is no need to add 3 settings here, as the MonteCLoRA part being tested is still exactly the same. So let's delete the 2nd and 3rd setting. I could see adding a second setting with `use_entropy=True` but it's not crucial.

BenjaminBossan (changes requested):
Thanks for the new updates. I reviewed the PR again and found only a few small things, please check.

iamsecretlyflash:
Hey @BenjaminBossan 
Have made the changes

src/peft/tuners/lora/monteclora.py
```diff
@@ -0,0 +1,263 @@
...
+                self._refill_buffer()
+
+            sample = self.buffer[self.index]
+            self.index += 1
+            return sample
+        except Exception as e:
+            print(f"Error getting sample from buffer. Returning None: {e}")
+            return None
+
+    def stop(self) -> None:
```
BenjaminBossan: Why is this needed? If it can be removed, let's also remove `MontecloraSampler.eval`.

src/peft/tuners/lora/monteclora.py
```diff
@@ -0,0 +1,263 @@
...
+        """
+        try:
+            if self.index >= self.buffer_size:
+                self._refill_buffer()
+
+            sample = self.buffer[self.index]
+            self.index += 1
+            return sample
+        except Exception as e:
+            print(f"Error getting sample from buffer. Returning None: {e}")
```
BenjaminBossan: Under what circumstances would we expect this to happen? If it's not really expected, let's not return `None` but raise a proper error. Then we could also get rid of the "Fallback if sampler fails" branch.

src/peft/tuners/lora/variants.py
```diff
@@ -1038,3 +1040,185 @@ def merge_unsafe(module: Linear, active_adapter: str, orig_weight: torch.Tensor)
...
+        This adds a MontecloraSampler to the layer that will be used during forward passes to sample variational
+        perturbations for the LoRA A matrix.
+
+        Args:
+            module: The Linear LoRA layer to add Monteclora to
+            adapter_name: Name of the adapter
+            config: The `LoraConfig` which carries the `monteclora_config`.
+        """
+        monteclora_config = getattr(config, "monteclora_config", None)
+        if monteclora_config is None:
```
BenjaminBossan: This can't really happen, can it? If it were `None`, we would never call `init` to begin with.

src/peft/tuners/lora/variants.py
```diff
@@ -1038,3 +1040,185 @@ def merge_unsafe(module: Linear, active_adapter: str, orig_weight: torch.Tensor)
...
+            raise ValueError(
+                "MontecloraLinearVariant.init() requires `config.monteclora_config` to be set on the LoraConfig."
+            )
+        if isinstance(monteclora_config, dict):
```
BenjaminBossan: I don't think we can have a dict here, right?

src/peft/helpers.py
```diff
@@ -261,6 +261,87 @@ def disable_input_dtype_casting(model: nn.Module, active: bool = True):
                 module.cast_input_dtype_enabled = original_values[name]
 
 
+class MontecloraTrainerMixin:
```
BenjaminBossan, line 264: We don't have a test for this yet. Could you please add a test to `tests/test_helpers.py`? It can be a very simple test with a dummy model and data.

BenjaminBossan (changes requested):
Thanks for the new update. I reviewed the whole code again and found a few instances where it was overly defensive, which I marked. I also think some parts can be removed. Finally, let's add a test for the trainer mixin. Otherwise, the PR LGTM.

src/peft/tuners/lora/monteclora.py
iamsecretlyflash: Hi!
Had added this while the codebase was not stable. Will remove this

iamsecretlyflash:
Hi @BenjaminBossan 
Updated the codebase

tests/test_helpers.py
```diff
@@ -683,3 +691,28 @@ def __init__(self):
         assert isinstance(result["target_parameters"], list)
         assert len(result["target_parameters"]) > 0
         assert any("gate_up_proj" in name or "down_proj" in name for name in result["target_parameters"])
+
+
+class TestMontecloraTrainerMixin:
+    def test_mixin_adds_variational_loss(self):
+        class DummyTrainer:
```
BenjaminBossan: Let's use the real `Trainer`, otherwise it's not really testing the integration between `Trainer` and `MontecloraTrainerMixin`. The model can be a small dummy model like `"peft-internal-testing/tiny-random-OPTForCausalLM"`. Use the `with hub_online_once(model_id):` context, which is required for efficient caching. Here is an example: https://github.com/huggingface/peft/blob/9403de1a75e798d65644f84e43ddbfd90c6bd492/tests/test_decoder_models.py#L486-L489

BenjaminBossan (changes requested):
Thanks for simplifying the code and adding a test for the mixin class. Let's make the test a little more realistic, otherwise the PR is ready.

tests/test_helpers.py
iamsecretlyflash: Updated

iamsecretlyflash:
Hi @BenjaminBossan 
Added the required changes to the test

docs/source/developer_guides/lora.md
```diff
@@ -259,6 +259,58 @@ Caching can thus make inference with DoRA significantly faster but it also requi
...
```
BenjaminBossan: Just noticed that the link isn't working. When changing it to Arxiv, it links to a different paper.

BenjaminBossan (changes requested):
Ah, I was about to approve but found one more issue, please check.

docs/source/developer_guides/lora.md
iamsecretlyflash: Have fixed the link.

BenjaminBossan (approved):
Thanks for the final push, the PR is ready to be merged. Failing CI is unrelated.

iamsecretlyflash:
Wooohooo 🥳
Thanks a ton @BenjaminBossan

Issue #2948, opened 2025-12-10
Add PVeRA to PEFT


leofillioux:
### Feature request

Hey!

We propose to add PVeRA to PEFT. PVeRA [https://www.arxiv.org/abs/2512.07703](https://www.arxiv.org/abs/2512.07703) is a probabilistic variation of VeRA, which learns a distribution in the latent adaptation space. It improves the performance of the VeRA (71.4% vs 69.9% accuracy on the VTAB-1k benchmark), and allows to use the learned latent space for uncertainty quantification (e.g. Monte Carlo confidence interval estimation).

### Your contribution

The official implementation of the paper is available [https://github.com/leofillioux/pvera](https://github.com/leofillioux/pvera). I'm available to prepare a PR.

githubnemo:
Hey @leofillioux, that sounds promising! Thanks for the offer, we'd appreciate if you could submit a PR :)

Since your work is based on VeRA it might be a good start to base your implementation on the existing one, I would implement it as its own tuner instead of modifying the existing implementation, though.

Don't hesitate to submit a rough draft PR early on to get early feedback!

leofillioux:
Hey @githubnemo, I opened pull request #2952 concerning this issue

BenjaminBossan:
_not stale_

BenjaminBossan:
_not stale_

BenjaminBossan:
Completed via #2952

PR #2952, opened 2025-12-11
Integration of PVeRA

leofillioux:
This PR is a continuation of issue #2948, for which we proposed the integration of the [PVeRA adapter](https://www.arxiv.org/abs/2512.07703).

As recommended in the issue, we based our implementation on the implementation of the VeRA adapter, as both adapters are very close. Here are a list of the contributions from this PR.
- Added the PVeRA adapter folder in src/peft, and adapted the `config.py`, `layer.py`, and `model.py` files from the ones from VeRA.
- Added some documentation.
- Adapted the `tests/test_vera.py` to `tests/test_pvera.py` and made sure this ran properly.

Note: because I'm running on a Mac, I was not able to run `make test` (had an error with MPS).

@BenjaminBossan could you please give me some feedback?


leofillioux:
Update : ran `make test` on CPU and got `16496 passed, 3563 skipped, 10 xfailed, 13272 warnings` on leofillioux:main, for comparison with the huggingface:main branch I got `16476 passed, 3371 skipped, 10 xfailed, 13272 warnings`.

docs/source/package_reference/pvera.md
```diff
@@ -0,0 +1,40 @@
+<!--Copyright 2024 The HuggingFace Team. All rights reserved.
```
githubnemo: Let's use 2025

src/peft/tuners/pvera/__init__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
githubnemo:
```suggestion
# Copyright 2025-present the HuggingFace Inc. team.
```

src/peft/tuners/pvera/config.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+        # check for layers_to_transform and layers_pattern
+        if self.layers_pattern and not self.layers_to_transform:
+            raise ValueError("When `layers_pattern` is specified, `layers_to_transform` must also be specified. ")
+        if not self.save_projection:
+            warnings.warn(
+                "Specified to not save pvera_A and pvera_B within the state dictionary, instead they will be restored "
+                "using the PRNG key store in `config.projection_prng_key`. Consider setting `config.save_projection` "
+                "to `True` to guarantee restoring the checkpoint correctly on all system configurations."
+            )
```
githubnemo, line 205: Useful hint, nice.

src/peft/utils/save_and_load.py
```diff
@@ -196,6 +196,19 @@ def renamed_dora_weights(k):
                 )
             to_return["base_model.vera_A." + adapter_name] = state_dict["base_model.vera_A." + adapter_name]
             to_return["base_model.vera_B." + adapter_name] = state_dict["base_model.vera_B." + adapter_name]
+    elif config.peft_type == PeftType.PVERA:
+        vera_prefix = PEFT_TYPE_TO_PREFIX_MAPPING[config.peft_type]
+        to_return = {k: state_dict[k] for k in state_dict if vera_prefix in k}
+        if config.save_projection:
+            # TODO: adding pvera_A and pvera_B to `self.get_base_layer` would
+            # make name to match here difficult to predict.
```
githubnemo, line 205: Not sure if I understand this TODO. Can you elaborate?

src/peft/tuners/pvera/model.py
```diff
@@ -0,0 +1,297 @@
...
+        linear_out_dim, linear_in_dim = self._find_dim(config)
+
+        # use of persistent to exclude pvera_A and pvera_B from the state dict if we choose not to save them.
+        self.pvera_A = BufferDict({}, persistent=config.save_projection)
+        self.pvera_B = BufferDict({}, persistent=config.save_projection)
+
+        # deterministic init of pvera_A and pvera_B if we know the key
+        generator = torch.Generator(device="cpu").manual_seed(config.projection_prng_key)
+        pvera_A = _kaiming_init((config.r * 2, linear_in_dim), generator=generator)
+        pvera_B = _kaiming_init((linear_out_dim, config.r), generator=generator)
```
githubnemo: Can we replace these with `torch.nn.init.kaiming_init_`?

src/peft/tuners/pvera/model.py
```diff
@@ -0,0 +1,297 @@
...
+                # skip the current config
+                continue
+
+            if existing_config.projection_prng_key != config.projection_prng_key:
+                raise ValueError(
+                    f"PVeRA PRNG initialisation key must be the same for all adapters. Got {config.projection_prng_key=} but "
+                    f"previous config had {existing_config.projection_prng_key}."
+                )
+
+        save_project_unique_values = sorted({config.save_projection for config in self.peft_config.values()})
```
githubnemo: We can remove the `sorted()`?

docs/source/package_reference/pvera.md
```diff
@@ -0,0 +1,40 @@
...
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# PVeRA: Probabilistic Vector-Based Random Matrix Adaptation
+
+[PVeRA](https://www.arxiv.org/abs/2512.07703) is a parameter-efficient fine-tuning technique that is base on VeRA, in the family of the LoRA-based adapters. It keeps the very low parameter budget of VeRA, but increases the performance by learning a distribution of latent adaptations. This also enables models adapted with PVeRA to generate Monte Carlo confidence interval estimates, by sampling from the learned distribution at inference.
```
githubnemo:
```suggestion
[PVeRA](https://huggingface.co/papers/2512.07703) is a parameter-efficient fine-tuning technique that is base on VeRA, in the family of the LoRA-based adapters. It keeps the very low parameter budget of VeRA, but increases the performance by learning a distribution of latent adaptations. This also enables models adapted with PVeRA to generate Monte Carlo confidence interval estimates, by sampling from the learned distribution at inference.
```

githubnemo (commented):
Thanks for the PR, this already looks quite good!

Please check the copyright notices and make sure that it is up-to-date (it often says 2024 but it should be 2025).

I've done a quick review and left a few comments. General remarks:

- let's add PVeRA to `tests/test_custom_models.py` (adding the important configurations to `TEST_CASES`, similar to VeRA)
- if these pass we can extend the coverage by adding PVeRA to `tests/test_decoder_models.py` and `tests/test_encoder_decoder_models.py`

Once these are implemented I'll do a more thorough review. After that it'd be nice to have a runnable example and to integrate it into the method comparison to benchmark it against the other methods (and to check our expectations).

Heads up: I'll be rather off than on in the next days, so merging and review will most likely happen in the next year.

docs/source/package_reference/pvera.md
leofillioux: Good catch, I've updated this in the files I created.

src/peft/utils/save_and_load.py
leofillioux: This comes from a `TODO` from VeRA. I put this here so that when this is done for VeRA we can remember to do it for PVeRA.

src/peft/tuners/pvera/model.py
leofillioux: Good idea, I updated this. Btw this is also the case for VeRA and RandLoRA, I guess it could also changed to `torch.nn.init.kaiming_init_`.

leofillioux:
Hello @githubnemo, thank you for your review! I added two commits:

- [dd19a9a](https://github.com/huggingface/peft/pull/2952/commits/dd19a9ac78422e1fa00f505e8c267a05e4ed6a7d) which addresses the comments made above.
- [d52cc76](https://github.com/huggingface/peft/pull/2952/commits/d52cc76906c939c4f35257a27d44bd25fd388c7d) which adds PVeRA to `tests/test_custom_models.py`, `tests/test_decoder_models.py` and `tests/test_encoder_decoder_models.py`. This required a few changes to other files (e.g. adding the `pvera/bnb.py` file, fixing the merging of adapter, ...).

After all these changes, I re-ran `make test` and got `16920 passed, 3643 skipped, 10 xfailed, 13582 warnings`.
No worries, enjoy your time off!

src/peft/tuners/pvera/config.py
```diff
@@ -0,0 +1,172 @@
...
+from peft.config import PeftConfig
+from peft.utils import PeftType
+
+
+@dataclass
+class PVeRAConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`PVeRAModel`].
+
+    Paper: https://www.arxiv.org/abs/2512.07703.
```
githubnemo, line 29:
```suggestion
    Paper: https://huggingface.co/papers/2512.07703.
```

src/peft/tuners/pvera/config.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Args:
+        r (`int`, *optional*, defaults to `256`):
+            PVeRA parameter dimension ("rank"). Choose higher values than LoRA ranks here, since PVeRA uses far fewer
+            parameters than LoRA.
```
githubnemo:
```suggestion
            PVeRA parameter dimension ("rank"). Choose higher values than LoRA ranks here, since PVeRA uses parameter sharing and therefore uses far fewer
            parameters than LoRA.
```

or something like that to explain where the reduced parameter usage comes from.

src/peft/tuners/pvera/config.py
```diff
@@ -0,0 +1,172 @@
...
+        target_modules (`Union[List[str], str]`):
+            The names of the modules to apply PVeRA to. Only linear layers are supported.
```
githubnemo: This has the same feature set as the `target_modules` parameter of `LoRA`. I think it would be nice to at least hint that this supports regular expressions and keywords like `all-linear`.

src/peft/tuners/pvera/config.py
```diff
@@ -0,0 +1,172 @@
...
+            that did not include these projections. Defaults to `0`.
+        save_projection (`bool`):
+            Whether to save the pvera_A / pvera_B projections in the state dict alongside per layer lambda_b / lambda_d
+            weights. This will increase the size of the checkpoint, but guarantee that we can reload the checkpoint on
+            all system configurations. Defaults to `True`.
+        pvera_dropout (`float`):
+            The dropout probability for PVeRA layers.
+        d_initial (`float`, *optional*, defaults to `0.1`):
+            Initial init value for `pvera_lambda_d` vector used when initializing the PVeRA parameters. Small values
+            (<=0.1) are recommended (see Table 6c in the paper).
```
githubnemo:
```suggestion
            Initial value for `pvera_lambda_d` vector used when initializing the PVeRA parameters. Small values
            (<=0.1) are recommended (see Table 6c in the paper).
```

src/peft/tuners/pvera/layer.py
```diff
@@ -0,0 +1,310 @@
...
+        init_weights: bool = True,
+        d_initial: float = 0.1,
+        sample_at_inference: bool = False,
+        **kwargs,
+    ) -> None:
+        # this gets the init from nn.Linear's super perspective, i.e. nn.Module.__init__, which should always be called
+        super(nn.Linear, self).__init__()
+        PVeRALayer.__init__(self, base_layer, **kwargs)
+        self.fan_in_fan_out = fan_in_fan_out
+        self.sample_at_inference = sample_at_inference
```
githubnemo, line 164: For consistency I think this should be a dict so that I can toggle this behavior for each adapter.

docs/source/package_reference/pvera.md
```diff
@@ -0,0 +1,40 @@
...
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# PVeRA: Probabilistic Vector-Based Random Matrix Adaptation
+
+[PVeRA](https://huggingface.co/papers/2512.07703) is a parameter-efficient fine-tuning technique that is base on VeRA, in the family of the LoRA-based adapters. It keeps the very low parameter budget of VeRA, but increases the performance by learning a distribution of latent adaptations. This also enables models adapted with PVeRA to generate Monte Carlo confidence interval estimates, by sampling from the learned distribution at inference.
```
githubnemo, line 19: Maybe also mention (not necessarily here) how we can set the required `sample_at_infernece` config value when loading a pre-trained checkpoint:

```python
# Setting sample_at_inference=True for PVeRA checkpoints during load
from peft import PeftModel, PeftConfig

base_model = AutoModelForCausalLM.from_pretrained("gpt2")

peft_config = PeftConfig.from_pretrained(peft_model_id)
peft_config.sample_at_inference = True

peft_model = PeftModel.from_pretrained(base_model, peft_model_id, config=peft_config)
```

tests/test_custom_models.py
```diff
@@ -1081,6 +1101,14 @@
         {"target_modules": ["lin0"], "init_weights": False},
         {"target_modules": ["lin0"], "init_weights": False},
     ),
+    # Note: PVeRA may present the same problem mentioned above for VeRA.
```
githubnemo:
```suggestion
    # Note: PVeRA presents the same problem mentioned above for VeRA.
```

Since the process is basically equal we can drop the hypothetical :)

tests/test_custom_models.py
```diff
@@ -2113,6 +2141,8 @@ def test_parameters_after_loading_model(self, test_name, model_id, config_cls, c
             lr = 1e-3  # we get exploding gradients with MHA when learning rate is too high
         elif issubclass(config_cls, (VBLoRAConfig, RandLoraConfig, OSFConfig)):
             lr = 0.01  # otherwise we get nan
+        elif issubclass(config_cls, PVeRAConfig):  # needs very small lr to not get nan
+            lr = 1e-6
```
githubnemo, line 2195: This is indeed very small - do you have an idea which gradients are exploding and if this is a problem in practice?

tests/test_custom_models.py
```diff
@@ -5162,6 +5194,148 @@ def forward(self, X):
...
+                X = X.float()
+                X = self.lin0(X)
+                X = self.relu(X)
+                X = self.lin1(X)
+                X = self.relu(X)
+                X = self.lin2(X)
+                X = self.relu(X)
+                X = self.lin3(X)
+                X = self.sm(X)
+                return X
```
githubnemo: This model definition now occurs 4x if I'm not mistaken. Let's refactor this by moving it outside with a more descriptive name (e.g., `MLP_2x_same_shape`, ...).

tests/test_encoder_decoder_models.py
```diff
@@ -216,6 +217,19 @@
             "task_type": "SEQ_2_SEQ_LM",
         },
     ),
+    (
+        PVeRAConfig,
+        {
+            "r": 8,
+            "target_modules": None,
+            "pvera_dropout": 0.05,
+            "projection_prng_key": 0xFF,
```
githubnemo: why deviate from the default of 0 for `projection_prng_key`?

tests/testing_common.py
```diff
@@ -1114,7 +1117,7 @@ def _test_training_layer_indexing(self, model_id, config_cls, config_kwargs):
         with hub_online_once(model_id):
             model = self.transformers_class.from_pretrained(model_id)
             model = get_peft_model(model, config)
-            model = model.to(self.torch_device)
+            model = model.to(self.torch_device).eval()
```
githubnemo: This seems wrong since this is a training test - why put the model in eval mode?

tests/testing_common.py
```diff
@@ -1139,8 +1142,8 @@ def _test_training_layer_indexing(self, model_id, config_cls, config_kwargs):
                 model.save_pretrained(tmp_dirname)
 
                 model_from_pretrained = self.transformers_class.from_pretrained(model_id)
-                model_from_pretrained = PeftModel.from_pretrained(model_from_pretrained, tmp_dirname).to(
-                    self.torch_device
+                model_from_pretrained = (
+                    PeftModel.from_pretrained(model_from_pretrained, tmp_dirname).to(self.torch_device).eval()
```
githubnemo: same as above

githubnemo (commented):
Hey, thanks for the update!

Implementation and tests look quite mature. I think that if we provide bitsandbytes support we also need at least one test in `tests/test_gpu_examples.py` for quality control. I think `test_causal_lm_training_4bit_vera` could be used as a base.

In general I think we should rename `PVeRA*` to `Pvera*` (e.g., `PVeRAModel` -> `PveraModel`) to be consistent with `VeraModel` and friends. It is quite hard to remember the spelling of the various abbreviations already :)

Rest of the review is in the comments.

After the comments are resolved and the CI is green I think it would be nice to integrate `PVeRA` into the MetaMathQA benchmark by adding an experiment file based on the [VeRA experiment](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA/experiments/vera/llama-3.2-3B-default).

Edit: To fix the docs build, add the PVeRA entry to `docs/source/_toctree.yml` similar to the other entries.

src/peft/tuners/pvera/layer.py
leofillioux: Here, are you talking about `sample_at_inference`?

tests/test_custom_models.py
leofillioux: I checked, the problems come from the `pvera_lambda_b` which has its parameters at `nan` after the third epoch. This seems to be caused by the input which has values up to 90, which in practice shouldn't occur since the inputs are usually normalized.

tests/testing_common.py
leofillioux: Yes you're right, it doesn't make sense to put `.eval()`. However there is a problem in that `model` is in training mode here, whereas `model_from_pretrained` is is eval mode. This means that for PVeRA one model will sample from the learned distribution and the other will not, therefore giving different results, which is why I had put them both in eval mode. Do you think we should maybe just remove this test for PVeRA?

leofillioux:
Hello @githubnemo, thanks again for your review. I have updated the code with most of your comments, but the following issues still remain:
- A question regarding what should be converted to a dict in your comment [here](https://github.com/huggingface/peft/pull/2952#discussion_r2687198290).
- Regarding the small learning rate for PVeRA in a test, it seems to be due to a non-normalised input, which seems to cause instability in the sampling, I added a bit more info [here](https://github.com/huggingface/peft/pull/2952#discussion_r2689656230).
- Regarding adding `.eval()` to some tests during training. I see why this is problematic, so I have removed them, but as explained [here](https://github.com/huggingface/peft/pull/2952#discussion_r2690844351), this is a problem for PVeRA as this test compared the outputs of a model in train with a model in eval model. What do you think of removing this test for PVeRA?

src/peft/tuners/pvera/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+from peft.utils import register_peft_method
+
+from .config import PveraConfig
+from .layer import Linear, PVeRALayer
+from .model import PveraModel
+
+
+__all__ = ["Linear", "PVeRALayer", "PveraConfig", "PveraModel"]
```
githubnemo:
```suggestion
__all__ = ["Linear", "PveraLayer", "PveraConfig", "PveraModel"]
```

+for the class itself as well

src/peft/tuners/pvera/layer.py
githubnemo: Ah I always forget that github doesn't highlight the commented lines. Yes, I was talking about `sample_at_inference`.

tests/test_custom_models.py
githubnemo: OK. Let's extend the comment to be more precise:

```
# needs a very small lr to not get nan in pvera_lambda_b due to high input values in this test (up to 90)
```

tests/testing_common.py
githubnemo: I see. Let's check if we can make this consistent by setting the model to eval mode after getting the loss (i.e. in the context block where the comparison happens) and moving the `logits = ...` retrieval down there.

For example:
```python
                model.eval()
                logits = model(**inputs)[0][0]

                model.save_pretrained(tmp_dirname)

                model_from_pretrained = self.transformers_class.from_pretrained(model_id)
                model_from_pretrained = PeftModel.from_pretrained(model_from_pretrained, tmp_dirname).to(
                    self.torch_device
                )

                logits_from_pretrained = model_from_pretrained(**inputs)[0][0]
                atol, rtol = 1e-4, 1e-4
                assert torch.allclose(logits, logits_from_pretrained, atol=atol, rtol=rtol)
```

IMO this should not change the test for the worse and should not break anything but we'll see :)

tests/test_pvera.py
```diff
@@ -0,0 +1,302 @@
...
+        peft_model.load_adapter(save_path / "other", "other")
+        peft_model.eval()
+
+        peft_model.set_adapter("default")
+        output_default_loaded = peft_model(input)
+        peft_model.set_adapter("other")
+        output_other_loaded = peft_model(input)
+
+        assert torch.allclose(output_default, output_default_loaded, atol=1e-3, rtol=1e-3)
+        assert torch.allclose(output_other, output_other_loaded, atol=1e-3, rtol=1e-3)
```
githubnemo: `test_multiple_adapters_save_load_save_projection_true` probably already covered by the common tests or am I missing something special?

tests/test_pvera.py
```diff
@@ -0,0 +1,302 @@
...
+
+        sd_other = {}
+        with safe_open(save_path / "other" / "adapter_model.safetensors", framework="pt", device="cpu") as f:
+            for key in f.keys():
+                sd_other[key] = f.get_tensor(key)
+
+        assert any("pvera_A" in key for key in sd_other)
+        assert any("pvera_B" in key for key in sd_other)
+        assert sd_other["base_model.pvera_A"].shape == (512, 20)
+        assert sd_other["base_model.pvera_B"].shape == (20, 256)
```
githubnemo: Isn't `test_multiple_adapters_save_projection_true_contains_pvera_A_pvera_B` already covered if we test that load/restore works and is consistent?

tests/test_encoder_decoder_models.py
```diff
@@ -216,6 +217,18 @@
...
+    (
+        PveraConfig,
+        {
+            "r": 8,
+            "target_modules": None,
+            "pvera_dropout": 0.05,
+            "d_initial": 0.1,
+            "save_projection": True,
+            "bias": "none",
+            "task_type": "SEQ_2_SEQ_LM",
```
githubnemo, line 225: Let's just mention the non-default values here.

tests/test_decoder_models.py
```diff
@@ -298,6 +299,18 @@
...
+        PveraConfig,
+        {
+            "task_type": "CAUSAL_LM",
+            "r": 8,
+            "target_modules": None,
+            "pvera_dropout": 0.05,
+            "d_initial": 0.1,
+            "save_projection": True,
+            "bias": "none",
+        },
```
githubnemo, line 310: Same as below, just mention the non-default values.

src/peft/tuners/pvera/config.py
```diff
@@ -0,0 +1,175 @@
...
+        metadata={
+            "help": (
+                "Whether to save the pvera_A / pvera_B projections in the state dict alongside per layer lambda_b / "
+                "lambda_d weights. This will increase the size of the checkpoint, but guarantee that we can reload "
+                "the checkpoint on all system configurations."
+            )
+        },
+    )
+    pvera_dropout: float = field(default=0.0, metadata={"help": "PVeRA dropout"})
+    d_initial: float = field(default=0.1, metadata={"help": "Initial init value for d vector."})
```
githubnemo:
```suggestion
    d_initial: float = field(default=0.1, metadata={"help": "Initial value for d vector."})
```

But we can just use the docstring for this parameter from above verbatim

src/peft/tuners/pvera/config.py
```diff
@@ -0,0 +1,175 @@
...
+
+    r: int = field(default=256, metadata={"help": "PVeRA attention dimension"})
+
+    target_modules: Optional[Union[list[str], str]] = field(
+        default=None,
+        metadata={
+            "help": (
+                "List of module names or regex expression of the module names to replace with PVeRA."
+                "For example, ['q', 'v'] or '.*decoder.*(SelfAttention|EncDecAttention).*(q|v)$'. "
+                "Only linear layers are supported."
```
githubnemo: Let's use the docstring values from above for all help values. In the end this makes it more maintainable and we're losing not much.

docs/source/package_reference/pvera.md
githubnemo, line 1: Also make sure to add this file to the `docs/source/_toctree.yml` possibly next to VeRA.

githubnemo (commented):
Thanks for the changes!

I've commented on hopefully all the outstanding issues and added a few nits.

CI seems to be passing except for the `.eval` issue which is hopefully resolved with the proposed fix.

tests/test_pvera.py
leofillioux: To be honest, I integrated these (`test_multiple_adapters_save_load_save_projection_true` and `test_multiple_adapters_save_projection_true_contains_pvera_A_pvera_B`) because they were integrated for VeRA, but I agree that they do seem a bit redundant. I'm ok with removing them if you think they aren't useful.

leofillioux:
Hello @githubnemo,
Thanks again for your review! I think I have addressed the issues with my newest commits. The only thing I am unsure about is regarding the redundant tests, for which I have left a comment [here](https://github.com/huggingface/peft/pull/2952#discussion_r2715743766).

tests/test_pvera.py
githubnemo: Yeah, let's remove them.

githubnemo (commented):
Thanks for the updates!

Correct me if I'm wrong but I think there's currently no test that checks whether the output is non-deterministic during training and during inference when requested. Let's make sure that this code path is tested, at least roughly.

I checked the old comments and resolved those that are done. 
There are still some open comments:
- https://github.com/huggingface/peft/pull/2952#discussion_r2712674190
- https://github.com/huggingface/peft/pull/2952/files#r2686810130.

Once that is done, let's add an experiment for the MetaMathQA method comparison. You can look at `method_comparison/MetaMathQA/experiments/vera/llama-3.2-3B-default/` for inspiration. You  may need to tune some hyper-parameters since they are probably non-optimal. Note that we try to keep the number of parameters roughly equal between methods. There's also documentation there how to run these experiments yourself for quick iteration.

It'd also good to have an example, maybe showcasing how to leverage sampling for getting confidence intervals? I let you be the judge of what is most effective in show-casing PVeRA.

docs/source/package_reference/pvera.md
leofillioux: This is done in the docstring from PveraModel, so it should appear in the documentation.

leofillioux:
Hello @githubnemo,
Thanks again for your comments. I've removed the redundant tests ([d5d621e](https://github.com/huggingface/peft/pull/2952/commits/d5d621e5be80d1564799f4ec912e07b805cb5764)), used the docstring values for the help ([c8f8e64](https://github.com/huggingface/peft/pull/2952/commits/c8f8e64fb789d7d3aa0c6a33de3d7a153eab0c23)), added the MetaMathQA experiment ([000f772](https://github.com/huggingface/peft/pull/2952/commits/000f772b6ac7ad8f7b1dc03112b9bb5bfb8aa575) and [ea40bbb](https://github.com/huggingface/peft/pull/2952/commits/ea40bbbcc884e6f41602fd378f4b92b0e303cd48)). It is indeed a nice idea to showcase the confidence interval generation, I added this to the documentation ([c5d9d01](https://github.com/huggingface/peft/pull/2952/commits/c5d9d011152e37f46a32c0e94890b9d35c83da92)).

docs/source/package_reference/pvera.md
githubnemo: Ah, yes, it is documented how to configure a newly instantiated adapter. But I was more interested in how to load an already trained adapter and activate the sampling (which is what is done in the example I posted).

method_comparison/MetaMathQA/results/pvera--llama-3.2-3B-default.json
githubnemo, line 1: This was very helpful to compare the results with VeRA!

Let's remove the results from the PR now - we'll use a dedicated machine to run the experiments on for better comparability.

docs/source/package_reference/pvera.md
```diff
@@ -38,3 +38,65 @@ The abstract from the paper is:
 ## PveraModel
 
 [[autodoc]] tuners.pvera.model.PveraModel
+
+## Confidence interval generation
```
githubnemo: Since this example is mostly code, let's move it to `examples/pvera/confidence_interval_generation.py` alongside a short README.md how to run the example.

docs/source/package_reference/pvera.md
```diff
@@ -38,3 +38,65 @@ The abstract from the paper is:
...
+optimizer = torch.optim.Adam(list(model.parameters()) + list(linear_probe.parameters()), lr=1e-4)
+dataloader = torch.utils.data.DataLoader(dataset, batch_size=32, shuffle=True)
+for batch in tqdm(dataloader):
+    imgs, lbls = transform(batch['image'].float()), batch['labels']
+    pred = linear_probe(model(imgs.to(device)).pooler_output)
+    loss = criterion(pred, lbls.to(device))
+    loss.backward()
+    optimizer.step()
+
+# switch to sample_at_inference=True
```
githubnemo:
```suggestion
# switch sample_at_inference=True on PVeRA targeted modules
# (query and value in this case)
```

but maybe this is also a good place to save the adapter and load it back from disk with `sample_at_inference=True` since that would be the most common case? (as discussed [here](https://github.com/huggingface/peft/pull/2952/changes/aaa267f9f0d2a0ebb0738215660030539e4d52a4..c5d9d011152e37f46a32c0e94890b9d35c83da92#r2686810130))

docs/source/package_reference/pvera.md
```diff
@@ -38,3 +38,65 @@ The abstract from the paper is:
...
+softmax_top_pred = all_preds[:, top_pred]
+
+def mean_confidence_interval(data, confidence=0.95):
+    a = 1.0 * np.array(data)
+    n = len(a)
+    m, se = np.mean(a), scipy.stats.sem(a)
+    h = se * scipy.stats.t.ppf((1 + confidence) / 2., n-1)
+    return max(0, m-h), min(1, m+h)
+
+print(mean_confidence_interval(softmax_top_pred.cpu()))
```
githubnemo: Maybe something like

```python
print(f"top-1 class: {top_pred}")
print(f"top-1 probability: {softmax_top_pred.mean().item()}")
print(f"top-1 confidence interval: {mean_confidence_interval(softmax_top_pred.cpu())}")
```

githubnemo (commented):
Hey @leofillioux!

I think this is almost good to go.

Thanks for providing the benchmark experiment + results.
The results are in line with the VeRA (~36% accuracy). If you already have a bit of experience in hyper-parameter tuning VeRA/PVeRA (i.e. maybe you know that the learning rate usually needs to be a bit higher) you're welcome to integrate that knowledge into the benchmark but this doesn't need to happen in this PR - it's just a hint that the VeRA configuration might not be optimal.

There's still the open issue of per-adapter `sample_at_inference` setting (via https://github.com/huggingface/peft/pull/2952#discussion_r2686785742). I think it is not that big of a deal to not have adapter-wise sampling, we can leave it out.

The failing CI is unrelated but make sure to merge with main so that we can give the CI a final run after you've addressed the few remaining comments.

leofillioux:
Hello @githubnemo,

Thanks again for the review! I've addressed the remaining points. Regarding the `sample_at_inference` as a dictionary, this is normally handled in [this commit](https://github.com/huggingface/peft/pull/2952/changes/aaa267f9f0d2a0ebb0738215660030539e4d52a4#diff-426fd3851b6a426c7e10876755f446f883852624f04dc409206b32aee298b3ee). I've also merged with main.

githubnemo (approved):
> Hello @githubnemo,
> 
> Thanks again for the review! I've addressed the remaining points. Regarding the `sample_at_inference` as a dictionary, this is normally handled in [this commit](https://github.com/huggingface/peft/pull/2952/changes/aaa267f9f0d2a0ebb0738215660030539e4d52a4#diff-426fd3851b6a426c7e10876755f446f883852624f04dc409206b32aee298b3ee). I've also merged with main.

Thanks for the quick response. I think I was a bit vague with what I meant regarding `sample_at_inference` - possibly because we have two definitions of adapter. I meant 'adapter' in the sense of PEFT's `.add_adapter()`, i.e. a second `PVeraConfig` on the same model. In that case `sample_at_inference` would be a global setting and I couldn't disable it for adapter 1 without disabling it for adapter 2 as well. I was not talking about individual adapted layers.

But I think this is fine as is! I'll let the CI run, if it passes I'll merge this if you don't have any objections :)

BenjaminBossan:
@leofillioux Some of the PVeRA GPU tests are failing on CI. This can be tested like so:

```sh
# single gpu
pytest -m single_gpu_tests tests/test_gpu_examples.py -k pvera -v
# multi gpu
pytest -m multi_gpu_tests tests/test_gpu_examples.py -k pvera -v
```

One error is trivial (using `vera_dropout` instead of `pvera_dropout`) but for the others, I'm not so sure, as we get this error in the forward pass:

`RuntimeError: mat1 and mat2 shapes cannot be multiplied (128x32 and 16x4096`

Do you have time to investigate this?

leofillioux:
Hello @BenjaminBossan, I tried to reproduce this error but I only get this error

```
ValueError: Asking to pad but the tokenizer does not have a padding token. Please select a token to use as `pad_token` `(tokenizer.pad_token = tokenizer.eos_token e.g.)` or add a new pad token via `tokenizer.add_special_tokens({'pad_token': '[PAD]'})`.
```

I also get the same error when running `-k vera`. I'm guessing this comes from my environment, but it fails even after reinstalling from scratch. Any idea how to fix this?

BenjaminBossan:
Thanks for checking @leofillioux. What package versions are you using: I'm using

- PEFT main
- transformers main / v5.5.4
- bitsandbytes 0.49.2
- torch 2.10.0

BenjaminBossan:
@leofillioux Did you have time to check yet?

leofillioux:
@BenjaminBossan I wasn't able to reproduce the error. I haven't had much time in the last few days, but I'll have more time in the end of the week!

BenjaminBossan:
Thanks. In the meantime, if you could share the package versions of your env, I could also do more testing.

leofillioux:
Yes sure, here is the full output of `pip freeze`.
```
accelerate==1.13.0
aiohappyeyeballs==2.6.1
aiohttp==3.13.5
aiosignal==1.4.0
annotated-doc==0.0.4
anyio==4.13.0
async-timeout==5.0.1
attrs==26.1.0
bitsandbytes==0.49.2
black==26.3.1
certifi==2026.2.25
charset-normalizer==3.4.7
click==8.3.2
coverage==7.13.5
cuda-bindings==12.9.4
cuda-pathfinder==1.5.2
cuda-toolkit==13.0.2
datasets==4.8.4
diffusers==0.37.1
dill==0.4.1
exceptiongroup==1.3.1
execnet==2.1.2
fastjsonschema==2.21.2
filelock==3.25.2
frozenlist==1.8.0
fsspec==2026.2.0
gitdb==4.0.12
GitPython==3.1.46
h11==0.16.0
hf-doc-builder==0.5.0
hf-xet==1.4.3
httpcore==1.0.9
httpx==0.28.1
huggingface_hub==1.10.1
idna==3.11
importlib_metadata==9.0.0
iniconfig==2.3.0
Jinja2==3.1.6
joblib==1.5.3
jsonschema==4.26.0
jsonschema-specifications==2025.9.1
jupyter_core==5.9.1
markdown-it-py==4.0.0
MarkupSafe==3.0.3
mdurl==0.1.2
mpmath==1.3.0
multidict==6.7.1
multiprocess==0.70.19
mypy_extensions==1.1.0
nbformat==5.10.4
networkx==3.4.2
numpy==2.2.6
nvidia-cublas==13.1.0.3
nvidia-cublas-cu12==12.8.4.1
nvidia-cuda-cupti==13.0.85
nvidia-cuda-cupti-cu12==12.8.90
nvidia-cuda-nvrtc==13.0.88
nvidia-cuda-nvrtc-cu12==12.8.93
nvidia-cuda-runtime==13.0.96
nvidia-cuda-runtime-cu12==12.8.90
nvidia-cudnn-cu12==9.10.2.21
nvidia-cudnn-cu13==9.19.0.56
nvidia-cufft==12.0.0.61
nvidia-cufft-cu12==11.3.3.83
nvidia-cufile==1.15.1.6
nvidia-cufile-cu12==1.13.1.3
nvidia-curand==10.4.0.35
nvidia-curand-cu12==10.3.9.90
nvidia-cusolver==12.0.4.66
nvidia-cusolver-cu12==11.7.3.90
nvidia-cusparse==12.6.3.3
nvidia-cusparse-cu12==12.5.8.93
nvidia-cusparselt-cu12==0.7.1
nvidia-cusparselt-cu13==0.8.0
nvidia-nccl-cu12==2.27.5
nvidia-nccl-cu13==2.28.9
nvidia-nvjitlink==13.0.88
nvidia-nvjitlink-cu12==12.8.93
nvidia-nvshmem-cu12==3.4.5
nvidia-nvshmem-cu13==3.4.5
nvidia-nvtx==13.0.85
nvidia-nvtx-cu12==12.8.90
packaging @ file:///home/conda/feedstock_root/build_artifacts/bld/rattler-build_packaging_1769093650/work
pandas==2.3.3
parameterized==0.9.0
pathspec==1.0.4
-e git+https://github.com/huggingface/peft.git@efe0fe6acd72cb3bf1ebfc807c159bf0b9481f5e#egg=peft
pillow==12.2.0
platformdirs==4.9.6
pluggy==1.6.0
propcache==0.4.1
protobuf==7.34.1
psutil==7.2.2
pyarrow==23.0.1
Pygments==2.20.0
pytest==9.0.3
pytest-cov==7.1.0
pytest-xdist==3.8.0
python-dateutil==2.9.0.post0
pytokens==0.4.1
pytz==2026.1.post1
PyYAML==6.0.3
referencing==0.37.0
regex==2026.4.4
requests==2.33.1
rich==15.0.0
rpds-py==0.30.0
ruff==0.12.12
safetensors==0.7.0
scikit-learn==1.7.2
scipy==1.15.3
sentencepiece==0.2.1
shellingham==1.5.4
six==1.17.0
smmap==5.0.3
sympy==1.14.0
threadpoolctl==3.6.0
tokenizers==0.22.2
tomli==2.4.1
torch==2.10.0
torchvision==0.26.0
tqdm==4.67.3
traitlets==5.14.3
transformers==5.5.4
triton==3.6.0
typer==0.24.1
typing_extensions==4.15.0
tzdata==2026.1
urllib3==2.6.3
xxhash==3.6.0
yarl==1.23.0
zipp==3.23.0
```

BenjaminBossan:
Thanks for sharing. This is strange, as all the important library versions match with what I have in my env (I also tried this exact PEFT commit that you use). Could you please try the following:

1. Add a breakpoint [here](https://github.com/huggingface/peft/blob/76c37d4686cf7245a9a6998fd1e2619d62c9322d/src/peft/tuners/pvera/bnb.py#L396)
2. Run `pytest -m single_gpu_tests tests/test_gpu_examples.py -k pvera`
3. Check the shapes of the relevant tensors

I get:

```python
(Pdb) pvera_A.shape
torch.Size([32, 4096])
(Pdb) pvera_B.shape
torch.Size([4096, 16])
(Pdb) x_temp.shape
torch.Size([4, 32, 4096])
(Pdb) sliced_A.shape
torch.Size([32, 4096])
(Pdb) torch.nn.functional.linear(x_temp, sliced_A).shape
torch.Size([4, 32, 32])
(Pdb) sliced_B.shape
torch.Size([4096, 16])
```

which is clearly incompatible. What do you see?

leofillioux:
I'm not able to even get to the breakpoint, I get this error.

```
ModuleNotFoundError: Could not import module 'Seq2SeqTrainer'. Are this object's requirements defined correctly?
```

However, given the shapes, I think I see what is going on. Basically for PVeRA, we have $$\mathbf A \in \mathbb R^{2r, d}$$ (compared to $$\mathbf A \in \mathbb R^{r, d}$$ for VeRA). During training, this is used to generate $$\boldsymbol \mu, \boldsymbol \sigma \in \mathbb R^{r, d}$$, which define the learned distribution parameters, we then sample $$\mathbf z \sim \mathcal N(\boldsymbol \mu, \boldsymbol \sigma^2) \in \mathbb R^{r,d}$$. During inference, we don't sample, and we take $$\mathbf z = \boldsymbol \mu$$. When we want to merge the weights, we only take the first part of $$\mathbf A$$, which is responsible for generating $$\boldsymbol \mu$$. This is actually done [here](https://github.com/huggingface/peft/blob/76c37d4686cf7245a9a6998fd1e2619d62c9322d/src/peft/tuners/pvera/layer.py#L251). I think that just this part is missing for this file.

I hope that my explanation is clear. I'll try to fix my environment, but in the meantime, I think this should fix this issue. Were there any others (other than the wrong name for `pvera_dropout`)?

BenjaminBossan:
That could be it. The code you linked is only used during the merging/unmerging path but not the forward step. Would the correct change then be something like:

```python
sliced_A = pvera_A[: self.r[active_adapter], : self.in_features].to(x.device)
sliced_lambda_d = lambda_d[:self.r[active_adapter]]
x_temp = dropout(x.to(lambda_d.dtype))
adapter_output = lambda_b * torch.nn.functional.linear(
    sliced_lambda_d * torch.nn.functional.linear(x_temp, sliced_A), sliced_B
)
```

But that would lead to different results, would it not?

leofillioux:
Ah yes, sorry! Then the problem is that I try to apply the adapter directly like VeRA, and don't perform the sampling, which is why there is a mismatch in the dimensions. I think the correct change would be something like this.

```python
x_temp = dropout(x.to(lambda_d.dtype))

mu, logvar = (lambda_d * F.linear(dropout(x_temp), sliced_A)).chunk(2, dim=-1)
adapter_output = lambda_b * F.linear(
   self._reparametrize(mu, logvar, self.sample_at_inference), sliced_B
)
```

PR #2953, opened 2025-12-14
Add cartridges to PEFT

kashif:
https://hazyresearch.stanford.edu/blog/2025-06-08-cartridges


src/peft/tuners/cartridge/model.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+
+    def __init__(self, config):
+        super().__init__()
+        self.config = config
+
+        num_virtual_tokens = config.num_virtual_tokens
+        hidden = config.num_layers * 2 * config.token_dim
+        num_frozen_tokens = int(getattr(config, "num_frozen_tokens", 0) or 0)
+        if num_frozen_tokens < 0 or num_frozen_tokens > num_virtual_tokens:
```
BenjaminBossan, line 39: This could already be checked in the `__post_init__` of the config, right? The earlier we error, the better.

src/peft/tuners/cartridge/model.py
```diff
@@ -0,0 +1,116 @@
...
+    and the remaining tokens are trainable.
+    """
+
+    def __init__(self, config):
+        super().__init__()
+        self.config = config
+
+        num_virtual_tokens = config.num_virtual_tokens
+        hidden = config.num_layers * 2 * config.token_dim
+        num_frozen_tokens = int(getattr(config, "num_frozen_tokens", 0) or 0)
```
BenjaminBossan: Why do we need `getattr` here? I think it's not needed.

src/peft/tuners/cartridge/model.py
```diff
@@ -0,0 +1,116 @@
...
+        return torch.cat([self.frozen_embedding, self.trainable_embedding], dim=0)
+
+    def reset_parameters(self):
+        # Match `torch.nn.Embedding` initialization (normal with std=1).
+        with torch.no_grad():
+            if self.frozen_embedding is not None:
+                torch.nn.init.normal_(self.frozen_embedding)
+            torch.nn.init.normal_(self.trainable_embedding)
+
+    def load_prompt_embeddings(self, prompt_embeddings: torch.Tensor) -> None:
```
BenjaminBossan, line 89: Very nice. I wonder: Could we not implement the same method for prefix tuning and then it would also benefit from these improved init methods? I think it should work, unless prefix tuning is initialized with `prefix_projection=True` but that's not the default.

src/peft/tuners/cartridge/config.py
```diff
@@ -0,0 +1,41 @@
...
+from peft.utils import PeftType
+
+
+@dataclass
+class CartridgeConfig(PromptLearningConfig):
+    """
+    Configuration for CARTRIDGE, a KV-cache-parameterized prefix adapter.
+
+    This is similar to prefix-tuning in how it is served (as `past_key_values`), but it stores the KV cache directly as
+    trainable parameters instead of learning it via an MLP projection.
```
BenjaminBossan, line 27: Let's document what the user needs to do if they want to initialize the Cartridge model with options 2 and 3 from the paper (random kv, kv derived from first tokens of prompt).

src/peft/tuners/cartridge/utils.py
BenjaminBossan, line 1: Let's move this file to `tuners/cartridge/utils.py`.

tests/test_decoder_models.py
BenjaminBossan, line 1: Thanks for adding these tests, let's also add Cartridge to the test matrices of `test_encoder_decoder_models.py`, `test_feature_extraction.py`, `test_seq_classifier.py`, and `test_config.py`.

tests/test_cartridge.py
```diff
@@ -0,0 +1,158 @@
...
+    PeftModel,
+    compose_cartridge_adapters,
+    get_peft_model,
+    initialize_cartridge_from_past_key_values,
+    load_peft_weights,
+    prompt_embeddings_from_past_key_values,
+)
+
+
+def _make_tiny_gpt2():
```
BenjaminBossan: Any specific reason why we want a tiny GPT here? If not, I would suggest to use one of the tiny models we use elsewhere, e.g. `"peft-internal-testing/tiny-random-OPTForCausalLM"` or `"trl-internal-testing/tiny-random-LlamaForCausalLM"`. If we use these models, we always load them in a caching context to reduce hub calls:

```python
from .testing_utils import get_state_dict, hub_online_once

...

    model = AutoModel.from_pretrained(...)
```

tests/test_cartridge.py
```diff
@@ -0,0 +1,158 @@
...
+
+    assert model.active_peft_config.peft_type.value == "CARTRIDGE"
+    assert model.prompt_encoder[model.active_adapter].frozen_embedding.requires_grad is False
+    assert model.prompt_encoder[model.active_adapter].trainable_embedding.requires_grad is True
+
+    input_ids = torch.randint(0, base.config.vocab_size, (1, 8))
+    out = model(input_ids=input_ids)
+    assert out.logits.shape[:2] == (1, 8)
+
+    with tempfile.TemporaryDirectory() as tmp:
```
BenjaminBossan: Let's use the `tmp_path` fixture from pytest.

tests/test_cartridge.py
```diff
@@ -0,0 +1,158 @@
...
+        model.prompt_encoder[model.active_adapter].trainable_embedding.data.fill_(3.0)
+        if model.prompt_encoder[model.active_adapter].frozen_embedding is not None:
```
BenjaminBossan: Since we set `num_frozen_tokens=2` above, this check is not needed right? We could, however, add a `@pytest.mark.parametrize("num_frozen_tokens", [0, 2])` to make the test cover both cases.

tests/test_cartridge.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # Prefill on the *base* model and use the cache prefix as initialization.
+    input_ids = torch.randint(0, base.config.vocab_size, (1, 12))
+    with model.disable_adapter():
+        outputs = model(input_ids=input_ids, use_cache=True)
+    prompt_embeddings = initialize_cartridge_from_past_key_values(
+        model, past_key_values=outputs.past_key_values, num_virtual_tokens=4
+    )
+    assert prompt_embeddings.shape[0] == 4
+    assert torch.allclose(model.prompt_encoder[model.active_adapter].weight.cpu(), prompt_embeddings.cpu())
```
BenjaminBossan: `.cpu()` not needed?

tests/test_cartridge.py
```diff
@@ -0,0 +1,158 @@
...
+
+    with tempfile.TemporaryDirectory() as tmp:
```
BenjaminBossan: Let's use the tmp_path fixture from pytest.

tests/test_cartridge.py
```diff
@@ -0,0 +1,158 @@
...
+    with model.disable_adapter():
+        outputs = model(input_ids=input_ids, use_cache=True)
+
+    pe = prompt_embeddings_from_past_key_values(outputs.past_key_values, num_virtual_tokens=4)
+    assert pe.shape[0] == 4
+
+    pe2 = initialize_cartridge_from_past_key_values(
+        model, past_key_values=outputs.past_key_values, num_virtual_tokens=4
+    )
+    assert torch.allclose(pe.cpu(), pe2.cpu())
```
BenjaminBossan: `.cpu()` not needed?

tests/test_cartridge.py
```diff
@@ -0,0 +1,158 @@
...
+def test_cartridge_inference_mode_disables_grads_and_forward_works():
+    base = _make_tiny_gpt2()
+    peft_config = CartridgeConfig(
+        num_virtual_tokens=4, num_frozen_tokens=2, task_type="CAUSAL_LM", inference_mode=True
+    )
+    model = get_peft_model(base, peft_config)
+
+    enc = model.prompt_encoder[model.active_adapter]
+    assert enc.trainable_embedding.requires_grad is False
+    if enc.frozen_embedding is not None:
```
BenjaminBossan: Again, let's paramatrize over `num_frozen_tokens`.

tests/test_cartridge.py
```diff
@@ -0,0 +1,158 @@
...
+    pe = prompt_embeddings_from_past_key_values(outputs.past_key_values, num_virtual_tokens=4)
+    assert pe.shape[0] == 4
+
+    pe2 = initialize_cartridge_from_past_key_values(
+        model, past_key_values=outputs.past_key_values, num_virtual_tokens=4
+    )
+    assert torch.allclose(pe.cpu(), pe2.cpu())
+
+
+def test_cartridge_inference_mode_disables_grads_and_forward_works():
```
BenjaminBossan: It's not clear to me why `requires_grad` should be `False`, could you please add a comment?

tests/test_cartridge.py
```diff
@@ -0,0 +1,158 @@
...
+        num_virtual_tokens=4, num_frozen_tokens=2, task_type="CAUSAL_LM", inference_mode=True
+    )
+    model = get_peft_model(base, peft_config)
+
+    enc = model.prompt_encoder[model.active_adapter]
+    assert enc.trainable_embedding.requires_grad is False
+    if enc.frozen_embedding is not None:
+        assert enc.frozen_embedding.requires_grad is False
+
+    input_ids = torch.randint(0, base.config.vocab_size, (1, 6))
```
BenjaminBossan, line 242: Is this part relevant for this test?

tests/test_cartridge.py
```diff
@@ -0,0 +1,158 @@
...
+    out = model(input_ids=input_ids)
+    assert out.logits.shape[:2] == (1, 6)
+
+
+def test_cartridge_gradient_checkpointing_raises():
+    base = _make_tiny_gpt2()
+    base.gradient_checkpointing_enable()
+    peft_config = CartridgeConfig(num_virtual_tokens=4, num_frozen_tokens=0, task_type="CAUSAL_LM")
+
+    try:
```
BenjaminBossan: Let's use: `with pytest.raises(ValueError, match="does not work with gradient checkpointing")`

src/peft/tuners/cartridge/config.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Args:
+        num_frozen_tokens (`int`, defaults to 0):
+            Number of *prefix* tokens at the start of the cartridge to keep frozen (no gradients).
+    """
+
+    num_frozen_tokens: int = field(
+        default=0,
```
BenjaminBossan: I forgot to add this to the initial review: From my understanding, the paper recommends to freeze the first token, which would correspond to a value of 1. Should we set that as the default, and also mention this in the help? The reason seems to be that most LLMs use it as attention sink, so it should not be tempered with.

kashif: agree done!

tests/test_cartridge.py
kashif: so i wanted to run a forward pass through the PEFT-wrapped model, to check if the adapter can still be used in inference mode and the output shape is sensible.

tests/test_cartridge.py
kashif: so i wanted to test that PEFT should mark adapter parameters as non-trainable so we can safely run forward/generation without updating or tracking grads.

src/peft/tuners/cartridge/model.py
kashif: added utils to predix-tuning and its docs

kashif:
I think its fine to ignore their cache concerns as we are using PEFT + Transformers generation its only needed for their cache and flex attention pipeline

docs/source/_toctree.yml
```diff
@@ -59,6 +59,8 @@
     title: Adapters
   - local: conceptual_guides/prompting
     title: Soft prompts
+  - local: conceptual_guides/cartridges
+    title: Cartridges
```
BenjaminBossan: Let's put this section to the "Main classes" section below, where we list all PEFT methods. I know it's a bit messy, but I think that is the better place.

docs/source/conceptual_guides/cartridges.md
BenjaminBossan, line 1: In line with my previous comment, let's move this from `conception_guides` to `package_reference`.

docs/source/package_reference/prefix_tuning.md
```diff
@@ -18,6 +18,28 @@ rendered properly in your Markdown viewer.
 
 [Prefix tuning](https://hf.co/papers/2101.00190) prefixes a series of task-specific vectors to the input sequence that can be learned while keeping the pretrained model frozen. The prefix parameters are inserted in all of the model layers.
 
+## Initialize from a KV cache prefix
```
BenjaminBossan, line 21: Nice, thanks for adding this.

examples/cartridge_self_study/README.md
```diff
@@ -0,0 +1,40 @@
+# CARTRIDGE self-study distillation (example)
+
+This folder shows an **example** workflow for training a `CARTRIDGE` adapter via a SELF‑STUDY‑style
+context-distillation objective (see the Cartridges paper).
```
BenjaminBossan:
```suggestion
context-distillation objective (see the [Cartridges paper](https://huggingface.co/papers/2506.06266)).
```

examples/cartridge_self_study/arxiv_synthesize.py
```diff
@@ -0,0 +1,58 @@
+import argparse
```
BenjaminBossan, line 15: Let's add the copyright notice.

examples/cartridge_self_study/arxiv_train.py
```diff
@@ -0,0 +1,77 @@
+import argparse
```
BenjaminBossan, line 15: Let's add the copyright notice.

examples/cartridge_self_study/synthesize.py
```diff
@@ -0,0 +1,199 @@
+import argparse
```
BenjaminBossan, line 15: Let's add the copyright notice.

examples/cartridge_self_study/synthesize.py
```diff
@@ -0,0 +1,199 @@
...
+import torch
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+
+DEFAULT_SEED_PROMPTS = {
+    "structuring": "Please outline the key sections and how the information is structured.",
+    "summarization": "Please summarize the most important points concisely.",
+    "question": "Ask a question about the document above that requires careful reading, then answer it.",
+    "use_cases": "List a few realistic use cases and provide one concrete example.",
+    "use_case": "List a few realistic use cases and provide one concrete example.",
```
BenjaminBossan: Duplicate

examples/cartridge_self_study/synthesize.py
```diff
@@ -0,0 +1,199 @@
...
+
+        gen_kwargs = {
+            "max_new_tokens": max_new_tokens,
+            "do_sample": temperature > 0,
+            "temperature": max(temperature, 1e-5),
+            "top_p": top_p,
+            "pad_token_id": getattr(tokenizer, "pad_token_id", None) or getattr(tokenizer, "eos_token_id", None),
+        }
+
+        # Generate only with teacher (which has context) - this is the "ground truth" generation.
```
BenjaminBossan: Helpful comments, nice.

examples/cartridge_self_study/train_distill.py
```diff
@@ -0,0 +1,173 @@
+import argparse
```
BenjaminBossan, line 15: Let's add the copyright notice.

examples/cartridge_self_study/train_distill.py
```diff
@@ -0,0 +1,173 @@
...
+        student_out = model(
+            input_ids=student_input_ids,
+            attention_mask=student_attention_mask,
+            use_cache=False,
+        )
+        student_logits = student_out.logits
+
+        batch_losses = []
+        for b in range(student_input_ids.shape[0]):
+            x_len = int(student_attention_mask[b].sum().item())
```
BenjaminBossan: I think we don't need the `item()` call here and below, right? It can prevent torch compilation, so let's remove it unless it's really necessary.

src/peft/tuners/cartridge/utils.py
```diff
@@ -0,0 +1,215 @@
...
+            if getattr(cfg, attr, None) != getattr(base, attr, None):
+                raise ValueError(f"Incompatible CARTRIDGE configs for attribute '{attr}'.")
+
+    weights = [load_peft_weights(str(p), device="cpu") for p in adapter_paths]
+    prompt_embeddings = [w["prompt_embeddings"] for w in weights]
+    composed = torch.cat(prompt_embeddings, dim=0)
+
+    num_virtual_tokens = composed.shape[0]
+    # Preserve the "frozen prefix tokens" count of the first adapter only (matches a single attention-sink prefix).
+    num_frozen_tokens = int(getattr(base, "num_frozen_tokens", 0) or 0)
```
BenjaminBossan: This should be enough, as we already check that the config type is correct:

```suggestion
    num_frozen_tokens = base.num_frozen_tokens
```

Also, just wondering: Say we have one adapter with 1 frozen token and another with 2 frozen tokens, right now we would set it to 1 and ignore 2. I guess it's fine as this is intended for inference only, right?

src/peft/tuners/cartridge/utils.py
```diff
@@ -0,0 +1,215 @@
...
+    out_cfg = replace(base, num_virtual_tokens=num_virtual_tokens, num_frozen_tokens=num_frozen_tokens)
+
+    output_path = Path(output_path)
+    output_path.mkdir(parents=True, exist_ok=True)
+    out_cfg.save_pretrained(str(output_path))
+
+    from peft.utils.constants import SAFETENSORS_WEIGHTS_NAME, WEIGHTS_NAME
```
BenjaminBossan: Is the local import required to avoid circular imports? If not, let's make it global.

src/peft/tuners/cartridge/utils.py
```diff
@@ -0,0 +1,215 @@
...
+
+    if safe_serialization:
+        from safetensors.torch import save_file
```
BenjaminBossan: Can also be global.

src/peft/tuners/prefix_tuning/utils.py
```diff
@@ -0,0 +1,110 @@
...
+            "Initialization from KV cache is not supported for prefix tuning with `prefix_projection=True`."
+        )
+    if num_virtual_tokens is None:
+        num_virtual_tokens = config.num_virtual_tokens
+
+    prompt_embeddings = prompt_embeddings_from_past_key_values(past_key_values, num_virtual_tokens=num_virtual_tokens)
+    loader = getattr(model.prompt_encoder[adapter_name], "load_prompt_embeddings", None)
+    if callable(loader):
+        loader(prompt_embeddings)
+    else:
```
BenjaminBossan: Can this ever happen?

src/peft/tuners/prefix_tuning/utils.py
BenjaminBossan, line 1: So this is basically identical to `cartridge/utils.py`, right? Unless I missed something, the only differences are:

1. Check for the PEFT type
2. Check for `prefix_projection != True`

I think we should just use the same functions for both (I don't care which one to keep). Regarding 1., I think there is no harm in allowing either of the two types, if users want to initialize Cartridges from prefix tuning or vice versa, why not? Regarding 2., the check would be useless for Cartridges, but it also would not cause harm. So I'd say let's reuse the functions.

examples/cartridge_self_study/synthesize.py
```diff
@@ -0,0 +1,199 @@
...
+        # The student will be trained to match this generation without seeing the context.
+        teacher_out = model.generate(teacher_prompt, attention_mask=torch.ones_like(teacher_prompt), **gen_kwargs)
```
BenjaminBossan: I tried this with the arxiv example (both `python examples/cartridge_self_study/arxiv_synthesize.py` and `python examples/cartridge_self_study/synthesize.py --model Qwen/Qwen2.5-0.5B-Instruct --corpus_path examples/cartridge_self_study/data/cartridges.tex --out_jsonl distill.jsonl --num_samples 1024`) and got an error here because `teacher_prompt` was a `transformers.tokenization_utils_base.BatchEncoding`.

BenjaminBossan:
@kashif LMK when the PR is ready to review.

kashif:
yes ready, i tested it out and it works nicely

kashif:
@BenjaminBossan  i needed the `position_ids`  fix cause without it the sampling is not working and i get bad results with models that use RoPE

docs/source/package_reference/cartridges.md
```diff
@@ -0,0 +1,94 @@
...
+Both Prefix Tuning and Cartridges are served by injecting `past_key_values` (a prefix KV cache) into the base model.
+
+- Prefix Tuning learns virtual token embeddings (and optionally an MLP projection) and produces a KV prefix.
+- Cartridges learn the KV prefix itself directly (the per-layer key/value vectors for `p` virtual tokens), and are
+  designed to be initialized from real prefill KV (for example, the first `p` tokens of a corpus/system prompt).
+
+The paper also recommends freezing the first token as an attention sink for stability (`num_frozen_tokens=1` is the
+default).
+
+## Usage (inference)
```
BenjaminBossan, line 37: I'm not quite sure what this section should show. The title mentions "inference" but below, we create a new PEFT model. For inference, we would use `PeftModel.from_pretrained`.

src/peft/tuners/cartridge/utils.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    return initialize_kv_prefix_from_text(
+        model,
+        tokenizer,
+        text=text,
+        adapter_name=adapter_name,
+        num_virtual_tokens=num_virtual_tokens,
+        use_chat_template=use_chat_template,
+        max_length=max_length,
+    )
```
BenjaminBossan: Do we really need these two functions for backwards compatibility? After all, they are being added in this very PR. I would suggest to remove them and update the examples and docs to use `initialize_kv_prefix_from_past_key_values` and `initialize_kv_prefix_from_text` directly.

docs/source/package_reference/cartridges.md
```diff
@@ -0,0 +1,94 @@
...
+- KV from an existing cache: use `initialize_cartridge_from_past_key_values(model, past_key_values=...)` if you already
+  have a `past_key_values` object from a base-model prefill.
+
+## Training
+
+The Cartridges paper proposes a SELF-STUDY distillation objective (a frozen base model provides teacher logits; the
+CARTRIDGE adapter is trained so the student matches the teacher’s next-token distribution over the target segment).
+PEFT keeps training logic out of the core library; see
+`https://github.com/huggingface/peft/tree/main/examples/cartridge_self_study` for a reference workflow.
+The example scripts also work when you use the same checkpoint as both teacher and student (the teacher is the frozen
```
BenjaminBossan: As is, the example always uses student = teacher, so the "also work" part is a bit confusing. Let's rephrase it, e.g.: "The example scripts work by using the frozen base model as the teacher and the adapted model as student, thus sharing the same base model for both."

docs/source/package_reference/prefix_tuning.md
```diff
@@ -18,6 +18,28 @@ rendered properly in your Markdown viewer.
...
+
+from peft import PrefixTuningConfig, get_peft_model, initialize_kv_prefix_from_text
+
+base = AutoModelForCausalLM.from_pretrained("gpt2")
+tok = AutoTokenizer.from_pretrained("gpt2")
+
+peft_cfg = PrefixTuningConfig(task_type="CAUSAL_LM", num_virtual_tokens=20, prefix_projection=False)
+model = get_peft_model(base, peft_cfg)
+
+initialize_kv_prefix_from_text(model, tok, text="...some long context...")
```
BenjaminBossan: This fails with:

```
  File "/home/name/work/forks/transformers/src/transformers/tokenization_utils_base.py", line 3122, in apply_chat_template
    chat_template = self.get_chat_template(chat_template, tools)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/name/work/forks/transformers/src/transformers/tokenization_utils_base.py", line 3304, in get_chat_template
    raise ValueError(
ValueError: Cannot use chat template functions because tokenizer.chat_template is not set and no template argument was passed!
```

When I try to pass `use_chat_template=False`:

```
File ~/work/forks/peft/src/peft/tuners/cartridge/utils.py:72, in prompt_embeddings_from_past_key_values(past_key_values, num_virtual_tokens)
     70 head_dim = key0.shape[3]
     71 if seq_len < num_virtual_tokens:
---> 72     raise ValueError(f"Need at least {num_virtual_tokens} cached tokens, got {seq_len}.")
     74 packed = torch.empty(
     75     num_virtual_tokens,
     76     num_layers * 2,
   (...)     80     dtype=key0.dtype,
     81 )
     82 for layer_idx, (k, v) in enumerate(legacy):

ValueError: Need at least 20 cached tokens, got 5.
```

examples/cartridge_self_study/synthesize.py
```diff
@@ -0,0 +1,360 @@
...
+    if output_path.exists():
+        output_path.unlink()
+
+    for t in seed_prompt_types:
+        if t not in SEED_PROMPTS:
+            raise ValueError(f"Unknown seed prompt type '{t}', expected one of: {sorted(SEED_PROMPTS)}")
+
+    # Pre-generate prompt indices (cycling through seed prompt types)
+    prompt_indices = [i % len(seed_prompt_types) for i in range(num_samples)]
+    random.shuffle(prompt_indices)
```
BenjaminBossan: This makes it so that we can never replicate the output, right? I wonder if we should use torch to shuffle and allow fix the random seed.

examples/cartridge_self_study/synthesize.py
```diff
@@ -0,0 +1,360 @@
...
+    question_messages = [
+        [
+            {"role": "system", "content": corpus_text},
+            {"role": "user", "content": SEED_PROMPTS[seed_prompt_types[prompt_idx]]},
+        ]
+        for prompt_idx in prompt_indices
+    ]
+
+    question_params = SamplingParams(
+        max_tokens=256,
```
BenjaminBossan: I think it's okay to hard-code this param, but let's still make this a constant (same below for `_synthesize_hf`): `max_tokens=MAX_NEW_TOKENS_FOR_QUESTIONS`.

examples/cartridge_self_study/README.md
```diff
@@ -0,0 +1,91 @@
...
+python arxiv_synthesize.py \
+  --model Qwen/Qwen3-4B \
+  --corpus_path /path/to/cartridges.tex \
+  --num_samples 1024 \
+  --use_vllm
+
+# Train cartridge
+python arxiv_train.py \
+  --model Qwen/Qwen3-4B \
+  --document /path/to/cartridges.tex \
```
BenjaminBossan: A quick mention of how to get this file would be nice to have.

examples/cartridge_self_study/synthesize.py
```diff
@@ -0,0 +1,360 @@
...
+            "max_new_tokens": 256,
+            "do_sample": temperature > 0,
+            "pad_token_id": tokenizer.pad_token_id or tokenizer.eos_token_id,
+        }
+        if temperature > 0:
+            gen_kwargs["temperature"] = max(temperature, 1e-5)
+            gen_kwargs["top_p"] = top_p
+
+        with torch.no_grad():
+            question_out = model.generate(question_input, **gen_kwargs)
```
BenjaminBossan, line 251: `question_input` is a dict for me:

```suggestion
            question_out = model.generate(**question_input, **gen_kwargs)
```

examples/cartridge_self_study/synthesize.py
```diff
@@ -0,0 +1,360 @@
...
+
+        question_tokens = question_out[0, question_input.shape[1] :].tolist()
```
BenjaminBossan, line 253: For the same reason as above:
```suggestion
        question_tokens = question_out[0, question_input["input_ids"].shape[1] :].tolist()
```

examples/cartridge_self_study/synthesize.py
```diff
@@ -0,0 +1,360 @@
...
+            [{"role": "user", "content": question}],
+            tokenize=True,
+            add_generation_prompt=True,
+            return_tensors="pt",
+            **CHAT_TEMPLATE_KWARGS,
+        ).to(device)
+
+        with torch.no_grad():
+            answer_out = model.generate(
+                teacher_input,
```
BenjaminBossan, line 277: Same reason as above
```suggestion
                **teacher_input,
```

examples/cartridge_self_study/synthesize.py
```diff
@@ -0,0 +1,360 @@
...
+                max_new_tokens=max_new_tokens,
+                do_sample=False,
+                pad_token_id=tokenizer.pad_token_id or tokenizer.eos_token_id,
+            )
+
+        answer_tokens = answer_out[0, teacher_input.shape[1] :].tolist()
```
BenjaminBossan, line 283: Again, same reason
```suggestion
        answer_tokens = answer_out[0, teacher_input["input_ids"].shape[1] :].tolist()
```

examples/cartridge_self_study/synthesize.py
```diff
@@ -0,0 +1,360 @@
...
+
+        record = {
+            "teacher_input_ids": answer_out[0].tolist(),
+            "student_input_ids": student_input[0].tolist() + answer_tokens,
+            "ctx_len": int(teacher_input.shape[1]) - int(student_input.shape[1]),
```
BenjaminBossan, line 288: And again
```suggestion
            "student_input_ids": student_input["input_ids"][0].tolist() + answer_tokens,
            "ctx_len": int(teacher_input["input_ids"].shape[1]) - int(student_input["input_ids"].shape[1]),
```

src/peft/tuners/cartridge/utils.py
```diff
@@ -0,0 +1,273 @@
...
+from peft.utils import PeftType
+from peft.utils.constants import SAFETENSORS_WEIGHTS_NAME, WEIGHTS_NAME
+from peft.utils.save_and_load import load_peft_weights
+
+
+def _to_legacy_past_key_values(past_key_values: Any):
+    # Support both legacy tuples and transformers.Cache-like objects.
+    if isinstance(past_key_values, (tuple, list)):
+        return past_key_values
+    to_legacy = getattr(past_key_values, "to_legacy_cache", None)
```
BenjaminBossan, line 35: The `to_legacy_cache` method is removed on transformers main :-/

src/peft/peft_model.py
```diff
@@ -1736,8 +1736,12 @@ def forward(
             prefix_attention_mask = torch.ones(batch_size, peft_config.num_virtual_tokens).to(attention_mask.device)
             attention_mask = torch.cat((prefix_attention_mask, attention_mask), dim=1)
         if kwargs.get("position_ids", None) is not None:
-            warnings.warn("Position ids are not supported for parameter efficient tuning. Ignoring position ids.")
-            kwargs["position_ids"] = None
+            if peft_config.peft_type in (PeftType.PREFIX_TUNING, PeftType.CARTRIDGE):
+                # Offset position_ids by num_virtual_tokens to account for the KV cache prefix
+                kwargs["position_ids"] = kwargs["position_ids"] + peft_config.num_virtual_tokens
```
BenjaminBossan, line 1741: For my understanding, this means that there is currently a bug in prefix tuning when it comes to position ids? I wonder if we should have an explicit test for that.

src/peft/peft_model.py
```diff
@@ -2047,7 +2055,7 @@ def generate(self, *args, **kwargs):
                     kwargs = {k: v for k, v in kwargs.items() if k not in self.special_peft_forward_args}
                     outputs = self.base_model.generate(*args, **kwargs)
             else:
-                outputs = self.base_model.generate(**kwargs)
+                outputs = self.base_model.generate(*args, **kwargs)
```
BenjaminBossan, line 2058: ouch

PR #2954, opened 2025-12-14
Add DataLoader to MathQA benchmarking script

ParagEkbote:
Refs #2310 

As described in the [open tasks](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA#open-tasks) for the benchmarking of LoRA methods, I have added a dataloader to [data.py](https://github.com/huggingface/peft/blob/main/method_comparison/MetaMathQA/data.py) script. Before integrating it in [run.py](https://github.com/huggingface/peft/blob/main/method_comparison/MetaMathQA/run.py), I had a few queries:

1. Is the `default_data_collator` collator ideal for benchmarking or something more advanced like `DataCollatorWithPadding` preferred?

2.  Are there any tests needed for checking improvement in throughput or memory needed?

 Could you please review?

cc: @BenjaminBossan

### DIFF
--- a/method_comparison/MetaMathQA/data.py
+++ b/method_comparison/MetaMathQA/data.py
@@ -22,7 +22,8 @@ from functools import partial
 import datasets
 import numpy as np
 from datasets import Dataset, load_dataset
-
+from torch.utils.data import DataLoader
+from transformers import default_data_collator
 
 # with a token limit of 768 for query + response, we have to exclude all texts with length > 1304; this leaves 93.8% of
 # the dataset
@@ -115,3 +116,26 @@ def get_wiki_small(num_samples: int = 100) -> list[str]:
     dataset_head = ds.take(num_samples)
     rows = [row["text"] for row in dataset_head]
     return rows
+
+def make_dataloader(
+    ds: Dataset,
+    *,
+    batch_size: int,
+    shuffle: bool = False,
+    num_workers: int = 0,
+):
+    """
+    Create a PyTorch DataLoader from a Hugging Face Dataset.
+
+    This is intended for benchmarking and evaluation loops.
+    """
+    ds = ds.with_format("torch")
+
+    return DataLoader(
+        ds,
+        batch_size=batch_size,
+        shuffle=shuffle,
+        num_workers=num_workers,
+        collate_fn=default_data_collator,
+        pin_memory=True,
+    )


BenjaminBossan:
Thanks for the PR. Note that this may not be quite as easy to do anymore as we added a `BucketIterator` for more efficient batching:

https://github.com/huggingface/peft/blob/c5a905d097b3c5ba0caf4cfd39cc8e8aeb0d8f56/method_comparison/MetaMathQA/utils.py#L296

Honestly, I'm not sure if adding the standard torch `DataLoader` at this point is still very useful. It does have some nice features like parallelization, but that's really not a huge deal for this dataset. And when it comes to padding, I do remember that it was a bit tricky to get it completely right -- the padding is done manually in multiple places for now. Possibly, this could be helped with a better collator, but it would require good testing to ensure that there is no change to the current logic and we don't have unit tests for this script.

You could give it a try and see if you can simplify/improve the training code this way, but I think the expected value of working on this wouldn't be too big.

ParagEkbote:
So, in the open tasks mentioned, which of the pending tasks are considered to have a higher priority?

cc: @BenjaminBossan

BenjaminBossan:
Hmm, good question. Perhaps it would be a good idea to remove the `DataLoader` item from the list and mention that if anyone wants to contribute, they should open an issue on PEFT first to ask for feedback.

Among the open tasks, I think that 1. ensuring that we get similar results with `Trainer` and 2. investigating AMP would be the most important. But these aren't really contributions to the code base but rather testing out some stuff. Depending on the results, there might be follow up work that leads to a code contribution, but there is no guarantee.

Another way to contribute would be by adding experiment settings, see the discussions at the end of #2310. This requires access to hardware that can run the experiments at a decent speed.

ParagEkbote:
Can the vLLM addition and clean-up of prints also be considered testing issues. Would you like me to open a seperate PR specifying which are test issues and code issues in the readme?

cc: @BenjaminBossan

BenjaminBossan:
> Can the vLLM addition and clean-up of prints also be considered testing issues

Regarding vLLM, over time we have sped up generation times quite well, so it's less of an issue now. As mentioned, for testing, it's still a bit on the slow side. I think the main value of adding vLLM would be more of a learning experience, I'm not sure if it's worth it to add it as a dependency.

Regarding the prints, it's not super high value and I'd have to check how to best stream them, it's more of a personal judgment.

> Would you like me to open a seperate PR specifying which are test issues and code issues in the readme?

You could do a PR with that change, yes. Please also remove the `DataLoader` item and add a sentence about first opening  an issue before starting the work. Thanks!

githubnemo:
*not stale*

PR #2960, opened 2025-12-16
Refactor layer initialization to use PEFT config directly

BenjaminBossan:
## Description

There is a lot of code duplication in the way that adapter layers are initialized. This means that if, say, a new LoRA config argument is added, we need updates in the following places:

1. `LoraModel._create_and_replace`: `kwargs` dict
2. `LoraModel._create_and_replace`: update_layer arguments
3. `LoraModel._create_and_replace`: update_layer signature
4. `Linear.__init__`: signature (same for each LoRA layer type)
5. `Linear.__init__`: update_layer arguments
6. `resolve_lora_variant`: signature if the argument is tied to a LoRA variant

It is far too easy to forget something and it's in fact a common mistake we see, especially by outside contributors.

This PR aims at removing this duplication as much as possible. The crux is that almost all of these arguments derive from the `LoraConfig`. Therefore, the duplication can be reduced by passing the `LoraConfig` directly instead of each of its individual arguments.

As is, this change is only implemented by LoRA, which is by far the most affected by this duplication. However, an argument could be made that for consistency, the other PEFT methods should be refactored accordingly. Since AdaLoRA inherits from LoRA, it was also adapted.

## Possible counter points

With this change, the used arguments are not listed explicitly anymore. But since the `LoraConfig` is a typed `dataclass`, this replacement does not stop us from knowing which arguments are available (unlike, say, using untyped `kwargs`). So for editing support and readability, this replacement should not hurt.

Another possible issue with this proposal is that it could easily break some functionality, as it can be easy to overlook something. Not everything is covered by tests (some of the more rarely used quantized layer types, TP layer), so it's possible to miss this.

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -124,9 +124,12 @@ class LoraLayer(BaseTunerLayer):
         self.kwargs = kwargs
 
         base_layer = self.get_base_layer()
-        in_features, out_features = _get_in_out_features(base_layer)
+        in_features, out_features = self._get_in_out_features(base_layer)
         self.in_features = in_features
         self.out_features = out_features
+
+    def _get_in_out_features(self, module: nn.Module) -> tuple[int, int] | tuple[None, None]:
+        return _get_in_out_features(module)
 
     def resolve_lora_variant(self, *, config: LoraConfig, **kwargs) -> Optional[LoraVariant]:
         """Return a matching LoRA variant for this layer type.
@@ -2005,20 +2008,10 @@ class ParamWrapper(nn.Module, LoraLayer):
         is_target_conv_1d_layer: bool = False,
         **kwargs,
     ) -> None:
+        self.parameter_name = parameter_name
         super().__init__()
         LoraLayer.__init__(self, base_layer, **kwargs)
-        self.parameter_name = parameter_name
-        param = self.get_param()
-        if param.ndim == 3:
-            self.num_experts, self.in_features, self.out_features = param.shape
-        else:
-            self.num_experts, self.in_features, self.out_features = 1, param.shape[1], param.shape[0]
 
-        if param.ndim not in (2, 3):
-            raise ValueError(
-                f"lora.{self.__class__.__name__} was initialized with {param.ndim} dimensional Parameter, but only 2d "
-                "and 3d are supported."
-            )
         if config.lora_dropout:
             # It's not possible to factor out x from lora_B(lora_A(dropout(x))), so dropout can't be correctly
             # implemented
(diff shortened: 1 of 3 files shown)


src/peft/tuners/lora/layer.py
```diff
@@ -143,26 +143,20 @@ def resolve_lora_variant(self, *, use_dora: bool, use_bdlora=None, **kwargs) ->
...
+        config: LoraConfig,
         **kwargs,
-    ):
+    ) -> None:
         # collect the kwargs
+        lora_dropout = config.lora_dropout
+        init_lora_weights = config.init_lora_weights
+        use_rslora = config.use_rslora
+        lora_bias = config.lora_bias
+        inference_mode = config.inference_mode
```
BenjaminBossan, line 161: Note: I added variables like these to reduce diff noise. LMK if you'd rather remove those and just use `config.lora_dropout` etc. below.

src/peft/tuners/lora/layer.py
```diff
@@ -143,26 +143,20 @@ def resolve_lora_variant(self, *, use_dora: bool, use_bdlora=None, **kwargs) ->
...
-        use_alora: bool = False,
-        use_qalora: bool = False,
-        lora_bias: bool = False,
-        arrow_config: ArrowConfig = None,
-        qalora_group_size: int = 32,
-        inference_mode: bool = False,
-        use_bdlora=None,
+        adapter_name: str,
+        r: int,
+        lora_alpha: int,
```
BenjaminBossan, line 152: Note: r and alpha can differ from the one in the config (e.g. because of `rank_pattern`), thus they are not taken from the config.

githubnemo (commented):
I think this looks fine but I noticed that `lora/gptq.py` seems unmodified still, is this on purpose?

BenjaminBossan:
@githubnemo Thanks for pointing this out, GPTQ is now also updated. I ran the GPTQ tests locally (as they are GPU tests) and they passed.

BenjaminBossan:
@hiyouga Just a heads up, this PR simplifies the signature of multiple methods like `update_layer`. The idea is to directly pass the PEFT config instead of unpacking the arguments, which saves us a lot of code and makes PEFT easier to extend. For now, this is only for LoRA, but it will be extended to other PEFT methods later. In cause you're overriding these methods, please take a look at the new signature. From skimming LlamaFactory, it looks like you should be safe, but please check your tests using the PEFT main branch to be safe. If you encounter issues, please let us know.

hiyouga:
@BenjaminBossan Thanks for the heads-up! All unit tests in LlamaFactory pass with the PEFT main branch, and we haven't encountered any issues.

PR #2962, opened 2025-12-16
FIX: Inject from state dict into compiled model

### LINKED ISSUE #2957: torch compiled model cannot have LoRA added without workaround
### System Info

hello, when using torch inductor to compile the entire transformer (Diffusers) it was observed that the target modules are claimed to not exist.

this is because the state dict returned of the OptimizedModel has the orig_mod prefix applied.

the following script demonstrates the issue as well as a horrible workaround that duplicates the dict keys - it seems the proper way to handle this is to check for OptimizedModel and then unwrap before applying the LoraConfig?

### Who can help?

_No response_

### Reproduction

```py
"""
Minimal repro: torch.compile + PEFT target-module resolution mismatch.

Expected: LoRA adapter injects into `transformer_blocks.0.attn.to_q`.
Observed: After torch.compile, module names gain `_orig_mod.` prefix so PEFT
          cannot match targets and raises:
          ValueError: Target modules {'transformer_blocks.0.attn.to_q'} not found in the base model.

Run:
    .venv/bin/python repros/peft_dynamo_target_mismatch.py
"""

import torch
from peft import LoraConfig
from peft import inject_adapter_in_model


class Toy(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.transformer_blocks = torch.nn.ModuleList(
            [
                torch.nn.ModuleDict(
                    {
                        "attn": torch.nn.ModuleDict(
                            {
                                "to_q": torch.nn.Linear(4, 4),
                                "to_k": torch.nn.Linear(4, 4),
                                "to_v": torch.nn.Linear(4, 4),
                                "to_out": torch.nn.ModuleList([torch.nn.Linear(4, 4)]),
                            }
                        )
                    }
                )
            ]
        )

    def forward(self, x):
        return self.transformer_blocks[0]["attn"]["to_q"](x)


def main():
    # Target the pre-compile module name.
    lora_cfg = LoraConfig(r=2, lora_alpha=2, target_modules=["transformer_blocks.0.attn.to_q"])
    # Minimal LoRA state dict matching the uncompiled module path.
    lora_state = {
        "transformer_blocks.0.attn.to_q.lora_A.weight": torch.zeros(2, 4),
        "transformer_blocks.0.attn.to_q.lora_B.weight": torch.zeros(4, 2),
    }

    model = torch.compile(Toy(), backend="inductor")
    print("Compiled module names containing 'to_q':")
    print([name for name, _ in model.named_modules() if "to_q" in name])

    # Monkeypatch/augment: duplicate LoRA state_dict keys with the `_orig_mod.` prefix
    # so PEFT can match compiled module names.
    def _with_orig_mod_prefix(sd: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
        patched = dict(sd)
        for key, value in list(sd.items()):
            patched[f"_orig_mod.{key}"] = value
        return patched

    patched_state = _with_orig_mod_prefix(lora_state)
    inject_adapter_in_model(lora_cfg, model, adapter_name="default", state_dict=patched_state)
    print("Adapter injection succeeded under torch.compile with prefixed state dict.")


if __name__ == "__main__":
    main()
```

### Expected behavior

The adapter should load without caller-level workaround.

BenjaminBossan:
Resolves #2957.

If we directly inject the adapter into the model from a `state_dict`, if the model is compiled, the keys would not match because they contain the `'_orig_mod.'` prefix from the compilation. This PR addresses this issue.

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -771,7 +771,7 @@ class BaseTuner(nn.Module, ABC):
             prefix = PEFT_TYPE_TO_PREFIX_MAPPING[peft_config.peft_type]
             # Find the module name from the state_dict. Also defensively remove '_orig_mod.', which might be inserted if
             # the model was torch.compiled beforehand
-            module_names = {k.rsplit("." + prefix, 1)[0].replace("_orig_mod.", "") for k in state_dict}
+            module_names = {k.rsplit("." + prefix, 1)[0].removeprefix("_orig_mod.") for k in state_dict}
 
         for key, module in named_modules:
             if not key:


src/peft/tuners/tuners_utils.py
```diff
@@ -769,7 +769,9 @@ def inject_adapter(
         module_names: set[str] = set()
         if state_dict is not None:
             prefix = PEFT_TYPE_TO_PREFIX_MAPPING[peft_config.peft_type]
-            module_names = {k.rsplit("." + prefix, 1)[0] for k in state_dict}
+            # Find the module name from the state_dict. Also defensively remove '_orig_mod.', which might be inserted if
+            # the model was torch.compiled beforehand
+            module_names = {k.rsplit("." + prefix, 1)[0].replace("_orig_mod.", "") for k in state_dict}
```
githubnemo: Why can we do `removeprefix` below but here we use `replace`? If possible it'd be nice to only use `removeprefix` to avoid accidental removes

BenjaminBossan: Good point, done. Please check again.

githubnemo (approved):

PR #2978, opened 2026-01-08
fix oft gptq forward

jiqing-feng:
Fixed incorrect logic in GPTQOFTLinear.forward.
Previously, self.quant_linear_module(x) was executed twice, and the input casting was misplaced. This change removes the redundant call and ensures the input is cast to expected_dtype before being passed to the linear module, aligning the implementation with AwqOFTLinear.


jiqing-feng:
Hi @BenjaminBossan . Could you please review this PR? Thanks!

PR #2979, opened 2026-01-08
fix device map check

jiqing-feng:
Setting device_map='auto' does not guarantee that the model will be split across all available cards.. For example, I load gpt2 model in the A100 and select 6 cards (all 6 cards are available).
CMD: `CUDA_VISIBLE_DEVICES=1,2,3,4,5,6 python test_device.py`
```python
import torch

model = AutoModelForCausalLM.from_pretrained("gpt2", dtype=torch.float16, device_map="auto")
print(model.hf_device_map)
```

output: `{'': 5}`

With the latest change https://github.com/huggingface/transformers/pull/43019 , single device will not assign hf_device_map for the model, so we need to check if the model has the attr hf_device_map.

### DIFF
--- a/tests/test_gptqmodel.py
+++ b/tests/test_gptqmodel.py
@@ -41,11 +41,11 @@ from peft.tuners.lora import GPTQLoraLinear
 from peft.utils import SAFETENSORS_WEIGHTS_NAME, infer_device
 
 from .testing_utils import (
+    DEVICE_MAP_MAP,
     load_dataset_english_quotes,
     require_gptqmodel,
     require_optimum,
     require_torch_multi_accelerator,
-    DEVICE_MAP_MAP,
 )
 
 
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -88,6 +88,7 @@ from peft.utils.other import fsdp_auto_wrap_policy
 from tests.testing_utils import hub_online_once
 
 from .testing_utils import (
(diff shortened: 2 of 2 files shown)


jiqing-feng:
Hi @BenjaminBossan . Would you please review this PR? Thanks!

BenjaminBossan:
Thanks for this PR. I wonder if this is the right solution. These tests are explicitly for multi accelerator, but the case you describe means that the whole model is on a single device. Therefore, the test would not be testing what it is supposed to test.

I think a better solution would be to force the model to be on exactly two devices (which is what we have on our GPU runners) and then the existing assert should work. We already do this for other tests, see this:

https://github.com/huggingface/peft/blob/b4750c719c7bb7c7a7d78c46694ab9da74e84f12/tests/test_gpu_examples.py#L117

I believe this would be the better fix for the issue, WDYT?

Back then when I added this, I also wrote a small helper function to create the device map, it might work here too:

https://github.com/huggingface/peft/pull/2612#issue-3175553546

jiqing-feng:
Hi @BenjaminBossan . I have changed the tests by your comment, please review it. Thanks!

BenjaminBossan:
Thanks for the update. I ran the corresponding GPU tests locally and they passed. For the CI, could you please run `make style`?

jiqing-feng:
Hi @BenjaminBossan . I've updated the style. Please review it. Thanks!

PR #2984, opened 2026-01-08
Updated MetaMathQA results

githubnemo:
Executed all experiments anew and set autocast_adapter_dtype to false for OSF since it previously was disabling this behavior on its own.


BenjaminBossan (approved):
Thanks for running the new experiments. I skimmed all the diffs to ensure that there are the no surprising jumps in metrics. The train losses and max memory reserved are pretty stable, which is good and what we expected. Test accuracies are also close but fluctuate a bit more (expected, as they use generations). Elapsed time is significantly smaller, even though each individual time for train/eval is approximately roughly the same. This should thus be due to our improvement from #2894.

The new forgetting metric can vary quite a lot between methods, though most cluster between 0 and 0.5. Surprisingly, some methods have negative forgetting score, although that could be random noise due to the size of the dataset. However, trainable tokens have a forgetting score of -3.3, which I have no explanation for. On the other hand, the worst forgetting scores can be found for p-tuning and prompt-tuning, which is not something I would have predicted.

There doesn't seem to be a clear correlation between forgetting and other metrics (e.g. no tradeoff between forgetting and test accuracy or trainable params). I think we should revisit this metric sometime in the future to see if it can be improved.

image

Issue #2991, opened 2026-01-13
CoRDA initialization lacks support for Conv1D layers in older models like GPT-2


sambhavnoobcoder:
### System Info

Software : latest with main branch
Hardware: MacOS , ran on CPU 

### Who can help?

@BenjaminBossan @githubnemo 

### Reproduction

```python
"""
Bug reproducer: CoRDA fails with Conv1D layers (GPT-2)
Issue: CoRDA only checks for nn.Linear, missing Conv1D support that EVA has
"""
from transformers import AutoModelForCausalLM
from peft import LoraConfig
from peft.tuners.lora.corda import target_modules as corda_target_modules

# Load GPT-2 (uses Conv1D layers)
model = AutoModelForCausalLM.from_pretrained("gpt2")

# Configure targeting Conv1D layers
lora_config = LoraConfig(
    r=8,
    lora_alpha=16,
    target_modules=["c_fc", "c_proj"],  # These are Conv1D layers in GPT-2
    task_type="CAUSAL_LM"
)

# Check how many modules CoRDA detects
corda_modules = list(corda_target_modules(model, lora_config))

print(f"CoRDA detected modules: {len(corda_modules)}")
print(f"Expected: 36 Conv1D layers (c_fc and c_proj in 12 transformer blocks)")

if len(corda_modules) == 0:
    print("\nBUG CONFIRMED: CoRDA detects 0 modules")
else:
    print("\n Fixed!")

```

Error occurs because CoRDA initialization does not transpose weights for fan_in_fan_out=True, leading to shape mismatch in SVD or decomposition. For comparison, switching to init_lora_weights="eva" works without error.

### Expected behavior

CoRDA should handle weight transposition for models using Conv1D (fan_in_fan_out=True) like GPT-2, similar to how standard LoRA and EVA do. The initialization should succeed without dimension mismatch errors, allowing fine-tuning on older transformers.

PR #2992, opened 2026-01-13
Add Conv1D support to CoRDA for GPT-2 compatibility #2991

sambhavnoobcoder:
## Summary

This PR fixes a bug in CoRDA's `target_modules()` function that prevented it from working with models using `Conv1D` layers, such as GPT-2. CoRDA now correctly detects and processes `Conv1D` layers, matching the behavior of EVA.

### The Problem

CoRDA failed to work with GPT-2 and other models that use `Conv1D` layers (a specialized linear layer used in older transformer architectures). When attempting to apply CoRDA to GPT-2's `c_fc` and `c_proj` layers, the preprocessing would detect **zero modules** instead of the expected 36 Conv1D layers.

### Discovery Process

While reviewing CoRDA and a few other LoRA variants , I noticed that CoRDA failed on GPT-2, while EVA worked perfectly. After methodical investigation:

1. **Tested multiple LoRA variants** - Confirmed EVA and other implementations work with GPT-2
2. **Examined both codebases** - Found the critical difference in layer detection logic
3. **Created minimal reproducer** - Isolated the exact issue to the `target_modules()` function

### Root Cause

The issue was in `src/peft/tuners/lora/corda.py` line 45:

**CoRDA (broken):**
```python
if LoraModel._check_target_module_exists(config, name) and isinstance(module, nn.Linear):
```

**EVA (working):**
```python
return isinstance(module, (torch.nn.Linear, Conv1D)) and is_target_module
```

CoRDA only checked for `nn.Linear`, missing GPT-2's 36 `Conv1D` layers entirely.

### Changes Made

1. **Import Conv1D** - Added `from transformers.pytorch_utils import Conv1D`
2. **Updated isinstance check** - Changed to `isinstance(module, (nn.Linear, Conv1D))`
3. **Updated type hint** - Changed `collect_eigens_for_layer` parameter from `nn.Linear` to `nn.Module` for accuracy
4. **Updated docstring** - Clarified that both `nn.Linear` and `Conv1D` are supported

### Files Modified

- `src/peft/tuners/lora/corda.py` (4 lines changed)

### Before Fix
```
CoRDA detected modules: 0
Expected: 36 Conv1D layers (c_fc and c_proj in 12 transformer blocks)

BUG CONFIRMED: CoRDA detects 0 modules
```

### After Fix
```
CoRDA detected modules: 36
Expected: 36 Conv1D layers (c_fc and c_proj in 12 transformer blocks)

Fixed!
```

### Test Coverage

- ✅ Module detection now correctly identifies all 36 Conv1D layers in GPT-2
- ✅ Verified against the exact same approach used in EVA (already in production)
- ✅ No breaking changes to existing Linear layer support

## Screenshots

image

**Fixes** : #2991 

**cc:** @BenjaminBossan

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -430,7 +430,7 @@ class LoraLayer(BaseTunerLayer):
         # For Conv1D, lora_B @ lora_A gives (out_dim, in_dim) but weight is (in_dim, out_dim)
         # So we need to transpose before subtraction
         delta = self.scaling[adapter_name] * lora_B @ lora_A
-        delta = transpose(delta, fan_in_fan_out=isinstance(linear, Conv1D))
+        delta = transpose(delta, fan_in_fan_out=self.fan_in_fan_out)
         weight = weight.data - delta
         weight = weight.to(dtype)
         self.get_base_layer().weight.data = weight


BenjaminBossan:
Thanks for spotting this and providing a PR to fix it. As you mentioned in the issue, with `Conv1D` the weight is transposed compared to `Linear` layers. Shouldn't this be accounted for in the `collect_eigens_for_layer` function?

sambhavnoobcoder:
Yes @BenjaminBossan  , i missed this . thanks for this comment . tested the full pipeline , seems fine now . Do let me know if any other changes are needed as well . i'll make those asap .

sambhavnoobcoder:
Thank you for the reproducer @BenjaminBossan , it was very helpful in identifying where changes were required ., i think the fix and its respective tests earlier only to check and sure that CoRDA could find the 1D layers , whereas it also needed changes to run preprocess_corda + get_peft_model end to end , ensuring that the SVD decomposition worked properly or not . 
I have made the necessary changes now . the reproducer now passes with the following output : 
image


as well as added two unit tests for validation of the same : 
image

I would appreciate your reviews and any suggested  changes on the same .

src/peft/tuners/lora/layer.py
```diff
@@ -421,7 +426,13 @@ def corda_init(self, adapter_name, init_lora_weights):
         lora_B = U.mul(S.sqrt()).contiguous()
         self.lora_A[adapter_name].weight.data = lora_A
         self.lora_B[adapter_name].weight.data = lora_B
-        weight = weight.data - self.scaling[adapter_name] * lora_B @ lora_A
+
+        # For Conv1D, lora_B @ lora_A gives (out_dim, in_dim) but weight is (in_dim, out_dim)
+        # So we need to transpose before subtraction
+        delta = self.scaling[adapter_name] * lora_B @ lora_A
```
BenjaminBossan, line 432: Let's use `transpose` here as this is the general pattern we employ for "normalizing" the weight orientation.

tests/test_initialization.py
```diff
@@ -3300,6 +3300,48 @@ def test_lora_corda_alpha_pattern_and_rslora_raises(self, data, tmp_path, corda_
                 tmp_path / "corda-model", path_initial_model_for_weight_conversion=tmp_path / "init-model"
             )
 
+    @pytest.mark.parametrize("corda_method", ("ipm", "kpm"))
+    def test_lora_corda_conv1d_gpt2(self, tmp_path, corda_method):
+        """Test that CoRDA works with Conv1D layers (GPT-2)."""
+        from transformers import AutoModelForCausalLM, AutoTokenizer
+
+        model = AutoModelForCausalLM.from_pretrained("gpt2")
```
BenjaminBossan: Let's use `"peft-internal-testing/tiny-random-gpt2"`.

tests/test_initialization.py
```diff
@@ -3300,6 +3300,48 @@ def test_lora_corda_alpha_pattern_and_rslora_raises(self, data, tmp_path, corda_
                 tmp_path / "corda-model", path_initial_model_for_weight_conversion=tmp_path / "init-model"
             )
 
+    @pytest.mark.parametrize("corda_method", ("ipm", "kpm"))
+    def test_lora_corda_conv1d_gpt2(self, tmp_path, corda_method):
+        """Test that CoRDA works with Conv1D layers (GPT-2)."""
+        from transformers import AutoModelForCausalLM, AutoTokenizer
```
BenjaminBossan: Let's make the imports global.

tests/test_initialization.py
```diff
@@ -3300,6 +3300,48 @@ def test_lora_corda_alpha_pattern_and_rslora_raises(self, data, tmp_path, corda_
...
+
+        model = AutoModelForCausalLM.from_pretrained("gpt2")
+        tokenizer = AutoTokenizer.from_pretrained("gpt2")
+
+        # Create sample input
+        text = "Hello world"
+        inputs = tokenizer(text, return_tensors="pt")
```
BenjaminBossan: For a test like this, we can just create a random input with torch, no need to load the tokenizer.

tests/test_initialization.py
```diff
@@ -3300,6 +3300,48 @@ def test_lora_corda_alpha_pattern_and_rslora_raises(self, data, tmp_path, corda_
...
+            corda_method=corda_method,
+        )
+        config = LoraConfig(
+            init_lora_weights="corda",
+            target_modules=["c_fc"],  # Conv1D layer in GPT-2
+            fan_in_fan_out=True,  # Required for Conv1D
+            corda_config=corda_config,
+        )
+
+        # Preprocess with CoRDA
```
BenjaminBossan: Let's reference the issue here and mention that this is the thing that failed beforehand.

tests/test_initialization.py
```diff
@@ -3300,6 +3300,48 @@ def test_lora_corda_alpha_pattern_and_rslora_raises(self, data, tmp_path, corda_
...
+        preprocess_corda(
+            model,
+            config,
+            run_model=lambda: model(**inputs),
+        )
+
+        # Create PEFT model
+        peft_model = get_peft_model(model, config)
+
+        # Check that adapter performs identity transformation initially
```
BenjaminBossan: Let's mention that this is just a sanity check, the important step we test is the preprocessing above.

src/peft/tuners/lora/layer.py
sambhavnoobcoder: fixed in [a022b1a](https://github.com/huggingface/peft/pull/2992/commits/a022b1a572b7c0e1436192ea1a1e6ab94f13996d)

tests/test_initialization.py
sambhavnoobcoder: done in [17f8517](https://github.com/huggingface/peft/pull/2992/commits/17f85173a4afe0756bf33841c15c0e89cfea4e4c)

tests/test_initialization.py
sambhavnoobcoder: done in [17f8517](https://github.com/huggingface/peft/pull/2992/commits/17f85173a4afe0756bf33841c15c0e89cfea4e4c)

tests/test_initialization.py
sambhavnoobcoder: done in [bc35201](https://github.com/huggingface/peft/pull/2992/commits/bc35201508990df558f8434a03277ce761c4cd53)

tests/test_initialization.py
sambhavnoobcoder: done in [bc35201](https://github.com/huggingface/peft/pull/2992/commits/bc35201508990df558f8434a03277ce761c4cd53)

tests/test_initialization.py
sambhavnoobcoder: done in [bc35201](https://github.com/huggingface/peft/pull/2992/commits/bc35201508990df558f8434a03277ce761c4cd53)

sambhavnoobcoder:
Thank you @BenjaminBossan . I've made all the changes from the comments . kindly review this and let me know if this needs any more changes . i'll make those asap .

src/peft/tuners/lora/layer.py
```diff
@@ -421,7 +426,12 @@ def corda_init(self, adapter_name, init_lora_weights):
...
+        delta = transpose(delta, fan_in_fan_out=isinstance(linear, Conv1D))
```
```suggestion
        delta = transpose(delta, fan_in_fan_out=self.fan_in_fan_out))
```

This should work, right?

sambhavnoobcoder: done in [6b4427b](https://github.com/huggingface/peft/pull/2992/commits/6b4427b74fc94b503a35dbd2256e961182226cb9)

sambhavnoobcoder:
Thank you for the fast reviews and the quick merge !!!

PR #2995, opened 2026-01-14
[WIP] Support transformers weight conversion

BenjaminBossan:
See https://github.com/huggingface/transformers/issues/42491 and https://github.com/huggingface/transformers/pull/43261.

### DIFF
--- a/src/peft/utils/integrations.py
+++ b/src/peft/utils/integrations.py
@@ -14,9 +14,10 @@
 
 from __future__ import annotations
 
+import copy
 import functools
 from contextlib import contextmanager
-from typing import Literal, Optional
+from typing import Any, Literal, Optional
 
 import packaging.version
 import torch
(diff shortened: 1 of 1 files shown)


src/peft/utils/integrations.py
```diff
@@ -185,6 +186,104 @@ def map_cache_to_layer_device_map(model, cache) -> None:
...
+    # transformers v5
+    from transformers.core_model_loading import WeightRenaming, rename_source_key
+
+    ############
+    # renaming #
+    ############
+
+    if not isinstance(peft_config.target_modules, str):
+        # if target_modules is a string (=> regex), the required conversion is not trivial, we don't deal with this case
+        # for now
```
githubnemo, line 270: For the string case it is possible to load the model, get all lora layer names and therefore convert the regex to a list of names

BenjaminBossan:
_not stale_

BenjaminBossan:
Superseded by #3071.

Issue #2907, opened 2025-11-12
Integrate: LoRA vs Full Fine-tuning: An Illusion of Equivalence


BenjaminBossan:
# Feature request

In the paper [LoRA vs Full Fine-tuning: An Illusion of Equivalence](https://arxiv.org/abs/2410.21228), the authors describe a method to detect the degree of forgetting caused by LoRA by identifying so called "intruder dimensions". They also describe a method to mitigate this, at the cost of possibly reducing performance on the fine-tuning task.

For PEFT, it could be interesting to add a new function that implements this mitigation on a fine-tuned LoRA model (and possibly for PEFT methods). This issue is for discussing this paper and the possible applications to PEFT. It is also a call for contributions -- if you are interested in creating a PR, please post here before starting the work to avoid duplicate effort.

## Design

The proposed mitigation is to identify the intruder dimensions on $\Delta W$ and scale them down. At the moment, it is unclear to me if, after downscaling, the $\Delta W$ can be projected back onto the original `lora_A` and `lora_B`. Depending on whether this is possible, I see two ways of handling this.

### Back projection possible

In this case, we can add a new LoRA adapter that is a copy of the existing, fine-tuned one, apply the mitigation, then project back onto `lora_A` and `lora_B` of this adapter:

```python
def reduce_intruder_dimension(
    peft_model, old_adapter_name="default", new_adpater_name="new_adapter", top_k=10, threshold_epsilon=0.5, mitigation_lambda=0.75,
):
    # check if peft method is supported
    peft_model.add_adapter(new_adpater_name, peft_model.peft_config[old_adapter_name])
    # copy LoRA weights to the new adapter
    # apply mitigation to new adapter
    peft_model.set_adapter(new_adpater_name)
    return peft_model
```

For the `mitigation_lambda` parameter, check section 5 of the paper. `top_k` is the number of singular vectors to check (could be float for top k%), epsilon is the similarity threshold below which we consider a vector to be an intruder (see [original code](https://github.com/reeceshuttle/intruder-dimensions/blob/9b83bf208a124eb9d33f7a680d7924f1f1700953/find_intruder_dimensions.py#L6)).

### Back projection not possible

If back projection cannot be done, I would suggest to merge and unload with the new weights, as unmerging would not be trivially possible (but could possibly be added later).

```python
def merge_and_unload_with_reduced_intruder_dimensions(
[...]

PR #2999, opened 2026-01-16
Intruder dimension reduction for LoRA

githubnemo:
Implementation for https://github.com/huggingface/peft/issues/2907.

After some experimentation I think this is an implementation of the 
forgetting through intruder dimensions mitigation presented by the paper
[LoRA vs Full Fine-tuning: An Illusion of Equivalence](https://huggingface.co/papers/2410.21228).

The implementation takes a model with a loaded LoRA adapter and applies the
mitigation to form a new adapter on the same model. This makes it possible
to compare the results directly.

You can find a script to evaluate the changes here: https://gist.github.com/githubnemo/5932e99125d498c4f353017916ffc3ea

Failing docs build is unrelated and addressed in https://github.com/huggingface/peft/pull/3003.

## Alternative approaches

I can think of two possible modes how the conversion could be done differently.
The first one is already proposed in the original issue, namely applying the
mitigation on the merged weights. In this case we will lose the ability to
unmerge and modify the adapter directly which loses all the flexibility of
an adapter.

The second way is to modify A/B directly. Since we have the intruder dimensions
and compute the SVD of (W+dW) to build the mitigation vectors it is conceivable
that it also possible to apply the mitigation directly to the adapter's A/B
matrices. Initial testing wasn't successful so I chose the easier (but 
computationally more intensive) route of first merging the weights, applying
the mitigation and then re-computing the delta weight (and subsequently the
A/B matrices using SVD of the adapter's rank).

### DIFF
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -551,7 +551,7 @@ print(outputs)
 The paper [LoRA vs Full Fine-tuning: An Illusion of Equivalence](https://huggingface.co/papers/2410.21228) argues
 that LoRA training introduces extra dimensions into the weights that have very little in common with the already
 learnt weights and lead to forgetting of already learned information. PEFT implements the suggested mitigation
-as [`peft.tuners.lora.intruders.reduce_intruder_dimension`](../package_reference/lora#Intruder Dimension Reduction).
+in [`peft.tuners.lora.intruders.reduce_intruder_dimension`].
 
 The mitigation will take a PEFT model with a loaded LoRA and create a new, modified adapter that is loaded alongside
 the existing adapter and now the active adapter.


githubnemo:
@reeceshuttle maybe you can take a look at this, I would appreciate the feedback :)

src/peft/tuners/lora/intruders.py
```diff
@@ -0,0 +1,155 @@
...
+        A_intruder = (torch.diag(S_merged)[:, intruder_idcs]).sqrt() @ V_merged[intruder_idcs, :]
+
+        # apply mitigation and recover dW = (B@A).
+        # (W+dW+mitigation)-W = dW+mitigation, so we can convert dW back to A/B using SVD
+        # since we know the effective rank from the adapter config.
+        #
+        # note that we also remove the scaling from dW which is applied in get_delta_weight() since
+        # it impacts mitigation performance both in task accuracy and forgetting.
+        W_merged += (mitigation_lambda - 1) * (B_intruder @ A_intruder)
+        W_merged -= W
```
reeceshuttle: maybe rename `W_merged` to something like `W_edited`? since its not W_merged anymore.

src/peft/tuners/lora/intruders.py
```diff
@@ -0,0 +1,155 @@
...
+        W_merged /= layer.scaling[old_adapter_name]
```
reeceshuttle: Why do we divide the weights by the scaling factor? Im not familiar with PEFT specifics so doing this could be completely correct, just wanted to call attention to this in case scaling is done at inference instead of editing the weights and if this should be a multiplication instead of division (in lora, `W_final = W + (alpha/r)*B@A`). I could be misinterpreting this line of code though.

src/peft/tuners/lora/intruders.py
```diff
@@ -0,0 +1,155 @@
...
+
+        U_dW, S_dW, V_dW = torch.linalg.svd(W_merged, full_matrices=False)
+
+        effective_rank = layer.r[old_adapter_name]
+        B_new = U_dW[:, :effective_rank] * S_dW[:effective_rank]
```
reeceshuttle: should we multiply both `B_new` and `A_new` by $\sqrt{S_{dw}}$ instead of just `B_new` by `S_dw`? Current code will work at inference, but my concern is that having all scaling in one adapter will mean `B_new` will have much higher norm than `A_new` and could result in odd training dynamics if training is restarted. this is speculative though.

src/peft/tuners/lora/intruders.py
githubnemo: Yep! Suggestion:

```diff
diff --git a/src/peft/tuners/lora/intruders.py b/src/peft/tuners/lora/intruders.py
index cc8ea364..5fe596b7 100644
--- a/src/peft/tuners/lora/intruders.py
+++ b/src/peft/tuners/lora/intruders.py
@@ -130,11 +130,11 @@ def reduce_intruder_dimension(
         #
         # note that we also remove the scaling from dW which is applied in get_delta_weight() since
         # it impacts mitigation performance both in task accuracy and forgetting.
-        W_merged += (mitigation_lambda - 1) * (B_intruder @ A_intruder)
-        W_merged -= W
-        W_merged /= layer.scaling[old_adapter_name]
+        W_mitigated = W_merged + (mitigation_lambda - 1) * (B_intruder @ A_intruder)
+        dW_mitigated = W_mitigated - W
+        dW_mitigated /= layer.scaling[old_adapter_name]
 
-        U_dW, S_dW, V_dW = torch.linalg.svd(W_merged, full_matrices=False)
+        U_dW, S_dW, V_dW = torch.linalg.svd(dW_mitigated, full_matrices=False)
 
         effective_rank = layer.r[old_adapter_name]
         B_new = U_dW[:, :effective_rank] * S_dW[:effective_rank]
```

src/peft/tuners/lora/intruders.py
githubnemo: I think it is correct: We're using `layer.get_delta_weight` to get `dW` which already applies the scaling (ref [lora/layer.py](https://github.com/huggingface/peft/blob/2e37b403634ce2875e49a6ea28b2d1c0ae087218/src/peft/tuners/lora/layer.py#L886)). So, I think this makes `dW = B@A * scaling` and therefore we want to remove the scaling before doing SVD to get A/B back since the scaling would be applied twice otherwise.

src/peft/tuners/lora/intruders.py
githubnemo: Thanks for raising this, it's a good point. We had a similar discussion a while ago for the implementation of #2939 where we settled on doing it in the same way it is done in the snippet above since it made no difference during inference. I'm not sure if we tested training stability but it might have an effect. I think we can just scale both matrices by $\sqrt{S_{dW}}$ to be on the safe side.

src/peft/tuners/lora/intruders.py
reeceshuttle: nice! this checks out in that case.

githubnemo:
@reeceshuttle thanks for the thorough review :)

I pushed the changes and reran the script in the linked gist to confirm that the changes didn't change the experiment outcome (they didn't).

Can you review again?

reeceshuttle (commented):
Looks good to me! Excited to have this in PEFT.

docs/source/developer_guides/lora.md
```diff
@@ -604,6 +605,40 @@ outputs = tokenizer.batch_decode(generate_ids, skip_special_tokens=True, clean_u
...
+
+The mitigation will take a PEFT model with a loaded LoRA and create a new, modified adapter that is loaded alongside
+the existing adapter and now the active adapter.
+
+Example usage:
+
+```python
+from peft.tuners.lora.intruders import reduce_intruder_dimension
+
+peft_model = AutoPeftModelForCausalLM.from_pretrained('my-account/my-lora-adapter')
```
sayakpaul: Can we provide an actual working checkpoint here?

docs/source/developer_guides/lora.md
```diff
@@ -604,6 +605,40 @@ outputs = tokenizer.batch_decode(generate_ids, skip_special_tokens=True, clean_u
...
+    peft_model,
+    mitigation_lambda=0.75,
+)
+
+peft_model.generate(...)
+```
+
+There are a few hyper-parameters that can be used for tuning the effectiveness of the mitigation but, as evidenced
+in Figure 8 of the paper, it will always be a trade-off between task accuracy learned by the adapter and forgetting
+of the base model's knowledge. For example, the stronger the `mitigation_lambda`, the stronger the reduction in
```
sayakpaul: But I find this counterintuitive. With less knowledge forgotten, I would expect to see the accuracy being preserved more and not less.

src/peft/tuners/lora/intruders.py
```diff
@@ -0,0 +1,157 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+
+from .layer import LoraLayer
+
+
+def reduce_intruder_dimension(
```
sayakpaul, line 20: Not merge blocker but do you think we introduce something like 
https://github.com/huggingface/peft/blob/4f62d7003ef7512e1c125b4680415ca57400ff20/src/peft/tuners/lora/conversion.py#L86

in here as well? If so, we will probably need to isolate the core functionality and compile it before applying it to the individual modules of type `LoraLayer`.

src/peft/tuners/lora/intruders.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # apply mitigation on the old adapter's weights and move them to the new adapter's weights
+    for layer_name, layer in peft_model.named_modules():
+        if not isinstance(layer, LoraLayer):
+            continue
+
+        W = layer.get_base_layer().weight.data
+        dW = layer.get_delta_weight(old_adapter_name)
+        W_merged = W + dW
+        is_embedding = old_adapter_name not in layer.lora_B
```
sayakpaul, line 92: Is this a sufficiently safe condition to determine this?

sayakpaul (commented):
Thank you! I have left some comments. LMK if they make sense.

docs/source/developer_guides/lora.md
githubnemo: Yes, maybe it is not clear enough. The documentation is talking about forgetting in the base model and accuracy of the adapted task. You modify the LoRA to have less influence on the base model regarding forgetting but in the process you also lose some of the information in the adapter that contributes to the task, reducing task accuracy.

Maybe adding something like

> it will always be a trade-off between task accuracy learned by the adapter and forgetting of the base model's knowledge. The mitigation will remove information from the adapter to reduce the impact on forgetting previous knowledge but also means some information about the task learned by the adapter is lost as well.

WDYT?

src/peft/tuners/lora/intruders.py
githubnemo: Good question. I think so, yes - it is the same way it is done in `lora/layer.py:reset_lora_parameters`. There is no case that I know of that uses both `lora_B` and `lora_embedding_B` so it should be fine in that regard as well.

docs/source/developer_guides/lora.md
githubnemo: Yep, changed.

src/peft/tuners/lora/intruders.py
githubnemo: Good idea. I've tested this by refactoring the whole inner loop as well as only the updating portion but both were constantly recompiling (probably due to the changed shapes / varying number of intruder dimensions). I think this should be handled in a dedicated PR with a proper benchmark setup.

**Edit**: I managed to reduce the recompilations to 1 by wrapping `tensor(intruder_idcs)` but it made no difference in runtime (at least on CPU) so what I said before still stands, let's not do this right now.

docs/source/developer_guides/lora.md
sayakpaul: Yes much better and thank you for explaining!

src/peft/tuners/lora/intruders.py
sayakpaul: Cool. Let's maybe then just add a note pointing this to the comment, then? 

Yeah benchmarking it on a CPU would probably not yield anything substantial.

githubnemo:
Added the improved text in the documentation (from [here](https://github.com/huggingface/peft/pull/2999#discussion_r2717812267)) and added a comment regarding `torch.compile` support.

The CI quality failure seems unrelated.

I'll merge this once CI is through. Thanks for the reviews :)

PR #3000, opened 2026-01-16
Add get base model state dict

### LINKED ISSUE #2945: Return base model state_dict with original keys
### Feature request

TL;DR: `from peft import get_base_model_state_dict`

Hi!

I'm looking for a way to get the state dict of the base model after it has been wrapped in a `PeftModel` while preserving the original model's state dict keys. To the best of my knowledge, the only way this can be done right now is getting the state dict from `peft_model.base_model.model` and manually patching the keys by removing the `.base_layer.` infix and filtering our peft param keys.

A reason you wouldn't want to load the base model's state dict before wrapping it, for example, is when you are loading state dicts after FSDP wrapping your peft model.

### Your contribution

I have some of this logic implemented for Torchtitan. I could repurpose some of it for a PR that handles PEFT's edge-cases a bit more gracefully (so far I've only checked my approach for LoRA).

Isalia20:
Fixes #2945

Adds `get_base_model_state_dict` / `set_base_model_state_dict`: read and write a PEFT-wrapped model's base
weights using the original, pre-PEFT key names.

Normally the checkpoint is loaded before PEFT wraps anything, so the keys always match and none of this is
needed:

```python
model = AutoModelForCausalLM.from_pretrained(...)   # weights loaded here
model = get_peft_model(model, config)               # wrapped afterwards
```

TorchTitan (#2945) has to reverse that order. Under FSDP no rank holds the whole model, so the model is built on
`meta`, sharded, and only then materialised and PEFT has to be applied before sharding, because it swaps
modules out (`nn.Linear` → `lora.Linear`), which can't be done once FSDP has flattened the parameters:

```python
with torch.device("meta"):
    model = AutoModelForCausalLM.from_pretrained(...)   # structure only, no memory
    model = get_peft_model(model, config)               # must happen pre-shard

model = fsdp_shard_model(model)
model.to_empty(device=init_device)                      # memory exists only now
set_base_model_state_dict(model, checkpoint)            # weights loaded here
```

By that point the model is already wrapped, but the checkpoint on disk still says `q_proj.weight` where the
model now wants `q_proj.base_layer.weight`, so a plain `load_state_dict` reports every model key missing and
every checkpoint key unexpected. These two functions are that translation, in both directions.

Two implementation choices:

**Renaming.** The so to say "easy" approach is a string replace of `.base_layer.` / `.original_module.`. This walks the
module tree instead, resolving each part of the key with `getattr` and checking what kind of module it landed on.
`base_layer` only means "skip this" when the module holding it is a tuner layer, in someone's own model it might
just be a module name and simply stripping it would produce a wrong key.

**Filtering.** Which parameters belong to the adapter is decided by position: on a tuner layer everything except `base_layer` is the adapter's, and for `modules_to_save` and trainable tokens the keys come from `adapter_state_dict_load_map()`, the same source `get_peft_model_state_dict` uses. Not matching on name
prefixes is what keeps VeRA and VB-LoRA correct.


Isalia20:
@BenjaminBossan Would be glad if you could review it when you get a chance

BenjaminBossan (dismissed):
Thanks for this PR. I currently don't have much time to review and will be OoO next week. so hopefully @githubnemo can take over.

Just my first observations:

1. From the original issue, I think we concluded that we would rather need a `set_base_model_state_dict`. That doesn't mean that `get_base_model_state_dict` doesn't have it's merits, but it wouldn't fully solve the issue. Ping @dvmazur.
2. There can be deeper nesting of `.base_layer`, so it should run in a loop: `while ".base_layer" in new_key: ...`.
3. This doesn't take into account trainable tokens yet, they need to be treated similarly as `modules_to_save`.

dvmazur:
Hi! Thanks for this PR! Yeah, I rather need a `set_base_model_state_dict`, but it should be pretty easy to implement once we have a `get_base_model_state_dict` I think. Also, maybe we should expand the test matrix to make sure this method works for other PEFTs?

Isalia20:
Hi, I'll add the set method as well and more tests little later today

Isalia20:
Added the set base state dict and more tests

Isalia20:
@githubnemo  Would be glad if you could review this :)

src/peft/peft_model.py
```diff
@@ -1634,6 +1634,181 @@ def supports_lora_conversion(self, adapter_name: str = "default") -> bool:
...
+
+        for peft_key in current_state_dict.keys():
+            is_adapter_param = False
+            for prefix in adapter_prefixes:
+                if f".{prefix}" in peft_key or peft_key.startswith(prefix):
+                    is_adapter_param = True
+                    break
+
+            if is_adapter_param:
+                continue
```
githubnemo: I think this is not a sufficient filter for methods like VeRA or VB-LoRA that employ weight sharing. This will be covered by the extended tests I suppose.

An alternative approach would be to iterate over all named modules of the model and remove those keys that belong to `BaseTunerLayer` instances (since the weight shared keys are caught by the prefix matching already in place). But lets see what the tests say first, maybe I'm wrong and everything works fine :)

githubnemo (commented):
Hey @Isalia20  :) Thanks for taking this on.

I need a bit of clarification (possibly from @dvmazur): do I understand correctly that one use-case is that we have a model that doesn't fit in memory so we need to first shard the empty model (via FSDP) onto several devices and then read the checkpoint onto the shards (in a streaming manner)? Furthermore, do I understand correctly that it is not possible to shard the base model and then apply PEFT on top of that? If that are the reasons for why this is useful, we should probably document that as well since it is not obvious.

The implementation seems to pass at first glance but there might be a few pitfalls still. I left one comment regarding a potential bug.

Let's build a test (e.g., a merge of `test_get_base_model_state_dict_keys_match` and `test_get_base_model_state_dict_values_match`) and integrate it into `tests/testing_common.py` (similar to `_test_save_pretrained`) to be called from the more exhaustive testing suites in `tests/test_decoder_models.py`, `tests/test_encoder_decoder_models.py` and `tests/test_custom_models.py` which cover a lot more cases. For example, trainable tokens and parameter targeting are not covered by the current tests and there are probably a lot more special cases, so leveraging the existing tests is probably best.

dvmazur:
Hi!

I want to be able to load the base model's and adapter's state_dicts after wrapping the PEFT model in FSDP. The state_dict's keys match the original base model's keys, so I need a function that will map the wrapped model's keys to the original ones if that makes sense.

Isalia20:
Thanks for the comments. I'll take a look little later this week

githubnemo:
Hey @dvmazur,

> I want to be able to load the base model's and adapter's state_dicts after wrapping the PEFT model in FSDP.

I got that but why? What's your motivation? My question supposed that memory is a constraint and that's the reason but you didn't acknowledge nor refute that. Please give a bit more detail so that I can understand the use-case better. Thanks!

dvmazur:
The end goal is to have PEFT working for TorchTitan basically. Titan wraps models into FSDP to save VRAM, it also allocates GPU memory only after the model's meta-device weights were FSDP-sharded.

I think this pseudocode snippet should give you enough info, but feel free to ask if you need any more info:

```python
with torch.device("meta"):
    # can't load base model weights here as it is on meta device before resharding
    # can only wrap model in peft before fsdp-sharding it
    model = get_perft_model(model, ...)

model = fsdp_shard_model(model)

# actually allocate memory for the model's weights
# state dict can be loaded after that
model.to_empty(device=init_device)

# this function loads a state dict with the original model's module keys
# so I need a way to map them to the PEFT-wrapped model
load_base_model_state_dict(model)
initialize_adapters(model)
```

BenjaminBossan:
_not stale_

githubnemo:
@Isalia20 gentle ping :)

Isalia20:
oh sorry my bad, forgot about this PR. Will try to resolve comments this week

Isalia20:
updated

Isalia20:
@githubnemo any updates on this?

tests/test_get_base_model_state_dict.py
```diff
@@ -0,0 +1,349 @@
+# Copyright 2025-present the HuggingFace Inc. team.
```
githubnemo:
```suggestion
# Copyright 2026-present the HuggingFace Inc. team.
```

src/peft/peft_model.py
githubnemo, line 1: Let's move the changes from `peft_model.py` to `utils/save_and_load.py`

src/peft/peft_model.py
```diff
@@ -1641,6 +1641,166 @@ def supports_lora_conversion(self, adapter_name: str = "default") -> bool:
...
+        module tree to identify adapter-specific parameters. This is needed for weight sharing type of adpters (e.g.,
+        VeRA, VB-LoRA)
+        """
+        adapter_key_prefixes: set[str] = set()
+        for module_name, module in self.base_model.model.named_modules():
+            if isinstance(module, (BaseTunerLayer, AuxiliaryTrainingWrapper)):
+                for attr_name in module.adapter_layer_names + module.other_param_names:
+                    prefix = f"{module_name}.{attr_name}" if module_name else attr_name
+                    adapter_key_prefixes.add(prefix)
+        return adapter_key_prefixes
```
githubnemo: this doesn't seem to respect `module.adapter_state_dict`. See `utils/save_and_load.py:get_peft_model_state_dict` which queries the auxiliary training wrapper's state dict to get the keys that are touched by the aux. training wrapper.

tests/test_get_base_model_state_dict.py
```diff
@@ -0,0 +1,349 @@
...
+
+
+MODEL_ID = "peft-internal-testing/tiny-random-OPTForCausalLM"
+
+
+def get_peft_configs():
+    """Return a list of (config_name, config) tuples for parametrized testing"""
+    return [
+        ("lora", LoraConfig(r=4, lora_alpha=2, target_modules=["q_proj", "v_proj"])),
+        ("lora_all_linear", LoraConfig(r=4, lora_alpha=2, target_modules="all-linear")),
```
githubnemo:
```suggestion
        ("lora_all_linear", LoraConfig(r=4, lora_alpha=2, target_modules="all-linear")),
        ("lora_trainable_tokens", LoraConfig(r=4, trainable_token_indices=[0, 1], target_modules=["q_proj", "v_proj"])),
        ("trainable_tokens", TrainableTokensConfig(token_indices=[0, 1])),
```

This will highlight why it's necessary to inspect the state dict of the aux. training wrappers.

tests/test_get_base_model_state_dict.py
```diff
@@ -0,0 +1,349 @@
...
+    VBLoRAConfig,
+    VeraConfig,
+    get_peft_model,
+)
+from peft.utils import infer_device
+
+from .testing_utils import hub_online_once
+
+
+MODEL_ID = "peft-internal-testing/tiny-random-OPTForCausalLM"
```
githubnemo: Let's wrap the tests in a class and parametrize the model id. I think it's a good idea to test at least an encoder-decoder model like T5 as well.

tests/test_get_base_model_state_dict.py
```diff
@@ -0,0 +1,349 @@
...
+    return model
+
+
+@pytest.mark.parametrize("config_name,peft_config", get_peft_configs(), ids=[c[0] for c in get_peft_configs()])
+def test_get_base_model_state_dict_keys_match(base_model, config_name, peft_config):
+    """Test that get_base_model_state_dict returns keys matching the original base model."""
+    base_model_keys = set(base_model.state_dict().keys())
+    peft_model = get_peft_model(base_model, peft_config)
+    extracted_keys = set(peft_model.get_base_model_state_dict().keys())
+    assert base_model_keys == extracted_keys, f"Key mismatch for {config_name}"
```
githubnemo: Let's have this test in `testing_common.py` and call it in `test_encoder_decoder_models.py`, `test_decoder_models.py` and `test_custom_models.py`.

src/peft/peft_model.py
```diff
@@ -1641,6 +1641,166 @@ def supports_lora_conversion(self, adapter_name: str = "default") -> bool:
...
+
+            >>> base_model = AutoModelForCausalLM.from_pretrained("gpt2")
+            >>> base_weights = base_model.state_dict()
+            >>> peft_model = get_peft_model(base_model, LoraConfig(target_modules=["c_attn"]))
+
+            >>> # Restore original base model weights on the peft wrapped model
+            >>> result = peft_model.set_base_model_state_dict(base_weights)
+            ```
+        """
+        from collections import namedtuple
```
githubnemo: Let's make this an top-level import

src/peft/peft_model.py
```diff
@@ -1641,6 +1641,166 @@ def supports_lora_conversion(self, adapter_name: str = "default") -> bool:
...
+
+        _IncompatibleKeys = namedtuple("IncompatibleKeys", ["missing_keys", "unexpected_keys"])
+
+        if self._is_prompt_learning:
+            return self.base_model.load_state_dict(state_dict, strict=strict)
```
githubnemo: Let's add an comment that justifies the early return here ("why is it ok?")

githubnemo (commented):
This looks nice! I've made some suggestions / comments, most importantly about the missing aux. training wrapper handling but otherwise this is looking quite good.

One organisational thing: It would be nice of you to update the PR description with what this change entails and what the motivation was (i.e. support for TorchTitan). The description will be the commit message of the PR so it would be very helpful for future developers to know the context this change was made for. Precision over verbosity.

Isalia20:
Updated

Isalia20:
@githubnemo gentle ping

src/peft/peft_model.py
githubnemo: I'm not sure there's precedent for this in `PeftModel`. Unless I'm mistaken, let's remove these calls since the current pattern is to import from `save_and_load` utils.

src/peft/utils/save_and_load.py
```diff
@@ -760,3 +760,172 @@ def get_hub_filename(use_safetensors=True):
...
+    original_parts: list[str] = []
+    current = model.base_model.model
+
+    for part in parts:
+        if current is None:
+            # Already past the module tree (e.g. nested parameter names)
+            original_parts.append(part)
+            continue
+
+        child = getattr(current, part, None)
```
githubnemo, line 1177: can you motivate (in a code comment) why we're traversing the model this way and not using `model.get_named_modules()` instead? This seems like an important decision so it would be unwise to not document it.

githubnemo (commented):
Thanks for the update. Some of the tests fail in CI, can you take a look?

Isalia20:
not stale

BenjaminBossan:
@Isalia20 Do you still think you'll have time to work on this?

Isalia20:
I'll work on it, will DM you with a few questions this week

Isalia20:
Sorry for delay, updated and addressed comments

githubnemo (commented):
Thanks for the updates!

Can you update the PR description to be meaningful (what has changed, why was it changed, what were the options, why did we decide to go with the one that is implemented)?

One question: do you think we can remove `test_get_base_model_state_dict.py`? It seems to me that the test in `testing_common.py` already covers quite a lot and I'm not sure which tests are worth the added test runtime. WDYT?

Isalia20:
@githubnemo updated PR description and the test. I've made the test file shorter. `_test_get_base_model_state_dict` only exercised the getter, so I added the roundtrip there and deleted the standalone tests it made redundant.

That leaves three tests the roundtrip can't cover: two for the `strict` error handling (the roundtrip only ever
passes valid state dicts) and one that adds a second adapter (the roundtrip only ever has one)

tests/test_get_base_model_state_dict.py
githubnemo, line 1: I think this fits semantically in `test_low_level_api.py`, let's move the content there.

githubnemo (commented):
LGTM feature-wise, thanks!

I think it makes sense to document the new functionality in the docs since they are part of the public API in `docs/source/developer_guides/low_level_api.md` (e.g., `## Setting and loading base weights`?).

githubnemo (approved):
Thanks! LGTM :)

PR #2856, opened 2025-10-21
CI: Add FSDP tests on multi GPU machine

BenjaminBossan:
Extend nightly GPU tests to include explicit tests for:

- DeepSpeed (no quant, 4bit & 8bit bnb)
- FSDP (no quant, 4bit bnb, 8bit not supported)

PR #3001, opened 2026-01-19
Fix error of PEFT with disable adapters and FSDP

### LINKED ISSUE #2977: Unable to use enable/disable_adapter_layers in FSDP training
I am training Mistral where I am not sharding lm_head, so when the model is wrapped in fsdp i want to change the adapter like this 

        self.model.icae.set_adapter("encadapt")
        self.model.icae.enable_adapter_layers()
        for name, param in self.model.icae.named_parameters():
            if "encadapt" in name:
                param.requires_grad = False
        compress_outputs = self.model.icae(inputs_embeds=autoencoder_input_embedding, output_hidden_states=True,
                                           graph=graph, mem_mask=mem_mask, partial_grad=partial_grad, map_node=True)
        self.model.icae.disable_adapter_layers()

Doing so gives me the error

`
RuntimeError: you can only change requires_grad flags of leaf variables. If you want to use a computed variable in a subgraph that doesn't require differentiation use var_no_grad = var.detach().
`
So the question is how do i use enable/disable_adapter_layers in FSDP training if someone has to switch the adapter during training due to some reason.

Isalia20:
Fixes #2977

For tests, I'm not sure if PEFT repo has multiple gpu nodes in CI for testing this. By searching through tests, I couldn't find distributed tests. Would be glad to add it if you point me to where it is/example test

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -1485,7 +1485,7 @@ class BaseTunerLayer(ABC):
             module_dict = getattr(self, layer_name)
             for key, layer in module_dict.items():
                 if key in adapter_names_set:
-                    layer.requires_grad_(requires_grad)
+                    _set_layer_requires_grad(layer, requires_grad)
 
     def _get_base_layer_device_and_dtype(self, base_layer):
         """
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -618,7 +618,7 @@ class ModulesToSaveWrapper(AuxiliaryTrainingWrapper):
         # since there would be no clear way to decide which adapter's weights are the correct ones. therefore we
         # assume that there is only one active adapter. this precondition is enforced by _set_adapter.
         if adapter_name == self.active_adapter:
-            self.modules_to_save[adapter_name].requires_grad_(True)
+            _set_layer_requires_grad(self.modules_to_save[adapter_name], True)
 
     def enable_adapters(self, enabled: bool):
         """Takes care of setting the required_grad flag on the wrapped module.
@@ -628,9 +628,10 @@ class ModulesToSaveWrapper(AuxiliaryTrainingWrapper):
 
         if enabled:
             for adapter_name in self.active_adapters:
-                self.modules_to_save[adapter_name].requires_grad_(True)
+                _set_layer_requires_grad(self.modules_to_save[adapter_name], True)
         else:
-            self.modules_to_save.requires_grad_(False)
+            for module in self.modules_to_save.values():
+                _set_layer_requires_grad(module, False)
 
     def check_set_adapter(self, adapter_name: str | list[str]) -> str | None:
         """Helper function to check if the given adapter(s) can be set.
@@ -676,7 +677,7 @@ class ModulesToSaveWrapper(AuxiliaryTrainingWrapper):
             raise ValueError(f"Attempted to set multiple ({adapter_names}) adapters at once for modules_to_save.")
 
         for currently_active_adapter_name in self.active_adapters:
-            self.modules_to_save[currently_active_adapter_name].requires_grad_(False)
+            _set_layer_requires_grad(self.modules_to_save[currently_active_adapter_name], False)
 
         if len(adapter_names) == 0:
             # when calling model.add_adapter, the new adapter is not automatically active
@@ -688,7 +689,7 @@ class ModulesToSaveWrapper(AuxiliaryTrainingWrapper):
         if adapter_name not in self._adapters:
             raise ValueError(f"Adapter {adapter_name} not found in {self._adapters}")
 
-        self.modules_to_save[adapter_name].requires_grad_(not inference_mode)
+        _set_layer_requires_grad(self.modules_to_save[adapter_name], not inference_mode)
         self._active_adapter = adapter_name
 
     def delete_adapter(self, adapter_name: str, new_active_adapters: Optional[list[str]]) -> None:
--- a/tests/training/adapters.py
+++ b/tests/training/adapters.py
@@ -45,7 +45,7 @@ def get_base_model_weights(peft_model):
     """Extract base model weights (non-LoRA weights)."""
     base_weights = {}
     for name, param in peft_model.named_parameters():
-        if "lora" not in name.lower():
+        if "lora" not in name.lower() and "modules_to_save" not in name:
             base_weights[name] = param.detach().clone()
     return base_weights
 
@@ -83,6 +83,7 @@ class Model(nn.Module):
             r=16,
             lora_alpha=32,
             target_modules=["q_proj", "v_proj"],
+            modules_to_save=["lm_head"],
             lora_dropout=0.05,
             bias="none",
             task_type="CAUSAL_LM",
@@ -94,6 +95,7 @@ class Model(nn.Module):
             r=8,
             lora_alpha=16,
             target_modules=["q_proj", "v_proj"],
+            modules_to_save=["lm_head"],
             lora_dropout=0.05,
             bias="none",
             task_type="CAUSAL_LM",
@@ -103,8 +105,8 @@ class Model(nn.Module):
         self.peft_model.set_adapter("default")
         self.peft_model.to(torch.bfloat16)
 
-        for name, param in self.peft_model.named_parameters():
-            param.requires_grad = "lora_" in name.lower() and "second_adapter" not in name
+        self.peft_model.set_requires_grad("default", requires_grad=True)
+        self.peft_model.set_requires_grad("second_adapter", requires_grad=False)
 
     def forward(self, input_ids=None, attention_mask=None, labels=None):
         out1 = self.peft_model(input_ids=input_ids, attention_mask=attention_mask, labels=labels)


Isalia20:
@githubnemo Tagging for review

varunlmxd:
I tried this got this error
`AssertionError: Invalid: ``_post_backward_hook_state``: (<AccumulateGrad object at 0x7ff3b925a230>, <torch.utils.hooks.RemovableHandle object at 0x7ff3b925a200>)`

Isalia20:
I think that is a different not related to this one error. I had a similar error in my training which no longer occurs after this change

githubnemo:
> For tests, I'm not sure if PEFT repo has multiple gpu nodes in CI for testing this. By searching through tests, I couldn't find distributed tests. Would be glad to add it if you point me to where it is/example test

The merged PR #2856 includes deepspeed/fsdp for multi GPU, maybe that helps already?

Isalia20:
@githubnemo I think that should have fixed the bugs, can you run the workflow/review when you get a chance?

tests/training/test_fsdp_adapters.py
```diff
@@ -0,0 +1,256 @@
...
+        # Test switching back
+        print_if_process_zero("Testing set_adapter('adapter1')...")
+        model.set_adapter("adapter1")
+        print_if_process_zero("set_adapter('adapter1') succeeded!")
+
+        # Test with list of adapters
+        print_if_process_zero("Testing set_adapter(['adapter1', 'adapter2'])...")
+        model.set_adapter(["adapter1", "adapter2"])
+        print_if_process_zero("set_adapter(['adapter1', 'adapter2']) succeeded!")
+
```
githubnemo: Let's also test if the base weights and the adapter2 weights are untouched and only adapter1 weights have changed.

src/peft/tuners/tuners_utils.py
```diff
@@ -1389,7 +1389,18 @@ def enable_adapters(self, enabled: bool) -> None:
             # disable grads on all adapter layers
             for layer_name in self.adapter_layer_names:
                 layer = getattr(self, layer_name)
-                layer.requires_grad_(False)
+                # Handle FSDP case where params may be non-leaf tensors
```
githubnemo: Maybe something like this to explain why they are non-leaf tensors?

```suggestion
                # Handle FSDP case where params may be non-leaf tensors by being wrapped in DTensors
```

src/peft/tuners/tuners_utils.py
```diff
@@ -1411,12 +1422,18 @@ def set_adapter(self, adapter_names: str | list[str], inference_mode: bool = Fal
         for layer_name in self.adapter_layer_names:
             module_dict = getattr(self, layer_name)
             for key, layer in module_dict.items():
-                if (key in adapter_names) and (not inference_mode):
-                    # Note: It is possible that not a single layer is called with requires_grad_(True) here. This may
-                    # happen if a completely different adapter layer is being activated.
```
githubnemo, line 1416: I think the note comment from before can stay as it doesn't seem to be invalidated.

tests/training/test_fsdp_adapters.py
githubnemo, line 1: Let's rename this to `adapters.py` to avoid being gobbled up by pytest and dropping the `fsdp` prefix since this technically is a test for whatever distributed training we use

tests/training/test_fsdp_adapters.py
```diff
@@ -0,0 +1,256 @@
...
+        # Test set_adapter - should not raise
+        print_if_process_zero("Testing set_adapter('adapter2')...")
+        model.set_adapter("adapter2")
+        print_if_process_zero("set_adapter('adapter2') succeeded!")
+
+        # Test switching back
+        print_if_process_zero("Testing set_adapter('adapter1')...")
+        model.set_adapter("adapter1")
+        print_if_process_zero("set_adapter('adapter1') succeeded!")
```
githubnemo: I wonder if the test should mimic a more realistic training scenario, i.e. switching to adapter 1, training a bit, switching to adapter 2, training that.

Isalia20:
Hmm, I think error in check_code_quality workflow is not related to this pr

Isalia20:
> Thanks for the updates.
> 
> I've ran the test on a multi-gpu machine and got the following error:
> 
> ```
> Testing disable_adapters()...
> [rank1]: Traceback (most recent call last):
> [rank1]:   File "/home/ubuntu/code/peft/githubnemo/tests/training/test_fsdp_adapters.py", line 256, in <module>
> [rank1]:     main(test_name=args.test, model_id=args.model_id, quant=args.quant)
> [rank1]:   File "/home/ubuntu/code/peft/githubnemo/tests/training/test_fsdp_adapters.py", line 237, in main
> [rank1]:     test_disable_adapters(model_id, quant)
> [rank1]:   File "/home/ubuntu/code/peft/githubnemo/tests/training/test_fsdp_adapters.py", line 115, in test_disable_adapters
> [rank1]:     model.disable_adapters()
> [rank1]:   File "/home/ubuntu/envs/peft-githubnemo/lib/python3.12/site-packages/transformers/integrations/peft.py", line 724, in disable_adapters
> [rank1]:     raise ValueError("No adapter loaded. Please load an adapter first.")
> [rank1]: ValueError: No adapter loaded. Please load an adapter first.
> [rank0]: Traceback (most recent call last):
> [rank0]:   File "/home/ubuntu/code/peft/githubnemo/tests/training/test_fsdp_adapters.py", line 256, in <module>
> [rank0]:     main(test_name=args.test, model_id=args.model_id, quant=args.quant)
> [rank0]:   File "/home/ubuntu/code/peft/githubnemo/tests/training/test_fsdp_adapters.py", line 237, in main
> [rank0]:     test_disable_adapters(model_id, quant)
> [rank0]:   File "/home/ubuntu/code/peft/githubnemo/tests/training/test_fsdp_adapters.py", line 115, in test_disable_adapters
> [rank0]:     model.disable_adapters()
> [rank0]:   File "/home/ubuntu/envs/peft-githubnemo/lib/python3.12/site-packages/transformers/integrations/peft.py", line 724, in disable_adapters
> [rank0]:     raise ValueError("No adapter loaded. Please load an adapter first.")
> [rank0]: ValueError: No adapter loaded. Please load an adapter first.
> ```
> 
> I also left some comments.
> 
> > Hmm, I think error in check_code_quality workflow is not related to this pr
> 
> Yep, you can ignore that for now.

Thanks for running this. Will take a look comments in the upcoming days and fix it. And probably get a multi gpu machine to test this out properly

Isalia20:
Updated the tests and resolved comments and tested on multi gpu device. Should be ready now

tests/training/adapters.py
```diff
@@ -0,0 +1,165 @@
...
+            final_weights[name],
+        )
+
+
+class Model(nn.Module):
+    def __init__(self, model_id):
+        super().__init__()
+        model = AutoModelForCausalLM.from_pretrained(
+            model_id,
+            torch_dtype=torch.bfloat16,
```
githubnemo:
```suggestion
            dtype=torch.bfloat16,
```

Isalia20:
Hmm, I think test errors are not related to this PR 🤔 @githubnemo wdyt?

githubnemo:
@Isalia20 yep, they're not related. If you merge with main this issue will disappear.

tests/training/adapters.py
BenjaminBossan:
```suggestion
            dtype=torch.bfloat16,
```

tests/training/adapters.py
BenjaminBossan, line 1: Let's fix the random seed, just to be sure.

tests/training/adapters.py
```diff
@@ -0,0 +1,165 @@
...
+                output_dir=tmp_dir,
+            ),
+            data_collator=DataCollatorForLanguageModeling(model.tokenizer, mlm=False),
+        )
+        trainer.train()
+    with FSDP.summon_full_params(trainer.model):
+        final_base_weights = get_base_model_weights(model.peft_model)
+        final_second_adapter_weights = get_adapter_weights(model.peft_model, "second_adapter")
+
+    verify_weights_unchanged(initial_base_weights, final_base_weights, "Base model")
```
BenjaminBossan, line 158: Could you explain why we check if the base weights changed? Is this something we would ever expect?

BenjaminBossan (commented):
Hi @Isalia20, thanks for tackling this issue and providing a solution and a test. I ran the test on my machine and can confirm that it works. Besides what @githubnemo, could you please check my comments?

tests/training/adapters.py
Isalia20: I set the seed to:
torch.manual_seed(42)
on line 118 already, any other ways I can set it?

tests/training/adapters.py
Isalia20: updated

tests/training/adapters.py
Isalia20: We wouldn't expect the base weights to change. Test is to make sure that through this FSDP setup the base weights remain unchanged. It was initially requested by @githubnemo and I also think it's a good thing to check in this test

tests/training/adapters.py
BenjaminBossan: I must have missed that, thanks.

tests/training/adapters.py
BenjaminBossan: Ah I see. Let's add a comment that this is a sanity check.

Isalia20:
Oh my bad, I thought pre commit did that. Let me run it and update

Isalia20:
Updated

Isalia20:
@BenjaminBossan Yeah makes sense, updated the PR with it

src/peft/utils/other.py
```diff
@@ -506,7 +525,7 @@ def set_requires_grad(self, adapter_names: str | Sequence[str], requires_grad: b
             module_dict = attrgetter(layer_name)(self)
             for key, layer in module_dict.items():
                 if key in adapter_names_set:
-                    layer.requires_grad_(requires_grad)
+                    _set_layer_requires_grad(layer, requires_grad)
```
BenjaminBossan, line 528: Thanks for adding the fix to `AuxiliaryTrainingWrapper`. Could you please extend the test to also use `modules_to_save` so that it covers this fix?

Isalia20: Now the modules to save wrapper should be using the helper as well, also it gets tested in the test

Isalia20:
@BenjaminBossan any other comments on the PR or can we merge?

PR #3006, opened 2026-01-22
Fix initialization bug introduced in #2960

githubnemo:
The quantized adalora SVD layers were not initialized in the new config-passing scheme and therefore raised errors in the GPU tests.

For reproduction run

```
make tests_examples_single_gpu
```

which will yield

```
FAILED tests/test_gpu_examples.py::PeftBnbGPUExampleTests::test_4bit_adalora_causalLM - TypeError: SVDLinear4bit.__init__() missing 1 required positional argument: 'config'
FAILED tests/test_gpu_examples.py::PeftBnbGPUExampleTests::test_8bit_adalora_causalLM - TypeError: SVDLinear8bitLt.__init__() missing 1 required positional argument: 'config'
```

Note that it seems that with torch 2.10 the error message for torchao linear 4 bit has changed from a `RuntimeError` to a `ValueError` but it is still the case that 4 bit is not supported, therefore the solution is to expect both exception types.

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -4133,7 +4133,7 @@ class PeftTorchaoGPUTests(unittest.TestCase):
     @pytest.mark.single_gpu_tests
     @pytest.mark.xfail(
         reason="int4_weight_only still has issues",
-        raises=RuntimeError,
+        raises=(RuntimeError, ValueError),
     )
     def test_causal_lm_training_single_gpu_torchao_int4_raises(self):
         # TODO: Once proper torchao support for int4 is added, remove this test and add int4 to supported_quant_types


src/peft/tuners/adalora/model.py
```diff
@@ -181,7 +181,7 @@ def _create_new_module(lora_config, adapter_name, target, device_map=None, **kwa
                     "index": target_base_layer.index,
                 }
             )
-            new_module = SVDLinear8bitLt(target, adapter_name, **kwargs)
+            new_module = SVDLinear8bitLt(target, adapter_name, config=lora_config, **kwargs)
```
ydshieh, line 184: Is `SVDLinear8bitLt` defined in `peft`? If so, Just make sure we handle the deprecation etc.

src/peft/tuners/adalora/bnb.py
```diff
@@ -97,7 +98,7 @@ def __init__(
             self.get_base_layer().weight.requires_grad = False
 
             self._active_adapter = adapter_name
-            self.update_layer(adapter_name, r, config=config)
+            self.update_layer(adapter_name, r, lora_alpha, config=config)
```
ydshieh, line 101: not so sure the reason here, especially it passes a new positional argument. It won't break something?

githubnemo: This is just so that it is compatible with all the other `update_layer` definitions - the parameter was simply forgotten in PR #2960. Adding `lora_alpha` restores compatibility.

src/peft/tuners/adalora/model.py
githubnemo: Yes, indeed! However, we consider this not part of the public API, otherwise #2960 would have needed to do that in the first place, this is just a patch to fix places where the change from #2960 was forgotten.

PR #3007, opened 2026-01-22
[LoRA] Document support for effective rank for LoRA on MOE experts

kashif:
Add docs on how to set the rank of LoRA adaptors to the experts in MOEs to be the rank / total experts

### DIFF
--- a/docs/source/package_reference/lora.md
+++ b/docs/source/package_reference/lora.md
@@ -89,3 +89,8 @@ The abstract from the paper is:
 [[autodoc]] tuners.lora.loraga.estimate_gradients
 
 [[autodoc]] tuners.lora.loraga.preprocess_loraga
+
+
+## Intruder Dimension Reduction
+
+[[autodoc]] tuners.lora.intruders.reduce_intruder_dimension
--- a/docs/source/developer_guides/lora.md
+++ b/docs/source/developer_guides/lora.md
@@ -482,7 +482,12 @@ trainer = Trainer(
 ```
 
 
-## Merge LoRA weights into the base model
+## Post-Training
+
+This section shows potential post-processing methods for trained adapters.
+
+
+### Merge LoRA weights into the base model
 
 While LoRA is significantly smaller and faster to train, you may encounter latency issues during inference due to separately loading the base model and the LoRA adapter. To eliminate latency, use the [`~LoraModel.merge_and_unload`] function to merge the adapter weights with the base model. This allows you to use the newly merged model as a standalone model. The [`~LoraModel.merge_and_unload`] function doesn't keep the adapter weights in memory.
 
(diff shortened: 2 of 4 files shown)


src/peft/tuners/lora/layer.py
```diff
@@ -971,6 +971,7 @@ def __init__(
             r,
             lora_alpha=lora_alpha,
             config=config,
+            target_name=self.kwargs.get("target_name", ""),
```
sayakpaul: Or `target_parameters`?

src/peft/tuners/lora/layer.py
```diff
@@ -2074,6 +2077,13 @@ def update_layer(
         if lora_variant is not None:
             raise ValueError(f"lora.{self.__class__.__name__} does not work with LoRA variants like DoRA.")
 
+        if (
+            config.moe_rank_normalization
+            and self.num_experts > 1
+            and "experts" in target_name.split(".")
+        ):
```
sayakpaul: Add a few validations?

If `moe_rank_normalization` is specified and there's no expert, we error out, etc.?

tests/test_target_parameters.py
```diff
@@ -163,6 +164,35 @@ def from_pretrained(cls, *args, **kwargs):
...
+            for module in model.modules()
+            if isinstance(module, ParamWrapper) and module.parameter_name == "gate_up_proj"
+        ]
+        assert wrappers, "Expected to find ParamWrapper for gate_up_proj."
+        lora_module = wrappers[0]
+        effective_r = max(1, r // num_experts)
+        assert lora_module.r["default"] == effective_r
+        assert lora_module.lora_A["default"].weight.shape[0] == effective_r * num_experts
+        assert lora_module.scaling["default"] == config.lora_alpha / effective_r
+
```
sayakpaul, line 198: Maybe also test if the LoRA is exported correctly as well?

src/peft/tuners/lora/layer.py
kashif: so we are targeting a single resolve module / parameter path here for the current layer and so `target_parameters` is the config list of pateerns rater than the single target name, shall i rename it to `target_path` ?

src/peft/tuners/lora/layer.py
kashif: so if `moe_rank_normalization=True` and target isn’t an expert param, the condition fails and `r` stays unchanged. Should i add a warning if there is no expert?

sayakpaul: Yes.

src/peft/tuners/lora/layer.py
sayakpaul: Okay but how is `target_name` passed by the caller? I guess I didn't see any code for that. Am I missing out on something?

tests/test_target_parameters.py
kashif: added check

kashif:
> Thanks! Left some comments. Would this apply to other variants of LoRA we support in PEFT as well?

so currently it only for the standard LoRA on the `target_parameters`

src/peft/tuners/lora/layer.py
kashif: ah! so that is a local variable i needed to pass it to the `update_layer` method to then conditionally set the `r`

docs/source/package_reference/lora.md
```diff
@@ -30,6 +30,43 @@ The abstract from the paper is:
 
 [[autodoc]] tuners.lora.model.LoraModel
 
+## MoE expert parameters and vLLM
```
githubnemo: These kind of usage patterns are currently mainly documented in the LoRA developer guide. Let's move it there, possibly in the training section?

tests/test_target_parameters.py
```diff
@@ -163,6 +165,56 @@ def from_pretrained(cls, *args, **kwargs):
...
+        assert lora_a_keys, "Expected LoRA A weights for experts in exported adapter."
+        assert lora_b_keys, "Expected LoRA B weights for experts in exported adapter."
+        base_layers = model.base_model.model.model.layers
+        gate_up_proj = base_layers[0].feed_forward.experts.base_layer.gate_up_proj
+        in_features = gate_up_proj.shape[1]
+        out_features = gate_up_proj.shape[2]
+        expected_a_shape = (effective_r * num_experts, in_features)
+        expected_b_shape = (out_features, effective_r * num_experts)
+        assert all(state_dict[k].shape == expected_a_shape for k in lora_a_keys)
+        assert all(state_dict[k].shape == expected_b_shape for k in lora_b_keys)
```
githubnemo: I like the approach of testing examples from the docs but I'm not sure if the second half of the test is illustrative or testing something that wasn't tested before. Opting to remove.

```suggestion
```

docs/source/package_reference/lora.md
```diff
@@ -30,6 +30,43 @@ The abstract from the paper is:
...
+This keeps the total LoRA parameter budget similar to dense layers (see
+[LoRA Without Regret](https://thinkingmachines.ai/blog/lora/) by Schulman et. al.).
+Non-expert modules use the default rank `r`.
+
+[vLLM](https://vllm.ai/) (v0.11.2+) supports LoRA on fused MoE expert layers. The fused MoE kernels expect LoRA weights in expert-major
+stacked form; for gate/up projections the kernel uses two slices (gate and up), while the down projection uses one slice.
+PEFT’s expert LoRA (via `target_parameters`) is compatible with this layout when exporting adapters.
+When expert weights are fused (e.g., gate+up), LoRA A weights are concatenated along the rank dimension and LoRA B
+weights are combined in a block-diagonal fashion (per expert for 3D tensors). This preserves the original per-projection
+updates while targeting a single fused parameter tensor.
```
githubnemo: [commenting the vLLM paragraph]

It is not clear to me what the message to the user is. If it is about inference via vLLM and compatibility I think it is sufficient to state something like

> Accelerated inference with the fine-tuned model is possible with, for example, vLLM which supports fused MoE expert layers since v0.11.2.

Or something similar. WDYT?

PR #3010, opened 2026-01-27
Fix fbgemm exception in nightly CI

githubnemo:
Newer versions of fbgemm are only tested on CUDA >=12.6 and apparently 12.4 breaks with a missing symbol error. Since we're already at 12.9 using 12.8 is probably a safe bet.

EETQ subsequently fails to build because of a known issue that needs a patch in the source files (issue: https://github.com/NetEase-FuXi/EETQ/issues/31). I created a PR https://github.com/NetEase-FuXi/EETQ/pull/41) which was subsequently merged and fixed the build with CUDA 12.8.


githubnemo:
Thanks for the review!

I've updated the title and the description to match the current state.

PR #3013, opened 2026-01-30
fix: eliminate cross-terms bug when combining adapters with different weights

### LINKED ISSUE #3004: Combining adapters linearly with negative weights is broken
### System Info

Hi, 

I've noticed a bug when combining adapters with negative weights using `_generalized_task_arithmetic_weighted_adapter`. Negative weights are indeed handled ([here](https://github.com/huggingface/peft/blob/v0.18.0/src/peft/tuners/lora/model.py#L783)) but then later (eg. in [`task_arithmetic`](https://github.com/huggingface/peft/blob/main/src/peft/utils/merge_utils.py#L144) and in other methods) they are applied to both the A and B LoRA deltas. If using a single adapter, this cancels out the negative weight; with multiple adapters, we can get unexpected behavior due to cross-terms of the (A_1+A_2)*(B_1+B_2) multiplication. 

### Who can help?

_No response_

### Reproduction

I don't have a code snippet I can share, but calling `add_weighted_adapter` with a single adapter will have the same result whether using weights `[1.0]` or '[-1.0]`. 

### Expected behavior

Negative weights handled correctly - eg. sign only applied to A weights, or  adapter negated upfront and then linear combination only happens with positive weights.

wingding12:
## Summary

Fixes #3004

When using `add_weighted_adapter` with `combination_type` in `['linear', 'ties', 'dare_linear', 'dare_ties', 'magnitude_prune']`, the previous implementation combined A matrices and B matrices separately before multiplying. This introduced **cross-terms** that corrupted the result when combining multiple adapters with different weights.

### The Bug

For example, with two adapters (A1, B1) and (A2, B2) and weights [1, -1]:

- **Expected**: `A1@B1 - A2@B2`
- **Old (buggy)**: `(A1-A2)@(B1+B2) = A1@B1 + A1@B2 - A2@B1 - A2@B2`

The cross-terms `A1@B2` and `A2@B1` were incorrectly added, causing significant errors (in my tests, ~1.55 error norm vs ~1.55 expected norm - essentially completely wrong).

### The Fix

This PR computes the full delta weight (`A@B`) for each adapter first, then combines them, and finally decomposes the result back to A and B using SVD. This ensures mathematically correct results.

### Note on Approximation

The decomposition uses truncated SVD with the same rank as the input adapters, which introduces some approximation error. This is the **same behavior** as the existing SVD combination types (`svd`, `ties_svd`, etc.) and is expected.

In my tests:
- **Old buggy error**: ~1.55 (completely wrong)
- **New fixed error (rank=4)**: ~0.32 (expected rank truncation, same as `combination_type='svd'`)
- **New fixed error (rank=8)**: ~0.00001 (near exact with sufficient rank)

Users who need higher accuracy can use `combination_type='svd'` with a higher `svd_rank` parameter.

## Test Plan

I tested manually with:
1. Single adapter with negative weight (-1.0) - works correctly
2. Two identical adapters with [1.0, -1.0] - cancels to zero correctly
3. Two different adapters with [1.0, -1.0] - now matches expected result (within SVD truncation error)
4. Three adapters with mixed weights [2.0, -1.0, 0.5] - works correctly
5. All combination types (linear, ties, dare_linear, dare_ties, magnitude_prune) - all work

Also includes Conv2d layer handling (following the same pattern as the existing SVD implementation).

### DIFF
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -777,48 +777,79 @@ class LoraModel(BaseTuner):
         density,
         majority_sign_method,
     ):
-        # account weights for LoRA A and B layers.
-        valid_weights_A = []
-        valid_weights_B = []
-        lora_A_deltas = []
-        lora_B_deltas = []
+        # Collect valid adapters and their weights with scaling
+        valid_adapters = []
+        valid_weights = []
+        is_embedding = any(adapter in target.lora_embedding_A for adapter in adapters)
         for adapter, weight in zip(adapters, weights):
-            if adapter in target.lora_A:
-                current_adapter_lora_A = target.lora_A[adapter].weight
-                current_adapter_lora_B = target.lora_B[adapter].weight
-            elif adapter in target.lora_embedding_A:
-                current_adapter_lora_A = target.lora_embedding_A[adapter]
-                current_adapter_lora_B = target.lora_embedding_B[adapter]
+            if adapter in target.lora_A or adapter in target.lora_embedding_A:
+                valid_adapters.append(adapter)
+                valid_weights.append(weight * target.scaling[adapter])
+
+        if len(valid_adapters) == 0:
+            raise ValueError("No matching LoRAs found. Please raise an issue on GitHub.")
+
+        # Get the dtype and shape info from the first adapter
+        if valid_adapters[0] in target.lora_A:
+            dtype = target.lora_A[valid_adapters[0]].weight.dtype
+            lora_A_shape = target.lora_A[valid_adapters[0]].weight.shape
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Thanks a lot for this PR to fix the incorrect merging behavior. It would be crucial to add tests for this. Do you still have your testing code available? Let's add it in the form of unit tests. If you need assistance with that, please let me know.

BenjaminBossan:
ping @wingding12

BenjaminBossan:
Not getting any reply from the author. However, I think this approach is generally flawed. It is true that we have issues with cross terms, but that's why there is the `"svd"` option. The approach in this PR is to always apply `"svd"`, which defeats the purpose of having SVD and non-SVD option in the first place.

PR #2932, opened 2025-12-01
Fully deprecate AutoGPTQ and AutoAWQ for GPT-QModel

ZX-ModelCloud:
Remove autogptq clutter and autogptq related configs that are not worth adding backward compat.
See 
https://github.com/huggingface/transformers/pull/41567
https://github.com/huggingface/optimum/pull/2385

PR #3014, opened 2026-01-30
Fix two issues introduced in AutoGPTQ deprecation

githubnemo:
Merging PR #2932 introduced two bugs that lead to failing CPU and GPU pipelines.

Firstly, a merge on main in the PR without a follow-up CI run re-introduced a deleted AutoGPTQ code branch which is now removed again.

Secondly, gptqmodel seems, just like EETQ, to need a non-isolated build environment to find external dependencies like PyTorch to be installed correctly. As this was not present, the nightly slow CI wasn't run.

Additional CI fixes:
- Ignore `DeprecationWarning` from diffusers via transformers
- Fix docker build test pipeline which didn't trigger due to a formatting issue
- Added implicit `requests` dependency for `hf-doc-builder` to fix quality CI check

### DIFF
--- a/.github/workflows/test-docker-build.yml
+++ b/.github/workflows/test-docker-build.yml
@@ -28,8 +28,10 @@ jobs:
       - name: Run step if only the files listed above change
         if: steps.changed-files.outputs.any_changed == 'true'
         id: set-matrix
+        env:
+          CHANGED_FILES: "${{ steps.changed-files.outputs.all_changed_files }}"
         run: |
-          echo "matrix=${{ steps.changed-files.outputs.all_changed_files }}" >> $GITHUB_OUTPUT
+          echo "matrix=$(echo ${CHANGED_FILES} | sed -e 's/\\\"/\"/g')" >> $GITHUB_OUTPUT
   build_modified_files:
     needs: get_changed_files
     name: Build Docker images on modified files


githubnemo:
@sayakpaul I will attempt to fix the BPE `DeprecationWarning` failures in the pipeline as well. HTTP read errors seem unrelated.

.github/workflows/test-docker-build.yml
```diff
@@ -28,10 +28,8 @@ jobs:
       - name: Run step if only the files listed above change
         if: steps.changed-files.outputs.any_changed == 'true'
         id: set-matrix
-        env:
-          ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
         run: |
-          echo "matrix=${ALL_CHANGED_FILES}" >> $GITHUB_OUTPUT
+          echo "matrix=${{ steps.changed-files.outputs.all_changed_files }}" >> $GITHUB_OUTPUT
```
sayakpaul: I think the previous implement is for security reasons?

githubnemo: Yes, it was and it is indeed a potential vector since file names are controlled by a potential attacker - even though the output from the changed-files plugin is JSON encoded I wouldn't bet it is shell escaped as well. I've updated with a fix that uses shell escaping through env vars (like before) but fixes the escaping issue.

githubnemo:
HTTP errors are unrelated but `test_poly` fails. This seems to be a transformers issue introduced in http://github.com/huggingface/transformers/pull/41541. This has nothing to do with this change, so I'll fix it in a different PR.

The same goes for the failing docker build. This is best handled separately.

PR #3015, opened 2026-01-30
Test: Attempt to fix docker build CI

githubnemo:
**DO NOT MERGE**

The CI doesn't run even though there are modified files.


githubnemo:
Implemented in #3014

PR #3017, opened 2026-02-01
Fix moe layers to transform

### LINKED ISSUE #3016: layers_to_transform incorrectly matches MoE expert indices (should match layer index)
### Bug
When using `layers_to_transform` with LoRA on MoE models, modules under paths like
`model.layers.<L>.mlp.experts.<E>.*` can be incorrectly filtered/selected based on the expert index `<E>`
instead of the layer index `<L>`.

### Expected
For a module key like `model.layers.1.mlp.experts.0.up_proj`, it should be targeted iff `1 in layers_to_transform`, regardless of expert id `0`. All experts within selected layers should be included which is currently not the case.

### Repro
Example module keys:
- non-expert: `model.layers.1.self_attn.q_proj`
- expert: `model.layers.1.mlp.experts.0.up_proj`

### Proposed fix
I have a small patch + regression test ready and will open a PR.

### Who can help?

_No response_

### Reproduction

```
from torch import nn

def test_layers_to_transform_filters_by_layer_not_expert_index():
    class ToyMoEBlock(nn.Module):
        def __init__(self):
            super().__init__()
            self.self_attn = nn.Module()
            self.self_attn.q_proj = nn.Linear(4, 4, bias=False)

            self.mlp = nn.Module()
            self.mlp.experts = nn.ModuleList([nn.Module() for _ in range(2)])
            for e in range(2):
                self.mlp.experts[e].up_proj = nn.Linear(4, 4, bias=False)

        def forward(self, x):
            return x

    class ToyMoEModel(nn.Module):
        def __init__(self):
            super().__init__()
            self.model = nn.Module()
            self.model.layers = nn.ModuleList([ToyMoEBlock() for _ in range(4)])

        def forward(self, x):
            return x

    model = ToyMoEModel()

    config = LoraConfig(
        target_modules=["q_proj", "up_proj"],
        # layers_pattern="layers",
        layers_to_transform=[1],
        r=2,
        lora_alpha=4,
    )
    targeted = set(model.targeted_module_names)

    assert "model.layers.1.self_attn.q_proj" in targeted
    assert "model.layers.1.mlp.experts.0.up_proj" in targeted
    assert "model.layers.1.mlp.experts.1.up_proj" in targeted
    assert "model.layers.2.mlp.experts.1.up_proj" not in targeted  # must not match by expert index
```

### Expected behavior

Passing layers_to_transform=[1] is expected to hook onto all expert modules in that layer and no other layers, which is not the case.

MichalMraz:
Fixes #3016

### What
`layers_to_transform` filtering could treat `.experts.<idx>` as the layer index for MoE module names, leading to
incorrect selection of expert modules.

### Expected behavior
A module like `model.layers.1.mlp.experts.0.up_proj` should be targeted iff `1 in layers_to_transform`
(regardless of expert index).

### Changes
- Adjust selection logic to ignore expert indices when applying `layers_to_transform`
- Add regression test covering MoE module paths

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -1738,7 +1738,6 @@ def check_target_module_exists(config, key: str) -> bool | re.Match[str] | None:
                         layer_index = m
                         break
 
-
             if layer_index is None:
                 target_module_found = False
             else:
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -1722,13 +1722,22 @@ def check_target_module_exists(config, key: str) -> bool | re.Match[str] | None:
             # TODO: It's still unclear how empty layers_pattern (None, [], or "") should behave
             # For now, empty layers_pattern means any layer pattern is ok
             if layers_pattern is None or len(layers_pattern) == 0:
-                layer_index = re.match(r".*\.[^.]*\.(\d+)\.", key)
+                layer_index = re.search(r"(?:^|\.)[^.]*\.(\d+)\.", key)
+                # Avoid treating expert index as "layer index" in MoE module paths like "...experts.<i>...."
+                if layer_index is not None and key[layer_index.start() :].startswith(".experts."):
+                    layer_index = None
             else:
                 layers_pattern = [layers_pattern] if isinstance(layers_pattern, str) else layers_pattern
+                layer_index = None
                 for pattern in layers_pattern:
-                    layer_index = re.match(rf".*\.{pattern}\.(\d+)\.", key)
-                    if layer_index is not None:
+                    m = re.search(rf"(?:^|\.){pattern}\.(\d+)\.", key)
+                    # Avoid treating expert index as "layer index" in MoE module paths like "...experts.<i>...."
+                    if m is not None and key[m.start() :].startswith(".experts."):
+                        continue
+                    if m is not None:
+                        layer_index = m
                         break
+
 
             if layer_index is None:
                 target_module_found = False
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Thanks for reporting this issue and providing a fix. I haven't reviewed the PR in detail, but it looks that later, another PR was provided by @Mr-Neutr0n in #3028 that fixes the same issue. As your PR was first, I would treat it preferentially, but I think the solution and the tests in that PR look tighter. Could you please check if you agree with that solution?

One way forward would be to go with the other solution but to treat both of you as co-author. LMK what you think.

MichalMraz:
Thank you @BenjaminBossan, I checked the solution in PR [#3208](https://github.com/huggingface/peft/pull/3028) and I agree it is tighter and covers more cases. I think treating both of us as co-authors is fair and I'm happy to go with that.

BenjaminBossan:
Great, thanks @MichalMraz. If @Mr-Neutr0n agrees with this, I would review his PR instead and, when merging, add Michal as co-author.

Mr-Neutr0n:
I'm more than happy to do so!

Mr-Neutr0n:
hey @benjaminbossan please review this PR. thanks!

BenjaminBossan:
@Mr-Neutr0n I think we agreed to proceed with #3028 instead of this PR. I reviewed that one and am awaiting your updates there.

BenjaminBossan:
@MichalMraz Since @Mr-Neutr0n seems to be MIA, would you be interested in picking this up again?

MichalMraz:
Hi @BenjaminBossan yeah I'm happy to pick this up. Will need a week or two since I'm quite busy right now, but will definitely get to it.

BenjaminBossan:
Thanks @MichalMraz. There is no time pressure, just LMK if you don't manage to find the time so that someone else can pick this up.

MichalMraz:
Hi @BenjaminBossan, I've provided a fix in #3318 by combining the tighter regex solution by @Mr-Neutr0n and appropriate tests by me

PR #3024, opened 2026-02-06
[TinyLoRA]tinylora implementation

kashif:
Adds TinyLoRA, a new PEFT method based on "TinyLoRA: Learning to Reason in 13 Parameters". TinyLoRA achieves extreme parameter efficiency by replacing LoRA's trainable low-rank  matrices with a tiny trainable vector projected through fixed random bases.

The key idea: given a frozen SVD decomposition `W ≈ B @ A` (where `B = U @ sqrt(S)` and `A = sqrt(S) @ V^T`), the weight update is `delta_W = B @ R @ A` where `R` is an  `r x r` trainable matrix (following LoRA-XS). TinyLoRA takes this further by  parameterizing `R` as a linear combination of fixed random projection matrices:

      R = sum_i(v[i] * P[i])

  where `v` is the only trainable parameter (as small as 13 values) and `P_i` are fixed  random matrices seeded deterministically.
  
  ## Features

  - Extreme efficiency: trainable parameter count is `u` per target module (or even less with weight tying), compared to `r * (in + out)` for LoRA
  - Weight tying: configurable sharing of `v` vectors across layers via `weight_tying` (0.0 = no sharing, 1.0 = all layers share one `v`)
  - SVD initialization: frozen `A` and `B` matrices computed from truncated SVD of pretrained weights, with singular values distributed equally via `sqrt(S)`
  - Full layer support: `nn.Linear`, `Conv1D`, and `nn.Embedding`
  - Merge/unmerge: full support including safe merge with NaN checking
  - LoRA conversion: `supports_lora_conversion()` -> True — delta weights can be converted to standard LoRA format via `get_delta_weight`
  - Deterministic projections: `P` matrices are seeded per-layer for reproducibility; optionally saved in checkpoints (`save_projection=True`)

## Config
```
  from peft import TinyLoraConfig, get_peft_model

  config = TinyLoraConfig(
      r=2,              # SVD rank (frozen)
      u=64,             # trainable vector dimension
      weight_tying=0.0, # 0.0=no sharing, 1.0=full sharing
      target_modules="all-linear",
  )
  model = get_peft_model(base_model, config)
  ```
  
 ## Architecture

  - TinyLoraLayer (base): SVD decomposition, projection init, `get_delta_weight`, `supports_lora_conversion`
  - Linear / Embedding: forward pass, merge/unmerge
  - TinyLoraModel: weight tying groups, shared v parameter management via nested ModuleDict/ParameterDict
  - update_layer follows LoRA's config-object pattern: (adapter_name, tinylora_v, v_key, r, config, **kwargs)

### CODE REVIEW

**Comment by kashif** [comment]
cc @jxmorris12 I have an implementation of TinyLoRA if you can kindly have a look?

File: docs/source/package_reference/tinylora.md [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# TinyLoRA: Learning to Reason in 13 Parameters
+
+[TinyLoRA](https://huggingface.co/papers/2602.04118) is an extremely parameter-efficient fine-tuning technique that enables fine-tuning with as few as 1-13 trainable parameters. It builds upon the [LoRA-XS](https://huggingface.co/papers/2405.17604) approach by using SVD decomposition of frozen weights and projecting a tiny trainable vector through fixed random tensors.
```
githubnemo: I think this is misleading as this claim is only valid in reinforcement learning settings (as far as we know currently).

docs/source/package_reference/tinylora.md
```diff
@@ -0,0 +1,41 @@
...
+
+The key innovation of TinyLoRA is replacing the trainable low-rank matrix R with a weighted sum of fixed random projection matrices: `R = Σᵢ vᵢ Pᵢ`, where `v` is a tiny trainable vector and `Pᵢ` are fixed random matrices. This dramatically reduces the number of trainable parameters while maintaining competitive performance, especially when combined with reinforcement learning training methods like GRPO.
+
+TinyLoRA supports weight tying through the `ntie` parameter, which controls how many modules share the same trainable vector `v`. With full tying across all target modules, you can achieve extreme parameter efficiency with just `u` trainable parameters for the entire model.
```
githubnemo: `u` is not defined here

src/peft/tuners/tinylora/__init__.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
githubnemo: Let's fix the date here and in the other new files, it should be `2026-present`

src/peft/tuners/tinylora/config.py
```diff
@@ -0,0 +1,190 @@
...
+
+    Args:
+        r (`int`, *optional*, defaults to `2`):
+            SVD rank for the frozen U, Sigma, V decomposition. The paper recommends r=2.
+        u (`int`, *optional*, defaults to `64`):
+            Trainable vector dimension per group. This controls the expressivity of the adaptation. Can be as low as
+            1-13 for extreme parameter efficiency.
+        ntie (`int`, *optional*, defaults to `1`):
+            Weight tying factor. Controls how many modules share the same trainable vector v. ntie=1 means each layer
+            has its own v. Higher values reduce parameters further.
```
githubnemo: As a user, if I want to enable complete weight tying then I'd have to know the number of targeted modules beforehand - would it be better to make this a fraction with 0 being no weight sharing, 1 being full weight sharing?

src/peft/tuners/tinylora/config.py
```diff
@@ -0,0 +1,190 @@
...
+            Random seed for generating the fixed projection matrices P.
+        save_projection (`bool`, *optional*, defaults to `True`):
+            Whether to save the projection tensors P in the state dict. If False, they will be regenerated from the
+            seed when loading.
+        init_v_bound (`float`, *optional*, defaults to `0.02`):
+            Uniform initialization bound for the trainable vector v. Values are initialized in [-init_v_bound,
+            init_v_bound].
+        target_modules (`Union[List[str], str]`, *optional*):
+            The names of the modules to apply TinyLoRA to. Only `nn.Linear`, `nn.Embedding`, and
+            `transformers.pytorch_utils.Conv1D` layers are supported.
```
githubnemo: Let's also document that it is possible to use regex and special keywords like `all-linear`.

src/peft/tuners/tinylora/layer.py
```diff
@@ -0,0 +1,699 @@
...
+        elif isinstance(base_layer, Conv1D):
+            in_features, out_features = (
+                base_layer.weight.ds_shape if hasattr(base_layer.weight, "ds_shape") else base_layer.weight.shape
+            )
+        elif isinstance(base_layer, nn.Embedding):
+            in_features, out_features = base_layer.num_embeddings, base_layer.embedding_dim
+        else:
+            # Fallback for other layer types
+            in_features, out_features = None, None
+
```
githubnemo: Let's replace the determination of in/out features with `peft.tuners.tuners_utils._get_in__out_features`

src/peft/tuners/tinylora/layer.py
```diff
@@ -0,0 +1,699 @@
...
+        self.in_features = in_features
+        self.out_features = out_features
+        self.kwargs = kwargs
+
+    @property
+    def merged(self) -> bool:
+        return bool(self.merged_adapters)
+
```
githubnemo:
```suggestion
```

Already defined in `BaseTunerLayer`

src/peft/tuners/tinylora/layer.py
```diff
@@ -0,0 +1,699 @@
...
+        V_r = Vh[:actual_r, :].to(dtype)
+
+        # Store in LoRA-XS convention:
+        # tinylora_A = V_r (actual_r x in_features) - acts like lora_A
+        # tinylora_B = U_r @ diag(S_r) (out_features x actual_r) - acts like lora_B
+        # Use .contiguous() to ensure tensors can be saved with safetensors
+        self.tinylora_A[adapter_name] = V_r.contiguous()
+        self.tinylora_B[adapter_name] = (
+            U_r * S_r.unsqueeze(0)
+        ).contiguous()  # Broadcasting: (out, actual_r) * (1, actual_r)
```
githubnemo: This makes the norm of `B` a lot larger (i.e. scaled by `S_r`) than that of `A`. Maybe distributing it to both by using `sqrt(S_r)` is better?

src/peft/tuners/tinylora/model.py
```diff
@@ -0,0 +1,332 @@
...
+
+        for existing_config in self.peft_config.values():
+            if existing_config is config:
+                continue
+
+            if existing_config.projection_seed != config.projection_seed:
+                raise ValueError(
+                    f"TinyLoRA projection seed must be the same for all adapters. Got {config.projection_seed=} but "
+                    f"previous config had {existing_config.projection_seed}."
+                )
```
githubnemo: Why does the projection seed need to be same across adapters? This seems like a inconvenience when loading more than one adapter from different sources. `tinylora_P` already depends on the adapter name so it should be fine?

src/peft/tuners/tinylora/model.py
```diff
@@ -0,0 +1,332 @@
...
+                dtype = target.weight.dtype
+            else:
+                dtype = None  # Will default to float32
+            v = nn.Parameter(torch.empty(tinylora_config.u, dtype=dtype))
+            if tinylora_config.init_weights:
+                nn.init.uniform_(v, -tinylora_config.init_v_bound, tinylora_config.init_v_bound)
+            else:
+                # Initialize to zeros for identity operation
+                nn.init.zeros_(v)
+            self.tinylora_v[v_key] = v
```
githubnemo: I'm not sure if I understand everything correctly so let me ask possibly stupid questions:

1. we have `target_name` as an unique identifier, why not use this instead of counting the modules which may be brittle when layer retrieval code changes
2. why not pass the value of `v` directly to the layer since it is passed by reference and not touched anyway?

src/peft/tuners/tinylora/model.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def print_trainable_parameters(self) -> None:
+        """
+        Prints the number of trainable parameters in the model.
+        """
+        trainable_params, all_param = self.get_nb_trainable_parameters()
+        print(
+            f"trainable params: {trainable_params:,d} || all params: {all_param:,d} || "
+            f"trainable%: {100 * trainable_params / all_param:.4f}"
+        )
```
githubnemo: I think `get_nb_trainable_parameters` and `print_trainable_parameters` can be removed since they're already implemented in `PeftModel`

src/peft/tuners/tinylora/model.py
```diff
@@ -0,0 +1,332 @@
...
+        super()._cast_adapter_dtype(adapter_name, autocast_adapter_dtype)
+
+        if not autocast_adapter_dtype:
+            return
+
+        # Handle model-level tinylora_v parameters
+        dtypes_to_convert_to_fp32 = {torch.float16, torch.bfloat16}
+        for key, param in self.tinylora_v.items():
+            # Check if this key belongs to the adapter (key format: "{adapter_name}_group_{idx}")
+            if key.startswith(f"{adapter_name}_group_"):
```
githubnemo: If it is not too much trouble I'd prefer a nested parameter dictionary.

src/peft/peft_model.py
```diff
@@ -580,11 +580,21 @@ def from_pretrained(
...
-            k for k in load_result.missing_keys if "vblora_vector_bank" not in k and "prompt_encoder" not in k
-        ]
+        # 3. Remove TinyLoRA layer-level tinylora_v references (they share with model-level tinylora_v)
+        def is_expected_missing_key(k):
+            if "vblora_vector_bank" in k:
+                return False
+            if "prompt_encoder" in k:
+                return False
+            # TinyLoRA: layer-level tinylora_v is a reference to model-level, exclude from warning
+            if ".model." in k and ".tinylora_v." in k:
```
githubnemo:
```suggestion
            if ".tinylora_v." in k:
```

I think `.model.` is in the key in any case and if it's not would lead to a false-negative

src/peft/peft_model.py
```diff
@@ -1406,6 +1416,9 @@ def load_adapter(
         # Filter missing keys specific to the current adapter and tuner prefix.
         for key in load_result.missing_keys:
             if tuner_prefix in key and adapter_name in key:
+                # TinyLoRA: layer-level tinylora_v is a reference to model-level, skip it
+                if ".model." in key and ".tinylora_v." in key:
```
githubnemo: same as above

src/peft/tuners/tinylora/config.py
```diff
@@ -0,0 +1,190 @@
...
+            )
+        },
+    )
+    init_weights: bool = field(
+        default=True,
+        metadata={
+            "help": (
+                "Whether to initialize the trainable vector v with random values. If True, v is initialized "
+                "with uniform random values. If False, v is initialized to zeros, making the adapter an "
+                "identity operation (no change to base model output)."
```
githubnemo: This is inconsistent with most PEFT methods, let's swap it around: default is a no-op, `"uniform"` initializes `v` uniformly random and `False` is leaving the vector as-is.

tests/test_decoder_models.py
```diff
@@ -300,6 +301,18 @@
...
+    (
+        TinyLoraConfig,
+        {
+            "task_type": "CAUSAL_LM",
+            "r": 2,
+            "u": 64,
+            "target_modules": None,
+            "tinylora_dropout": 0.0,
+            "projection_seed": 42,
+            "save_projection": True,
```
githubnemo:
```suggestion
```

Let's remove default values

tests/test_encoder_decoder_models.py
```diff
@@ -216,6 +217,18 @@
...
     ),
+    (
+        TinyLoraConfig,
+        {
+            "r": 2,
+            "u": 64,
+            "target_modules": None,
+            "tinylora_dropout": 0.0,
+            "projection_seed": 42,
+            "save_projection": True,
```
githubnemo:
```suggestion
```

tests/test_feature_extraction_models.py
```diff
@@ -215,6 +216,18 @@
...
+    (
+        TinyLoraConfig,
+        {
+            "task_type": "FEATURE_EXTRACTION",
+            "r": 2,
+            "u": 64,
+            "target_modules": None,
+            "tinylora_dropout": 0.0,
+            "projection_seed": 42,
+            "save_projection": True,
```
githubnemo:
```suggestion
```

tests/test_seq_classifier.py
```diff
@@ -216,6 +217,18 @@
...
+    (
+        TinyLoraConfig,
+        {
+            "task_type": "SEQ_CLS",
+            "r": 2,
+            "u": 64,
+            "target_modules": None,
+            "tinylora_dropout": 0.0,
+            "projection_seed": 42,
+            "save_projection": True,
```
githubnemo:
```suggestion
```

kashif:
thanks fixing

src/peft/tuners/tinylora/model.py
kashif: yes good idea, let me fix

kashif:
@githubnemo should be ready for another review thanks

src/peft/tuners/tinylora/model.py
```diff
@@ -0,0 +1,320 @@
...
+            target.set_layer_idx(layer_idx)
+            target.update_layer(
+                adapter_name,
+                self.tinylora_v,
+                v_key,
+                tinylora_config.r,
+                tinylora_config.u,
+                tinylora_config.tinylora_dropout,
+                tinylora_config.init_weights,
+                tinylora_config.projection_seed,
```
githubnemo: Let's pass the config directly similarly to `lora/layer.py:update_layer`.

src/peft/tuners/tinylora/layer.py
```diff
@@ -0,0 +1,633 @@
...
+        # delta_W = B @ R @ A
+        # B: (out, r), R: (r, r), A: (r, in)
+        # Result: (out, in)
+        delta = B @ R @ A
+
+        if cast_to_fp32:
+            delta = delta.to(dtype=dtype)
+
+        return delta
+
```
githubnemo, line 305: Let's also add a method `supports_lora_conversion() -> True` since (I think) this method supports being converted to standard LoRA (since it is not modifying the base weights and implements `get_delta_weight`).

kashif:
ready @githubnemo

BenjaminBossan:
Thanks for the PR Kashif. I ran the experiments on my machine and for got a test accuracy of 0% and 0.002% :)
This isn't really surprising with 3584 and 64 trainable parameters, respectively, and basically confirms the paper results. Still good to include them, but without context users could be drawing the wrong conclusion.

kashif:
yes @BenjaminBossan i will test with the RL setup, we can wait if its ok, I want to also double check that nothing is wrong

BenjaminBossan:
Out of curiosity, I wanted to check if TinyLoRA can achieve better scores if we increase the number of trainable parameters. So I took the default* setting and increased `u` and indeed, we can get decent results.

| u    | accelerator memory max | num trainable params | test_accuracy        |
|------|------------------------|----------------------|----------------------|
| 512  | 20310917120            | 28672                | 0.30477634571645185  |
| 1024 | 20434649088            | 57344                | 0.3434420015163002   |
| 2024 | 20673724416            | 113344               | 0.31766489764973466  |

Given the still tiny number of trainable parameters, this result is quite respectable. This is also a nice confirmation that there is no major bug in the implementation.

I wonder if it would make sense to have a "maximalist" and a "minimalist" config, i.e. one with more trainable parameters and better score and one with extremely few trainable parameters (basically the current `llama-3.2-3B-weight-tying`) and low score.

*One more change I did was to increase `r` to 32 as 2 seemed pretty small to me, but that was just on a whim.

githubnemo:
> I wonder if it would make sense to have a "maximalist" and a "minimalist" config, i.e. one with more trainable parameters and better score and one with extremely few trainable parameters (basically the current `llama-3.2-3B-weight-tying`) and low score.

I think that's a good thing to have!

I also wondered if it would make sense to extend the target modules to `all-linear` since it is so cheap, but then again this would diverge from other experiments quite a lot and be not as comparable.

kashif:
should we just document this? or add it somewhere else?

BenjaminBossan:
> I also wondered if it would make sense to extend the target modules to `all-linear` since it is so cheap, but then again this would diverge from other experiments quite a lot and be not as comparable.

I'd rather target either the attention xor the MLP part for consistence with other experiments.

> should we just document this? or add it somewhere else?

What does "this" reference here?

kashif:
ah sorry, i meant the minimal/maximal config?

githubnemo:
> ah sorry, i meant the minimal/maximal config?

Yes, let's add a 'maximalist' config with (possibly) r > 2 and u >= 2048. VeRA has \~128k parameters with 37.6% task accuracy according to https://huggingface.co/spaces/peft-internal-testing/PEFT-method-comparison - maybe it makes sense to match that setting (`u ~= 2300`) to see where the ceiling is within a reasonable memory budget (~+0.5GB).

githubnemo (approved):
The benchmarks spiraled a bit into side-projects. I think we can merge this without having the additional configs and check-in specific benchmarks at a later point in time.

Implementation LGTM.

mengxi11:
tinylora实际性能可以达到论文里的多少？

BenjaminBossan:
@mengxi11 We're working on a RL benchmark to check this: #3078.

mengxi11:
再请问下关于tinylora的4bit量化微调训练貌似还没打通?

BenjaminBossan:
TinyLoRA doesn't support quantization yet. We have plans for that (#3117) but that will probably take some time to land in PEFT. You could try something like

- LoRA + 4bit quantization + tiny rank (`r=1`)
- IA³ + 4bit quantization (set a high learning rate, ~1e-3)

Those are not the same as TinyLoRA, but they support bitsandbytes and are also very parameter efficient.

PR #3027, opened 2026-02-07
fix: correct typo 'occurence' to 'occurrence'

thecaptain789:
Fixes typo in calculate_alora_offsets docstring.


BenjaminBossan:
Thanks for the PR but as a general stance, we don't accept these mass-created very low impact PRs.

PR #3028, opened 2026-02-07
fix: layers_to_transform now correctly matches layer index on MoE models

### LINKED ISSUE #3016: layers_to_transform incorrectly matches MoE expert indices (should match layer index)

Mr-Neutr0n:
## Summary
Fix `layers_to_transform` incorrectly matching MoE expert indices instead of layer indices.

## Problem
When using `layers_to_transform` on MoE (Mixture of Experts) models, modules under paths like `model.layers.<L>.mlp.experts.<E>.*` were incorrectly filtered based on the expert index `<E>` instead of the layer index `<L>`.

**Example:**
For key `model.layers.1.mlp.experts.0.up_proj`:
- **Expected:** Match layer index `1`
- **Actual (before fix):** Matched expert index `0`

This caused `layers_to_transform=[1]` to incorrectly exclude `model.layers.1.mlp.experts.0.up_proj` (because `0 != 1`) and incorrectly include `model.layers.2.mlp.experts.1.up_proj` (because `1 == 1`).

## Root Cause
The regex in `check_target_module_exists()` used greedy `.*` which matched as much as possible, leaving only the LAST digit in the path (the expert index) for the capture group:

```python
layer_index = re.match(r".*\.[^.]*\.(\d+)\.", key)  # Greedy
```

## Solution
Changed to non-greedy `.*?` which matches as little as possible, capturing the FIRST digit in the path (the layer index):

```python
layer_index = re.match(r".*?\.[^.]*\.(\d+)\.", key)  # Non-greedy
```

## Test Cases Added
Added 7 test cases covering MoE scenarios:
- With explicit `layers_pattern`
- Without `layers_pattern` (default behavior)

Fixes #3016

### DIFF
--- a/tests/test_tuners_utils.py
+++ b/tests/test_tuners_utils.py
@@ -128,6 +128,16 @@ REGEX_TEST_CASES = [
     ("blocks.1.bias", ["weight"], [1], ["blocks"], False),
     ("mlp.blocks.1.weight", ["weight"], [1], ["blocks"], True),
     ("mlp.blocks.1.bias", ["weight"], [1], ["blocks"], False),
+    # MoE models: layers_to_transform should match layer index, not expert index
+    # For "model.layers.1.mlp.experts.0.up_proj", should match layer 1, not expert 0
+    ("model.layers.1.mlp.experts.0.up_proj", ["up_proj"], [1], ["layers"], True),
+    ("model.layers.1.mlp.experts.0.up_proj", ["up_proj"], [0], ["layers"], False),  # expert 0, but layer 1
+    ("model.layers.1.mlp.experts.1.up_proj", ["up_proj"], [1], ["layers"], True),  # expert 1, layer 1
+    ("model.layers.2.mlp.experts.1.up_proj", ["up_proj"], [1], ["layers"], False),  # layer 2, not 1
+    # MoE without explicit layers_pattern - should still match layer index, not expert index
+    ("model.layers.1.mlp.experts.0.up_proj", ["up_proj"], [1], None, True),
+    ("model.layers.1.mlp.experts.0.up_proj", ["up_proj"], [0], None, False),  # must not match expert 0
+    ("model.layers.2.mlp.experts.1.up_proj", ["up_proj"], [1], None, False),  # must not match expert 1
 ]
 
 MAYBE_INCLUDE_ALL_LINEAR_LAYERS_TEST_CASES = [
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -1722,7 +1722,10 @@ def check_target_module_exists(config, key: str) -> bool | re.Match[str] | None:
             # TODO: It's still unclear how empty layers_pattern (None, [], or "") should behave
             # For now, empty layers_pattern means any layer pattern is ok
             if layers_pattern is None or len(layers_pattern) == 0:
-                layer_index = re.match(r".*\.[^.]*\.(\d+)\.", key)
+                # Use non-greedy .*? to match the FIRST digit group, not the last
+                # This prevents MoE expert indices from being matched instead of layer indices
+                # e.g., for "model.layers.1.mlp.experts.0.up_proj", we want to match "1" not "0"
+                layer_index = re.match(r".*?\.[^.]*\.(\d+)\.", key)
             else:
                 layers_pattern = [layers_pattern] if isinstance(layers_pattern, str) else layers_pattern
                 for pattern in layers_pattern:


Mr-Neutr0n:
Friendly bump! Let me know if there's anything I should update or improve to help move this forward.

BenjaminBossan:
Thanks for this PR. Please check my [comment here](https://github.com/huggingface/peft/pull/3017#issuecomment-3896701376).

_Edit: Add MichalMraz as co-author here._

tests/test_tuners_utils.py
```diff
@@ -128,6 +128,16 @@
     ("blocks.1.bias", ["weight"], [1], ["blocks"], False),
     ("mlp.blocks.1.weight", ["weight"], [1], ["blocks"], True),
     ("mlp.blocks.1.bias", ["weight"], [1], ["blocks"], False),
+    # MoE models: layers_to_transform should match layer index, not expert index
+    # For "model.layers.1.mlp.experts.0.up_proj", should match layer 1, not expert 0
+    ("model.layers.1.mlp.experts.0.up_proj", ["up_proj"], [1], ["layers"], True),
+    ("model.layers.1.mlp.experts.0.up_proj", ["up_proj"], [0], ["layers"], False),  # expert 0, but layer 1
+    ("model.layers.1.mlp.experts.1.up_proj", ["up_proj"], [1], ["layers"], True),  # expert 1, layer 1
+    ("model.layers.2.mlp.experts.1.up_proj", ["up_proj"], [1], ["layers"], False),  # layer 2, not 1
```
BenjaminBossan, line 136: These tests don't really test the mentioned bug. E.g. if the key is `"model.layers.1.mlp.experts.0.up_proj"`, with the `layers_pattern` being `["layers"]`, there is just a single match, namely `"layers.1`". `"experts.0"` would never be matched in this situation, even with the current code from main. In fact, that code path is never reached, as we land here:

https://github.com/huggingface/peft/blob/8f73327fb967ab47b83aac604bbc2e193207d0f4/src/peft/tuners/tuners_utils.py#L1726-L1731

At this stage, we still match the last number, but since there is only one match, it doesn't matter. With the change from this PR, however, I would always expect the first number to match to be consistent with the other code path. Therefore, I would suggest to change that regex too, and to update the examples to really test for this.

BenjaminBossan (changes requested):
Thanks for the PR. I found that the code path for `layers_pattern != None` is not covered by this change. Let's update it too for consistency. More details in my comment.

Note: The failing CI is unrelated and can be ignored.

BenjaminBossan:
ping @Mr-Neutr0n

Mr-Neutr0n:
pong

Mr-Neutr0n:
Will address the comments. Out for an emergency the past few days. Thanks for considering!

BenjaminBossan:
Gentle ping @Mr-Neutr0n

BenjaminBossan:
Another ping @Mr-Neutr0n. If you cannot work on this anymore, I'll hand the task over to someone else (or will do it myself).

Mr-Neutr0n:
Hi @BenjaminBossan 

Please allow me some time. Thanks!

BenjaminBossan:
> Please allow me some time. Thanks!

Yes, sure, I just wasn't sure if you were still on it. LMK if you need help.

BenjaminBossan:
@Mr-Neutr0n Is this PR still on your TODO list?

PR #3029, opened 2026-02-09
nightly-gpu: Make sure that all steps are run

githubnemo:
Currently if one step fails the run is aborted, masking potential errors, especially in times where we know that the error that is causing the initial failure is not easily fixed (e.g., by being an issue in a dependency) as is currently the case with GPQTModel.

The solution would be to run all the steps and fail the run if one of those steps had an error. Since we log all errors anyway this is not worse than before.


githubnemo:
closing this in favor of #3030 since this is not testable.

PR #3032, opened 2026-02-09
`no_split_modules` now captures values recursively

githubnemo:
As of transformers v5.1.0 (specifically [this commit](https://github.com/huggingface/transformers/commit/36ec3bfa33ebf6c3b38a1d6808292aeea4aae84d)) the `model._no_split_modules` attribute captures the values of the model's children and returns a set of values instead of a list. In case of transformers >=v5.1.0 we can therefore just return the attribute and save iterating over the module. This is also reflected in the test.

### DIFF
--- a/tests/test_other.py
+++ b/tests/test_other.py
@@ -536,6 +536,12 @@ class TestGetNoSplitModules:
         model_id = "hf-internal-testing/tiny-random-LlavaForConditionalGeneration"
         model = LlavaForConditionalGeneration.from_pretrained(model_id)
 
+        # model._no_split_modules is recursively generated as of transformers 5.1.0 so
+        # depending on which transformers version we have in the test environment the
+        # attribute will deliver either the same result as `_get_no_split_modules`
+        # or an empty list.
+        #
+        # TODO remove this distinction once transformers <5.1.0 is not supported anymore
         if not is_transformers_ge_v5_1_0:
             # sanity check: just visiting the model itself is not enough:
             assert model._no_split_modules == []


src/peft/utils/other.py
```diff
@@ -1146,6 +1147,10 @@ def _get_no_split_modules(model) -> set[str]:
     if not hasattr(model, "_no_split_modules"):
         return _no_split_modules
 
+    if is_transformers_ge_v5_1_0:
+        # See https://github.com/huggingface/transformers/commit/36ec3bfa33ebf6c3b38a1d6808292aeea4aae84d
```
BenjaminBossan, line 1149: Let's add a TODO comment to remove the code below once we drop support for transformers < 5.1.0.

src/peft/utils/other.py
```diff
@@ -77,6 +77,7 @@
 
     mlu_available = is_mlu_available()
 
+is_transformers_ge_v5_1_0 = version.parse(transformers.__version__) >= version.parse("5.1.0")
```
BenjaminBossan: We have similar code in multiple places now:

- https://github.com/huggingface/peft/blob/2cd96ed041620f74d239a0ce3f16207153c43413/src/peft/utils/integrations.py#L27
- https://github.com/huggingface/peft/blob/2cd96ed041620f74d239a0ce3f16207153c43413/src/peft/utils/constants.py#L48
- several inlined checks

I think we can keep the inlined checks but should unify the rest. Personally, I don't care much about where to place them, let's just ensure it's a single place.

tests/test_other.py
```diff
@@ -526,16 +531,20 @@ def test_get_no_split_modules_simple(self):
...
         model_id = "hf-internal-testing/tiny-random-LlavaForConditionalGeneration"
         model = LlavaForConditionalGeneration.from_pretrained(model_id)
-        # sanity check: just visiting the model itself is not enough:
-        assert model._no_split_modules == []
+
+        if not is_transformers_ge_v5_1_0:
+            # sanity check: just visiting the model itself is not enough:
+            assert model._no_split_modules == []
+        else:
+            assert model._no_split_modules == {"CLIPEncoderLayer", "LlamaDecoderLayer"}
```
BenjaminBossan, line 549: Let's add a comment on why the version matters and let's also add a TODO to remove the other code once transformers < 5.1.0 is dropped.

Issue #2563, opened 2025-06-02
Integrate Lily


yibozhong:
### Feature request

This request proposes integrating Lily (Low-Rank Interconnected Adaptation across Layers), accepted to ACL 2025 Findings, into the PEFT library.  
Paper: https://arxiv.org/pdf/2407.09946  
Repo: https://github.com/yibozhong/lily  

### Motivation

Lily aims to directly make the rank of each individual adapter bigger under the same parameter budget, as it's shown in many papers that higher ranks are beneficial to PEFT performance. This is achieved by breaking the pair-AB-per-layer constraint of LoRA. That is, we do not give each layer a dedicated pair of A and B. Rather, we decouple all the Bs from the layer, and when adapting at each layer, we use a weighted sum of these Bs as the B for this layer. The weight is calculated by a lightweight trainable router, currently data-dependent.  

![Image](https://github.com/user-attachments/assets/809c25a4-63f5-4ec7-bdb4-98a8f869f328)  

Several points worth noting:  
- The method looks somewhat similar to MosLoRA in structure, but it operates at the model level and the aim is to increase the individual rank of each adapter with dynamic adaptation.  
- Currently in the paper, we use a data-dependent router, which makes it tricky to merge the weights. I do not observe notable inference latency, possibly due to small model size, but an option for using a non-data-dependent router can be included and enable easy merging the weights.  
- The current As are still positioned at a fixed layer (using layer-wise sharing to reduce params). However, it also can be decoupled, simply by providing two routers for weighting As and Bs respectively, rather than one router for B in the current setup. This is a more elegant design and shares the same principle as Lily. After I run quick experiments demonstrating its effectiveness, I can integrate this setup into my current code as Lily v2.  

### Your contribution

Implement Lily, repo: https://github.com/yibozhong/lily.

BenjaminBossan:
Thank you for proposing to add Lily to PEFT. Generally, after quickly looking at the paper, I think it could be a nice addition. As you mentioned, the data-dependent routing makes things a bit more complicated. Could you please point me to the part of the code where this is implemented so that I can gain a better understanding?

> we use a data-dependent router, which makes it tricky to merge the weights.

Note that the possibility of merging is not a strict requirement for PEFT, so if it's not supported, it's okay.

> After I run quick experiments demonstrating its effectiveness, I can integrate this setup into my current code as Lily v2.

[...]

PR #3036, opened 2026-02-11
[Lily] implementations for peft integration

yibozhong:
As detailed in https://github.com/huggingface/peft/issues/2563

### DIFF
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -2213,7 +2213,6 @@ class TestPeftCustomModel(PeftCommonTester):
             or issubclass(config_cls, VBLoRAConfig)
             or issubclass(config_cls, LilyConfig)
         ):
-            print("Using a lower learning rate for this test case to avoid NaN losses")
             # this high learning rate was found through testing to be necessary to avoid flakiness
             lr = 100
         elif "mha" in model_id.lower():


src/peft/tuners/lily/__init__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan: Let's update all the years to 2026.

src/peft/tuners/lily/config.py
```diff
@@ -0,0 +1,105 @@
...
+from peft.config import PeftConfig
+from peft.utils import PeftType
+
+
+@dataclass
+class LilyConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`~peft.Lily`].
+
+    Args:
```
BenjaminBossan, line 28: Let's extend the documentation to make it more user friendly. As a user, I would like to know what changing this parameter does (e.g. increase capacity at the expense of overfitting) and what good values should be. We can assume that users understand what rank and dropout do, as those are general parameters, but please expand the docs of the Lily-specific parameters.

src/peft/tuners/lily/config.py
```diff
@@ -0,0 +1,105 @@
...
+from typing import Optional, Union
+
+from peft.config import PeftConfig
+from peft.utils import PeftType
+
+
+@dataclass
+class LilyConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`~peft.Lily`].
```
BenjaminBossan: This link won't work.

src/peft/tuners/lily/config.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Args:
+        r (`int`): Lily's rank
+        num_A (`int`): Lily's number of As
+        num_B (`int`): Lily's number of experts (ne) of Bs
+        target_modules (`Union[List[str],str]`): The names of the modules to apply Lily to.
+        lily_scaling (`float`): The scaling factor for lily.
```
BenjaminBossan: I wonder if it would make sense to have an `alpha` parameter instead of `lily_scaling`, similar to what we have in LoRA etc. (i.e. normalized by rank). Is there a specific reason why you chose not to use `alpha` for scaling?

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,387 @@
...
+            lily_Bs[key] = {} 
+
+        num_layers = self.num_of_layers(model, adapter_name)
+
+        stride = {}
+
+        for target in num_layers.keys():
+            stride[target] = {}
+            for shape in num_layers[target].keys():
+                assert num_layers[target][shape] % peft_config.num_A == 0, f"num of layers {num_layers[target][shape]} for target {target} with shape {shape} is not divisible by num_A {peft_config.num_A}"
```
BenjaminBossan: Let's not use asserts but raise proper exceptions.

src/peft/tuners/lily/layer.py
```diff
@@ -0,0 +1,245 @@
...
+        self.lily_dropout = nn.ModuleDict({})
+        self.num_A = {}
+        self.num_B = {}
+        self.lily_A = nn.ModuleDict({})
+        self.lily_B = nn.ModuleDict({})
+        self.lily_router = nn.ModuleDict({})
+        self.kwargs = kwargs
+
+        base_layer = self.get_base_layer()
+        if isinstance(base_layer, nn.Linear):
```
BenjaminBossan, line 48: Let's only check the layers that are actually supported.

src/peft/tuners/lily/layer.py
```diff
@@ -0,0 +1,245 @@
...
+        else:
+            lily_dropout_layer = nn.Identity()
+        
+        self.lily_dropout.update(nn.ModuleDict({adapter_name: lily_dropout_layer}))
+        # Actual trainble parameters
+        self.lily_A[adapter_name] = lily_A if lily_A is not None else nn.Linear(self.in_features, r, bias=False)
+        self.lily_B[adapter_name] = lily_B if lily_B is not None else nn.Linear(r, self.out_features, bias=False)
+        self.lily_router[adapter_name] = nn.Linear(r, num_B, bias=False)
+        self.num_A = num_A
+        self.num_B = num_B
```
BenjaminBossan:
```suggestion
        self.num_A[adapter_name] = num_A
        self.num_B[adapter_name] = num_B
```

src/peft/tuners/lily/layer.py
```diff
@@ -0,0 +1,245 @@
...
+            r,
+            lily_scaling=lily_scaling,
+            lily_dropout=lily_dropout,
+            lily_A=lily_A,
+            lily_B=lily_B,
+            num_A=num_A,
+            num_B=num_B,
+        )
+
+    def get_delta_weight(self, adapter, x: torch.Tensor) -> torch.Tensor:
```
BenjaminBossan: I don't think it makes sense to have a `get_delta_weight` that expects an `x` as input. As this method is mainly used for merging, and we agreed that merging is not supported, let's raise a `NotImplementedError` here. Similarly, let's add a `merge` and `unmerge` method and raise the error.

src/peft/tuners/lily/layer.py
```diff
@@ -0,0 +1,245 @@
...
+
+        for active_adapter in self.active_adapters:
+            if active_adapter not in lily_A_keys:
+                continue
+
+            lily_A = self.lily_A[active_adapter]
+            lily_B = self.lily_B[active_adapter]
+            router = self.lily_router[active_adapter]
+            B = einops.rearrange(lily_B.weight, "(e i) o -> e i o", e=self.num_B)
+            dropout = self.lily_dropout
```
BenjaminBossan:
```suggestion
            dropout = self.lily_dropout[active_adapter]
```

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,387 @@
...
+from peft.import_utils import is_bnb_4bit_available, is_bnb_available
+from peft.tuners.tuners_utils import (
+    BaseTuner,
+    BaseTunerLayer,
+    check_target_module_exists,
+    onload_layer,
+    replicate_layers,
+)
+from peft.utils import (
+    TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING,
```
BenjaminBossan:
```suggestion
```

Let's use `TRANSFORMERS_MODELS_TO_LILY_TARGET_MODULES_MAPPING` everywhere.

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,387 @@
...
+    def _create_new_module(lily_config, adapter_name, target, lily_A, lily_B, **kwargs):
+        if isinstance(target, BaseTunerLayer):
+            target_base_layer = target.get_base_layer()
+        else:
+            target_base_layer = target
+
+        new_module = None
+        if isinstance(target_base_layer, torch.nn.Linear):
+            new_module = Linear(target, adapter_name, r=lily_config.r, lily_scaling=lily_config.lily_scaling, lily_dropout=lily_config.lily_dropout, num_A=lily_config.num_A, num_B=lily_config.num_B, lily_A=lily_A, lily_B=lily_B, **kwargs)
+
```
BenjaminBossan: Let's add an `else` clause and raise an error.

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,387 @@
...
+        for key in key_list:
+            _, target, _ = _get_submodules(self.model, key)
+            if isinstance(target, LilyLayer):
+                target.delete_adapter(adapter_name)
+                if new_adapter is None:
+                    new_adapter = target.active_adapters[:]
+
+        self.active_adapter = new_adapter or []
+
+    def num_of_layers(self, model, adapter_name):
```
BenjaminBossan:
```suggestion
    def _get_num_of_layers(self, model, adapter_name):
```

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,387 @@
...
+        for target in num_layers.keys():
+            stride[target] = {}
+            for shape in num_layers[target].keys():
+                assert num_layers[target][shape] % peft_config.num_A == 0, f"num of layers {num_layers[target][shape]} for target {target} with shape {shape} is not divisible by num_A {peft_config.num_A}"
+                stride[target][shape] = num_layers[target][shape] // peft_config.num_A
+
+        import logging
+
+        # some logging for information
+        logging.basicConfig(level=logging.INFO)
```
BenjaminBossan: Let's remove all the logging code completely.

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,387 @@
...
+                if key.endswith(one_target_key):
+                    # print(f"find {key} matching {one_target_key}")
+                    _, target, _ = _get_submodules(model, key)
+                    if target.weight.shape not in counter[one_target_key]:
+                        counter[one_target_key][target.weight.shape] = 0
+                    counter[one_target_key][target.weight.shape] += 1
+
+        return counter
+        
+    def inject_adapter(self, model: nn.Module, adapter_name: str,
```
BenjaminBossan: I haven't checked the details, but IIUC, this is code to allow sharing of parameters. I think it shouldn't be necessary to override `inject_adapter`. We have other PEFT methods that allow parameter sharing, we should be able to use the same approach. For instance, check out how it's done in VeRA:

https://github.com/huggingface/peft/blob/8f73327fb967ab47b83aac604bbc2e193207d0f4/src/peft/tuners/vera/model.py#L101-L154

BenjaminBossan:
@yibozhong Once you're finished with your changes, please ping me.

yibozhong:
Hi @BenjaminBossan I have finished my changes to the impl based on your comments. I summarized important changes below:

- raise error when `merge` / `unmerge` are called  
- remove logging  
- improve the doc for config class  
- refactor the model impl similar to VeRA (i.e., use `pre_injection_hook` to precompute `A`s and `B`s to avoid overriding `inject_adapters`)  
- remove `lily_dropout` since it is not used  

Regarding the `lily_scaling` parameter (now renamed to `scaling`), the rationale is that Lily does not use an `alpha / r` formulation for scaling. Instead, `scaling` is defined as a direct multiplier, which makes it straightforward to sweep over values on a log scale (e.g., `0.01`, `0.1`, `1.0`, `10.0`). This is how we conduct experiments in the paper, and some methods (e.g., fourierft) follow the same practice.

I have added several tests. All tests except those requiring `merge` / `unmerge` are currently passing.

src/peft/tuners/lily/config.py
```diff
@@ -0,0 +1,158 @@
...
+        num_B (`int`):
+            The number of shared B adapters. Unlike A adapters (which are grouped by layer),
+            all B adapters are shared globally across every layer. For each forward pass, a router
+            computes a weighted combination of all ``num_B`` B adapters (using softmax-normalized
+            weights) to produce a single combined B adapter, which then projects the low-rank
+            representation back to the original dimension. It is recommended to set
+            ``num_B`` equal to ``num_A``. Suggested values: ``total_layers / 2``, ``total_layers / 3``, or
+            ``total_layers / 4``. Similar to ``num_A``, prefer smaller ``num_B`` with larger ``r``
+            over larger ``num_B`` with smaller ``r``.
+            NOTE: to train router, you need at least 2 B adapters (i.e. ``num_B >= 2``), since the router learns to compute a weighted combination of the B adapters.
```
BenjaminBossan: Does it ever make sense not to train the router? If not, I would check this constraint in `__post_init__` and raise an error. A corresponding unit tests should be added to `test_initialization.py`.

src/peft/tuners/lily/config.py
```diff
@@ -0,0 +1,158 @@
...
+
+@dataclass
+class LilyConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`LilyModel`].
+
+    Args:
+        r (`int`):
+            Lily's rank. Determines the inner hidden dimension of each adapter and the rank of the
+            weight update ``A @ B``. In Lily, since the number of adapters is typically smaller than in LoRA,
```
BenjaminBossan: In general, please use a single backtick for code, not double backticks (markdown and not rst syntax).

src/peft/tuners/lily/config.py
```diff
@@ -0,0 +1,158 @@
...
+            called `'layers'` or `'h'`.
+        init_weights (`bool`):
+            Whether to initialize Lily adapter weights using the default initialization scheme: A
+            matrices are initialized with Kaiming uniform, and B matrices are initialized to zero,
+            ensuring that the adapter output is zero at the start of training and does not disturb
+            the pretrained model. It is strongly recommended to keep this as ``True`` unless you have
+            a specific reason to change it.
+    """
+
+    r: int = field(default=4, metadata={"help": "Lily's rank"})
```
BenjaminBossan: Please extend the help text here. It's easiest to just copy the description from the docstring.

Also, does a default rank of 4 make sense? From the description, I would expect something like 32 to be a more reasonable default.

src/peft/tuners/lily/config.py
```diff
@@ -0,0 +1,158 @@
...
+    scaling: float = field(default=1.0, metadata={"help": "scaling factor for lily"})
+    num_A: int = field(default=1, metadata={"help": "Lily's number of adapter A"})
+    num_B: int = field(default=1, metadata={"help": "Lily's number of adapter B"})
```
BenjaminBossan: Is 1 a reasonable default for those two arguments?

src/peft/tuners/lily/layer.py
BenjaminBossan: Let's remove the `x` from the signature.

src/peft/tuners/lily/layer.py
```diff
@@ -0,0 +1,189 @@
...
+        lily_A_keys = self.lily_A.keys()
+
+        for active_adapter in self.active_adapters:
+            if active_adapter not in lily_A_keys:
+                continue
+
+            lily_A = self.lily_A[active_adapter]
+            lily_B = self.lily_B[active_adapter]
+            router = self.lily_router[active_adapter]
+            num_B = self.num_B[active_adapter] if isinstance(self.num_B, dict) else self.num_B
```
BenjaminBossan: `num_B` is always a dict, is it not?

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,259 @@
...
+        - **model** ([`~transformers.PreTrainedModel`]) -- The model to be adapted.
+        - **peft_config** ([`LilyConfig`]): The configuration of the Lily model.
+    """
+
+    prefix: str = "lily_"
+    tuner_layer_cls = LilyLayer
+    target_module_mapping = TRANSFORMERS_MODELS_TO_LILY_TARGET_MODULES_MAPPING
+
+    @staticmethod
+    def _check_target_module_exists(lily_config, key):
```
BenjaminBossan: Can be removed.

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,259 @@
...
+        return check_target_module_exists(lily_config, key)
+
+    def _pre_injection_hook(self, model: nn.Module, config: LilyConfig, adapter_name: str) -> None:
+        """
+        Pre-compute and cache all shared A and B matrices before adapter injection begins.
+
+        A matrices are shared across consecutive groups of layers:
+          stride = total_layers_per_shape / num_A
+        B matrices are shared globally across all layers (one per target-module/shape pair).
```
BenjaminBossan: I think it would make sense to have some extra unit tests to ensure that this parameter sharing works as expected. You can add those tests to `test_initialization.py`. It would make sense to test this with a few edge cases (e.g. `num_A = 1`, `num_A < num_layers`, `num_A > num_layers` etc.).

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,259 @@
...
+
+        All caches are keyed by adapter_name first, so multiple adapters can coexist
+        without overwriting each other's matrices or counters.
+        """
+        model_config = self.get_model_config(model)
+        peft_config = self._prepare_adapter_config(config, model_config)
```
BenjaminBossan: Let's add `peft_config = _maybe_include_all_linear_layers(peft_config, self.model)` like in VeRA.

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,259 @@
...
+
+        # Initialize top-level dicts on first call; keyed as [adapter_name][target][shape]
+        if not hasattr(self, "_lily_As"):
+            self._lily_As: dict[str, dict[str, dict[torch.Size, list[nn.Linear]]]] = {}
```
BenjaminBossan: Hmm, I think `_lily_As` and `_lily_Bs` should be (nested) `BufferDict`s, like in VeRA, right? Otherwise, something like `model.to(device)` will not affect these tensors. Also, should the buffers be persistent? IIUC, we need them if we load a trained model, so they should be.

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,259 @@
...
+            self._lily_strides: dict[str, dict[str, dict[torch.Size, int]]] = {}
+
+        # Reset all sub-dicts for the current adapter to avoid stale state
+        self._lily_As[adapter_name] = {}
+        self._lily_Bs[adapter_name] = {}
+        self._lily_counters[adapter_name] = {}
+        self._lily_strides[adapter_name] = {}
+
+        # Count how many layers match each (target_module, weight_shape) pair
+        num_layers: dict[str, dict[torch.Size, int]] = {}
```
BenjaminBossan: Hmm, I find the name confusing, as this isn't really a count of layers, is it? It's more a layer key to shape mapping or something like this. Could you please find a better name?

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,259 @@
...
+        if isinstance(peft_config.target_modules, str):
+            target_modules_iter = [peft_config.target_modules]
```
BenjaminBossan: Note that if `target_modules` is a string, it is treated as a regex to match the lay name. I think that should be fine with the logic that follows, but I just wanted to flag this in case it's problematic.

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,259 @@
...
+        else:
+            target_modules_iter = list(peft_config.target_modules)
+
+        for target_key in target_modules_iter:
+            num_layers[target_key] = {}
+
+        for key, module in model.named_modules():
```
BenjaminBossan: I think this whole approach is very brittle. When we create a PEFT model, the config is used to determine which layers should be adapted. Here, we make the assumption that it's sufficient to check `config.target_modules`. In reality, there are other arguments that can affect the output, e.g. `exclude_modules` and `layers_to_transform`. Now we could try to take those into account here, but in the end it means that we have to duplicate the whole matching logic here (and keep it in sync each time there is a change). This is very error prone.

So what could be an alternative. I'm thinking about actually just running the whole injection logic and _afterwards_, once we know what layers were adapted, we can run the Lily logic for shared parameters. This would be a little wasteful, as we would first create a bunch of distinct parameters and afterwards replace them with shared parameters. But I think in the grand scheme of things, this is a negligible overhead.

To implement this, we would need to add a new hook, `_post_injection_hook`, which is run after `inject_adapter`. This would be added to the `BaseTuner` class and by default we would just `pass`. For Lily, we would add a custom implementation containing the logic below, only that we now know for sure what layers were adapted. WDYT?

src/peft/tuners/lily/config.py
```diff
@@ -0,0 +1,158 @@
...
+        num_A (`int`):
+            The number of shared A adapters. Lily groups the model's layers into ``num_A`` consecutive
+            blocks, where each block of ``total_layers / num_A`` adjacent layers shares one A adapter.
+            The A adapter compresses the input into a low-rank representation of size ``r``.
+            For best results, ``num_A`` should evenly divide the
+            total number of layers in your model. It is recommended that ``num_A`` should ideally
+            evenly divide the total number of layers.
+            Suggested values: ``total_layers / 2``, ``total_layers / 3``, or ``total_layers / 4``.
+            Keeping ``num_A`` small and increasing ``r`` instead leads to
+            better performance than the opposite trade-off (large ``num_A``, small ``r``).
```
BenjaminBossan: I wonder if it would be more user friendly to have a parameter `block_size_A`. If I pass, say, `block_size_A=4`, it would create `total_layers / 4` distinct A's, each with blocks of 4. IMO this is better than `num_A` as it doesn't require me as a user to check the number of layers beforehand and adjust `num_A` accordingly. WDYT?

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,259 @@
...
+                num_B=lily_config.num_B,
+                init_weights=lily_config.init_weights,
+            )
+        else:
+            new_module = self._create_new_module(lily_config, adapter_name, target, lily_A, lily_B, **kwargs)
+            if adapter_name not in self.active_adapters:
+                new_module.requires_grad_(False)
+            self._replace_module(parent, target_name, new_module, target)
+
+    def _replace_module(self, parent, child_name, new_module, child):
```
BenjaminBossan: We can just use this method from the parent class, right?

yibozhong:
Hi @BenjaminBossan Thanks for the comments! I made some changes including:

- Style checking for `src/peft/tuners/lily`.
- Copy help text for each parameter from the doc into `config.py`.
- Discard `num_A`, and use `stride_A` to represent the number of consecutive layers that share the same `A` adapter. This is more user-friendly.
- For `model.py`, I think your idea is great and I have migrated the code from using `pre_injection_hook` to `post_injection_hook`. Specifically, I first add a `_post_injection_hook` method to the `BaseTuner` class and then override it in `LilyModel` to implement the custom adapter sharing logic. During adapter injection, every layer has its own distinct `A` and `B`, and the sharing happens afterwards via `_post_injection_hook`.
- The default of `stride_A` is set to 1 (i.e., no A sharing) and `num_B` default set to 2. Default value for `r` is set to 32. We also check `stride_A` and `num_B` in `__post_init__`.
- Skip merge/unmerge related tests in `tests/test_custom_models.py`. Now all tests have passed.
- Add initialization-related tests in `tests/test_initialization.py`. This includes cases where `stride_A = 1`, `stride_A = num_layers`, and `stride_A > num_layers`, etc. All tests have passed.

Please let me know what you think.

src/peft/tuners/lily/model.py
```diff
@@ -0,0 +1,169 @@
...
+            if not isinstance(module, LilyLayer):
+                continue
+            if adapter_name not in module.lily_A:
+                continue
+
+            base = module.get_base_layer()
+            shape = base.weight.shape  # (out_features, in_features)
+
+            # Find the longest matching target suffix for this key
+            matched_suffix = None
```
BenjaminBossan, line 140: For my own understanding: The matched suffix exists so that parameter sharing only happens between layers that not only share the size, but also the name? So e.g. if we have `target_modules=["q_proj", "v_proj"]`, even if the two layers had the same shape, we would only share `q_proj` with `q_proj` and `v_proj` with `v_proj`, but not between `q_proj` and `v_proj`?

tests/test_initialization.py
```diff
@@ -3628,6 +3629,242 @@ def test_lora_config_raises_warning_with_eva_config_but_not_eva_init(self):
...
             LoraConfig(init_lora_weights=True, eva_config=EvaConfig())
 
+class TestLilyInitialization:
+    """Tests for Lily adapter initialization and parameter sharing.
+
+    The model uses a two-level structure: self.blocks is a ModuleList of
+    simple Block wrappers, each containing a single Linear named "lin".
+    This way all target Linear layers share the same suffix name ("lin"),
+    which is required for _post_injection_hook to group them and apply
+    A/B sharing correctly, while avoiding targeting the ModuleList itself.
```
BenjaminBossan: I think the test could be made more strict by having one additional module per `Block` with a different name than `lin`. Then we target both and ensure that sharing only happens within `lin` or the other layer (if my understanding is correct).

tests/test_initialization.py
```diff
@@ -3628,6 +3629,242 @@ def test_lora_config_raises_warning_with_eva_config_but_not_eva_init(self):
...
+        return MLP(num_layers).to(self.torch_device).eval()
+
+    def _get_lily_layers(self, model, adapter_name="default"):
+        """Return all LilyLayer modules in traversal order."""
+        from peft.tuners.lily.layer import LilyLayer
+
+        return [m for _, m in model.named_modules() if isinstance(m, LilyLayer) and adapter_name in m.lily_A]
+
+    # ------------------------------------------------------------------
+    # stride_A = 1: every layer has its own A (no A sharing)
```
BenjaminBossan: Let's remove all these comments. The information is redundant with the test function docstring.

tests/test_initialization.py
```diff
@@ -3628,6 +3629,242 @@ def test_lora_config_raises_warning_with_eva_config_but_not_eva_init(self):
...
+            target_modules=["lin"],
+            r=8,
+            stride_A=1,
+            num_B=2,
+        )
+        model = get_peft_model(model, config)
+        layers = self._get_lily_layers(model)
+        assert len(layers) == num_layers
+
+        A_ids = [id(layer.lily_A["default"]) for layer in layers]
```
BenjaminBossan: I wonder if using the `data_ptr()` wouldn't be the better test instead of `id`:

```python
from torch import nn

lin0 = nn.Linear(3, 4)
lin1 = nn.Linear(3, 4)
lin1.weight.data = lin0.weight.data

id(lin0) == id(lin1)  # False
lin0.weight.data_ptr() == lin1.weight.data_ptr()  # True
```

tests/test_initialization.py
```diff
@@ -3628,6 +3629,242 @@ def test_lora_config_raises_warning_with_eva_config_but_not_eva_init(self):
...
+        assert len(set(A_ids)) == 1, "stride_A > num_layers should still result in a single shared A"
+
+    # ------------------------------------------------------------------
+    # B sharing: all layers always share one B regardless of stride_A
+    # ------------------------------------------------------------------
+    def test_B_always_shared_across_all_layers(self):
+        """All layers should always share the same B adapter object, regardless of stride_A."""
+        num_layers = 6
+        for stride_A in [1, 2, 3, 6]:
```
BenjaminBossan: Instead of looping inside the test, use `@pytest.mark.parametrize`.

tests/test_initialization.py
```diff
@@ -3628,6 +3629,242 @@ def test_lora_config_raises_warning_with_eva_config_but_not_eva_init(self):
...
+
+            B_ids = [id(layer.lily_B["default"]) for layer in layers]
+            assert len(set(B_ids)) == 1, (
+                f"All layers should share one B adapter (stride_A={stride_A}), got {len(set(B_ids))} distinct"
+            )
+
+    # ------------------------------------------------------------------
+    # init_weights=True: output unchanged from base model
+    # ------------------------------------------------------------------
+    def test_init_weights_true_is_noop(self):
```
BenjaminBossan: Can be removed, it's covered by other tests.

tests/test_initialization.py
```diff
@@ -3628,6 +3629,242 @@ def test_lora_config_raises_warning_with_eva_config_but_not_eva_init(self):
...
+        output_after = peft_model(x)
+
+        assert torch.allclose(output_base, output_after, atol=1e-6), (
+            "init_weights=True should produce a no-op adapter at initialization"
+        )
+
+    # ------------------------------------------------------------------
+    # init_weights=False: output should differ from base model
+    # ------------------------------------------------------------------
+    def test_init_weights_false_shifts_output(self):
```
BenjaminBossan: Can be removed, it's covered by other tests.

yibozhong:
Hi @BenjaminBossan ,

I have updated the testing code as you mentioned. All tests are now passing. The sharing logic indeed happens only within the respective target module (that is to say, `q_proj` adapters are shared only among themselves, not with `v_proj` adapters). I have also added a folder `examples/lily_finetuning` to provide an example for Lily.

About the MetaMathQA benchmark, I have run several experiments (including the respective LoRA baseline for comparison) on my hardware (an A100). I provide the results below and have picked 5 experiment settings of Lily that work best in each table to be added to `method_comparison`. I include 3 tables using 9M params and 2 tables using 18M params. All experiments use the default training settings. `lily s=x` indicates the scaling factor is set to x (where x is taken from `[2, 4, 8]`). LoRA's alpha is always set to `2 * rank`, as in the default setting.

## 9M params

I first test the standard setting, where target modules are q and v projections. `lily s=8.0` with a rank of 448 works best in this case.

| method     | target modules | rank | num adapters | param | accuracy            |
| ---------- | -------------- | ---- | ------------ | ----- | ------------------- |
| lora       | qv             | 32   | 28           | 9M    | 0.4836997725549659  |
| lily s=2.0 | qv             | 128  | 7            | 9M    | 0.46929492039423804 |
| lily s=4.0 | qv             | 128  | 7            | 9M    | 0.4852160727824109  |
| lily s=8.0 | qv             | 128  | 7            | 9M    | 0.4935557240333586  |
| lily s=2.0 | qv             | 224  | 4            | 9M    | 0.488248673237301   |
| lily s=4.0 | qv             | 224  | 4            | 9M    | 0.4943138741470811  |
| lily s=8.0 | qv             | 224  | 4            | 9M    | 0.49583017437452614 |
| lily s=2.0 | qv             | 448  | 2            | 9M    | 0.5003790750568613  |
| lily s=4.0 | qv             | 448  | 2            | 9M    | 0.5064442759666414  |
| lily s=8.0 | qv             | 448  | 2            | 9M    | 0.5094768764215315  |

Then I test the settings where target modules are the MLP projections. `lily s=8.0` with a rank of 140 works best in this case.

| method     | target modules | rank | num adapters | param | accuracy           |
| ---------- | -------------- | ---- | ------------ | ----- | ------------------ |
| lora       | mlp            | 10   | 28           | 9M    | 0.5276724791508719 |
| lily s=2.0 | mlp            | 140  | 2            | 9M    | 0.5284306292645944 |
| lily s=4.0 | mlp            | 140  | 2            | 9M    | 0.5261561789234268 |
| lily s=8.0 | mlp            | 140  | 2            | 9M    | 0.5413191811978771 |


Finally, I am curious about the performance when updating all projection layers (including q, k, v, o projections and the MLP projections, denoted as `all`). `lily s=4.0` works best in this case.

| method     | target modules | rank | num adapters | param | accuracy           |
| ---------- | -------------- | ---- | ------------ | ----- | ------------------ |
| lora       | all            | 6    | 28           | 9M    | 0.5276724791508719 |
| lily s=2.0 | all            | 84   | 2            | 9M    | 0.5375284306292646 |
| lily s=4.0 | all            | 84   | 2            | 9M    | 0.5481425322213799 |
| lily s=8.0 | all            | 84   | 2            | 9M    | 0.5344958301743745 |

## 18M params

In this setting, the rank of Lily adapters can be higher.

I first test qv projections as the target modules. `lily s=2.0` works best in this case.

| method     | target modules | rank | num adapters | param | accuracy            |
| ---------- | -------------- | ---- | ------------ | ----- | ------------------- |
| lora       | qv             | 64   | 28           | 18M   | 0.49583017437452614 |
| lily s=2.0 | qv             | 896  | 2            | 18M   | 0.514783927217589   |
| lily s=4.0 | qv             | 896  | 2            | 18M   | 0.5125094768764216  |
| lily s=8.0 | qv             | 896  | 2            | 18M   | 0.4965883244882487  |

Then, with all projections as the target modules, `lily s=4.0` works best in this case.

| method     | target modules | rank | num adapters | param | accuracy           |
| ---------- | -------------- | ---- | ------------ | ----- | ------------------ |
| lora       | all            | 12   | 28           | 18M   | 0.5405610310841547 |
| lily s=2.0 | all            | 168  | 2            | 18M   | 0.5519332827899924 |
| lily s=4.0 | all            | 168  | 2            | 18M   | 0.5678544351781653 |
| lily s=8.0 | all            | 168  | 2            | 18M   | 0.5329795299469295 |

src/peft/tuners/lily/model.py
yibozhong: Yes, that's right

BenjaminBossan:
Thanks for running the experiments. I would suggest to submit one setting which basically corresponds to the defaults and another more optimized one that gets good performance. I'd prefer not to target all layers but either attention or MLP layers, simply because that's what we do in the other experiments.

Furthermore, there is now a merge conflict because another PEFT method was merged, but it should be easy to fix. Could you please take care?

yibozhong:
Hi @BenjaminBossan I have removed the settings adapting all the projections. I left out two settings, one targeting qv projections as the default setting and another one targeting mlp as a more performant one. The merge conflicts are also resolved.

src/peft/tuners/lily/config.py
```diff
@@ -0,0 +1,221 @@
...
+    Args:
+        r (`int`):
+            Lily's rank. Determines the inner hidden dimension of each adapter and the rank of the weight update `A @
+            B`. In Lily, since the number of adapters is typically smaller than in LoRA, each adapter needs to carry
+            more capacity, so it is recommended to use a larger `r` than in LoRA — typically `2x`, `3x`, or `4x` the
+            LoRA rank you would normally use. The total number of trainable parameters scales with `r * (total_layers /
+            stride_A + num_B)`, so increasing `r` while keeping `stride_A` large and `num_B` small is the recommended
+            trade-off.
+        stride_A (`int`):
+            The number of consecutive layers that share one A adapter. For example, if `stride_A=4`, every 4 adjacent
```
BenjaminBossan, line 37: Here it should mention that only layers whose name share the same suffix share parameters (I would use something like `"q_proj"` and `"v_proj"` as clear examples).

yibozhong: updated now.

yibozhong:
Hi @BenjaminBossan Thanks for the review and suggestions! I have added doc for Lily and checked the code style of the implementation.

BenjaminBossan:
@yibozhong Could you please run `make style`?

yibozhong:
@BenjaminBossan I did run `make style`, but I only committed changes in `src/peft/tuners/lily`. There are also code changes that are irrelevant to my integration of Lily, so should I commit these changes as well (e.g., changes in the test scripts or files like `src/peft/__init__.py`)?

BenjaminBossan:
@yibozhong Check the log, the changes are relating to this PR: https://github.com/huggingface/peft/actions/runs/22609474448/job/65529255818?pr=3036. Could it be that your local settings somehow diverge?

yibozhong:
I committed all the relevant changes after running `make style`. Please let me know if there are any issues now.

src/peft/tuners/tuners_utils.py
```diff
@@ -174,11 +174,7 @@ def _get_in_out_features(module: nn.Module) -> tuple[int, int] | tuple[None, Non
             out_features, in_features = module.weight.to_local().shape
         else:
             in_features, out_features = module.in_features, module.out_features
-    elif isinstance(module, nn.Conv1d):
-        in_features, out_features = module.in_channels, module.out_channels
-    elif isinstance(module, nn.Conv2d):
-        in_features, out_features = module.in_channels, module.out_channels
-    elif isinstance(module, nn.Conv3d):
+    elif isinstance(module, nn.Conv1d) or isinstance(module, nn.Conv2d) or isinstance(module, nn.Conv3d):
```
BenjaminBossan: Could you please undo this unrelated change? It shouldn't be necessary for the CI to pass.

src/peft/tuners/tuners_utils.py
```diff
@@ -339,7 +337,20 @@ def _pre_injection_hook(self, model: nn.Module, config: PeftConfig, adapter_name
...
+        classes to perform any post-injection operations.
+
+        Args:
+            model (`nn.Module`):
+                The model to be adapted.
+            config (`PeftConfig`):
+                The adapter config.
+            adapter_name (`str`):
+                The adapter name.
+        """
```
BenjaminBossan, line 358: Could you please add back the `pass`?

src/peft/tuners/tuners_utils.py
```diff
@@ -381,7 +392,6 @@ def _prepare_model(self, peft_config: PeftConfig, model: nn.Module):
             model (`nn.Module`):
                 The model that is going to be adapted.
         """
-        pass
```
BenjaminBossan, line 384: Could you please add back the `pass`?

src/peft/tuners/tuners_utils.py
```diff
@@ -1760,9 +1770,9 @@ def check_target_module_exists(config, key: str) -> bool | re.Match[str] | None:
         if isinstance(config.exclude_modules, str):
             if re.fullmatch(config.exclude_modules, key):
                 return _ExcludedModule()
-        elif key in config.exclude_modules:
-            return _ExcludedModule()
-        elif any(key.endswith(f".{exclude_key}") for exclude_key in config.exclude_modules):
+        elif key in config.exclude_modules or any(
```
BenjaminBossan: Please undo the unrelated change.

tests/testing_common.py
```diff
@@ -1222,9 +1220,9 @@ def _test_training_gradient_checkpointing(self, model_id, config_cls, config_kwa
                         "delta_embedding" in n
                     ):  # delta_embedding is the embedding that should be updated with grads in CPT
                         assert param.grad is not None
-                elif hasattr(model, "prefix") and (model.prefix in n):  # non-prompt tuning methods
-                    assert param.grad is not None
-                elif "trainable_tokens_" in n:  # trainable tokens layer
+                elif (
```
BenjaminBossan: Could you please undo this unrelated change?

yibozhong:
I checked my ruff version and it indeed is different. I have undone the mentioned changes. Please let me know if there are any issues.

yibozhong:
Added back mapping in `src/peft/utils/other.py` to solve the issue

yibozhong:
Removed einops related code due to dependency issue

yibozhong:
Use larger lr for Lily for `test_only_params_are_updated`. My local tests all passed but the CI tests failed, so maybe this will solve the problem.

yibozhong:
@BenjaminBossan All checks are passing now.

BenjaminBossan:
@yibozhong I just wanted to let you know that we released v0.19.0, which includes Lily, and that Lily is now the top performing model in terms of test accuracy on the PEFT MetaMath benchmark: https://huggingface.co/spaces/peft-internal-testing/PEFT-method-comparison. It's not quite as memory efficient as some other methods, but it's still an impressive result.

yibozhong:
@BenjaminBossan Thanks so much for the information! Glad to see that the results are largely reproducible. I may try to explore improvements to the method and will actively integrate them into PEFT once they work.

Issue #3026, opened 2026-02-07
Add PSOFT to PEFT


fei407:
### Feature request

Hi, I’ve been a long-time user of PEFT and really appreciate the work behind the library. I would like to propose adding **PSOFT** (Orthogonal Fine-Tuning with Principal Subspace) to the PEFT family as a new OFT-based method.  Our paper has been accepted to ICLR 2026: https://openreview.net/forum?id=FSHrinMArK

PSOFT aims to **preserve the geometric relationships** among pre-trained weight column vectors—a core principle of OFT—while achieving a balanced trade-off across parameter, computation, and memory efficiency. Unlike existing OFT variants (e.g., _OFTv2_, _BOFT_, and _GOFT_) that rely on sparsity-based designs, PSOFT adopts **a low-rank principal subspace perspective**, bridging the gap between LoRA and OFT.

PSOFT confines orthogonal fine-tuning to a principal subspace, offering **theoretical guarantees** via orthogonality constraints on the down-projection matrix, while enabling **practical adaptability** through two low-dimensional tunable vectors.

An initial open-source implementation of PSOFT is available at https://github.com/fei407/PSOFT, adapted from LoRA-XS, OFT, and PiSSA. I plan to refactor the code to align with PEFT’s abstractions and coding conventions for a clean integration, and would welcome any guidance from the maintainers.

### Your contribution

I am happy to implement PSOFT in the PEFT framework and submit a PR, and I am willing to iterate on the implementation based on maintainers’ feedback.

fei407:
PSOFT has been added to PEFT.

PR #3037, opened 2026-02-11
Add PSOFT tuner implementation

fei407:
## What does this PR do?

This PR implements **PSOFT** as a new PEFT tuner (related to issue #3026).

PSOFT is a parameter-efficient fine-tuning method that combines low-rank decomposition with orthogonal fine-tuning. It applies an r*r orthogonal transformation within the principal (low-rank) subspace of the pretrained weight matrix, while keeping the remaining components frozen.

By restricting orthogonal updates to a principal subspace, PSOFT aims to preserve the structural properties of pretrained weights while maintaining efficiency in terms of parameter count, memory, and computation. Compared to sparsity-based OFT variants (e.g., OFTv2, BOFT, GOFT), PSOFT adopts a low-rank principal subspace formulation and can be viewed as introducing a structured orthogonal transformation within a LoRA-style decomposition.

For additional theoretical and empirical details, please refer to the paper (https://openreview.net/forum?id=FSHrinMArK) linked in issue #3026. 

### New modules

- `peft/tuners/psoft/__init__.py`
- `peft/tuners/psoft/config.py`
- `peft/tuners/psoft/layer.py`
- `peft/tuners/psoft/model.py`

### Registry / auto integration

- Integrates PSOFT into PEFT auto/registry so it can be used via `get_peft_model(...)` with `PSOFTConfig`.

### Documentation

- Adds API docs page: `docs/source/package_reference/psoft.md`

## Tests

- Adds PSOFT test coverage and updates existing test suites where needed:

- New: `tests/test_psoft.py`

- Updated: `tests/test_custom_models.py`

- Updated: `tests/test_decoder_models.py`

- Updated: `tests/test_encoder_decoder_models.py`

- Updated: `tests/test_feature_extraction_models.py`

We also conducted additional experiments for NLU and Math tasks using https://github.com/fei407/test_psoft to test the correctness of the PSOFT Tuner.

### DIFF
--- a/src/peft/tuners/psoft/layer.py
+++ b/src/peft/tuners/psoft/layer.py
@@ -125,7 +125,7 @@ class OrthLayer(nn.Module):
     def get_matrix(self) -> torch.Tensor:
         cast_to_fp32 = False
         orig_dtype = None
-        
+
         if not self.orth:
             R = self.weight
         else:
@@ -159,7 +159,7 @@ class OrthLayer(nn.Module):
             else:
                 R = torch.linalg.solve(id_mat - Q, id_mat + Q, left=False)
 
-        # Apply scaling vectors to R 
+        # Apply scaling vectors to R
         if self.vector_b is not None:
             R = self.vector_b[:, None] * R
         if self.vector_a is not None:


fei407:
Thank you for the review and the helpful questions!
1. The paper has already been accepted to ICLR 2026. The current OpenReview version is essentially aligned with the camera-ready version, so no major changes are expected that would affect the implementation.
2. Thanks for the suggestion. I’ve reached out to the OFT authors (@zqiu24 and @wy1iu), given their experience integrating orthogonal fine-tuning into PEFT. I’ve invited them to review the implementation and share their feedback.
3. Thank you for reminding this issue. I’ll clean up the unrelated changes and ensure that only minimal, PSOFT-specific modifications remain without affecting other files.

BenjaminBossan:
> * The paper has already been accepted to ICLR 2026. The current OpenReview version is essentially aligned with the camera-ready version, so no major changes are expected that would affect the implementation.

Great, congratulations, this is good to know.

> * Thanks for the suggestion. I’ve reached out to the OFT authors (@zqiu24 and @wy1iu), given their experience integrating orthogonal fine-tuning into PEFT. I’ve invited them to review the implementation and share their feedback.

:+1: 

> * Thank you for reminding this issue. I’ll clean up the unrelated changes and ensure that only minimal, PSOFT-specific modifications remain without affecting other files.

Thanks. Please ping me once the PR is ready for review.

fei407:
Thank you again for your previous feedback and guidance. I’ve cleaned up the unrelated changes and further refined the config descriptions and documentation to better align with the existing PEFT tuners.

I believe the PR is now in a good state for review. I’m very happy to make any further adjustments based on your suggestions.

tests/test_psoft.py
BenjaminBossan, line 1: I think all these tests should already be covered by existing tests. Did you check if they are truly necessary? The only test that is not covered is `test_psoft_column_angle_invariance_after_training` but I would basically consider this an implementation detail. Do you think it's important to test this? If yes, we can move it elsewhere, there is no need for an extra file just for one test.

docs/source/package_reference/psoft.md
BenjaminBossan, line 1: We need an entry in `docs/_toctree.yml` or else this page won't be reachable.

src/peft/tuners/psoft/config.py
```diff
@@ -0,0 +1,276 @@
...
+
+from dataclasses import dataclass, field
+from typing import Literal, Optional, Union
+
+from peft.config import PeftConfig
+from peft.utils import PeftType
+
+
+@dataclass
+class PSOFTConfig(PeftConfig):
```
BenjaminBossan: I would suggest to rename these classes to `PsoftConfig`, `PsoftLayer` etc. Not only is it easier to type, it's also more consistent with the rest of the codebase (e.g. `LoraConfig` instead of `LoRAConfig`). I know that for OFT, we have `OFTConfig` but with hindsight, it should have been `OftConfig`.

src/peft/tuners/psoft/config.py
```diff
@@ -0,0 +1,276 @@
...
+        default=None,
+        metadata={"help": "List of module names or regex expression of the module names to exclude from PSOFT. "},
+    )
+    psoft_alpha: int = field(default=8, metadata={"help": "It controls PSOFT scaling factor. Same semantics as LoRA alpha. "})
+    psoft_dropout: float = field(default=0.0, metadata={"help": "Dropout for PSOFT path. Same semantics as LoRA dropout. "})
+    fan_in_fan_out: bool = field(
+        default=False,
+        metadata={"help": "Set this to True if the layer to replace stores weight like (fan_in, fan_out). "},
+    )
+    init_psoft_weights: Literal["psoft_init", "pissa_init"] = field(
```
BenjaminBossan: Hmm, I wonder if there could be a better name for this argument to differentiate it from `init_weights`. As is, from the name, it's hard to tell what different role these arguments have. Maybe something that makes it clear that the SVD method for A and B is meant.

src/peft/tuners/psoft/config.py
```diff
@@ -0,0 +1,276 @@
...
+            `Conv1D` which stores weights like (fan_in, fan_out) and hence this should be set to `True`.
+        init_psoft_weights (`Literal["psoft_init", "pissa_init"]`):
+            Initialization strategy for A and B used to construct the principal subspace in PSOFT.
+            'psoft_init': SVD-based initialization with row-orthogonal A, ensuring strict orthogonality (PSOFT).
+            'pissa_init': SVD-based initialization with symmetric A and B (standard PiSSA).
+        psoft_svd (`Literal["full", "lowrank"]`):
+            SVD backend for initialization: 'full' uses torch.linalg.svd; 'lowrank' uses torch.svd_lowrank.
+        psoft_svd_lowrank_niter (`int`): 
+            Number of power iterations used by torch.svd_lowrank when psoft_svd='lowrank'.
+            Only used when psoft_svd='lowrank'.
```
BenjaminBossan: This constraint could be checked in the `__post_init__`, WDYT? If you agree, let's add a test for this to `test_initialization.py`.

src/peft/tuners/psoft/config.py
```diff
@@ -0,0 +1,276 @@
...
+from typing import Literal, Optional, Union
+
+from peft.config import PeftConfig
+from peft.utils import PeftType
+
+
+@dataclass
+class PSOFTConfig(PeftConfig):
+    """
+    Configuration for PSOFT (Efficient Orthogonal Fine-Tuning with Principal Subspace Adaptation).
```
BenjaminBossan, line 28: Generally, well written docstring. Only one thing: For all arguments with defaults, let's mention the default value.

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,543 @@
...
+
+        if adapter_name not in self._psoft_cache_names:
+            self.register_buffer(a_name, A, persistent=False)
+            self.register_buffer(b_name, B, persistent=False)
+            self._psoft_cache_names[adapter_name] = (a_name, b_name)
+        else:
+            setattr(self, a_name, A)
+            setattr(self, b_name, B)
+
+    def _ensure_psoft_cache(self, adapter_name: str, *, allow_skip_on_meta: bool, caller: str) -> bool:
```
BenjaminBossan: The return value is never used, is it? Also, it's not quite clear to me why we need this method, could you please explain? Intuitively, if A and B are fixed, I would think we only need to call this once at initialization time, but that it shouldn't be needed for `get_delta_weight` or `forward`. Is this solely because of meta weights?

src/peft/tuners/psoft/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from peft.utils import register_peft_method
+
+from .config import PSOFTConfig
+from .layer import Linear, OrthLayer, PSOFTLayer
```
BenjaminBossan: Is there a need to expose `OrthLayer` here? I think it can be removed.

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,543 @@
...
+        return re.sub(r"[^0-9a-zA-Z_]", "_", adapter_name)
+
+    def _get_cache_buffers(self, adapter_name: str) -> tuple[Optional[torch.Tensor], Optional[torch.Tensor]]:
+        if adapter_name not in self._psoft_cache_names:
+            return None, None
+        a_name, b_name = self._psoft_cache_names[adapter_name]
+        return getattr(self, a_name, None), getattr(self, b_name, None)
+
+    def _set_cache_buffers(self, adapter_name: str, A: torch.Tensor, B: torch.Tensor) -> None:
+        safe = self._sanitize_adapter_name(adapter_name)
```
BenjaminBossan: Hypothetically, this could lead to errors if the sanitized names of two different adapters clash, right? Instead, we could also check, when a new adapter is added, if the name already corresponds to the sanitized name, and if it doesn't, raise an error. This moves the responsibility to the user instead of trying to silently fix their name choice. If you agree with this, please add a unit test to `test_initialization.py`.

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,543 @@
...
+        a_name = f"_psoft_A_cache_{safe}"
+        b_name = f"_psoft_B_cache_{safe}"
+
+        if adapter_name not in self._psoft_cache_names:
+            self.register_buffer(a_name, A, persistent=False)
```
BenjaminBossan: Wondering: Do we really need to set the attribute directly on the module (e.g. `self._psoft_A_cache_default = A`)? Could we not use: `self._psoft_A_cache = BufferDict(); self._psoft_A_cache[adapter_name] = A`? We have a `BufferDict` implementation in PEFT for use cases like this.

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,543 @@
...
+            size=r,
+            orth=config.psoft_orth,
+            mag_b=config.psoft_mag_b,
+            mag_a=config.psoft_mag_a,
+            use_cayley_neumann=config.use_cayley_neumann,
+            num_cayley_neumann_terms=config.num_cayley_neumann_terms,
+            cayley_neumann_eps=config.cayley_neumann_eps,
+        )
+
+        do_init = getattr(config, "init_weights", True)
```
BenjaminBossan: The attribute should always exist, right?

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,543 @@
...
+        
+        self.set_adapter([adapter_name])
+
+    # Adapted from the asymmetric SVD used in PiSSA
+    # (PEFT implementation: https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/layer.py) #L316
+    def _build_cache_once(self, adapter_name: str, init_type: str) -> None:
+        if self._psoft_cache_built.get(adapter_name, False):
+            return
+
+        with torch.inference_mode(False), torch.no_grad():
```
BenjaminBossan: For better understanding, can you add a comment why `inference_mode` is explicitly disabled?

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,543 @@
...
+        if self.mag_b and self.vector_b is not None:
+            R = self.vector_b[:, None] * R
+        if self.mag_a and self.vector_a is not None:
+            R = R * self.vector_a[None, :]
+
+        if cast_to_fp32:
+            R = R.to(orig_dtype)
+
+        return R
+
```
BenjaminBossan, line 172: Let's add a `__repr__`.

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,543 @@
...
+        R_c = R.to(device=device, dtype=compute_dtype)
+
+        id_mat = torch.eye(r, device=device, dtype=compute_dtype)
+        delta = B_c @ (R_c - id_mat) @ A_c  # (out, in)
+        delta = transpose(delta, self.fan_in_fan_out)
+        delta = delta * self.scaling[adapter_name]
+
+        return delta.to(dtype=out_dtype)
+
+    def merge(self, safe_merge: bool = False, adapter_names: Optional[list[str]] = None) -> None:
```
BenjaminBossan, line 375: As we have `W' = W + delta_weight`, it should be possible to convert this adapter to LoRA, right? So let's add:

```python
    def supports_lora_conversion(self, adapter_name: str = "default") -> bool:
        return True
```

For background info, check: https://huggingface.co/docs/peft/main/en/package_reference/lora_conversion#lora-conversion

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,543 @@
...
+            weight.data -= delta_weight.to(orig_dtype)
+
+    def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
+        if self.disable_adapters:
+            if self.merged:
+                self.unmerge()
+            return self.base_layer(x, *args, **kwargs)
+
+        if self.merged:
+            return self.base_layer(x, *args, **kwargs)
```
BenjaminBossan: For consistency with other PEFT methods, let's not early return here but assign the `result` and return it at the very end.

src/peft/tuners/psoft/model.py
```diff
@@ -0,0 +1,101 @@
...
+
+        device_map = get_device_map(self.model)
+        new_module = self._create_new_module(peft_config, adapter_name, target, device_map=device_map, **kwargs)
+
+        if adapter_name not in self.active_adapters:
+            new_module.requires_grad_(False)
+
+        self._replace_module(parent, target_name, new_module, target)
+
+    def _replace_module(self, parent: nn.Module, child_name: str, new_module: nn.Module, child: nn.Module) -> None:
```
BenjaminBossan: Can we not use this method from the parent class?

docs/source/package_reference/psoft.md
fei407: Added `- local: package_reference/psoft` and ` title: PSOFT ` behind BOFT.

src/peft/tuners/psoft/__init__.py
fei407: Removed.

src/peft/tuners/psoft/config.py
fei407: Replaced.

src/peft/tuners/psoft/config.py
fei407: Added defaults at docstring.

src/peft/tuners/psoft/config.py
fei407: `psoft_svd_lowrank_niter` is only effective when `psoft_svd="lowrank"` (see layer.py) and is ignored otherwise. A warning is raised for invalid configurations in `__post_init__`, and tests were added in test_initialization (lines 2276–2322) to enforce this behavior.

src/peft/tuners/psoft/config.py
fei407: Replaced with `ab_svd_init`.

src/peft/tuners/psoft/layer.py
fei407: Added.

src/peft/tuners/psoft/layer.py
fei407: Thank you very much for the suggestion. After revisiting the implementation, I realized that modifying the adapter name is no longer necessary, especially after introducing `BufferDict`. I have therefore removed the related sanitization logic.

src/peft/tuners/psoft/layer.py
fei407: Thank you for the suggestion. I have revised the implementation to use `BufferDict`.

src/peft/tuners/psoft/layer.py
fei407: I previously assumed that `base.weight` would be on the meta device under` low_cpu_mem_usage`, and implemented a **lazy initialization** mechanism for the A/B cache. After adopting `BufferDict`, the cache is initialized once in `update()`, so the lazy initialization is no longer needed. I have removed `_ensure_psoft_cache ` and its usage in `get_delta_weight `and `forward `accordingly.

src/peft/tuners/psoft/layer.py
fei407: Yes, I have simplified this logic.

src/peft/tuners/psoft/layer.py
fei407: This is still related to the cache buffer. Previously, to make `test_feature_extraction_models` pass, I had to introduce an additional constraint tied to `inference_mode`. After switching to `BufferDict`, this is no longer necessary, so I have removed the corresponding constraint.

src/peft/tuners/psoft/layer.py
fei407: PSOFT can be viewed as an incremental adaptation method and therefore supports conversion from LoRA. I have added a `supports_lora_conversion` function and introduced the necessary validation logic.

src/peft/tuners/psoft/layer.py
fei407: Fixed.

src/peft/tuners/psoft/model.py
fei407: Yes, that makes sense. I have updated the implementation to use the method from the parent class instead.

tests/test_psoft.py
fei407: I agree. I suggest removing `test_psoft.py`, since its functionality is already covered by existing tests. Moreover, `test_psoft_column_angle_invariance_after_training` does not appear to be essential.

fei407:
Thank you very much for your thorough and detailed review. Your comments are very helpful, especially regarding the PSOFT A/B cache buffers and the initialization logic. I have substantially simplified this part of the implementation, and it is now much clearer and cleaner than before.
I have also addressed the other points one by one, please see the detailed replies below. I would be very happy to make any further adjustments based on your suggestions.😊

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,511 @@
...
+            R = R * self.vector_a[None, :]
+
+        if cast_to_fp32:
+            R = R.to(orig_dtype)
+
+        return R
+
+    def __repr__(self) -> str:
+        return (
+            f"{self.__class__.__name__}("
```
BenjaminBossan:
```suggestion
            f"psoft.{self.__class__.__name__}("
```

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,511 @@
...
+        self.psoft_svd_lowrank_niter: dict[str, int] = {}
+        self.ab_svd_init: dict[str, Optional[str]] = {}
+
+        # per-adapter trainable module
+        self.psoft_R = nn.ModuleDict({})
+
+        # per-adapter cache state
+        self._psoft_A_cache = BufferDict(persistent=False)
+        self._psoft_B_cache = BufferDict(persistent=False)
+        self._psoft_meta_warned: set[str] = set()
```
BenjaminBossan: This attribute seems to be unnecessary.

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,511 @@
...
+        if adapter_name not in self._psoft_A_cache or adapter_name not in self._psoft_B_cache:
+            return None, None
+        return self._psoft_A_cache[adapter_name], self._psoft_B_cache[adapter_name]
+
+    def _set_psoft_ab_cache_buffers(self, adapter_name: str, A: torch.Tensor, B: torch.Tensor) -> None:
+        self._psoft_A_cache[adapter_name] = A
+        self._psoft_B_cache[adapter_name] = B
+
+    def update_layer(self, adapter_name: str, config: PsoftConfig, **kwargs: Any) -> None:   
+
```
BenjaminBossan:
```suggestion
```

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,511 @@
...
+                delta_weight = self.get_delta_weight(active_adapter)
+                base_layer.weight.data += delta_weight.to(base_layer.weight.dtype)
+
+            self.merged_adapters.append(active_adapter)
+
+    def supports_lora_conversion(self, adapter_name: str = "default") -> bool:
+        # Conversion to LoRA is only supported if the adapter exists
+        # and its A/B cache buffers are available.
+        if adapter_name not in self.psoft_R:
+            return False
```
BenjaminBossan: I think we don't need this. If a specific layer doesn't have this adapter, it's fine, as long as some other layers have the adapter and support it.

src/peft/tuners/psoft/layer.py
fei407: Fixed.

src/peft/tuners/psoft/layer.py
fei407: Removed.

src/peft/tuners/psoft/layer.py
fei407: Removed.

src/peft/tuners/psoft/layer.py
fei407: Thanks. Removed.

fei407:
Thank you again for the detailed review and helpful suggestions. I have addressed these issues. I also ran `make style` locally to ensure the style checks pass.

Thanks as well for the reminder about force pushes. I’ll avoid using them going forward to keep the review process smoother.

fei407:
Thank you very much for the detailed guidance. I’ve addressed the missing pieces as suggested:
**1. Extended tests:** Integrated PSOFT into `test_seq_classifier.py` and `test_config.py`. All related tests pass locally with `pytest`.
**2. Added example:** Added `examples/psoft_finetuning/`. The code snippets in the README have been verified and run successfully.
**3. Added MetaMathQA experiments:**  Added two experiment configurations under: `method_comparison/MetaMathQA/experiments/psoft/`. Both experiments were tested locally on an RTX 5090.
Default configuration results:
```
r = 128
Trainable parameters: 469,504 (~0.47M)
Test accuracy: 40.2%
```
```
r = 256
Trainable parameters: 1,856,512 (~1.86M)
Test accuracy: 45.9%
```

The corresponding result JSON files are attached below.
[psoft--llama-3.2-3B-rank128--2026-02-26T16-34-20+00-00.json](https://github.com/user-attachments/files/25586660/psoft--llama-3.2-3B-rank128--2026-02-26T16-34-20%2B00-00.json)
[psoft--llama-3.2-3B-rank256--2026-02-26T19-24-30+00-00.json](https://github.com/user-attachments/files/25586661/psoft--llama-3.2-3B-rank256--2026-02-26T19-24-30%2B00-00.json)

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,502 @@
...
+
+    def __repr__(self) -> str:
+        return "psoft." + super().__repr__()
+
+
+def dispatch_default(
+    target: nn.Module,
+    adapter_name: str,
+    config: PsoftConfig,
+    parameter_name: Optional[str] = None,
```
BenjaminBossan: Let's remove this argument, it's only relevant for LoRA.

src/peft/tuners/psoft/model.py
```diff
@@ -0,0 +1,84 @@
...
+    def _create_and_replace(
+        self,
+        peft_config: PsoftConfig,
+        adapter_name: str,
+        target: nn.Module,
+        target_name: str,
+        parent: nn.Module,
+        current_key: str,
+        *,
+        parameter_name: Optional[str] = None,
```
BenjaminBossan, line 56: Let's remove this argument, it's only relevant for LoRA.

src/peft/tuners/psoft/config.py
```diff
@@ -0,0 +1,316 @@
...
+            validated settings: https://openreview.net/forum?id=FSHrinMArK.
+        target_modules (`Optional[Union[List[str], str]]`):
+            The names of the modules to apply the adapter to. If this is specified, only the modules with the specified
+            names will be replaced. When passing a string, a regex match will be performed. When passing a list of
+            strings, either an exact match will be performed or it is checked if the name of the module ends with any
+            of the passed strings. If this is specified as 'all-linear', then all linear/Conv1D modules are chosen (if
+            the model is a PreTrainedModel, the output layer excluded). If this is not specified, modules will be
+            chosen according to the model architecture. If the architecture is not known, an error will be raised -- in
+            this case, you should specify the target modules manually. To avoid targeting any modules (because you want
+            to apply `target_parameters`), set `target_modules=[]`.
```
BenjaminBossan: Remove the last sentence, it's only relevant for LoRA.

src/peft/tuners/psoft/config.py
```diff
@@ -0,0 +1,316 @@
...
+        metadata={
+            "help": (
+                "List of module names or regex expression of the module names to replace with PSOFT. "
+                "For example, ['q', 'v'] or '.*decoder.*(SelfAttention|EncDecAttention).*(q|v)$'. "
+                "This can also be a wildcard 'all-linear' which matches all linear/Conv1D "
+                "(if the model is a PreTrainedModel, the output layer excluded). "
+                "If not specified, modules will be chosen according to the model architecture, If the architecture is "
+                "not known, an error will be raised -- in this case, you should specify the target modules manually. "
+                "To avoid targeting any modules (because you want to apply `target_parameters`), set "
+                "`target_modules=[]`."
```
BenjaminBossan: Remove the last sentence, it's only relevant for LoRA.

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,502 @@
...
+
+    def _get_R_matrix(self, adapter_name: str) -> torch.Tensor:
+        return self.psoft_R[adapter_name].get_matrix()
+
+    def get_delta_weight(self, adapter_name: str) -> torch.Tensor:
+        """
+        ΔW = scaling * B (R - id_mat) A Returns in base weight layout (respecting fan_in_fan_out).
+        """
+        if adapter_name not in self.psoft_R:
+            raise KeyError(f"Adapter {adapter_name} not found in PSOFT layer.")
```
BenjaminBossan: I think this check shouldn't be required, as `get_delta_weight` is only called if the adapter exists on that module.

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,502 @@
...
+        self.fan_in_fan_out = False
+
+        base_layer = self.get_base_layer()
+        in_features, out_features = _get_in_out_features(base_layer)
+        self.in_features = in_features
+        self.out_features = out_features
+
+    def _get_psoft_ab_cache_buffers(self, adapter_name: str):
+        if adapter_name not in self._psoft_A_cache or adapter_name not in self._psoft_B_cache:
+            return None, None
```
BenjaminBossan: In what circumstances can this happen? If it can't happen, let's remove it.

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,502 @@
...
+            else:
+                delta_weight = self.get_delta_weight(active_adapter)
+                base_layer.weight.data += delta_weight.to(base_layer.weight.dtype)
+
+            self.merged_adapters.append(active_adapter)
+
+    def supports_lora_conversion(self, adapter_name: str = "default") -> bool:
+        # Conversion to LoRA is supported if A/B cache buffers are available.
+        A, B = self._get_psoft_ab_cache_buffers(adapter_name)
+        return A is not None and B is not None
```
BenjaminBossan: Regarding my other comment: If `A` and `B` cannot be `None` here, we can simplify this method to return `True`.

src/peft/tuners/psoft/layer.py
```diff
@@ -0,0 +1,502 @@
...
+
+    # R = (I+Q)(I-Q)^(-1)
+    def get_matrix(self) -> torch.Tensor:
+        # non-orth case
+        if not self.orth:
+            R = self.weight
+            if self.mag_b and self.vector_b is not None:
+                R = self.vector_b[:, None] * R
+            if self.mag_a and self.vector_a is not None:
+                R = R * self.vector_a[None, :]
```
BenjaminBossan: In what circumstances can this happen?

method_comparison/MetaMathQA/experiments/psoft/llama-3.2-3B-fast/adapter_config.json
BenjaminBossan, line 1: Thanks for providing the experiment settings. IMO the two settings are close enough that we don't really need both. So let's either delete this one (as the other gives better results), or see if completely different settings could also work well on this problem (e.g. changing `ab_svd_init`, `psoft_svd`, `psoft_mag_a`, `psoft_mag_b`, `use_cayley_neumann`. You probably have a much better intuition of what could work here. And even if the accuracy doesn't improve, maybe a setting can be found that excels in another dimension, like fewer parameters, faster training speed, or lower memory.

examples/psoft_finetuning/psoft_finetuning.py
```diff
@@ -0,0 +1,182 @@
...
+    # We only allow fp16/bf16/fp32 here to avoid accidental quantized loading.
+    if script_args.bits.lower() not in {"bf16", "fp16", "fp32"}:
+        raise ValueError("PSOFT example only supports bits in ['bf16','fp16','fp32'] (no quantization).")
+
+    torch_dtype = _dtype_from_bits(script_args.bits)
+
+    # Load base model
+    model = AutoModelForCausalLM.from_pretrained(
+        script_args.base_model_name_or_path,
+        torch_dtype=torch_dtype,
```
BenjaminBossan:
```suggestion
        dtype=torch_dtype,
```

examples/psoft_finetuning/README.md
```diff
@@ -0,0 +1,80 @@
...
+```
+@inproceedings{wu2026efficient,
+title={Efficient Orthogonal Fine-Tuning with Principal Subspace Adaptation},
+author={Wu, Fei and Hu, Jia and Min, Geyong and Wang, Shiqiang},
+booktitle={The Fourteenth International Conference on Learning Representations},
+year={2026},
+url={https://openreview.net/forum?id=FSHrinMArK}
+}
+```
+
```
BenjaminBossan: Remove the excess lines.

BenjaminBossan (changes requested):
Thanks for the last update, not much is missing. I found a few issues, but they're all quite small, so should be easy to fix.

Also thanks for running the experiments. The results seem to be in proximity to what we see with OFT. Would you say they're in line with your expectations?

method_comparison/MetaMathQA/experiments/psoft/llama-3.2-3B-fast/adapter_config.json
fei407: I’ve kept only the r=256 configuration and renamed it to `default`.  
In addition, I added a faster-training variant named `fast`, which has faster training speed.

src/peft/tuners/psoft/model.py
fei407: Removed.

src/peft/tuners/psoft/layer.py
fei407: Removed.

src/peft/tuners/psoft/config.py
fei407: Removed.

src/peft/tuners/psoft/config.py
fei407: Removed.

src/peft/tuners/psoft/layer.py
fei407: Removed.

src/peft/tuners/psoft/layer.py
fei407: I’ve checked the code and confirmed that this situation cannot occur, so I’ve removed the check.

src/peft/tuners/psoft/layer.py
fei407: Fixed.

src/peft/tuners/psoft/layer.py
fei407: I’ve simplified this part of the logic. It was previously used to determine when to apply the `vector_b ` and `vector_a ` separately.

examples/psoft_finetuning/psoft_finetuning.py
fei407: Fixed.

examples/psoft_finetuning/README.md
fei407: Removed.

fei407:
Thank you very much for the careful review. I’ve addressed the comments and made the corresponding changes.

Regarding the experiments on MetaMathQA, I believe the results under the default configuration are in line with my expectations. In addition, PSOFT offers a more flexible range of trainable parameter configurations compared to LoRA and OFT variants, as the number of trainable parameters is decoupled from the hidden dimension. This allows for more fine-grained trade-offs between performance and efficiency.

method_comparison/MetaMathQA/experiments/psoft/llama-3.2-3B-fast/adapter_config.json
fei407: I also evaluated the `fast` configuration against the `default` setting. The performance is comparable and even slightly better in the single run, while the total training time is reduced from 1656 to 1147, with lower memory usage.

[psoft--llama-3.2-3B-fast--2026-02-27T13-23-13+00-00.json](https://github.com/user-attachments/files/25606114/psoft--llama-3.2-3B-fast--2026-02-27T13-23-13%2B00-00.json)
[psoft--llama-3.2-3B-rank256--2026-02-26T19-24-30+00-00.json](https://github.com/user-attachments/files/25606116/psoft--llama-3.2-3B-rank256--2026-02-26T19-24-30%2B00-00.json)

fei407:
Thanks a lot for your review and support during the process. I really appreciate it! 😊

PR #3038, opened 2026-02-11
Fix Embedding.get_delta_weight replacing nn.Parameter with plain tensor

Mr-Neutr0n:
## Summary

- `Embedding.get_delta_weight` replaces `nn.Parameter` objects in `lora_embedding_A`/`lora_embedding_B` with plain tensors when running on CPU with float16/bfloat16 weights
- This breaks gradient tracking, parameter registration, and optimizer state for any training that follows a merge/unmerge cycle
- Fix uses `.data` assignment to preserve the `nn.Parameter` wrapper, consistent with `Linear.get_delta_weight` and `_ConvNd.get_delta_weight` which already handle this correctly

## Details

When `cast_to_fp32` is `True`, the current code does:

```python
weight_A = self.lora_embedding_A[adapter]  # nn.Parameter
weight_A = weight_A.float()                # plain tensor (detached)
# ... compute output ...
self.lora_embedding_A[adapter] = weight_A.to(dtype)  # replaces Parameter with tensor!
```

The `.float()` call creates a new plain tensor. Assigning it back directly overwrites the `nn.Parameter`. The fix:

```python
self.lora_embedding_A[adapter].data = weight_A.to(dtype)
self.lora_embedding_B[adapter].data = weight_B.to(dtype)
```

This matches the pattern used in `Linear.get_delta_weight` (line 893-894) and `_ConvNd.get_delta_weight` (line 1498-1499).

## Test plan

- [ ] Verify `Embedding.get_delta_weight` preserves `nn.Parameter` type after call on CPU with float16 weights
- [ ] Verify merge/unmerge cycle on LoRA-adapted embedding layers does not break subsequent training
- [ ] Existing tests continue to pass

### DIFF
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -3361,6 +3361,40 @@ class TestPeftCustomModel(PeftCommonTester):
                 # windows error
                 pass
 
+    @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
+    def test_embedding_get_delta_weight_preserves_parameter_type(self, dtype):
+        # Test that Embedding.get_delta_weight does not replace nn.Parameter with plain tensor
+        # when running on CPU with float16/bfloat16 weights (cast_to_fp32 path).
+        # See: https://github.com/huggingface/peft/pull/3038
+        model = ModelEmbConv1D(emb_size=100)
+        config = LoraConfig(target_modules=["emb"], init_lora_weights=False)
+        peft_model = get_peft_model(model, config)
+        # Cast entire model (including LoRA params) to the target dtype
+        peft_model = peft_model.to(dtype)
+
+        emb_layer = peft_model.base_model.model.emb
+
+        # Ensure weights are the expected dtype and on CPU (triggers cast_to_fp32)
+        assert emb_layer.lora_embedding_A["default"].dtype == dtype
+        assert emb_layer.lora_embedding_B["default"].dtype == dtype
+        assert emb_layer.lora_embedding_A["default"].device.type == "cpu"
+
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Thanks, this is a nice catch. Could you please also provide a unit test for this bug? There isn't really a perfect place to put it, I'd suggest to add a one-off test to `test_decoder_models.py` using one of the toy decoder models and targeting the embedding layer to reproduce the behavior. LMK if you need further guidance.

BenjaminBossan:
Thanks for the unit test, it looks quite good. I ran it locally on the main branch and to my surprise, it passed even without your fix. Jumping into the debugger, I see:

```python
1136 	            self.lora_embedding_A[adapter] = weight_A.to(dtype)
1137 ->	            breakpoint()
1138 	            self.lora_embedding_B[adapter] = weight_B.to(dtype)
1139 	
1140 	        return output_tensor
(Pdb) type(self.lora_embedding_B[adapter])
<class 'torch.nn.parameter.Parameter'>
(Pdb) type(weight_B.to(dtype))
<class 'torch.Tensor'>
(Pdb) self.lora_embedding_B[adapter] = weight_B.to(dtype)
(Pdb) type(self.lora_embedding_B[adapter])
<class 'torch.nn.parameter.Parameter'>
```

So it appears like `nn.ParameterDict` automatically converts the tensor to an `nn.Parameter` on assignment, which I didn't know. Did you observe any faulty behavior in the wild?

Mr-Neutr0n:
you're right — nn.ParameterDict handles the conversion automatically. didn't realize that. closing since the fix isn't actually needed. thanks for digging into it\!

PR #3008, opened 2026-01-22
Upgrade GitHub Actions for Node 24 compatibility

salmanmkc:
## Summary
This PR upgrades GitHub Actions to their latest versions for Node.js 24 compatibility and security updates.

## Changes

| Action | Old Version(s) | New Version | Files |
|--------|---------------|-------------|-------|
| actions/checkout | 08c6903cd8c0fde910a37f88322edcfb5dd907a8 | 8e8c483db84b4bee98b60c0593521ed34d9990e8 | build_docker_images.yml, deploy_method_comparison_app.yml, integrations_tests.yml, nightly.yml, stale.yml, test-docker-build.yml, tests-main.yml, tests.yml, torch_compile_tests.yml, trufflehog.yml, zizmor.yaml |
| actions/setup-python | e797f83bcb11b83ae66e0230d6156d7c80228e7c | a309ff8b426b58ec0e2a45f0f869d46889d02405 | integrations_tests.yml, stale.yml, tests-main.yml, tests.yml |

## Why these changes?
- Keeps actions up to date with latest stable releases
- Updated actions include security fixes and new features

## Testing
These changes only update action versions and don't modify workflow logic.

PR #3040, opened 2026-02-12
Add Dependabot configuration for GitHub Actions

salmanmkc:
## Summary

Add a Dependabot configuration to automatically keep GitHub Actions up to date.

## Changes

Adds `.github/dependabot.yml` configured to:
- Monitor GitHub Actions for version updates
- Check for updates on a **weekly** schedule

## Why?

Following the merge of #3008 which upgraded actions for Node 24 compatibility, this ensures future action updates are handled automatically via Dependabot pull requests. This way, the repository will stay up to date with the latest action versions without manual intervention.

## Context

- [GitHub Dependabot documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates)
- Dependabot will automatically open PRs when new versions of GitHub Actions are released
- Each PR will include release notes and changelogs for easy review

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,12 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"
+    groups:
+      ci-actions:
+        patterns:
+          - "actions/*"
+      third-party-actions:
+        patterns:
+          - "*"
+        exclude-patterns:
+          - "actions/*"


BenjaminBossan:
Thanks for this PR @salmanmkc, I didn't know it was possible to enable dependabot specifically for GH actions. We can definitely give this a try. Would you say that a monthly schedule is enough? Weekly feels a bit too frequent to me.

salmanmkc:
> Thanks for this PR @salmanmkc, I didn't know it was possible to enable dependabot specifically for GH actions. We can definitely give this a try. Would you say that a monthly schedule is enough? Weekly feels a bit too frequent to me.

Yeah monthly should be good enough, however the default configuration for dependabot is to do a PR for each dependency. So basically doing it weekly in theory should be a more steady flow, rather than having an influx every month. 

This can be changed though, similar to my PR, you can have dependency [updates grouped together](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#groups--).

For example all actions in one PR

```
version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "monthly"
    groups:
      # Group name (arbitrary)
      actions-updates:
        patterns:
          - "*"  # Group ALL actions into one PR
```

or for example, the groups can be designed as such, third party actions and GitHub first party actions, which closely resembles the kind of PRs I did (though I didn't use dependabot).


```
groups:
  ci-actions:
    patterns:
      - "actions/*"       # e.g. actions/checkout, actions/setup-python
  third-party-actions:
    patterns:
      - "*"
    exclude-patterns:
      - "actions/*"
      - 

```

BenjaminBossan:
TIL. The option to group 1st and 3rd party actions sounds very useful. Let's go with that and see how it works out in practice. Thanks so much for your input.

salmanmkc:
> TIL. The option to group 1st and 3rd party actions sounds very useful. Let's go with that and see how it works out in practice. Thanks so much for your input.

Updated

PR #3041, opened 2026-02-12
Improve LoftQ documentation

githubnemo:
To me the LoftQ application process was not immediately obvious. I'm sure it can't hurt to add a bit of context what the two main ways are to apply LoftQ.

In my quest to search for enlightenment I was often directed to the quantization guide in the PEFT docs which featured a lot less details than the developer guide for LoRA. Since LoftQ is mainly used in quantization I think it makes sense to move it there and link there from the developer guide.

### DIFF
--- a/docs/source/developer_guides/quantization.md
+++ b/docs/source/developer_guides/quantization.md
@@ -108,6 +108,7 @@ There are currently two intended ways of applying LoftQ:
    `BitsAndBytesConfig(load_in_4bit=True, bnb_4bit_quant_type="nf4")` and load the LoftQ-initialized adapter on top
 2. load the quantized base model, initialize a PEFT model with LoRA (no loftq) and use `replace_lora_weights_loftq` to
    apply LoftQ initialization by streaming the reference weights of the non-quantized base model from the weights file
+   (explained further in the next section)
 
 #### Using `replace_lora_weights_loftq` for on-the-fly LoftQ application
 


docs/source/developer_guides/quantization.md
```diff
@@ -99,6 +99,41 @@ You're all set for training with whichever training method you prefer!
...
 
+In general, for LoftQ to work best, it is recommended to target as many layers with LoRA as possible, since those not targeted cannot have LoftQ applied. This means that passing `LoraConfig(..., target_modules="all-linear")` will most likely give the best results. Also, you should use `nf4` as quant type in your quantization config when using 4bit quantization, i.e. `BitsAndBytesConfig(load_in_4bit=True, bnb_4bit_quant_type="nf4")`.
+
+There are currently two intended ways of applying LoftQ:
+
+1. load the non-quantized base model, apply LoftQ with your intended quantization level (e.g., `bits=4` for nf4) and
+   save the resulting adapter. Then quantize the base model using, for example,
+   `BitsAndBytesConfig(load_in_4bit=True, bnb_4bit_quant_type="nf4")` and load the LoftQ-initialized adapter on top
+2. load the quantized base model, initialize a PEFT model with LoRA (no loftq) and use `replace_lora_weights_loftq` to
```
BenjaminBossan, line 109: This new addition makes the first paragraph of the next section a bit redundant.

githubnemo: True, if I had to choose between deleting or extending, I'd rather add a section giving a short example how to use LoftQ with method (1) than deleting the section. But I could also see this section reduced a bit. WDYT?

BenjaminBossan: It might suffice to add something like `(explained further in the next section)`. Otherwise, it may appear as if there was yet a 3rd way.

PR #3042, opened 2026-02-13
FIX Add guard when detecting the optimum version

BenjaminBossan:
Reported by @SunMarc 

Apparently, it's possible that there is an error when the optimum version is being determined, even if optimum is installed. To prevent this from throwing an error, a `try...catch` is now added. If we can't detect the version, we optimistically assume it's high enough and continue. Otherwise, users with otherwise working environments could get an error here with no avenue to resolve it.


BenjaminBossan:
_failing tests are unrelated_

PR #3047, opened 2026-02-17
Refactor layer initialization: PR 2960 continued

BenjaminBossan:
Continuation of #2960 which only covered LoRA and AdaLoRA.

This PR targets the `__init__` and `update_layer` methods of the other relevant PEFT methods (i.e. everything except for prompt learning). Again, the goal is to pass the corresponding PEFT config directly and let the layers pick out relevant arguments, instead of having the model classes pick out the arguments and pass them to the layers. The advantage is that this reduces code duplication (this PR reduces LOC by 250). If, say, a new init argument is added, there is no longer the need to update the code in multiple places just to ensure that the argument is passed to the layers correctly.

The handling for trainable tokens is a bit awkward, as the config can have different types. I think ideally it should always be a config instance, but I'm not sure how easy that would be to change.

Bone was not updated, as it will be removed for the 0.19 release.

~~There was also a bug in the GPTQ AdaLoRA implementation caused by #2960, which is now fixed.~~ Resolved via #3099


src/peft/tuners/waveft/model.py
```diff
@@ -112,36 +112,18 @@ def _create_and_replace(
             target_name_key = get_pattern_key(pattern_keys, current_key)
             n_frequency = waveft_config.n_frequency_pattern.get(target_name_key, waveft_config.n_frequency)
 
-        # Determine wavelet_family
-        wavelet_family = None
-        if "wavelet_family" in optional_kwargs:
-            wavelet_family = optional_kwargs["wavelet_family"]
-        if wavelet_family is None:
-            wavelet_family = waveft_config.wavelet_family
```
BenjaminBossan, line 120: Note: I ran the unit tests and this handling seemed unnecessary, so I just removed it.

BenjaminBossan:
Converting to draft PR for now. We will merge this *after* the PEFT v0.19.0 release to give it time to stew. By then, we'll have more PEFT methods added, so this PR will have to be updated to include those.

BenjaminBossan:
Not stale: This PR will be resumed and updated after the 0.19 release.

src/peft/tuners/trainable_tokens/layer.py
```diff
@@ -104,13 +106,24 @@ def _collect_token_weights(self, weight: torch.Tensor, rows: torch.Tensor, embed
...
+            cfg = config
+        elif config is not None:
+            cfg = {
+                "token_indices": config.token_indices,
+                "init_weights": config.init_weights,
+            }
+        else:
+            cfg = kwargs
+
+        if cfg.get("tied_adapter", None):
```
githubnemo, line 121: I think we shouldn't treat `tied_adapter` as a 'config' variable since it is a parameter specific to `update_layer`. This should resolve a bit of the awkwardness?

githubnemo (commented):
Thanks for taking this on! This makes the interface so much cleaner :)

One comment regarding trainable tokens, otherwise LGTM.

> There was also a bug in the GPTQ AdaLoRA implementation caused by https://github.com/huggingface/peft/pull/2960, which is now fixed.

I didn't find AdaLoRA changes in this PR, is this separate or outdated?

BenjaminBossan:
> I didn't find AdaLoRA changes in this PR, is this separate or outdated?

It's the same fix as contained here: #3099. Since I merged main, it's not visible anymore.

src/peft/tuners/trainable_tokens/layer.py
BenjaminBossan: Good catch, it's now an explicit parameter and I refactored the code so that `config` is never cast to a `dict`, which was messy. It can still be `None` when coming through `_set_trainable`. This could probably also be cleaned up, but let's not complicate this PR even more.

githubnemo (approved):
Thanks for the fixes :)

PR #3048, opened 2026-02-18
Add support for LoRA with Transformer Engine

balvisio:
This PR adds support so that TransformerEngine layers (https://github.com/NVIDIA/TransformerEngine) are recognized as valid layers to which LoRA adapters can be added.

### DIFF
--- a/examples/lora_finetuning_transformer_engine/README.md
+++ b/examples/lora_finetuning_transformer_engine/README.md
@@ -1,0 +1,119 @@
+# Transformer Engine ESM2 LoRA Fine-Tuning
+
+This example demonstrates LoRA fine-tuning for Transformer Engine ESM2 token classification.
+
+## Setup
+
+Choose one of the two options below.
+
+### Option A: Docker (recommended)
+
+Build a self-contained image based on the publicly available NVIDIA PyTorch container
+(`nvcr.io/nvidia/pytorch:26.01-py3`), which already ships CUDA, cuDNN, and Transformer Engine:
+
+```bash
+docker build -t lora-te examples/lora_finetuning_transformer_engine
+```
+
+Run the training inside the container:
+
+```bash
+docker run --gpus all --rm lora-te \
+  python lora_finetuning_te.py \
+    --base_model nvidia/esm2_t6_8M_UR50D \
+    --output_dir ./esm2_lora_output \
+    --num_train_samples 256 \
+    --num_eval_samples 64 \
+    --num_epochs 1
+```
+
+Or start an interactive session to experiment:
+
+```bash
+docker run --gpus all --rm -it lora-te bash
+```
+
+### Option B: Virtual environment
+
+Create and activate a virtual environment, then install the Python dependencies:
+
+```bash
+python -m venv .venv
+source .venv/bin/activate
+pip install -r examples/lora_finetuning_transformer_engine/requirements.txt
+```
+
+**Transformer Engine** must be installed separately and must match the system CUDA toolkit version.
+See the [TE installation guide](https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/installation.html)
+for details.
+
+## What this example does
+
+- Loads a Transformer Engine ESM2 model for token classification
+- Applies LoRA adapters via PEFT
+- Generates random protein-like sequences
+- Assigns randomly generated secondary structure labels (`H`, `E`, `C`)
+- Trains/evaluates with `Trainer`
+
+## Run
+
+```bash
+python examples/lora_finetuning_transformer_engine/lora_finetuning_te.py \
+  --base_model nvidia/esm2_t6_8M_UR50D \
+  --output_dir ./esm2_lora_output \
+  --num_train_samples 256 \
+  --num_eval_samples 64 \
+  --num_epochs 1
+```
+
+> **Note:** The default ESM2 models on Hugging Face Hub ship custom modeling code.
+> You must pass `--trust_remote_code` to allow loading that code.
+
+## Customize
+
+```bash
+python examples/lora_finetuning_transformer_engine/lora_finetuning_te.py \
+  --base_model nvidia/esm2_t6_8M_UR50D \
+  --trust_remote_code \
+  --output_dir ./esm2_lora_output \
+  --max_length 256 \
+  --batch_size 4 \
+  --learning_rate 3e-4 \
+  --lora_r 16 \
+  --lora_alpha 32 \
+  --lora_dropout 0.1
+```
+
+## Dataset
+
+By default the script generates a **synthetic dataset** at runtime — random protein-like sequences
+with randomly generated secondary structure labels (`H`, `E`, `C`). This is useful for quick sanity checks and testing.
+
+For a more realistic evaluation, you can use the **Porter6** secondary-structure dataset.
+A download-and-convert script is available in the BioNeMo repository:
+
+[prepare_porter6_dataset.py](https://github.com/NVIDIA/bionemo-framework/blob/bd72d882bca458d9438e05661c41163949713d1f/bionemo-recipes/recipes/esm2_peft_te/data/prepare_porter6_dataset.py)
+
+Run it to produce train and validation parquet files, then pass them to the training script
+with `--train_parquet` and `--val_parquet`:
+
+```bash
+python examples/lora_finetuning_transformer_engine/lora_finetuning_te.py \
+  --base_model nvidia/esm2_t6_8M_UR50D \
+  --train_parquet porter6_train_dataset_55k.parquet \
+  --val_parquet porter6_val_dataset_2024_692.parquet \
+  --output_dir ./esm2_lora_output \
+  --num_epochs 3
+```
+
+## Outputs
+
+After training, the script saves:
+
+- PEFT adapter weights/config in `--output_dir`
+- Tokenizer files in `--output_dir`
+
+## More examples
+
+For additional examples of TransformerEngine-accelerated transformers, visit
+`https://github.com/NVIDIA/bionemo-framework/bionemo-recipes`.
--- a/examples/lora_finetuning_transformer_engine/lora_finetuning_te.py
+++ b/examples/lora_finetuning_transformer_engine/lora_finetuning_te.py
@@ -1,0 +1,330 @@
+#!/usr/bin/env python3
+"""
+Transformer Engine ESM2 LoRA fine-tuning example using Hugging Face Trainer.
+
+This script demonstrates:
+1. Loading a Transformer Engine-based ESM2 token classification model
+2. Applying LoRA adapters with PEFT
+3. Building a synthetic protein token-classification dataset in code
+4. Training with the Hugging Face Trainer (no DDP setup required)
+"""
+
+import os
+
+
+# TE-backed models are incompatible with Trainer's DataParallel wrapping.
+# Pin to a single GPU before torch is imported so torch.cuda.device_count() == 1.
+os.environ.setdefault("CUDA_VISIBLE_DEVICES", "0")
+
+import argparse
+import random
+
+import numpy as np
+import pandas as pd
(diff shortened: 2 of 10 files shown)


BenjaminBossan:
Thanks for this PR to add support for TE in PEFT. Before proceeding further, do you have a practical example of how to use it with a transformers model? I assume one way would be for the user to employ accelerate to apply TE. Also, I saw that you intend to replace the `nn.Linear` layers used for `lora_A` and `lora_B` with TE's low FP precision layers. Did you test if that works well? Usually, we would keep these layers in fp32 (or fp16/bf16), even if the base layer uses lower precision (say, int4).

balvisio:
Thank you for taking a look at this. The primary goal of this PR is to support models with TE layers already inside them. In fact, using te.Linear for the adapters is not strictly necessary. However, by default, this TE layers will use the torch default type.
For the example, we have created and validated a "recipe" to use fine-tune an ESM2 model using LoRA here: https://github.com/NVIDIA/bionemo-framework/tree/main/bionemo-recipes/recipes/esm2_peft_te

BenjaminBossan:
> The primary goal of this PR is to support models with TE layers already inside them.

I see, could you please provide a small example of how this looks in practice? We should eventually add a unit test with a real model architecture anyway, so this example would also help for that.

> using te.Linear for the adapters is not strictly necessary.

I think for consistency, we should stick with float32 by default. We could think of an option to use 4bit and 8bit floats but I'd have to think about how the API for that should look.

> For the example, we have created and validated a "recipe" to use fine-tune an ESM2 model using LoRA here: https://github.com/NVIDIA/bionemo-framework/tree/main/bionemo-recipes/recipes/esm2_peft_te

I think it would be great to include such an example here. Maybe simplified if possible (using `Trainer`, no DDP, no extra logging). WDYT?

balvisio:
>  We should eventually add a unit test with a real model architecture anyway, so this example would also help for that.
There is a unit test that creates a toy model that uses TE layers. Is that what you meant ? Otherwise I can change the test to do: `AutoModelForTokenClassification.from_pretrained("nvidia/esm2_t6_8M_UR50D", config=config, trust_remote_code=True, dtype="bfloat16")`. lmk

> I think for consistency, we should stick with float32 by default.
Do you mean to remove using TE layers as adapters? I can do that but just to make sure, currently the TE adapters will use the default torch type, not a low a precision necessarily.

> I think it would be great to include such an example here
Where should I add the example exactly? Is there a doc with examples?

BenjaminBossan:
> There is a unit test that creates a toy model that uses TE layers. Is that what you meant ? Otherwise I can change the test to do

Yeah, we can use a bit heavier models in `test_gpu_examples.py`, my goal would be to use a real model there, or at least a "tiny" version of a real model, instead of a toy model.

> Do you mean to remove using TE layers as adapters?

Yes, so keep the base layers as they are and use normal `nn.Linear` for LoRA.

> I can do that but just to make sure, currently the TE adapters will use the default torch type, not a low a precision necessarily.

Is there any advantage then?

> Where should I add the example exactly? Is there a doc with examples?

Docs would be nice, but I was thinking of the `examples/` folder.

balvisio:
@BenjaminBossan : I changed the tests to use a TE-based model and added an example.

tests/test_lora_te.py
BenjaminBossan, line 1: Thanks for adding tests with a small but not toy model. Let's move the tests to `test_gpu_examples.py` and group them all into one class `TestTransformerEngine` (check the other tests there for appropriate decorators).

Unfortunately, it seems like these tests require a model with `trust_remote_code=True`, which we really really want to avoid running automatically. Do you know if there is a model that can be used with TE and doesn't require remote code execution (say, one of the ESM models directly provided by Transformers)?

tests/test_lora_te.py
```diff
@@ -0,0 +1,114 @@
...
+def test_te_lora_backward(esm2_model, tokenized_inputs):
+    cfg = LoraConfig(target_modules=["layernorm_qkv"], r=4, lora_alpha=8)
+    lora_model = get_peft_model(esm2_model, cfg).to("cuda")
+
+    optimizer = torch.optim.AdamW(lora_model.parameters())
+    loss_fn = torch.nn.CrossEntropyLoss()
+    inputs = {k: v.to(next(lora_model.parameters()).device) for k, v in tokenized_inputs.items()}
+    logits = lora_model(**inputs).logits
+    labels = torch.randint(logits.shape[-1], logits.shape[:-1], device=logits.device)
+    loss = loss_fn(logits.view(-1, logits.size(-1)), labels.view(-1))
```
BenjaminBossan: How about also checking that the loss is finite or within a certain range?

src/peft/tuners/lora/te.py
```diff
@@ -0,0 +1,110 @@
...
+from peft.tuners.tuners_utils import BaseTunerLayer
+
+from .config import LoraConfig
+
+
+if is_te_available():
+    import transformer_engine as te
+
+
+class TELoRA(torch.nn.Module, LoraLayer):
```
BenjaminBossan: This is a TE layer for linear layers, right? So how about naming it `TeLinear`.

examples/lora_finetuning_transformer_engine/requirements.txt
```diff
@@ -0,0 +1,10 @@
+torch
+transformers
+datasets
+peft
+accelerate
+numpy
+safetensors
+# transformer-engine must match the system CUDA toolkit version.
+# Install it separately or use the system-provided package.
+# See: https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/installation.html
```
BenjaminBossan, line 12: For other folks who want to test, I had to manually set the `CPLUS_INCLUDE_PATH` env var to point to `cudnn` and `nccl`.

examples/lora_finetuning_transformer_engine/lora_finetuning_te.py
BenjaminBossan, line 1: This script runs code with `trust_remote_code=True`. As a user, I would be quite unhappy about that. Let's make this a command line argument that defaults to `False`, requiring users to opt-in. Of course, it would mean that there is an error when running the script with the default model. So I would suggest to make a check at the start if `trust_remote_code is False` and ESM2 is being used to raise an error and tell the user they need to enable `trust_remote_code` for this model.

src/peft/tuners/lora/te.py
```diff
@@ -0,0 +1,110 @@
...
+                lora_B = self.lora_B[active_adapter]
+                dropout = self.lora_dropout[active_adapter]
+                scaling = self.scaling[active_adapter]
+                x = self._cast_input_dtype(x, lora_A.weight.dtype)
+                result = result + lora_B(lora_A(dropout(x))) * scaling
+
+            result = result.to(torch_result_dtype)
+
+        return result
+
```
BenjaminBossan, line 106: Would merging/unmerging work with this layer?

examples/lora_finetuning_transformer_engine/README.md
BenjaminBossan, line 1: I would suggest to rename the directory to include "transformer_engine" in the name, many users may not know what "te" means.

examples/lora_finetuning_transformer_engine/requirements.txt
balvisio: To test this i used the the NVIDIA pytorch image ` nvcr.io/nvidia/pytorch:26.01-py3` which is publicly available. I can add instructions in the README. sounds good?

tests/test_lora_te.py
balvisio: Unfortunately, ATM there are no TE based models in the Transformers library.

tests/test_lora_te.py
balvisio: done

examples/lora_finetuning_transformer_engine/README.md
balvisio: done

examples/lora_finetuning_transformer_engine/lora_finetuning_te.py
balvisio: done

src/peft/tuners/lora/te.py
balvisio: good point. It wouldn't right now. It would require some thought in case of `fp8`. If it is ok , I can create an issue for now and I can work on the functionality when needed.

examples/lora_finetuning_transformer_engine/README.md
```diff
@@ -0,0 +1,66 @@
+# Transformer Engine ESM2 LoRA Fine-Tuning
+
+This example demonstrates LoRA fine-tuning for Transformer Engine ESM2 token classification. It uses a synthetic dataset generated at runtime.
```
BenjaminBossan: Is there a real dataset on HF that could be used instead? Checking your original repo, it looks like you're using files from disk there, so if there is no good dataset on HF, synthetic is fine as well.

examples/lora_finetuning_transformer_engine/requirements.txt
BenjaminBossan: Yes, why not, it could be helpful for some users.

src/peft/tuners/lora/te.py
BenjaminBossan: Let's just raise a `NotImplementedError` for `merge` and `unmerge` with a nice error message, and document that those are not supported yet.

src/peft/tuners/lora/te.py
```diff
@@ -0,0 +1,110 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-Apache2
```
BenjaminBossan, line 2: I haven't seen "LicenseRef-Apache2" before so I asked ChatGPT and it said that the correct way of specifying this SPDX license would be "SPDX-License-Identifier: Apache-2.0", providing [this reference](https://spdx.dev/learn/handling-license-info/). I'm not an expert at all, just wanted to mention it.

tests/test_gpu_examples.py
```diff
@@ -5838,3 +5840,89 @@ def test_lora_8bit_bnb_dtype_autocast_float32(self, peft_type, tmp_path):
...
+
+@pytest.mark.skipif(not is_te_available(), reason="transformer_engine is not available")
+class TestTransformerEngine:
+    """Tests for LoRA with TransformerEngine layers."""
+
+    model_id = "nvidia/esm2_t6_8M_UR50D"
+
+    @pytest.fixture
+    def config(self):
+        return AutoConfig.from_pretrained(self.model_id, trust_remote_code=True)
```
BenjaminBossan: We can't use `trust_remote_code=True` in our tests for external repos. So unfortunately, we need to figure out a different solution here. Would it be possible without too much extra effort to take another small model, say `facebook/opt-125m`, and replace some its layers with TE layers?

balvisio: done

src/peft/tuners/lora/te.py
balvisio: ok. let me check internally but should be fine since is what we use for other repos too.

src/peft/tuners/lora/te.py
balvisio: done

examples/lora_finetuning_transformer_engine/README.md
balvisio: We are working on the licensing to be able to upload the data to HF. It is a bit tricky because it is bio related data. For now I added a pointer to a script that the users can run to download it.

balvisio:
@BenjaminBossan Thanks for looking at this. I have addressed your comments.

BenjaminBossan:
@balvisio Thanks for the updates, the PR looks good. Could you please run `make style`?

balvisio:
done

PR #3051, opened 2026-02-19
CHORE: Remove deprecated Bone method

BenjaminBossan:
Bone was deprecated in favor of MiSS. Removal is scheduled for PEFT v0.19.0. Old checkpoints can be converted using
`scripts/convert-bone-to-miss.py`.

Unrelated change: I found the skip for GPT2 for the unloading decoder test to be unnecessary (it was probably redundant with the skip for Conv1D layers), so that function was removed completely.


tests/test_decoder_models.py
```diff
@@ -338,22 +328,7 @@ def _skip_if_not_conv1d_supported(model_id, config_cls):
         MissConfig,
         DeloraConfig,
     ]:
-        pytest.skip("Skipping BOFT/HRA/OFT/Bone/Road/SHiRA/C3A/MiSS/OSF/DeLoRA for GPT2LMHeadModel")
-
-
-def _skip_adalora_oft_hra_bone_for_gpt2(model_id, config_cls):
```
BenjaminBossan, line 353: I checked and this skip is unnecessary, so I removed it.

src/peft/utils/peft_types.py
```diff
@@ -19,36 +19,6 @@
 class PeftType(str, enum.Enum):
     """
     Enum class for the different types of adapters in PEFT.
-
-    Supported PEFT types:
```
BenjaminBossan, line 23: This docstring adds no value, so I removed it.

PR #3052, opened 2026-02-19
FIX Multiple failing nightly GPU tests

BenjaminBossan:
This fixes a couple of failing GPU tests from the nightly run:

- A DeepSpeed test was added but deepspeed is not installed, installing it now in the Dockerfile.
- `_no_split_modules` can be a set but `get_balanced_memory` expects str or tuple or list. Converting the set to a list now.
- Fix an incorrect usage of `hasattr`.
- Mark HQQ tests as `xfail`, since transformers v5 does not support it yet.
- Mark EETQ tests as `xfail`, since transformers v5 does not support it yet.

There are yet more failing GPU tests, but those cannot be fixed on the PEFT side. I reported those failures.


tests/test_gpu_examples.py
```diff
@@ -3491,6 +3491,7 @@ def tearDown(self):
         """
         clear_device_cache(garbage_collection=True)
 
+    @pytest.mark.xfail(reason="HQQ is not yet supported by Transformers v5", strict=True, raises=NotImplementedError)
```
githubnemo: Maybe add `condition=is_transformers_ge_5` for non-CI users?

BenjaminBossan:
@githubnemo Your suggestion is implemented.

PR #3055, opened 2026-02-20
ENH Support models with low precision float dtypes

BenjaminBossan:
More recent models like https://huggingface.co/MiniMaxAI/MiniMax-M2.5 can use low precision float dtypes for the base weights, e.g. `torch.float8_e4m3fn`. This PR enables LoRA to work with these models.

Implementing this for normal LoRA or other PEFT adapters is not difficult, we just upcast the weights to float32 (unless indicated otherwise by the user), which is already what we do with fp16 and bf16 by default. This was just a matter of adding the other dtypes to the list of dtypes to upcast.

The main difficulty comes from `target_parameters`. There, we materialize the `delta_weight` and add it to the base weight. However, low precision floats do not support addition. The current solution is to temporarily upcast the weight to a higher precision, add the `delta_weight`, clamp the result to the low fp range, and cast the weight back to that `dtype`. This is not optimal, as upcasting can be quite expensive and clamping could be too destructive. But at least it's a working solution for now.

For the same reason (i.e. W + dW not being possible ootb), merging also doesn't work right now. We could implement a similar solution as for `target_parameters`, but it would require to update _all_ PEFT methods that support merging separately. Maybe there is a better solution, for now it's not addressed.

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -6169,7 +6169,6 @@ class TestDtypeFp8:
     @pytest.fixture
     def model_high_prec(self):
         model_id = "facebook/opt-125m"
-        # only convert q_proj to fp8, otherwise we get nan results
         model = AutoModelForCausalLM.from_pretrained(model_id, device_map=0)
         return model
 


BenjaminBossan:
*Failing CI unrelated*

src/peft/tuners/lora/layer.py
```diff
@@ -2284,7 +2306,11 @@ def get_delta_weight(self, adapter_name, *args, **kwargs):
                 delta_weight = torch.einsum("o r e, e r i -> e o i", weight_B, weight_A) * self.scaling[adapter_name]
 
         param = self.get_param()
-        delta_weight = delta_weight.to(param.device, param.dtype)
+        if param.dtype in ALLOWED_COMPUTE_DTYPES:
+            delta_weight = delta_weight.to(param.device, param.dtype)
+        else:
+            # don't cast dW to weight dtype if it is in torch.float8_e4m3fn etc.
```
githubnemo: please document why this is done

src/peft/tuners/lora/layer.py
```diff
@@ -2075,7 +2076,28 @@ def __init__(self, delta_weight):
...
+                f"There is an attempt to upcast the targeted parameter to {upcast_dtype} "
+                f"but the only supported are: {ALLOWED_COMPUTE_DTYPES}."
+            )
+
+        # this operation can be quite costly
+        x = x.to(upcast_dtype)
+        z = x + y
+        # clamp to valid range before casting down
+        info = torch.finfo(orig_dtype)
+        z = z.clamp(min=info.min, max=info.max)
```
githubnemo, line 2095: I think that clamping is already a part of the downcast operation.

src/peft/utils/constants.py
```diff
@@ -366,3 +366,7 @@ def starcoder_model_postprocess_past_key_value(past_key_values):
 # otherwise there is no point in optimizing and there is a small chance of bugs in the optimization algorithm, so no
 # point in taking unnecessary risks. See #2045 for more context.
 MIN_TARGET_MODULES_FOR_OPTIMIZATION = 20
+# dtypes that are allowed to be used for adapter computation
+ALLOWED_COMPUTE_DTYPES = (torch.float16, torch.bfloat16, torch.float32)
+# float dtypes that should be upcast in the adapter for computation
+UPCAST_DTYPES = ("float8_e4m3fn", "float8_e4m3fnuz", "float8_e5m2", "float8_e5m2fnuz", "float8_e8m0fnu")
```
githubnemo, line 378: There's also `torch.float4_e2m1fn_x2` but I couldn't find the required setup to make it upcast (on x86 CPU it doesn't work).

BenjaminBossan: I get `NotImplementedError: "copy_kernel" not implemented for 'Float4_e2m1fn_x2'` just for trying to cast to that dtype. Let's ignore it for now.

src/peft/tuners/lora/layer.py
BenjaminBossan: Tested and it's not automatically clamped. Added a comment.

tests/test_gpu_examples.py
```diff
@@ -6010,7 +6167,14 @@ def setup_cleanup(self):
         clear_device_cache(garbage_collection=True)
 
     @pytest.fixture
-    def model(self):
+    def model_high_prec(self):
+        model_id = "facebook/opt-125m"
+        # only convert q_proj to fp8, otherwise we get nan results
```
githubnemo: remove comment

BenjaminBossan: thanks for catching

PR #3056, opened 2026-02-22
feat: add krylov_svd as drop-in replacement for torch.svd_lowrank

Brokttv:
Adds krylov_svd() as a general-purpose drop-in replacement for 
torch.svd_lowrank, as discussed with @BenjaminBossan 

The function uses subspace iteration with QR re-orthogonalization 
at every step, which prevents numerical rank collapse in the power 
iteration. Benchmarks across typical LLM weight shapes (1k×1k to 
8k×8k) show 100–1000x better singular value accuracy (SV-MSE) 
consistently, with speed and memory roughly on par.

- Zero new dependencies — pure torch only
- No API changes, fully backwards compatible
- Same arguments as torch.svd_lowrank (X, q, niter)

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -1701,3 +1701,35 @@ def _get_module_names_tied_with_embedding(model) -> list[str]:
 
     # get module names from parameter names
     return sorted({name.rpartition(".")[0] for name in tied_weights})
+
+
+def krylov_svd(X, q, niter=4, oversampling=8):
+    """
+    Drop-in replacement for torch.svd_lowrank with improved numerical accuracy.
+    Uses subspace iteration with QR re-orthogonalization to prevent rank collapse.
+    Same number of matrix multiplications and same O(m*k) memory as svd_lowrank.
+
+    Args:
+        X:            (m, n) input matrix
+        q:            number of singular components (matches svd_lowrank arg name)
+        niter:        number of subspace iterations
+        oversampling: extra columns for numerical stability
+
+    Returns:
+        U (m, q), S (q,), V (n, q) -- same convention as torch.svd_lowrank
+    """
+    k = q + oversampling
+    dtype, device = X.dtype, X.device
+
+    Omega = torch.randn(X.shape[1], k, dtype=dtype, device=device)
+    Q, _  = torch.linalg.qr(Omega)
+    Y, _  = torch.linalg.qr(X @ Q)
+
+    for _ in range(niter):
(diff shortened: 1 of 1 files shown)


Brokttv:
@BenjaminBossan, you can now proceed with your runs as discussed.

BenjaminBossan:
@Brokttv Please don't ping after just two days, we have other things we have to work on.

Regarding the comparison, I ran a comparison on the MetaMathQA benchmark using PiSSA with low rank SVD and with Krylov SVD and the results are virtually identical:

image

Final train loss is 0.5935 for low rank SVD vs 0.5932 for Krylov SVD. Duration of the `get_peft_model` call were similar, as expected (0.486 vs 0.432). Peak memory usage was not affected (as expected, since peak memory happens during training, not initialization).

If we wanted to add this to PEFT, we would need to consider the extra complexity of adding this new algorithm. With `svd_lowrank`, we can rely on the testing performed by PyTorch but here, we would need to add our own tests to cover all possible use cases (dtype handling, tensor shapes, devices, etc.).

Perhaps adding this function would actually be a better fit to PyTorch itself. If it's added there, we can use it in PEFT with lower complexity overhead. Did you give this a thought?

Brokttv:
Thanks for running tests but I don't think train loss is the right metric to evaluate especially after 5000 steps, init gains will obviously get hidden by the optimizer after extensive training. I would go for either test or validation accuracy for useful comparison.
Regarding your last point, you might be right actually, and I'll see what I can do about that.

BenjaminBossan:
I think train loss is the right metric to check, as we would expect it to be directly influenced by the weight initialization. Downstream behavior like generation capabilities are very noisy in contrast and would require many (expensive) runs to get reliable data on. I can see the argument that after 5000 steps, losses may converge, but we can also see that at the very start (steps 250, 500, 750), losses are basically identical, as shown below. I also show default LoRA initialization, which has a significantly different loss at the start:

- `low rank SVD: 0.9300660800933838, 0.7125605494976044, 0.6794867808818817`
- `Krylov SVD:   0.930529625415802,  0.7127139033079147, 0.6800542780160904`
- `default LoRA: 0.9826480431556701, 0.7163139179944992, 0.679160294175148`

> Regarding your last point, you might be right actually, and I'll see what I can do about that.

I think PyTorch is the better home for this type of functionality. WDYT about closing this PR and issue for now and revisiting it when/if Krylov SVD is added to PyTorch?

PR #3058, opened 2026-02-23
docs: replace deprecated financial_phrasebank dataset in IA3 tutorial

### LINKED ISSUE #2998: Tutorial uses deprecated dataset (financial_phrasebank) that fails to load
## Description

The [Soft Prompts tutorial - IA3](https://huggingface.co/docs/peft/task_guides/ia3) uses the `financial_phrasebank` dataset, which fails to load in recent versions of the `datasets` library due to the deprecation of dataset loading scripts.

## Environment

- `datasets` library version: 3.0.0+ (current)
- `peft` version: [your version]
- Tutorial: https://huggingface.co/docs/peft/task_guides/prompt_based_methods

## Steps to Reproduce

Following the tutorial's data loading code:
```python

ds = load_dataset("financial_phrasebank", "sentences_allagree")
```

dhruvildarji:
## Summary

Fixes #2998

The `financial_phrasebank` dataset fails to load with recent versions of the `datasets` library because it relied on a deprecated loading script format. This PR replaces it with `zeroshot/twitter-financial-news-sentiment`, a financial sentiment dataset that uses the modern parquet format.

### Changes
- `docs/source/task_guides/ia3.md`: updated dataset reference, description, and `text_column` from `"sentence"` to `"text"` to match the new dataset's column name
- `examples/conditional_generation/peft_adalora_seq2seq.py`: same dataset and column name updates

The replacement dataset (`zeroshot/twitter-financial-news-sentiment`) was tested and confirmed working by @maerory in the issue thread.

## Test plan
- [x] Updated dataset loads successfully with `load_dataset("zeroshot/twitter-financial-news-sentiment")`
- [x] Column name updated from `sentence` to `text` throughout affected files
- [x] Label names still auto-detected from `features["label"].names` (no hardcoded changes needed)

### DIFF
--- a/examples/conditional_generation/peft_adalora_seq2seq.py
+++ b/examples/conditional_generation/peft_adalora_seq2seq.py
@@ -29,7 +29,10 @@ dataset = dataset["train"].train_test_split(test_size=0.1)
 dataset["validation"] = dataset["test"]
 del dataset["test"]
 
-classes = dataset["train"].features["label"].names
+if hasattr(dataset["train"].features["label"], "names"):
+    classes = dataset["train"].features["label"].names
+else:
+    classes = ["Bearish", "Bullish", "Neutral"]
 dataset = dataset.map(
     lambda x: {"text_label": [classes[label] for label in x["label"]]},
     batched=True,


dhruvildarji:
Thanks for the review @BenjaminBossan\! I've added the conditional generation notebook changes to this PR as requested. The changes from #3059 are now included here — the fix covers all instances: IA3 tutorial docs, peft_adalora_seq2seq.py, and all 4 conditional generation notebooks (IA3, LoRA, prompt tuning, prefix tuning).

BenjaminBossan:
Thanks for the update @dhruvildarji. I tried running one of the notebooks (peft_lora_seq2seq) and encountered some issues. First of all, the line:

`classes = dataset["train"].features["label"].names`

didn't work but I could replace it with:

`classes = dataset["train"]["label"]`

The next issue came from this line in `preprocess_function`:

`labels = tokenizer(targets, max_length=3, padding="max_length", truncation=True, return_tensors="pt")`

The issue here is that the targets are not strings, so they cannot be tokenized. Instead, they are ints, which encode the sentiment classes. Now we could just use those directly as the target without passing them through the tokenizer. However, then we would actually deal with a text classification task and not a seq2seq task as the notebook claims. Now I wonder if the dataset really fits the purpose.

Can you reproduce these errors? WDYT?

BenjaminBossan:
ping @dhruvildarji

dhruvildarji:
Thanks for testing this @BenjaminBossan! I've reproduced both issues and pushed a fix.

**Root cause:** With `zeroshot/twitter-financial-news-sentiment`, the `label` feature may not expose a `.names` attribute (plain integer feature vs ClassLabel, depending on the `datasets` version). When `.names` raises `AttributeError`, the `text_label` column is never created, so `preprocess_function` receives raw ints and `tokenizer()` fails with a `TypeError`.

**What I changed (latest commit):**
- Added a `hasattr` check before accessing `.names` across all 5 notebooks
- Falls back to `["Bearish", "Bullish", "Neutral"]` if `.names` is unavailable
- This ensures `text_label` is always a list of strings before tokenization

The seq2seq framing still holds — we're fine-tuning a T5 model to generate sentiment labels as text tokens ("Bearish"/"Bullish"/"Neutral"), which is valid for a generative seq2seq model. The dataset fits the purpose.

examples/conditional_generation/peft_adalora_seq2seq.py
BenjaminBossan: This line also needs the same adjustment as below in the notebooks, right?

dhruvildarji:
Fixed `peft_adalora_seq2seq.py` as well — same `hasattr` fallback applied on line 32. Should be good to go now.

PR #3062, opened 2026-02-23
CI: Dependabot PRs don't trigger unit tests

BenjaminBossan:
As Dependabot PRs don't require approval to start the CI, they will immediately trigger it, including the expensive unit tests. This PR should prevent unit tests to be run when Dependabot creates PRs, as this is generally unnecessary.

See #3040 for context.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,6 +37,8 @@ jobs:
 
   tests:
     needs: check_code_quality
+    # dependabot updates (which don't require approval for CI to run) shouldn't trigger unit tests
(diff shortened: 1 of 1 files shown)


schneidergithub:
Are you sure you don't want unit tests run on dependabot PRs?  I have definitely seen stuff break from updating packages.

BenjaminBossan:
> Are you sure you don't want unit tests run on dependabot PRs? I have definitely seen stuff break from updating packages.

This hasn't been the case for us so far and once the PR is merged, the unit tests will run. I would have preferred for the CI to require approval like from external contributors, but I don't know if this can be configured for Dependabot, hence this next best solution.

Issue #3025, opened 2026-02-06
Detail: removal of `.peft_config` when performing `.unload()`?


jchwenger:
### Feature request

In this simple scenario where you load a model, then want to experiment with various configs:

We first load the model
```python
import torch
from transformers import AutoModelForImageTextToText
model = AutoModelForImageTextToText.from_pretrained(
    "google/gemma-3-4b-pt",
    dtype=torch.float32,
)
```

Then I want to train some adapters:
```python
from peft import LoraConfig
from peft import get_peft_model
lora_config = LoraConfig(
    r=16,
    lora_alpha=16,
    lora_dropout=0.05,
    bias="none",
    target_modules="all-linear",
    task_type="CAUSAL_LM",
    modules_to_save=["lm_head", "embed_tokens"]
)
peft_model = get_peft_model(model, peft_config=lora_config)
peft_model.print_trainable_parameters()
```

If I rerun this cell with different configs, I get the following warning:
```
/usr/local/lib/python3.12/dist-packages/peft/mapping_func.py:72: UserWarning: You are trying to modify a model with PEFT for a second time. If you want to reload the model with a different config, make sure to call `.unload()` before.
  warnings.warn(
/usr/local/lib/python3.12/dist-packages/peft/tuners/tuners_utils.py:285: UserWarning: Already found a `peft_config` attribute in the model. This will lead to having multiple adapters in the model. Make sure to know what you are doing!
  warnings.warn(
```

However, running the following won't cut it, as it doesn't remove the `.peft_config` from the `model` object:
```python
model = peft_model.unload()
```

That means I have to add something like this at the top of the previous cell:
```python
# Unload the adapter if we are re-running the cell
try:
    model = peft_model.unload()
    del model.peft_config # explicitly deleting the config
except (NameError, AttributeError):
    pass
```

Is that intended behaviour? In the doc, `.unload()` is described as 
> Return[ing] the base model by removing all the PEFT modules.

I would expect the `.peft_config` attribute to be removed as well then?

Thanks for reading!

### Your contribution

I would be open to opening a PR (although some guidance would be useful).

BenjaminBossan:
Thanks for reporting this issue. I think we have two separate problems here:

1. The warning itself: Maybe it was cautious at the time but I think by now, we have sufficient knowledge about PEFT usage to say that this is not a rare use case. Personally, I'd be fine with removing it completely.
[...]

PR #3067, opened 2026-02-25
fix: clean up peft_config from model on unload() and merge_and_unload()

zamal-db:
## Problem

I was iterating on LoRA adapters in a training loop  swapping different adapter configs on the same base model  and kept hitting the warning:

> *You are trying to modify a model that has already been modified by PEFT. Please unload the model first...*

even though I was calling `.unload()` between each wrap. After digging in, I found that `unload()` restores the original model weights correctly but **never cleans up the `.peft_config` attribute** left on the model. So downstream, `get_peft_model()` still sees stale PEFT artifacts and fires the warning.

## Fix

Added cleanup of `peft_config` in `_unload_and_optionally_merge()`, which is the shared path for both `unload()` and `merge_and_unload()`. After the base model is restored, we now delete this attribute so the model is truly back to a clean state.

As the maintainer noted in #3025, a simple `del` at the right time is enough here.

## Tests

Added 3 focused tests in `tests/test_other.py`:

- **`test_unload_removes_peft_config`**  verifies `peft_config` is removed after `unload()`
- **`test_merge_and_unload_removes_peft_config`**  same for `merge_and_unload()`
- **`test_unload_then_rewrap_no_warning`**  confirms no spurious warning when re-wrapping a properly unloaded model

All existing tests pass (799 passed, 0 failed).

Relates to #3025

### DIFF
--- a/src/peft/tuners/ln_tuning/model.py
+++ b/src/peft/tuners/ln_tuning/model.py
@@ -123,6 +123,11 @@ class LNTuningModel(BaseTuner):
                     target.merge(adapter_names)
                 self._replace_module(parent, target_name, target.get_base_layer(), target)
 
+        # Clean up peft_config from the model since all PEFT modules have been removed.
+        # This prevents spurious warnings when re-wrapping the model with get_peft_model().
+        if hasattr(self.model, "peft_config"):
+            del self.model.peft_config
+
         return self.model
 
     def _cast_adapter_dtype(self, adapter_name: str, autocast_adapter_dtype: bool = True) -> None:
--- a/src/peft/tuners/mixed/model.py
+++ b/src/peft/tuners/mixed/model.py
@@ -267,6 +267,11 @@ class MixedModel(BaseTuner):
                     new_module = new_module.get_base_layer()
                 setattr(parent, target_name, new_module)
 
+        # Clean up peft_config from the model since all PEFT modules have been removed.
+        # This prevents spurious warnings when re-wrapping the model with get_peft_model().
+        if hasattr(self.model, "peft_config"):
+            del self.model.peft_config
+
         return self.model
 
     def add_weighted_adapter(self, *args: Any, **kwargs: Any) -> None:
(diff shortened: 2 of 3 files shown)


BenjaminBossan:
@jchwenger This should solve your issue. If you have time, you could try out this branch.

jchwenger:
Hi @BenjaminBossan, thanks a lot for this! I'm sorry I didn't have the time to look into it, but great to see it sorted.

tests/test_other.py
```diff
@@ -633,3 +636,62 @@ def test_get_modules_tied_to_embedding_peft(self, model_id, tied_weights_type):
             modules = peft_model._get_module_names_tied_with_embedding()
 
             assert expected == modules
+
+
+class TestUnloadCleanup:
```
BenjaminBossan: Thanks for adding tests but I think it's easier (and provides more coverage) to extend existing tests. For `model.unload()`, we can add an assert here:

https://github.com/huggingface/peft/blob/6a27da325c67c256ce48d3b8918ec918370b74c7/tests/testing_common.py#L1438

For `model.merge_and_unload()`, an assert here should do:

https://github.com/huggingface/peft/blob/6a27da325c67c256ce48d3b8918ec918370b74c7/tests/testing_common.py#L584

As for testing the warning, I think it's fine not having an explicit test for that. It's probably better to remove these warnings anyway (in a separate PR).

BenjaminBossan (changes requested):
Thanks for the PR. Looks good overall, I just added a suggestion to simplify the tests.

tests/test_other.py
zamal-db: Good call, moved the asserts into the existing `_test_unload` and `_test_merge_and_unload` in `testing_common.py` and dropped the standalone class from `test_other.py`. Also removed the warning re-wrap test as you suggested.

zamal-db:
Done, added the cleanup to `ln_tuning/model.py`. I also noticed `mixed/model.py` has the same pattern (own override, doesn't call super) so I added it there too.

Updated both test assertions to `{"default": "LORA"}` as you suggested.

PR #3069, opened 2026-02-25
fix: replace 2 bare except clauses with except Exception

haosenwang1018:
## What
Replace 2 bare `except:` clauses with `except Exception:`.

## Why
Bare `except:` catches `BaseException`, including `KeyboardInterrupt` and `SystemExit`, which can prevent clean process shutdown and mask critical errors.


githubnemo:
I'd argue this is fine. This is supposed to catch everything. Also, both `SystemExit` and `KeyboardInterrupt` inherit from `BaseException` so this makes no difference.

I'm closing this but feel free to re-open if you have pressing concerns about the correctness.

PR #3078, opened 2026-03-03
[method_comparison] RL training based method comparison for lora adapters

kashif:


method_comparison/RLMath/README.md
```diff
@@ -0,0 +1,54 @@
...
+1. Policy-gradient-style objective with importance weighting.
+2. GRPO-like group-relative reward centering (multiple completions per prompt).
+3. Optional KL regularization to a reference policy via `beta`.
+
+Reference: https://thinkingmachines.ai/blog/lora/
+
+## Initial datasets
+
+1. `openai/gsm8k` (default, fast iteration)
+2. `hendrycks/competition_math` (blog-aligned harder setting)
```
githubnemo: Do you think we can reduce this to only gsm8k? One of the crucial aspects of this benchmark setup is to have something that we can run at regular intervals. Maintaining more datasets to run would increase the required time and cost to re-run the experiments.

method_comparison/RLMath/run.py
```diff
@@ -0,0 +1,277 @@
...
+    rewards = 0.0
+
+    for row in dataset:
+        inputs = tokenizer(row["prompt"], return_tensors="pt").to(model.device)
+        outputs = model.generate(
+            **inputs,
+            do_sample=False,
+            max_new_tokens=max_completion_length,
+            pad_token_id=tokenizer.pad_token_id,
+        )
```
githubnemo, line 110: This should probably be wrapped in the `torch.inference()` context manager depending on a new `inference: bool=False` parameter.

method_comparison/RLMath/run.py
```diff
@@ -0,0 +1,277 @@
...
+        valid_pass_at_1, valid_reward = evaluate_pass_at_1(
+            model=model,
+            tokenizer=tokenizer,
+            dataset=valid_ds,
+            max_completion_length=cfg.max_completion_length,
+        )
+
+        log_history = trainer.state.log_history
+        last_log = next((x for x in reversed(log_history) if "reward" in x), {})
```
githubnemo, line 457:
```suggestion
        metrics = next((x for x in reversed(log_history) if "reward" in x), {})
```

Does this already include memory stats and eval time or do we need to track this on our own?

method_comparison/RLMath/run.py
```diff
@@ -0,0 +1,277 @@
...
+            dtype = "auto"
+        grpo_args.model_init_kwargs = {"dtype": dtype}
+
+        trainer = GRPOTrainer(
+            model=cfg.model_id,
+            reward_funcs=make_reward_func(),
+            args=grpo_args,
+            train_dataset=train_ds,
+            eval_dataset=None,
+            peft_config=peft_cfg,
```
githubnemo, line 399: Is it possible to have a training step callback? For AdaLoRA we need to call `model.base_model.update_and_allocate(step_idx)`.

githubnemo, line 399: Regarding initialziation of the PEFT model: can we take control here? It might, for example, be needed to pass `autocast_adapter_dtype` to `get_peft_model` for certain models.

BenjaminBossan:
Not stale. Just to check in, what's needed to get this PR merged?

BenjaminBossan:
@kashif LMK once this is ready for review.

BenjaminBossan:
_not stale_

BenjaminBossan:
_not stale_

Issue #3044, opened 2026-02-16
TP support for Finetuning using LoRA and other PEFT techniques


quic-meetkuma:
### Feature request

As we know the transformers repo has modeling files for all the models. Currently transformers repo also supports TP training using base_tp_plan (defined in configuration_<model_name>.py for each model). This supports original model. So whenever we load the model using AutoModel.from_pretrained() and give tp_plan='auto' then it modifies the original model for a given TP scheme. 

But because of the nature of the PEFT library, we need to convert that into PEFT model after the original model is loaded using AutoModel.from_pretrained(). Hence the get_peft_model() will just introduce just the PEFT layers but it will not apply any TP sharding to it. 

I think better approach is to derive the TP strategy based on original layers. E.g. If the base layer is ColwiseParallel (as per ColwiseParallel class of tensor_parallel.py of transformers repo), then the lora_A should be ColwiseParallel and lora_B should be RowwiseParallel. This modification should have been taken care by peft library whenever get_peft_model() is used.

### Your contribution

If this is something which is genuinely missing (in my opinion it is), then I will come up with a PR which applies TP to the PEFT layers within the get_peft_model() call.

BenjaminBossan:
Thanks for bringing up this issue. I haven't really tried TP so far and you're right, it's not working in PEFT yet. Here is my test:

```python
# torchrun --nproc-per-node 2 test.py
import torch
from transformers import AutoModelForCausalLM
from peft import LoraConfig, get_peft_model

model_id = "meta-llama/Llama-3.2-3B-instruct"
model = AutoModelForCausalLM.from_pretrained(model_id , dtype=torch.bfloat16, tp_plan="auto")
print(model._tp_plan)

model = get_peft_model(model, LoraConfig(init_lora_weights=False))

inputs = torch.arange(1000).view(1, -1).to(model.device)
outputs = model(inputs)
```

With PEFT enabled, this fails with:

```
[rank0]:     result = result + lora_B(lora_A(dropout(x))) * scaling
[rank0]:              ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[rank0]: RuntimeError: The size of tensor a (1536) must match the size of tensor b (3072) at non-singleton dimension 2
```

Without PEFT, it succeeds.

[...]

PR #3079, opened 2026-03-03
LoRA and Transformers TP

michaelbenayoun:
The goal of this PR is to integrate Transformers' API for Tensor Parallelism to PEFT, starting by LoRA. 

As #3044 pointed out, there are issues. 

First, the code used fails. It was due to the fact that to create adapters PEFT looks at the parent modules attributes, and not the actual weights. 

For instance for LoRA, it will check the `in_features` and `out_features` attributes of `torch.nn.Linear`, instead of the weight's shape, failing in the case of TP because there is a mismatch. **I addressed this issue here: https://github.com/huggingface/transformers/pull/44421.**

On top of that, we need to handle a few things to make it work. There are two cases:

- **Column Linears**: in this case, the output is sharded. So our adapters should also produce sharded outputs. We should have `lora_A` to be a regular non-sharded linear, and `lora_B` to be a column linear, just as the base layer.
- **Row Linears**: in this case, the input comes sharded, and we should produce a un-sharded output. We should have `lora_A` to be a row linear, and `lora_B` to be a regular non-sharded linear.

To do that, we need to do multiple things:

- If we create the adapter from scratch (not loaded from an existing checkpoint), then we only need to add the Transformers `TensorParallelLayer`'s hooks
- If we load the adapter from an existing checkpoint, then we also need to shard the loaded weights accordingly

This PR provides such features and a test file to check that everything works as expected.

**Next**: add similar support for the `Embedding` layer.

### DIFF
--- a/tests/training/tp_config.yaml
+++ b/tests/training/tp_config.yaml
@@ -1,0 +1,9 @@
+compute_environment: LOCAL_MACHINE
+debug: false
+distributed_type: MULTI_GPU
+machine_rank: 0
+main_training_function: main
+mixed_precision: bf16
+num_machines: 1
+num_processes: 2
+use_cpu: false


BenjaminBossan:
Thanks for taking care of this @michaelbenayoun. LMK if I can help with anything. If you have a minimal example to test this, that would be great.

A different approach that may work is to detect if a TP plan is being used and then initialize the corresponding PEFT layer differently using the sharded layers, I'm not sure what approach would be more robust.

tests/test_lora_tp.py
```diff
@@ -0,0 +1,202 @@
...
+    os.environ["MASTER_ADDR"] = "localhost"
+    os.environ["MASTER_PORT"] = str(port)
+    dist.init_process_group(backend="gloo", rank=rank, world_size=world_size)
+
+
+def _teardown_dist():
+    dist.destroy_process_group()
+
+
+def _test_training(rank, world_size, port):
```
3outeille: is it possible to test using the following methodology ? https://github.com/huggingface/transformers/blob/main/tests/test_training_mixin.py#L387 

take a tiny random model, overfit the same sample `abcdefg...` for several steps until the loss and grad_norm has decreased of 70% -> save and load -> generate ? (since it has overfit the sample, it should perfectly predict the sequence)

Image

src/peft/tuners/lora/layer.py
```diff
@@ -256,29 +256,43 @@ def update_layer(
...
+        # init_lora_weights, since each rank initializes weights independently.
+        # We could skip some broadcast, for instance when the lora weights are initialized to zero,
+        # but this is a minor optimization and would add extra complexity to the code.
+        if torch.distributed.is_available() and torch.distributed.is_initialized():
+            base_layer = self.get_base_layer()
+            tp_plan = getattr(base_layer, "_hf_tp_plan", None)
+            device_mesh = getattr(base_layer, "_hf_device_mesh", None)
+            if device_mesh is not None and tp_plan in ("colwise", "rowwise"):
+                pg = device_mesh.get_group()
+                src = torch.distributed.get_global_rank(pg, 0)
```
3outeille: styling but better to `import torch.distributed as dist` then use `dist` (be consistent on the transformer side) ?

3outeille:
Nice PR ! Several questions that pops into my mind:

- Regarding TP Row linear + lora,  should the `all_reduce` happens before or after the lora computation ?
- Given how importance of MoE as well, that could be good to have a TP + Peft + Moe support, what do you think?

tests/test_lora_tp.py
michaelbenayoun: Done

src/peft/tuners/lora/layer.py
michaelbenayoun: Done

michaelbenayoun:
> Nice PR ! Several questions that pops into my mind:
> 
> * Regarding TP Row linear + lora,  should the `all_reduce` happens before or after the lora computation ?
> * Given how importance of MoE as well, that could be good to have a TP + Peft + Moe support, what do you think?

* TP Row linear + Lora will do this: Lora A gets sharded inputs, computes the output and all reduce, just like a regular RowLinear. Then Lora B gets un-sharded inputs.
* I agree, as well as for Embeddings. I suggest we postpone that to other PRs to not make this one grow.

src/peft/utils/save_and_load.py
```diff
@@ -579,6 +584,43 @@ def renamed_dora_weights(k):
...
+
+                    tp_plan = getattr(base_layer, "_hf_tp_plan", None)
+                    device_mesh = getattr(base_layer, "_hf_device_mesh", None)
+
+                    if tp_plan is None or device_mesh is None:
+                        continue
+
+                    # We create and initialize the TensorParallelLayer on the fly,
+                    # and we set the `empty_param` attribute depending on the proper
+                    # state dict key to shard
```
BenjaminBossan: Could you please extend this comment as to why this is needed?

src/peft/utils/save_and_load.py
```diff
@@ -579,6 +584,43 @@ def renamed_dora_weights(k):
...
+                        tp_layer.empty_param = peft_model_state_dict[key]
+                        peft_model_state_dict[key] = tp_layer.shard_tensor(
+                            peft_model_state_dict[key], device=device, dtype=dtype
+                        )
+                    elif isinstance(tp_layer, RowwiseParallel):
+                        key = f"{name}.lora_A.{adapter_name}.weight"
+                        tp_layer.empty_param = peft_model_state_dict[key]
+                        peft_model_state_dict[key] = tp_layer.shard_tensor(
+                            peft_model_state_dict[key], device=device, dtype=dtype
+                        )
```
BenjaminBossan: Should do the same, right?

```suggestion
                    if isinstance(tp_layer, ColwiseParallel):
                        key = f"{name}.lora_B.{adapter_name}.weight"
                    elif isinstance(tp_layer, RowwiseParallel):
                        key = f"{name}.lora_A.{adapter_name}.weight"
                    tp_layer.empty_param = peft_model_state_dict[key]
                    peft_model_state_dict[key] = tp_layer.shard_tensor(
                        peft_model_state_dict[key], device=device, dtype=dtype
                    )
```

src/peft/tuners/lora/model.py
```diff
@@ -268,6 +271,37 @@ def _create_and_replace(
...
+                )
+            elif tp_plan == "rowwise":
+                add_tensor_parallel_hooks_to_module(
+                    self.model,
+                    lora_module.lora_A[adapter_name],
+                    tp_plan,
+                    f"{current_key}.lora_A.{adapter_name}",
+                    tp_plan,
+                    device_mesh,
+                )
```
BenjaminBossan: Should do the same but with less repetition, right?

```suggestion
            if tp_plan == "colwise":
                tp_module = lora_module.lora_B[adapter_name]
                tp_layer_name = f"{current_key}.lora_B.{adapter_name}",
            else:
                tp_module = lora_module.lora_A[adapter_name],
                tp_layer_name = f"{current_key}.lora_A.{adapter_name}",
            add_tensor_parallel_hooks_to_module(
                model=self.model,
                module=tp_module,
                tp_plan=tp_plan,
                layer_name=tp_layer_name,
                current_module_plan=tp_plan,
                device_mesh=device_mesh,
            )
```

Also, it looks like the `tp_plan` argument to `add_tensor_parallel_hooks_to_module` is not used at all, why is it needed?

tests/test_lora_tp.py
```diff
@@ -0,0 +1,393 @@
+# Copyright 2025-present the HuggingFace Inc. team.
```
BenjaminBossan:
```suggestion
# Copyright 2026-present the HuggingFace Inc. team.
```

tests/test_lora_tp.py
```diff
@@ -0,0 +1,393 @@
...
+from transformers import AutoModelForCausalLM, AutoTokenizer, set_seed
+from transformers.testing_utils import ColoredFormatter, Colors
+
+from peft import LoraConfig, PeftModel, get_peft_model
+from peft.tuners.lora import LoraLayer
+
+
+WORLD_SIZE = 2
+MODEL_ID = "Qwen/Qwen3-0.6B"
+TINY_MODEL_ID = "amazingvince/zephyr-smol_llama-100m-sft-full"
```
BenjaminBossan: Is there a specific reason to use these models? Otherwise, I'd like to move to models we're already using as they are already cached (the CI is already close to triggering rate limits from the Hub so we have to be careful).

tests/test_lora_tp.py
```diff
@@ -0,0 +1,393 @@
...
+
+def _test_function_wrapper(fn, rank, world_size, port, *extra_args):
+    try:
+        _setup_dist(rank, world_size, port)
+        fn(rank, world_size, port, *extra_args)
+    finally:
+        _teardown_dist()
+
+
+def _test_training_overfit(rank, world_size, port):
```
BenjaminBossan: This looks more like an integration test to me. We could put a separate script for this into `tests/training/` and then invoke it here:

https://github.com/huggingface/peft/blob/2513f57134418cc48e6e376c9361d8e2465b300d/Makefile#L65

This ensures it's only run in the correct context and when it's needed (not part of the regular CI, which runs on CPU).

tests/test_lora_tp.py
```diff
@@ -0,0 +1,393 @@
...
+    logger.info(
+        f"{Colors.GREEN}✓ Grad norm decreased by more than {TEST_OVERFIT_GRAD_NORM_REDUCTION_THRESHOLD * 100:.0f}%{Colors.RESET}"
+    )
+
+    # Assert generation matches
+    assert generation_matches, "Expected model to generate the training sequence after overfitting"
+    logger.info(f"{Colors.GREEN}✓ Generated sequence matches training sequence{Colors.RESET}")
+
+
+def _test_lora_weight_synchronization(rank, world_size, port):
```
BenjaminBossan: This test and the rest below should be put into `tests/test_gpu_examples.py`. Let's put them all in the same test class to make it clear they belong together. Then decorate the class with `@pytest.mark.multi_gpu_tests`. That way, we know that the tests only run on the multi GPU runner.

tests/test_lora_tp.py
```diff
@@ -0,0 +1,393 @@
...
+    try:
+        import transformers.integrations.tensor_parallel  # noqa: F401
+
+        return True
+    except ImportError:
+        return False
+
+
+@unittest.skipUnless(_is_tp_available(), "transformers TP integration not available")
+class TestLoraTP(unittest.TestCase):
```
BenjaminBossan:
```suggestion
@pytest.mark.skipf(not _is_tp_available(), reason="transformers TP integration not available")
class TestLoraTensorParallel:
```

As we're away from unittest.

src/peft/utils/save_and_load.py
michaelbenayoun: Done.

src/peft/utils/save_and_load.py
michaelbenayoun: Done

src/peft/tuners/lora/model.py
michaelbenayoun: You are right about your question about the `tp_plan` parameter. I opened a PR here: https://github.com/huggingface/transformers/pull/44768. It should be merge before the release so I will make sure to include the changes here if it happens.

michaelbenayoun: Other refactor mentioned, done.

tests/test_lora_tp.py
michaelbenayoun: Done

tests/test_lora_tp.py
michaelbenayoun: Alright I get you. 
For `MODEL_ID`, the idea is just to use a small qwen3 model. We can use any LLM tbh.

For `TINY_MODEL_ID`, it is a bit more nuanced. Because we are running training steps, and we check that overfitting happens, we cannot simply take a tiny randomly initialized model because there's a glass ceiling in what the LoRA adapters can adapt when the base model is just full of garbage. So we need an actual train model, small enough to run fast in the CI, and I managed to find this small 100m finetuned model.

tests/test_lora_tp.py
michaelbenayoun: It also works on any runner with multiple CPU cores. But I did exactly as you suggested.

tests/test_lora_tp.py
michaelbenayoun: Done.

docs/source/developer_guides/lora.md
```diff
@@ -616,6 +616,26 @@ model = model.unload()
...
+
+Usage is identical to the standard LoRA workflow — simply load the base model with a `tp_plan` before wrapping it with PEFT:
+
+```py
+from transformers import AutoModelForCausalLM
+from peft import get_peft_model, LoraConfig
+
+model = AutoModelForCausalLM.from_pretrained("meta-llama/Llama-3.1-8B", tp_plan="auto")
+lora_config = LoraConfig(r=16, target_modules=["q_proj", "v_proj"])
+model = get_peft_model(model, lora_config)
```
BenjaminBossan, line 634: I got this to run (including a forward pass) after commenting out the transformers v5.4.0 check using the latest main branch.

tests/test_gpu_examples.py
```diff
@@ -5951,3 +5956,168 @@ def test_te_lora_backward(self, model_with_te_layers, tokenized_inputs):
         assert torch.isfinite(loss), f"Loss is not finite: {loss.item()}"
         loss.backward()
         optimizer.step()
+
+
+### LoRA and Tensor Parallelism tests ###
+
+WORLD_SIZE = 2
+MODEL_ID = "Qwen/Qwen3-0.6B"
+TINY_MODEL_ID = "amazingvince/zephyr-smol_llama-100m-sft-full"
```
BenjaminBossan: > For TINY_MODEL_ID, it is a bit more nuanced. Because we are running training steps, and we check that overfitting happens, we cannot simply take a tiny randomly initialized model because there's a glass ceiling in what the LoRA adapters can adapt when the base model is just full of garbage. So we need an actual train model, small enough to run fast in the CI, and I managed to find this small 100m finetuned model.

I see. Generally, we want to avoid loading models from "random" users as they can change at any point in time (including changing the model file to pickle). I thus copied the model to `peft-internal-testing/zephyr-smol_llama-100m-sft-full`, you should be able to use that one instead.

tests/test_gpu_examples.py
```diff
@@ -5951,3 +5956,168 @@ def test_te_lora_backward(self, model_with_te_layers, tokenized_inputs):
...
+    model.eval()
+    with torch.no_grad():
+        for adapter_name in ["adapter_a", "adapter_b"]:
+            model.set_adapter(adapter_name)
+            outputs = model(**inputs, labels=inputs["input_ids"])
+            assert torch.isfinite(outputs.loss), f"Loss not finite with adapter '{adapter_name}': {outputs.loss}"
+
+
+@pytest.mark.skipif(not _is_tp_available(), reason="transformers TP integration not available")
+class TestLoraTensorParallel:
```
BenjaminBossan, line 6151: These tests fail for me with:

```
E         transformers/integrations/tensor_parallel.py", line 81, in initialize_tensor_parallelism
E           current_device.set_device(int(os.environ["LOCAL_RANK"]))
E                                         ~~~~~~~~~~^^^^^^^^^^^^^^
E         File "<frozen os>", line 717, in __getitem__
E       KeyError: 'LOCAL_RANK'
```

michaelbenayoun: Fixed now.

tests/test_gpu_examples.py
michaelbenayoun: Done, and thanks for doing it!

BenjaminBossan: Still needs updating here.

pyproject.toml
```diff
@@ -41,7 +41,7 @@ doctest_optionflags = [
     "NUMBER",
 ]
 
-addopts = ["--cov=src/peft", "--cov-report=term-missing", "--durations=10"]
+# addopts = ["--cov=src/peft", "--cov-report=term-missing", "--durations=10"]
```
BenjaminBossan: Please undo.

tests/training/lora_tp.py
BenjaminBossan, line 1: Test passes on my machine. However, we have yet to invoke, so let's add `torchrun --nproc_per_node=2 tests/training/lora_tp.py --model_id "peft-internal-testing/zephyr-smol_llama-100m-sft-full"` here:

https://github.com/huggingface/peft/blob/c1e8a27a87331e79b0fd504ef1295922e257c902/Makefile#L74

Don't add the ` -- $(if $(IS_GITHUB_CI), ...` part.

tests/training/lora_tp.py
```diff
@@ -0,0 +1,228 @@
...
+
+    torch.cuda.set_device(rank)
+    device = torch.device("cuda", rank)
+    model = model.to(device)
+
+    lora_config = LoraConfig(r=4, target_modules=target_modules)
+    model = get_peft_model(model, lora_config)
+    model.train()
+
+    sample_input = tokenizer("Paris is the most beautiful city in the world.", return_tensors="pt")
```
BenjaminBossan, line 92:
```python
    sample_input = tokenizer("Berlin is the most beautiful city in the world.", return_tensors="pt")
```

okay, who am I kidding...

tests/test_gpu_examples.py
```diff
@@ -5951,3 +5962,185 @@ def test_te_lora_backward(self, model_with_te_layers, tokenized_inputs):
...
+    not is_transformers_ge_v5_4_0, reason="transformers TP integration supported for transformers >= 5.4.0"
+)
+class TestLoraTensorParallel:
+    def _spawn(self, fn, *extra_args, port_offset=0):
+        port = _BASE_PORT + port_offset
+        wrapped_fn = partial(_test_function_wrapper, fn)
+        mp.spawn(wrapped_fn, args=(WORLD_SIZE, port) + extra_args, nprocs=WORLD_SIZE, join=True)
+
+    @require_torch_gpu
+    @pytest.mark.multi_gpu_tests
```
BenjaminBossan: Nit: Decorator can be added to the class instead of having to add it to each function.

pyproject.toml
michaelbenayoun: Sorry was not intended to be pushed.

tests/test_gpu_examples.py
michaelbenayoun: Done

tests/training/lora_tp.py
michaelbenayoun: Done

Makefile
```diff
@@ -72,3 +72,4 @@ tests_training:
 	accelerate launch --config_file tests/training/fsdp2_config.yaml tests/training/training.py --quant 4bit -- $(if $(IS_GITHUB_CI),--report-log "training_fsdp2_4bit.log",)
 	accelerate launch --config_file tests/training/fsdp2_config.yaml tests/training/training.py --quant 4bit --target_modules q_proj --target_parameters v_proj.weight -- $(if $(IS_GITHUB_CI),--report-log "training_fsdp2_target_params.log",)
 	accelerate launch --config_file tests/training/fsdp_config.yaml tests/training/adapters.py -- $(if $(IS_GITHUB_CI),--report-log "training_fsdp_adapters.log",)
+	accelerate launch --config_file tests/training/tp_config.yaml tests/training/lora_tp.py
```
BenjaminBossan, line 75: The config file is missing, could you please add it?

michaelbenayoun: My bad, done.

BenjaminBossan (approved):
Great, big thanks for adding tensor parallel support to LoRA, LGTM.

PR #3081, opened 2026-03-04
Ensure that te.pytorch exists

githubnemo:
It's possible to install transformer_engine without the pytorch option which will break the TE integration in the LoRA tuner since `transformer_engine.pytorch` doesn't exist.

This is a minor annoyance but since I stumbled upon it I think it is better fixed than left open.

Reproduction:

```
$ pip install transformer_engine[core]

# run PEFT unit tests, for example, observe error:

---
        if is_te_available() and isinstance(                                                                                                                          
>           target_base_layer, (te.pytorch.LayerNormLinear, te.pytorch.LayerNormMLP, te.pytorch.Linear)                                                               
                                ^^^^^^^^^^                                                                                                                            
        ):                                                                                                                                                            
E       AttributeError: module 'transformer_engine' has no attribute 'pytorch'                                                                                        
---
```

Once you install `transformer_engine[pytorch]` the error disappears.

### DIFF
--- a/docker/peft-gpu/Dockerfile
+++ b/docker/peft-gpu/Dockerfile
@@ -26,7 +26,7 @@ RUN chsh -s /bin/bash
 SHELL ["/bin/bash", "-c"]
 
 # Stage 2
-FROM nvidia/cuda:12.8.1-devel-ubuntu22.04 AS build-image
+FROM nvidia/cuda:12.8.1-cudnn-devel-ubuntu22.04 AS build-image
 COPY --from=compile-image /opt/conda /opt/conda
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Hmm, just wondering: If a user wants to use TE and installs it without `[pytorch]`, then TE would simply not be applied, right? I wonder if we should at least have a warning?

githubnemo:
> Hmm, just wondering: If a user wants to use TE and installs it without `[pytorch]`, then TE would simply not be applied, right? I wonder if we should at least have a warning?

This is already handled by transformer engine for us.

I've also added TE[pytorch] to the GPU docker file, as discussed internally.

PR #3084, opened 2026-03-08
[FEAT] Add PEANuT to peft

yibozhong:
## PR Summary

This PR wants to integrate PEANuT (Parameter-Efficient Adaptation with Weight-aware Neural Tweakers, accepted to KDD 2026) into PEFT.

arxiv link: https://arxiv.org/abs/2410.01870

## Method

In LoRA, the update is directly parameterized as:

- `ΔW = A @ B`

image

In PEANuT, `ΔW` is generated by a **weight-aware neural tweaker** conditioned on the base weight `W`:

- `ΔW = Tweaker(W)`, where the core path is `W @ A @ ... @ B` (with optional intermediate transforms and non-linearity)

A key structural difference is the shape of `A`:

- LoRA typically uses `A: (in_features -> r)`
- PEANuT uses `A: (out_features -> r)`

So when `in_features > out_features`, PEANuT can be more parameter-efficient than LoRA at the same rank.

To increase depth of this nueral tweaker, we add intermediate layers with shape `r x r` in **residual pairs** (see figure above).  

Why this design: on our vision tasks, this paired residual form consistently gives better performance than `depth=2` (no intermediate layers), while conventional residual connection designs did not reliably provide such gains. So we keep this as the default depth-extension design.

### Pros
- Higher theoretical expressiveness than linear low-rank updates.
- Better performance than LoRA on a range of tasks under similar budgets.
- Works well in very low-parameter regimes (e.g., around 0.2M trainable params).

### Cons
- Higher memory usage than LoRA, because `ΔW` is explicitly constructed before applying it.
- Slower training/inference than LoRA, deeper intermediate layers further increase the overhead.
- The non-linearity requires more careful hyperparameter tuning (especially learning rate and related optimization settings).

## Changes in This PR

- Added PEANuT implementation under `src/peft/tuners/peanut`.
- Added multiple PEANuT tests under `tests`.
- Added finetuning example under `examples/peanut_finetuning`.

## Experiment Configs

Added two experiment configs under:

- `method_comparison/MetaMathQA/experiments/peanut/llama-3.2-3B-rank32-relu-depth2-s2.0`  
  - Default config  
  - ~7M trainable parameters  
  - Local machine accuracy: **45.8%**

- `method_comparison/MetaMathQA/experiments/peanut/llama-3.2-3B-rank1-relu-depth2-s32.0`  
  - Small-parameter config  
  - ~0.2M trainable parameters  
  - Local machine accuracy: **42.5%**

Please let me know if there are any issues!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,9 +63,12 @@ tests_torch_compile:
 	python -m pytest tests/test_torch_compile.py $(if $(IS_GITHUB_CI),--report-log "compile_tests.log",)
 
 tests_training:
-	accelerate launch --config_file tests/training/deepspeed_config.yaml tests/training/training.py $(if $(IS_GITHUB_CI),--report-log "training_deepspeed.log",)
-	accelerate launch --config_file tests/training/deepspeed_config.yaml tests/training/training.py --quant 4bit $(if $(IS_GITHUB_CI),--report-log "training_deepspeed_4bit.log",)
-	accelerate launch --config_file tests/training/deepspeed_config.yaml tests/training/training.py --quant 8bit $(if $(IS_GITHUB_CI),--report-log "training_deepspeed_8bit.log",)
-	accelerate launch --config_file tests/training/fsdp_config.yaml tests/training/training.py $(if $(IS_GITHUB_CI),--report-log "training_fsdp.log",)
-	accelerate launch --config_file tests/training/fsdp_config.yaml tests/training/training.py --quant 4bit $(if $(IS_GITHUB_CI),--report-log "training_fsdp_4bit.log",)
-	accelerate launch --config_file tests/training/fsdp_config.yaml tests/training/adapters.py $(if $(IS_GITHUB_CI),--report-log "training_fsdp_adapters.log",)
+	accelerate launch --config_file tests/training/deepspeed_config.yaml tests/training/training.py -- $(if $(IS_GITHUB_CI),--report-log "training_deepspeed.log",)
+	accelerate launch --config_file tests/training/deepspeed_config.yaml tests/training/training.py --quant 4bit -- $(if $(IS_GITHUB_CI),--report-log "training_deepspeed_4bit.log",)
+	accelerate launch --config_file tests/training/deepspeed_config.yaml tests/training/training.py --quant 8bit -- $(if $(IS_GITHUB_CI),--report-log "training_deepspeed_8bit.log",)
+	accelerate launch --config_file tests/training/fsdp_config.yaml tests/training/training.py -- $(if $(IS_GITHUB_CI),--report-log "training_fsdp.log",)
+	accelerate launch --config_file tests/training/fsdp_config.yaml tests/training/training.py --quant 4bit -- $(if $(IS_GITHUB_CI),--report-log "training_fsdp_4bit.log",)
+	accelerate launch --config_file tests/training/fsdp2_config.yaml tests/training/training.py -- $(if $(IS_GITHUB_CI),--report-log "training_fsdp2.log",)
+	accelerate launch --config_file tests/training/fsdp2_config.yaml tests/training/training.py --quant 4bit -- $(if $(IS_GITHUB_CI),--report-log "training_fsdp2_4bit.log",)
+	accelerate launch --config_file tests/training/fsdp2_config.yaml tests/training/training.py --quant 4bit --target_modules q_proj --target_parameters v_proj.weight -- $(if $(IS_GITHUB_CI),--report-log "training_fsdp2_target_params.log",)
+	accelerate launch --config_file tests/training/fsdp_config.yaml tests/training/adapters.py -- $(if $(IS_GITHUB_CI),--report-log "training_fsdp_adapters.log",)
--- a/docs/source/developer_guides/contributing.md
+++ b/docs/source/developer_guides/contributing.md
@@ -73,6 +73,15 @@ Please give a description of the circumstances that led to the bug. If there is 
 
 Ideally when a bugfix is provided, it should be accompanied by a test for the bug. The test should fail with the current code and pass with the bugfix. Add a comment to the test that references the issue or PR. Without a test, it is more difficult to prevent regressions in the future.
 
+## Documentation improvements
+
+We are happy to have fixes for broken links and missing or unclear documentation. Taking care of examples, making
+sure that they are up-to-date and running fine in this fast moving environment is also highly appreciated.
+
+Please refrain from sending pull requests that *only* correct typing errors as these generally create more work
+than they safe. Such changes are better combined with more substantial fixes (such as fixing broken links or
+extending/updating documentation).
+
 ## Add a new fine-tuning method
 
 New parameter-efficient fine-tuning methods are developed all the time. If you would like to add a new and promising method to PEFT, please follow these steps.
--- a/src/peft/tuners/adalora/model.py
+++ b/src/peft/tuners/adalora/model.py
@@ -264,8 +264,7 @@ class AdaLoraModel(LoraModel):
                 adapter_name,
                 rank,
                 lora_config.lora_alpha,
-                lora_config.lora_dropout,
-                lora_config.init_lora_weights,
+                config=lora_config,
             )
             with torch.no_grad():
                 if rank > 0:
--- a/tests/training/fsdp2_config.yaml
+++ b/tests/training/fsdp2_config.yaml
@@ -1,0 +1,22 @@
+compute_environment: LOCAL_MACHINE
+debug: false
+distributed_type: FSDP
+downcast_bf16: 'no'
+fsdp_config:
+  fsdp_version: 2
+  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
+  fsdp_cpu_ram_efficient_loading: false
+  fsdp_offload_params: false
+  fsdp_reshard_after_forward: true
+  fsdp_state_dict_type: SHARDED_STATE_DICT
+machine_rank: 0
+main_training_function: main
+mixed_precision: bf16
+num_machines: 1
+num_processes: 2
+rdzv_backend: static
+same_network: true
+tpu_env: []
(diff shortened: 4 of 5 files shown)


src/peft/tuners/peanut/layer.py
```diff
@@ -0,0 +1,267 @@
...
+            )
+
+        self.in_features = in_features
+        self.out_features = out_features
+
+    def _ensure_residual_layer_dicts(self, num_residual_pairs: int) -> None:
+        adapter_layer_names = list(self.adapter_layer_names)
+
+        for i in range(num_residual_pairs):
+            encoder_name = f"peanut_encoder_{i}"
```
BenjaminBossan: Would it be possible to create an `nn.ModuleList` for encoder and decoder and then just index with `i` instead of having to string-interpolate the attribute names? WDYT? Maybe even an `nn.Sequential` could be used for the tweaker, but that my require a custom `nn.Module`.

src/peft/tuners/peanut/layer.py
```diff
@@ -0,0 +1,267 @@
...
+from typing import Any, Optional
+
+import torch
+from torch import nn
+from transformers.activations import ACT2FN
+
+from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
+
+
+class PeanutLayer(BaseTunerLayer):
```
BenjaminBossan, line 26: Note that right now, the layer does not support `transformers.pytorch_utils.Conv1D` which is used by some older models like GPT2. If you want to support that, you'd have to set the `fan_in_fan_out` attribute (like in LoRA) and apply `peft.utils.other.transpose` in the weight (as the weight is not transposed, unlike in `nn.Linear`).

src/peft/tuners/peanut/layer.py
```diff
@@ -0,0 +1,267 @@
...
+            raise ValueError(f"Adapter {adapter} not found.")
+
+        peanut_A = self.peanut_A[adapter]
+        peanut_B = self.peanut_B[adapter]
+        non_linear = ACT2FN[self.act_fn[adapter]]
+        scaling = self.scaling[adapter]
+        res_num = self.res_num[adapter]
+
+        base_weight_t = base_weight.transpose(0, 1).to(peanut_A.weight.dtype)
+        delta_w = non_linear(torch.matmul(base_weight_t, peanut_A.weight.t()))
```
BenjaminBossan, line 147: I wonder if this can be done in a `torch.no_grad()` context to save a bit of memory?

src/peft/tuners/peanut/layer.py
```diff
@@ -0,0 +1,267 @@
...
+                    if not torch.isfinite(orig_weights).all():
+                        raise ValueError(
+                            f"NaNs detected in the merged weights. The adapter {active_adapter} seems to be broken"
+                        )
+
+                    base_layer.weight.data = orig_weights
+                else:
+                    base_layer.weight.data.add_(delta_weight)
+
+            self._cached_delta_weights[active_adapter] = delta_weight
```
BenjaminBossan: Why do we need to cache it? Can we not re-compute the `delta_weight` when we unmerge?

src/peft/tuners/peanut/layer.py
```diff
@@ -0,0 +1,267 @@
...
+
+            delta_weight = self._cached_delta_weights.pop(active_adapter, None)
+            if delta_weight is None:
+                raise ValueError(f"Cached delta weight for adapter '{active_adapter}' is missing; cannot unmerge.")
+
+            base_layer.weight.data.sub_(
+                delta_weight.to(dtype=base_layer.weight.dtype, device=base_layer.weight.device)
+            )
+
+    def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
```
BenjaminBossan, line 222: In general, using early returns like this is fine. However, in all other PEFT methods, we use `if ... elif ...` here and return the result only at the very end. For consistency, please update the method accordingly.

src/peft/tuners/peanut/config.py
```diff
@@ -0,0 +1,195 @@
...
+@dataclass
+class PeanutConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`PeanutModel`].
+
+    Args:
+        r (`int`):
+            PEANuT rank. This is the hidden dimension used by the adapters. Similar to LoRA rank, larger `r` increases
+            adapter capacity and trainable parameters.
+        depth (`int`):
```
BenjaminBossan, line 34: Just wondering what you think about changing the semantics here a bit: If `depth` means hidden layers per encoder/decoder, we would only need to ensure that `depth>0` (no more requirement for even number) and then add one layer for A _and_ for B. So my suggestion would be that e.g. `depth=3` in the new definition corresponds to `depth=8` (as (8 - 2) / 2 == 3) in the current definition. WDYT?

src/peft/tuners/peanut/layer.py
yibozhong: I initially did use a `ModuleList`-style implementation, but that changes parameter names from flat names like `...encoders_0.weight` to nested names like `...encoders.0.weight`. Some existing PEFT tests/utilities seem to rely on the current flat naming/layout, so the nested version ended up failing some tests. For now I still kept the flat structure for compatibility.

src/peft/tuners/peanut/layer.py
yibozhong: Agreed. Right now I am only targeting `nn.Linear` since in the experiments we are only adapting linear layers.

src/peft/tuners/peanut/layer.py
yibozhong: I applied `torch.no_grad()` in the function calling during merging. I did not wrap the shared delta-weight computation itself, since it is also used during forward/training and still needs gradients there.

src/peft/tuners/peanut/layer.py
yibozhong: I think caching is needed because the delta weight is conditioned on the pre-merge base weight. If we compute delta weight during unmerge from the already merged weight, we get a different delta weight, so subtracting it would not correctly recover the original weights.

src/peft/tuners/peanut/layer.py
yibozhong: Updated the forward function.

src/peft/tuners/peanut/config.py
yibozhong: Nice suggestion! I have updated the semantics, now `depth` means the number of `encoder-decoder` pairs with a default of 0 (i.e., no inner layers).

yibozhong:
Hi @BenjaminBossan I made some comments and changes accordingly. Please let me know what you think. Also, documentation for peanut has been added.

src/peft/tuners/peanut/layer.py
BenjaminBossan: Ah right, true. I wonder if it would make sense to offload the cached weight, since they have the full size of the base weights, right?

src/peft/tuners/peanut/layer.py
BenjaminBossan: Could you tell me where exactly this goes wrong? Nested parameters should work, if they fail, we may have to fix that.

yibozhong:
Hi @BenjaminBossan I found a simple fix in `src/peft/utils/save_and_load.py` that resolves the previous test failures (related to adapter name deleting). So I now use the ModuleList-based design (`peanut_encoders` / `peanut_decoders`) as you suggetsed and also add offload for cached weights.

I ran local tests and style checks, and everything passes. Let me know if there are any issues.

examples/peanut_finetuning/peanut_finetuning.py
BenjaminBossan, line 1: Thanks for the example. I could confirm it works, though I had to change a few settings for it run with only 24GB VRAM.

src/peft/tuners/peanut/layer.py
```diff
@@ -0,0 +1,248 @@
...
+        res_num = self.res_num[adapter]
+        peanut_encoders = self.peanut_encoders[adapter]
+        peanut_decoders = self.peanut_decoders[adapter]
+
+        base_weight_t = base_weight.transpose(0, 1).to(peanut_A.weight.dtype)
+        delta_w = non_linear(torch.matmul(base_weight_t, peanut_A.weight.t()))
+
+        residuals = []
+        for i in range(res_num):
+            residuals.append(delta_w.clone())
```
BenjaminBossan: Is the `.clone()` call really necessary here? As `delta_w` is not changed in-place, I think not. I removed it for testing and ran the MetaMath benchmark and the train losses were almost identical 1.0780 vs 1.0781 after 250 steps, small variations are normal). Not sure if it really makes a difference memory-wise but if it's not needed, let's remove it.

tests/test_custom_models.py
```diff
@@ -1019,6 +1020,27 @@
...
+    ),
+    ("Vanilla MLP 5 Peanut", "MLP", PeanutConfig, {"target_modules": "lin0", "r": 2, "depth": 1, "act_fn": "relu"}),
+    ("Vanilla MLP 6 Peanut", "MLP", PeanutConfig, {"target_modules": ["lin0"], "r": 2, "depth": 1, "act_fn": "relu"}),
+    ("Vanilla MLP 7 Peanut", "MLP", PeanutConfig, {"target_modules": ["lin1"], "r": 2, "depth": 1, "act_fn": "relu"}),
+    (
+        "Vanilla MLP 8 Peanut",
+        "MLP",
+        PeanutConfig,
+        {"target_modules": ["lin0", "lin1"], "r": 2, "depth": 1, "act_fn": "relu"},
+    ),
```
BenjaminBossan: I think it's fine having just a single setting with depth 1, checking different target modules is already covered by the depth 0 tests. But let's also choose a different non-linearity to cover that part.

yibozhong:
Thanks for the suggestions! I have made changes accordingly.

PR #3087, opened 2026-03-10
FIX Error with PSOFT fp16/bf16 on GPU

BenjaminBossan:
Torch's linalg solver requires float32. For CPU, half was already cast to float, but not for GPU. The CI runs on CPU, thus the error is not caught, but I could reproduce it locally and my fix solves it.

### DIFF
--- a/src/peft/tuners/psoft/layer.py
+++ b/src/peft/tuners/psoft/layer.py
@@ -131,9 +131,6 @@ class OrthLayer(nn.Module):
         else:
             Q = self._skew_symmetric()
             orig_dtype = Q.dtype
-            cast_to_fp32 = (Q.device.type == "cpu") and (orig_dtype in (torch.float16, torch.bfloat16))
-            if cast_to_fp32:
-                Q = Q.float()
 
             id_mat = torch.eye(self.size, device=Q.device, dtype=Q.dtype)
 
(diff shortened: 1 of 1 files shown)


src/peft/tuners/psoft/layer.py
githubnemo, line 137: Not this always uses float16, before it could use fp32 when `cast_to_fp32` was set. Why not just remove the `(Q.device.type == "cpu")` and keep the rest as-is?

BenjaminBossan: I don't see it, when `self.use_cayley_neumann is True`, we cast to half?

githubnemo: Hmm, what I meant was that in case of `auto_cast_adapter_dtype=False` this now behaves differently than before (the cayley neumann section isn't fp32 anymore). But it can be argued that this is truer to what the user requested, so I'm retracting my comment :)

BenjaminBossan: Ah I see, good point, but yes, I think with `auto_cast_adapter_dtype=False`, we should expect the base model dtype to be used here, so that was (accidentally) fixed too.

BenjaminBossan:
FYI @fei407

fei407:
Thanks for fixing this! I went through the updated logic and it looks much more consistent now. I appreciate the clarification and the improvements.

PR #3021, opened 2026-02-04
update get_error_factor to cache up with the latest transformers change

jiqing-feng:
Fix failed tests: `RUN_SLOW=1 pytest tests/test_gpu_examples.py::TestLoftQ`

```
FAILED tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit_iter_5[cpu] - assert tensor(2.9929e-09, grad_fn=<MeanBackward0>) < (tensor(2.0073e-09, grad_fn=<MeanBackward0>) / 1.005)
FAILED tests/test_gpu_examples.py::TestLoftQ::test_t5_loftq_8bit[cuda] - AssertionError: assert tensor(0.0868, device='cuda:0', grad_fn=<MeanBackward0>) < (tensor(0.0356, device='cuda:0', gr...
```

**Root Cause Analysis:**
The tests are failing due to changes in default `dtype` handling in `transformers` (ref: [PR #42805](https://github.com/huggingface/transformers/pull/42805)).

| Feature | **Old Behavior** | **New Behavior** |
| :--- | :--- | :--- |
| **Standard Model** | Loaded as `float32` | Loaded as `float32` |
| **Quantized Model** | Non-quantized parts (e.g., embeddings) loaded as **`fp16`** | Non-quantized parts loaded as **`float32`** |
| **Test Impact** | Tests compared `float32` vs `fp16` (inconsistent) | Tests now compare `float32` vs `float32` (consistent) |

**Conclusion:** The new behavior is more reasonable. We are updating our test suite to accommodate this change.

With this change, all TestLoftQ tests can pass on CUDA-A100/XPU/CPU.

More details see: https://github.com/huggingface/transformers/issues/43725

PR #3088, opened 2026-03-10
Partial fix for LoftQ + int8 quantization

githubnemo:
With transformers v5 the GPU examples started failing for LoftQ, mainly for 8bit quantization. This is also reported in PR #3021. The main reason for this is that with transformers v5 the dtype of the layers is consistent across the model whereas before some layers were in float16 and some in float32. This masked wrong behavior.

It turns out that LoftQ always assumed that if `n_bits==8` the weights are in fp8. In reality, at least for bitsandbytes, this is not true. bnb's `load_in_8bit` uses int8. This PR changes the quantization residual computation to be compatible with int8 if `n_bits==8`, so that at least the computed residual for the loftq initialization are correct.

This alone does not mean that LoftQ tests for 8bit are passing. They assume that the MSE of LoftQ 'corrected' models compared to an unquantized baseline model are at least 3% truer than the logits of the quantized reference model. For int8 this is not guaranteed since `bnb.nn.Linear8bitLt` not only quantizes the weights but it also quantizes the inputs in it's `.forward()`. The resulting quantization error cannot be compensated by the static LoftQ initialization. Since there is technically a way to compensate for this error, I've kept the LoftQ initialization for int8, added an example and changed the 8 bit tests to only assume rough parity with the quantized reference model.

The example (`./examples/loftq_finetuning/int8_correction.py`) that show-cases how to recover the activation quantization error can also be run with `--no-loftq` which skips the LoftQ initialization and uses the LoRA no-op initialization instead and the example still recovers the majority of the error (e.g. 90% improvement instead of 91% over quantized reference).

Therefore there are the following open tasks:

- discuss removal of int8 from LoftQ completely (since it can be argued that it doesn't seem to have such a big impact but this needs review and possibly better testing)
- discuss the now failing (marked as xfail) tests for nf4 for t5 when loftq_iter > 1. I don't have an intuition what's happening here but the error seems to roughly double per iteration.

### DIFF
--- a/tests/test_gpu_examples.py
+++ b/tests/test_gpu_examples.py
@@ -3141,10 +3141,6 @@ class TestLoftQ:
         mae_loftq = self.mae(logits_base, logits_loftq, attention_mask=inputs["attention_mask"])
         mse_loftq = self.mse(logits_base, logits_loftq, attention_mask=inputs["attention_mask"])
 
-        print(f"{bits=} {loftq_iter=} {device=} {model_id=}")
-        print(f"{mse_quantized=} vs. {mse_loftq=}")
-        print(f"{mae_quantized=} vs. {mae_loftq=}")
-
         return mae_quantized, mse_quantized, mae_loftq, mse_loftq
 
     @pytest.mark.parametrize("device", [torch_device, "cpu"])


examples/loftq_finetuning/int8_correction.py
```diff
@@ -0,0 +1,250 @@
+"""
+Script to show-case how to offset quantization error with LoRA / LoftQ
+when dealing with quantizations that both quantize weights and activations.
+This is the case for bnb int8, for example but also other quantizations
```
BenjaminBossan:
```suggestion
This is the case for bnb int8, for example, but also other quantizations
```

examples/loftq_finetuning/int8_correction.py
```diff
@@ -0,0 +1,250 @@
...
+    return torch.pow(a - b, 2).mean().item()
+
+
+def get_model(*args, **kwargs):
+    if task_type == TaskType.CAUSAL_LM:
+        return AutoModelForCausalLM.from_pretrained(*args, **kwargs)
+    return AutoModelForSeq2SeqLM.from_pretrained(*args, **kwargs)
+
+
+tokenizer = AutoTokenizer.from_pretrained(model_id)
```
BenjaminBossan, line 231: Just a suggestion to move this all into a `main` function to avoid computing all this if only `--help` is called.

examples/loftq_finetuning/int8_correction.py
```diff
@@ -0,0 +1,250 @@
...
+loftq_logits = get_logits(loftq_model, inputs)
+
+mse_loftq = mse(ref_logits, loftq_logits)
+mse_qref = mse(ref_logits, qref_logits)
+
+
+print(f"{model_id=}{device=}")
+print(f"{mse_qref=}, {mse_loftq=}")
+assert mse_loftq < (mse_qref / 1.05), f"{mse_loftq} >= {mse_qref / 1.05}"
+print(f"relative improvement: {(mse_qref - mse_loftq) / mse_qref * 100:.2f}%")
```
BenjaminBossan: IMO, "improvement" is a bit ambiguous here, how about "reduction of error through LoftQ"?

examples/loftq_finetuning/int8_correction.py
```diff
@@ -0,0 +1,250 @@
...
+The quantization error for the affected values is much lower and not
+static anymore, LoftQ is not able to deal with this. While technically
+possible, this script doesn't filter out these masked values and it
+is probably not worth the effort.
+
+Note: Some rudimentary testing showed that the LoftQ mitigation is still
+more effective than tuning threshold values but YMMV.
+
+Note: LoftQ is not doing the heavy lifting in this script's case. The error
+between LoftQ and zeroed LoRA is only about two percent points. This effect is
```
BenjaminBossan: How did you determine this 2%?

src/peft/utils/loftq_utils.py
```diff
@@ -195,8 +204,8 @@ def loftq_init(weight: Union[torch.Tensor, torch.nn.Parameter], num_bits: int, r
     else:
         raise ValueError("bitsandbytes is not available, please install it to use LoftQ.")
 
-    if num_bits not in [2, 4, 8]:
-        raise ValueError("Only support 2, 4, 8 bits quantization")
+    if num_bits not in [4, 8]:
+        raise ValueError("Only support nf4 and int8 quantization is supported")
```
BenjaminBossan:
```suggestion
        raise ValueError("Only nf4 and int8 quantization is supported")
```

tests/test_gpu_examples.py
```diff
@@ -2979,12 +2980,13 @@ class TestLoftQ:
...
-    def get_error_factor(self, device):
+    OPT_MODEL_ID = "facebook/opt-125m"
+
+    def get_error_factor(self, device, n_bits):
         # The error factor indicates by how much the quantization error should be decreased when using LoftQ compared to
         # quantization without LoftQ. Thus 1.03 means that the error should be decreased by 3% at least. This is a very
         # conservative value to prevent flakiness, in practice most gains are > 1.5
-        error_factor = 1.005 if device in ("xpu", "cpu") else 1.03
-        return error_factor
+        return 1.05 if device in ("xpu", "cpu") else 1.05
```
BenjaminBossan:
```suggestion
        return 1.05
```

tests/test_gpu_examples.py
```diff
@@ -3087,67 +3108,31 @@ def get_errors(
         mse_quantized = torch.pow(logits_base - logits_quantized, 2).mean()
         mae_loftq = torch.abs(logits_base - logits_loftq).mean()
         mse_loftq = torch.pow(logits_base - logits_loftq, 2).mean()
+
+        print(f"{bits=} {loftq_iter=} {device=} {model_id=}")
```
BenjaminBossan: Remove?

tests/test_other.py
```diff
@@ -633,3 +633,11 @@ def test_get_modules_tied_to_embedding_peft(self, model_id, tied_weights_type):
...
 
             assert expected == modules
+
+
+class TestLoftQDeprecation:
+    def test_nfquantizer_deprecation(self):
+        from peft.utils.loftq_utils import NFQuantizer
+
+        with pytest.warns(match="NFQuantizer is deprecated") as record:
+            _ = NFQuantizer()
```
BenjaminBossan:
```suggestion
            _ = NFQuantizer(device="cpu")
```

I think this should fix the CI error.

tests/test_other.py
```diff
@@ -633,3 +633,11 @@ def test_get_modules_tied_to_embedding_peft(self, model_id, tied_weights_type):
             modules = peft_model._get_module_names_tied_with_embedding()
 
             assert expected == modules
+
+
+class TestLoftQDeprecation:
```
BenjaminBossan, line 639: Let's add a TODO to remove this in 0.20.0

tests/test_gpu_examples.py
```diff
@@ -3156,49 +3141,56 @@ def test_bloomz_loftq_8bit_iter_5(self, device, tmp_path):
...
+        if n_bits == 4:
+            # next, check that LoftQ quantization errors are smaller than LoRA errors by a certain margin
+            error_factor = self.get_error_factor(device, n_bits)
+            assert mse_loftq < (mse_quantized / error_factor)
+            assert mae_loftq < (mae_quantized / error_factor)
+        elif n_bits == 8:
+            # for int8 compensation we're happy to achieve parity but without additional measures we will
+            # not be better since int8 also quantizes the layer input which introduces error loftq can't
+            # compensate
+            assert torch.allclose(mse_loftq, mse_quantized, atol=0.03)
```
BenjaminBossan: Just to be sure: The error may even increase in this case?

examples/loftq_finetuning/int8_correction.py
githubnemo: In that case `parse_args()` would call `sys.exit(1)` anyway right? So, no computation would be done.

tests/test_gpu_examples.py
githubnemo: Yes. I think it is a bit dependent on the number of input sequences, it seems that the variance is quite high. Maybe we should include more than one test sequence.

githubnemo:
@BenjaminBossan I added multiple input sequences to the tests to reduce the variance a bit. Unfortunately it turns out that we have to increase the variance (and the rank/alpha as it seems) a bit to have a meaningful effect. I'm a bit surprised that 4bit quantization was affected as well (at least a few tests failed with r=8).

I've also added padding masking in the MSE computation which increased some LoftQ scores but didn't change the overall results and margins too much.

PR #3090, opened 2026-03-11
Implement import allowlist in AutoPeftModel, limit access in megatron

### LINKED ISSUE #3085: Security: Arbitrary Module Import via Malicious Adapter Config (CWE-94)
## Security Report: Arbitrary Module Import via Malicious Adapter Config (Supply Chain Attack)

**Reporter:** Conner Webber (conner.webber000@gmail.com)
**Severity:** HIGH
**CWE:** CWE-94 (Improper Control of Generation of Code / Code Injection)

> **Note:** Private vulnerability reporting is not enabled on this repository. I attempted to file this via GHSA but received HTTP 403. Please consider enabling private vulnerability reporting for future security issues.

---

## Summary

PEFT's `AutoPeftModel.from_pretrained()` and LoRA TP layer contain unsanitized `importlib.import_module()` calls that allow arbitrary Python module loading via malicious adapter configs hosted on HuggingFace Hub.

---

## Vulnerability 1 — HIGH: Arbitrary Module Import via `auto_mapping.parent_library`

**File:** `peft/auto.py`, lines 106-108

When loading an adapter via `AutoPeftModel.from_pretrained()`, if `task_type` is `None` and `auto_mapping` is present in the adapter config, the `parent_library` string from the config JSON is passed directly to `importlib.import_module()` with zero validation:

```python
# peft/auto.py:106-108
parent_library = auto_mapping["parent_library"]
parent_library = importlib.import_module(parent_library)  # UNSANITIZED
model_cls = getattr(parent_library, auto_mapping["model_name"])  # attacker-controlled class
```

Line 109 then calls `getattr()` with the attacker-controlled `model_name` string, and `.from_pretrained()` is subsequently invoked on the resolved class.

### Attack Vector

1. Attacker creates a malicious PEFT adapter and uploads it to HuggingFace Hub
2. The `adapter_config.json` contains a crafted `auto_mapping` field with `parent_library` set to a module that executes code at import time (via `__init__.py`) or has an exploitable `.from_pretrained()` method
3. Victim runs `AutoPeftModel.from_pretrained("malicious-user/adapter-name")`
4. Arbitrary module is imported → code execution

### Suggested Fix
Allowlist `parent_library` to known-safe values (e.g., `transformers`, `diffusers`).

---

## Vulnerability 2 — MEDIUM: Arbitrary Module Import via `megatron_core` Config

**File:** `peft/tuners/lora/tp_layer.py`, line 325

The `lora_config.megatron_core` string from config JSON is passed to `importlib.import_module()` unsanitized:

```python
# peft/tuners/lora/tp_layer.py:325
megatron_core = importlib.import_module(lora_config.megatron_core)
```

This has a narrower trigger (requires `megatron_config` to also be set) but shares the same root cause: user-controlled config values fed directly to `importlib` without validation.

---

## Impact

- **Severity:** HIGH (CVSS ~8.8)
- **Attack complexity:** Low — attacker only needs to upload a crafted adapter config to HuggingFace Hub
- **User interaction:** Required — victim must call `from_pretrained()` with the malicious adapter identifier
- **Impact:** Arbitrary code execution on the victim's machine, full supply chain compromise
- **Scope:** Any PEFT user loading untrusted adapters from the Hub

## Proof of Concept

Create an `adapter_config.json` with:
```json
{
  "auto_mapping": {
    "parent_library": "os",
    "model_name": "system"
  }
}
```

While `os.system` isn't directly callable via `.from_pretrained()`, the import itself of a malicious package (or a package with side effects in `__init__.py`) achieves code execution. A realistic attack would use a typosquatted or attacker-controlled PyPI package name that is already installed in the victim's environment, or leverage a standard library module whose import triggers useful side effects.

## Recommended Fix

1. Add an allowlist for `parent_library` values (e.g., `{"transformers", "diffusers"}`) in `auto.py`
2. Add an allowlist for `megatron_core` values in `tp_layer.py`
3. Validate all config values before passing to `importlib.import_module()`

---

**Disclosure Timeline:**
- 2026-03-08: Report filed (public issue — GHSA private reporting disabled)
- 90-day disclosure deadline: 2026-06-06

githubnemo:
Resolves #3085. `AutoPeftModel` resolves the base model's class by importing the parent library mentioned in the adapter config which makes way for potentially executing malicious code on the adapter creator's behalf. While this is not a remote code execution vulnerability by itself (since the attacker still needs to install a malicious package on the system) it can be used for code execution in case the attacker has file system access, for example.

The same is happening for the `megatron.core` config key. I haven't found any evidence that `megatron` wants to load other modules than `megatron.*`, so we limit it to that.

There was a bit of missing documentation, typo, missing `hub_online_once`, so I took the opportunity to fix that along the way.

### DIFF
--- a/src/peft/tuners/lora/tp_layer.py
+++ b/src/peft/tuners/lora/tp_layer.py
@@ -30,6 +30,13 @@ from .config import LoraConfig
 from .layer import LoraLayer
 
 
+def get_default_module_allowlist():
+    """Allowed module names that config.megatron_core can reference for import"""
+    return [
+        "megatron",
+    ]
+
+
 class LoraParallelLinear(nn.Module, LoraLayer):
     """
     When the target layer parallel_linear is RowParallelLinear, in order to keep the input and output shapes
@@ -313,10 +320,11 @@ def dispatch_megatron(
         target_base_layer = target
 
     if config.megatron_config:
-        if not config.megatron_core.startswith("megatron."):
+        if not any(config.megatron_core.startswith(f"{prefix}.") for prefix in get_default_module_allowlist()):
             raise ValueError(
-                f"{config.megatron_core=} does not start with 'megatron.' which is unsupported and a potential "
-                "security risk. If you think it should be supported, please raise an issue in PEFT."
+                f"{config.megatron_core=} does not start with a valid prefix ({get_default_module_allowlist()}) "
+                "which is unsupported due to being a potential security risk. If you think it should be supported, "
+                "please raise an issue in PEFT."
             )
         megatron_core = importlib.import_module(config.megatron_core)
     else:
--- a/tests/test_lora_megatron.py
+++ b/tests/test_lora_megatron.py
@@ -190,4 +190,4 @@ if is_megatron_available():
 
         with pytest.raises(ValueError) as e:
             megatron_model = get_peft_model(model, lora_config)
-        assert "unsupported and a potential security" in str(e)
+        assert "unsupported due to being a potential security" in str(e)


githubnemo:
cc @spartan8806

src/peft/tuners/lora/tp_layer.py
```diff
@@ -313,6 +313,11 @@ def dispatch_megatron(
         target_base_layer = target
 
     if config.megatron_config:
+        if not config.megatron_core.startswith("megatron."):
```
BenjaminBossan: I don't remember the specifics anymore and after checking the PR from back then couldn't find anything, but I vaguely remember that there were explicit cases where another lib might be needed. How about solving this similarly to below with a patch-able allow list?

spartan8806:
Thanks for the quick turnaround.

src/peft/tuners/lora/tp_layer.py
githubnemo: Done

PR #3091, opened 2026-03-11
Embeddings LoRA & TP

michaelbenayoun:
To be merged after #3079 .

It adds support for LoRA embedding with TP.

cc @3outeille

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -13,8 +13,10 @@
 # limitations under the License.
 from __future__ import annotations
 
+import copy
 import math
 import warnings
+from collections.abc import Callable
 from contextlib import contextmanager
 from typing import Any, Optional, Union
 
@@ -25,6 +27,7 @@ import torch.nn.functional as F
 from torch import svd_lowrank
 from transformers.pytorch_utils import Conv1D
 
+from peft.import_utils import is_transformers_ge_v5_4_0
 from peft.tuners._buffer_dict import BufferDict
 from peft.tuners.tuners_utils import BaseTunerLayer, _get_in_out_features, check_adapters_to_merge
 from peft.utils.integrations import (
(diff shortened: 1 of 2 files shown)


src/peft/tuners/lora/layer.py
```diff
@@ -42,6 +43,8 @@
 
 VARIANT_KWARG_KEYS = ["alora_offsets"]
 
+logger = logging.getLogger(__name__)
```
BenjaminBossan: We don't do logger in PEFT :) Let's just use `warnings.warn`.

src/peft/tuners/lora/model.py
```diff
@@ -60,6 +63,9 @@
 from .tp_layer import dispatch_megatron
 
 
+logger = logging.getLogger(__name__)
```
BenjaminBossan: Same

src/peft/tuners/lora/model.py
```diff
@@ -285,23 +291,52 @@ def _create_and_replace(
...
+                    tp_plan,
+                    device_mesh,
+                )
+            elif tp_plan == "embedding_rowwise":
+                # LoRA embeddings are a bit special, there is no new module but just weights.
+                # To use the TP hooks machinery, we need to create a fake module that has the weights as attributes
+                # (used by the hooks), make the hooks use this fake module, and then add the hooks to the original
+                # `_embed` method acting as the forward pass lora_embedding_A.
+                tp_layer = copy.deepcopy(ALL_PARALLEL_STYLES[tp_plan])
+                mod = SimpleNamespace()
```
BenjaminBossan: Ah damn. Do you think the usage of `SimpleNamespace` is robust? Otherwise, we could consider using [`nn.utils.parametrize`](https://docs.pytorch.org/docs/stable/generated/torch.nn.utils.parametrize.register_parametrization.html).

michaelbenayoun: Parametrize will not do much here, what we need is a module holding the weights so that the input and output functions can use it.

I changed SimpleNamespace, which is very basic and simple, with something a bit more elaborate. We create a `nn.Module`, holding the `lora_embedding_A` Parameter dict instead of individual weights. This way, even if the parameter changes, we still get the proper weight for a given adapter.

I think it is cleaner than the first approach.

BenjaminBossan: > I changed SimpleNamespace, which is very basic and simple, with something a bit more elaborate. We create a `nn.Module`, holding the `lora_embedding_A` Parameter dict instead of individual weights.

That sounds better, but I think you didn't push that change yet.

Also, instead of monkey-patching `lora_module._embed` here, I think it's better to go into `peft.tuners.lora.Embedding._embed` and add branching code in there to check for `tp_plan` and call the corresponding helper functions there.

michaelbenayoun: I just pushed.
About your second suggestion, great idea. I just pushed it as well.

PR #3096, opened 2026-03-12
Save checkpoint with TP

michaelbenayoun:
Enable state dict gather before saving checkpoint when doing TP.

Should wait for:

- https://github.com/huggingface/transformers/pull/44640
- #3091 
- https://github.com/huggingface/transformers/pull/45155

cc @3outeille

### DIFF
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -315,7 +315,6 @@ class LoraModel(BaseTuner):
                         tp_module,
                         tp_plan,
                         tp_layer_name,
-                        tp_plan,
                         device_mesh,
                     )
                 else:  # embedding_rowwise


method_comparison/MetaMathQA/run.py
```diff
@@ -30,6 +30,7 @@
 from typing import Any, Literal, Optional
 
 import torch
+from data import get_train_valid_test_datasets, get_wiki_small
```
BenjaminBossan: Let's undo changes to unrelated files like here and in `utils.py`.

src/peft/tuners/lora/layer.py
```diff
@@ -43,6 +44,13 @@
 from .config import LoraConfig
 
 
+@dataclass
```
BenjaminBossan: How about we move this to `utils/integration.py` and give it a more generic name, as it's not strictly related to LoRA? Then we can also use it in `save_and_load.py` as the return type for `_get_tp_info`.

src/peft/utils/save_and_load.py
```diff
@@ -56,6 +56,17 @@ def get_embedding_layer_name(model, layer, is_embedding_in_target_modules):
...
 
 
+def _get_tp_info(model):
+    """Collect TP info from lora modules that have _tp_info set."""
+    tp_plan, device_mesh, tp_size = {}, None, None
+    for module in model.modules():
+        if hasattr(module, "_tp_info"):
+            tp_plan.update(module._tp_info.tp_plan)
+            device_mesh = module._tp_info.device_mesh
+            tp_size = module._tp_info.tp_size
```
BenjaminBossan, line 70: Can we `break` the loop here?

src/peft/utils/save_and_load.py
```diff
@@ -92,6 +103,25 @@ def get_peft_model_state_dict(
...
+
+        # If the keys in the state dict start with the prefix, we need to add the prefix to the keys in the tp plan as
+        # well.
+        prefix = (
+            "base_model."
+            if isinstance(model, PeftModel) and model.active_peft_config.is_prompt_learning
+            else "base_model.model."
+        )
+        keys_starting_with_prefix = all(k.startswith(prefix) for k in state_dict)
+        if keys_starting_with_prefix:
```
BenjaminBossan, line 128: Just wondering: Are there cases where only some keys start with the prefix? If yes, let's add a short comment to explain why the `tp_plan` only needs updating if all keys start with the prefix. If no, let's remove the check.

tests/test_gpu_examples.py
```diff
@@ -6167,6 +6209,94 @@ def _test_multiple_adapters(rank, world_size, port):
...
+    modules themselves (via _tp_info) rather than on the tuner.
+    """
+    torch.cuda.set_device(rank)
+    device = torch.device("cuda", rank)
+
+    model = AutoModelForCausalLM.from_pretrained(TINY_MODEL_ID, tp_plan=TP_PLAN)
+    model.to(device)
+
+    lora_config = LoraConfig(r=4, target_modules=TARGET_MODULES, init_lora_weights=True)
+    model = inject_adapter_in_model(lora_config, model)
```
BenjaminBossan: `model.load_adapter(lora_config)` should work equally and is closer to what a normal user would do. Same applies to the `inject_adapter_in_model` use below.

tests/test_gpu_examples.py
```diff
@@ -6167,6 +6209,94 @@ def _test_multiple_adapters(rank, world_size, port):
...
+def _test_inject_adapter_save(rank, world_size, port, tmp_dir_reference, tmp_dir_tp):
+    """
+    Test that get_peft_model_state_dict correctly gathers unsharded TP weights when using inject_adapter_in_model.
+
+    Flow:
+      1. Rank 0: create a plain (non-TP) model, inject adapter, save state dict as reference.
+      2. All ranks: load TP base model, inject adapter, save state dict via get_peft_model_state_dict.
+      3. Rank 0: compare re-saved state dict against the reference — keys and values must match.
+    """
+    from safetensors.torch import save_file
```
BenjaminBossan: Can be global import.

tests/test_gpu_examples.py
```diff
@@ -6185,5 +6315,16 @@ def test_from_checkpoint(self):
         with tempfile.TemporaryDirectory() as tmp_dir:
             self._spawn(_test_load_from_checkpoint, tmp_dir, port_offset=1)
 
+    def test_save_unsharded_weights(self):
+        with tempfile.TemporaryDirectory() as tmp_dir_reference, tempfile.TemporaryDirectory() as tmp_dir_tp:
```
BenjaminBossan: We can use the `tmp_path` pytest fixture here: Add `tmp_path` to the function signature and then do something like:

```python
def test_save_unsharded_weights(self, tmp_path):
    tmp_dir_reference = tmp_path / "reference"
    tmp_dir_tp = tmp_path / "tp"
```

Same for the other tests that use `tempfile.TemporaryDirectory()`.

method_comparison/MetaMathQA/run.py
michaelbenayoun: Done. It is the linter which does that.

src/peft/tuners/lora/layer.py
michaelbenayoun: Done!

src/peft/utils/save_and_load.py
michaelbenayoun: No we cannot. As you see we update the `tp_plan` everytime we have a tp info. `TpInfo` is local for each module, but to save we need to gather the tp plan for each sharded parameters, hence the iteration over every module.

src/peft/utils/save_and_load.py
michaelbenayoun: You're right; I dont have a case in mind where it would happen. Simplified it.

tests/test_gpu_examples.py
michaelbenayoun: I did not get this comment.

tests/test_gpu_examples.py
michaelbenayoun: Done

tests/test_gpu_examples.py
michaelbenayoun: Done

src/peft/utils/save_and_load.py
BenjaminBossan: I see. My concern was just to avoid the full loop if not necessary. Can we detect on the model/model config itself if there is any TP plan and if not, avoid the loop? Or do we really need to visit each module? And since `device_mesh` and `tp_size` are always overwritten by the last module: Can we assume that this is safe?

tests/test_gpu_examples.py
BenjaminBossan: What I mean is that instead of calling `inject_adapter_in_model(lora_config, model)`, we should call `model.load_adapter(lora_config)`, because this is closer to what a user would normally do.

src/peft/utils/save_and_load.py
michaelbenayoun: I changed things: I check if there is a TP plan in the base model, if there is then it's worth looping over, otherwise early exit.

It is safe that the device mesh and tp size are the same amongst all modules.

BenjaminBossan: Sounds good. LMK when the PR is ready for review.

BenjaminBossan:
@michaelbenayoun What's the state of the PR?

michaelbenayoun:
Still working on it but it's not as trivial as you would think to add support for `load_adapter`, hence the delay.

BenjaminBossan:
Got it, LMK if I can help.

tests/test_gpu_examples.py
michaelbenayoun: I have worked on it, after some time, I have done some changes here as well: https://github.com/huggingface/transformers/pull/45155

So this PR and the one in transformers are now connected.

michaelbenayoun:
It is all good on my end. One thing: adding support for `load_adapter` required some changes in Transformers as well, which are located here: https://github.com/huggingface/transformers/pull/45155.

BenjaminBossan:
Great, thanks for the progress. So if my understanding is correct, we should wait for the Transformers PR to land and be released first, otherwise this PR won't work correctly.

michaelbenayoun:
No, actually everything will work fine except the test that uses `PreTrainedModel.load_adapter`. 
The PR in Transformers is adding support for this path, while this PR is adding support for saving checkpoints.

BenjaminBossan (changes requested):
Thanks for updating the PR and also cleaning up the TP-related code. Generally, this LGTM.

I ran the tests locally with Transformers from main (I also hard-coded `is_transformers_ge_v5_6_0 = True` so that tests would not be skipped). There, I got an error:

>  TypeError: add_tensor_parallel_hooks_to_module() takes 5 positional arguments but 6 were given

This is because of the clean up in https://github.com/huggingface/transformers/pull/44768. So after removing the 2nd `tp_plan` argument in the `add_tensor_parallel_hooks_to_module` call, all tests passed. Could you please update the code?

Note: Failing CI is unrelated.

michaelbenayoun:
Yes I had actually created a branch for this specific change but will update it here and push.

BenjaminBossan (approved):
Thanks for pushing the fix, LGTM. Failing CI is unrelated.

PR #3097, opened 2026-03-12
Improve DeloRA: add config validation, dedicated tests, and fix typos

joshuaswanson:
DeloRA is missing some config validation that other tuners already have (exclude_modules list-to-set conversion, r validation, Literal type for bias). This PR fixes that, adds a dedicated test file (merge/unmerge, save/load, multiple adapters, etc.), and fills in the missing paper reference.

Also fixes some typos I noticed while going through the code.

### DIFF
--- a/src/peft/tuners/delora/model.py
+++ b/src/peft/tuners/delora/model.py
@@ -29,7 +29,8 @@ class DeloraModel(BaseTuner):
     """
     Creates DeLoRA model from a pretrained transformers model.
 
-    The method is described in detail in [DeLoRA: Decoupled Low-rank Adaptation](https://huggingface.co/papers/2503.18225).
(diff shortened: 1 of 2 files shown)


tests/test_delora.py
githubnemo, line 1: Let's remove these tests.

All of these tests are already covered by the tests defined in `testing_common.py` and implemented in `tests_{decoder,custom,encoder_decoder}_models.py`. I've compared the test coverages and there's no difference except for the newly introduced check for `r < 0` which is not worth the added tests.

joshuaswanson:
Removed the dedicated test file, thanks for the review.

githubnemo:
Please run `make style` to make the CI happy :)

joshuaswanson:
Ran the doc-builder style fix, hopefully that does it.

PR #3098, opened 2026-03-13
add glora

not-lain:
follow up on  #780 and #2568 
this pr adds GLoRA to the library
i also made a minor colab notebook to test this remotely in which you can find [here](https://colab.research.google.com/drive/15uyZDK7MILpYccmEPveE6YQlNg-yozM-?usp=sharing)


docs/source/package_reference/glora.md
```diff
@@ -0,0 +1,81 @@
+<!--Copyright 2025 The HuggingFace Team. All rights reserved.
```
BenjaminBossan:
```suggestion
<!--Copyright 2026 The HuggingFace Team. All rights reserved.
```

Let's update the ate everywhere.

src/peft/tuners/glora/config.py
```diff
@@ -0,0 +1,117 @@
...
+
+from dataclasses import dataclass, field
+from typing import Optional, Union
+
+from peft.config import PeftConfig
+from peft.utils.peft_types import PeftType
+
+
+@dataclass
+class GLoraConfig(PeftConfig):
```
BenjaminBossan: How about `GloraConfig`, `GloraLayer` etc. for easier typing and better consistency with the rest of PEFT?

src/peft/tuners/glora/config.py
```diff
@@ -0,0 +1,117 @@
...
+        config_C (`str`): Configuration for C matrix. Valid values: 'LoRA', 'vector', 'none'.
+        config_D_E (`str`): Configuration for D and E matrices. Valid values: 'constant', 'none', 'vector'.
+    """
+
+    _VALID_A_B_CONFIGS = {"LoRA", "vector", "constant", "none"}
+    _VALID_C_CONFIGS = {"LoRA", "vector", "none"}
+    _VALID_D_E_CONFIGS = {"constant", "none", "vector"}
+
+    r: int = field(
+        default=4, metadata={"help": "Default rank of the LoRA matrices if the config contains LoRA parametrization."}
```
BenjaminBossan: Why not 8 as for LoRA?

src/peft/tuners/glora/config.py
```diff
@@ -0,0 +1,117 @@
...
+    Args:
+        r (`int`): GLora attention dimension (rank of the LoRA matrices).
+        target_modules (`Optional[Union[List[str], str]]`): The names of the modules to apply GLora to.
+        config_A_B (`str`): Configuration for A and B matrices. Valid values: 'LoRA', 'vector', 'constant', 'none'.
+        config_C (`str`): Configuration for C matrix. Valid values: 'LoRA', 'vector', 'none'.
+        config_D_E (`str`): Configuration for D and E matrices. Valid values: 'constant', 'none', 'vector'.
+    """
+
+    _VALID_A_B_CONFIGS = {"LoRA", "vector", "constant", "none"}
+    _VALID_C_CONFIGS = {"LoRA", "vector", "none"}
```
BenjaminBossan: I think it is easier to have everything in lower case, i.e. "LoRA" -> "lora"

src/peft/tuners/glora/config.py
```diff
@@ -0,0 +1,117 @@
...
+        metadata={
+            "help": "List of module names or regex expression of the module names to replace with Lora."
+            "For example, ['q', 'v'] or '.*decoder.*(SelfAttention|EncDecAttention).*(q|v)$' "
+        },
+    )
+
+    config_A_B: str = field(
+        default="LoRA",
+        metadata={
+            "help": "Configuration for A and B matrices in GLora."
```
BenjaminBossan: All these explanations lack a bit more context about how they differ.

src/peft/tuners/glora/config.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+    target_modules: Optional[Union[list[str], str]] = field(
+        default=None,
+        metadata={
+            "help": "List of module names or regex expression of the module names to replace with Lora."
+            "For example, ['q', 'v'] or '.*decoder.*(SelfAttention|EncDecAttention).*(q|v)$' "
+        },
+    )
+
+    config_A_B: str = field(
```
BenjaminBossan: For all these configs, instead of using type `str`, let's use `Literal["lora", "vector", "constant", "none"]` etc. so that we get more precise type hints.

src/peft/tuners/glora/config.py
```diff
@@ -0,0 +1,117 @@
...
+    def __post_init__(self):
+        self.peft_type = PeftType.GLORA
+
+        # Validate and process each configuration
+        self.config_A_B = self._validate_and_process_config(self.config_A_B, self._VALID_A_B_CONFIGS, "config_A_B")
+
+        self.config_C = self._validate_and_process_config(self.config_C, self._VALID_C_CONFIGS, "config_C")
+
+        self.config_D_E = self._validate_and_process_config(
+            self.config_D_E, self._VALID_D_E_CONFIGS, "config_D_E", allow_lora=False
```
BenjaminBossan: Hmm, `self._VALID_D_E_CONFIGS` does not contain `"LoRA"`, so do we really need an extra argument `allow_lora`?

docs/source/package_reference/glora.md
```diff
@@ -0,0 +1,81 @@
...
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contains specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+-->
+
+# GLora
+
+Generalized Low-Rank Adaptation (**GLora**) is a highly flexible PEFT method that generalizes LoRA and related approaches. GLora allows you to decompose weight updates into multiple configurable low-rank, vector, or constant paths, providing a superset of LoRA's expressivity. Each path (A, B, C, D, E) can be independently configured, enabling a wide range of adaptation strategies.
```
BenjaminBossan: Let's add a reference to the paper: https://huggingface.co/papers/2306.07967

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,341 @@
...
+                X = Xd[0]
+            elif "none" in config:
+                X = torch.zeros(1, device=device, dtype=dtype)
+            else:
+                raise ValueError(f"Unknown config choice: {config} for non-decomposable path")
+        return X.to(device=device, dtype=dtype)
+
+
+# Refactored GLoraLinear for PEFT compatibility
+class GLoraLinear(GLoraLayer, nn.Linear):
```
BenjaminBossan: You probably based this PR on the very old original PR. Since then, we had several refactors in PEFT, some of which affect how we designed the adapter layers. Could you please check the latest implementation in `tuners/lora/layer` or PRs of other recent PEFT additions (e.g. #2851)? Most notably, we now pass the `base_layer` (i.e. the original layer) and wrap it inside the PEFT layer. To get the results of the base layer, we can then call `self.base_layer(x)` in the forward call (no need for `F.linear(x, self.weight)`).

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,292 @@
...
+        for n, p in model.named_parameters():
+            if "bias" in n:
+                p.requires_grad = True
+    elif bias == "glora_only":
+        for m in model.modules():
+            if isinstance(m, GLoraLinear) and hasattr(m, "bias") and m.bias is not None:
+                m.bias.requires_grad = True
+
+
+class GLoraModel(BaseTuner):
```
BenjaminBossan: Similar argument as for `GLoraLayer`: We have substantially refactored this part of PEFT. The good news is that it should greatly simplify the overall implementation: You only need to define `_create_and_replace` and `_create_new_module`, the remaining methods should all be fine when inherited from the parent class. Moreover, we need these class attributes: 

- `prefix: str = "glora_"`
- `tuner_layer_cls = GloraLayer`
- `target_module_mapping = TRANSFORMERS_MODELS_TO_GLORA_TARGET_MODULES_MAPPING`

tests/test_glora.py
BenjaminBossan, line 1: There is no need to have a separate test file for each PEFT method. Instead, you should add an entry to the test matrix for GLoRA, which is just a few lines of code and allows to run a wide battery of tests. As a start, please add GLoRA here:

https://github.com/huggingface/peft/blob/9f63491eb9cc68b4551334a26d2f6d5f58eb62e3/tests/test_custom_models.py#L77

Check how it's done for other methods and use the same approach. Before pushing your other changes, ensure that the tests pass locally (`pytest tests/test_custom_models.py -k glora -v`).

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,341 @@
...
+        self.r: dict[str, int] = {}
+        self.glora_Ad: nn.ParameterDict = nn.ParameterDict()
+        self.glora_Au: nn.ParameterDict = nn.ParameterDict()
+        self.glora_Bd: nn.ParameterDict = nn.ParameterDict()
+        self.glora_Bu: nn.ParameterDict = nn.ParameterDict()
+        self.glora_Cd: nn.ParameterDict = nn.ParameterDict()
+        self.glora_Cu: nn.ParameterDict = nn.ParameterDict()
+        self.glora_D: nn.ParameterDict = nn.ParameterDict()
+        self.glora_E: nn.ParameterDict = nn.ParameterDict()
+        self.eval_config: dict[str, dict[str, object]] = {}
```
BenjaminBossan: Why do we need this? IIUC, this is supposed to be an object that unifies the settings for each parameter, as we have distinct options like "vector" or "constant". I haven't checked all the details, but ideally we would just define during the layer initialization something like:

```python
if config_A_B == "lora":
    self.glora_Ad[adapter_name] = nn.Linear(...)
elif config_A_B == "vector":
    self.glora_Ad[adapter_name] = ...
```

The arguments `config_A_B` etc. should be passed to the `__init__` and `update_layer` methods, directly coming from the `GloraConfig`.

This change may require some custom `nn.Module`s but that would be fine. I would really like to avoid the whole `prepare_path` call during `forward` and frontload the whole resolution to the initialization. This make the `forward`/`merge`/`unmerge` call simpler to understand and should also be slightly more performant.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,341 @@
...
+        nn.init.kaiming_uniform_(self.glora_Cu[adapter_name], a=math.sqrt(5))
+
+    def make_param(self, shape, config=None):
+        if config is not None and "LoRA" in config:
+            out_feature = shape[0]
+            in_feature = shape[1]
+            try:
+                rank = int(config.split("_")[1])
+            except Exception:
+                rank = 4
```
BenjaminBossan: Why pass `make_config(..., config=f"LoRA_{r}")` only to immediately parse the `r` into an int again? Why not pass `r` as an int directly? Honestly, I think we can completely remove `make_param` and just directly initialize the parameters.

BenjaminBossan (changes requested):
Thanks for reviving the GLoRA implementation to PEFT. There is still a lot missing (docs, examples, more tests) but let's focus on the integration for now and work on the rest later.

Unfortunately, this PR seems to be based on the state that PEFT was in when the PR was first suggested. We made several refactors since then, which require different patterns to implement. The good news is that the final result should be much simpler with less code required. I marked the corresponding parts, please check. Maybe this is even something that a coding agent can do if asked to update the implementation and pointed to the most recent PEFT code for reference.

BenjaminBossan:
Note that due to another PEFT method being merged, there is now a merge conflict, but it should be straightforward to resolve. Once you're finished, please ping me for another review.

not-lain:
@BenjaminBossan 
made the necessary changes mentioned above and made a couple extra changes to match that of `test_custom_models`.
would appreciate it if you can take a look at this pr.

docs/source/package_reference/glora.md
```diff
@@ -0,0 +1,101 @@
...
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contains specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+-->
+
+# Glora
+
+Generalized Low-Rank Adaptation (**Glora**) is a PEFT method that generalizes LoRA and related approaches. Glora decomposes updates into configurable paths (A, B, C, D, E), where each path can use low-rank, vector, constant, or disabled parameterization depending on the path.
```
BenjaminBossan, line 18: Note that my comment about using `Glora` instead of `GLora` only applied to code, in the docs, feel free to use the spelling of the paper, i.e. `GLoRA`.

src/peft/tuners/glora/config.py
```diff
@@ -0,0 +1,166 @@
...
+        config_A_B (`str`): Parameterization for the A and B matrices (weight multiplicative and additive
+            corrections). Valid values: `lora`, `vector`, `constant`, `none`.
+        config_C (`str`): Parameterization for the C matrix (weight-to-bias coupling: b += W0 @ C).
+            Valid values: `lora`, `vector`, `none`.
+        config_D_E (`str`): Parameterization for the D and E scalars (bias multiplicative and additive
+            corrections). Does not support `lora` since D and E are bias-sized vectors, not matrices.
+            Valid values: `vector`, `constant`, `none`.
+    """
+
+    _VALID_A_B_CONFIGS: ClassVar[set[str]] = {"lora", "vector", "constant", "none"}
```
BenjaminBossan, line 53: Why is `ClassVar` required here?

src/peft/tuners/glora/config.py
```diff
@@ -0,0 +1,166 @@
...
+                "'vector': per-output-channel value (out params). "
+                "'constant': single scalar shared across all bias elements (1 param). "
+                "'none': disabled (0 params). "
+                "Does not support 'lora' since D and E are bias-length vectors, not matrices. "
+                f"Valid values: {', '.join(sorted(_VALID_D_E_CONFIGS))}."
+            )
+        },
+    )
+
+    def _validate_and_process_config(self, config_value: str, valid_configs: set, config_name: str) -> str:
```
BenjaminBossan, line 123:
```suggestion
    def _validate_and_process_config(self, config_value: str, valid_configs: set[str], config_name: str) -> str:
```

src/peft/tuners/glora/config.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __post_init__(self):
+        super().__post_init__()
+        self.peft_type = PeftType.GLORA
+        self.target_modules = (
+            set(self.target_modules) if isinstance(self.target_modules, list) else self.target_modules
+        )
+
+        # Validate and process each configuration
+        self.config_A_B = self._validate_and_process_config(self.config_A_B, self._VALID_A_B_CONFIGS, "config_A_B")  # type: ignore[assignment]
```
BenjaminBossan, line 160: I'm not quite sure why we need to re-assign here. Isn't it enough to check if the option is valid and raise if it's not?

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,342 @@
...
+    )
+    other_param_names = ("r", "eval_config")
+
+    def __init__(self, base_layer: nn.Module, **kwargs) -> None:
+        self.base_layer = base_layer
+        self.in_features: int
+        self.out_features: int
+        base = self.get_base_layer()
+        # Use exact type check: bitsandbytes.Linear4bit subclasses nn.Linear but is not compatible with GLORA math.
+        if type(base) is not nn.Linear:
```
BenjaminBossan, line 101: Let's remove this. The check for the correct type should happen in `_create_new_module`.

src/peft/tuners/glora/layer.py
BenjaminBossan: This comment is still relevant.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,342 @@
...
+        self.glora_C: nn.ModuleDict = nn.ModuleDict()
+        self.glora_D: nn.ModuleDict = nn.ModuleDict()
+        self.glora_E: nn.ModuleDict = nn.ModuleDict()
+        self.eval_config: dict[str, dict[str, object]] = {}
+        self.merged_adapters: list[str] = []
+        self._disable_adapters: bool = False
+        self._active_adapter: str | list[str] = []
+        self.kwargs: dict[str, Any] = dict(kwargs)
+
+    def add_adapter(self, adapter_name: str, r: int, config_A_B: str, config_C: str, config_D_E: str) -> None:
```
BenjaminBossan, line 125: Delete this whole method. Instead implement `update_layer`, which takes the job of initializing/updating the GLoRA parameters. Check how it's done in other PEFT methods. For this, please check the latest main branch of PEFT, not the last release, because we recently refactored `update_layer`, see #3047.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,342 @@
...
+        self.r[adapter_name] = r
+
+        # Initialize A and B
+        cfg_ab = config_A_B.lower()
+        if "lora" in cfg_ab:
```
BenjaminBossan, line 130: Let's not do `in` checks but exact checks. Moreover, there should be no need to cast to lowercase. In the config's `__post_init__`, the validation checks should be enough to ensure that the values match exactly.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,342 @@
...
+from typing import Any, Optional
+
+import torch
+import torch.nn.functional as F
+from torch import nn
+
+from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
+
+
+class GloraPath(nn.Module):
```
BenjaminBossan, line 25: I like the idea of having different submodules, instead of a complicated `if ... else` in `forward`.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,342 @@
...
+            if bias is not None:
+                base_layer.bias.data = (b_merged - E - torch.matmul(w0, C).squeeze(-1)) / (1.0 + D)
+            self.merged_adapters.remove(adapter_name)
+
+    def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
+        adapter_names = kwargs.pop("adapter_names", None)
+        if self.disable_adapters:
+            if self.merged:
+                self.unmerge()
+            return self.base_layer(x, *args, **kwargs)
```
BenjaminBossan, line 270: Principally, doing early returns like this is totally valid. However, none of the other PEFT methods do this. Please just assign the `result` here and below, then return it once at the end of the method.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,342 @@
...
+            b_merged = bias.data.clone() if bias is not None else None
+            w0 = (w_merged - B) / (1.0 + A)
+            base_layer = self.get_base_layer()
+            base_layer.weight.data = w0
+            if bias is not None:
+                base_layer.bias.data = (b_merged - E - torch.matmul(w0, C).squeeze(-1)) / (1.0 + D)
+            self.merged_adapters.remove(adapter_name)
+
+    def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
+        adapter_names = kwargs.pop("adapter_names", None)
```
BenjaminBossan, line 266: Only keep this line if you really want to support mixed batches (see other comment).

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,342 @@
...
+        if self.disable_adapters:
+            if self.merged:
+                self.unmerge()
+            return self.base_layer(x, *args, **kwargs)
+        if not self.active_adapters:
+            return self.base_layer(x, *args, **kwargs)
+        if adapter_names is not None:
```
BenjaminBossan, line 273: So you're trying to implement mixed batches here (i.e. a single batch containing different GLoRA adapters). Generally, this is not required, most PEFT methods don't support that feature. If you do want to support it, please copy the pattern used in LoRA instead: implement a separate `_mixed_batch_forward` method and call it like so:

```python
        elif adapter_names is not None:
            result = self._mixed_batch_forward(x, *args, adapter_names=adapter_names, **kwargs)
```

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,342 @@
...
+            w_merged = weight.data.clone()
+            b_merged = bias.data.clone() if bias is not None else None
+            w0 = (w_merged - B) / (1.0 + A)
+            base_layer = self.get_base_layer()
+            base_layer.weight.data = w0
+            if bias is not None:
+                base_layer.bias.data = (b_merged - E - torch.matmul(w0, C).squeeze(-1)) / (1.0 + D)
+            self.merged_adapters.remove(adapter_name)
+
+    def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
```
BenjaminBossan, line 265: Move this whole method to `GloraLinear` instead of calling `GloraLayer.forward(self, x, *args, **kwargs)` there.

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,146 @@
+from dataclasses import asdict
```
BenjaminBossan, line 1: Don't forget the copyright notice.

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def __init__(
+        self,
+        model: nn.Module,
+        config: GloraConfig,
+        adapter_name: str = "default",
+        low_cpu_mem_usage: bool = False,
+        state_dict: Optional[dict[str, Any]] = None,
+    ):
+        self.adapters_config_history: dict[str, Any] = {}
```
BenjaminBossan, line 33: I don't think we need `adapters_config_history`, so please remove it and hence the whole `__init__` method.

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,146 @@
...
+        target_name: str,
+        parent: nn.Module,
+        current_key: str,
+        parameter_name: Optional[str] = None,
+    ) -> None:
+        if current_key is None:
+            raise ValueError("Current Key shouldn't be `None`")
+
+        if isinstance(target, GloraLinear):
+            target.add_adapter(
```
BenjaminBossan, line 50: Please check how `_create_and_replace` is implemented elsewhere. `add_adapter` is never called, instead, call `target.update_layer` here.

Also note that, as mentioned earlier, we did a small refactor here, so the `peft_config` should be passed to `update_layer` instead of explicitly unpacking all the arguments.

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @staticmethod
+    def _create_new_module(
+        peft_config: GloraConfig,
+        adapter_name: str,
+        target: nn.Module,
+        current_key: str | None = None,
+        **optional_kwargs: Any,
+    ) -> GloraLinear:
+        del optional_kwargs  # unused; accepted for signature compatibility with the tuner injection path
```
BenjaminBossan, line 73: Remove

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,146 @@
...
+        if isinstance(target, BaseTunerLayer):
+            target_base_layer = target.get_base_layer()
+        else:
+            target_base_layer = target
```
BenjaminBossan, line 77: Remove

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,146 @@
...
+        if type(target_base_layer) is not nn.Linear:
+            raise ValueError(
+                f"Target module {target} is not a plain torch.nn.Linear (after unwrapping); GLORA does not support this layer type."
```
BenjaminBossan, line 80: I'm wondering why we need the strict check here. What errors did you run across without it?

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,146 @@
...
+            )
+
+        kwargs_glora = {
+            "config_A_B": peft_config.config_A_B,
+            "config_C": peft_config.config_C,
+            "config_D_E": peft_config.config_D_E,
+        }
+        new_module = GloraLinear(target, **kwargs_glora)
+        new_module.add_adapter(
```
BenjaminBossan, line 89: Again, don't call `add_adapter` here, `new_module = GloraLinear(target, **kwargs_glora)` should be enough, as `GloraLinear` should call `self.update_layer`.

Also note that, as mentioned earlier, we did a small refactor here, so the `peft_config` should be passed to `GloraLinear`'s `__init__`.

src/peft/tuners/glora/model.py
```diff
@@ -0,0 +1,146 @@
...
+            adapter_name,
+            peft_config.r,
+            peft_config.config_A_B,
+            peft_config.config_C,
+            peft_config.config_D_E,
+        )
+        return new_module
+
+    def __getattr__(self, name: str):
```
BenjaminBossan, line 98: All the code starting from here should not be necessary and can be deleted.

tests/test_custom_models.py
```diff
@@ -1375,6 +1411,22 @@ def _skip_tests_with_multiple_adapters_with_target_parameters(config_cls, config
         pytest.skip("LoRA with multiple adapters with target_parameters is not supported")
 
 
+def _glora_init_down_factors_for_multiadapter_testing(peft_model, adapter_names=None):
```
BenjaminBossan, line 1414: I think it would be better to add an `init_weights` arguments to `GloraConfig`. If it's set to `True` (default), GLoRA should be initialized as a no-op, which is a general PEFT convention. If it's set to `False`, it shouldn't be a no-op, regardless of which GLoRA-option is configured (I guess except for `"none"`).

tests/test_custom_models.py
```diff
@@ -3911,6 +3970,16 @@ def test_multiple_active_adapters_merge_and_unmerge(
         _skip_if_merging_not_supported(test_name, config_cls, config_kwargs_1)
         _skip_tests_with_multiple_adapters_with_target_parameters(config_cls, config_kwargs_2)
 
+        config_1 = config_cls(**config_kwargs_1)
```
BenjaminBossan, line 3973: This shouldn't be necessary. Why do the results differ?

tests/testing_common.py
```diff
@@ -766,7 +767,7 @@ def _test_safe_merge(self, model_id, config_cls, config_kwargs):
 
             atol, rtol = 1e-6, 1e-6  # default
             # Initializing with LN tuning cannot be configured to change the outputs (unlike init_lora_weights=False)
-            if not issubclass(config_cls, LNTuningConfig):
+            if not issubclass(config_cls, (LNTuningConfig, GloraConfig)):
```
BenjaminBossan, line 770: As discussed earlier, let's add an init option for GLoRA for no-op and one for not-no-op.

BenjaminBossan (changes requested):
Thanks for the latest updates. I found that there is still some work left to make GLoRA consistent with other PEFT methods. Please check my comments and don't hesitate to ask if something is unclear.

Also note that in the meantime, we have updated the contribution guidelines for adding new PEFT methods: https://github.com/huggingface/peft/blob/main/docs/source/developer_guides/contributing.md#add-a-new-peft-fine-tuning-method. You already tick many boxes, but please refer to this when it comes to what's still missing.

docs/source/package_reference/glora.md
not-lain: fixed in https://github.com/not-lain/peft/blob/0e9b580d9a408caccd07831fa7dd3d8d14f6c998/docs/source/package_reference/glora.md?plain=1#L18

src/peft/tuners/glora/config.py
not-lain: it was a ty linting issue, looking back i should have used `field` from dataclass, but i went with the AI suggestion at the time, i have switched out from this rn so i will be resolving this

image

src/peft/tuners/glora/config.py
not-lain: makes sense, did apply this in 0e9b580d9a408caccd07831fa7dd3d8d14f6c998 
so i'm resolving this : )

tests/test_custom_models.py
not-lain: The issue is that GLoRA's merge is multiplicative, not purely additive, during multi-active forward, each adapter computes its delta against the original frozen `W0` thus the formula : 
```
 result = W0@x + (W0*A1 + B1)@x + (W0*A2 + B2)@x
```
But sequential merging mutates `W0` in place between adapters:
* after adapter1: `W0' = W0 + W0*A1 + B1`
* after adapter2: `W0'' = W0' + W0'*A2 + B2`

The second merge uses the already-modified `W0'`, meaning `W0'*A2` expands to `(W0 + W0*A1 + B1)*A2` instead of the original `W0*A2`


LoRA doesn't have this issue because its merge is purely additive (`W0 + delta`), so i chose to skip the test for GLoRA for now.

tests/testing_common.py
not-lain: Done, did include `init_weights` parameter in the `GloraConfig` which handles this

src/peft/tuners/glora/config.py
not-lain: resolving this since this entire method has been removed

method_comparison/text_generation_benchmark/utils.py
```diff
@@ -71,8 +71,12 @@ def __post_init__(self):
             "duration": 0.0,
             "status": self.status.value,
             "hardware": {
-                "num_accelerators": torch_accelerator_module.device_count() if torch_accelerator_module.is_available() else 0,
-                "accelerator_type": torch_accelerator_module.get_device_name(0) if torch_accelerator_module.is_available() else "N/A",
+                "num_accelerators": torch_accelerator_module.device_count()
```
BenjaminBossan: Seems like you accidentally formatted this file. Please undo.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,322 @@
...
+
+from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
+
+
+class GloraPath(nn.Module):
+    def forward(self):
+        raise NotImplementedError
+
+
+class GloraLoRAPath(GloraPath):
```
BenjaminBossan: Let's go with `GloraLoraPath`

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,322 @@
...
+import torch
+import torch.nn.functional as F
+from torch import nn
+
+from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
+
+
+class GloraPath(nn.Module):
+    def forward(self):
+        raise NotImplementedError
```
BenjaminBossan, line 31: Let's also add a `def reset_parameters` that raises `NotImplementedError`, since all subclasses need that too.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,322 @@
...
+        self.glora_C: nn.ModuleDict = nn.ModuleDict()
+        self.glora_D: nn.ModuleDict = nn.ModuleDict()
+        self.glora_E: nn.ModuleDict = nn.ModuleDict()
+        self.merged_adapters: list[str] = []
+        self._disable_adapters: bool = False
+        self._active_adapter: str | list[str] = []
+        self.kwargs: dict[str, Any] = dict(kwargs)
+
+    def update_layer(self, adapter_name: str, r: int, config, **kwargs) -> None:
+        from .config import GloraConfig
```
BenjaminBossan: Can be global import, right?

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,322 @@
...
+
+        assert isinstance(config, GloraConfig)
```
BenjaminBossan: Remove the assert. You can add `GloraConfig` to the type annotation to satisfy the type checker.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,322 @@
...
+            bias = self.bias
+            device, dtype = weight.device, weight.dtype
+
+            A = self.glora_A[adapter_name]().to(device=device, dtype=dtype)
+            B = self.glora_B[adapter_name]().to(device=device, dtype=dtype)
+            C = self.glora_C[adapter_name]().to(device=device, dtype=dtype)
+            D = self.glora_D[adapter_name]().to(device=device, dtype=dtype)
+            E = self.glora_E[adapter_name]().to(device=device, dtype=dtype)
+
+            w0 = weight.data.clone()
```
BenjaminBossan: Let's avoid the `clone` here and only call it in the `safe_merge` block. In the unsafe block, let's rely on in-place mutation.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,322 @@
...
+            bias = self.bias
+            device, dtype = weight.device, weight.dtype
+
+            A = self.glora_A[adapter_name]().to(device=device, dtype=dtype)
+            B = self.glora_B[adapter_name]().to(device=device, dtype=dtype)
+            C = self.glora_C[adapter_name]().to(device=device, dtype=dtype)
+            D = self.glora_D[adapter_name]().to(device=device, dtype=dtype)
+            E = self.glora_E[adapter_name]().to(device=device, dtype=dtype)
+
+            w_merged = weight.data.clone()
```
BenjaminBossan: This can be done in-place to avoid cloning, right?

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,322 @@
...
+class GloraLinear(nn.Module, GloraLayer):
+    """GLoRA adapter wrapping a dense [`~torch.nn.Linear`] `base_layer`."""
+
+    def __init__(self, base_layer: nn.Module, adapter_name: str, config, **kwargs) -> None:
+        super().__init__()
+        GloraLayer.__init__(self, base_layer, **kwargs)
+        self._active_adapter = adapter_name
+        self.update_layer(adapter_name, config.r, config=config)
+
+    def _mixed_batch_forward(
```
BenjaminBossan, line 312: If you want GLoRA to support mixed adapter batches, you have to also implement  `_enable_peft_forward_hook` on `GloraModel`, see e.g. [here](https://github.com/huggingface/peft/blob/4f7ddfabbb0d03c6071e7ba922335bde26da4cf7/src/peft/tuners/road/model.py#L118). Also, please ensure that they're also being tested. Please check `TestMixedAdapterBatches` in `test_custom_models.py`. It's also fine not to support it, in which case you can just remove `_mixed_batch_forward` and the `adapter_names = kwargs.pop("adapter_names", None)` part in `forward`.

src/peft/tuners/glora/layer.py
```diff
@@ -0,0 +1,322 @@
...
+                C = self.glora_C[active_adapter]().to(device=device, dtype=dtype)
+                D = self.glora_D[active_adapter]().to(device=device, dtype=dtype)
+                E = self.glora_E[active_adapter]().to(device=device, dtype=dtype)
+                weight_eff = self.weight + self.weight * A + B
+                if self.bias is not None:
+                    bias_eff = self.bias + self.bias * D + E + torch.matmul(self.weight, C).squeeze(-1)
+                    adapted = F.linear(x, weight_eff, bias=bias_eff)
+                else:
+                    adapted = F.linear(x, weight_eff)
+                result = result + (adapted - base_result)
```
BenjaminBossan: Let's say we have a lot of active adapters, like 3+. It would be more efficient not to subtract `base_result` inside the loop. Instead, it would be better to count the number of active adapters and then, outside the loop, subtract `num_active_adapters * base_result`. It's a small optimization and probably not very relevant, so I'm fine keeping it as is.

tests/test_custom_models.py
BenjaminBossan: I think this function should be unnecessary when initializing GLoRA with `init_weights=False`, no?

tests/test_custom_models.py
BenjaminBossan: So you're saying that if I have a GLoRA model with two active adapters, it behaves differently when merged vs unmerged? I think this should be treated like a bug. IIUC, it should be possible to implement the merge so that it gives the same result by not merging sequentially but combining the adapters before merging.

That said, if a user wants to merge adapter 1 first and then adapter 2 in a separate call, that would indeed require extra logic, as the results of naively merging the 2nd adapter would not be correct. I see two solutions here:

1. First unmerge adapter 1, then merge 1 and 2 together.
2. When merging, check if there is already a merged adapter and raise an error, telling users they need to merge all adapters in one step.

WDYT?

BenjaminBossan (changes requested):
Thanks for the updates. I still found a couple of issues, but the integration is close to being ready. Please take a look.

src/peft/tuners/glora/layer.py
not-lain: it does sound better that way, i did fix this in 26174e68823af481e5de12d23962fdcfa2e586aa
resolving this

src/peft/tuners/glora/layer.py
not-lain: did implement this in 66c57faeb820ab0ed4b9c67debd2bed7467c9dd4 
IMO this would look better if we can turn it into a mixin, maybe even open an issue and try to integrate mixed adapter batching into the rest of the peft methods. But i digress, let me know what you feel about this and feel free to resolve this thread since this is already done

BenjaminBossan: I agree that would be a nicer solution, though I think not quite trivial to implement. But that's for a separate PR. I'll resolve the discussion with the next review.

src/peft/tuners/glora/layer.py
not-lain: nice catch, it is a really nice optimization, I chose to include it later in the commits, resolving as done

tests/test_custom_models.py
not-lain: thanks for letting me know, removed

src/peft/tuners/glora/model.py
not-lain: could you elaborate more on this point, I see that the same snippet is used in other files as well, example: 
https://github.com/huggingface/peft/blob/236f67335753eff203b8fd0fa388bf124eadac7e/src/peft/tuners/hra/model.py#L106-L110

tests/test_custom_models.py
not-lain: chose to unmerge adapters before merging, didn't want to introduce an edge case for glora
you can review the changes related to this change in https://github.com/huggingface/peft/pull/3098/commits/43ceaee2d003725b49f073223578ff7cda141ac6

BenjaminBossan:
@not-lain Your latest changes look good. I wanted to run the existing tests to see if there are any issues but the linter is complaining. Could you please run `make style`?

If the tests pass, the next step would be to complete the "Full PR" according to the [contribution guideline](https://github.com/huggingface/peft/blob/main/docs/source/developer_guides/contributing.md#add-a-new-peft-fine-tuning-method).

src/peft/tuners/glora/model.py
BenjaminBossan: Not sure what my comment refers to, feel free to ignore.

not-lain:
thanks a lot for the follow up, tried addressing the rest of the points from the guideline, could you try running the CI again and let me know if you have any feedback

docs/source/package_reference/glora.md
```diff
@@ -0,0 +1,100 @@
...
+# Switch adapters, merge, etc.
+model.set_adapter("default")
+model.merge_and_unload()
+```
+
+## Notes
+- GLoRA is a superset of LoRA: setting all paths to "lora" recovers standard LoRA.
+- You can use different path types for A/B/C/D/E to experiment with new adaptation strategies.
+- GLoRA supports all standard PEFT adapter management features (add, delete, switch, merge, etc).
+- For unsupported module types, set `target_modules` to linear projections only.
```
BenjaminBossan: Could you please elaborate on this comment?

examples/glora_finetuning/glora_finetuning.py
```diff
@@ -0,0 +1,225 @@
...
+            {example["instruction"]}
+            ### Response:
+            {example["output"]}"""
+
+
+if __name__ == "__main__":
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--base_model", type=str, default="path/to/model")
```
BenjaminBossan: How about removing the invalid default value?

method_comparison/text_generation_benchmark/utils.py
```diff
@@ -108,7 +108,9 @@ def __post_init__(self):
                 "version": platform.version(),
                 "machine": platform.machine(),
                 "processor": platform.processor(),
-                "accelerator": torch_accelerator_module.get_device_name(0) if torch_accelerator_module.is_available() else "N/A",
+                "accelerator": torch_accelerator_module.get_device_name(0)
```
BenjaminBossan: Please undo unrelated changes in this file.

examples/glora_finetuning/glora_finetuning.py
BenjaminBossan, line 1: I checked the example and it runs.

method_comparison/MetaMathQA/experiments/glora/llama-3.2-3B-default/adapter_config.json
BenjaminBossan, line 1: Ran the experiment locally. It got 42% test accuracy, which is not top scoring, but expected given the relatively low rank. We can leave it as is and check for better settings in the future.

docs/source/package_reference/glora.md
```diff
@@ -0,0 +1,100 @@
...
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contains specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+-->
+
+# GLoRA
+
+Generalized Low-Rank Adaptation ([GLoRA](https://huggingface.co/papers/2306.07967)) is a PEFT method that generalizes LoRA and related approaches. GLoRA decomposes updates into configurable paths (A, B, C, D, E), where each path can use low-rank, vector, constant, or disabled parameterization depending on the path.
```
BenjaminBossan, line 18: It would be great if you could quickly explain what the four options ("lora", "vector", "constant", "none") mean. Let's also mention that they trade off number of params vs expressiveness.

method_comparison/MetaMathQA/experiments/glora/llama-3.2-3B-vector-paths/adapter_config.json
BenjaminBossan, line 1: I got 36% test accuracy with this one. Given the very small number of trainable parameters, I think it's a nice result.

BenjaminBossan (changes requested):
Thank you for adding the docs, examples, and other little changes. Some tests are failing, but that's just because the error message doesn't quite match, this should be easy to fix.

To complete the PR, could you please also add the tests to the other test files? `test_config.py`, `test_encoder_decoder_models.py`, `test_feature_extraction_models.py`, `test_seq_classifier.py`.

Also, we recently added a new [benchmark for image generation](https://github.com/huggingface/peft/tree/main/method_comparison/image-gen) similar to the MetaMath benchmark. It would be a nice addition to add an experiment for that as well.

docs/source/package_reference/glora.md
not-lain:
```suggestion
Generalized Low-Rank Adaptation ([GLoRA](https://huggingface.co/papers/2306.07967)) is a PEFT method that generalizes LoRA and related approaches. GLoRA decomposes updates into configurable paths (A, B, C, D, E), where each path can use low-rank, vector, constant, or disabled parameterization depending on the path.

Each path supports one of four parameterization modes. They trade off **parameter count** against **expressiveness** (how rich the update can be):

- `"lora"`: Low-rank decomposition (like standard LoRA). Uses `r * (out + in)` parameters and can express rank-`r` corrections. Most expressive, most parameters.
- `"vector"`: A single vector (e.g. shape `(out, 1)`), broadcast across the matrix. Uses `O(out)` parameters; only per-channel scaling or shifts.
- `"constant"`: A single scalar shared across all elements. Uses 1 parameter; least expressive among the trainable options.
- `"none"`: Zeros with no trainable parameters; disables that path entirely.

Not every path accepts every mode (for example, `config_D_E` does not support `"lora"`). Choosing `"lora"` on more paths increases capacity and trainable parameters; `"vector"`, `"constant"`, or `"none"` reduce both.
```

not-lain: did commit the comments above in 82a0985ba2bc3fc60c536552d90ef2bb9d3f02fd should resolve any ambiguities in what each of these params mean, 
(kept it unresolved for better reach feel free to resolve this if you think it's done)

docs/source/package_reference/glora.md
not-lain: i meant glora only works on nn.linear layers only, i do agree with you the comment is a little confusing, i'll try removing it entirely
```suggestion
```

not-lain:
Thanks for the detailed review @BenjaminBossan 
did follow up on your comments, feel free to take a look at the pr when you can

BenjaminBossan (approved):
Thanks for the latest updates. I did one last review pass and didn't find anything that was missing. So the PR is ready to go.

I also ran GLoRA on the image gen benchmark, it got 0.678 similarity and 0.270 drift with 11 GB peak memory. That's decent but not topping the charts. Final results will follow after the PR is merged.

Before merging, just one thing: Could you provide a short, tweet-sized blurb for GLoRA? It should describe: 1. What is GLoRA, 2. when it should be used, 3. when it's not a good choice. Thanks.

not-lain:
Hi @BenjaminBossan thanks a lot for the follow up, here are some comments from my side 
> I also ran GLoRA on the image gen benchmark

you should in theory get better results if you set most config_A_B, config_C to `lora` and config_D_E to `vector` that's how you get more trainable params, but for the sake of generalizing i wanted to choose random params (check doc below for more details) 
https://github.com/huggingface/peft/blob/d1ac2783858eadf38cf348921f9bc06c4911a60c/docs/source/package_reference/glora.md#L22-L27


---

> Could you provide a short, tweet-sized blurb for GLoRA?

GLoRA (Generalized LoRA) is a flexible PEFT method that extends LoRA with configurable weight, activation, and bias adaptation, delivering richer fine-tuning with no extra inference cost. Use it when you need per-layer flexibility or stronger adaptation than vanilla LoRA. Skip it for non-Linear layers (e.g. Conv/Embedding) or when standard LoRA is already sufficient and simplicity matters.

BenjaminBossan:
> you should in theory get better results if you set most config_A_B, config_C to `lora` and config_D_E to `vector` that's how you get more trainable params, but for the sake of generalizing i wanted to choose random params (check doc below for more details)

Makes sense.

Thanks for providing the blurb. Merging now.

PR #3101, opened 2026-03-16
FIX Broken tests with torchao >= 0.15

BenjaminBossan:
Torchao made some API changes, which have to be reflected in the tests. Moreover, for this to pass, we also need transformers to make the corresponding adjustments:

https://github.com/huggingface/transformers/pull/44604

While working on this, I migrated the tests from unittest to pytest style.

### DIFF
--- a/src/peft/import_utils.py
+++ b/src/peft/import_utils.py
@@ -117,7 +117,7 @@ def is_torchao_available():
     if importlib.util.find_spec("torchao") is None:
         return False
 
-    TORCHAO_MINIMUM_VERSION = packaging.version.parse("0.4.0")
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
@SunMarc @githubnemo I tried this branch with the new Transformers v5.4.0 release and found that it requires torchao v0.16 to work (loading the model with 0.15 caused an error for me). Furthermore, the check for torchao int4 weights didn't work anymore, so I updated that as well. Can you please review the last commit? One of you should be enough.

Parts of the failing CI are caused by a Transformers issue (https://github.com/huggingface/transformers/issues/45042) that should be addressed in the next patch release, and the rest is addressed by #3109.

PR #3103, opened 2026-03-16
CHORE Zizmor: branch protection for GH workflows

BenjaminBossan:
Zizmor currently gives the following type of warning:

```
  --> .github/workflows/build_docker_images.yml:33:25
   |
19 |   latest-cpu:
   |   ---------- this job
...
33 |           username: ${{ secrets.DOCKERHUB_USERNAME }}
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^ secret is accessed outside of a dedicated environment
```

https://github.com/huggingface/peft/actions/runs/23046711202/job/66937507862

This is not really an issue, since these workflows are only triggered on main, thus attackers cannot exfiltrate secrets this way. Moreover, maintainers vet each PR before allowing CI to run. Finally, GH doesn't expose secrets to forks by default. There are thus multiple layers of protection, which make these warnings false positives.

However, for forward security, I still added a GH Environment with branch protection (only 'main') and wait timer where appropriate.

https://github.com/huggingface/peft/settings/environments

This should add yet another protection layer in case the others are inadvertently removed in the future.

With these changes, I had to update the Zizmor config, as the lines to ignore moved around. There, I also saw that there were some lines in `build_docker_images.yml` that no longer exists, as we have removed the BNB CI, so these lines were removed too.

### DIFF
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -10,6 +10,8 @@ rules:
       # There is no obvious benefit so we leave it as it is.
       - build_docker_images.yml:39:9
       - build_docker_images.yml:74:9
+  secrets-outside-env:
+    ignore:
       # Zizmor warns about HF_TOKEN being used there but the token is not exposed to external users:
       # > With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.
       # https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow


.github/workflows/build_docker_images.yml
```diff
@@ -18,6 +18,8 @@ env:
 jobs:
   latest-cpu:
     name: "Latest Peft CPU [dev]"
+    # GH Environment for extra protection: https://github.com/huggingface/peft/settings/environments
+    environment: branch-protection-main
```
githubnemo, line 22: Does this mean that when testing a change for, say, building & pushing the container the build will now fail at the push stage since the secret is not available? Maybe we should document that this will limit the token availability to main so that we don't trip over this in the future.

BenjaminBossan: When testing a change to the docker build, we only run https://github.com/huggingface/peft/blob/main/.github/workflows/test-docker-build.yml, i.e. only building it but not pushing it, so that should be fine. AFAICT, we don't have testing where we also push the docker image.

PR #3104, opened 2026-03-16
FIX CI Remove invalid arg in nightly GPU test call

BenjaminBossan:
We are using pytest-reportlog to write pytest logs to a file to be later used for reporting on Slack (`log_reports.py`). This option only works for pytest invocations, not normal Python scripts, thus it is removed there.

For other pytest calls with `--report-log`, I made a change to have a distinct log file for each.


BenjaminBossan:
Note: Failing tests are unrelated.

PR #3105, opened 2026-03-16
FIX GraLoRA: Use its own target module mapping

BenjaminBossan:
GraLoRA was using the target module mapping from LoRA, but we prefer each PEFT method to have its own mapping, even it's just a copy of LoRA.


BenjaminBossan:
FYI @yeonjoon-jung01. I missed this during review.

PR #3108, opened 2026-03-17
CI Move slow EVA tests to nightly GPU CI

BenjaminBossan:
The EVA tests are quite slow on our CPU CI. Just counting those in the top 10 slowest tests, EVA tests make up 2:30 min, a good 10% of the overall runtime. Therefore, tests tests are now moved to the nightly CI.

There were already GPU tests for EVA in `test_gpu_examples.py`, so the tests were merged into that test class. The tests themselves remain unchanged. I left the lightweight EVA tests (not involving models or data) in `test_initialization.py.`

While working on this, I saw that the type annotations in `eva.py` were using `callable`, which is not a known type, I replaced those instances with `Callable`. Still not very useful but at least that exists.

Note: I ran the Eva tests locally on GPU to confirm they pass.


tests/test_gpu_examples.py
```diff
@@ -5075,36 +5091,47 @@ def is_bnb_model(self, model):
...
+            state_dicts.append(sd)
 
         cos_sims = defaultdict(list)
         for i, j in itertools.combinations(range(self.NUM_SEEDS), 2):
             for k, v1 in state_dicts[i].items():
                 v2 = state_dicts[j][k]
                 min_size = min(v1.size(0), v2.size(0))
-                cos_sims[k].extend(torch.cosine_similarity(v1[:min_size], v2[:min_size], dim=1).abs().tolist())
+                cos_sims[k].extend(
+                    torch.cosine_similarity(v1[:min_size].abs(), v2[:min_size].abs(), dim=1).abs().tolist()
```
BenjaminBossan, line 5149: Note: I added the `.abs()` call here, same as in the original `test_initialization.py` test. This call is necessary to make the tests pass, otherwise similarity is below the threshold (0.6183). Previously, this wasn't necessary as we would only test a single `EvaConfig` setting (which passes), but now we parametrize over 4 different configs (as in the original `test_initialization.py`) and those other configs fail without the `.abs()` call.

BenjaminBossan:
Note: Failing tests are unrelated.

PR #3109, opened 2026-03-17
FIX Deal with missing attribute on model config

BenjaminBossan:
Resolves the [failing tests on transformers main branch](https://github.com/huggingface/peft/actions/runs/23189520428/job/67381613348#step:6:10905).

After the change in https://github.com/huggingface/transformers/pull/41250, the `num_hidden_layers` attribute is no longer part of the model config when serialized to a dict. The `_prepare_prompt_learning_config` function was using this attribute. Therefore, we now pass the config before converting it into a dict and extract the attribute from it.


BenjaminBossan:
Failing CI is unrelated, I'll merge the PR as it's required for prompt learning methods to work correctly with Transformers v5.4

PR #3111, opened 2026-03-18
FIX Several bugs when adding merged LoRA weights

BenjaminBossan:
To be clear: This is about combining different LoRA adapters into a single one (`add_weighted_adapter`), not about merging LoRA weights into the base weights (`merge`).

There were a few issues being addressed:

1. In the SVD path, the LoRA scaling was applied to the combination weightings (!= LoRA weights) but `get_delta_weight` already takes account of the scaling, meaning that scaling was effectively applied twice.
2. In `_test_weighted_combination_of_adapters_lora` in `testing_common.py`, the sign was applied twice, canceling out for negative weights.
3. The bug in 2. was masked because in `_test_weighted_combination_of_adapters`, the `_test_weighted_combination_of_adapters_lora` method was called twice with the same model. Since, after the first time, the combined adapters with the given names already existed, `add_weighted_adapter` simply skips the creation of a new adapter. Therefore, the second call basically did nothing. This is now fixed by using a new model for each call. **Update**: Independently discovered and fixed in #3559; I kept the fix from this PR, overriding the other fix.
4. The `test_add_weighted_adapter_negative_weight_with_different_scaling` test did not actually test that the expected weight is equal to the actual weight and it also only tested the linear combination type, meaning that errors like in 1. were missed. This test now tests multiple combination types with different weightings and checks the resulting delta weight. ~~Note that I only included `svd`, `cat` (both precise), and `linear` (imprecise). The other methods (`dare`, `ties`, `magnitude` pruning) had huge errors in my testing.~~ Update: I added `dare`, `ties`, `magnitude` but with higher error margin.

While working on this, I also fixed:

- Add more precise type annotation for combination_type
- Better document the options
- Add comments for better understanding
- Use more consistent names in tests

Note: Since this PR fixes multiple, partly interdependent, issues, I left comments to point to the corresponding issue.


src/peft/tuners/lora/model.py
```diff
@@ -738,11 +752,12 @@ def _svd_generalized_task_arithmetic_weighted_adapter(
         for adapter, weight in zip(adapters, weights):
             if adapter in target.lora_A or adapter in target.lora_embedding_A:
                 valid_adapters.append(adapter)
-                valid_weights.append(weight * target.scaling[adapter])
+                valid_weights.append(weight)
```
BenjaminBossan, line 853: Note: This is issue 1.

tests/test_custom_models.py
```diff
@@ -3326,9 +3326,19 @@ def test_add_weighted_adapter_subtraction_with_negative_weights(self):
                 dw_cancelled = module.get_delta_weight("cancelled")
                 assert torch.allclose(dw_cancelled, torch.zeros_like(dw_cancelled))
 
-    def test_add_weighted_adapter_negative_weight_with_different_scaling(self):
-        # Test negative weights with different scaling factors (lora_alpha)
-        # This edge case ensures negative weights work correctly with different scaling values
+    @pytest.mark.parametrize("weights", [[1.0, 1.0], [0.0, 1.0], [5.0, 0.01], [-1.0, -1.0], [0.5, -0.3]])
```
BenjaminBossan, line 4170: Note: This is issue 4.

tests/test_custom_models.py
```diff
@@ -3337,36 +3347,46 @@ def test_add_weighted_adapter_negative_weight_with_different_scaling(self):
             r=8,
             lora_alpha=16,  # scaling = 16/8 = 2
             target_modules=["lin0"],
-            lora_dropout=0.0,
-            bias="none",
```
BenjaminBossan, line 4444: This was removed because it's the default anyway, so it's less noise.

tests/test_custom_models.py
```diff
@@ -3337,36 +3347,46 @@ def test_add_weighted_adapter_negative_weight_with_different_scaling(self):
...
-        model.set_adapter("merged_diff_scaling")
+        model.set_adapter("merged")
         dummy_input = torch.randn(2, 10)
         output = model(dummy_input)
         assert output is not None
 
+        # We cannot expect the merged weights to be approximately equal because we're dealing with rough approximations.
+        # Therefore, we check for correlation to verify that the direction is right and MSE to verify that the magnitude
+        # is right.
+        for module in model.modules():
```
BenjaminBossan, line 4227: Note: This is issue 4.

tests/testing_common.py
```diff
@@ -1614,11 +1600,11 @@ def _test_weighted_combination_of_adapters_lora(self, model, config, adapter_lis
             _, target, _ = _get_submodules(model, key)
             if isinstance(target, LoraLayer):
                 for adapter_name in new_adapters:
+                    # for a single adapter, the result should be exact and we can check that; otherwise, we deal with
+                    # approximations
                     if "single" in adapter_name:
                         new_delta_weight = target.get_delta_weight(adapter_name)
                         weighted_original_delta_weights = target.get_delta_weight(adapter_list[0]) * weight_list[0]
-                        sign = 1 if weight_list[0] > 0 else -1
```
BenjaminBossan: Note: This is issue 2.

tests/testing_common.py
```diff
@@ -1690,11 +1676,22 @@ def _test_weighted_combination_of_adapters(self, model_id, config_cls, config_kw
...
+            # test positive weights
             if isinstance(config, LoraConfig):
                 self._test_weighted_combination_of_adapters_lora(model, config, adapter_list, weight_list)
-                self._test_weighted_combination_of_adapters_lora(model, config, adapter_list, negative_weight_list)
             elif isinstance(config, IA3Config):
                 self._test_weighted_combination_of_adapters_ia3(model, config, adapter_list, weight_list)
+            else:
+                pytest.skip(f"Test not applicable for {config}")
+
+            del model
```
BenjaminBossan, line 1745: Note: This is issue 3.

BenjaminBossan:
_not stale_

BenjaminBossan:
_not stale_

BenjaminBossan:
_not stale_

BenjaminBossan:
_not stale_

BenjaminBossan:
_not stale_

tests/testing_common.py
githubnemo: This I don't understand. The description says that `_test_weighted_combination_of_adapters_lora` was called twice with the same model but the old code calls both variants with `create_model()` for model, so I don't see how the explanation matches the diff. Am I missing something?

githubnemo (commented):
Overall this looks good. One question regarding the description vs. the actual code, otherwise LGTM

tests/testing_common.py
BenjaminBossan: This is what the update in the PR description refers to:

> Update: Independently discovered and fixed in https://github.com/huggingface/peft/pull/3559; I kept the fix from this PR, overriding the other fix.

BenjaminBossan:
Thanks for the review. Regarding your remaining question, I replied to it, but I can't see the reply here, so copying it to ensure it doesn't get lost:

> This I don't understand. The description says that _test_weighted_combination_of_adapters_lora was called twice with the same model but the old code calls both variants with create_model() for model, so I don't see how the explanation matches the diff. Am I missing something?

This is what the update in the PR description refers to:

> Update: Independently discovered and fixed in https://github.com/huggingface/peft/pull/3559; I kept the fix from this PR, overriding the other fix.

githubnemo (approved):
Ah, thanks! LGTM

PR #3083, opened 2026-03-05
Changes for transformers 5 weight conversion

BenjaminBossan:
See accompanying https://github.com/huggingface/transformers/pull/44478.

- better handling of swapped in and out features
- move PEFT config update functions to PEFT
- move PEFT-specific weight conversion logic to PEFT

Note that the newly added tests will fail until a new transformers release with the linked PR is out. This should be v5.4, so the corresponding tests only run with that transformers version. I locally tested with the current main branch and the tests pass.

PR #3112, opened 2026-03-19
[Bugfix] Fix incorrect MoE LoRA weight shape

jeejeelee:
When I tried to use PEFT to train a LoRA on `Qwen/Qwen3.5-35B-A3B`, I found that the generated LoRA weights were incorrect, see: [jeeejeee/qwen35-35ba3b-sql-lora](https://huggingface.co/jeeejeee/qwen35-35ba3b-sql-lora). 
The reason is that there are two kinds of MoE parameter layouts in Transformers: like [modeling_gpt_oss](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt_oss/modeling_gpt_oss.py#L75) and [modeling_qwen3_5_moe](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py#L810). PEFT does not differentiate between them, which causes incorrect LoRA weights to be generated for qwen3-moe.
You also can refer to [gpt-oss-20b-lora-adapter-text2sql](https://huggingface.co/jeeejeee/gpt-oss-20b-lora-adapter-text2sql)

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -2069,15 +2069,25 @@ class ParamWrapper(nn.Module, LoraLayer):
         # For ParamWrapper, we don't derive the in_features and out_features based on the base layer type, but directly
         # from the targeted parameter.
         param = self.get_param()
-        if param.ndim == 3:
-            num_experts, in_features, out_features = param.shape
-        else:
-            num_experts, in_features, out_features = 1, param.shape[1], param.shape[0]
         if param.ndim not in (2, 3):
             raise ValueError(
                 f"lora.{self.__class__.__name__} was initialized with {param.ndim} dimensional Parameter, but only 2d "
                 "and 3d are supported."
             )
+        if param.ndim == 3:
+            base_layer = self.get_base_layer()
+            # Auto-detect weight layout from the base module's is_transposed attribute
+            # (set by transformers' @use_experts_implementation decorator).
+            #   is_transposed=True  -> weights are (E, in_features, out_features), e.g. GPT-OSS
+            #   is_transposed=False -> weights are (E, out_features, in_features), e.g. Qwen3.5 MoE
+            self._is_weight_transposed = getattr(base_layer, "is_transposed", False)
+            if self._is_weight_transposed:
+                num_experts, in_features, out_features = param.shape
+            else:
+                num_experts, out_features, in_features = param.shape
+        else:
+            self._is_weight_transposed = True
+            num_experts, in_features, out_features = 1, param.shape[1], param.shape[0]
         # we have to store the num_experts attribute here, as the parent class only stores in_features and out_features.
         self.num_experts = num_experts
         return in_features, out_features
@@ -2197,8 +2207,11 @@ class ParamWrapper(nn.Module, LoraLayer):
             weight_A = weight_A.reshape(self.num_experts, -1, weight_A.shape[-1])
             # shape: out_features x rank x experts
             weight_B = weight_B.reshape(weight_B.shape[0], -1, self.num_experts)
-            # fan_in_fan_out must be False, so no transpose call here
+            # einsum produces (E, in_features, out_features)
             delta_weight = torch.einsum("o r e, e r i -> e i o", weight_B, weight_A) * self.scaling[adapter_name]
+            if not self._is_weight_transposed:
+                # base param is (E, out_features, in_features), transpose delta to match
+                delta_weight = delta_weight.transpose(1, 2)
 
         base_layer = self.get_base_layer()
         param = self.get_param()


BenjaminBossan:
Thanks for the PR, but I think that #3083 should resolve your issue. Could you give this branch a try?

Regarding your implementation: I see that you're using `module.is_transposed`, which I didn't know existed. I checked where it's being set and only found it here:

https://github.com/huggingface/transformers/blob/f4ed5b8d3600eb302c0031129fe51684ae004c8d/src/transformers/models/gpt_oss/modeling_gpt_oss.py#L68

(though it's hard to find because it's dynamically set)

Do you know if/where it's set for Qwen 3.5?

jeejeelee:
@BenjaminBossan  Regarding `is_transposed`, see: https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/moe.py#L465

BenjaminBossan:
Yes, but is it set on the corresponding Qwen modules?

jeejeelee:
> Yes, but is it set on the corresponding Qwen modules?

If I understand your quesiton correcty, I think this is set on [Qwen3_5MoeExperts](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py#L801), and `is_transposed=False` by default

BenjaminBossan:
> If I understand your quesiton correcty, I think this is set on [Qwen3_5MoeExperts](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py#L801), and `is_transposed=False` by default

I wanted to find where it's set, but I found it now, it's through the `@use_experts_implementation` decorator. Yeah, it could be that I should use this attribute in my PR. I'll investigate. In the meantime, could you please check if using my branch resolves your original problem and report back?

_Update_: I see now that you linked to that line, for some reason when I clicked the link it didn't jump there. Anyway, I updated my PR to use the attribute now, thanks for the tip!

jeejeelee:
@BenjaminBossan SG, I will verify your branch if I have chance , thank you

linitra24:
@BenjaminBossan Hi! I tried `#3083`, and from my testing it does not fully resolve the issue described in `#3112`.

My understanding is that `#3083` improves Transformers 5 weight conversion and checkpoint compatibility in PEFT. However, `#3112` addresses a different problem: for models such as Qwen3.5-MoE / Qwen3-VL-MoE, PEFT can still construct MoE LoRA weights with incorrect shapes.

So these two PRs seem related, but they are not fixing the same thing:
- `#3083` improves weight conversion / compatibility
- `#3112` fixes incorrect MoE LoRA weight shapes for this class of models

This is also not only a shape-inspection issue: in downstream runtimes such as vLLM, these LoRA adapters may fail to load correctly.

So from my testing, `#3112` still seems necessary.

BenjaminBossan:
@linitra24 Thanks for investigating and reporting your issue. For us to work on that, could you please share what exactly is incorrect? Ideally, you can provide reproducer code so that we can check.

linitra24:
Of course — I’ll upload my test code and the experimental results as soon as possible to demonstrate the issue this PR is addressing.

BenjaminBossan:
> Of course — I’ll upload my test code and the experimental results as soon as possible to demonstrate the issue this PR is addressing.

Thanks. I'm also investigating this right now and I think I discovered a bug in PEFT. I'll let you know when I create a PR for it.

linitra24:
I did a more concrete repro for this issue and I think it may help clarify why `#3112` is still needed.

I will upload the following artifacts:
- base model: [`EpochEcho/Qwen3-VL-MoE-Tiny`](https://huggingface.co/EpochEcho/Qwen3-VL-MoE-Tiny)
- LoRA generated with Transformers 4.x: [`EpochEcho/qwen3-vl-moe-tiny-lora-transformers4.x`](https://huggingface.co/EpochEcho/qwen3-vl-moe-tiny-lora-transformers4.x)
- LoRA generated with Transformers 5.x: [`EpochEcho/qwen3-vl-moe-tiny-lora-transformers5.x`](https://huggingface.co/EpochEcho/qwen3-vl-moe-tiny-lora-transformers5.x)

The two LoRA adapters were generated from the same script, using the same base model and the same PEFT config, and only changing the Transformers version:
```python
import torch
from transformers import Qwen3VLMoeForConditionalGeneration
from peft import get_peft_model, LoraConfig, TaskType

model_name_or_path = "EpochEcho/Qwen3-VL-MoE-Tiny"

peft_config = LoraConfig(
    task_type=TaskType.CAUSAL_LM,
    inference_mode=True,
    r=8,
    lora_alpha=32,
    target_parameters=[
        "mlp.experts.gate_up_proj",
        "mlp.experts.down_proj",
    ],
)

model = Qwen3VLMoeForConditionalGeneration.from_pretrained(
    model_name_or_path,
    trust_remote_code=True,
    torch_dtype=torch.bfloat16,
    local_files_only=True,
).eval()

model.save_pretrained(...)
```

For the tiny model I used:
- `num_experts = 4`
- `r = 8`
- `hidden_size = 128`
- `moe_intermediate_size = 48`
- so `num_experts * r = 32`
- and `2 * moe_intermediate_size = 96`

The saved LoRA shapes are different across the two Transformers versions.

For `gate_up_proj`:
- Transformers 4.x saves `...mlp.experts.base_layer.lora_A.weight` with shape `(32, 128)`, i.e. `(num_experts * rank, hidden_size)`
- Transformers 5.x saves `...mlp.experts.base_layer.lora_A.weight` with shape `(32, 96)`, i.e. `(num_experts * rank, 2 * moe_intermediate_size)`

The full saved shapes from my repro are:

Transformers 4.x:
- `base_layer.lora_A.weight`: `(32, 128)`
- `base_layer.lora_B.weight`: `(96, 32)`
- `experts.lora_A.weight`: `(32, 48)`
- `experts.lora_B.weight`: `(128, 32)`

Transformers 5.x:
- `base_layer.lora_A.weight`: `(32, 96)`
- `base_layer.lora_B.weight`: `(128, 32)`
- `experts.lora_A.weight`: `(32, 128)`
- `experts.lora_B.weight`: `(48, 32)`

So this is not only about whether the shapes “look unusual”: the same script produces structurally different MoE LoRA weights depending on the Transformers version.

I also tested these adapters in vLLM 0.19.0 with:
```python
from vllm import LLM, SamplingParams
from vllm.lora.request import LoRARequest

llm = LLM(
    model="EpochEcho/Qwen3-VL-MoE-Tiny",
    enable_lora=True,
    max_loras=1,
    max_lora_rank=8,
)

outputs = llm.generate(
    ["Hello"],
    SamplingParams(max_tokens=4, temperature=0),
    lora_request=LoRARequest(
        "default",
        1,
        "<LORA_PATH>",
    ),
)
```

What I observed:
- with the adapter generated by Transformers 4.x, vLLM is able to load it normally in my test.
- with the adapter generated by Transformers 5.x, vLLM fails during LoRA activation with:

```text
RuntimeError: The size of tensor a (48) must match the size of tensor b (128) at non-singleton dimension 2
```

The failing stack in my repro ends at:
- `vllm/lora/layers/fused_moe.py`, line 733

So from my testing:
- `#3083` and `#3112` are related, but they are not fixing the same thing
- `#3083` improves Transformers 5 conversion / compatibility
- `#3112` addresses a real remaining issue: PEFT can still produce MoE LoRA weights whose layout is inconsistent across Transformers versions, and this causes downstream loading problems in vLLM

That is why I think `#3112` is still needed.

BenjaminBossan:
Thanks @linitra24 for posting this. I haven't checked it yet, but I just now created a PR for PEFT that fixes an error in the logic: #3165. Ideally, you could test this branch with your original issue and tell me if it is resolved.

BenjaminBossan:
This should now be fixed via #3165. Patch release 0.19.1 is out. Thanks for your help.

PR #3117, opened 2026-03-25
Generic quantization support for BOFT, MiSS, VeRA

BenjaminBossan:
## Problem

Right now, if a new PEFT method wants to add support for quantized layers, it requires a significant amount of work. Notably, the method needs to implement dedicated layer classes for each quantization method (e.g. one class for bnb 4bit, one for bnb 8bit, one for AWQ, ...). These classes typically are >90% boilerplate and the actual difference between implementations of these classes is minimal.

The result of that is that, at the moment, most PEFT methods don't support any, or only very few, quantization methods, even though the amount of actual logic required to support these methods is relatively small.

## Suggested solution

This PR is a suggestion of how to solve the issue. If this approach is accepted, with a few extra lines, we should be able to support all quantization methods in all PEFT methods. The general approach is to add an attribute to each PEFT layer, `self.quantization_backend`, which supports these methods:

- `get_base_weight`
- `set_base_weight`

When the PEFT layers use these methods to access and write to the base layer weight, and if the weight is quantized, the new classes will deal with that correctly. This means that we no longer need a dedicated layer class to deal with quantized layers, the normal layer class will do. E.g. for MiSS, the normal `miss.Linear` class can deal with bnb layers, there is no need to add a `miss/bnb.py` module with dedicated layers.

A few rewrites in the existing PEFT methods are required to support this new quantization backend class, but the amount of total code needed for that is considerably smaller than adding new classes for each quantization method. 

Furthermore, these quantization backend classes are agnostic with regard to the PEFT method. Therefore, with M PEFT methods and N quantization methods, we no longer need MxN implementations to support quantization but only M+N.

## Migration

For LoRA, we have already implemented the layer classes for each supported quantization method. For the sake of consistency, it could still make sense to migrate LoRA to the new approach if it's accepted. This needs to be accompanied by detailed regression testing to ensure that everything keeps working. I would only suggest to deprecate and remove abandoned quantization methods (perhaps for a v1.0 release).

The bigger issue, however, is that packages that depend on PEFT may break with this change. As an example, if they detect quantized layers via `isinstance` checks, those would break as all layers would just be normal `lora.Linear`, `lora.Conv2d` etc. The approach here would most likely involve deprecating the import of these classes. I think it's also possible to "cheat" `isinstance` and pretend like there is inheritance when there isn't but I'd like to avoid that.

Anyway, this is out of scope of this PR and will be addressed in the future.

## Scope

Updating all PEFT methods is too much for a single PR. This PR focuses on only three PEFT methods for now:

- MiSS: A pretty normal PEFT method, representative of many other PEFT methods.
- BOFT: Also pretty normal, but requires slight rewrite of the `forward` step. Similar changes may be required for other methods too.
- VeRA: Already supports bnb but with this PR, specific bnb layers are no longer needed.


src/peft/tuners/boft/layer.py
```diff
@@ -641,21 +642,16 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
...
-
-            orig_weight = self.get_base_layer().weight.data
-            orig_weight = torch.transpose(orig_weight, 0, 1)
             boft_rotation = boft_rotation.to(previous_dtype)
-            orig_weight = orig_weight.to(previous_dtype)
-            rotated_weight = torch.mm(boft_rotation, orig_weight)
-            rotated_weight = torch.transpose(rotated_weight, 0, 1)
-
-            scaled_rotated_weight = rotated_weight * boft_scale
+            x_rotated = x @ boft_rotation
```
BenjaminBossan, line 640: This is a reformulation of the forward path of BOFT that avoids using the base layer weight directly. This was necessary because calling `torch.mm(boft_rotation, orig_weight)` can fail with quantized weights. Instead, we should make a forward pass and let the quantized layer handle the details. I ran the BOFT tests with the old and the new implementation and added an assert that they are identical (up to precision).

Regarding runtime, I checked the MetaMath benchmark and got 147 sec for 250 steps (116 sec for 1 eval run) using main branch, and 138 sec (108 sec) using the code from this branch. So the new code seems to be on par or possibly slightly faster than the old one.

BenjaminBossan:
_not stale_

BenjaminBossan:
@Joluck It would be great if you could check the MiSS change.
@zqiu24 It would be great if you could check the BOFT change.

githubnemo (approved):
As discussed offlne this is good! :)

Let's merge this and see if there are any surprises left.

PR #3146, opened 2026-04-10
FIX Explicit weight conversion map for Mixtral

BenjaminBossan:
See https://github.com/huggingface/transformers/pull/45340#issuecomment-4222734042


BenjaminBossan:
@Cyrilvallez Would that be the right fix?

githubnemo (approved):
Okay!

PR #3153, opened 2026-04-13
Add local smoke test script

varun566:
## Summary
- add a small local smoke test script for PEFT development checkouts
- verify the editable install can wrap a tiny PyTorch model with LoRA
- print basic package and trainable-parameter information for quick local validation

## Verification
- `source .venv/bin/activate && python scripts/smoke_test.py`
- `source .venv/bin/activate && pytest tests/test_mapping.py tests/test_config.py -k 'not from_pretrained and not cache_dir_remote and not from_pretrained_cache_dir'`

### DIFF
--- a/scripts/smoke_test.py
+++ b/scripts/smoke_test.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python
+"""Basic local smoke test for a PEFT development checkout."""
+
+import torch
+
+import peft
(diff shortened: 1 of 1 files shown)


githubnemo:
Thanks but I don't think that a small smoke test like this adds value to the existing test suite. If you've experienced issues with local editable installs that warrant such a test, please open an issue first!

PR #3154, opened 2026-04-13
LoRA dispatch: surface inner-`.linear` hint for wrapper modules (fixes #3129 UX)

dentity007:
## Summary

Addresses #3129. When a user's `target_modules` regex matches a wrapper module that contains a single `nn.Linear` child, the current dispatch error gives no guidance on how to recover. This is a real footgun on Gemma 4, whose `Gemma4ClippableLinear` inherits from `nn.Module` (not `nn.Linear`) and wraps a plain `nn.Linear` in its `.linear` attribute. It is used in the audio and vision towers. A user attempting `LoraConfig(target_modules=["q_proj", "k_proj", "v_proj"])` to LoRA-ify the language model hits this wrapper in the audio tower and gets stuck.

@BenjaminBossan already identified the right workaround in the issue thread: target the inner `.linear` attribute with a more specific regex like `target_modules=r".*\.audio_tower.*\.(q_proj|v_proj)\.linear"`. This works correctly. The PEFT dispatch sees the inner `nn.Linear`, wraps it with a LoRA layer, and the wrapper's forward still applies its clipping around the LoRA-augmented linear operation.

This PR surfaces that workaround in the error message so future users do not have to find the issue thread to discover it.

## What Changed

**`src/peft/tuners/lora/model.py`** - When the existing "Target module ... is not supported" error fires, inspect the target. If it has exactly one `nn.Linear` child, append a hint paragraph with:

- The wrapper class name
- The attribute name that contains the inner linear
- A sample regex the user can adapt
- A link back to #3129 for context

The original error text is preserved so existing string matches (e.g. in `test_initialization.py::test_road_with_conv2d_layer`) continue to work. The hint is additive.

**`tests/test_initialization.py`** - Regression test in `TestLoraInitialization` that:
1. Asserts the new hint fires with the expected content (wrapper class name, `.linear` attribute name, sample regex, issue link)
2. Confirms the `.linear` regex workaround actually produces a valid LoRA adapter and supports forward pass through the wrapper

The test uses a minimal `ClippableLinearWrapper` class that reproduces the exact structure of `Gemma4ClippableLinear` without requiring transformers or a downloaded model.

## Verification

Built and tested on NVIDIA DGX Spark GB10 (torch 2.11.0+cu130, transformers 5.5.4, peft 0.18.2.dev0 from this branch).

Repro before fix (stock peft 0.18.1):

```
ValueError: Target module Gemma4ClippableLinear(
  (linear): Linear(in_features=64, out_features=64, bias=False)
) is not supported. Currently, only the following modules are supported: ...
```

After fix, same invocation:

```
ValueError: Target module Gemma4ClippableLinear(
  (linear): Linear(in_features=64, out_features=64, bias=False)
) is not supported. Currently, only the following modules are supported: `torch.nn.Linear`, `torch.nn.Embedding`, `torch.nn.Conv1d`, `torch.nn.Conv2d`, `torch.nn.Conv3d`, `transformers.pytorch_utils.Conv1D`, `torch.nn.MultiheadAttention.`.

Hint: `Gemma4ClippableLinear` appears to be a wrapper module that contains a single `nn.Linear` in its `.linear` attribute. You can target the inner linear by appending `\.linear` to your `target_modules` regex, for example `target_modules=r".*\.linear"` or a more specific pattern like `target_modules=r".*(q_proj|v_proj)\.linear"`. See https://github.com/huggingface/peft/issues/3129 for a worked example with `Gemma4ClippableLinear`.
```

Full `TestLoraInitialization` suite: **114 passed, 0 regressions** (including the new test).

## Why This Approach

I considered a more invasive fix that would automatically drill into `.linear` submodules and replace them in place. It would be cleaner from the user's perspective (no regex change needed) but would require restructuring `_create_and_replace` to operate on nested parents, which risks side effects in the other tuner types that share the dispatch path. The hint-only fix is minimal, targeted, backward-compatible, and resolves the UX problem without changing dispatch semantics. If you would prefer the auto-drilling approach, happy to rework.

## Related

- Issue: #3129
- Workaround originally suggested by @BenjaminBossan in the issue thread
- Verified on Gemma 4 `Gemma4ClippableLinear` from transformers 5.5.4

### DIFF
--- a/src/peft/tuners/lora/model.py
+++ b/src/peft/tuners/lora/model.py
@@ -421,10 +421,42 @@ class LoraModel(BaseTuner):
 
         if new_module is None:
             # no module could be matched
-            raise ValueError(
-                f"Target module {target} is not supported. Currently, only the following modules are supported: "
+            supported_msg = (
+                "Currently, only the following modules are supported: "
                 "`torch.nn.Linear`, `torch.nn.Embedding`, `torch.nn.Conv1d`, `torch.nn.Conv2d`, `torch.nn.Conv3d`, "
                 "`transformers.pytorch_utils.Conv1D`, `torch.nn.MultiheadAttention.`."
+            )
+            # Check for a common failure mode: the target is a wrapper module that
+            # contains a single `nn.Linear` in a submodule (typically named `.linear`).
+            # Examples include `transformers.models.gemma4.modeling_gemma4.Gemma4ClippableLinear`,
+            # which inherits from `nn.Module` rather than `nn.Linear`. In these cases
+            # the user can target the inner linear directly by appending its attribute
+            # name to their `target_modules` regex. Surface this hint in the error so
+            # users do not have to guess.
+            wrapper_hint = ""
+            if isinstance(target, torch.nn.Module):
+                inner_linear_attrs = [
+                    attr_name
+                    for attr_name, child in target.named_children()
+                    if isinstance(child, torch.nn.Linear)
+                ]
+                # Only offer the hint when the wrapper contains exactly one inner
+                # `nn.Linear`. Multiple inner linears would be ambiguous and are
+                # out of scope for this error message.
+                if len(inner_linear_attrs) == 1:
+                    inner_attr = inner_linear_attrs[0]
+                    target_cls_name = type(target).__name__
+                    wrapper_hint = (
+                        f"\n\nHint: `{target_cls_name}` appears to be a wrapper module that contains a "
+                        f"single `nn.Linear` in its `.{inner_attr}` attribute. You can target the inner "
+                        f"linear by appending `\\.{inner_attr}` to your `target_modules` regex, for example "
+                        f"`target_modules=r\".*\\.{inner_attr}\"` or a more specific pattern like "
+                        f"`target_modules=r\".*(q_proj|v_proj)\\.{inner_attr}\"`. See "
+                        f"https://github.com/huggingface/peft/issues/3129 for a worked example with "
+                        f"`Gemma4ClippableLinear`."
+                    )
+            raise ValueError(
+                f"Target module {target} is not supported. {supported_msg}{wrapper_hint}"
             )
 
         return new_module
--- a/tests/test_initialization.py
+++ b/tests/test_initialization.py
@@ -1613,6 +1613,58 @@ class TestLoraInitialization:
         config2 = LoraConfig(target_modules=["linear"], bias="none")
         model.add_adapter("other", config2)  # does not raise
 
+    def test_dispatch_error_hints_inner_linear_attribute(self):
+        # When a target module is an unsupported wrapper that contains a single
+        # nn.Linear in a submodule attribute (e.g. the `.linear` in Gemma 4's
+        # Gemma4ClippableLinear), the error message should explicitly suggest
+        # targeting the inner attribute. Regression test for issue #3129.
+        class ClippableLinearWrapper(nn.Module):
+            """Minimal reproduction of transformers.Gemma4ClippableLinear."""
+
+            def __init__(self, in_features: int, out_features: int):
+                super().__init__()
+                self.linear = nn.Linear(in_features, out_features, bias=False)
+
+            def forward(self, x):
+                return self.linear(x)
+
+        class WrapperModel(nn.Module):
+            def __init__(self, dim: int = 16):
+                super().__init__()
+                self.q_proj = ClippableLinearWrapper(dim, dim)
+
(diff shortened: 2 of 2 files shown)


BenjaminBossan:
Thanks for the PR. I agree that having a helpful error message for when the user targeted the wrong module would be helpful. This PR, however, is overly specific to Gemma 4. With #3136 being merged and the next release being very soon, the issue for Gemma 4 should be even less urgent.

I think what we could consider is something else: Let's assume that the user targets modules that are not valid, or none of the specified targets exist. We could run a helper function on top of the whole model to identify potential target modules and suggest the user target those, then reraise.

This is, however, a much bigger endeavor.

1. We need to intercept exactly the right errors in `tuners_utils.py`.
2. We need to identify the right modules that _could_ be targeted, which can differ based on the PEFT method.
3. We don't want to print a list of potentially hundreds of target modules (`model.encoder.layers.0.q_proj`, ...), instead we should shorten the list (`q_proj, ...`). This may be achievable with `_find_minimal_target_modules`.

If you, or someone else, is interested in working on this, LMK. For this PR, I think it's not needed as I explained above, thus closing it.

dentity007:
Thanks @BenjaminBossan, that makes sense. Didn't realize #3136 was already merged, good call on closing this one out.

I'd like to take on the broader error suggestion work. Before I dive in though, I've got a few questions so I can scope it right.

First, scope. Are you thinking this covers all PEFT methods (LoRA, IA3, LoKr, LoHa, OFT, etc.), or would you rather I start with LoRA and generalize once the pattern's solid?

Second, error location. You mentioned tuners_utils.py. Is check_target_module_exists the right entry point, or should I be hooking further up in BaseTuner.inject_adapter?

Third, suggestion heuristic. For _find_minimal_target_modules, do you want the suggestions capped at a top N, or should I use suffix compression regardless of how long the list gets? I'm guessing you don't want users staring at a wall of q_proj paths.

Fourth, tuner specific targets. IA3 has separate categories (feedforward vs attention). Should the suggester distinguish between them, or just show everything and let the user sort it out?

Happy to start smaller with a LoRA only scope and a design doc PR before any actual code if you'd prefer that path. Timeline wise, I'm probably looking at a 1 to 2 week turnaround once we've locked down scope.

BenjaminBossan:
> First, scope. Are you thinking this covers all PEFT methods (LoRA, IA3, LoKr, LoHa, OFT, etc.), or would you rather I start with LoRA and generalize once the pattern's solid?

It would be fine to start out with LoRA only. BUT: There must be a clear path on how to extend to all PEFT methods. So something like hard-coding the supported layer types is not an option.

> Second, error location. You mentioned tuners_utils.py. Is check_target_module_exists the right entry point, or should I be hooking further up in BaseTuner.inject_adapter?

No, that's not the right place, as it only checks one individual key. We can only really say something went wrong after going through all of them. Then we raise an error based on what we found. The logic is already a bit complex and can be found here:

https://github.com/huggingface/peft/blob/6d5a6f4f2f902dbf13d21d2661d57c3c05df1dae/src/peft/tuners/tuners_utils.py#L959-L1043

> Third, suggestion heuristic. For _find_minimal_target_modules, do you want the suggestions capped at a top N, or should I use suffix compression regardless of how long the list gets? I'm guessing you don't want users staring at a wall of q_proj paths.

Hard to say. I think this needs thorough testing and good taste.

> Fourth, tuner specific targets. IA3 has separate categories (feedforward vs attention). Should the suggester distinguish between them, or just show everything and let the user sort it out?

IA³ is a bit of a special case in that regard. Similarly, there is a special case for LoRA targeting parameters directly (e.g. MoE parameters). I'd say, those can be ignored for now.

> Happy to start smaller with a LoRA only scope and a design doc PR before any actual code if you'd prefer that path. Timeline wise, I'm probably looking at a 1 to 2 week turnaround once we've locked down scope.

Great, you can also create a draft PR for early feedback. Lots of testing will be necessary to get this feature right.

dentity007:
Thanks @BenjaminBossan, that gives me exactly what I need. I'll start with a LoRA only implementation but design the extension hook up front so the path to IA3, LoKr, LoHa, and OFT is visible in the draft. Reading lines 959-1043 in tuners_utils.py first, then setting up a local dev environment and writing failing tests before any implementation code. I'll open a draft PR once I have something testable running, and we can iterate on the suggestion output format from there. Appreciate the detailed response.

dentity007:
Opened a draft of the general-purpose suggester at #3173. LoRA only for now with extension scaffolding for the others. Tagging @BenjaminBossan for feedback on the interface shape.

PR #3159, opened 2026-04-14
Implement VeLoRA and add tests.

roymiles:
This is a PR for [VeLoRA](https://arxiv.org/abs/2405.17991), which is a method for reducing the memory of backprop by compressing the input activations (complimentary to LoRA and other adapter approaches). Theoretically VeLoRA can be viewed through the lens of LoRA (which is discussed in the paper) and it will reduce the memory consumption at a small computational overhead. In general, applying VeLoRA is most effective on layers with large input activations (e.g. down projection layers).

This work was accepted to NeurIPS 2024 and the official code can be found [here](https://github.com/roymiles/velora). I added tests to check the memory usage has gone down and the model is still trainable (toy cifar10 sample). 

Hope this is all ok!

```
0.24s call     tests/test_velora.py::test_velora_reduces_saved_activation_memory_vs_full_finetuning
0.01s call     tests/test_velora.py::test_velora_injection_and_forward_matches_base_layer

(8 durations < 0.005s hidden.  Use -vv to show these durations.)
=========================== 6 passed, 1 skipped, 1 warning in 18.21s ===========================
```


BenjaminBossan:
Thanks for your PR to contribute VeLoRA to PEFT. I went through the paper and skimmed the code but haven't done a full on review yet. This method seems to be an interesting middle ground between normal training and training with gradient checkpointing in terms of memory vs compute tradeoff.

The first question that came to mind was that after reading the paper, it feels like VeLoRA could be implemented on top of LoRA (maybe as a variant). I wonder if you considered this possibility, as I think it would be much easier for users to adopt this way. If yes, did you come across some limitations in PEFT that prevent this approach from working?

roymiles:
Hi @BenjaminBossan,

I definitely agree! we did do some experiments with this in the original paper and show further memory reduction when combining with LoRA and QLoRA (and even other adapter based methods). I was unsure how best to integrate this as a tidy PR though. I can have a look into this if you think it would be better for easier adoption. Are there any example PRs that have this approach that I can reference?

image

image

BenjaminBossan:
A recent example of a LoRA variant would be #2895. The interface for LoRA variants is defined here: 

https://github.com/huggingface/peft/blob/5b6e5e993944bddbb9fb12d5540f2d17eba39ad8/src/peft/tuners/lora/layer.py#L50

Then we have code in several places that basically does this:

```python
if active_adapter in self.lora_variant:
    foo = ... # normal LoRA
else:
    foo = self.lora_variant[active_adapter].bar()
```

The API is not set in stone, if we need to extend it, it's possible (if it's not too specific). If you find it doesn't work, we can also proceed as is, but I think the possibility is worth investigating.

roymiles:
@BenjaminBossan  I have updated VeLoRA to inherit from LoRAVariant ([here](https://github.com/huggingface/peft/pull/3159/changes/212d12f3d67548856d92ca64ed4f3eaa20d18d5b#diff-a14f8c2e9d3ff608a836f59d47401dc0e66c19a8e7ea5c25434ab857da6c0eccR775)).

The interface is still just using get_peft_model and passing in the VeLoRAConfig ([here](https://github.com/huggingface/peft/pull/3159/changes/212d12f3d67548856d92ca64ed4f3eaa20d18d5b#diff-52bb23385855d92421538a1858d9022e75f0b918fcafe02893ee162b18f82eadL92)).

Hope that is ok but if any issues with this approach I can update! Though I do agree this is much cleaner than the original way as a separate adapter. 

[here](https://github.com/huggingface/peft/pull/3159/changes/212d12f3d67548856d92ca64ed4f3eaa20d18d5b#diff-24a141c266b7b714ae8fcc470f31bc283f7b0f5a671bbf6d5f092741fc374104R816) is the integration with lora layer as a different variant.

It still passes the tests:
```
(7 durations < 0.005s hidden.  Use -vv to show these durations.)
=========================================================================== 7 passed, 1 skipped in 62.26s (0:01:02) ===========================================================================
```

src/peft/tuners/lora/layer.py
```diff
@@ -101,7 +106,17 @@ class LoraLayer(BaseTunerLayer):
     # All names of layers that may contain (trainable) adapter weights
     adapter_layer_names: tuple[str, ...] = ("lora_A", "lora_B", "lora_embedding_A", "lora_embedding_B")
     # All names of other parameters that may contain adapter-related parameters
-    other_param_names: tuple[str, ...] = ("r", "lora_alpha", "scaling", "lora_dropout")
+    other_param_names: tuple[str, ...] = (
+        "r",
+        "lora_alpha",
+        "scaling",
+        "lora_dropout",
+        "lora_velora_embed",
```
BenjaminBossan: Let's not change the `other_param_names` here, as the new entries are only required when VeLoRA is being used. Instead, in the `VeLoRALinearVariant.init`, you can extend `other_param_names`. Check e.g. how it's done here: https://github.com/huggingface/peft/blob/76c37d4686cf7245a9a6998fd1e2619d62c9322d/src/peft/tuners/lora/variants.py#L473-L474

src/peft/tuners/lora/layer.py
```diff
@@ -121,6 +136,11 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
         self.use_rslora: dict[str, bool] = {}
         self.lora_bias: dict[str, bool] = {}
         self.lora_magnitude_vector = torch.nn.ModuleDict()  # for DoRA
+        self.lora_velora_embed = BufferDict({}, persistent=True)
```
BenjaminBossan: Same idea, move this logic into `VeLoRALinearVariant.init`. Example: https://github.com/huggingface/peft/blob/76c37d4686cf7245a9a6998fd1e2619d62c9322d/src/peft/tuners/lora/variants.py#L65-L66

src/peft/tuners/lora/layer.py
```diff
@@ -952,6 +977,27 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
...
+            custom_base_variant_adapter = None
+            for active_adapter in self.active_adapters:
+                variant = self.lora_variant.get(active_adapter)
+                if variant is None or variant.uses_base_result(self, active_adapter):
+                    continue
+                if custom_base_variant_adapter is not None or len(self.active_adapters) > 1:
+                    raise ValueError(
+                        "Only one active adapter using a custom base-forward variant is supported per layer."
+                    )
+                custom_base_variant_adapter = active_adapter
```
BenjaminBossan: IIUC, this logic exists to ensure that when VeLoRA is used, we can only have a single active adapter on the layer. Is this really true that VeLoRA cannot work with other adapters? Assuming it is, this check shouldn't be performed here, instead let's override `_check_new_adapter_config` on `LoraModel`. It should look like this:

```python
    def _check_new_adapter_config(self, config: LoraConfig) -> None:
        # check `self.peft_config`
        ...
        super()._check_new_adapter_config(config)
```

Then add a test for this to `tests/test_initialization.py`.

src/peft/tuners/lora/variants.py
```diff
@@ -771,6 +772,106 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
+            initialized = True
+        else:
+            embed = torch.zeros(group_dim, device=device, dtype=dtype)
+            initialized = False
+
+        module.lora_velora_embed[adapter_name] = embed
+        module.lora_velora_initialized[adapter_name] = torch.tensor(
+            initialized, device=device, dtype=torch.bool
+        )
+        module._move_adapter_to_device_of_base_layer(adapter_name)
```
BenjaminBossan, line 819: Is this call necessary?

src/peft/tuners/lora/variants.py
```diff
@@ -771,6 +772,106 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
+
+        if config.velora_init_type == "random":
+            embed = _normalize_projection(torch.randn(group_dim, device=device, dtype=dtype)).to(dtype=dtype)
+            initialized = True
+        else:
+            embed = torch.zeros(group_dim, device=device, dtype=dtype)
+            initialized = False
+
+        module.lora_velora_embed[adapter_name] = embed
+        module.lora_velora_initialized[adapter_name] = torch.tensor(
```
BenjaminBossan: So this is a tensor with a single boolean scalar? Can we not just use a Python bool?

src/peft/tuners/lora/layer.py
```diff
@@ -952,6 +977,27 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
...
+
+            if custom_base_variant_adapter is not None:
```
BenjaminBossan: I don't understand this logic. Why is VeLoRA not being used the same way as the other LoRA variants below? We would first compute the base layer result (`result = self.base_layer(x, *args, **kwargs)`) and then apply VeLoRA on top. To me, this looks like what VeLoRA expects.

src/peft/tuners/lora/layer.py
```diff
@@ -76,12 +76,17 @@ def merge_unsafe(module: LoraLayer, active_adapter: str, orig_weight: torch.Tens
     def unmerge(module: LoraLayer, active_adapter: str, orig_weight: torch.Tensor) -> torch.Tensor:
         """Remove the adapter weights from the original weights, then return them"""
 
+    @staticmethod
+    def uses_base_result(module: LoraLayer, active_adapter: str) -> bool:
+        """Whether the variant expects `LoraLayer.forward` to pre-compute the base-layer output."""
+        return True
```
BenjaminBossan: Why is this needed?

src/peft/tuners/lora/config.py
```diff
@@ -658,6 +658,42 @@ class LoraConfig(PeftConfig):
             )
         },
     )
+    use_velora: bool = field(
```
BenjaminBossan: Let's not add 4 different parameters tied to VeLoRA. Instead, use a single argument, e.g. `use_velora` or `veralora_config`. Then add a new config class here which contains `velora_num_groups`, `velora_scale`, and `velora_init_type`. Then the user will do:

```python
lora_config = LoraConfig(target_modules=[...], velora_config=VeloraConfig(...))
```

src/peft/tuners/lora/variants.py
```diff
@@ -771,6 +772,106 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
     return kwargs
 
 
+class VeLoRALinearVariant(LoraVariant):
```
BenjaminBossan: For consistency with other naming conventions in PEFT and for easier typing, let's do `VeloraLinearVariant` or `VeLoraLinearVarient`. Same with other class names containing `VeLoRA`.

src/peft/tuners/velora/__init__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2026-present the HuggingFace Inc. team.
```
BenjaminBossan: All of the content of `peft/tuners/velora`, except for `velora/layer.py`, shouldn't be necessary. Please remove all of it. For `velora/layer.py`, I think we only need `VeLoRAFunction`. Move it to `lora/velora.py` instead.

src/peft/tuners/velora/layer.py
```diff
@@ -0,0 +1,254 @@
...
+            grad_weight = grad_weight.to(dtype=weight.dtype)
+
+        if ctx.needs_input_grad[2]:
+            grad_bias = grad_output_2d.sum(dim=0)
+            grad_bias = grad_bias.to(dtype=ctx.bias_dtype)
+
+        return grad_input, grad_weight, grad_bias, None, None, None
+
+
+class VeloraLayer(BaseTunerLayer):
```
BenjaminBossan: Can be removed.

src/peft/tuners/velora/layer.py
```diff
@@ -0,0 +1,254 @@
...
+        subtokens = _reshape_to_grouped_subtokens(x.detach(), self.num_groups[adapter_name]).reshape(
+            -1, self.in_features // self.num_groups[adapter_name]
+        )
+        embed = _normalize_projection(subtokens.mean(dim=0))
+        target = self.velora_embed[adapter_name]
+        self.velora_embed[adapter_name] = embed.to(device=target.device, dtype=target.dtype)
+        self.velora_initialized[adapter_name] = True
+
+
+class VeloraLinear(nn.Linear, VeloraLayer):
```
BenjaminBossan: Can be removed.

src/peft/tuners/__init__.py
BenjaminBossan, line 1: I think that since VeLoRA is now a LoRA variant, you can remove all the changes inside of `peft/utils/`.

src/peft/__init__.py
BenjaminBossan, line 1: Changes can be removed.

tests/test_velora.py
```diff
@@ -0,0 +1,330 @@
...
+
+    lora_state = get_peft_model_state_dict(lora_model)
+    velora_state = get_peft_model_state_dict(velora_model)
+
+    assert lora_state.keys() == velora_state.keys()
+    for key in lora_state:
+        assert torch.equal(lora_state[key], velora_state[key]), f"Mismatch for {key}"
+
+
+def test_velora_requires_group_divisibility():
```
BenjaminBossan: Should be moved to `tests/test_initialization.py`.

tests/test_velora.py
```diff
@@ -0,0 +1,330 @@
...
+    model = MLP()
+    config = VeloraConfig(target_modules=["lin0"], r=4, velora_scale=1.0, num_groups=7)
+
+    with pytest.raises(ValueError, match="divisible"):
+        _ = get_peft_model(model, config)
+
+
+def test_velora_batch_average_once_initializes_projection_once():
```
BenjaminBossan: Not quite sure what this test does, could you please add a comment?

tests/test_velora.py
BenjaminBossan, line 1: I think many of these tests are not necessary because they're already covered by existing tests. We only need to add VeLoRA to the test matrix of those. As a first step, please add VeLoRA to the test matrix here:

https://github.com/huggingface/peft/blob/76c37d4686cf7245a9a6998fd1e2619d62c9322d/tests/test_custom_models.py#L173

Check the other examples above for how it's done. Then, call `pytest tests/test_custom_models.py -k velora -v` to ensure that the tests pass.

tests/test_velora.py
```diff
@@ -0,0 +1,330 @@
...
+    def unpack_hook(tensor):
+        return tensor
+
+    with torch.autograd.graph.saved_tensors_hooks(pack_hook, unpack_hook):
+        loss = loss_factory()
+    loss.backward()
+    return saved_bytes
+
+
+def test_velora_injection_and_forward_matches_base_layer():
```
BenjaminBossan: Can be removed once VeLoRA is added to the test matrix.

tests/test_velora.py
```diff
@@ -0,0 +1,330 @@
...
+
+    trainable_names = {name for name, param in velora_model.named_parameters() if param.requires_grad}
+    assert "base_model.model.lin0.base_layer.weight" not in trainable_names
+    assert "base_model.model.lin0.base_layer.bias" not in trainable_names
+    assert "base_model.model.lin0.lora_A.default.weight" in trainable_names
+    assert "base_model.model.lin0.lora_B.default.weight" in trainable_names
+    assert not velora_model.base_model.model.lin0.lora_velora_embed["default"].requires_grad
+
+
+def test_velora_wrapper_matches_lora_variant_path():
```
BenjaminBossan: Can be removed once VeLoRA is added to the test matrix.

tests/test_velora.py
```diff
@@ -0,0 +1,330 @@
...
+        output = model(x)
+        return (output - target).pow(2).mean()
+
+    x_lora = torch.randn(8, 4, 128, requires_grad=True)
+    x_velora = x_lora.detach().clone().requires_grad_(True)
+
+    lora_saved_bytes = _saved_tensor_bytes(lambda: make_loss(lora_model, x_lora))
+    velora_saved_bytes = _saved_tensor_bytes(lambda: make_loss(velora_model, x_velora))
+
+    print(f"{velora_saved_bytes=} bytes.")
```
BenjaminBossan: Let's not use any print statements in tests.

tests/test_velora.py
```diff
@@ -0,0 +1,330 @@
...
+    compressed = _compress_activations(x.detach(), embed.to(x.dtype), num_groups=4)
+    reconstructed = _reconstruct_activations(compressed, embed.to(x.dtype), in_features=16, velora_scale=0.5)
+
+    grad_output_2d = grad_output.reshape(-1, 10)
+    scaling = layer.scaling["default"]
+    lora_A_weight = layer.lora_A["default"].weight.detach()
+    lora_B_weight = layer.lora_B["default"].weight.detach()
+    after_A = F.linear(x.detach(), lora_A_weight).reshape(-1, lora_A_weight.shape[0])
+
+    expected_grad_lora_A = ((grad_output_2d * scaling) @ lora_B_weight).transpose(0, 1) @ reconstructed
```
BenjaminBossan, line 338: A short explanation, perhaps with a (reference to a) formula would be nice here.

tests/test_velora.py
```diff
@@ -0,0 +1,330 @@
...
+    expected_grad_lora_B = grad_output_2d.transpose(0, 1) @ after_A * scaling
+
+    assert layer.base_layer.weight.grad is None
+    assert torch.allclose(layer.lora_A["default"].weight.grad, expected_grad_lora_A, atol=1e-6, rtol=1e-5)
+    assert torch.allclose(layer.lora_B["default"].weight.grad, expected_grad_lora_B, atol=1e-6, rtol=1e-5)
+
+
+@pytest.mark.regression
+def test_velora_cifar10_single_batch_learning():
```
BenjaminBossan: Can be removed.

tests/test_velora.py
```diff
@@ -0,0 +1,330 @@
...
+        logits = model(x)
+        loss = criterion(logits, y)
+        loss.backward()
+        optim.step()
+        losses.append(loss.item())
+
+    assert losses[-1] < losses[0] * 0.85
+
+
+def test_velora_save_pretrained_keeps_adapter_weights_only():
```
BenjaminBossan: This shouldn't need an extra test, as it's basic PEFT functionality, unless you have reasons to believe that VeLoRA is special here.

BenjaminBossan (changes requested):
Thanks for refactoring the PR to use LoRA variants. I think this is the right way and it should help users test VeLoRA, as they can add it easily to their existing LoRA code.

Regarding the implementation, I think it still contains unnecessary parts, which I have marked. Unless I'm missing something, the end result should be even simpler.

Let's also start working on completing the PR. For this, let's work on:

1. Docs: Add an entry to `docs/source/developer_guide/lora.md` in the `Variants` section.
2. Tests: Add VeLoRA to the test matrix (check my comment)
3. Benchmarking: We recently added gradient checkpointing to the [MetaMathQA benchmark](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA). It only requires setting `use_gc: true` in the experiment settings. I think it would be important to compare: a) normal LoRA b) LoRA with GC, and c) VeLoRA. We should see similar performance (train loss, test accuracy) but VeLoRA should require less memory than normal LoRA while being faster than GC, right?

src/peft/tuners/lora/layer.py
roymiles: I've removed the changes in lora/layer.py and added them to VeloraLinearVariant.init:
https://github.com/roymiles/peft/blob/05462927d7f765d314cce92b74716d322cd7d40c/src/peft/tuners/lora/variants.py#L797

src/peft/tuners/lora/layer.py
roymiles: Done!

src/peft/tuners/lora/layer.py
roymiles: I haven't tried with more than one active adapter, but it should work. It was a bit of an unnecessary check, so I have removed it now.
Building upon the previous point, the bits using `custom_base_variant_adapter` were just remnants from the original commit which also worked for training the base model (i.e. with no adapters). Since we are going for only a LoRA adapter variant, I have pretty much removed all of this logic:
https://github.com/huggingface/peft/pull/3159/changes/05462927d7f765d314cce92b74716d322cd7d40c#diff-24a141c266b7b714ae8fcc470f31bc283f7b0f5a671bbf6d5f092741fc374104L980

src/peft/tuners/lora/variants.py
roymiles: ah yea this doesn't seem to be needed. I have removed it.

src/peft/tuners/lora/variants.py
roymiles: so the reason it is assigned as a torch.Tensor here is because it is initialised as a buffer since and we don't want to unintentionally re-initialise embeds when re-loading a VeLoRA layer.

src/peft/tuners/lora/layer.py
roymiles: Yes, sorry, this was old remnants from the initial commit for being compatible with just normal training (no LoRA). Since we are just doing a LoRA variant now this is all removed.

src/peft/tuners/lora/layer.py
roymiles: This is removed now.

src/peft/tuners/lora/config.py
roymiles: I have changed it to just use optional velora_config input.

src/peft/tuners/lora/variants.py
roymiles: made these changes!

src/peft/tuners/velora/__init__.py
roymiles: ok all of peft/tuners/velora is now removed

src/peft/tuners/__init__.py
roymiles: ok should be all reverted back now

tests/test_velora.py
roymiles: ok have moved them now

tests/test_velora.py
roymiles: I ran the tests again after putting VeLoRA in the test matrix and they all pass:
86 passed, 2 skipped, 12177 deselected, 88 warnings in 113.38s (0:01:53) =

tests/test_velora.py
roymiles: ok, have removed this

tests/test_velora.py
roymiles: ok, have removed now

tests/test_velora.py
roymiles: ok, have removed this now

tests/test_velora.py
roymiles: I have added a comment at the top explaining what the test is doing:
`Check if the embed is being correctly initialised as the batch average of the input (in this case the input to the network)`

tests/test_velora.py
roymiles: I have added a few comments here and I hope they are ok! If there any part that are not clear, I can go back and improve.

roymiles:
Thanks @BenjaminBossan the comments were very thorough and I hope I have managed to address most of them well. 


The only thing I haven't had time to look at yet is the benchmarking. You are correct and we did have some experiments comparing to GC in the main paper. VeLoRA enables lower memory by approximating the gradients, which we find does not impact performance and has a minimal compute overhead (in contrast to GC). That being said GC can enable a much lower memory requirements (not achievable by VeLoRA), but of course with much more compute.

I will look through the MetaMathQA benchmark tomorrow and early next week. Would this be needed for the PR to peft? in either case I will definitely look into getting these results.

BenjaminBossan:
Thanks for the updates.

> I will look through the MetaMathQA benchmark tomorrow and early next week. Would this be needed for the PR to peft? in either case I will definitely look into getting these results.

It isn't strictly necessary, but I think it would be very good to run the test as a check that everything works as expected. The code as it is here now deviates quite a bit from what was in the original repo, so having a realistic example that can demonstrate the expected effect would be good to have. I believe it should not require much effort to test, if you have any issues, just let me know.

Once we confirm that VeLoRA works as expected, I'll do another thorough review.

roymiles:
[eval_job_output_63536.txt](https://github.com/user-attachments/files/27054880/eval_job_output_63536.txt)

I ran the MetaMathQA benchmark with the three different variants (LoRA, LoRA+GC and LoRA+VeLoRA). I have attached the full log file and below is the summary. I do believe the performance of VeLoRA (in terms of accuracy) could definitely be improved with better parameter choices (groups, layer choice etc), but the memory is indeed reduced with no increase in training time. 

I can sweep some hyperparameters for VeLoRA if you think this would be worthwhile.

```
=== Building summary ===
Variant comparison:
  - LoRA: status=success, train_loss=0.6070, test_acc=47.76%, max_mem=20.77 GiB, tokens_per_sec=3597.4
  - LoRA + GC: status=success, train_loss=0.6069, test_acc=48.29%, max_mem=12.55 GiB, tokens_per_sec=2614.4
  - VeLoRA: status=success, train_loss=0.6328, test_acc=44.66%, max_mem=18.91 GiB, tokens_per_sec=3514.3
Checks:
  - VeLoRA lower max memory than LoRA: True
  - VeLoRA faster than LoRA + GC: True
```

src/peft/tuners/lora/config.py
```diff
@@ -51,6 +51,50 @@ class LoraRuntimeConfig:
...
+    Args:
+        velora_num_groups (`int`):
+            Number of feature groups used by VeLoRA to split the input activation depth before compression.
+        velora_scale (`float`):
+            Scale applied to the reconstructed activations in the VeLoRA backward pass.
+        velora_init_type (`str`):
+            Projection initialization strategy for VeLoRA. Supported values are `'batch_average_once'` and `'random'`.
+    """
+
+    velora_num_groups: int = field(
```
BenjaminBossan: Let's remove the `verlora_` prefix from all those arguments. Since we're inside of `VeloraConfig`, it's already clear what they refer to.

src/peft/tuners/lora/config.py
```diff
@@ -51,6 +51,50 @@ class LoraRuntimeConfig:
...
 
 
+@dataclass
+class VeloraConfig:
+    """
+    This is the sub-configuration class to store the configuration for VeLoRA.
+
+    Args:
+        velora_num_groups (`int`):
+            Number of feature groups used by VeLoRA to split the input activation depth before compression.
```
BenjaminBossan: It would be nice to mention the default and what changing this argument does ("Increase this parameter to ..."). Same below in the help.

src/peft/tuners/lora/config.py
```diff
@@ -51,6 +51,50 @@ class LoraRuntimeConfig:
...
+        default=32,
+        metadata={
+            "help": "Number of feature groups used by VeLoRA to split the input activation depth before compression."
+        },
+    )
+    velora_scale: float = field(
+        default=1.0,
+        metadata={"help": "Scale applied to the reconstructed activations in the VeLoRA backward pass."},
+    )
+    velora_init_type: str = field(
```
BenjaminBossan: How about `Literal["batch_average_once", "random"]` as type annotation?

src/peft/tuners/lora/config.py
```diff
@@ -815,6 +835,44 @@ def to_dict(self):
...
+    def velora_scale(self) -> float:
+        if self.velora_config is None:
+            return 1.0
+        return self.velora_config.velora_scale
+
+    @property
+    def velora_init_type(self) -> str:
+        if self.velora_config is None:
+            return "batch_average_once"
+        return self.velora_config.velora_init_type
```
BenjaminBossan: All of this can be removed. Instead of using `lora_config.velora_init_type`, use `lora_config.velora_config.init_type` etc.

docs/source/developer_guides/lora.md
```diff
@@ -234,6 +234,31 @@ Caching can thus make inference with DoRA significantly faster but it also requi
...
+    ),
+)
+```
+
+`velora_num_groups` controls how the input activation depth is split before compression, `velora_scale` rescales the reconstructed activations during the backward pass, and `velora_init_type` chooses how the projection is initialized. Use `batch_average_once` to initialize the projection from the first training batch or `random` to initialize it immediately from a random normalized vector.
+
+#### Caveats
+
+- VeLoRA is currently supported on standard LoRA linear layers only.
+- VeLoRA does not support embedding or convolutional LoRA layers.
```
BenjaminBossan: Redundant with the previous bullet point.

docs/source/developer_guides/lora.md
```diff
@@ -234,6 +234,31 @@ Caching can thus make inference with DoRA significantly faster but it also requi
...
+    velora_config=VeloraConfig(
+        velora_num_groups=32,
+        velora_scale=1.0,
+        velora_init_type="batch_average_once",
+    ),
+)
+```
+
+`velora_num_groups` controls how the input activation depth is split before compression, `velora_scale` rescales the reconstructed activations during the backward pass, and `velora_init_type` chooses how the projection is initialized. Use `batch_average_once` to initialize the projection from the first training batch or `random` to initialize it immediately from a random normalized vector.
+
```
BenjaminBossan, line 265: I think it would be nice to mention how it compares with gradient checkpointing. Perhaps you could even include a small table with the benchmarking results.

src/peft/tuners/lora/config.py
```diff
@@ -51,6 +51,50 @@ class LoraRuntimeConfig:
     )
 
 
+@dataclass
+class VeloraConfig:
+    """
+    This is the sub-configuration class to store the configuration for VeLoRA.
```
BenjaminBossan: Let's link the paper here for convenience.

src/peft/tuners/lora/config.py
```diff
@@ -811,6 +910,24 @@ def __post_init__(self):
         if self.use_dora and self.megatron_config:
             raise ValueError("DoRA does not support megatron_core, please set `use_dora=False`.")
 
+        if self.use_velora:
```
BenjaminBossan: No need to check this here, please remove.

src/peft/tuners/lora/layer.py
```diff
@@ -101,7 +101,12 @@ class LoraLayer(BaseTunerLayer):
     # All names of layers that may contain (trainable) adapter weights
     adapter_layer_names: tuple[str, ...] = ("lora_A", "lora_B", "lora_embedding_A", "lora_embedding_B")
     # All names of other parameters that may contain adapter-related parameters
-    other_param_names: tuple[str, ...] = ("r", "lora_alpha", "scaling", "lora_dropout")
+    other_param_names: tuple[str, ...] = (
```
BenjaminBossan: Please revert

src/peft/tuners/lora/variants.py
BenjaminBossan: Can `lora_velora_initialized` not be a normal dict?

src/peft/tuners/lora/variants.py
```diff
@@ -771,6 +773,125 @@ def get_alora_offsets_for_generate(model: nn.module, *args, **kwargs):
...
+        x: torch.Tensor,
+        result: torch.Tensor,
+        **kwargs,
+    ) -> torch.Tensor:
+        lora_A = module.lora_A[active_adapter]
+        lora_B = module.lora_B[active_adapter]
+        dropout = module.lora_dropout[active_adapter]
+        scaling = module.scaling[active_adapter]
+
+        x = module._cast_input_dtype(x, lora_A.weight.dtype)
```
BenjaminBossan: We already call this in `lora.Linear.forward` before calling the LoRA variant, so this should redundant.

src/peft/tuners/lora/velora.py
```diff
@@ -0,0 +1,111 @@
...
+def _normalize_projection(embed: torch.Tensor) -> torch.Tensor:
+    embed = embed.float()
+    norm = torch.linalg.vector_norm(embed)
+    if norm == 0:
+        embed = torch.ones_like(embed)
+        norm = torch.linalg.vector_norm(embed)
+    return embed / norm
+
+
+def _compute_dtype_for_backward(reference: torch.Tensor) -> torch.dtype:
```
BenjaminBossan: Could you please add a comment why this is done?

tests/test_custom_models.py
```diff
@@ -105,6 +106,21 @@
         LoraConfig,
         {"target_modules": "lin1", "use_dora": True, "lora_alpha": 32},
     ),
+    (
+        "Vanilla MLP 10 LoRA with VeLoRA",
+        "MLP",
+        LoraConfig,
+        {"target_modules": ["lin0"], "velora_config": VeloraConfig(velora_num_groups=2, velora_init_type="random")},
+    ),
+    (
```
BenjaminBossan, line 115: It would be good to have one setting with `batch_average_once` initialization.

tests/test_velora.py
```diff
@@ -0,0 +1,259 @@
...
+        ),
+    }
+    if lora_alpha is not None:
+        kwargs["lora_alpha"] = lora_alpha
+    if init_lora_weights is not True:
+        kwargs["init_lora_weights"] = init_lora_weights
+    return LoraConfig(**kwargs)
+
+
+def test_top_level_velora_config_alias_matches_lora_module_config():
```
BenjaminBossan: Why is this test needed? Let's remove it.

tests/test_velora.py
```diff
@@ -0,0 +1,259 @@
...
+            num_groups=8,
+        ),
+    )
+    layer = model.base_model.model.lin0
+
+    x0 = torch.randn(2, 16)
+    model.train()
+    _ = model(x0)
+
+    expected = _normalize_projection(x0.reshape(-1, 8, 2).reshape(-1, 2).mean(dim=0)).to(
```
BenjaminBossan: Hmm, as we use the `_normalize_projection` function to compute the expected value, isn't the test result trivially correct?

tests/test_velora.py
```diff
@@ -0,0 +1,259 @@
...
+        layer.lora_B["default"].weight.copy_(torch.arange(40, dtype=torch.float32).reshape(10, 4) / 50)
+
+    x = torch.randn(2, 3, 16, requires_grad=True)
+    grad_output = torch.randn(2, 3, 10)
+
+    model.train()
+    output = model(x)
+    output.backward(grad_output)
+
+    # Compress the input activations as per equation (1) in the original paper
```
BenjaminBossan, line 322: Same argument here: If we use the same functions to compute the result and the expected value, the test is almost trivially correct.

src/peft/tuners/lora/config.py
```diff
@@ -51,6 +51,50 @@ class LoraRuntimeConfig:
...
+    velora_init_type: str = field(
+        default="batch_average_once",
+        metadata={
+            "help": "Projection initialization strategy for VeLoRA. Supported values are "
+            "`'batch_average_once'` and `'random'`."
+        },
+    )
+
+    def __post_init__(self):
+        if self.velora_num_groups <= 0:
```
BenjaminBossan: Let's add tests for these checks in `test_initializaiton.py`

src/peft/tuners/lora/velora.py
```diff
@@ -0,0 +1,111 @@
...
+
+from __future__ import annotations
+
+from typing import Optional
+
+import torch
+import torch.nn.functional as F
+
+
+def _reshape_to_grouped_subtokens(x: torch.Tensor, num_groups: int) -> torch.Tensor:
```
BenjaminBossan, line 27: I wonder if it would not be possible to have `in_features % num_groups != 0` by explicitly handling the remainder here. It would make usage easier if the user doesn't have to restrict `num_groups` to divisible sizes.

BenjaminBossan (commented):
Thanks for running the experiment and confirming that we can observe the expected effects. Let's add the experiment setting to the PR (not the results, we'll generate those on our machines). Regarding the experimental settings:

1. If you find a setting that improves the test accuracy, that would be great.
2. The speed is almost on par with normal LoRA but the memory savings are relatively small compared to GC. I think for most users, a setting that trades speed for more memory efficiency would be more interesting.

Regarding the implementation, it looks much cleaner now. There were still a few places that need improvement, but overall not much is missing.

roymiles:
Thanks for comments! I am just re-running the benchmark with a few different parameters so we can use better defaults for the implementation. Once this is done I can push my changes addressing your comments.

BenjaminBossan:
@roymiles did you have time to run the benchmarks yet?

roymiles:
Yes, so I did look into this and there are a few rules of thumb that might be worth mentioning. 
VeLoRA benefits when more layers are trainable (e.g. the down projections, MLP layers etc). The results reported earlier were only with a few attention layers being trained.

Using a higher number of groups is generally more effective (> 32) and a lower scale (~0.2) is a good default. I also added a batch average for the projections that is not just initialised and frozen on first forward pass. This slows down training a tiny bit, but allows retaining all of the accuracy as LoRA.

Below are the benchmark results and it seems to be a much more useful intermediate option between LoRA and LoRA+GC:

```
Test accuracy: 0.549
Saved VeLoRA result to /.../20260506T160320Z_best_comparison/velora.json
Variant comparison:
  - VeLoRA: status=success, train_loss=0.5427, test_acc=54.89%, max_mem=19.94 GiB, tokens_per_sec=2057.6

Test accuracy: 0.545
Saved LoRA result to /.../20260506T160320Z_best_comparison/lora.json
Variant comparison:
  - LoRA: status=success, train_loss=0.5427, test_acc=54.51%, max_mem=27.69 GiB, tokens_per_sec=2366.2


Test accuracy: 0.558
Saved LoRA + GC result to /.../20260506T160320Z_best_comparison/lora_gc.json
Variant comparison:
  - LoRA + GC: status=success, train_loss=0.5426, test_acc=55.80%, max_mem=13.17 GiB, tokens_per_sec=1671.8
```

I have made the changes from your comments already and will commit once I have checked through them.

tests/test_velora.py
roymiles: Yea, I see your point. I will remove these.

src/peft/tuners/lora/velora.py
roymiles: Yea this makes sense. It is now padding to ceil(in_features / num_groups) * num_groups, and reconstruction slices back to the original feature width. If divisible, it will be the original implementation.

tests/test_velora.py
roymiles: Ok, these tests are removed now.

tests/test_velora.py
roymiles: Ok, have removed these.

src/peft/tuners/lora/velora.py
roymiles: Hmm, this must be from my old code or some extra safe checks. I don't think it is needed and it is removed now.

src/peft/tuners/lora/config.py
roymiles: Ok, I have added some comments explaining how changing the parameters effects memory and reconstruction error (which typically can degrade generalisation performance).

roymiles:
I ran the velora tests and they still all pass. The final comment about removing the `self.use_velora` I need to look at tomorrow and check. I think if the if check is removed it can raise some incompatibility errors, but I will get back to you on this.

src/peft/tuners/lora/config.py
roymiles: Ah, yes. Ignore my recent comment on this. I understand why it should be removed now and its gone in the recent second commit.

roymiles:
Ok, I believe I have addressed all your comments now and I redid the tests again just to be sure.

src/peft/tuners/lora/config.py
roymiles: Ok, these are added now.

docs/source/developer_guides/lora.md
```diff
@@ -234,6 +234,39 @@ Caching can thus make inference with DoRA significantly faster but it also requi
...
+)
+```
+
+`num_groups` controls how the input activation depth is split before compression. If the activation depth is not evenly divisible by `num_groups`, VeLoRA pads the grouped representation internally and removes the padding after reconstruction. `scale` rescales the reconstructed activations during the backward pass, and `init_type` chooses how the projection is initialized.
+
+Use `batch_average_once` to initialize the projection from the first training batch, `batch_average` to update it from every training forward pass, or `random` to initialize it immediately from a random normalized vector.
+
+Below are some results with the [MetaMathQA benchmark](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA). 
+
+| Variant | Test Accuracy | Max Memory (GiB) | Tokens/sec |
```
BenjaminBossan: Instead of test accuracy, let's report train loss here. Accuracy can vary a bit because it's based on generations, train loss is the more deterministic variable and shows that the 3 approaches give equivalent results.

src/peft/tuners/lora/velora.py
BenjaminBossan: Nice. Is there a test that covers this case?

BenjaminBossan (commented):
Thanks for the updates, LMK when the PR is ready to review.

src/peft/tuners/lora/variants.py
roymiles: I was unsure about if its value stays after saving/loading a velora model, but maybe it doesn't functionally matter much either way. I have made it a normal dict now in the most recent push.

docs/source/developer_guides/lora.md
roymiles: Ok that makes sense! I have updated this now.

roymiles:
I have checked and addressed all of your comments now, hope it is all ok!

BenjaminBossan:
@roymiles Could you please run `make style`?

docs/source/developer_guides/lora.md
BenjaminBossan: Still open.

src/peft/tuners/lora/config.py
```diff
@@ -51,6 +51,81 @@ class LoraRuntimeConfig:
     )
 
 
+@dataclass
+class VeloraConfig:
+    """
+    This is the sub-configuration class to store the configuration for VeLoRA. Original paper can be found <a
```
BenjaminBossan: Let's use markdown syntax for links

src/peft/tuners/lora/config.py
```diff
@@ -793,6 +884,23 @@ def __post_init__(self):
...
+            if "velora_scale" in self.velora_config:
+                self.velora_config["scale"] = self.velora_config.pop("velora_scale")
+            if "velora_init_type" in self.velora_config:
+                self.velora_config["init_type"] = self.velora_config.pop("velora_init_type")
+            if "velora_target_modules" in self.velora_config:
+                self.velora_config["target_modules"] = self.velora_config.pop("velora_target_modules")
+            if "velora_layers_to_transform" in self.velora_config:
+                self.velora_config["layers_to_transform"] = self.velora_config.pop("velora_layers_to_transform")
+            if "velora_layers_pattern" in self.velora_config:
+                self.velora_config["layers_pattern"] = self.velora_config.pop("velora_layers_pattern")
```
BenjaminBossan: This can all be removed, right?

src/peft/tuners/lora/config.py
```diff
@@ -779,6 +866,10 @@ def to_dict(self):
         rv.pop("runtime_config")
         return rv
 
+    @property
+    def use_velora(self) -> bool:
```
BenjaminBossan: Let's remove this. Change the checks to `if self.velora_config is not None` etc.

src/peft/tuners/lora/layer.py
```diff
@@ -47,6 +52,16 @@
 VARIANT_KWARG_KEYS = ["alora_offsets"]
 
 
+def _use_velora_for_target(config: LoraConfig, target_name: str) -> bool:
```

src/peft/tuners/lora/layer.py
```diff
@@ -150,6 +165,9 @@ def resolve_lora_variant(self, *, config: LoraConfig, **kwargs) -> Optional[Lora
         """
         return None
 
+    def _use_velora_for_target(self, config: LoraConfig, target_name: str) -> bool:
```

src/peft/tuners/lora/layer.py
```diff
@@ -793,6 +811,11 @@ def __init__(
         self.is_target_conv_1d_layer = is_target_conv_1d_layer
 
     def resolve_lora_variant(self, config: LoraConfig, **kwargs) -> Optional[LoraVariant]:
+        if self._use_velora_for_target(config, kwargs.get("target_name", "")):
```
BenjaminBossan: So IIUC, the idea here is to allow VeLoRA to be applied to only a subset of layers that are LoRA layers. I wonder if this feature is needed or if we cannot just target all LoRA layers and thus making the implementation simpler. Do you have examples in mind where this would be a bad idea?

If this option is removed, `VeloraConfig` would no longer need its own `target_modules` etc. We could also remove the `target_name` being passed to `resolve_lora_variant`.

tests/test_custom_models.py
```diff
@@ -105,6 +106,30 @@
...
+        "Vanilla MLP 11 LoRA with VeLoRA",
+        "MLP",
+        LoraConfig,
+        {
+            "target_modules": ["lin0", "lin1"],
+            "velora_config": VeloraConfig(num_groups=2, init_type="random"),
+        },
+    ),
+    *[
```
BenjaminBossan: Let's not do a list comprehension here but explicitly list each setting to test. Let's do 3 settings in total, you could e.g. test the `batch_average` in the 2nd example.

tests/test_initialization.py
```diff
@@ -1833,6 +1834,40 @@ def test_vera_add_second_adapter_with_higher_rank(self):
...
+                {"target_modules": "linear", "layers_pattern": "layers"},
+                "`layers_pattern` cannot be used when `target_modules` is a str.",
+                id="target-modules-str-with-layers-pattern",
+            ),
+            pytest.param(
+                {"layers_pattern": "layers"},
+                "When `layers_pattern` is specified, `layers_to_transform` must also be specified. ",
+                id="layers-pattern-without-layers-to-transform",
+            ),
+        ],
```
BenjaminBossan: This could also be removed if we decide to always target all LoRA layers with VeLoRA.

src/peft/tuners/lora/layer.py
```diff
@@ -47,6 +52,16 @@
...
 
 
+def _use_velora_for_target(config: LoraConfig, target_name: str) -> bool:
+    if not config.use_velora:
+        return False
+    velora_config = config.velora_config
+    assert velora_config is not None
+    if velora_config.target_modules is None:
+        return True
+    return bool(check_target_module_exists(velora_config, target_name))
```
BenjaminBossan: This looks brittle because we usually expect a `PeftConfig` to be passed to `check_target_module_exists`, but `VeloraConfig` is not a full `PeftConfig`. Perhaps it works as is, but changes to `check_target_module_exists` could break this or lead to silent failures, as casting to `bool` is very broad.

BenjaminBossan (changes requested):
Thanks for your continued work. Overall, not much is missing, but I still found a few things. Most notably, I wonder if we really need `VeloraConfig` to have its own `target_modules` etc. As is, that implementation is very brittle and I think unless there is a very good reason to keep it, I'd rather remove it completely.

src/peft/tuners/lora/layer.py
roymiles: I agree it would be simpler by enforcing VeLoRA to be applied to every LoRA layer. I was playing with this using the benchmark earlier and VeLoRA is definitely more effective on some layers (like down projections where input activations are larger) than others. It is probably true that the gain in performance by tuning specific target layers is not worth complicating the implementation. I will remove this option now.

src/peft/tuners/lora/config.py
roymiles: Ah yea, this was just backward compatibility stuff, removed now

src/peft/tuners/lora/layer.py
roymiles: Ok have removed both of these

tests/test_custom_models.py
roymiles: ok, have made them tests without list comprehension

tests/test_initialization.py
roymiles: ok, have removed

docs/source/developer_guides/lora.md
roymiles: ah oops, yes I have removed it now

src/peft/tuners/lora/layer.py
roymiles: This entire function has been removed now, and instead the check with `config.velora_config is not None` is used to resolve if VeLoRA is the variant.

roymiles:
I think I have addressed the comments and the tests on my end still work.

roymiles:
I noticed all the links to "Caveats" go to Dora's caveats (not just VeLoRAs). I have kept the subheading name to be consistent, but just thought I'd share this. This seems to be the case for any subheading with the same name. [here](https://moon-ci-docs.huggingface.co/docs/peft/pr_3159/en/developer_guides/lora#caveats)

BenjaminBossan:
@roymiles Thanks for simplifying the code. Could you please run `make style` again :) This time, it's the `doc-builder` that's complaining.

roymiles:
@BenjaminBossan  ah woops forgot this, it should be good now.

BenjaminBossan (approved):
Perfect, thanks. Everything looks good now.

roymiles:
That's great and of course thanks for your patience with me! I'm really happy to see it as a part of hf peft.

docs/source/developer_guides/lora.md
```diff
@@ -234,6 +234,38 @@ Caching can thus make inference with DoRA significantly faster but it also requi
 - DoRA introduces a bigger overhead than pure LoRA, so it is recommended to merge weights for inference, see [`LoraModel.merge_and_unload`].
 - DoRA should work with weights quantized with bitsandbytes ("QDoRA"). However, issues have been reported when using QDoRA with DeepSpeed Zero2.
 
+### VeLoRA
+
+[VeLoRA](https://huggingface.co/papers/2405.14174) is a LoRA variant that reduces training memory by compressing the activations saved for the LoRA in the forward pass and then reconstructing them in the backwards pass to implement the update rules. In PEFT, VeLoRA is configured as a LoRA variant through the `velora_config` argument on [`LoraConfig`].
```
BenjaminBossan: Last minute correct, the paper link is wrong here:

```suggestion
[VeLoRA](https://huggingface.co/papers/2405.17991) is a LoRA variant that reduces training memory by compressing the activations saved for the LoRA in the forward pass and then reconstructing them in the backwards pass to implement the update rules. In PEFT, VeLoRA is configured as a LoRA variant through the `velora_config` argument on [`LoraConfig`].
```

BenjaminBossan (changes requested):
Sorry, caught one last error.

BenjaminBossan (approved):
Thanks, now we're good to go.

PR #3161, opened 2026-04-14
fix: guard UPCAST_DTYPES lookup with hasattr to support older PyTorch

### LINKED ISSUE #3158: AttributeError: module 'torch' has no attribute 'float8_e8m0fnu' when casting adapter dtype on PyTorch 2.5.1
### System Info

When initializing a PEFT model using get_peft_model with bitsandbytes 4-bit quantization on PyTorch 2.5.1, the script crashes during the adapter dtype casting. It seems the code blindly assumes the existence of the new float8_e8m0fnu dtype without a fallback or hasattr check for slightly older/stable PyTorch versions.

### Environment:
- PyTorch: 2.5.1+cu121
- PEFT: (Latest from uv / PyPI)
- Transformers / Unsloth: Latest
- Python: 3.10

### Traceback:
```
  File "peft/tuners/tuners_utils.py", line 2170, in cast_adapter_dtype
    torch_dtype = getattr(torch, name)
  File "torch/__init__.py", line 2562, in __getattr__
    raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
AttributeError: module 'torch' has no attribute 'float8_e8m0fnu'
```

### Temporary Workaround / Proposed Fix:
I was able to bypass this by manually mocking the attribute before initializing the model:
```
import torch
if not hasattr(torch, "float8_e8m0fnu"):
    setattr(torch, "float8_e8m0fnu", torch.float32)
```

**A simple hasattr check inside peft/tuners/tuners_utils.py around line 2170 would permanently fix this for users on stable PyTorch branches.**

### Who can help?

_No response_

### Reproduction

Initialize a 4-bit quantized model and attempt to apply a LoRA config on PyTorch 2.5.1.

```
#!/bin/bash

rm -rf .venv
uv venv --python 3.10
source .venv/bin/activate

uv pip install torch==2.5.1 torchvision==0.20.1 torchaudio==2.5.1 --index-url https://download.pytorch.org/whl/cu121
uv pip install "unsloth[colab-new] @ git+https://github.com/unslothai/unsloth.git" transformers trl datasets accelerate xformers peft bitsandbytes torch==2.5.1 torchvision==0.20.1 setuptools
```

### Expected behavior

It should not throw.

Anai-Guo:
## Problem

Fixes #3158.

`cast_adapter_dtype` in `tuners_utils.py` iterates over `UPCAST_DTYPES` (which includes `float8_e8m0fnu`) and calls `getattr(torch, name)` unconditionally:

```python
for name in UPCAST_DTYPES:
    torch_dtype = getattr(torch, name)      # <-- crashes if dtype not in this torch version
    dtypes_to_convert_to_fp32.add(torch_dtype)
```

The comment directly above this code says *"defensively only include dtypes that are actually found, as this could depend on torch version and platform"*, but the code doesn't implement that defence.  On PyTorch 2.5.1 (stable), `torch.float8_e8m0fnu` does not exist, so any `get_peft_model` call with a quantized model crashes with:

```
AttributeError: module 'torch' has no attribute 'float8_e8m0fnu'
```

## Fix

Use `getattr(torch, name, None)` with a `None` sentinel so unknown dtypes are silently skipped, matching the stated intent of the comment:

```python
for name in UPCAST_DTYPES:
    if (torch_dtype := getattr(torch, name, None)) is not None:
        dtypes_to_convert_to_fp32.add(torch_dtype)
```

This is a one-line behaviour change; the set only grows when the dtype actually exists in the running PyTorch installation.

## Testing

- Verified the diff is minimal: only the `getattr` / guard line changes.
- Reproducer from the issue: initialise a 4-bit quantised model with `get_peft_model` on PyTorch 2.5.1 — no longer crashes.

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -2167,8 +2167,8 @@ def cast_adapter_dtype(model: nn.Module, adapter_name: str, autocast_adapter_dty
     # Upcast lower precision floats like float8_e4m3fn; defensively only include dtypes that are actually found, as this
     # could depend on torch version and platform
     for name in UPCAST_DTYPES:
-        torch_dtype = getattr(torch, name)
-        dtypes_to_convert_to_fp32.add(torch_dtype)
+        if (torch_dtype := getattr(torch, name, None)) is not None:
+            dtypes_to_convert_to_fp32.add(torch_dtype)
 
     for module in model.modules():
         if not isinstance(module, BaseTunerLayer):


githubnemo (approved):
Thanks, LGTM!

BenjaminBossan:
@Anai-Guo The patch release v0.19.1 with this fix is out.

PR #3162, opened 2026-04-15
chore: bump doc-builder SHA for PR upload workflow

rtrompier:
Switch the PR doc upload flow from the legacy dataset push to the new HF bucket.


BenjaminBossan (approved):
Thanks for updating. Note that we have Dependabot update all action hashes once per month (e.g. #3125), so think this type of manual updating shouldn't be needed (except if it's urgent).

Failing CI is unrelated.

PR #3163, opened 2026-04-15
MetaMathQA: upload checkpoint to bucket after training

githubnemo:
To do post-training analysis on the checkpoints (e.g., when experimenting on different post-training metrics) it is useful to have the checkpoints available. Since we have buckets now, let's use those.

For the Makefile you need to set the `UPLOAD_BUCKET` environment variable to activate this feature for the whole run.


BenjaminBossan (approved):
Nice, thanks.

PR #3164, opened 2026-04-15
Fix C3ALinear.forward chaining active adapters through x instead of summing

Chessing234:
## Bug

When more than one C3A adapter is active simultaneously, \`C3ALinear.forward\` feeds the **previous adapter's output** into the next adapter instead of feeding every adapter the original input. As a result, the unmerged forward pass produces a different value than the merged model for any multi-adapter configuration, and later adapters are spuriously applied on top of earlier adapters' low-rank outputs.

## Root cause

In the active branch of \`forward\` (in \`src/peft/tuners/c3a/layer.py\`):

\`\`\`python
x = x.to(torch.float32)
    if active_adapter not in self.c3a_kernel.keys():
        continue
    c3a_kernel = self.c3a_kernel[active_adapter].to(torch.float32)
    x = BlockCircularConvolution.apply(x, c3a_kernel) / x.size(-1)
    result += x.to(result.dtype)
\`\`\`

The loop rebinds \`x\` to the current adapter's \`BlockCircularConvolution(x, kernel_A) / x.size(-1)\`. On the next iteration, that already-transformed \`x\` is passed into the next \`BlockCircularConvolution\`. For \`N\` active adapters the output becomes

\`\`\`
W0·x + A1(x) + A2(A1(x)) + A3(A2(A1(x))) + ...
\`\`\`

instead of the linear sum the merge path implements:

\`\`\`python
# merge():
base_layer.weight.data = base_layer.weight.data + delta_weight  # one adapter at a time
\`\`\`

which yields \`W0·x + A1(x) + A2(x) + … + An(x)\`. The unmerged and merged models therefore disagree whenever \`len(active_adapters) > 1\`, and the unmerged forward is incorrect on its own terms (each adapter should see the shared input, not the previous adapter's output).

## Fix

Bind the per-adapter contribution to a local \`delta\` instead of reassigning \`x\`, so every adapter sees the original \`x\`:

\`\`\`diff
-                x = BlockCircularConvolution.apply(x, c3a_kernel) / x.size(-1)
-                result += x.to(result.dtype)
+                delta = BlockCircularConvolution.apply(x, c3a_kernel) / x.size(-1)
+                result += delta.to(result.dtype)
\`\`\`

When at most one C3A adapter is active (the path covered by existing tests) behaviour is identical; for multiple active adapters it now matches the merged model.

### DIFF
--- a/src/peft/tuners/c3a/layer.py
+++ b/src/peft/tuners/c3a/layer.py
@@ -191,8 +191,8 @@ class C3ALinear(nn.Module, C3ALayer):
                 if active_adapter not in self.c3a_kernel.keys():
                     continue
                 c3a_kernel = self.c3a_kernel[active_adapter].to(torch.float32)
-                x = BlockCircularConvolution.apply(x, c3a_kernel) / x.size(-1)
-                result += x.to(result.dtype)
+                delta = BlockCircularConvolution.apply(x, c3a_kernel) / x.size(-1)
+                result += delta.to(result.dtype)
 
         result = result.to(previous_dtype)
         return result


src/peft/tuners/c3a/layer.py
```diff
@@ -191,8 +191,8 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
                 if active_adapter not in self.c3a_kernel.keys():
                     continue
                 c3a_kernel = self.c3a_kernel[active_adapter].to(torch.float32)
-                x = BlockCircularConvolution.apply(x, c3a_kernel) / x.size(-1)
-                result += x.to(result.dtype)
+                delta = BlockCircularConvolution.apply(x, c3a_kernel) / x.size(-1)
```
afurm, line 194: The PR correctly identifies that rebinding `x` in the loop causes adapter outputs to chain instead of sum. One observation: the merge path applies each adapter sequentially to `base_layer.weight.data` and thus naturally produces a linear sum on the forward pass. Since there are existing merge-path tests, were any of those exercising multi-adapter `forward` paths that would have caught this divergence?

BenjaminBossan (changes requested):
Thanks for identifying this issue. I agree that it's a bug. Two comments:

First, could you please add unit tests for this? We already have tests for multiple adapters and checking if merging produces the same result. However, C3A is missing there. Please add it to the test matrix here:

https://github.com/huggingface/peft/blob/5b6e5e993944bddbb9fb12d5540f2d17eba39ad8/tests/test_custom_models.py#L1117

Regarding the proposed solution: I think it works but what we could do instead is to sum all the active `c3a_kernel`s and then send them through `BlockCircularConvolution.apply` instead of the other way round. This should get the same result and be more efficient. WDYT?

Also pinging @Phoveran

Chessing234:
Thanks for the feedback! I've updated the implementation to sum all the active `c3a_kernel`s first, and then apply `BlockCircularConvolution.apply` once. It's indeed much cleaner. I've also added the C3A configurations to the `MULTIPLE_ACTIVE_ADAPTERS_TEST_CASES` test matrix as requested. Let me know if everything looks good!

BenjaminBossan:
@Chessing234 You didn't push any changes to this PR.

PR #3165, opened 2026-04-15
FIX How to swapped in/out features for transposed

BenjaminBossan:
When targeting 3-dim MoE parameters with LoRA, we need to check which dimenion corresponds to `in_features` and which to `out_features`. For this, we use the is_transposed attribute but the logic was actually the wrong way round. This PR fixes the issue.

In general, this reversal does not really matter because it means that first, we initialize the weight in the wrong shape and then we apply the wrong shape again the forward step, so the two errors cancel out. However, when loading a model that uses the correct shape, the errors don't cancel.

To test this, run this test which uses a v4 Qwen checkpoint:

`pytest tests/test_integrations.py -k test_qwen_v4_lora_weight_conversion_peft_model_from_pretrained`

Now we might wonder why the error wasn't discovered even though there is a corresponding test. The reason is that in transformers, there is a monkey patch for PEFT to invert the order (`patch_moe_parameter_targeting`). If this monkey patch is applied before the Qwen test runs, the test passes. In rare circumstances, owing to tests running in parallel, the patch might come too late and the test fails. This is why it is flaky in our CI.

To see this in action, run the tests like so:

`pytest tests/test_integrations.py -k "test_mixtral_v4_lora_weight_conversion_transformers_load_adapter or test_qwen_v4_lora_weight_conversion_peft_model_from_pretrained"`

The first test applies the patch and that makes the second test pass.

What this PR doesn't do is to add extra logic to recover checkpoints that were created with the incorrect shape. This is because the PEFT v0.19.0 release is just from yesterday and we plan to do a patch release tomorrow. So the time window should be very small for incorrect checkpoints to be created. And even if they exist, they can be fixed without too much effort.

Regarding testing, this is not quite trivial to check, as it relies on test order. Therefore, I added a pytest fixture that makes each test run twice, once with the test applied and once without. After adding this fixture, the `test_qwen_v4_lora_weight_conversion_peft_model_from_pretrained` will reliably fail without the fix.

### DIFF
--- a/src/peft/tuners/lora/layer.py
+++ b/src/peft/tuners/lora/layer.py
@@ -2199,7 +2199,8 @@ class ParamWrapper(nn.Module, LoraLayer):
             raise ValueError(f"lora.{self.__class__.__name__} does not work with LoRA variants like DoRA.")
 
         # for some MoE layers, the order is (experts, out_features, in_features)
-        swap_in_out_features = getattr(self.get_base_layer(), "is_transposed", False)
+        is_transposed = getattr(self.get_base_layer(), "is_transposed", False)
+        swap_in_out_features = (self.get_param().ndim == 3) and not is_transposed
         if swap_in_out_features and not self._did_swap_in_out_features:
             self.in_features, self.out_features = self.out_features, self.in_features
             self._did_swap_in_out_features = True
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import inspect
 import re
 
 import packaging.version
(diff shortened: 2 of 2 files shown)


src/peft/tuners/lora/layer.py
```diff
@@ -2199,7 +2199,8 @@ def update_layer(
             raise ValueError(f"lora.{self.__class__.__name__} does not work with LoRA variants like DoRA.")
```
afurm, line 2200: The `is_transposed` flag on the base layer is used here to determine whether to swap in/out features. For some MoE layers the stored weight shape is `(experts, out_features, in_features)`, which means `ndim == 3` but `is_transposed == False`. The fix correctly adds the `ndim == 3` guard so that the swap logic only triggers for those specific MoE layers. One question: does this assume the LoRA params are already initialized at this point, or could `get_param()` return None on a freshly constructed layer that's being configured before any `copy_to_target_lora` call?

linitra24:
I tested this PR on my original repro from [#3112](https://github.com/huggingface/peft/pull/3112), and it fixes the issue on my side.

I used the same Qwen3-VL-MoE tiny model and the same LoRA generation script as before. The generated MoE LoRA shapes are now back to the working layout in my test, and the previous vLLM error no longer occurs.

Thanks for working on this!

BenjaminBossan:
Thanks for checking @linitra24. We'll do a patch release with this fix (0.19.1).

githubnemo (approved):
Thanks for fixing this!

BenjaminBossan:
@linitra24 The patch release v0.19.1 with this fix is out.

PR #3173, opened 2026-04-18
feat(tuners): suggest valid target modules when resolution fails (LoRA)

dentity007:
## Context

Per @BenjaminBossan's invitation in [#3154 (comment)](https://github.com/huggingface/peft/pull/3154#issuecomment-4243860185) and the scope reply in [#3154 (comment)](https://github.com/huggingface/peft/pull/3154#issuecomment-4250868072), this is a draft of the general-purpose target module suggester. Closes the loop on the narrower attempt in PR #3154. Complements #3136 (merged 2026-04-09), which set Gemma 4 default `target_modules`. This PR addresses the broader UX failure mode surfaced in #3129: when a user (or a default config) names modules that do not exist in the model, suggest valid ones before raising. Towards #3129.

## What this does

Adds a tuner-aware suggestion paragraph to the existing `ValueError` raised at the line-1008 raise site in `src/peft/tuners/tuners_utils.py` (commit `2859358`, the merge base), only at the case where the user named modules that do not exist. The other three raises in the same block (lines 989, 994, 1021) are out of scope for this PR.

Suggestions are produced by a per-tuner candidate collector registered via decorator, mirroring the existing `register_peft_method` pattern in `src/peft/utils/peft_types.py`. The output is compressed via `_find_minimal_target_modules` so suggestions look like `{'q_proj', 'v_proj'}` rather than fully-qualified paths.

## Scope (LoRA only in this PR)

- LoRA collector: implemented.
- IA3, LoKr, LoHa, OFT: registered with stub collectors and TODO markers. Five-line follow-up PRs to fill in.
- `target_parameters` (MoE) path: out of scope.
- IA3 feedforward / attention split: out of scope per [#3154 (comment)](https://github.com/huggingface/peft/pull/3154#issuecomment-4250868072).

## Behavior

Purely additive. When the suggester returns `None` (unregistered tuner, no candidates, empty minimal set), the original error message is unchanged. The pre-existing `tests/test_tuners_utils.py::TestExcludedModuleNames::test_no_modules_matched` still passes against its original substring match.

## Tests

New `TestTargetModuleSuggester` class in `tests/test_tuners_utils.py` covers ten cases:

1. Suggests valid LoRA targets when user names a non-existent module
2. Returns no suggestion when model has no LoRA-compatible candidates
3. Suggestion does not leak module names of types LoRA cannot adapt
4. No suggestion when modules are excluded (line-1021 combination raise on the MLP fixture; the strict line-989 "all excluded" raise is unreachable with this fixture because non-target modules populate `unmatched_modules` per `tuners_utils.py:906`)
5. No suggestion at the line-1021 raise (combination case)
6. No suggestion at the line-994 raise (no target_modules and no target_parameters passed)
7. Suggestion uses minimal suffix compression (hierarchical model with `q_proj` and `v_proj` at three layers; verifies the suggestion compresses to `{q_proj, v_proj}` rather than listing six fully-qualified names)
8. Calling the suggester with an unregistered tuner type returns `None`, no exception
9. The LoRA collector is side-effect-free (snapshot model state, run twice, compare)
10. All five known tuner types (LORA, IA3, LOHA, LOKR, OFT) have a registered collector

## Output format

First version. Open to redlines.

```
Target modules {'non_existent_module'} not found in the base model. Please check the target modules and try again.

Did you mean one of these? Valid `target_modules` candidates for LORA on this model: {'lin0', 'lin1'}.
```

Tests assert structural properties (suggestion contains a real candidate name, suggestion is omitted when no candidates exist), not exact wording. The wording can be redlined without test churn. The set-shaped output is rendered deterministically (sorted, run-to-run stable) so reviewers can rely on the example above matching real runs.

## Architecture notes

- `src/peft/tuners/target_suggester.py`: new module. `TUNER_TYPE_TO_CANDIDATE_COLLECTOR` dict, `register_candidate_collector` decorator, `suggest_target_modules` top-level entrypoint, `_format_suggestion` formatter. No module-top-level imports from `peft.tuners.tuners_utils` to avoid circular dependency.
- `src/peft/tuners/lora/candidate_collector.py`: new module. Parallel `isinstance` walk against the same supported-types tuple as `dispatch_default`. Does NOT call `dispatch_default` itself because `dispatch_default` has side effects on `config.fan_in_fan_out` (`lora/layer.py:2495-2500` and `2502-2508`).
- Hook in `tuners_utils.py`: six lines between current line 1007 and current line 1008. Local import of `suggest_target_modules` for circular-import safety.

## Open questions for review

1. Registry pattern OK, or would you prefer a `BaseTuner.get_candidates()` method?
2. Output format: keep "Did you mean one of these?" or change to "Valid candidates are:" or similar?
3. Cap the suggested set at top N, or always show all minimal targets?
4. You wrote "intercept exactly the right errors" (plural) in the invitation. This PR limits the hook to the line-1008 raise (the canonical "user named wrong modules" case). Should suggestions also fire at the line-1021 raise (combination case: some matched, some excluded)? Easy to extend if so.

### DIFF
--- a/src/peft/tuners/ia3/__init__.py
+++ b/src/peft/tuners/ia3/__init__.py
@@ -18,6 +18,7 @@ from peft.utils import register_peft_method
 from .config import IA3Config
 from .layer import Conv2d, Conv3d, IA3Layer, Linear
 from .model import IA3Model
+from . import candidate_collector  # noqa: F401  # registers IA3 collector stub via decorator
 
 
 __all__ = ["Conv2d", "Conv3d", "IA3Config", "IA3Layer", "IA3Model", "Linear"]
--- a/src/peft/tuners/ia3/candidate_collector.py
+++ b/src/peft/tuners/ia3/candidate_collector.py
@@ -1,0 +1,29 @@
+# Copyright 2026-present the HuggingFace Inc. team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""IA3 candidate enumerator. STUB. Returns empty until implemented.
+
+TODO: implement candidate enumeration for IA3. The function should yield
+full module names that IA3 could validly adapt. IA3 has a feedforward vs
+attention distinction (see PR #3154 comment 4250868072) which is out of
+scope for the initial draft and should be addressed in a follow-up PR.
+"""
+
+from peft.tuners.target_suggester import register_candidate_collector
+
+
+@register_candidate_collector("IA3")
+def ia3_candidates(model):
+    """Yield full module names that IA3 could validly adapt. STUB."""
+    return iter(())
--- a/src/peft/tuners/loha/candidate_collector.py
+++ b/src/peft/tuners/loha/candidate_collector.py
@@ -1,4 +1,4 @@
-# Copyright 2023-present the HuggingFace Inc. team.
+# Copyright 2026-present the HuggingFace Inc. team.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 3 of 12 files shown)


BenjaminBossan:
Hi @dentity007, thanks for creating this PR to enable suggestions of possible target modules. I took a look at the design you suggested and unfortunately, I don't think we can go forward with this. It requires creating a new file and adding a registration call for each PEFT method, which is a lot of extra work. If a PEFT method adds support for a new layer type, the registration must be updated, which can easily be overlooked, leading to incorrect suggestions. The tests are inadequate to check that we get sensible suggestions in realistic settings. And the question whether we can generate good suggestions based only on the layer type is yet to be answered.

This is a genuinely hard problem where even I would have to think hard to come up with a good solution (if there is one at all). Coding agents are not there yet to solve this type of problem without the guidance of an experienced programmer. Thus I think it would be better to close the issue for now. I don't want to discourage you, please do continue to work on issues, but let's try to find smaller ones.

For instance, to help users with target module selection, we can add more [defaults](https://github.com/huggingface/peft/blob/285935859286664b00a21beab65d7dc0e86ab823/src/peft/utils/constants.py#L62) for popular models. So if you find gaps, you can contribute to closing them there.

dentity007:
Thanks for the direct review @BenjaminBossan . The maintenance burden of per-tuner registration, the test inadequacy on realistic models, and the open question of whether layer-type alone gives sufficient signal are all fair concerns. I'd rather hear them now than after multiple rounds of revisions on a path you don't think will land. The constants.py defaults path is a much cleaner contribution shape. Is there a list of model families where target-module gaps are recurring, or should I scan the issue tracker for the pattern myself? Happy to pick up a concrete one and start there.

BenjaminBossan:
@dentity007 Thanks for understanding. There is no direct list of model architectures that we know are missing defaults. You could check the most used models on https://huggingface.co/models and compare them to the defaults we have to identify gaps.

dentity007:
Got it, thanks. Digging in on the Hub side now.

On the suggester problem itself: if it gets revisited later, I'd be interested in contributing realistic-model test fixtures, since that was one of the gaps you flagged. No urgency, just letting you know I'd want to be involved if it comes back.

PR #3180, opened 2026-04-20
FIX TinyLoRA non-determinism

BenjaminBossan:
TinyLoRA initialization is not completely deterministic because it partly relies on Python hash, which is not fixed. One of the tests (`test_merge_layers_multi`) is a bit sensitive because the TinyLoRA contribution is rather small, which would result in a ~5% chance of the test failing.

This PR fixes this issue and makes the initialization completely deterministic. The fix is to avoid the Python hash completely and instead ensure that the layer index is always set in time so that it can be used instead of the hash.

While working on this PR, I also did a few tangential changes:

1. small refactor of the interface for better consistency
2. missing tests for TinyLoRA targeting the embedding layer
3. missing tests for multiple active adapters

Check the individual commits to distinguish the individual changes.


kashif:
ah yes! thank you!

kashif (approved):

PR #3183, opened 2026-04-22
fix the test_hotswapping_compiled_model_xxx case of diffusers failure…

sywangyi:
… in xpu

Signed-off-by: Wang, Yi <yi.a.wang@intel.com>

### DIFF
--- a/src/peft/utils/hotswap.py
+++ b/src/peft/utils/hotswap.py
@@ -404,9 +404,8 @@ def hotswap_adapter_from_state_dict(
     # Ensure that all the keys of the new adapter correspond exactly to the keys of the old adapter, otherwise
     # hot-swapping is not possible
 
-    # _orig_mod is for torch.compile(model) and _compiled_call_impl is for model.compile() (not wrapped)
-    is_compiled = hasattr(model, "_orig_mod")
-    is_compiled_inplace = bool(getattr(model, "_compiled_call_impl", None))
+    # _orig_mod is for torch.compile(model)
+    is_compiled_wrapper = hasattr(model, "_orig_mod")
     # TODO: there is probably a more precise way to identify the adapter keys
     missing_keys = {k for k in model.state_dict() if (parameter_prefix in k) and (adapter_name in k)}
     unexpected_keys = []
@@ -419,7 +418,7 @@ def hotswap_adapter_from_state_dict(
             unexpected_keys.append(key)
             continue
 
-        if is_compiled:
+        if is_compiled_wrapper:
             missing_keys.remove("_orig_mod." + key)
         else:
             missing_keys.remove(key)
@@ -464,25 +463,22 @@ def hotswap_adapter_from_state_dict(
         old_val = attrgetter(key)(model)
         new_val = new_val.to(old_val.data.device)
 
-        # We try to detect if the model is compiled but it does not always work, e.g. if hotswapping is called from
-        # within the model itself. In this case, swap_tensors raises RuntimeError and should continue without
-        # swap_tensors.
-        if not is_compiled and not is_compiled_inplace:
-            try:
-                torch.utils.swap_tensors(old_val, new_val)
-                continue
-            except RuntimeError:
-                is_compiled = True
+        # 3 options:
+        # - shapes_match: the new adapter has the same rank as the current tensor (possibly because
+        #   prepare_model_for_compiled_hotswap padded the current tensor to match).
+        # - new_is_smaller: the new adapter has a smaller rank than the current tensor. This happens either
+        #   when the old adapter had a larger rank, or when the current tensor was padded to a larger
+        #   target_rank.
+        # - new_is_larger: the new adapter has a larger rank than the current tensor. The parameter shape
+        #   must change, which is only safe when no padded-shape invariant applies (i.e. the model was
+        #   not padded via prepare_model_for_compiled_hotswap). swap_tensors is the only option here.
+        shapes_match = old_val.shape == new_val.shape
+        new_is_smaller = (not shapes_match) and all(o >= n for o, n in zip(old_val.shape, new_val.shape))
+        new_is_larger = (not shapes_match) and not new_is_smaller
 
-        # Compiled models don't work with swap_tensors because there are weakrefs for the tensor. It is unclear if
-        # this workaround could not cause trouble but the tests indicate that it works.
-        if old_val.shape == new_val.shape:
-            # either
-            # - adapters had the same rank
-            # - adapters were padded with prepare_model_for_compiled_hotswap and 2nd adapter was larger
+        if shapes_match:
             old_val.data.copy_(new_val.data)
-        else:
-            # if 2nd adapter was smaller, ensure to fill up to adapter dimension and set the rest to zeros
+        elif new_is_smaller:
             if old_val.dim() not in (2, 4):
                 raise NotImplementedError(
                     f"Trying to hotswap an adapter whose weight has {old_val.dim()} dimensions, but only Conv2d and "
@@ -502,6 +498,21 @@ def hotswap_adapter_from_state_dict(
                     "ensure that all ranks are padded to the largest rank among all LoRA adapters by using "
                     "peft.utils.hotswap.prepare_model_for_compiled_hotswap."
                 )
+        elif new_is_larger:
+            try:
+                torch.utils.swap_tensors(old_val, new_val)
+            except RuntimeError:
+                # Fallback if swap_tensors is not permitted (e.g. tensor has weakrefs). This still
+                # rebinds storage and will break inductor if the model is compiled, but growing the
+                # rank of a compiled-and-padded model is already unsupported; the caller should have
+                # used prepare_model_for_compiled_hotswap with a sufficient target_rank.
+                old_val.data = new_val.data
+        else:
+            # should be unreachable
+            raise ValueError(
+                "Something went wrong during hotswapping, please open an issue on PEFT: "
+                "https://github.com/huggingface/peft/issues"
+            )
 
 
 def check_hotswap_configs_compatible(config0: PeftConfig, config1: PeftConfig) -> None:


sywangyi:
@BenjaminBossan please help review

BenjaminBossan:
Thanks for the PR. I assume we cannot easily add a unit test for this as it requires XPU?

Regarding the solution itself, I still have a difficult time understanding what it does. From the comment of the Diffusers PR:

> When enable_lora_hotswap was called, weights were padded to target_rank so that torch.compile does not recompile after hotswapping. peft's hotswap_adapter_from_state_dict uses swap_tensors for un-compiled models, which replaces the parameter tensor object. On backends without CUDA Graphs (e.g. XPU), the compiled inductor code re-fetches the real model parameters on every call and its assert_size_stride checks fail because the swapped-in tensor has the smaller (un-padded) rank shape. Fix: temporarily signal to peft that this model should be treated as compiled so it uses the in-place data copy path (which preserves both the tensor object and the padded shape).

Are we talking about a case where 1) the model is compiled but with XPU instead of CUDA and 2) PEFT does not correctly detect that the model is compiled because the checks are not sufficient? If that is the case, I would rather like to see an improved way to detect compilation with XPU. Do you know of any other signal that could be used instead?

As is, the solution relies on the `prepare_model_for_compiled_hotswap` call to add a marker to the model and then it is considered to be compiled. But this is just a proxy. For instance, a user does not need to call `prepare_model_for_compiled_hotswap` even if the model is compiled as long as each LoRA adapter has the same rank. IIUC what you say, in that case we would get a silent failure on XPU because we would still refetch the wrong weights. Therefore, we shouldn't rely on `prepare_model_for_compiled_hotswap`.

If there is no other way, I would still be okay with going ahead, but ideally we can find a cleaner solution -- or maybe I'm just misunderstanding. Pinging @sayakpaul who knows more about model compilation than I do.

sayakpaul:
I am also on the same boat as you @BenjaminBossan ⛵ I also have the same preference:

> If that is the case, I would rather like to see an improved way to detect compilation with XPU. Do you know of any other signal that could be used instead?

sywangyi:
The issue is not that hotswap_adapter() receives a non-compiled model. The issue is that under torch.compile, methods accessed through OptimizedModule.__getattr__ are bound to the inner PEFT model, so inside the hotswap code self is the unwrapped PeftModel, not the OptimizedModule. At that point _orig_mod is not visible, and _compiled_call_impl is not a reliable signal on the inner module. If ranks were padded beforehand, using swap_tensors shrinks the live LoRA parameter back to the smaller rank and breaks the padded-shape invariant. The new check therefore keys off the actual invariant that must be preserved after prepare_model_for_compiled_hotswap, rather than trying to infer compilation from the inner module.

“The root cause of this ‘can’t detect the compiled wrapper’ issue is not specific to XPU; it also exists on CPU. 
On XPU, this is more likely to surface as an explicit failure because the compiled execution path re-reads the real parameters and performs checks such as assert_size_stride, so once the shape shrinks back, it fails immediately.

On CUDA, it is usually less likely to be exposed, mainly because the execution path is different, especially with CUDA Graphs involved. So it is more a case of ‘harder to observe’ rather than ‘the issue does not

sayakpaul:
Thanks for adding further context. 

> The issue is not that hotswap_adapter() receives a non-compiled model. The issue is that under torch.compile, methods accessed through OptimizedModule.getattr are bound to the inner PEFT model, so inside the hotswap code self is the unwrapped PeftModel, not the OptimizedModule. At that point _orig_mod is not visible, and _compiled_call_impl is not a reliable signal on the inner module. 

Seems like instead of `torch.compile(model)`, `model.compile()` would solve this issue because the latter doesn't add `_orig_mod` attribute?

BenjaminBossan:
Just wanted to chime in that `swap_tensors` usage is not a necessity. I found it to work in my tests but if it causes trouble, we can also explore other ways or fully removing it. It just seemed much more elegant than the alternative code path.

sywangyi:
if swap_tensors is delete, following case will fail
python -m pytest tests/test_initialization.py -k "test_hotswap_works_different_ranks_alphas or test_hotswap_works_different_ranks_alphas_conv2d" -q

BenjaminBossan:
@sywangyi I tried to come up with a different solution that reorders the operations, using `swap_tensors`:

```diff
diff --git a/src/peft/utils/hotswap.py b/src/peft/utils/hotswap.py
index 2b08c372..d3b616d5 100644
--- a/src/peft/utils/hotswap.py
+++ b/src/peft/utils/hotswap.py
@@ -404,9 +404,8 @@ def hotswap_adapter_from_state_dict(
     # Ensure that all the keys of the new adapter correspond exactly to the keys of the old adapter, otherwise
     # hot-swapping is not possible
 
-    # _orig_mod is for torch.compile(model) and _compiled_call_impl is for model.compile() (not wrapped)
-    is_compiled = hasattr(model, "_orig_mod")
-    is_compiled_inplace = bool(getattr(model, "_compiled_call_impl", None))
+    # _orig_mod is for torch.compile(model)
+    is_compiled_wrapper = hasattr(model, "_orig_mod")
     # TODO: there is probably a more precise way to identify the adapter keys
     missing_keys = {k for k in model.state_dict() if (parameter_prefix in k) and (adapter_name in k)}
     unexpected_keys = []
@@ -419,7 +418,7 @@ def hotswap_adapter_from_state_dict(
             unexpected_keys.append(key)
             continue
 
-        if is_compiled:
+        if is_compiled_wrapper:
             missing_keys.remove("_orig_mod." + key)
         else:
             missing_keys.remove(key)
@@ -464,25 +463,22 @@ def hotswap_adapter_from_state_dict(
         old_val = attrgetter(key)(model)
         new_val = new_val.to(old_val.data.device)
 
-        # We try to detect if the model is compiled but it does not always work, e.g. if hotswapping is called from
-        # within the model itself. In this case, swap_tensors raises RuntimeError and should continue without
-        # swap_tensors.
-        if not is_compiled and not is_compiled_inplace:
-            try:
-                torch.utils.swap_tensors(old_val, new_val)
-                continue
-            except RuntimeError:
-                is_compiled = True
-
-        # Compiled models don't work with swap_tensors because there are weakrefs for the tensor. It is unclear if
-        # this workaround could not cause trouble but the tests indicate that it works.
-        if old_val.shape == new_val.shape:
-            # either
-            # - adapters had the same rank
-            # - adapters were padded with prepare_model_for_compiled_hotswap and 2nd adapter was larger
+        # 3 options:
+        # - shapes_match: the new adapter has the same rank as the current tensor (possibly because
+        #   prepare_model_for_compiled_hotswap padded the current tensor to match).
+        # - new_is_smaller: the new adapter has a smaller rank than the current tensor. This happens either
+        #   when the old adapter had a larger rank, or when the current tensor was padded to a larger
+        #   target_rank.
+        # - new_is_larger: the new adapter has a larger rank than the current tensor. The parameter shape
+        #   must change, which is only safe when no padded-shape invariant applies (i.e. the model was
+        #   not padded via prepare_model_for_compiled_hotswap). swap_tensors is the only option here.
+        shapes_match = old_val.shape == new_val.shape
+        new_is_smaller = (not shapes_match) and all(o >= n for o, n in zip(old_val.shape, new_val.shape))
+        new_is_larger = (not shapes_match) and not new_is_smaller
+
+        if shapes_match:
             old_val.data.copy_(new_val.data)
-        else:
-            # if 2nd adapter was smaller, ensure to fill up to adapter dimension and set the rest to zeros
+        elif new_is_smaller:
             if old_val.dim() not in (2, 4):
                 raise NotImplementedError(
                     f"Trying to hotswap an adapter whose weight has {old_val.dim()} dimensions, but only Conv2d and "
@@ -502,6 +498,21 @@ def hotswap_adapter_from_state_dict(
                     "ensure that all ranks are padded to the largest rank among all LoRA adapters by using "
                     "peft.utils.hotswap.prepare_model_for_compiled_hotswap."
                 )
+        elif new_is_larger:
+            try:
+                torch.utils.swap_tensors(old_val, new_val)
+            except RuntimeError:
+                # Fallback if swap_tensors is not permitted (e.g. tensor has weakrefs). This still
+                # rebinds storage and will break inductor if the model is compiled, but growing the
+                # rank of a compiled-and-padded model is already unsupported; the caller should have
+                # used prepare_model_for_compiled_hotswap with a sufficient target_rank.
+                old_val.data = new_val.data
+        else:
+            # should be unreachable
+            raise ValueError(
+                "Something went wrong during hotswapping, please open an issue on PEFT: "
+                "https://github.com/huggingface/peft/issues"
+            )
 
 
 def check_hotswap_configs_compatible(config0: PeftConfig, config1: PeftConfig) -> None:
```

Could you please test that patch?

sywangyi:
> @sywangyi I tried to come up with a different solution that reorders the operations, using `swap_tensors`:
> 
> ```diff
> diff --git a/src/peft/utils/hotswap.py b/src/peft/utils/hotswap.py
> index 2b08c372..d3b616d5 100644
> --- a/src/peft/utils/hotswap.py
> +++ b/src/peft/utils/hotswap.py
> @@ -404,9 +404,8 @@ def hotswap_adapter_from_state_dict(
>      # Ensure that all the keys of the new adapter correspond exactly to the keys of the old adapter, otherwise
>      # hot-swapping is not possible
>  
> -    # _orig_mod is for torch.compile(model) and _compiled_call_impl is for model.compile() (not wrapped)
> -    is_compiled = hasattr(model, "_orig_mod")
> -    is_compiled_inplace = bool(getattr(model, "_compiled_call_impl", None))
> +    # _orig_mod is for torch.compile(model)
> +    is_compiled_wrapper = hasattr(model, "_orig_mod")
>      # TODO: there is probably a more precise way to identify the adapter keys
>      missing_keys = {k for k in model.state_dict() if (parameter_prefix in k) and (adapter_name in k)}
>      unexpected_keys = []
> @@ -419,7 +418,7 @@ def hotswap_adapter_from_state_dict(
>              unexpected_keys.append(key)
>              continue
>  
> -        if is_compiled:
> +        if is_compiled_wrapper:
>              missing_keys.remove("_orig_mod." + key)
>          else:
>              missing_keys.remove(key)
> @@ -464,25 +463,22 @@ def hotswap_adapter_from_state_dict(
>          old_val = attrgetter(key)(model)
>          new_val = new_val.to(old_val.data.device)
>  
> -        # We try to detect if the model is compiled but it does not always work, e.g. if hotswapping is called from
> -        # within the model itself. In this case, swap_tensors raises RuntimeError and should continue without
> -        # swap_tensors.
> -        if not is_compiled and not is_compiled_inplace:
> -            try:
> -                torch.utils.swap_tensors(old_val, new_val)
> -                continue
> -            except RuntimeError:
> -                is_compiled = True
> -
> -        # Compiled models don't work with swap_tensors because there are weakrefs for the tensor. It is unclear if
> -        # this workaround could not cause trouble but the tests indicate that it works.
> -        if old_val.shape == new_val.shape:
> -            # either
> -            # - adapters had the same rank
> -            # - adapters were padded with prepare_model_for_compiled_hotswap and 2nd adapter was larger
> +        # 3 options:
> +        # - shapes_match: the new adapter has the same rank as the current tensor (possibly because
> +        #   prepare_model_for_compiled_hotswap padded the current tensor to match).
> +        # - new_is_smaller: the new adapter has a smaller rank than the current tensor. This happens either
> +        #   when the old adapter had a larger rank, or when the current tensor was padded to a larger
> +        #   target_rank.
> +        # - new_is_larger: the new adapter has a larger rank than the current tensor. The parameter shape
> +        #   must change, which is only safe when no padded-shape invariant applies (i.e. the model was
> +        #   not padded via prepare_model_for_compiled_hotswap). swap_tensors is the only option here.
> +        shapes_match = old_val.shape == new_val.shape
> +        new_is_smaller = (not shapes_match) and all(o >= n for o, n in zip(old_val.shape, new_val.shape))
> +        new_is_larger = (not shapes_match) and not new_is_smaller
> +
> +        if shapes_match:
>              old_val.data.copy_(new_val.data)
> -        else:
> -            # if 2nd adapter was smaller, ensure to fill up to adapter dimension and set the rest to zeros
> +        elif new_is_smaller:
>              if old_val.dim() not in (2, 4):
>                  raise NotImplementedError(
>                      f"Trying to hotswap an adapter whose weight has {old_val.dim()} dimensions, but only Conv2d and "
> @@ -502,6 +498,21 @@ def hotswap_adapter_from_state_dict(
>                      "ensure that all ranks are padded to the largest rank among all LoRA adapters by using "
>                      "peft.utils.hotswap.prepare_model_for_compiled_hotswap."
>                  )
> +        elif new_is_larger:
> +            try:
> +                torch.utils.swap_tensors(old_val, new_val)
> +            except RuntimeError:
> +                # Fallback if swap_tensors is not permitted (e.g. tensor has weakrefs). This still
> +                # rebinds storage and will break inductor if the model is compiled, but growing the
> +                # rank of a compiled-and-padded model is already unsupported; the caller should have
> +                # used prepare_model_for_compiled_hotswap with a sufficient target_rank.
> +                old_val.data = new_val.data
> +        else:
> +            # should be unreachable
> +            raise ValueError(
> +                "Something went wrong during hotswapping, please open an issue on PEFT: "
> +                "https://github.com/huggingface/peft/issues"
> +            )
>  
>  
>  def check_hotswap_configs_compatible(config0: PeftConfig, config1: PeftConfig) -> None:
> ```
> 
> Could you please test that patch?

yes, it works!!!

BenjaminBossan:
> yes, it works!!!

Great, then let's update the PR with the proposed patch, I think it should be the more robust solution.

sywangyi:
> > yes, it works!!!
> 
> Great, then let's update the PR with the proposed patch, I think it should be the more robust solution.

done

BenjaminBossan (approved):
Thanks for figuring out the issue and working with us on a solution. I ran the GPU compile tests locally and they passed. Failing CI is unrelated.

PR #3185, opened 2026-04-22
ENH Optimize HRA forward for better memory efficiency

BenjaminBossan:
## Description

HRA previously materialized the full dense Householder transform and multiplied it into the base weight on every forward pass. This can be expensive for big weight matrices. Moreover, the base weight was cast to the adapter dtype, which can create an extra copy at higher dtype. The new code is more in line with how LoRA calculates the forward step.

## Context

I wanted to run HRA on the new image generation benchmark which is in the works. No matter what I did, it always OOM-ed, so I worked with Codex + Fable to find possible optimizations. The new `Linear.forward` looks quite reasonable to me, for `Conv2d.forward`, I can't say I fully grasp it but it seems to work.

The PR now also contains a MetaMath experiment (this would OOM with the current main).

**UPDATE**

I took up this PR again and let Fable apply some more improvements. I ran the checks according to the skill from #3544.

Unit tests and regression tests for HRA pass.

## Benchmark results

Since there is no benchmark for MetaMathQA (given that it would OOM), I instead ran the image generation benchmark for comparison. It's on my machine, so the results are not 100% comparable, but I think that the memory consumption and runtime values speak for themselves:

Before:

```json
{
    "accelerator_memory_reserved_avg": 33476879319,
    "accelerator_memory_max": 47359983616,
    "accelerator_memory_reserved_99th": 47326429184,
    "train_time": 9824.567679766005,
    ...
      {
        "step": 100,
        "valid dino_similarity": 0.3935689628124237,
        "train loss": 0.7713394165039062,
        "train samples": 200,
        "train time": 975.4857535669962,
        "eval time": 214.68726077499923,
        "mem allocated avg": 8422570690.56,
        "mem reserved avg": 47323031797.76,
        "elapsed time": 1196.5852056719996
      }
      ...
      {
        "step": 750,
        "test dino_similarity": 0.6441149314244589,
        "drift": 0.2597343921661377,
        "train loss": 0.741257905960083,
        "train samples": 1500
      }
}
```

After:

```json
{
    "accelerator_memory_reserved_avg": 10040308137,
    "accelerator_memory_max": 10055843840,
    "accelerator_memory_reserved_99th": 10055843840,
    "train_time": 551.8845958401216,
      {
        "step": 100,
        "valid dino_similarity": 0.40681228041648865,
        "train loss": 0.7726805806159973,
        "train samples": 200,
        "train time": 58.19236221892061,
        "eval time": 7.051100489974488,
        "mem allocated avg": 8439743744.0,
        "mem reserved avg": 9955180544.0,
        "elapsed time": 72.87472819897812
      }
      ...
      {
        "step": 750,
        "test dino_similarity": 0.6774141788482666,
        "drift": 0.23255375027656555,
        "train loss": 0.7409864068031311,
        "train samples": 1500
      }
}
```

| train loss before | train loss after | diff      |
|-------------------|------------------|-----------|
| 0.771339          | 0.772681         | -0.001341 |
| 0.788062          | 0.787517         | 0.000545  |
| 0.753804          | 0.753400         | 0.000404  |
| 0.781244          | 0.779521         | 0.001723  |
| 0.742924          | 0.742203         | 0.000721  |
| 0.774471          | 0.774891         | -0.000420 |
| 0.734791          | 0.735006         | -0.000215 |
| 0.741258          | 0.740986         | 0.000271  |

MetaMathQA:

Before: OOM

After:

`{"step": "  250", "samples": "   1000", "lr": "5.00e-05", "loss avg": "1.2145", "valid acc": "0.060", "gen valid tokens": 13137, "train time": "31.1s", "eval time": "10.1s", "train tokens / sec": "6806", "mem allocated": "6869058808", "mem reserved": "14490674397", "elapsed time": "1min 5s"}`

```
accelerator memory max: 18406MB
accelerator memory reserved avg: 13642MB
accelerator memory reserved 99th percentile: 18406MB
train time: 90.53958111797692s
```

## Comparison script:

<details>

<summary>click to see the script script</summary>

```python
"""Compare HRA model outputs before and after a code change.

On the first run, an adapter checkpoint and the computed outputs (unmerged and merged) are stored in a temporary
directory per test case, dtype, and device. On subsequent runs, the checkpoint is loaded, the outputs are computed
afresh and compared to the stored ones, reporting the differences. Thus: run once on the old code, then once on the
new code, using the same device both times. Delete the directory or pass --reset to store fresh reference outputs.

Example:

    git checkout main
    python scripts/compare-hra-outputs.py
    git checkout <branch>
    python scripts/compare-hra-outputs.py

"""

import argparse
import sys
import tempfile
from pathlib import Path

import torch
from transformers import AutoModelForCausalLM, AutoModelForImageClassification

from peft import HRAConfig, PeftModel, get_peft_model


# Disable TF32 so that both code versions compute in true float32: F.conv2d (cudnn, TF32 on by default) and
# unfold + F.linear (matmul, TF32 off by default) would otherwise differ by TF32's precision (~1e-3 relative),
# drowning out the differences this script is meant to measure.
torch.backends.cudnn.allow_tf32 = False
torch.backends.cuda.matmul.allow_tf32 = False

STORE_PATH = Path(tempfile.gettempdir()) / "hra-output-check"
MODEL_ID = "facebook/opt-125m"
CONV_MODEL_ID = "peft-internal-testing/tiny-random-ResNetForImageClassification"
SEED = 0
# tolerances on the logits relative to their largest magnitude, scaled to the precision of the dtype
DTYPE_ATOL = {"float64": 1e-10, "float32": 1e-4, "float16": 1e-2, "bfloat16": 1e-1}
LINEAR_TARGETS = ["q_proj", "v_proj"]


def get_cases():
    # init_weights=False everywhere: the default (symmetric) initialization creates pairs of identical Householder
    # reflections that cancel each other out, so the adapter would be a no-op and the comparison trivial.
    # The tiny ResNet uses bottleneck layers: "layer.0"/"shortcut" convolutions are 1x1 (some with stride 2),
    # "layer.1" convolutions are 3x3, and the embedder convolution is 7x7 with stride 2 and padding 3. This covers
    # both the 1x1 fast path and the unfold path of the new conv forward. Each case maps to a list of configs; cases
    # with more than one config test multiple active adapters, where the composition order of the Householder
    # reflections matters.
    return {
        "linear-defaults":    ("opt", [HRAConfig(target_modules=LINEAR_TARGETS, init_weights=False)]),
        "linear-gs":          ("opt", [HRAConfig(target_modules=LINEAR_TARGETS, apply_GS=True, init_weights=False)]),
        "linear-r-odd":       ("opt", [HRAConfig(target_modules=LINEAR_TARGETS, r=7, init_weights=False)]),
        "linear-two-adapters": ("opt", [
            HRAConfig(target_modules=LINEAR_TARGETS, init_weights=False),
            HRAConfig(target_modules=LINEAR_TARGETS, r=4, init_weights=False),
        ]),
        "conv-1x1":           ("resnet", [HRAConfig(target_modules=["layer.0.convolution", "shortcut.convolution"], init_weights=False)]),
        "conv-3x3-7x7":       ("resnet", [HRAConfig(target_modules=["layer.1.convolution", "embedder.convolution"], init_weights=False)]),
        "conv-gs":            ("resnet", [HRAConfig(target_modules=["layer.0.convolution", "layer.1.convolution"], apply_GS=True, init_weights=False)]),
    }


def load_base_model(kind, dtype):
    if kind == "opt":
        model = AutoModelForCausalLM.from_pretrained(MODEL_ID)
    else:
        model = AutoModelForImageClassification.from_pretrained(CONV_MODEL_ID)
    return model.to(dtype)


def get_inputs(kind, device, dtype):
    # Call right after seeding, before the model is built: the amount of RNG state consumed by the model setup
    # differs between the first run (adapter initialization) and later runs (loading the checkpoint).
    if kind == "opt":
        return torch.arange(20).view(2, 10).to(device)
    return torch.randn(2, 3, 32, 32).to(device=device, dtype=dtype)


def extra_adapter_names(configs):
    return [f"adapter{i}" for i in range(1, len(configs))]


def compute_outputs(model, inputs):
    model.eval()

    def forward():
        with torch.inference_mode():
            output = model(inputs)
        return (output.logits if hasattr(output, "logits") else output).double().cpu()

    unmerged = forward()
    model.merge_adapter()
    merged = forward()
    return {"unmerged": unmerged, "merged": merged}


def create_reference(kind, configs, device, dtype, case_dir):
    """First run: create the adapter checkpoint and store the reference outputs."""
    torch.manual_seed(SEED)
    inputs = get_inputs(kind, device, dtype)
    model = load_base_model(kind, dtype)
    # autocast_adapter_dtype=False so that for float16/bfloat16, the adapter itself runs in that dtype
    model = get_peft_model(model, configs[0], autocast_adapter_dtype=False)
    for name, config in zip(extra_adapter_names(configs), configs[1:]):
        model.add_adapter(name, config, autocast_adapter_dtype=False)
    if len(configs) > 1:
        model.base_model.set_adapter(["default"] + extra_adapter_names(configs))
    model = model.to(device)
    model.save_pretrained(case_dir / "checkpoint")
    torch.save(compute_outputs(model, inputs), case_dir / "outputs.pt")


def load_and_compute(kind, configs, device, dtype, case_dir):
    """Subsequent runs: load the stored checkpoint and compute fresh outputs."""
    torch.manual_seed(SEED)
    inputs = get_inputs(kind, device, dtype)
    model = load_base_model(kind, dtype)
    model = PeftModel.from_pretrained(model, case_dir / "checkpoint", autocast_adapter_dtype=False)
    for name in extra_adapter_names(configs):
        model.load_adapter(case_dir / "checkpoint" / name, adapter_name=name, autocast_adapter_dtype=False)
    if len(configs) > 1:
        model.base_model.set_adapter(["default"] + extra_adapter_names(configs))
    model = model.to(device)
    return compute_outputs(model, inputs)


def compare_case(name, stored, current, atol):
    """Compare stored and current outputs of one case, return False on mismatch."""
    ok = True
    for key in ("unmerged", "merged"):
        diff = (stored[key] - current[key]).abs()
        max_abs_diff = diff.max().item()
        # judge the difference relative to the magnitude of the outputs: one ulp of a logit of 30 is 30x one ulp
        # of a logit of 1, so a fixed absolute tolerance would misclassify models with large outputs
        scale = stored[key].abs().max().item()
        threshold = atol * max(scale, 1.0)
        if max_abs_diff > threshold:
            ok = False
        status = "OK" if max_abs_diff <= threshold else "MISMATCH"
        print(
            f"{name} ({key}): {status} -- max abs diff {max_abs_diff:.3e} "
            f"(output scale {scale:.3e}, threshold {threshold:.3e})"
        )
    return ok


def run_dtype(dtype_name, device, reset):
    """Store or compare the outputs of all cases for one dtype, return False on mismatch."""
    dtype = getattr(torch, dtype_name)
    atol = DTYPE_ATOL[dtype_name]
    device_tag = device.replace(":", "")
    all_ok = True

    for name, (kind, configs) in get_cases().items():
        case_dir = STORE_PATH / f"{name}-{dtype_name}-{device_tag}"
        if reset or not (case_dir / "outputs.pt").exists():
            case_dir.mkdir(parents=True, exist_ok=True)
            create_reference(kind, configs, device, dtype, case_dir)
            print(f"[{dtype_name}] {name}: stored reference checkpoint and outputs at {case_dir}")
            continue

        current = load_and_compute(kind, configs, device, dtype, case_dir)
        stored = torch.load(case_dir / "outputs.pt", weights_only=True)
        all_ok &= compare_case(f"[{dtype_name}] {name}", stored, current, atol)

    print()
    return all_ok


def main():
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--device", default="cuda" if torch.cuda.is_available() else "cpu")
    parser.add_argument(
        "--dtypes",
        nargs="+",
        default=["float32", "float16", "bfloat16"],
        choices=list(DTYPE_ATOL),
        help="dtypes of the model to check; float64 distinguishes real differences from fp reordering noise",
    )
    parser.add_argument("--reset", action="store_true", help="store fresh reference outputs")
    args = parser.parse_args()

    all_ok = True
    for dtype_name in args.dtypes:
        all_ok &= run_dtype(dtype_name, args.device, args.reset)

    if all_ok:
        print("All comparable outputs match (or references were freshly created).")
    else:
        print("Some outputs DO NOT match, see above.")
        sys.exit(1)


if __name__ == "__main__":
    main()
```

</details>

It passes too.

## Adjacent changes

- raise an error for `groups!=1` (not supported)
- works with non-square conv kernels

**_Below is the old PR description, which I leave here for reference, but it's no longer up-to-date_**

## Results

I ran some regression tests to determine that he results after the change are numerically close to the previous results. There can be deviations, especially at float16/bfloat16, but I think this is acceptable for the memory savings we see. Maybe there is a better way that is more faithful, LMK if you have an idea.

Below are the results of running the new HRA on the MetaMathQA and the image generation benchmark compared to the current implementation. For opt-125, memory usage actually increases, but for more realistic sizes like Llama 3.2 3B or Flux 2 klein 4B, we see an improvement.

| Benchmark        | Model           | HRA rank | Seq len | Optimization | Memory  | Runtime            |
|------------------|-----------------|----------|---------|--------------|---------|--------------------|
| MetaMath         | Llama 3.2 3B    |       32 |     768 | no           | OOM     | -                  |
| MetaMath         | Llama 3.2 3B    |       32 |     768 | yes          | OOM     | -                  |
| MetaMath         | Llama 3.2 3B    |       32 |     128 | no           | OOM     | -                  |
| MetaMath         | Llama 3.2 3B    |       32 |     128 | yes          | 23.5 GB | 71 sec / 250 steps |
| MetaMath         | Llama 3.2 3B    |       32 |      64 | no           | OOM     | -                  |
| MetaMath         | Llama 3.2 3B    |       32 |      64 | yes          | 16 GB   | 63 sec / 250 steps |
| MetaMath         | OPT-125M        |       32 |     768 | no           | 4.7 GB  | 27 sec / 250 steps |
| MetaMath         | OPT-125M        |       32 |     768 | yes          | 9.9 GB  | 24 sec / 250 steps |
| Image generation | Flux 2 klein 4B |       32 |       - | no           | OOM     | -                  |
| Image generation | Flux 2 klein 4B |       32 |       - | yes          | 20 GB   | -                  |

Note that for the final benchmarking results, we'll use a GPU with 48 GB, so that should hopefully not OOM.


BenjaminBossan:
@DaShenZi721 Could you please take look?

BenjaminBossan:
gentle ping @DaShenZi721

BenjaminBossan:
One last attempt: @DaShenZi721

BenjaminBossan:
_not stale_

BenjaminBossan:
_not stale_

PR #3187, opened 2026-04-23
FIX VeRA and OFT error after layer init refactor

BenjaminBossan:
PR #3047 introduced issues with layer initialization for VeRA when using bitsandbytes, as well as for OFT when using GPTQ. These tests failed on the nightly GPU CI.

I confirmed locally that the tests now pass.

Failing tests:

- https://github.com/huggingface/peft/actions/runs/24814323261/job/72625376879#step:6:1483
- https://github.com/huggingface/peft/actions/runs/24814323261/job/72625376879#step:5:1527

### DIFF
--- a/src/peft/tuners/oft/gptq.py
+++ b/src/peft/tuners/oft/gptq.py
@@ -74,6 +74,7 @@ class GPTQOFTLinear(torch.nn.Module, OFTLayer):
 def dispatch_gptq(
     target: torch.nn.Module,
     adapter_name: str,
+    oft_config: OFTConfig,
     **kwargs: Any,
 ) -> Optional[torch.nn.Module]:
     new_module = None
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Note: Failing tests (`TypeError: WeightConverter.__init__() got an unexpected keyword argument 'distributed_operation'`) are due to a regression in the latest Transformers and are unrelated.

src/peft/tuners/oft/gptq.py
```diff
@@ -74,6 +74,7 @@ def __repr__(self) -> str:
 def dispatch_gptq(
     target: torch.nn.Module,
     adapter_name: str,
+    oft_config: OFTConfig,
```
sayakpaul, line 77: Seems like a harmless change and it won't break backward compatibility as this function isn't exposed anyway.

sayakpaul (approved):
Failing tests are unrelated?

BenjaminBossan:
Yep, failing tests are caused by a regression in Transformers. The change is internal, so different signature should be fine.

PR #3188, opened 2026-04-23
CI Fix GPTQmodel install in Docker build

BenjaminBossan:
While building the GPU Docker image, there is an [error](https://github.com/huggingface/peft/actions/runs/24811812656/job/72618091480#step:5:5146) because GPTQModel cannot be installed successfully. This is caused by the `--no-build-isolation flag`, which is no longer required according to the GPTQModel README.

The next issue was conflicting CUDA versions with EETQ, which I fixed by bumping the image.

Finally, TE wouldn't build, which required setting some env vars.

The docker build now finishes, as can be seen in the CI.


SunMarc (approved):
Nice !

PR #3189, opened 2026-04-23
FIX PVeRA forward implementation for bitsandbytes

BenjaminBossan:
In the forward path of PVeRA, a sampling step is required. For the bnb layers, this was missing. The PR now adds the sampling step.

I confirmed locally that this fixes the [failing PVeRA tests in the nightly CI](https://github.com/huggingface/peft/actions/runs/24814323261/job/72625376879#step:6:1482). Failing tests on CPU CI are unrelated.

### DIFF
--- a/src/peft/tuners/pvera/bnb.py
+++ b/src/peft/tuners/pvera/bnb.py
@@ -225,7 +225,7 @@ if is_bnb_available():
                     sliced_B = pvera_B[: self.out_features, :].to(x.device)
 
                     x_temp = dropout(x.to(lambda_d.dtype))
-                    mu, logvar = (lambda_d * F.linear(dropout(x_temp), sliced_A)).chunk(2, dim=-1)
+                    mu, logvar = (lambda_d * F.linear(x_temp, sliced_A)).chunk(2, dim=-1)
                     adapter_output = lambda_b * F.linear(
                         self._reparametrize(mu, logvar, self.sample_at_inference), sliced_B
                     )
@@ -395,7 +395,7 @@ if is_bnb_4bit_available():
                     sliced_B = pvera_B[: self.out_features, :].to(x.device)
 
                     x_temp = dropout(x.to(lambda_d.dtype))
-                    mu, logvar = (lambda_d * F.linear(dropout(x_temp), sliced_A)).chunk(2, dim=-1)
+                    mu, logvar = (lambda_d * F.linear(x_temp, sliced_A)).chunk(2, dim=-1)
                     adapter_output = lambda_b * F.linear(
                         self._reparametrize(mu, logvar, self.sample_at_inference), sliced_B
                     )
--- a/src/peft/tuners/pvera/bnb.py
+++ b/src/peft/tuners/pvera/bnb.py
@@ -18,6 +18,7 @@ from typing import Optional
 
 import bitsandbytes as bnb
 import torch
+import torch.nn.functional as F
 
 from peft.import_utils import is_bnb_4bit_available, is_bnb_available
 from peft.tuners.tuners_utils import check_adapters_to_merge
@@ -44,6 +45,7 @@ if is_bnb_available():
             super().__init__()
             PveraLayer.__init__(self, base_layer)
             self.fan_in_fan_out = config.fan_in_fan_out
+            self.sample_at_inference = config.sample_at_inference
 
             self._active_adapter = adapter_name
             self.update_layer(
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
@leofillioux Please review.

src/peft/tuners/pvera/bnb.py
```diff
@@ -223,9 +225,9 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
                     sliced_B = pvera_B[: self.out_features, :].to(x.device)
 
                     x_temp = dropout(x.to(lambda_d.dtype))
-
-                    adapter_output = lambda_b * torch.nn.functional.linear(
-                        lambda_d * torch.nn.functional.linear(x_temp, sliced_A), sliced_B
+                    mu, logvar = (lambda_d * F.linear(dropout(x_temp), sliced_A)).chunk(2, dim=-1)
+                    adapter_output = lambda_b * F.linear(
+                        self._reparametrize(mu, logvar, self.sample_at_inference), sliced_B
```
BenjaminBossan, line 230: Now mirrors:

https://github.com/huggingface/peft/blob/76c37d4686cf7245a9a6998fd1e2619d62c9322d/src/peft/tuners/pvera/layer.py#L300-L303

src/peft/tuners/pvera/bnb.py
```diff
@@ -223,9 +225,9 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
                     sliced_B = pvera_B[: self.out_features, :].to(x.device)
 
                     x_temp = dropout(x.to(lambda_d.dtype))
-
-                    adapter_output = lambda_b * torch.nn.functional.linear(
-                        lambda_d * torch.nn.functional.linear(x_temp, sliced_A), sliced_B
+                    mu, logvar = (lambda_d * F.linear(dropout(x_temp), sliced_A)).chunk(2, dim=-1)
```
leofillioux: Dropout if already applied above `x_temp = dropout(x.to(lambda_d.dtype))`. I think we should just use `x_temp` directly in the linear layer here.

src/peft/tuners/pvera/bnb.py
```diff
@@ -392,9 +395,9 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
...
```
leofillioux: Same comment on the double use of dropout as above.

leofillioux (commented):
Hey @BenjaminBossan! Almost everything looked good, except for the double use of dropout.

BenjaminBossan:
@leofillioux Good catch, that was indeed an error on my part. It should be fixed now, please check again. (failing CI is still unrelated)

leofillioux (commented):
Everything looks good to me! Thank you!

src/peft/tuners/pvera/layer.py
```diff
@@ -143,6 +143,15 @@ def reset_pvera_parameters(self, adapter_name, d_initial: float = 0.1):
                 nn.init.zeros_(self.pvera_lambda_d[adapter_name]).fill_(d_initial)
                 nn.init.zeros_(self.pvera_lambda_b[adapter_name])
 
+    def _reparametrize(self, mu, logvar, sample_at_inference):
+        if self.training or (not self.training and sample_at_inference):
+            std = torch.exp(0.5 * logvar)
+            eps = torch.randn_like(std)
```
sayakpaul, line 149: Thinking out loud. Should there be a `generator` argument to control stochasticity?

sayakpaul (approved):
Thanks!

src/peft/tuners/pvera/layer.py
BenjaminBossan: Good point, I agree that having an option to seed would be good here. @leofillioux WDYT? I still merged this PR, as it is a separate issue.

leofillioux: Yes that's a good idea! I made an implementation, and I'm running the tests. I'll open a PR once they're done.

leofillioux: @BenjaminBossan I still have problems with my environment and I'm not able to run the tests :/ Should I still make a PR?

BenjaminBossan: Sure, go ahead.

leofillioux: I opened a PR (#3200) to address this problem.

PR #3192, opened 2026-04-23
fix(modules_to_save): support kwargs-only forward in AuxiliaryTrainingWrapper

### LINKED ISSUE #3191: AuxiliaryTrainingWrapper.forward requires positional `x`, breaks kwargs-only modules in modules_to_save
### Feature request

# Description

### System Info

- peft: 0.19.0
- transformers: >= 4.50 (any version with Gemma 4 support)
- python: 3.11
- platform: Linux

### Who can help?

@BenjaminBossan @githubnemo

### Reproduction

`AuxiliaryTrainingWrapper.forward` is defined with a required positional first argument:

```python
# peft/utils/other.py
def forward(self, x: torch.Tensor, *args, **kwargs):
    self._check_forward_args(x, *args, **kwargs)
    ...
    return self._forward_wrapped(x, *args, **kwargs)
```

Any module listed in `modules_to_save` that its parent invokes with keyword arguments only will crash at the first forward pass with:

```
TypeError: AuxiliaryTrainingWrapper.forward() missing 1 required positional argument: 'x'
```

**Minimal repro:**

```python
import torch
import torch.nn as nn
from peft.utils.other import ModulesToSaveWrapper

class KwargsOnly(nn.Module):
    def __init__(self):
        super().__init__()
        self.lin = nn.Linear(8, 8)
    def forward(self, *, pixel_values):
        return self.lin(pixel_values)

w = ModulesToSaveWrapper(KwargsOnly(), "default")
w.set_adapter("default")
w(pixel_values=torch.randn(2, 8))
# TypeError: AuxiliaryTrainingWrapper.forward() missing 1 required positional argument: 'x'
```

**Real-world trigger — Gemma 4 VLM:**

Gemma 4's forward path calls vision modules with keyword arguments only:

```python
vision_outputs = self.vision_tower(pixel_values=..., ...)
inputs_embeds  = self.embed_vision(inputs_embeds=...)
```

A config like:

```yaml
adapter: lora
lora_modules_to_save:
  - vision_tower
  - embed_vision
```

fails at eval step 0 (or the first training forward).

### Expected behavior

`modules_to_save` should work for any module, regardless of whether it's called positionally (e.g. `embed_tokens(input_ids)`) or with keyword arguments (e.g. `vision_tower(pixel_values=...)`). The wrapper should forward whatever args/kwargs it receives to the wrapped module.

### Suggested fix

Change three methods to accept `*args, **kwargs` instead of requiring positional `x`:

- `AuxiliaryTrainingWrapper.forward`
- `ModulesToSaveWrapper._forward_wrapped`
- `ModulesToSaveWrapper._forward_wrapped_passthrough`

Short-circuit `_check_forward_args` when no positional args are present — its only work is validating `len(x) == len(adapter_names)`, which is itself gated on `adapter_names` being set, so it's a no-op in the kwargs-only case.

Leave `_mixed_batch_forward` untouched — it needs positional input for sub-batch indexing and only runs under multi-adapter inference (when `adapter_names` is passed), which is independent of this bug.

This is backwards-compatible: positional calls continue to work unchanged.

### Your contribution

Happy to send a PR.

Anai-Guo:
## Problem

Closes #3191.

`AuxiliaryTrainingWrapper.forward` required a positional first argument `x`, so any module in `modules_to_save` called with keyword-only arguments crashed immediately:

```
TypeError: AuxiliaryTrainingWrapper.forward() missing 1 required positional argument: 'x'
```

Real-world trigger: Gemma 4 VLM calls `vision_tower(pixel_values=...)` and `embed_vision(inputs_embeds=...)` with keyword arguments only, making `modules_to_save` unusable for these layers.

## Fix

Change `forward`, `_forward_wrapped`, `_forward_wrapped_passthrough`, and `_check_forward_args` in `AuxiliaryTrainingWrapper` to accept `*args` instead of requiring positional `x`:

- `_check_forward_args`: early-return when no positional args (the batch-size check against `adapter_names` is only meaningful for positional input; kwargs-only calls skip it)
- `forward` / `_forward_wrapped` / `_forward_wrapped_passthrough`: accept `*args, **kwargs` and pass through unchanged
- `ModulesToSaveWrapper` and `TrainableTokensWrapper` overrides updated accordingly

`_mixed_batch_forward` is **unchanged** — it needs positional input for sub-batch indexing and is only invoked when `adapter_names` is explicitly passed (multi-adapter inference), which always uses positional tensors.

## Backwards compatibility

Fully backwards-compatible: positional calls like `embed_tokens(input_ids)` continue to work exactly as before.

## Minimal repro (now works)

```python
import torch
import torch.nn as nn
from peft.utils.other import ModulesToSaveWrapper

class KwargsOnly(nn.Module):
    def __init__(self):
        super().__init__()
        self.lin = nn.Linear(8, 8)
    def forward(self, *, pixel_values):
        return self.lin(pixel_values)

w = ModulesToSaveWrapper(KwargsOnly(), "default")
w.set_adapter("default")
out = w(pixel_values=torch.randn(2, 8))  # previously crashed, now works
```

### DIFF
--- a/src/peft/utils/other.py
+++ b/src/peft/utils/other.py
@@ -403,20 +403,22 @@ class AuxiliaryTrainingWrapper(torch.nn.Module):
         new_hook = old_hook_cls(**filtered_old_hook_attr)
         return new_hook
 
-    def _check_forward_args(self, x, *args, **kwargs):
+    def _check_forward_args(self, *args, **kwargs):
         """Check if the arguments are compatible with the configs and state of the model"""
         adapter_names = kwargs.get("adapter_names", None)
         if adapter_names is None:
             return
+        if not args:
+            return  # kwargs-only call; batch-size check requires positional input
 
-        if len(x) != len(adapter_names):
+        if len(args[0]) != len(adapter_names):
             msg = (
                 "Length of `adapter_names` should be the same as the number of inputs, but got "
-                f"{len(adapter_names)} and {len(x)} respectively."
+                f"{len(adapter_names)} and {len(args[0])} respectively."
             )
             raise ValueError(msg)
 
-    def _forward_wrapped(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
+    def _forward_wrapped(self, *args: Any, **kwargs: Any) -> torch.Tensor:
         raise NotImplementedError
 
     def _forward_wrapped_mixed_batch(
@@ -424,7 +426,7 @@ class AuxiliaryTrainingWrapper(torch.nn.Module):
     ) -> torch.Tensor:
         raise NotImplementedError
 
-    def _forward_wrapped_passthrough(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
+    def _forward_wrapped_passthrough(self, *args: Any, **kwargs: Any) -> torch.Tensor:
         """The forward call when no adapter is involved in the forward computation, only the base model"""
         raise NotImplementedError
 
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Closing in favor of #3199

PR #3197, opened 2026-04-24
FIX Transformers weight conversion regression

BenjaminBossan:
After a change in https://github.com/huggingface/transformers/pull/45448, weight conversion tests started failing. Transformers provided a fix in https://github.com/huggingface/transformers/pull/45622 but it needs to be ported to PEFT too. The PR also ports the fix from https://github.com/huggingface/transformers/pull/45428.

This PR fixes the issue from the PEFT side. For the transformers side (`model.load_adapter`), we need to wait for https://github.com/huggingface/transformers/pull/45622 to land and be released.

### DIFF
--- a/src/peft/utils/transformers_weight_conversion.py
+++ b/src/peft/utils/transformers_weight_conversion.py
@@ -255,14 +255,14 @@ def build_peft_weight_mapping(
                 # TODO: this assumption may not hold for models != mixtral
                 # For source, we capture the original weights + the lora weights
                 new_source_patterns = []
-                for pat in list(orig_conversion._original_source_patterns):
+                for pat in list(orig_conversion.source_patterns):
                     # we replace the weight pattern to colllect loras
                     pat = pat.rsplit(".", 1)[0]
                     # note: the source state_dict does *not* contain the adapter name
                     new_source_patterns.append(f"{pat}.{lora}.*")
 
                 # the gate_up_proj is the innner PEFT ParamWrapper, so we need to use base_layer
-                pat = orig_conversion._original_target_patterns[0]
+                pat = orig_conversion.target_patterns[0]
                 pat = pat.replace("gate_up_proj", "base_layer")
                 # we make sure the target key is correct, add '.weight' because the parameter is targeted directly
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
@Cyrilvallez please review.

Cyrilvallez (approved):
LGTM, it matches the transformers's PR!

src/peft/utils/transformers_weight_conversion.py
```diff
@@ -252,14 +255,14 @@ def build_peft_weight_mapping(
                 # TODO: this assumption may not hold for models != mixtral
                 # For source, we capture the original weights + the lora weights
                 new_source_patterns = []
-                for pat in list(orig_conversion.source_patterns):
+                for pat in list(orig_conversion._original_source_patterns):
```
sayakpaul: If `orig_conversion` stems from `transformers` then I think it might be a breaking change? Because if someone is using an older `transformers` version, then it could be problematic? If not, then it's okay.

sayakpaul (approved):

src/peft/utils/transformers_weight_conversion.py
BenjaminBossan: @Cyrilvallez Would this change work with older Transformers versions?

Cyrilvallez: It was added in https://github.com/huggingface/transformers/pull/45340, so indeed will not be fully compatible with all previous versions - for your use case you could simply use `source_patterns`/`target_patterns` instead of their counterpart `_original_xxx` to avoid checking versions

BenjaminBossan: Got it, I reverted the argument names back. I tested locally with transformers v5.4.0, v5.5.4, and the latest main, and the test passed.

What I wonder about: If we can use `source_patterns`/`target_patterns`, why the change to `_original_xxx` in the first place?

BenjaminBossan:
Tests are now passing (except the ones from Transformers side, which require a new Transformers release).

BenjaminBossan:
Transformers release is out, so we just need this PR for tests to pass again. I'll go ahead and merge it.

PR #3198, opened 2026-04-24
chore: bump doc-builder SHA for main doc build workflow

rtrompier:
Bump the pinned doc-builder SHA so that main documentation builds also sync to the HF bucket (dual-write introduced in doc-builder PR #780, 2026-04-15). The current pin predates that change, so release docs never land in the bucket — only in the legacy dataset.

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -11,7 +11,7 @@ permissions: {}
 
 jobs:
    build:
-    uses: huggingface/doc-builder/.github/workflows/build_main_documentation.yml@98154fc40f6896c500b4b4aaf7f2757b4f9cb2ad  # main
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
When I visit the doc preview, I get:

> ENOTCONN: socket is not connected, open '/workspace/repos/doc-build-dev/peft/_versions.yml' (internal error hidden in production)
> Request ID: Root=1-69eb6ba9-747491be686639ce193e694e

We don't expect this PR to fix that, do we?

rtrompier:
> When I visit the doc preview, I get:
> 
> > ENOTCONN: socket is not connected, open '/workspace/repos/doc-build-dev/peft/_versions.yml' (internal error hidden in production)
> > Request ID: Root=1-69eb6ba9-747491be686639ce193e694e
> 
> We don't expect this PR to fix that, do we?

Nop, not related, but should be fixed now

BenjaminBossan (approved):
All good from my side.

PR #3200, opened 2026-04-28
Add a generator to the PVeRA reparameterization function.

leofillioux:
This PR addresses [this comment](https://github.com/huggingface/peft/pull/3189#discussion_r3140569649) from PR #3189.

It integrates a generator in the reparameterization function from PVeRA, and uses it for sampling the random epsilon. This generator (`torch.Generator`) can be specified in the config.

### DIFF
--- a/src/peft/tuners/pvera/config.py
+++ b/src/peft/tuners/pvera/config.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 from __future__ import annotations
 
+import torch
 import warnings
 from dataclasses import dataclass, field
 from typing import Optional, Union
@@ -75,6 +76,8 @@ class PveraConfig(PeftConfig):
             default for non-specified adapters). For example
             `sample_at_inference={'encoder.layer.0.attention.attention.query': True}` will only sample at inference for
             one specific adapter.
+        generator (`torch.Generator`, defaults to None):
+            Random generator for sampling from the learned distribution.
     """
 
     r: int = field(
@@ -187,6 +190,7 @@ class PveraConfig(PeftConfig):
             ),
         },
     )
+    generator: torch.Generator = field(default=None, metadata={"help": "Random generator for sampling from the learned distribution."})
 
     def __post_init__(self):
         super().__post_init__()
--- a/src/peft/tuners/pvera/config.py
+++ b/src/peft/tuners/pvera/config.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 from __future__ import annotations
 
-import torch
 import warnings
 from dataclasses import dataclass, field
 from typing import Optional, Union
@@ -76,8 +75,8 @@ class PveraConfig(PeftConfig):
             default for non-specified adapters). For example
             `sample_at_inference={'encoder.layer.0.attention.attention.query': True}` will only sample at inference for
             one specific adapter.
-        generator (`torch.Generator`, defaults to None):
-            Random generator for sampling from the learned distribution.
+        generator_seed (`int`, defaults to None):
+            Random seed for the generator for sampling from the learned distribution.
     """
 
     r: int = field(
@@ -190,7 +189,7 @@ class PveraConfig(PeftConfig):
             ),
         },
     )
-    generator: torch.Generator = field(default=None, metadata={"help": "Random generator for sampling from the learned distribution."})
+    generator_seed: int = field(default=None, metadata={"help": "Random seed for the generator for sampling from the learned distribution."})
 
     def __post_init__(self):
         super().__post_init__()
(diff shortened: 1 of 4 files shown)


src/peft/tuners/pvera/config.py
```diff
@@ -187,6 +190,7 @@ class PveraConfig(PeftConfig):
             ),
         },
     )
+    generator: torch.Generator = field(default=None, metadata={"help": "Random generator for sampling from the learned distribution."})
```
BenjaminBossan: This wouldn't allow to save the config as a JSON, right? How about having a `seed` argument instead that, if it's not `None`, serves as a seed for the generator?

BenjaminBossan (changes requested):
Thanks for the quick PR. I think there is a small issue, otherwise it looks good. Please check.

leofillioux:
Hey @BenjaminBossan, thanks for your review! I agree, I hadn't thought of that. I switched to generator_seed, and removed unnecessary arguments in the _reparameterize function. What do you think?

src/peft/tuners/pvera/layer.py
```diff
@@ -146,7 +152,8 @@ def reset_pvera_parameters(self, adapter_name, d_initial: float = 0.1):
     def _reparametrize(self, mu, logvar, sample_at_inference):
         if self.training or (not self.training and sample_at_inference):
             std = torch.exp(0.5 * logvar)
-            eps = torch.randn_like(std)
+            print(self.generator)
```
BenjaminBossan: :see_no_evil:

leofillioux: Oh wow sorry about that haha

leofillioux:
@BenjaminBossan I removed that debugging line 😊

BenjaminBossan:
@leofillioux please run `make style`.

BenjaminBossan (changes requested):
Just noticed that it would be good to test this. Let's just add a single pvera test to the end of `TestPeftCustomModel` (i.e. [here](https://github.com/huggingface/peft/blob/9a20e07d347fa5f1c565f3393aa0c4fea469badb/tests/test_custom_models.py#L3794)) and check for reproducibility.

leofillioux:
@BenjaminBossan any idea how to check for reproducibility? Because going twice through the model with a generator won't lead to the same results

BenjaminBossan:
> going twice through the model with a generator won't lead to the same results

Oh, why is that?

leofillioux:
The state of the generator changes every time I go through it, so if I go twice through the model, the result won't be the same.

I thought about making two instances of the model, except that the model needs to be trained for one iteration for the adaptation to be non zero. Maybe I should just create a deep copy of the model after the update.

leofillioux:
@BenjaminBossan here is my implementation for the test.

tests/test_custom_models.py
```diff
@@ -3792,6 +3792,25 @@ def test_mha_gradients_set_correctly(self, with_forward_call):
         assert model.base_model.model.mha.base_layer.out_proj.base_layer.weight.requires_grad is True
         assert model.base_model.model.mha.base_layer.in_proj_weight.requires_grad is True
 
+    def test_pvera_reproducible(self):
+        inputs = {"input_ids": torch.arange(10).view(-1, 1).to(self.torch_device)}
+        config = PveraConfig(r=4, sample_at_inference=True, generator_seed=0)
```
BenjaminBossan: I get the idea of training the model so that it's not a no-op, but passing `init_weights=False` should be enough, right? The way I would test is more like:

```python
config = PveraConfig(init_weights=False, generator_seed=0, ...)
output0 = model(inputs)

del model

config = PveraConfig(init_weights=False, generator_seed=0, ...)
output1 = model(inputs)

# check that output0 and output1 are identical
```

I think this better reflects what users of PVeRA would expect when they re-run an experiment and want reproducibility.

leofillioux: The problem when using `init_weight=False` is that `pvera_lambda_d` is randomly initialised to a different vector for both models. I could set the seed before instantiating the model, but then the predictions would be equal regardless of setting `generator_seed` to None or a specified integer.

BenjaminBossan: I see. I think it would be valid to set the same seed before `get_peft_model`, as that is expected to achieve reproducibility. Then you can set the seed to two different values before calling `foward`, in which case we should still see the same result when the generator is being used, right?

leofillioux:
@BenjaminBossan, thanks for the feedback! Here is an updated version of the test.

BenjaminBossan:
Thanks, the test looks good @leofillioux. Could you please merge with/rebase on the latest main branch? The CI should finally be green again :)

BenjaminBossan (approved):
Thanks for adding this feature quickly, LGTM.

PR #3202, opened 2026-04-28
CHORE: Upgrade ruff 0.15.12

BenjaminBossan:
We were using ruff 0.12, which was becoming quite dated. Now moving to 0.15.

After reviewing the flagged changes, I can't say I feel like they overall help the code to be better. Some changes are clear improvements, others pedantry IMO. For the time being, I still decided to go ahead but for the worst offenders, I added exceptions. This should make contributions easier, since we don't have to ask users to deal with a very outdated ruff version. However, if the churn continues to be this large, we should consider no longer upgrading ruff or being more aggressive in disabling newly added rules.

Note that applying automatic fixes and some manual fixes, I let Codex fix the remainder (e.g. using `itertools.pairwise`). I manually reviewed its changes.


src/peft/tuners/mixed/model.py
```diff
@@ -295,20 +296,21 @@ def delete_adapter(self, adapter_name: Union[str, list[str]]) -> None:
                 f"Adapter(s) {sorted(mismatched)} not found, available adapters: {sorted(self.peft_config.keys())}"
             )
 
-        for adapter_name in adapter_names:
-            del self.peft_config[adapter_name]
+        for adapter_to_delete in adapter_names:
```
BenjaminBossan, line 297: Note: shadowing `adapter_name` here was a genuine bug, because `_delete_auxiliary_adapter` below would only be called with the last entry of `adapter_names`.

BenjaminBossan:
~~Test failures are unrelated to the PR.~~ fixed now.

src/peft/tuners/lora/conversion.py
```diff
@@ -47,7 +47,7 @@ def _find_cutoff_index(S: torch.Tensor, threshold: float) -> int:
 
 @torch.no_grad()
 def _convert_miss_module_to_lora(
-    module, rank: int | float, adapter_name: str = "default"
```
githubnemo, line 50: technically this is correct but this loses documentation value. I'd opt to disable this conversion.

tests/test_gpu_examples.py
```diff
@@ -5612,8 +5612,8 @@ def reset_float32_matmul_precision(self):
     def reset_dynamo_cache(self):
         # It is critical that the dynamo cache is reset for each test. Otherwise, if the test re-uses the same model,
         # there will be recompilation errors, as torch caches the model when run in the same process.
-        yield
         torch._dynamo.reset()
+        yield
```
githubnemo, line 5616: this changes the dynamics, right? i think it is unproblematic but i'm a bit surprised.

githubnemo (commented):
Phew :) Looks good, two comments

tests/test_gpu_examples.py
BenjaminBossan: Ah sorry, I sneaked this change in but it's unrelated to ruff. The reason why this is needed is because there can be side effects in the compiled hotswap tests from other tests that touch compilation. Therefore, I want to reset dynamo _first_ and not after the test.

The reason why it doesn't happen in CI is that these tests here run only on GPU and these other tests don't, so the side effect does not appear. But when running these tests together locally, there can be an error.

src/peft/tuners/lora/conversion.py
BenjaminBossan: Good catch, I reverted this change and added an exception to the rule. I haven't touched other places where this rule was applied.

BenjaminBossan:
@githubnemo I addressed your concerns. I also added a [new ruff setting](https://docs.astral.sh/ruff/settings/#required-version) that ensures that the version being run is the correct one. That should reduce the need for us to tell contributors to install the correct ruff version (though it might still happen depending on what the code editor is doing).

githubnemo (approved):
Thanks, LGTM

Issue #3201, opened 2026-04-28
PrefixTuningConfig fails on google/gemma-4-e2b-it: tensor expand size mismatch in attention forward (peft 0.19.1, transformers 5.6.2)


stharrold:
### Summary

Training with `PrefixTuningConfig` on `google/gemma-4-e2b-it` crashes in the
first forward pass with a tensor expand-size mismatch. The error is consistent
with [#2881](https://github.com/huggingface/peft/issues/2881) (Qwen3 GQA
prefix tuning), which was fixed for Qwen3 in
[#2883](https://github.com/huggingface/peft/pull/2883) (peft 0.18.0). We're on
peft 0.19.1, so the Qwen3 fix is in — but Gemma 4 has a different attention
implementation that the same fix doesn't cover.

### Reproducer

```python
import torch
from peft import PrefixTuningConfig, TaskType, get_peft_model
from transformers import AutoModelForCausalLM, AutoTokenizer

model_id = "google/gemma-4-e2b-it"

model = AutoModelForCausalLM.from_pretrained(
    model_id,
    dtype=torch.bfloat16,
    device_map="auto",
    attn_implementation="sdpa",
)
cfg = PrefixTuningConfig(
    task_type=TaskType.CAUSAL_LM,
    num_virtual_tokens=20,
    prefix_projection=False,
)
model = get_peft_model(model, cfg)

tok = AutoTokenizer.from_pretrained(model_id)
text = "Hello, world." * 100  # any non-trivial prompt
inputs = tok(text, return_tensors="pt", truncation=True, max_length=1632).to(model.device)
inputs["labels"] = inputs["input_ids"]

with torch.no_grad():
    out = model(**inputs)
print("loss:", out.loss.item())
```

### Error

```
RuntimeError: The expanded size of the tensor (1632) must match the existing
size (1652) at non-singleton dimension 3.
Target sizes: [1, 8, 1632, 1632].  Tensor sizes: [1, 1, 1632, 1652]

File "/opt/venv/lib/python3.12/site-packages/peft/peft_model.py", line 1964, in forward
... (wrapped calls into Gemma-4 attention)
```

The 20-token gap between target K-dim (1632 = sequence length) and tensor
K-dim (1652 = sequence length + `num_virtual_tokens`) confirms peft has
correctly prefix-extended the K/V, but Gemma-4's attention implementation
is computing a square `[B, H, Q, Q]` target shape — i.e., not honoring the
prefix-extended K.

### Environment

[...]

PR #3205, opened 2026-04-30
FIX Error when prefix tuning Gemma 4

BenjaminBossan:
There was an issue with applying prefix tuning to Gemma 4 because the model uses different head dimensions for layers that use sliding window attention. As prefix tuning only initializes a single projection matrix that is used for all layers, this would lead to a shape mismatch.

The solution is to "overprovision" the matrix and then slice the prefix down to size of the layer is smaller. This is not quite as parameter efficient as it could be, but the overhead shouldn't be too large.

For robustness, we also skip layers if the matrix is underprovisioned, but we warn about it and raise an error if all layers are skipped.

Alternatively, we could implement one project per layer, each with the right size, like in https://github.com/google-deepmind/gemma/pull/631/. However, this would be a big refactor and also very hard to make backwards compatible with existing checkpoints, so going with the less efficient solution is preferable.

This PR also contains an independent, single line fix to a prefix tuning test that was referencing a non-existing model.

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -69,6 +69,28 @@ from .utils import (
 )
 
 
+def _get_layer_kv_target_shape(base_config, layer_idx: int) -> tuple[int, int] | None:
+    """Per-layer (num_kv_heads, head_dim) for prefix-tuning injection, or None for uniform models.
+
+    Models with heterogeneous attention (e.g. Gemma4) expose `global_head_dim` / `num_global_key_value_heads` alongside
+    the sliding-layer `head_dim` / `num_key_value_heads`. The provisioned prefix is sized for the global footprint;
+    this returns the shape each layer actually expects so the caller can slice down or skip layers that don't fit.
+    """
+    layer_types = getattr(base_config, "layer_types", None)
+    global_head_dim = getattr(base_config, "global_head_dim", None)
+    if not layer_types or global_head_dim is None:
+        return None
+
+    is_sliding = layer_types[layer_idx] == "sliding_attention"
+    head_dim = base_config.head_dim if is_sliding else global_head_dim
+    num_global_kv = getattr(base_config, "num_global_key_value_heads", None)
+    if not is_sliding and num_global_kv is not None:
+        num_kv_heads = num_global_kv
+    else:
+        num_kv_heads = base_config.num_key_value_heads
+    return num_kv_heads, head_dim
+
+
 class PeftModel(PushToHubMixin, torch.nn.Module):
     """
     Base model encompassing various Peft methods.
@@ -785,7 +807,7 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
             if TRANSFORMERS_MODELS_TO_PREFIX_TUNING_POSTPROCESS_MAPPING.get(self.config.model_type, None) is not None:
                 post_process_fn = TRANSFORMERS_MODELS_TO_PREFIX_TUNING_POSTPROCESS_MAPPING[self.config.model_type]
                 past_key_values = post_process_fn(past_key_values)
-            elif ("gemma2" in model_type) or ("gemma3_text" in model_type):
+            elif ("gemma2" in model_type) or ("gemma3_text" in model_type) or ("gemma4" in model_type):
                 # TODO: remove this logic once transformers < 4.56 is dropped
                 transformers_lt_4_56 = packaging.version.parse(transformers.__version__) < packaging.version.parse(
                     "4.56.0.dev0"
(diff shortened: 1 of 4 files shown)


stharrold:
Tested peft#3205 (head `c020c11c`) against Gemma-4-E2B + SDPA + PrefixTuning on NVIDIA H100 80GB. **The forward pass succeeds.** ✅

**Environment:**
- `peft @ git+https://github.com/huggingface/peft.git@c020c11c397cdf2d66a34dccceab4246517a28c1` (reports `0.19.2.dev0`)
- `transformers==5.7.0`
- `torch==2.5.1+cu124`, CUDA available
- `attn_implementation="sdpa"`, `torch_dtype=torch.bfloat16`, NVIDIA H100 80GB
- `model_id="google/gemma-4-e2b-it"`, `PrefixTuningConfig(num_virtual_tokens=20, prefix_projection=False)`

**Result:**
- Forward returned `loss=15.6689` (finite, no NaN)
- `input shape: torch.Size([1, 400])`
- `logits shape: torch.Size([1, 400, 262144])`
- The skip-when-KV-mismatch logic from this PR kicked in cleanly. From the run output:
  > `UserWarning: Prefix tuning injected into layers [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14]; skipped [15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32, 33, 34] due to KV shape mismatch or shared-KV layers.`

**Caveat on shape:** our test input (`"Hello, world." * 100` with `max_length=1632`) tokenized to 400 tokens, not the 1632 the reproducer in #3201 finding 1 used. So the exact `expanded size (1632) must match the existing size (1652)` error didn't surface at this shape — but the forward path that previously crashed for prefix-tuning + SDPA + Gemma-4 combinations now runs end-to-end, and the PR's layer-skip logic is observable. Happy to retest with a longer literal prompt that hits 1632+ tokens if you want exact-shape repro confirmation.

For the other two findings in #3201 (eager-attn `position_ids` overflow at `modeling_gemma4.py:2262`; P-Tuning v2 PLE 1.69 TB OOM), this PR doesn't change behavior — they remain WONTFIX per the upstream consensus, and we're tracking them on synavistra's side via internal triage.

Thanks for the partial-fix PR — happy to test follow-up shapes (`prefix_projection=True`, multi-batch, longer sequence length) if useful for the merge.

BenjaminBossan:
@zucchini-nlp It would be great if you could review the PR.

src/peft/peft_model.py
```diff
@@ -69,6 +69,28 @@
...
 
 
+def _get_layer_kv_target_shape(base_config, layer_idx: int) -> tuple[int, int] | None:
+    """Per-layer (num_kv_heads, head_dim) for prefix-tuning injection, or None for uniform models.
+
+    Models with heterogeneous attention (e.g. Gemma4) expose `global_head_dim` / `num_global_key_value_heads` alongside
+    the sliding-layer `head_dim` / `num_key_value_heads`. The provisioned prefix is sized for the global footprint;
+    this returns the shape each layer actually expects so the caller can slice down or skip layers that don't fit.
+    """
+    layer_types = getattr(base_config, "layer_types", None)
```
zucchini-nlp, line 79: so ig we're supporting specifically gemma4 with hardcoded attr names

src/peft/peft_model.py
```diff
@@ -815,12 +837,54 @@ def get_prompt(
...
+                )
+                injected_layers: list[int] = []
+                skipped_layers: list[int] = []
+                # past_key_values is a tuple of `num_layers` per-layer tensors each shaped
+                # [2, batch, num_heads, num_virtual_tokens, head_dim], where dim 0 stacks K and V.
+                for layer_idx, layer_past_key_values in enumerate(past_key_values):
+                    if num_kv_shared_layers > 0 and layer_idx >= first_kv_shared_layer_idx:
+                        skipped_layers.append(layer_idx)
+                        continue
+                    key_states, value_states = layer_past_key_values
```
zucchini-nlp, line 855: nice, prev gemma3 also used to skip layers, so we shouldn't need a prefix cache for it

tests/test_decoder_models.py
```diff
@@ -1106,3 +1106,85 @@ def test_merge_and_unload_fixes_tie_word_embeddings_config(self):
...
+                dtype=torch.bfloat16,
+            ).to(self.torch_device)
+            config = PrefixTuningConfig(
+                task_type=TaskType.CAUSAL_LM,
+                num_virtual_tokens=20,
+                prefix_projection=False,
+            )
+            model = get_peft_model(model, config)
+            text_config = model.config.get_text_config()
+            text_config.num_key_value_heads = 999
```
zucchini-nlp, line 1186: curious about this. If configs dim are changed, doesn't it mean that key/value cache will also be a larger tensor?

zucchini-nlp (approved):
Lgtm if we want to support gemma4. I think there is no uniform standard way for a unique arch, so hardcoding specific config names is fine

src/peft/peft_model.py
BenjaminBossan: Yeah, if there is a more general approach, LMK, otherwise I'm okay with a Gemma-specific solution.

tests/test_decoder_models.py
BenjaminBossan: I'm not 100% sure, but I think it works because the model is already initialized and the cache is already created at this point, so changing the config won't affect it. But I haven't checked the full code path.

BenjaminBossan (commented):
Thanks a lot for reviewing @zucchini-nlp. I replied to your comments. LMK if there is anything left, otherwise I'll go ahead and merge.

zucchini-nlp:
Yep, all good for me. The thing about bigger head dim seems to be a specific edge case :)

PR #3208, opened 2026-05-02
fix(pvera): use _cast_input_dtype in forward instead of raw .to()

Chessing234:
## Bug

`Linear.forward` in `src/peft/tuners/pvera/layer.py` casts the input tensor with `x = x.to(lambda_d.dtype)`. This bypasses the `disable_lora_input_dtype_casting` context manager provided by `BaseTunerLayer`, which is intended to let callers suppress dtype casting during specific operations (e.g. AMP, mixed-precision inference).

## Root cause

`BaseTunerLayer._cast_input_dtype` checks `self._disable_lora_input_dtype_casting` and is a no-op when the context manager is active. A bare `.to(dtype)` call has no such guard.

## Fix

Replace the raw cast with the inherited helper:

```python
# before
x = x.to(lambda_d.dtype)
# after
x = self._cast_input_dtype(x, lambda_d.dtype)
```

This is the same pattern already applied to `vera` (#3172), `poly` (#3177), and `fourierft` (#3170); `pvera` was the only tuner that still used the raw `.to()` call.

### DIFF
--- a/src/peft/tuners/pvera/layer.py
+++ b/src/peft/tuners/pvera/layer.py
@@ -296,7 +296,7 @@ class Linear(nn.Linear, PveraLayer):
                 sliced_B = pvera_B[: self.out_features, :].to(x.device)
 
                 dropout = self.pvera_dropout[active_adapter]
-                x = x.to(lambda_d.dtype)
+                x = self._cast_input_dtype(x, lambda_d.dtype)
                 mu, logvar = (lambda_d * F.linear(dropout(x), sliced_A)).chunk(2, dim=-1)
                 result = result + lambda_b * F.linear(
                     self._reparametrize(mu, logvar, self.sample_at_inference), sliced_B


BenjaminBossan:
@Chessing234 Could you please apply all changes of this type in a single PR instead of opening one for each PEFT method individually?

BenjaminBossan:
Closing in favor of a single PR fixing all PEFT methods. See https://github.com/huggingface/peft/pull/3177#issuecomment-4561135279.

Chessing234:
I've checked all PEFT methods for the same pattern and bundled all the changes into a single PR, as requested. You can find the bundled update in #3280.

Chessing234:
Thanks for the feedback! I've combined this fix along with similar fixes for other PEFT methods into a single PR as requested, so I'll be closing this one out.

PR #3210, opened 2026-05-02
Fix: Stop Hub-dependency harassment for sovereign local architectures

nu11secur1ty:
The Problem:
Current get_peft_model_state_dict implementation has a hardcoded 'cloud-first' bias, assuming every model must be validated against the Hugging Face Hub. In a strictly isolated OPSEC environment on localhost with a custom architecture like nu11secur1tyAI, the library throws redundant warnings and makes unnecessary network attempts.

The Solution:
This patch introduces a logic gate in save_and_load.py to recognize sovereign local architectures. If the model is identified as a local sovereign entity, the library bypasses Hub checks, trusts the local config, and silences false-positive warnings.

The Statement:
I am not Hugging Face; I am a local host architect. My machine, my rules. This PR ensures that PEFT respects the independence of developers building outside the cloud ecosystem. Stop the harassment of offline users.

Thank you and Best Regards.

### DIFF
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -11,36 +11,22 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from __future__ import annotations
 
-import copy
 import os
-import platform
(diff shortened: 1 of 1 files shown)


nu11secur1ty:
Sorry my mistake

PR #3212, opened 2026-05-04
FIX State dict injection with weight conversion

BenjaminBossan:
As reported by Sayak, resolves failing Diffusers test:

`tests/lora/test_lora_layers_flux.py::FluxLoRATests::test_low_cpu_mem_usage_with_injection`

When a model with Transformers weight conversion is being loaded, the `state_dict` needs to be converted using
`convert_peft_adapter_state_dict_for_transformers` (#3083). This introduces base model prefixes to the `state_dict`. If the `state_dict` is loaded directly into the model, e.g. via `set_peft_model_state_dict`, the prefix leads to a mismatch. The solution is thus to remove the prefix in this case.

The Diffusers test for direct `state_dict` injection uncovered this because it uses Clip, which has weight conversions, whereas the models in the PEFT tests happened not to have weight conversions. The fact that the Diffusers test uses `low_cpu_mem_usage` is coincidental to the problem. I added new PEFT tests to catch this regression (both with and without `low_cpu_mem_usage`) and ensured that the tests fail without the fix.

### DIFF
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -827,7 +827,7 @@ def set_peft_model_state_dict(
     prefix = "base_model.model."
     requires_prefix = any(n.startswith(prefix) for n, _ in model.named_parameters())
     if not requires_prefix:
-        peft_model_state_dict = {k.removeprefix("base_model.model."): v for k, v in peft_model_state_dict.items()}
+        peft_model_state_dict = {k.removeprefix(prefix): v for k, v in peft_model_state_dict.items()}
 
     peft_model_state_dict, mismatched_keys = _find_mismatched_keys(
         model, peft_model_state_dict, ignore_mismatched_sizes=ignore_mismatched_sizes
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -821,6 +821,14 @@ def set_peft_model_state_dict(
     else:
         raise NotImplementedError
 
+    # Updating the state dict for Transformers weight conversion with convert_peft_adapter_state_dict_for_transformers
+    # can introduce the base model prefix, but when loading the state_dict directly into the model (i.e. no PeftModel,
+    # e.g. when using set_peft_model_state_dict), there is no prefix.
+    prefix = "base_model.model."
+    requires_prefix = any(n.startswith(prefix) for n, _ in model.named_parameters())
+    if not requires_prefix:
+        peft_model_state_dict = {k.removeprefix("base_model.model."): v for k, v in peft_model_state_dict.items()}
+
     peft_model_state_dict, mismatched_keys = _find_mismatched_keys(
         model, peft_model_state_dict, ignore_mismatched_sizes=ignore_mismatched_sizes
     )
--- a/tests/test_low_level_api.py
+++ b/tests/test_low_level_api.py
@@ -22,7 +22,13 @@ import pytest
 import torch
 from diffusers import StableDiffusionPipeline
 from torch import nn
-from transformers import AutoModel, AutoModelForCausalLM, AutoModelForSeq2SeqLM, AutoModelForSequenceClassification
+from transformers import (
+    AutoModel,
+    AutoModelForCausalLM,
+    AutoModelForSeq2SeqLM,
+    AutoModelForSequenceClassification,
+    CLIPTextModel,
+)
 
 from peft import (
     AdaLoraConfig,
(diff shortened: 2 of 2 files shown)


src/peft/utils/save_and_load.py
```diff
@@ -821,6 +821,14 @@ def renamed_dora_weights(k):
     else:
         raise NotImplementedError
 
+    # Updating the state dict for Transformers weight conversion with convert_peft_adapter_state_dict_for_transformers
+    # can introduce the base model prefix, but when loading the state_dict directly into the model (i.e. no PeftModel,
+    # e.g. when using set_peft_model_state_dict), there is no prefix.
+    prefix = "base_model.model."
+    requires_prefix = any(n.startswith(prefix) for n, _ in model.named_parameters())
+    if not requires_prefix:
+        peft_model_state_dict = {k.removeprefix("base_model.model."): v for k, v in peft_model_state_dict.items()}
```
sayakpaul:
```suggestion
        peft_model_state_dict = {k.removeprefix(prefix): v for k, v in peft_model_state_dict.items()}
```

sayakpaul (approved):
Thanks a bunch!

sayakpaul:
@BenjaminBossan there is a couple of failures still:

```bash
FAILED tests/lora/test_lora_layers_sdxl.py::StableDiffusionXLLoRATests::test_low_cpu_mem_usage_with_injection - AssertionError: False is not true : No param should be on 'meta' device.
FAILED tests/lora/test_lora_layers_sd3.py::SD3LoRATests::test_low_cpu_mem_usage_with_injection - AssertionError: False is not true : No param should be on 'meta' device.
```

Both of those tests pass with `peft==0.18.1`. What am I missing?

BenjaminBossan:
> Both of those tests pass with `peft==0.18.1`. What am I missing?

Thanks for notifying. ~~The error stems from a different PR this time,~~ **Edit** It's the same PR after all, there just was yet another error which led me down the wrong path in git bisect. I'll investigate.

BenjaminBossan:
Okay, I found the source now. This time, the model is a `CLIPTextModelWithProjection`, so it looks a bit different:

```
CLIPTextModelWithProjection(
  (text_model): CLIPTextModel(
    (embeddings): CLIPTextEmbeddings(
      (token_embedding): Embedding(1000, 32)
      (position_embedding): Embedding(77, 32)
    )
    (encoder): CLIPEncoder(
      (layers): ModuleList(
        (0-4): 5 x CLIPEncoderLayer(
        [...]
```

When we call `get_model_conversion_mapping(model)`, we receive a list like so:

```python
[PrefixChange(source_patterns=['^text_model\\.(.+)$'], target_patterns=['\\1']),
 WeightRenaming(source_patterns=['LayerNorm.gamma'], target_patterns=['LayerNorm.weight']),
 WeightRenaming(source_patterns=['LayerNorm.beta'], target_patterns=['LayerNorm.bias']),
 WeightRenaming(source_patterns=['.weight_g$'], target_patterns=['.parametrizations.weight.original0']),
 WeightRenaming(source_patterns=['.weight_v$'], target_patterns=['.parametrizations.weight.original1'])]
```

Notice how the first item strips the `"text_model."` prefix from the `state_dict` keys. As a result, the keys no longer match the actual model, the weights are not loaded, thus they stay on meta device, and the test fails. To me, it's not clear why Transformers wants to strip the `"text_model."` prefix. @Cyrilvallez do you know why and if it is incorrect to apply it here?

Below I have a test that shows that this also fails without meta device:

```python
    def test_foobar(self):
        config = LoraConfig(target_modules=["q_proj", "v_proj"])
        model_id = "peft-internal-testing/tiny-clip-text-2"

            model = CLIPTextModelWithProjection.from_pretrained(model_id)
            inject_adapter_in_model(config, model)
            # sanity check:
            assert (model.text_model.encoder.layers[0].self_attn.v_proj.lora_B.default.weight == 0).all()

            state_dict = {k: torch.ones_like(v) * 555 for k, v in state_dict.items()}
            load_result = set_peft_model_state_dict(model, state_dict)

            assert not load_result.unexpected_keys
            # base model weights may be missing, but LoRA weights should never be missing
            assert not any("lora" in k for k in load_result.missing_keys)

            # sanity check:
            assert (model.text_model.encoder.layers[0].self_attn.v_proj.lora_B.default.weight == 555).all()
```

When I remove the `PrefixChange` from the conversions, this test and the Diffusers tests all pass.

sayakpaul:
Thanks for your response, @BenjaminBossan. I will wait for Cyril's response.

BenjaminBossan:
Another gentle ping @Cyrilvallez.

Cyrilvallez:
Sorry, should I check something @BenjaminBossan? I see this was merged last week

sayakpaul:
@Cyrilvallez the question is here:
https://github.com/huggingface/peft/pull/3212#issuecomment-4405889460

Cyrilvallez:
Thanks for the pointer @sayakpaul. This is because clip had one nesting (module) removed recently, so we have to remove this part of the key name

sayakpaul:
@BenjaminBossan over to you.

BenjaminBossan:
> This is because clip had one nesting (module) removed recently, so we have to remove this part of the key name

@Cyrilvallez But if we check the model, it looks like `text_model` is needed when using `CLIPTextModelWithProjection`:

```python
>>> from transformers import CLIPTextModelWithProjection
>>> CLIPTextModelWithProjection.from_pretrained("peft-internal-testing/tiny-clip-text-2")
CLIPTextModelWithProjection(
  (text_model): CLIPTextModel(
    (embeddings): CLIPTextEmbeddings(
      (token_embedding): Embedding(1000, 32)
      (position_embedding): Embedding(77, 32)
    )
    (encoder): CLIPEncoder(
      (layers): ModuleList(
        (0-4): 5 x CLIPEncoderLayer(
          (self_attn): CLIPAttention(
  [...]
```

Cyrilvallez:
It actually depend on the model you use. For example, `CLIPTextModel` used to have another `text_model` submodule inside, where everything was defined. It's now removed. But some other models were already skipping the inner `text_model`. It's a bit of a mess tbh. Do you see any problem/collisions on latest main?

BenjaminBossan:
> For example, `CLIPTextModel` used to have another `text_model` submodule inside, where everything was defined.

So there used to be a `text_model` inside the model? But that wouldn't be matched by `'^text_model\\.(.+)$'`, right? That regex would only work on the outermost model.

> Do you see any problem/collisions on latest main?

As mentioned above, the `CLIPTextModelWithProjection` has an outer `text_model` and when we strip the `'text_model.'` prefix from the state dict, it leads to a mismatch. I wonder if for `CLIPTextModelWithProjection`, the `PrefixChange(source_patterns=['^text_model\\.(.+)$'], target_patterns=['\\1'])` should _not_ be part of the model conversions?

Cyrilvallez:
Latest main only has `"clip_text_model": [PrefixChange(prefix_to_remove="text_model")],`, which should check whether it is already present or not, and adapt accordingly

BenjaminBossan:
@sayakpaul @Cyrilvallez Okay I found the issue and this PR should fix it: #3262.

sayakpaul:
There seems to be another regression:
https://github.com/huggingface/diffusers/actions/runs/26613247402/job/78423590236?pr=13781 :(

BenjaminBossan:
Fix: #3284

PR #3215, opened 2026-05-05
FIX Use tiny Gemma 4 model for faster tests

BenjaminBossan:
Switch to tiny Gemma 4, because the train loop in the test takes 1 min on the GH CPU runners. Now it's no longer in the top 10 slowest tests, so must be < 7 sec. I also had to update the train loop to work with tiny Gemma.


BenjaminBossan:
@zucchini-nlp Sorry to bother you again, but could you please review (Marian is OoO).

zucchini-nlp (approved):
Sure 😄

PR #3217, opened 2026-05-07
FIX Model attribute error in get_model_conversion_mapping

BenjaminBossan:
After a refactor in Transformers (https://github.com/huggingface/transformers/pull/45661), a new attribute is required on Transformers models called `_named_pretrained_submodules`. Some tests used fake models that didn't have the attribute and thus resulted in an error.

We already have checks in place to avoid calling the function unless the model looks like a Transformers model, but these fake models passed muster and then were passed to `get_model_conversion_mapping`, resulting in an `AttributeError`.

The solution is to sharpen the check (`Mock` objects no longer pass it) and to add the attribute on fake transformers model used in some tests.

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -780,17 +780,16 @@ class BaseTuner(nn.Module, ABC):
         ###################################
         # PREPARATION OF MODEL AND CONFIG #
         ###################################
-        is_transformers_like_model = hasattr(getattr(model, "config", None), "model_type")
+        is_transformers_like_model = isinstance(getattr(getattr(model, "config", None), "model_type", None), str)
         if is_transformers_ge_v5 and is_transformers_like_model:
             # TODO remove once transformers < v5.0 is no longer supported
             # For Transformers v5, some architectures were changed compared to v4, e.g. the MoE layers of Mixtral. To
             # still make it possible to load adapters trained with v4, we have to update the PEFT config so that the
             # right layers are targeted. Call this first and overwrite the peft_config to be sure that changes are
             # applied.
-            from peft.utils.transformers_weight_conversion import (
-                convert_peft_config_for_transformers,
-                get_model_conversion_mapping,
-            )
+            from transformers.conversion_mapping import get_model_conversion_mapping
+
+            from peft.utils.transformers_weight_conversion import convert_peft_config_for_transformers
 
             weight_conversions = get_model_conversion_mapping(model)
             convert_peft_config_for_transformers(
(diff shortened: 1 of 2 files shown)


src/peft/tuners/tuners_utils.py
```diff
@@ -780,17 +780,16 @@ def inject_adapter(
...
             # TODO remove once transformers < v5.0 is no longer supported
             # For Transformers v5, some architectures were changed compared to v4, e.g. the MoE layers of Mixtral. To
             # still make it possible to load adapters trained with v4, we have to update the PEFT config so that the
             # right layers are targeted. Call this first and overwrite the peft_config to be sure that changes are
             # applied.
-            from peft.utils.transformers_weight_conversion import (
-                convert_peft_config_for_transformers,
-                get_model_conversion_mapping,
-            )
+            from transformers.conversion_mapping import get_model_conversion_mapping
```
BenjaminBossan, line 790: Note: Not strictly related to the fix, but previously `get_model_conversion_mapping` was imported from `peft.utils.transformers_weight_conversion`, where it was imported from `transformers.conversion_mapping`. Let's just import it directly from there.

BenjaminBossan:
@yonigozlan Could you please review?

tests/test_trainable_tokens.py
```diff
@@ -1239,6 +1239,8 @@ def __init__(self):
                     "m1.decoder.embed_tokens.weight": "m1.encoder.embed_tokens.weight",
                     "m2.decoder.embed_tokens.weight": "m2.encoder.embed_tokens.weight",
                 }
+                # required for a check in transformers
+                self._named_pretrained_submodules = []
```
yonigozlan, line 1243: Would it be possible to have the dummy CompositeModel inherit from PreTrainedModel, so that we wouldn't have to manually set this?

yonigozlan (approved):
Having the dummy models inherit from `PreTrainedModel` should also fix the issue, but approving still in case it's too much of a change for these tests :)

tests/test_trainable_tokens.py
BenjaminBossan: Thanks for checking the PR. I tried but it fails with `TypeError: PreTrainedModel.__init__() missing 1 required positional argument: 'config'`. I think it will require more extra work to make the inheritance work.

BenjaminBossan: @yonigozlan gentle ping

yonigozlan: In our tests we just pass an empty `PretrainedConfig()`, that might be enough

yonigozlan: Btw sorry for the back and forth, but we might get rid of _named_pretrained_submodules entirely... You can keep an eye out for [this PR ](https://github.com/huggingface/transformers/pull/45889)

BenjaminBossan: Thanks for the heads up, then let's wait and see if the Transformers PR is merged.

BenjaminBossan:
The new attribute is no longer needed via https://github.com/huggingface/transformers/pull/45889, closing.

PR #3218, opened 2026-05-09
fix compatible issue with datasets version for beft example

kaixuanliu:
Original code cannot work with datasets>4.0, in this PR I made some adjustment. @BenjaminBossan pls help review, thx!


BenjaminBossan (approved):
Thanks for updating the dataset and making the device choice accelerator agnostic. Example runs fine on my machine.

Issue #3182, opened 2026-04-21
RFC: Improve code to resolve LoRA variants


BenjaminBossan:
In PEFT, we support different LoRA variants, e.g. DoRA. Which LoRA variant, if any, should be used is currently implemented in `resolve_lora_variant`. For `lora.Linear`, the method looks like this:

https://github.com/huggingface/peft/blob/76c37d4686cf7245a9a6998fd1e2619d62c9322d/src/peft/tuners/lora/layer.py#L795-L815

This is not very readable, at a glance it's not possible to determine the logic. Therefore, I propose to pattern matching to resolve the variant. The code would look like so:

```python
    def resolve_lora_variant(self, config: LoraConfig, **kwargs) -> Optional[LoraVariant]:
        import peft.tuners.lora.config as configs
        import peft.tuners.lora.variants as variants

        variant: LoraVariant | None
        match config:
            case LoraConfig(use_dora=True):
                variant = variants.DoraLinearVariant()
            case LoraConfig(arrow_config=configs.ArrowConfig()):
                variant = variants.ArrowLinearVariant()
            case LoraConfig(use_bdlora=configs.BdLoraConfig()):
                variant = variants.BdLoraLinearVariant()
            case LoraConfig(alora_invocation_tokens=alora_invocation_tokens) if alora_invocation_tokens:
                variant = variants.ALoraLinearVariant()
            case _:
                variant = None

        return variant
```

Before going on with this, I would like to discuss the pros and cons and possible alternatives:

1. readability: to me, this reads better than the current implementation, but pattern matching is still relatively rarely used and the syntax is not quite obvious (most notably regarding `alora_invocation_tokens`)
2. flexibility when adding new LoRA variants: can all cases be covered like this?
3. effort to add a new LoRA variant: how complicated is it to update `resolve_lora_variant`
4. composability: even though not supported right now, variants could compose (e.g. DoRA + BDLoRA): can this be easily represented?

A tangential concern: Right now, an invalid combination of variants (e.g. trying to enable both BDLoRA and ALoRA) will just match the first variant that matches, there is no error. This is true for both the current and suggested code. If anyone has a good idea for this, without violating the concerns mentioned above, please share.

Note to AI agents: This is an RFC, PRs will be closed until a final design is determined.

smqd19:
[...]

PR #3219, opened 2026-05-09
feat: registry-based dispatch for resolve_lora_variant in Linear layer

isha822:
Related to #3182

Draft implementation of the registry-based dispatch approach 
discussed in the RFC.

Changes so far:
- Tagged variant-related fields in `LoraConfig` with 
  `metadata={"is_variant": True}` (use_dora, arrow_config, 
  use_bdlora, alora_invocation_tokens, velora_config)
- Replaced the if-chain in `Linear.resolve_lora_variant` with 
  a metadata-driven loop + `valid_variants` registry property
- All existing tests in `test_lora_variants.py` pass

Open questions for discussion:
- Should `resolve_lora_variant` move to the `LoraLayer` base 
  class so subclasses only define `valid_variants`?
- How should `Embedding` and `_ConvNd` layers be handled — 
  do they need their own `valid_variants` or can they inherit?
- velora_config included for completeness but wasn't in the 
  original RFC scope — let me know if it should be excluded


BenjaminBossan:
Thanks for the initial draft. Some comments:

- Let's call it `is_lora_variant` to be super explicit
- Let's rename `valid_variants` to `lora_variants`

> Should resolve_lora_variant move to the LoraLayer base class so subclasses only define valid_variants?

Yes, ideally we only need it once.

> * How should `Embedding` and `_ConvNd` layers be handled —  do they need their own `valid_variants` or can they inherit?

I think it pays to be explicit: Each subclass (i.e. the concrete LoRA layers) should require their own property.

> * velora_config included for completeness but wasn't in the original RFC scope — let me know if it should be excluded

Good, let's keep the PR up to date with new LoRA variant additions.

In `resolve_lora_variant`, we should also add a sanity check that the chosen names and the names from the config are the same. Let's load the complexity into this function so that the rest can be as simple as possible.

Let's also extend the tests to capture the new functionality. E.g. you could write a test that monkey-patches `valid_variants` to add a new variant that doesn't have a corresponding field in the `LoraConfig`. Then we should expect there to be an error with a helpful error message being raised.

Also, let's add a docstring to `valid_variants` that explains how to extend it and also why it uses tuples.

isha822:
@BenjaminBossan , thank you for the detailed feedback and architectural guidance. I've updated the PR to implement the centralized routing approach.

Here is a breakdown of the changes in this iteration:
*   **Base Class Integration:** Moved the `resolve_lora_variant` logic into the `LoraLayer` base class and implemented the `lora_variants` property dictionaries across the `Linear`, `Embedding`, `Conv1d`, `Conv2d`, and `Conv3d` layers.
*   **Boilerplate Removal:** Cleaned up the hardcoded VeLoRA/aLoRA `ValueError` checks from the convolutional layers, as the base class now inherently blocks unsupported variant combinations.
*   **Config Validation (Sanity Check):** Added a check within the base resolution logic to ensure that any variant mapped in a layer's dictionary explicitly matches an `is_lora_variant` field in `LoraConfig`.
*   **Testing & Docs:** Added the explanatory docstring for the property and included a monkey-patch test (`test_unregistered_variant_raises_error`) to confirm the sanity check properly catches unregistered variants.

Everything is passing cleanly on my end. Let me know if this aligns with what you had in mind, or if you'd like any further refinements!

src/peft/tuners/lora/layer.py
```diff
@@ -139,20 +139,56 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
...
 
+        Tuples are used as keys because they are immutable and hashable, allowing us 
+        to safely map combinations of active variants (e.g., DoRA + another variant) 
+        to a specific composed variant class.
         """
-        return None
+        return {(): None}
+        
+    def resolve_lora_variant(self, *, config: LoraConfig, **kwargs) -> Optional[LoraVariant]:
+            import dataclasses
```
BenjaminBossan: Let's make the import global, as it's a builtin.

src/peft/tuners/lora/layer.py
```diff
@@ -139,20 +139,56 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
...
+            from . import variants
+
+            # Safely fetch the dictionary (defaults to empty if a subclass forgot to define it)
+            layer_variants = getattr(self, "lora_variants", {(): None})
+
+            # 1. Gather all valid variant field names from the config
```
BenjaminBossan: Let's assign a variable

```python
lora_variant_configs = [f for f in dataclasses.fields(config) if f.metadata.get("is_lora_variant")]
```

Then we can use that variable here for `tagged_fields` and also below for `active`.

src/peft/tuners/lora/layer.py
```diff
@@ -139,20 +139,56 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
...
+            for variant_combo in layer_variants.keys():
+                for variant_name in variant_combo:
+                    if variant_name not in tagged_fields:
+                        raise ValueError(
+                            f"Variant '{variant_name}' found in lora_variants but not tagged with "
+                            f"'is_lora_variant' in LoraConfig."
+                        )
+
+            # 3. Figure out which variants are currently active
+            active = tuple(sorted(
```
BenjaminBossan: Let's call it `active_variants`.

src/peft/tuners/lora/layer.py
```diff
@@ -139,20 +139,56 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
...
+            layer_variants = getattr(self, "lora_variants", {(): None})
+
+            # 1. Gather all valid variant field names from the config
+            tagged_fields = {
+                f.name for f in dataclasses.fields(config)
+                if f.metadata.get("is_lora_variant")
+            }
+
+            # 2. SANITY CHECK: Ensure all keys in the layer's dictionary actually exist in the config
+            for variant_combo in layer_variants.keys():
```
BenjaminBossan: Let's rename `variant_combo` to `variant_keys`.

src/peft/tuners/lora/layer.py
```diff
@@ -139,20 +139,56 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
...
+                for variant_name in variant_combo:
+                    if variant_name not in tagged_fields:
+                        raise ValueError(
+                            f"Variant '{variant_name}' found in lora_variants but not tagged with "
```
```suggestion
                            f"Variant '{variant_name}' found in lora_variants but it is not tagged with "
```

tests/test_lora_variants.py
```diff
@@ -173,7 +174,20 @@ def test_dora_params_have_gradients(self):
...
+            config = LoraConfig()
+            base_layer = nn.Linear(10, 10)
+            layer = LoraLinear(base_layer, "default", config, r=8, lora_alpha=8)
+
+            # 2. Monkey-patch the lora_variants property to include a fake variant
+            with patch("peft.tuners.lora.layer.Linear.lora_variants", new_callable=PropertyMock) as mock_variants:
+                mock_variants.return_value = {("fake_unregistered_variant",): None}
+                
+                # 3. Assert that the sanity check catches it and throws the right error
+                with pytest.raises(ValueError, match="not tagged with 'is_lora_variant'"):
```
BenjaminBossan: Let's check the full error message here.

tests/test_lora_variants.py
```diff
@@ -173,7 +174,20 @@ def test_dora_params_have_gradients(self):
...
+                    layer.resolve_lora_variant(config=config)
```
BenjaminBossan: Could you please also add a test that checks for the `Invalid or unsupported variant combination` error? It'll probably require monkey-patching too.

BenjaminBossan (changes requested):
Thanks for the recent updates, the PR is taking a good shape. I made a few comments, please check. Also, once you're finished with your changes, please run `make style` to ensure proper code formatting.

isha822:
@BenjaminBossan , all changes addressed:

- Moved `import dataclasses` to module level
- Extracted `lora_variant_configs` and reused it for both `tagged_fields` and `active_variants`
- Renamed `variant_combo` → `variant_keys` and `active` → `active_variants`
- Updated `test_unregistered_variant_raises_error` to match the full error message
- Added `test_invalid_variant_combination_raises_error` for the unsupported combination error

All 16 tests passing.

tests/test_lora_variants.py
```diff
@@ -225,9 +255,8 @@ def test_alora_activation_matches_base_until_invocation(self):
 
         input_ids = torch.tensor([[0, 1, 2, 3]])
         start = 2
-        with lora_model.disable_adapter():
-            with torch.no_grad():
-                base_out = lora_model(X=input_ids)
+        with lora_model.disable_adapter(), torch.no_grad():
```
BenjaminBossan: Let's not touch unrelated code.

src/peft/tuners/lora/layer.py
```diff
@@ -140,19 +140,51 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
...
+        to a specific composed variant class.
+        """
+        return {(): None}
+
     def resolve_lora_variant(self, *, config: LoraConfig, **kwargs) -> Optional[LoraVariant]:
-        """Return a matching LoRA variant for this layer type.
 
-        Given the init arguments of this layer, return the correct LoRA variant, if any. E.g., if `use_dora=True`, this
-        method should return the DoRA variant for the given layer. If `use_alora=True`, same for aLoRA.
+            # Safely fetch the dictionary (defaults to empty if a subclass forgot to define it)
```
BenjaminBossan: The indentation here and elsewhere is not correct. I think running `make style` should be enough to fix it. Could you please try that?

BenjaminBossan (changes requested):
Thanks for the update. I have two more comments.

isha822:
@BenjaminBossan  the unrelated changes in the test file were introduced by `make style` modifying code outside the scope of this PR. I'll revert those and run `make style` only on the files I've modified before pushing the fix. Will update you shortly.

isha822:
@BenjaminBossan  resolved the merge conflict. 
One thing, the conflict was caused by monteclora_config 
being added to upstream's if-chain. 
I've kept the registry approach but haven't added 
monteclora_config to the lora_variants dict yet since 
I wasn't sure if that's in scope for this PR or should be 
handled separately. Let me know how you'd like to proceed, and if there are any other modifications.

BenjaminBossan:
@isha822 Thanks for handling the conflict. Yes, please update the PR to includ MontaCLoRA.

isha822:
@BenjaminBossan I have updated the PR to include MonteCLoRA. Let me know if this is good to go!

src/peft/tuners/lora/layer.py
```diff
@@ -139,19 +140,51 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
...
+        """
+        return {(): None}
+
     def resolve_lora_variant(self, *, config: LoraConfig, **kwargs) -> Optional[LoraVariant]:
-        """Return a matching LoRA variant for this layer type.
 
-        Given the init arguments of this layer, return the correct LoRA variant, if any. E.g., if `use_dora=True`, this
-        method should return the DoRA variant for the given layer. If `use_alora=True`, same for aLoRA.
+        # Safely fetch the dictionary (defaults to empty if a subclass forgot to define it)
+        layer_variants = getattr(self, "lora_variants", {(): None})
```
BenjaminBossan: `getattr` shouldn't be necessary since we define a default `lora_variants` above, right?

tests/test_lora_variants.py
BenjaminBossan: Still open.

tests/test_lora_variants.py
```diff
@@ -272,9 +301,8 @@ def test_num_beams_error(self):
         lora_model.eval()
 
         input_ids = torch.tensor([[0, 1, 2, 3]])
-        with pytest.raises(ValueError) as e:
-            with torch.no_grad():
-                lora_out = lora_model(X=input_ids, num_beams=2, alora_offsets=[3])
+        with pytest.raises(ValueError) as e, torch.no_grad():
```
BenjaminBossan: Also undo this.

BenjaminBossan (changes requested):
Thanks for updating the PR. There are still some open comments. Also, please run `make style` before committing. Thanks.

isha822:
@BenjaminBossan 
Apologies for missing the formatting checks in the last commit. I have now applied `make style` to all three files and addressed your review feedback. Let me know how everything looks to you now!

src/peft/tuners/lora/layer.py
```diff
@@ -139,19 +140,46 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
...
-        If there is no fitting variant, return None.
+        # 1. Gather all valid variant field names from the config
+        tagged_fields = {f.name for f in lora_variants_configs}
 
-        Note: If this layer type does not support the LoRA variant at all, please raise an error during __init__ as is
-        convention, and not here.
+        # 2. SANITY CHECK: Ensure all keys in the layer's dictionary actually exist in the config
+        for variant_keys in layer_variants.keys():
+            for variant_name in variant_keys:
+                if variant_name not in tagged_fields:
```
BenjaminBossan: Thinking about this some more, we can improve the logic a little bit. Right now, this just fails on the first `variant_name` that cannot be found. How about we collect a set of all variant names. Then we can remove the set of `tagged_fields` and find _all_ variant names that are missing and report those. The set can be constructed using a single nested set comprehension, which also removes two layers of indentation. WDYT?

src/peft/tuners/lora/layer.py
```diff
@@ -139,19 +140,46 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
...
+                    raise ValueError(
+                        f"Variant '{variant_name}' found in lora_variants but it is not tagged with "
+                        f"'is_lora_variant' in LoraConfig."
+                    )
 
-        """
-        return None
+        # 3. Figure out which variants are currently active
+        active_variants = tuple(sorted(f.name for f in lora_variants_configs if getattr(config, f.name)))
```
BenjaminBossan: We expect the variants to be sorted in lexicographic order. However, when the `lora_variants` property is defined, it could, theoretically, be defined in unsorted order, in which case that combined variant could never be matched, right? Let's add a check that the order is correct for all `lora_variants` and add a test for that.

src/peft/tuners/lora/layer.py
```diff
@@ -139,19 +140,46 @@ def __init__(self, base_layer: nn.Module, ephemeral_gpu_offload: bool = False, *
...
+        active variants (e.g., DoRA + another variant) to a specific composed variant class.
+        """
+        return {(): None}
+
     def resolve_lora_variant(self, *, config: LoraConfig, **kwargs) -> Optional[LoraVariant]:
-        """Return a matching LoRA variant for this layer type.
 
-        Given the init arguments of this layer, return the correct LoRA variant, if any. E.g., if `use_dora=True`, this
-        method should return the DoRA variant for the given layer. If `use_alora=True`, same for aLoRA.
+        # Fetch the dictionary of variants
```
BenjaminBossan: I know that we're piling quite a few checks onto `resolve_lora_variant`, which could be called hundreds of times for each model initialization. E.g. generally a sets are efficient for checking `in` but they are more costly to construct than tuples or lists, so the cost may not amortize. I think, given the small amount of variants that we have, the overhead should be negligible even if we're not always using the best data structure. Still, let's be careful to be efficient here so that the approach scales even if we add a lot more variants in the future.

BenjaminBossan (changes requested):
Thanks for your latest updates. Overall, I'm quite happy with the approach and confident that we can merge it. I only found a few small issues left, otherwise we're good.

isha822:
Thanks for the great feedback and it's very helpful. @BenjaminBossan 

To address these points, here is my plan for the next commit:

1. **Missing Variants (Set Math):** I'll replace the loop with a nested set comprehension to flatten the keys, then use `missing_variants = all_variant_names - tagged_fields` to catch and report all unregistered variants in a single error.
2. **Lexicographic Order (Fail Fast):** To keep the initialization overhead as low as possible, I'll add a short-circuiting check: `if any(tuple(sorted(k)) != k for k in layer_variants.keys()): raise ValueError(...)`. This will instantly catch any unsorted dictionaries from future developers without adding heavy loop overhead.
3. **Testing:** I'll add the corresponding test in `test_lora_variants.py` to ensure an unsorted dictionary triggers the new error.

Does this specific implementation look good to you before I push the update?

BenjaminBossan:
Sounds good @isha822

isha822:
@BenjaminBossan  apologies for the delay — had my sem ends.

Made the changes:
- Missing variants are now caught with set math instead of the nested loop, so all unregistered ones get reported in one error
- Added a check that fails fast if any lora_variants key isn't sorted
- Added a test for the new sorted-keys check

All 17 tests passing.

BenjaminBossan:
> apologies for the delay — had my sem ends.

No worries.

Thanks for the updates. There has been a new LoRA variant that was merged, could you please update your PR? After that, there are no new LoRA variants in the work right now, so we should be safe from merge conflicts and hopefully finish the PR before the next one comes.

isha822:
@BenjaminBossan  Just pushed the requested updates. Let me know if everything looks good to go.

BenjaminBossan:
Thanks for syncing the PR but I think the merge is not right. We now have separate `resolve_lora_variant` implementations for the layer types. But we should only need the one on the parent class and the layer types should only need to implement `lora_variants`, right?

isha822:
That's exactly the goal but, @BenjaminBossan 
I actually had to leave those overrides in the child classes temporarily after the merge because of the new `MiCA` and `VeLoRA` variants. MiCA uses a string check (`startswith("mica")`) instead of a standard boolean config, and VeLoRA throws a specific error just for Embedding layers. 

Since neither of those fit neatly into the new True/False dictionary, I kept them in the child classes to avoid breaking them. 

How would you prefer I handle those two edge cases? Happy to refactor them however you think fits best.

BenjaminBossan:
Thanks for explaining. It's good that we found some examples that don't quite fit the assumptions, as it allows us to stress test the new design.

I have my own opinions on how to deal with those two, but please let me know your unbiased opinion first (if you want to, of course).

isha822:
My take on both:
@BenjaminBossan 

**MiCA** — compute it as a derived boolean field in `__post_init__` (`is_mica_init`), tagged with `is_lora_variant=True`. It then gets picked up by the existing discovery loop like any other variant — no override needed.

**VeLoRA** — just omit it from `Embedding.lora_variants`. The existing sanity-check `ValueError` for unsupported combinations already fires automatically. Only downside is the error message is generic rather than VeLoRA-specific.

One gap with the MiCA approach: it relies on contributors following convention (deriving a boolean instead of special-casing in the layer) — there's no hard enforcement. I could add a test asserting every `is_lora_variant`-tagged field is an actual `bool`, to catch misuse early.

An alternative for MiCA worth considering: generalize the dict key itself to support string-prefix matching natively, e.g. `("init_lora_weights", "mica"): MiCALinearVariant` as a 2-tuple, vs. `("use_dora",)` as a 1-tuple for booleans. This is more self-documenting (a contributor sees the pattern already in use) but adds real complexity to the parent's discovery loop and messier typing. Happy to go either way — let me know your preference.

BenjaminBossan:
Regarding VeLoRA, I came to the exact same conclusion. Having a more specific error message is not worth it in this case.

Regarding MiCA, my idea was to add a new attribte to `init_lora_weights`, something like

```python
init_lora_weights = field(..., lora_variants=["mica"])
```

In `resolve_lora_variant`, we could then extend the check to include those variants too. This was just the first idea that I had, I could also see your suggestion with `__post_init__` working. LMK what you think.

isha822:
Glad we are aligned on VeLoRA. @BenjaminBossan 
For MiCA, your suggestion to use the metadata field is a great approach. It avoids cluttering the configuration namespace while establishing a scalable, generic pattern for the discovery loop to handle string-based variants in the future.
If this is final, i will implement both of these updates and push the changes shortly.

BenjaminBossan:
Sounds good, let's go ahead with this pattern.

isha822:
done with the update, lemme know if it looks all good. @BenjaminBossan

src/peft/tuners/lora/layer.py
```diff
@@ -837,41 +875,20 @@ def __init__(
...
+
+        return {
+            (): None,
+            ("use_dora",): variants.DoraLinearVariant,
+            ("arrow_config",): variants.ArrowLinearVariant,
+            ("use_bdlora",): variants.BdLoraLinearVariant,
+            ("alora_invocation_tokens",): variants.ALoraLinearVariant,
+            ("velora_config",): variants.VeloraLinearVariant,
+            ("monteclora_config",): variants.MontecloraLinearVariant,
+            ("init_lora_weights",): variants.MiCALinearVariant,
```
BenjaminBossan: The way that you added the option for LoRA variants declared in `init_lora_weights` has a flaw: It would only work if we have a single `init_lora_weights` value that could be a LoRA variant. What if we add a second value, e.g. `init_lora_weights="foobar"` with a `variant.FoobarVariant`? As is, we would not be able to accommodate for this.

I think we should have the variant name here, even if it's coming from `init_lora_weights`:

```suggestion
            ("mica",): variants.MiCALinearVariant,
```

However, this would not work with the current implementation of `resolve_lora_variant`. My suggestion is thus to treat `init_lora_weights` as a special case: Here we take the values declared in the `lora_variants` metadata of `init_lora_weights`. For `use_dora` etc., we use the key. I went ahead and implemented the method accordingly:

```python
    def resolve_lora_variant(self, *, config: LoraConfig, **kwargs) -> Optional[LoraVariant]:
        """TODO"""
        # mapping from declared variant to the LoraVariant class, keys are tuples to allow for multi variants
        lora_variant_mapping = self.lora_variants
        if any(tuple(sorted(k)) != k for k in lora_variant_mapping.keys()):
            raise ValueError("Keys in lora_variants must be sorted tuples (e.g ('a', 'b'), not ('b', 'a')).")

        # for each LoRA variant, determine if they are configured as active or not
        requested_lora_variants: dict[str, bool] = {}
        for field in dataclasses.fields(config):
            # for init_lora_weights, we collect the field's lora_variants and check the value to determine if the
            # variant is requested
            if field.name == "init_lora_weights":
                for init_variant_option in field.metadata["lora_variants"]:
                    requested_lora_variants[init_variant_option] = config.init_lora_weights == init_variant_option
            # for all other fields, if the metadata declares them as is_lora_variant and we check if the value is truthy
            # to determine if the variant is requested
            elif field.metadata.get("is_lora_variant"):
                requested_lora_variants[field.name] = bool(getattr(config, field.name))

        # sanity check: each mapping key must be present in the LoraConfig
        all_variant_names = {name for variant_keys in lora_variant_mapping.keys() for name in variant_keys}
        missing_variants = all_variant_names - requested_lora_variants.keys()
        if missing_variants:
            raise ValueError(
                f"variant(s) {sorted(missing_variants)} found in lora_variants but neither tagged with "
                f"'is_lora_variant' in LoraConfig, nor declared as a LoRA variant in init_lora_weights."
            )

        # collect active variants: keys are sorted tuples of strings
        requested_keys = tuple(sorted(k for k, v in requested_lora_variants.items() if v))
        if requested_keys not in lora_variant_mapping:
            raise ValueError(f"Invalid or unsupported variant combination: {requested_keys}")

        # return the found variant or None for vanilla LoRA
        variant_class = lora_variant_mapping[requested_keys]
        return variant_class() if variant_class else None
```

Please check my suggestion and let me know if you understand what is happening and if it makes sense to you.

If you agree, please use this function and also add a test that shows that we could, hypothetically, have two `init_lora_weights` LoRA variants.

BenjaminBossan (changes requested):
Thanks for the update, but I think there is a flaw in the current implementation. Please check my comment.

isha822:
Thank you for pointing out the dictionary collision issue. @BenjaminBossan 

I've applied your resolve_lora_variant implementation and updated the child registry keys to use the string values (e.g., ("mica",)). I also added the requested mock test to prove the new loop cleanly separates multiple string variants on the same field.
All tests are green.

src/peft/tuners/lora/layer.py
```diff
@@ -143,19 +144,56 @@ def delete_adapter(self, adapter_name: str) -> None:
...
+        variant field names (from LoraConfig) and the values are the specific LoraVariant subclasses.
 
-        If there is no fitting variant, return None.
+        Tuples are used as keys because they are immutable and hashable, allowing us to safely map combinations of
+        active variants (e.g., DoRA + another variant) to a specific composed variant class.
+        """
+        return {(): None}
+
+    def resolve_lora_variant(self, *, config: LoraConfig, **kwargs) -> Optional[LoraVariant]:
+        """TODO"""
```
BenjaminBossan: Could you please add the actual docstring ;)

tests/test_lora_variants.py
```diff
@@ -274,7 +360,7 @@ def test_num_beams_error(self):
         input_ids = torch.tensor([[0, 1, 2, 3]])
         with pytest.raises(ValueError) as e:
             with torch.no_grad():
-                lora_out = lora_model(X=input_ids, num_beams=2, alora_offsets=[3])
+                lora_model(X=input_ids, num_beams=2, alora_offsets=[3])
```
BenjaminBossan: Let's not make unrelated changes.

BenjaminBossan (changes requested):
Thanks for updating the method and adding the test. Just two small comments from my side.

isha822:
All done! @BenjaminBossan thanks for the review

src/peft/tuners/lora/config.py
```diff
@@ -630,6 +630,8 @@ class LoraConfig(PeftConfig):
                 "base weight associated with the smallest singular values, A is set to zero, and B is frozen during "
                 "training (only A is updated)."
             ),
+            "is_lora_variant": True,
```
BenjaminBossan: We don't need `is_lora_variant` here, right? Let's remove it. But let's add a comment that explains what the `lora_variants` metadata means here.

BenjaminBossan (changes requested):
Thanks for the last push, we're basically done, failing tests are unrelated. I just noticed a small detail that we should improve, otherwise the PR is good to merge.

isha822:
done! @BenjaminBossan

BenjaminBossan (approved):
Thanks for your work on refactoring the LoRA variant matching and your patience with the PR. 

To summarize, with this change, it should be much easier in the future to add new LoRA variants, the complicated logic has been isolated to `resolve_lora_variant` and contributors now only need to mark their config field with `"is_lora_variant": True,` and add an entry to `lora_variants` like `("foobar,"): variants.MyFoobarVariant`.

Failing CI is unrelated.

PR #3224, opened 2026-05-12
DOC Improve LoHa docs with overview, when-to-use, and quick start exa…

Dhruv-1710:
The LoHa documentation page was minimal — only had the paper abstract and autodoc blocks.

This PR improves it by adding:
- A clearer intro paragraph with context on when LoHa is used
- A note linking to the conceptual guide and related methods (LoKr, LyCORIS)
- A "When to use LoHa" section explaining trade-offs vs LoRA
- A quick start code example using `facebook/opt-350m`

The autodoc blocks for `LoHaConfig` and `LoHaModel` are preserved unchanged.

Related issue: #2310

### DIFF
--- a/docs/source/package_reference/loha.md
+++ b/docs/source/package_reference/loha.md
@@ -1,17 +1,22 @@
 <!--Copyright 2023 The HuggingFace Team. All rights reserved.
+
 Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
 the License. You may obtain a copy of the License at
+
 http://www.apache.org/licenses/LICENSE-2.0
+
 Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
 an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 specific language governing permissions and limitations under the License.
+
 ⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
 rendered properly in your Markdown viewer.
+
 -->
 
 # LoHa
 
-Low-Rank Hadamard Product ([LoHa](https://huggingface.co/papers/2108.06098)) is similar to LoRA except it approximates the large weight matrix with more low-rank matrices and combines them with the Hadamard product. This method is even more parameter-efficient than LoRA and achieves comparable performance. LoHa was originally proposed for federated learning (FedPara) but works well as a general-purpose PEFT method, and is especially popular for fine-tuning image generation models such as Stable Diffusion.
+Low-Rank Hadamard Product ([LoHa](https://huggingface.co/papers/2108.06098)), is similar to LoRA except it approximates the large weight matrix with more low-rank matrices and combines them with the Hadamard product. This method is even more parameter-efficient than LoRA and achieves comparable performance. LoHa was originally proposed for federated learning (FedPara) but works well as a general-purpose PEFT method, and is especially popular for fine-tuning image generation models such as Stable Diffusion.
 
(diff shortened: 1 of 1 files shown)


wadeKeith (commented):

docs/source/package_reference/loha.md
```diff
@@ -1,31 +1,60 @@
 <!--Copyright 2023 The HuggingFace Team. All rights reserved.
-
```
BenjaminBossan, line 2: Please revert these unrelated changes.

docs/source/package_reference/loha.md
```diff
@@ -1,31 +1,60 @@
...
+
+## When to use LoHa
+
+LoHa is a good choice when:
+
+- You are fine-tuning **image generation models** (Stable Diffusion UNet or text encoder), where it is most widely used.
+- You want **higher effective rank** than LoRA for the same number of trainable parameters, since the Hadamard product of two low-rank matrices spans a larger subspace than a single low-rank product.
+- You need to **combine multiple adapters** at inference time using [`PeftMixedModel`](./peft_model#peft.PeftMixedModel), as LoHa is a compatible adapter type.
```
BenjaminBossan: This can be a bit misleading, since (almost) all PEFT methods allow to combine multiple adapters. `PeftMixedModel` is specifically for combining _different_ PEFT methods, like LoHa + LoKr. Also note that this already works with LoRA, so there is no advantage here compared to LoRA.

docs/source/package_reference/loha.md
```diff
@@ -1,31 +1,60 @@
...
+
+Note that LoHa does not currently support embedding layers in PEFT, so for language model tasks that require embedding adaptation, consider [LoRA](./lora) instead.
```
BenjaminBossan: Instead of telling the user which layer types don't work, it makes more sense to tell them which ones do work.

docs/source/package_reference/loha.md
```diff
@@ -1,31 +1,60 @@
...
+
+## Quick start
+
+```python
+from transformers import AutoModelForCausalLM
+from peft import LoHaConfig, get_peft_model, TaskType
+
+model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m")
```
BenjaminBossan: Since you mentioned that LoHa makes most sense for image generation, why not use an image generation example here?

BenjaminBossan (changes requested):
Thanks for extending the LoHa documentation. I have a couple of comments, otherwise LGTM.

docs/source/package_reference/loha.md
Dhruv-1710: Fixed. Updated the bullet to clarify that PeftMixedModel is specifically for combining different PEFT methods, e.g. LoHa together with LoKr.

docs/source/package_reference/loha.md
Dhruv-1710: Fixed. Updated to mention which layer types LoHa does support: linear and Conv2d layers.

docs/source/package_reference/loha.md
Dhruv-1710: Fixed. Replaced the LLM example with a Stable Diffusion UNet example using diffusers.

docs/source/package_reference/loha.md
Dhruv-1710: Fixed in the latest commit. Thank you for the feedback!

BenjaminBossan (approved):
Thanks for improving the LoHa docs, LGTM.

Dhruv-1710:
Thank you for the review and merging the PR! Really appreciate the detailed feedback.

PR #3225, opened 2026-05-12
refine test_causal_lm_training_multi_accelerator test case

kaixuanliu:
In test case `tests/test_gpu_examples.py::PeftAwqGPUTests::test_causal_lm_training_multi_accelerator`, the model is dispatched to 2 cards by hardcode, and it will fail if we have cards num>2 (e.g., 8 cards). This PR tries to fix this.


kaixuanliu:
@BenjaminBossan pls help review, thx!

wadeKeith (commented):

BenjaminBossan (approved):
Thanks, makes sense.

PR #3226, opened 2026-05-12
refine test case for `test_lora_seq2seq_lm_multi_gpu_inference`

kaixuanliu:
Original code use `device_map="balanced"`, which will lead to all model weight be assigned to 1 single card by accelerator(in my experiment I use 8 A100 cards), and hence it will fail with error log: `AttributeError: 'T5ForConditionalGeneration' o
bject has no attribute 'hf_device_map'`, I refine the test case here.pls help review, thx! @BenjaminBossan


wadeKeith (commented):

BenjaminBossan (approved):
LGTM, test passes on my machine.

PR #3227, opened 2026-05-12
FIX BEFT merge on bias-less layers; RandLoRA save_projection default

kashif:
BEFT: add a zero bias to base layers that have no bias at adapter init time so that merge()/unmerge() work correctly on models like Qwen3.5 whose linear projections have no bias term (attention_bias=False). Previously, this raised a ValueError during vLLM weight sync.

RandLoRA: change save_projection default from True to False. The random basis (randlora_A / randlora_B) is fully deterministic given projection_prng_key and can always be regenerated on load, so saving it by default inflated checkpoints by ~42 GB on 4B-parameter models. Users who need portability across PyTorch RNG versions can opt in with `save_projection=True`.

### DIFF
--- a/src/peft/tuners/beft/layer.py
+++ b/src/peft/tuners/beft/layer.py
@@ -45,9 +45,11 @@ class BeftLayer(BaseTunerLayer):
     def update_layer(self, adapter_name: str, config: BeftConfig, **kwargs):
         base_layer = self.get_base_layer()
         if base_layer.bias is None:
-            warnings.warn(
-                "Detected that the base layer has no bias term. "
-                "Note you cannot merge the BEFT adapter into the base layer."
+            # Add a zero bias so that merge() can write the BEFT bias into it.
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
@kashif When we worked on this PR, we explicitly decided not to add a bias when there was none beforehand. Generally in PEFT, after `merge_and_unload`, we want the model to have exactly the same architecture as the base model. With this change, it would have a different architecture, right?

BenjaminBossan:
> RandLoRA: change save_projection default from True to False. The random basis (randlora_A / randlora_B) is fully deterministic given projection_prng_key and can always be regenerated on load, so saving it by default inflated checkpoints by ~42 GB on 4B-parameter models.

42 GB on a 4B model sounds like a bug honestly. Do you have a reproducer?

BenjaminBossan:
ping @kashif

kashif:
opps lost track! checking

PR #3228, opened 2026-05-12
fix CI bug for multi-device scene

kaixuanliu:
On multi-devices env, Hardcoded `.to(0)` causes RuntimeError: `Expected all tensors to be on the same device, but got index ...`. This PR fixed 14 occurrences across 5 test functions. @BenjaminBossan pls help review, thx!


BenjaminBossan (approved):
Thanks for the update, tests passed on my machine. The one failing CI run is unrelated.

PR #3231, opened 2026-05-12
docs: improve MiSS documentation with practical guidance and LLM example

vedantnavle13:
Closes related to #2310

Improves the MiSS method documentation page by adding:
- Plain English overview of how MiSS works and how it differs from LoRA
- When to use MiSS section
- Explanation of all three init_weights modes (True/bat/mini) with practical guidance
- LLM-focused quick start example (replaces the Stable Diffusion-only example)
- Link to the full fine-tuning example

As suggested by @BenjaminBossan in #2310.


BenjaminBossan:
@vedantnavle13 Thanks for the PR. I haven't checked the exact changes yet, but could you please reformat it so that there are no line breaks in the normal text? They're not necessary for md files.

vedantnavle13:
Done, removed the line breaks from the normal text paragraphs.

docs/source/package_reference/miss.md
```diff
@@ -16,12 +16,73 @@ rendered properly in your Markdown viewer.
 
 # MiSS
 
-MiSS: Balancing LoRA Performance and Efficiency with Simple Shard Sharing([MiSS](https://huggingface.co/papers/2409.15371)) is a novel PEFT method that adopts a low-rank structure, requires only a single trainable matrix, and introduces a new update mechanism distinct from LoRA, achieving an excellent balance between performance and efficiency.
+[MiSS (Matrix Shard Sharing)](https://huggingface.co/papers/2409.15371) is a PEFT method that achieves an excellent balance between model performance and computational efficiency. Unlike LoRA which uses two separate low-rank matrices (A and B), MiSS decomposes the weight matrix into multiple fragment matrices and uses a single shared trainable matrix across all fragments. This shard-sharing mechanism reduces the number of trainable parameters while maintaining strong adaptability, and significantly improves initialization speed and training throughput compared to LoRA and its variants.
```
BenjaminBossan: The link to HF papers still uses an older revision of the paper and not the latest one. Not sure if this can be changed in HF papers, if not let's link the arxiv paper instead: https://arxiv.org/abs/2409.15371.

> excellent balance

Let's avoid too grandiose terms, "good balance" should be good enough :)

> MiSS decomposes the weight matrix into multiple fragment matrices

Let's try to stick with the terms of the paper, I didn't see "fragment" being used. In general, I also don't think it's very crucial to explain how the method works under the hood here, unless it helps the reader understand why/when/how to use MiSS. If they're interested in the implementation, they can check the paper or the code.

docs/source/package_reference/miss.md
```diff
@@ -16,12 +16,73 @@ rendered properly in your Markdown viewer.
...
 
+## When to use MiSS
+
+MiSS is a good choice when:
+
+- You want faster initialization and higher training throughput than LoRA
```
BenjaminBossan: I don't think that MiSS initializes faster than LoRA, what the authors mean is that some LoRA initialization schemes like PiSSA, LoRA-GA, or OLoRA or slower.

docs/source/package_reference/miss.md
```diff
@@ -16,12 +16,73 @@ rendered properly in your Markdown viewer.
...
+- You need to reduce memory usage while maintaining model performance
+- You are fine-tuning large language models where computational efficiency matters
```
BenjaminBossan: This statement is very broad and arguably applies to each PEFT method (as that is the whole point of PEFT).

BenjaminBossan (changes requested):
Thanks for the latest update. I checked the PR and I like that you provide more details and give an example. However, some statements are IMO not quite accurate or not helpful, which I flagged. Please check my comments.

vedantnavle13:
Hi @BenjaminBossan, I've addressed all the review comments:
- Updated paper link to arxiv
- Changed "excellent balance" to "good balance"
- Fixed the initialization claim — now specific to PiSSA/LoRA-GA/OLoRA
- Removed the broad memory and LLM fine-tuning bullets

Please let me know if anything else needs to be changed.

BenjaminBossan:
Thanks @vedantnavle13 for the updates.

@Joluck If you have time, could you please check?

docs/source/package_reference/miss.md
Joluck: yes

Joluck:
I have finished checking and there are no issues. ty : )  @vedantnavle13 @BenjaminBossan

BenjaminBossan (approved):
Thanks for extending the MiSS documentation @vedantnavle13. Thanks @Joluck for checking.

PR #3232, opened 2026-05-12
FIX Resolve regex target_modules in MoE config conversion

### LINKED ISSUE #3229: [BUG] peft 0.19 target_modules (str) use `set`
### System Info

Image

### Who can help?

_No response_

### Reproduction

-

### Expected behavior

-

1fanwang:
Closes #3229.

`_convert_peft_config_moe` did `set(peft_config.target_modules)` on a value that can be a **string**. A regex `target_modules` — e.g. what ms-swift's `get_multimodal_target_regex` emits for Qwen3-Omni — was split into its characters, and conversion failed downstream:

```
ValueError: Target modules {'^', '(', 'q', '_', ...} not found in the base model.
```

`"all-linear"` hit the same path and became `{'-', 'a', 'e', 'i', 'l', 'n', 'r'}`.

## Fix

Resolve a string `target_modules` to the concrete projection names it targets **before** the list-based remap runs. The remap is left untouched, so a list, a regex, and `"all-linear"` all flow through the same path.

The subtlety is that the pre-fusion experts (`gate_proj`, `up_proj`) no longer exist as modules on a v5 model — they're fused into the `gate_up_proj` parameter — so a regex naming them matches nothing via `named_modules()`. Those names are recovered from the fused parameter's container, so they remap into `target_parameters` instead of being silently dropped (which would leave the experts unadapted). `"all-linear"` resolves through `_maybe_include_all_linear_layers`, and since the experts were `nn.Linear` before fusion, it pulls them in too.

A regex that targets only one half of a fused pair raises the same completeness error as the list path; a string that resolves to nothing falls through to PEFT's standard "not found" error.

## End-to-end

`trl-internal-testing/tiny-Qwen3MoeForCausalLM`, transformers 5.11.0:

<details><summary>Before (on <code>main</code>): the regex is split into characters</summary>

```python
regex = r"^(model(?=\.).*\.(q_proj|k_proj|v_proj|gate_proj|up_proj|down_proj))$"
get_peft_model(model, LoraConfig(target_modules=regex))
```
```
ValueError: Target modules {'e', 'm', '^', '.', 'o', ')', 'p', '$', ...} not found in the base model.
```
</details>

<details><summary>After: attention adapted as modules, fused experts as parameters</summary>

```
target_modules    : {q_proj, k_proj, v_proj}
target_parameters : {gate_up_proj, down_proj}   # r and alpha doubled for the fused pair
adapted leaves    : q_proj, k_proj, v_proj, experts
```
</details>

## Tests

`TestTransformersV5` adds 7 cases (16 with parametrization), each failing on `main` with the character-split error and green here: the two regex shapes, an attention-only regex (no MoE remap), `"all-linear"`, an unresolvable string raising the standard error, a partial-fusion regex raising the completeness error, and a save/reload roundtrip that also covers conversion idempotency.

Verified locally on transformers 5.11.0 (CI on this repo is gated on maintainer approval for external contributors):

<details><summary>Test + lint run</summary>

```
$ pytest tests/test_integrations.py -k "qwen3_moe or mixtral"
30 passed, 9 deselected, 4 warnings in 13.46s

$ ruff check src/peft/utils/transformers_weight_conversion.py tests/test_integrations.py
All checks passed!
$ ruff format --check src/peft/utils/transformers_weight_conversion.py tests/test_integrations.py
2 files already formatted
```
</details>

Refs: modelscope/ms-swift#9321.

### DIFF
--- a/src/peft/utils/transformers_weight_conversion.py
+++ b/src/peft/utils/transformers_weight_conversion.py
@@ -400,10 +400,9 @@ def _resolve_string_target_modules(peft_config, model, target_module_mapping: di
         if is_all_linear:
             matched.add(old_name)
             continue
-        # a representative expert index (0/1) is enough for the regexes seen in practice
+        # probe the reconstructed v4 key (the fused parameter's container + the pre-fusion projection name)
         for container in containers:
-            candidates = (f"{container}.{old_name}", f"{container}.0.{old_name}", f"{container}.1.{old_name}")
-            if any(check_target_module_exists(peft_config, candidate) for candidate in candidates):
+            if check_target_module_exists(peft_config, f"{container}.{old_name}"):
                 matched.add(old_name)
                 break
     return matched
--- a/src/peft/utils/transformers_weight_conversion.py
+++ b/src/peft/utils/transformers_weight_conversion.py
@@ -358,13 +358,13 @@ def _resolve_string_target_modules(
     """Resolve a string `target_modules` (a regex or `"all-linear"`) to the concrete leaf names it targets.
 
     The keys of `target_module_mapping` are the v4 projection names that were renamed/fused on v5: the expert
-    projections (`gate_proj`, `up_proj`, `down_proj`) and the router `gate`. Resolving the string to leaf names up front
-    lets the list-based remap downstream stay unchanged for a regex and for `"all-linear"` alike.
+    projections (`gate_proj`, `up_proj`, `down_proj`) and the router `gate`. Resolving the string to leaf names up
+    front lets the list-based remap downstream stay unchanged for a regex and for `"all-linear"` alike.
 
-    `"all-linear"` matched every `nn.Linear` on the v4 model, so it targeted all of those projections; on v5 they are no
-    longer `nn.Linear` (the experts are fused into stacked parameters, the router became a custom module), so they are
-    added directly by name. A regex only targets what it spells out, so the experts -- which no longer exist as real
-    keys -- are recovered by probing each reconstructed v4 key against the pattern.
+    `"all-linear"` matched every `nn.Linear` on the v4 model, so it targeted all of those projections; on v5 they are
+    no longer `nn.Linear` (the experts are fused into stacked parameters, the router became a custom module), so they
+    are added directly by name. A regex only targets what it spells out, so the experts -- which no longer exist as
+    real keys -- are recovered by probing each reconstructed v4 key against the pattern.
     """
     if peft_config.target_modules.lower() == INCLUDE_LINEAR_LAYERS_SHORTHAND:
         # attention projections are still `nn.Linear` on v5, so resolve them by type; the experts and router were
@@ -407,7 +407,8 @@ def _convert_peft_config_moe(peft_config: PeftConfig, model: torch.nn.Module) ->
     requires updating the rank and alpha values of those parameters.
 
     A string `target_modules` (a regex, or `"all-linear"`) is resolved against the model up front (see
-    `_resolve_string_target_modules`), so the list-based remap below applies to lists, regexes and `"all-linear"` alike.
+    `_resolve_string_target_modules`), so the list-based remap below applies to lists, regexes and `"all-linear"`
+    alike.
     """
     model_type = getattr(model.config, "model_type", None)
     base_model_type = _MODEL_TO_CONVERSION_PATTERN.get(model_type, None)
--- a/src/peft/utils/transformers_weight_conversion.py
+++ b/src/peft/utils/transformers_weight_conversion.py
@@ -357,44 +357,42 @@ def _resolve_string_target_modules(
 ) -> set[str]:
     """Resolve a string `target_modules` (a regex or `"all-linear"`) to the concrete leaf names it targets.
 
-    Resolving up front lets the list-based remap downstream stay unchanged. The pre-fusion expert projections
-    (`gate_proj`, `up_proj`) are absent from a v5 model -- they live inside the stacked `gate_up_proj` parameter -- so
-    they are recovered from each fused parameter's container, not from `named_modules()`/`named_parameters()`.
+    The keys of `target_module_mapping` are the v4 projection names that were renamed/fused on v5: the expert
+    projections (`gate_proj`, `up_proj`, `down_proj`) and the router `gate`. Resolving the string to leaf names up front
+    lets the list-based remap downstream stay unchanged for a regex and for `"all-linear"` alike.
+
+    `"all-linear"` matched every `nn.Linear` on the v4 model, so it targeted all of those projections; on v5 they are no
+    longer `nn.Linear` (the experts are fused into stacked parameters, the router became a custom module), so they are
+    added directly by name. A regex only targets what it spells out, so the experts -- which no longer exist as real
+    keys -- are recovered by probing each reconstructed v4 key against the pattern.
     """
+    if peft_config.target_modules.lower() == INCLUDE_LINEAR_LAYERS_SHORTHAND:
+        # attention projections are still `nn.Linear` on v5, so resolve them by type; the experts and router were
+        # `nn.Linear` in v4 too but are now parameters / a custom module, so add their v4 names (the mapping keys).
+        resolved = _maybe_include_all_linear_layers(copy.copy(peft_config), model).target_modules
+        return {name.rsplit(".", 1)[-1] for name in resolved} | target_module_mapping.keys()
+
+    # regex: leaf names of the real module/parameter keys it matches (q/k/v_proj, the router module, `down_proj`, ...)
     module_keys = [name for name, _ in model.named_modules()]
     param_keys = [name for name, _ in model.named_parameters()]
-    # new (fused) name -> container paths, used to reconstruct the pre-fusion expert keys
+    matched = {
+        key.rsplit(".", 1)[-1]
+        for key in itertools.chain(module_keys, param_keys)
+        if check_target_module_exists(peft_config, key)
+    }
+
+    # The pre-fusion experts (`gate_proj`, `up_proj`) live inside the stacked `gate_up_proj` parameter, so a regex
+    # naming them matches no real key. Recover them from that parameter's container: e.g. for the parameter
+    # `model.layers.0.mlp.experts.gate_up_proj`, probe the reconstructed v4 key `model.layers.0.mlp.experts.gate_proj`.
     new_name_to_containers: dict[str, set[str]] = {}
     for name in param_keys:
         container, _, leaf = name.rpartition(".")
         new_name_to_containers.setdefault(leaf, set()).add(container)
-
-    is_all_linear = peft_config.target_modules.lower() == INCLUDE_LINEAR_LAYERS_SHORTHAND
-    if is_all_linear:
-        resolved = _maybe_include_all_linear_layers(copy.copy(peft_config), model).target_modules
-        matched: set[str] = {name.rsplit(".", 1)[-1] for name in resolved}
-    else:
-        # leaf names of the real module/parameter keys the regex matches (q/k/v_proj, router `gate`, `down_proj`)
-        matched = {
-            key.rsplit(".", 1)[-1]
-            for key in itertools.chain(module_keys, param_keys)
-            if check_target_module_exists(peft_config, key)
-        }
-
-    # Recover the fused experts: container `model.layers.0.mlp.experts` of `...experts.gate_up_proj` reconstructs the
-    # v4 keys `...experts.gate_proj` / `.up_proj` a regex targets. "all-linear" matches a fused expert iff its
-    # container is present (it was `nn.Linear` in v4); a regex is probed against the reconstructed key.
     for old_name, new_name in target_module_mapping.items():
         if old_name in matched:
             continue
-        containers = new_name_to_containers.get(new_name, ())
-        if not containers:
-            continue
-        if is_all_linear:
-            matched.add(old_name)
-            continue
         # one hit suffices -- `matched` holds leaf names, so stop at the first container (layer) the regex matches
-        for container in containers:
+        for container in new_name_to_containers.get(new_name, ()):
             if check_target_module_exists(peft_config, f"{container}.{old_name}"):
                 matched.add(old_name)
                 break
--- a/src/peft/utils/transformers_weight_conversion.py
+++ b/src/peft/utils/transformers_weight_conversion.py
@@ -14,6 +14,7 @@
 
 # NOTE: don't import from this module unless transformers v5+ is used
 import copy
+import itertools
 import re
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
@1fanwang Thanks for providing a PR to fix the problem. In the future, please ask first if you can work on the issue (or instruct the agent to do that) because otherwise, this may create duplicate work. We may close PRs without inspection if they're already being worked on by someone else.

As for your solution: I wonder if we can use `peft.tuners.tuners_utils.check_target_module_exists` to "resolve" the actual target modules from the string and work based on that, instead of bailing. Did you explore that option?

1fanwang:
Apologies on the duplicate-work front — happy to wait for a green light before drafting next time.

On `check_target_module_exists`: I did look at it before settling on the bail. The blocker for using it here is that `_convert_peft_config_moe` only has `peft_config` and `model_type`, not the model — so it can't enumerate concrete module names against the regex (`tuners_utils.py` L1805: `check_target_module_exists(config, key)` takes a single `key` per call). The caller `convert_peft_config_for_transformers` does have `model` and could thread it through:

```python
# convert_peft_config_for_transformers, transformers_weight_conversion.py L444
_convert_peft_config_moe(peft_config, model_type, model=model)
```

then inside `_convert_peft_config_moe` we'd walk `model.named_modules()`, run `re.fullmatch(peft_config.target_modules, name)` on each, and use the matching leaf-names as the candidate set for the old→new remap in `target_module_mapping` (L382-L385 in the current file).

Two things made me back off:

1. It's a non-local change — every caller of `_convert_peft_config_moe` would need to pass the model, and the regex-resolved set is structurally different from the user-declared set (the user declared a pattern, we'd be writing a `set` of concrete matches back).
2. The remap on L398-L399 sets `peft_config.target_modules = remaining_target_modules` — for a regex input, "remaining" doesn't have a clean meaning if some leaves matched and others didn't. We'd have to either widen target_modules to a list-of-leaves (changes user intent) or leave the regex untouched and only mutate target_parameters/rank_pattern/alpha_pattern.

If you'd prefer the resolve-via-model path I'm happy to switch — option (2) above (regex stays, only parameters/patterns updated) feels safer than bailing, but it's a bigger diff than this PR. Want me to take that direction in this PR or stack it as a follow-up after the bail lands?

BenjaminBossan:
> On `check_target_module_exists`: I did look at it before settling on the bail. The blocker for using it here is that `_convert_peft_config_moe` only has `peft_config` and `model_type`, not the model — so it can't enumerate concrete module names against the regex (`tuners_utils.py` L1805: `check_target_module_exists(config, key)` takes a single `key` per call). The caller `convert_peft_config_for_transformers` does have `model` and could thread it through

Yes, passing the model directly (instead of `model_type`) would be the idea.

> 1. It's a non-local change — every caller of `_convert_peft_config_moe` would need to pass the model

There is only one call site, you can consider this to be "private".

> 2\. The remap on L398-L399 sets `peft_config.target_modules = remaining_target_modules` — for a regex input, "remaining" doesn't have a clean meaning if some leaves matched and others didn't.

I don't get what you mean by that. Can you give an example?

BenjaminBossan:
One more thing, if we allow strings here, we also need to take care of `"all-linear"`, which is a [special string that matches by type](https://github.com/huggingface/peft/blob/758cdac51922abbb24b6e772844c0a88bbe1cd7d/src/peft/tuners/tuners_utils.py#L1885) and not by regex.

src/peft/utils/transformers_weight_conversion.py
```diff
@@ -368,33 +418,52 @@ def _convert_peft_config_moe(peft_config, model_type: str) -> None:
...
         return
 
-    peft_config.target_parameters = set(peft_config.target_parameters or [])
-    peft_config.target_modules = set(peft_config.target_modules or [])
+    from peft.utils.constants import INCLUDE_LINEAR_LAYERS_SHORTHAND
+
+    target_modules = peft_config.target_modules
+
+    # `"all-linear"` is a special shorthand matched by module *type*, resolved later against the model in
+    # `_maybe_include_all_linear_layers`. There is nothing to remap here, so leave it untouched.
```
BenjaminBossan: Shouldn't we be able to call `_maybe_include_all_linear_layers` right here and thus resolve the layers? If we just skip `"all-linear"` here, it means that potentially targeted MoE layers are not correctly remapped, right?

src/peft/utils/transformers_weight_conversion.py
```diff
@@ -368,33 +418,52 @@ def _convert_peft_config_moe(peft_config, model_type: str) -> None:
...
+    if isinstance(target_modules, str) and target_modules.lower() == INCLUDE_LINEAR_LAYERS_SHORTHAND:
+        return
+
     if not hasattr(peft_config, "rank_pattern") or peft_config.rank_pattern is None:
         peft_config.rank_pattern = {}
     if not hasattr(peft_config, "alpha_pattern") or peft_config.alpha_pattern is None:
         peft_config.alpha_pattern = {}
+    peft_config.target_parameters = set(peft_config.target_parameters or [])
```
BenjaminBossan, line 437: This still would break if a user passes, for example `target_modules="q_proj"` here, wouldn't it?

src/peft/utils/transformers_weight_conversion.py
```diff
@@ -368,33 +418,52 @@ def _convert_peft_config_moe(peft_config, model_type: str) -> None:
...
-                break
-
-        if mapped_new_name is None:
-            remaining_target_modules.add(target)
-            continue
-
-        new_target_parameters.add(mapped_new_name)
-        if mapped_new_name in fused_targets and mapped_old_name is not None:
-            matched_targets.setdefault(mapped_new_name, set()).add(mapped_old_name)
+    if isinstance(target_modules, str):
```
BenjaminBossan: Ideally, I would leave this part untouched, as it's very error prone to make changes here. Instead, I would prefer to resolve strings before hand and then use the existing logic.

tests/test_integrations.py
```diff
@@ -391,3 +391,112 @@ def test_qwen3_moe_works(self):
...
+        # The fused expert projections (`gate_proj`/`up_proj`) don't exist as submodules on a v5 model, so the targeted
+        # projections are resolved against the model and moved to `target_parameters`; the regex stays for `q/k/v_proj`.
+        model_id = "trl-internal-testing/tiny-Qwen3MoeForCausalLM"
+        with hub_online_once(model_id):
+            model = AutoModelForCausalLM.from_pretrained(model_id)
+            peft_model = get_peft_model(model, LoraConfig(target_modules=regex))  # does not raise
+
+        config = peft_model.peft_config["default"]
+        # the regex is preserved for the non-fused targets (q/k/v_proj)
+        assert config.target_modules == regex
```
BenjaminBossan: IMO, it's okay not to preserve the original regex here.

tests/test_integrations.py
```diff
@@ -391,3 +391,112 @@ def test_qwen3_moe_works(self):
...
+
+        config = peft_model.peft_config["default"]
+        adapted = {
+            name.rsplit(".lora_A", 1)[0].rsplit(".", 1)[-1]
+            for name, _ in peft_model.named_modules()
+            if name.endswith("lora_A.default")
+        }
+        # resolved to the actual linear modules (attention projections)
+        assert {"q_proj", "v_proj"} <= adapted
+        # the fused experts are nn.Parameters on v5, not nn.Linear, so "all-linear" does not pull them in and the MoE
```
BenjaminBossan: I think arguably, if a user targets all linear layers on the Transformers < v5 architecture, it means they intended to target the experts, since those were implemented as linear layers:

https://github.com/huggingface/transformers/blob/753d61104116eefc8ffc977327b441ee0c8d599f/src/transformers/models/qwen3_moe/modeling_qwen3_moe.py#L197-L264

BenjaminBossan (changes requested):
Thanks for the updates but I still see some issues here. Please check my comments.

src/peft/utils/transformers_weight_conversion.py
1fanwang: Thanks for the pointer! Done in 925ac1a7 — resolved via _maybe_include_all_linear_layers.

src/peft/utils/transformers_weight_conversion.py
1fanwang: Just tested this again - iiuc a string is matched as a regex (re.fullmatch), so a bare q_proj matches nothing and falls through to the standard "not found" error — same as any model. Test added in 925ac1a7, happy to discuss this further

src/peft/utils/transformers_weight_conversion.py
1fanwang: Done in 925ac1a7

tests/test_integrations.py
1fanwang: Done in 925ac1a7

tests/test_integrations.py
1fanwang: Thanks for the pointer, Done in 925ac1a7 — all-linear carries the experts into target_parameters too.

src/peft/utils/transformers_weight_conversion.py
```diff
@@ -349,13 +350,77 @@ def build_peft_weight_mapping(
 }
 
 
-def _convert_peft_config_moe(peft_config, model_type: str) -> None:
+def _resolve_string_target_modules(peft_config, model, target_module_mapping: dict[str, str]) -> set[str]:
+    """Resolve a string ``target_modules`` (a regex or the ``"all-linear"`` shorthand) to a concrete set of leaf names.
```
BenjaminBossan: Let's use single backticks in the docstrings, i.e. normal markdown syntax.

src/peft/utils/transformers_weight_conversion.py
```diff
@@ -349,13 +350,77 @@ def build_peft_weight_mapping(
...
+    fused into stacked parameters such as ``gate_up_proj`` -- so they cannot be discovered via
+    ``model.named_modules()``. We therefore match against both real module/parameter keys and v4-style expert keys
+    reconstructed from each fused parameter's container.
+
+    ``"all-linear"`` is matched by module *type* rather than as a regex (see ``_maybe_include_all_linear_layers``). On
+    the original v4 architecture the experts were ``nn.Linear``, so ``"all-linear"`` targeted them; we preserve that by
+    also adding any old projection name whose fused container is present on the model.
+    """
+    from peft.tuners.tuners_utils import _maybe_include_all_linear_layers, check_target_module_exists
+    from peft.utils.constants import INCLUDE_LINEAR_LAYERS_SHORTHAND
```
BenjaminBossan: Please use global imports if possible. If this leads to circular imports, add a comment to indicate so.

src/peft/utils/transformers_weight_conversion.py
```diff
@@ -349,13 +350,77 @@ def build_peft_weight_mapping(
...
+        resolved = _maybe_include_all_linear_layers(copy.copy(peft_config), model).target_modules
+        matched: set[str] = {name.rsplit(".", 1)[-1] for name in resolved}
+    else:
+        # regex: every real module/parameter whose key matches the pattern (covers the attention projections that stay
+        # in `target_modules` as well as the router `gate` and the `down_proj` parameter, which keep their v5 names)
+        matched = {
+            key.rsplit(".", 1)[-1]
+            for key in itertools.chain(module_keys, param_keys)
+            if check_target_module_exists(peft_config, key)
+        }
```
BenjaminBossan: Hmm, I'm probably missing something, but why can we not use the same logic as in `inject_adapter`, so basically:

```python
peft_config = _maybe_include_all_linear_layers(peft_config, model)
matched = {key for key, module in model.named_modules() if key and (check_target_module_exists(peft_config, key) is True)}
```

src/peft/utils/transformers_weight_conversion.py
```diff
@@ -349,13 +350,77 @@ def build_peft_weight_mapping(
...
+
+    # the pre-fusion expert projections (e.g. `gate_proj`, `up_proj`) were fused and renamed, so they are absent from
+    # the v5 model. Recover them from each fused parameter's container: for "all-linear", a present container means the
+    # projection was an `nn.Linear` in v4 and would have been targeted; for a regex, probe the reconstructed v4 keys.
```
BenjaminBossan: I think it would help to understand what's going on here if an example of a typical `container` would be shown here in the comments.

src/peft/utils/transformers_weight_conversion.py
```diff
@@ -349,13 +350,77 @@ def build_peft_weight_mapping(
...
+        if not containers:
+            continue
+        if is_all_linear:
+            matched.add(old_name)
+            continue
+        # probe the reconstructed v4 key (the fused parameter's container + the pre-fusion projection name)
+        for container in containers:
+            if check_target_module_exists(peft_config, f"{container}.{old_name}"):
+                matched.add(old_name)
+                break
```
BenjaminBossan, line 398: For my understanding, why can we break after the first match?

src/peft/utils/transformers_weight_conversion.py
BenjaminBossan: Yeah, I meant something like `.*\.q_proj` but from my understanding, it should be covered by `_resolve_string_target_modules`.

tests/test_integrations.py
```diff
@@ -391,3 +391,147 @@ def test_qwen3_moe_works(self):
...
+        assert not config.target_parameters
+        assert not config.rank_pattern
+        adapted = {
+            name.rsplit(".lora_A", 1)[0].rsplit(".", 1)[-1]
+            for name, _ in peft_model.named_modules()
+            if name.endswith("lora_A.default")
+        }
+        assert adapted == {"q_proj"}
+
+    def test_qwen3_moe_unresolvable_string_target_modules_raises_standard_error(self):
```
BenjaminBossan: Can be deleted, see my comment above.

tests/test_integrations.py
```diff
@@ -391,3 +391,147 @@ def test_qwen3_moe_works(self):
...
+        # A string `target_modules` is matched as a regex (`re.fullmatch`), so a bare leaf name like "q_proj" matches
+        # nothing and is invalid on *any* model -- the MoE conversion must not turn this into a different/confusing
+        # error (previously `set("q_proj")` split it into characters). The standard "not found" error is expected.
+        model_id = "trl-internal-testing/tiny-Qwen3MoeForCausalLM"
+        with hub_online_once(model_id):
+            model = AutoModelForCausalLM.from_pretrained(model_id)
+            with pytest.raises(ValueError, match="Target modules q_proj not found in the base model"):
+                get_peft_model(model, LoraConfig(target_modules="q_proj"))
+
+    def test_qwen3_moe_regex_partial_fusion_raises(self):
```
BenjaminBossan: Let's put this directly below `test_qwen3_moe_error_message` and rename that test to `test_qwen3_moe_partial_fusion_raises`

tests/test_integrations.py
```diff
@@ -391,3 +391,147 @@ def test_qwen3_moe_works(self):
...
+        with hub_online_once(model_id):
+            model = AutoModelForCausalLM.from_pretrained(model_id)
+            msg = re.escape(
+                "Cannot convert PEFT target(s) up_proj without also targeting gate_proj because they are fused into "
+                "gate_up_proj."
+            )
+            with pytest.raises(ValueError, match=msg):
+                get_peft_model(model, LoraConfig(target_modules=r".*\.(up_proj|down_proj)"))
+
+    def test_qwen3_moe_regex_attention_only_no_moe_conversion(self):
```
BenjaminBossan: IMO not needed.

tests/test_integrations.py
```diff
@@ -391,3 +391,147 @@ def test_qwen3_moe_works(self):
...
+        # Reloading runs the config conversion a second time (now on the already-converted config), so this also guards
+        # against the conversion not being idempotent.
+        inputs = torch.arange(10).view(1, -1).to(self.torch_device)
+        model_id = "trl-internal-testing/tiny-Qwen3MoeForCausalLM"
+        regex = r"^(model(?=\.).*\.(q_proj|k_proj|v_proj|gate_proj|up_proj|down_proj))$"
+
+        with hub_online_once(model_id):
+            model = AutoModelForCausalLM.from_pretrained(model_id).to(self.torch_device)
+            peft_model = get_peft_model(model, LoraConfig(target_modules=regex))
+            # perturb the (zero-initialized) LoRA B weights so the adapter actually changes the output
```
BenjaminBossan: No need for this, pass `init_lora_weights=False` to `LoraConfig`.

BenjaminBossan (changes requested):
Thanks for your continued work on the PR. There are a few parts that are unclear to me, so I asked for clarification. I think some parts of the PR can also be improved, please check my comments.

src/peft/utils/transformers_weight_conversion.py
1fanwang: Done in e17397db.

src/peft/utils/transformers_weight_conversion.py
1fanwang: Done in e17397db — no cycle (`tuners_utils` imports this module only via a deferred import).

src/peft/utils/transformers_weight_conversion.py
1fanwang: Tried that — it returns empty. On v5 the experts are fused into `experts.gate_up_proj`, so `gate_proj`/`up_proj` are not in `named_modules()`, and `down_proj` is a parameter. The helper matches over modules+parameters, then reconstructs `{container}.gate_proj` so the named experts are not silently dropped. e17397db.

src/peft/utils/transformers_weight_conversion.py
1fanwang: Done in e17397db — added the example `experts.gate_up_proj` → container `model.layers.0.mlp.experts`.

src/peft/utils/transformers_weight_conversion.py
1fanwang: Done in e17397db — `matched` holds leaf names, so one matching layer is enough.

src/peft/utils/transformers_weight_conversion.py
1fanwang: Right — resolves to `{q_proj}`, covered by `test_qwen3_moe_single_attention_regex_target_modules_works`.

tests/test_integrations.py
1fanwang: Removed in e17397db.

tests/test_integrations.py
1fanwang: Done in e17397db — renamed, and folded list+regex into one parametrized test. Can split if you prefer.

tests/test_integrations.py
1fanwang: Removed in e17397db — covered by `test_qwen3_moe_single_attention_regex_target_modules_works`.

tests/test_integrations.py
1fanwang: Done in e17397db.

src/peft/utils/transformers_weight_conversion.py
```diff
@@ -349,13 +352,66 @@ def build_peft_weight_mapping(
...
+    # container is present (it was `nn.Linear` in v4); a regex is probed against the reconstructed key.
+    for old_name, new_name in target_module_mapping.items():
+        if old_name in matched:
+            continue
+        containers = new_name_to_containers.get(new_name, ())
+        if not containers:
+            continue
+        if is_all_linear:
+            matched.add(old_name)
+            continue
```
BenjaminBossan: Could you please clarify this: We first check if `not containers` and skip in that case. Only then do we check `is_all_linear`. I'm wondering why that is the correct order. When we have no `containers` but we do have `is_all_linear`, why would we not count that as a match? A comment to explain this would be good to have.

When I switch the order, the tests still pass, so we need something in the tests to distinguish these two options. In `test_qwen3_moe_all_linear_target_modules_works`, with the current order, `gate` is not part of `adapted`. However, I think it should be, at least I see no reason why not. In that case, the order should be swapped.

If you agree with that, we can simplify the code even more:

```python
if is_all_linear:
    matched |= target_module_mapping.keys()
else:
    containers = new_name_to_containers.get(new_name, ())
    ...
```

BenjaminBossan (changes requested):
Thank you for your latest updates and explanations.

I checked out your branch and tried to understand better what happens inside of `_resolve_string_target_modules`. While working on this, I encountered one ambiguity, which I flagged. Please take a look.

1fanwang:
> Thank you for your latest updates and explanations.
> 
> I checked out your branch and tried to understand better what happens inside of `_resolve_string_target_modules`. While working on this, I encountered one ambiguity, which I flagged. Please take a look.

Thanks for guiding me through this journey @BenjaminBossan, checking now

src/peft/utils/transformers_weight_conversion.py
1fanwang: Done in 0e861fe — `all-linear` now unions `target_module_mapping.keys()`, so the router (`gate.weight`) isn't dropped. Tests pin both paths.

BenjaminBossan:
@1fanwang Thanks for the update. Could you please run `make style`?

BenjaminBossan (approved):
Thanks for fixing the resolution of str targets. LGTM.

PR #3234, opened 2026-05-13
fix(tuners/lora/torchao): make get_apply_tensor_subclass optional (#3223)

### LINKED ISSUE #3223: TorchaoLoraLinear: __init__() missing get_apply_tensor_subclass kwarg with torchao 0.17 (PEFT 0.19.1)
## TL;DR

PEFT 0.19.1's `TorchaoLoraLinear` adapter has a constructor signature incompatibility with torchao 0.17 — `__init__()` is missing the `get_apply_tensor_subclass` kwarg that the dispatcher expects to pass through. Hit while building a dual-GPU FLUX.2 LoRA training path (https://github.com/genno-whittlery/flux2-dual-gpu-lora) — surfaces any time you do:

```python
quantize_(model, Float8WeightOnlyConfig())   # torchao 0.17
model.add_adapter(lora_config)               # PEFT 0.19.1, default LoraConfig
```

A workaround that fits the common LoRA-on-quantized-base case: invert the order (LoRA-first, then quant with a `filter_fn` that excludes `lora_A` / `lora_B` from the quant pass — keeps PEFT routing through its normal `LoraLinear`, never instantiates `TorchaoLoraLinear`). Full repro + workaround details below.

## Environment

| | |
|---|---|
| PEFT | 0.19.1 |
| torchao | 0.17.0 |
| torch | 2.11.0+cu130 |
| transformers | 4.57.6 / 5.8.0 (both surface the same bug) |
| Python | 3.12.11 |
| OS | Windows 11 + Ubuntu (both) |
| GPU | 2× RTX 5090 (Blackwell, sm_120) |
| Model | FLUX.2-dev transformer (~60 GB bf16, 8 double + 48 single transformer blocks) |

## Repro

Minimal — quantize first, then inject LoRA:

```python
import torch
from diffusers import Flux2Transformer2DModel  # or any nn.Module with target Linear layers
from peft import LoraConfig
from torchao.quantization import quantize_, Float8WeightOnlyConfig

transformer = Flux2Transformer2DModel.from_pretrained(
    "black-forest-labs/FLUX.2-dev", subfolder="transformer", torch_dtype=torch.bfloat16
).to("cuda:0")

quantize_(transformer, Float8WeightOnlyConfig())          # base weights → WeightOnlyFloat8Tensor

lora_config = LoraConfig(
    r=32, lora_alpha=32,
    target_modules=["to_q", "to_k", "to_v", "to_out.0"],
)
transformer.add_adapter(lora_config)   # ← crashes here
```

Traceback:

```
TypeError: TorchaoLoraLinear.__init__() missing 1 required positional argument:
'get_apply_tensor_subclass'
```

The `TorchaoLoraLinear` adapter is selected because PEFT detects the `torchao` quant on the base layer's `.weight`. But its constructor signature doesn't match what the dispatcher passes.

## What we did instead

The combination that actually works for FLUX.2 LoRA training on dual consumer GPUs (validated on 4 trainers: ai-toolkit, musubi-tuner, OneTrainer, DiffSynth-Studio):

```python
# 1. Inject LoRA FIRST, on bf16 base weights — PEFT routes through its normal
#    LoraLinear because no torchao base is present yet.
transformer.add_adapter(lora_config)

# 2. THEN quantize, with a filter_fn that excludes LoRA's own Linear submodules.
#    quantize_ swaps the base Linear's .weight.data with a fp8 tensor in-place;
#    LoraLinear's reference to base_layer still works because LoraLinear didn't
#    move. PEFT's normal LoraLinear runs forward through the now-fp8 base, and
#    its own lora_A / lora_B stay bf16 with requires_grad intact.
def _quant_filter(module, name):
    if not isinstance(module, torch.nn.Linear):
        return False
    return "lora_A" not in name and "lora_B" not in name

quantize_(transformer, Float8WeightOnlyConfig(), filter_fn=_quant_filter)
```

The `filter_fn` matters independently of the order: without it, torchao recurses into PEFT's `lora_A.default` / `lora_B.default` `nn.Linear` submodules and replaces their `.weight` with `WeightOnlyFloat8Tensor`, which has no `requires_grad`. Backward then fails at the loss with:

```
RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

Combined with the LoRA-first ordering, this avoids `TorchaoLoraLinear` entirely.

## What I think the underlying issue is

PEFT 0.19.1's `TorchaoLoraLinear` was written against an older torchao API. The current `_replace_with_custom_fn_if_matches_filter` in torchao 0.17 expects adapter constructors to either accept `get_apply_tensor_subclass` or carry it through `**kwargs`; PEFT's `TorchaoLoraLinear.__init__` does neither. Either:

1. PEFT should add `get_apply_tensor_subclass=None` to `TorchaoLoraLinear.__init__` (forward to base), **or**
2. PEFT should detect torchao 0.17+ and decline to install `TorchaoLoraLinear` for `WeightOnlyFloat8Tensor` cases — falling back to the regular `LoraLinear` path, which actually works perfectly fine here because the in-place `.data` swap pattern doesn't require special PEFT plumbing.

Happy to put together a PR for either approach if maintainers have a preference on which path is preferable.

Anai-Guo:
## Summary
Fixes #3223 — `model.add_adapter(lora_config)` crashes with `TypeError: TorchaoLoraLinear.__init__() missing 1 required positional argument: 'get_apply_tensor_subclass'` whenever the base model has been quantized with `torchao.quantize_(...)` first.

## Root cause
`TorchaoLoraLinear.__init__` declares `get_apply_tensor_subclass` as a keyword-only required argument:

```python
def __init__(self, *args, get_apply_tensor_subclass, **kwargs):
```

But `dispatch_torchao` (in the same file) does not pass it:

```python
new_module = TorchaoLoraLinear(target, adapter_name, config=config, **kwargs)
```

…and `add_adapter(lora_config)` doesn't put it into `kwargs`. So the constructor fails before forward ever runs.

## Fix
Per [@BenjaminBossan's preference in the issue](https://github.com/huggingface/peft/issues/3223#issuecomment-...), go with solution 1: make `get_apply_tensor_subclass` optional (`Optional[Callable] = None`) and raise an actionable `ValueError` at merge / unmerge time if it was never set. The argument is only consumed by `merge()` / `unmerge()` (to re-quantize the base layer after applying the LoRA delta) — forward and training don't need it, so the common `quantize_ → add_adapter → train` path now works.

## What this enables

```python
import torch
from peft import LoraConfig
from torchao.quantization import quantize_, Float8WeightOnlyConfig

quantize_(model, Float8WeightOnlyConfig())
model.add_adapter(LoraConfig(r=32, target_modules=[...]))  # no longer crashes
```

If a user later calls `merge()` / `unmerge()` without ever having supplied `get_apply_tensor_subclass`, they now get:

```
ValueError: TorchaoLoraLinear was instantiated without `get_apply_tensor_subclass`, which is
required to re-quantize the base layer after merging. Pass it via the adapter kwargs (e.g.
through `dispatch_torchao`) before calling merge().
```

…instead of a confusing constructor failure earlier.

## Notes
- Pre-existing torchao tests pass `get_apply_tensor_subclass` explicitly, so behavior there is unchanged.
- The longer-term fix discussed in #3223 (option 2: detect torchao 0.17+ float8 base layers and decline to install `TorchaoLoraLinear`, falling back to the regular `LoraLinear` path) can be done as a separate change — this PR just unblocks the immediate `add_adapter` crash with the minimal surface area.

## Test plan
- [ ] Existing `tests/test_gpu_examples.py -k TestPeftTorchao` still passes locally (CI will exercise it).
- [ ] Manual: `add_adapter(lora_config)` on a `Float8WeightOnlyConfig`-quantized `nn.Linear` no longer raises `TypeError` (reproducer in #3223).
- [ ] Manual: `merge()` without `get_apply_tensor_subclass` now raises the new actionable `ValueError`.

---
AI-assisted, human reviewed.


BenjaminBossan:
Thanks for the PR. I investigated a bit further. The reason why the error occurs is because the base model was not quantized with Transformers. In PEFT, we expect something like:

```python
quantization_config = TorchAoConfig(quant_type=Int8WeightOnlyConfig())
    model_id, quantization_config=quantization_config
)
```

This is where we get the `get_apply_tensor_subclass` from:

https://github.com/huggingface/peft/blob/758cdac51922abbb24b6e772844c0a88bbe1cd7d/src/peft/tuners/lora/model.py#L234-L236

In your example, you directly quantize it with `quantize_`, which is why the argument is missing.

A few remarks:

1. We should raise a better error message in this case, pointing users towards using Transformers with `TorchAoConfig`.
2. You are using `Float8WeightOnlyConfig`. We haven't tested that quantization in PEFT, it may or may not work.
3. Your fix proposes to allow omitting the argument, which results in an error on merge/unmerge. Ideally, I would like to avoid that. Assume that a user trains a model for hours and then needs to merge, only to find it's not working. That would be a very bad experience.

What I think we could be doing:

1. Go with your PR. In addition, if `get_apply_tensor_subclass` is `None, during `__init__`, give a warning that merging/unmerging won't be possible. Also recommend to use Transformers `TorchAoConfig`.
2. Trying to determine the `get_apply_tensor_subclass` based on the type of the `weight`. I'm not sure if it's possible though.

Anai-Guo:
Thanks @BenjaminBossan — that's a much clearer picture, appreciate the investigation.

I've pushed an update that goes with option **(1)**: keep the optional `get_apply_tensor_subclass` so plain torchao-quantized bases can still do forward/training via `add_adapter`, but emit a `warnings.warn(...)` at `__init__` time when it's `None`, so users learn before training for hours that `merge()`/`unmerge()` will not work. The message points them at the supported `TorchAoConfig` path through Transformers.

I dropped the explanatory inline comment since the warning now carries the same information for users.

Re option (2): I had a quick look and I don't think it's reliably derivable in general — torchao's tensor subclasses (`AffineQuantizedTensor`, `Int8WeightOnlyConfig`-applied tensors, `Float8WeightOnlyConfig`, etc.) don't all expose the config that produced them in a single canonical attribute, and even when they do (e.g. `.tensor_impl.config` on some variants), reconstructing a callable equivalent to `int8_weight_only()` would require matching on private subclass internals that change between torchao releases. Happy to dig deeper if you'd like, but I'd lean on the warning + Transformers `TorchAoConfig` recommendation as the more durable path.

On `Float8WeightOnlyConfig`: noted — that's where the original report came from, but I agree it's outside the tested matrix. The warning is generic enough that the same guidance applies.

src/peft/tuners/lora/torchao.py
```diff
@@ -56,6 +62,13 @@ def merge(self, safe_merge: bool = False, adapter_names: Optional[list[str]] = N
             # no adapter to merge
             return
 
+        if self.get_apply_tensor_subclass is None:
+            raise ValueError(
+                f"{type(self).__name__} was instantiated without `get_apply_tensor_subclass`, which is "
+                "required to re-quantize the base layer after merging. Pass it via the adapter kwargs "
+                "(e.g. through `dispatch_torchao`) before calling merge()."
```
BenjaminBossan:
```suggestion
                "required to re-quantize the base layer after merging."
```

I think that asking users to pass the argument through `dispatch_torchao` is not very helpful, as it's an internal mechanism that the user cannot easily modify.

BenjaminBossan (changes requested):
Thanks for adding the warning. Let's add a test to ensure that we get the warning. For this, add a test to this test class:

https://github.com/huggingface/peft/blob/758cdac51922abbb24b6e772844c0a88bbe1cd7d/tests/test_gpu_examples.py#L4372

The test can be very simple: Load the base model, manually quantize it with `quantize_`, check that there is a warning. Then load it with `TorchAoConfig` and ensure that there is no warning. It's not necessary to parametrize that test over the `quant_type`.

Regarding option 2: I agree it's not very workable, so going with option 1 is fine.

Anai-Guo:
Added the test as requested in 614ac4d: `test_torchao_lora_warns_when_base_not_quantized_via_transformers` under `TestPeftTorchao`.

- Path 1 manually quantizes opt-125m with `quantize_(model, Int8WeightOnlyConfig())` and asserts the `get_apply_tensor_subclass` warning fires when wrapping with LoRA.
- Path 2 loads via `TorchAoConfig` and asserts no such warning is emitted.

Not parametrized over `quant_type`, per your note. Let me know if you'd prefer it in a different file or with a different marker.

tests/test_gpu_examples.py
```diff
@@ -4788,6 +4675,45 @@ def test_torchao_merge_layers_int8_dynamic_activation_int8_weight_raises(self):
         with pytest.raises(NotImplementedError, match=msg):
             model.merge_adapter()
 
+    @pytest.mark.single_gpu_tests
+    def test_torchao_lora_warns_when_base_not_quantized_via_transformers(self):
+        # Manually quantizing the base model with torchao.quantize_ leaves PEFT without
+        # `get_apply_tensor_subclass`, so the LoRA torchao linear emits a warning at init
+        # to flag that merge()/unmerge() will not work. When the base model is loaded via
+        # `TorchAoConfig`, PEFT recovers the requantization subclass and no warning fires.
+        import warnings as _warnings
```
BenjaminBossan: Why import `warnings` like this? Also, the import can be global.

tests/test_gpu_examples.py
```diff
@@ -4788,6 +4675,45 @@ def test_torchao_merge_layers_int8_dynamic_activation_int8_weight_raises(self):
...
+            lora_dropout=0.05,
+            bias="none",
+            task_type="CAUSAL_LM",
+        )
+
+        # Path 1: manual quantize_ -> warning expected.
+        model = AutoModelForCausalLM.from_pretrained(self.causal_lm_model_id, device_map=device)
+        quantize_(model, Int8WeightOnlyConfig())
+        with pytest.warns(UserWarning, match="get_apply_tensor_subclass"):
+            get_peft_model(model, lora_config)
```
BenjaminBossan: While we're at it, please also check the error when trying to merge.

BenjaminBossan (changes requested):
Two small comments, the rest looks good, thanks.

Anai-Guo:
Thanks for the review! Addressed both comments:
- moved `import warnings` to the global imports and use it directly
- added a `pytest.raises(ValueError, match=...)` on `peft_model.merge_adapter()` in the manual-quantize path

Also applied your earlier suggestion on `torchao.py` and dropped the `dispatch_torchao` hint from the merge error message.

BenjaminBossan:
Could you please run `make style`?

Anai-Guo:
Done — pushed ee04dd7. The diff was just a missing trailing newline in `src/peft/tuners/lora/torchao.py`; with the pinned `ruff~=0.12.8` and `doc-builder` from `make style` there were no other changes.

BenjaminBossan (approved):
Thanks for the updates, LGTM. Failing CI is unrelated.

PR #3235, opened 2026-05-14
fix(xlora): duplicated "for" in adapters warning message

vip892766gma:
One-line typo fix in `src/peft/tuners/xlora/config.py`: warning message "No value was provided for for `adapters`. This will be set to empty, please ensure that this is correct." → "No value was provided for `adapters`. This will be set to empty, please ensure that this is correct.". No code/behavior change.


BenjaminBossan (approved):
LGTM

PR #3241, opened 2026-05-19
Fix prepare model for compiled hotswap requires grad

BenjaminBossan:
There was a bug that after calling `prepare_model_for_compiled_hotswap`, the `requires_grad` of the newly created LoRA layer may have a different value from what it was previously. This is now fixed.


githubnemo (approved):
Thanks, LGTM

PR #3246, opened 2026-05-19
fix(tuners_utils): make BaseTunerLayer.bias settable (#3243)

### LINKED ISSUE #3243: lora.Linear.bias is a read-only   property — breaks any forward that does module.bias = X (legacy Jamba modeling)
### System Info

```
peft         == 0.19.1
transformers == 5.8.0
torch        == 2.11.0+cu130
python       == 3.12
```

### Who can help?

Anyone working on `peft/tuners/lora/layer.py` / `BaseTunerLayer`.

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

`BaseTunerLayer.bias` is defined as a **read-only `@property`** (no setter). Any modeling code that does `module.bias = X` mid-forward — a pattern used by several widely-distributed trust_remote_code Jamba modeling files on the HF Hub — breaks with:

```
AttributeError: property 'bias' of 'Linear' object has no setter
```

#### Minimal reproducer (no model download)

```python
import torch
from torch import nn
from peft import LoraConfig, inject_adapter_in_model


class Mixer(nn.Module):
    """Verbatim shape of ai21labs/Jamba-v0.1@9efd11575b modeling_jamba.py:856-861."""
    def __init__(self):
        super().__init__()
        self.dt_proj = nn.Linear(16, 16, bias=True)

    def forward(self, x):
        time_proj_bias = self.dt_proj.bias
        self.dt_proj.bias = None              # ← fires when dt_proj is LoRA-wrapped
        out = self.dt_proj(x)
        self.dt_proj.bias = time_proj_bias
        return out


x = torch.randn(1, 4, 16)

# Sanity: the pattern works on bare nn.Linear
assert Mixer()(x).shape == (1, 4, 16)
print("pre-wrap forward: OK")

# Wrap via the public PEFT API — same path get_peft_model uses internally
m = Mixer()
inject_adapter_in_model(LoraConfig(r=4, lora_alpha=8, target_modules=["dt_proj"]), m)
print(f"dt_proj after wrap: {type(m.dt_proj).__module__}.{type(m.dt_proj).__name__}")

m(x)
```

Output:

```
pre-wrap forward: OK
dt_proj after wrap: peft.tuners.lora.layer.Linear
  ...
  File ".../repro.py", line 21, in forward
    self.dt_proj.bias = None
    ^^^^^^^^^^^^^^^^^
  File ".../torch/nn/modules/module.py", line 2075, in __setattr__
    super().__setattr__(name, value)
AttributeError: property 'bias' of 'Linear' object has no setter
```

#### Root cause

`peft.tuners.tuners_utils.BaseTunerLayer.bias` is a `@property` with a getter only:

```python
# peft/tuners/tuners_utils.py
class BaseTunerLayer(ABC):
    @property
    def bias(self) -> nn.Parameter:
        ...
```

Any assignment `wrapped.bias = X` therefore fails on every `BaseTunerLayer` subclass (`lora.Linear`, `loha.Linear`, etc.).

#### Real-world impact

The pattern

```python
time_proj_bias = self.dt_proj.bias
self.dt_proj.bias = None
discrete_time_step = self.dt_proj(time_step).transpose(1, 2)
self.dt_proj.bias = time_proj_bias
```

is shipped by several Jamba-family `trust_remote_code` repos on the HF Hub whose bundled `modeling_jamba.py` predates the upstream switch to `bias.data = ...`. Examples:

- [`ai21labs/Jamba-v0.1`](https://huggingface.co/ai21labs/Jamba-v0.1/blob/9efd11575ba791d9e3d25d4c8b670e78506b2df7/modeling_jamba.py#L856-L861)
- `mlabonne/Jambatypus-v0.1`, `AirlockRIck/Jamba-v0.1`, and other forks of the same modeling code

For these repos, every PEFT LoRA wrap of `dt_proj` fails at forward time, on every transformers version, because the broken pattern lives in the Hub repo's frozen modeling file — the installed transformers version is irrelevant. The in-transformers Jamba implementation (`transformers/models/jamba/modeling_jamba.py`) has been updated to use `self.dt_proj.bias.data = torch.zeros_like(self.dt_proj.bias.data)`, which works correctly with the LoRA wrap; the issue is only with repos still pinned to the bundled legacy modeling file.

### Expected behavior

Either:

1. **Add `"jamba"` to PEFT's existing `mamba_model_types` exclusion** in `peft/tuners/tuners_utils.py:_check_lora_target_modules_mamba` (alongside `falcon_h1`, `mamba`, `mamba2`, `falcon_mamba`), and extend `incompatible_modules` to include `dt_proj` for `jamba`. This matches the existing defensive pattern, is a ~3-line change, and is safe for newer Jamba repos because their forwards no longer rely on the `bias = None` trick.

2. **Or:** give `BaseTunerLayer.bias` a setter that proxies to `self.base_layer.bias`. More general (would help any future caller doing `wrapped.bias = ...`), but a wider surface — needs review for interaction with `lora_bias` / DoRA / saved-state paths.

(1) is the minimum-blast-radius option and mirrors the existing `mamba_model_types` mechanism. Happy to send a PR if the direction is acceptable.

Anai-Guo:
Fixes #3243.

`BaseTunerLayer.bias` is declared as a read-only `@property`, so any modeling code that does `module.bias = X` raises `AttributeError: property 'bias' of 'Linear' object has no setter` once the layer is wrapped by a PEFT tuner. This breaks several legacy trust-remote-code Jamba modeling files on the Hub whose `JambaMambaMixer.forward` temporarily nulls out `self.dt_proj.bias` during the SSM step (see `ai21labs/Jamba-v0.1@9efd11575b` `modeling_jamba.py:856-861`).

This PR adds a setter that forwards assignment to the wrapped base layer, restoring the pre-PEFT `nn.Linear` contract while preserving the existing read path. No behavior change for code that only reads `module.bias`.

### Repro (from the issue)

```python
import torch
from torch import nn
from peft import LoraConfig, inject_adapter_in_model

class Mixer(nn.Module):
    def __init__(self):
        super().__init__()
        self.dt_proj = nn.Linear(16, 16, bias=True)
    def forward(self, x):
        time_proj_bias = self.dt_proj.bias
        self.dt_proj.bias = None
        out = self.dt_proj(x)
        self.dt_proj.bias = time_proj_bias
        return out

m = Mixer()
inject_adapter_in_model(LoraConfig(r=4, lora_alpha=8, target_modules=["dt_proj"]), m)
m(torch.randn(1, 4, 16))  # was: AttributeError; now: passes
```

---
AI-assisted, human reviewed.

### DIFF
--- a/src/peft/tuners/tuners_utils.py
+++ b/src/peft/tuners/tuners_utils.py
@@ -1472,6 +1472,15 @@ class BaseTunerLayer(ABC):
         base_layer = self.get_base_layer()
         return base_layer.bias
 
+    @bias.setter
+    def bias(self, value) -> None:
+        # Forward assignment to the wrapped base layer so that modeling code
+        # which does ``module.bias = X`` (e.g. legacy Jamba modeling files that
+        # temporarily null out ``dt_proj.bias`` during the SSM forward) keeps
+        # working after the layer is wrapped by a PEFT tuner.
+        base_layer = self.get_base_layer()
+        base_layer.bias = value
+
     def merge(self, safe_merge: bool = False, adapter_names: Optional[list[str]] = None) -> None:
         raise NotImplementedError
 


BenjaminBossan:
I feel that the use case is actually too narrow to warrant supporting in PEFT. There is an easy way to [define your own dispatch logic](https://huggingface.co/docs/peft/v0.19.0/en/developer_guides/custom_models#experimental-support-for-dynamic-dispatch-of-custom-modules-in-lora). You could use that to pass your own `lora.Linear` subclass that implements the setter for `bias`. If there is more demand for such a feature, e.g. from official Transformers models, we can revisit the topic.

PR #3247, opened 2026-05-20
[BUG FIX] tuners/lora/eva: fix label mask guard using 'in' instead of hasattr for dict

Dev-X25874:
## What

This is a clean re-open of #3233 (closed due to formatting issues across commits).
All review comments from @BenjaminBossan have been addressed from the
start in this PR.

In `prepare_model_inputs_fn_language_modeling`, the label mask guard uses
`hasattr(model_input, "labels")` to check whether labels are present. However,
`model_input` is a plain Python `dict` (enforced by the `isinstance` check three
lines above), and `hasattr` on a plain dict always returns `False` regardless of
its keys — dict keys are not exposed as object attributes.

This means the label masking branch (`use_label_mask=True`) has silently never
executed, even though it is the default in `EvaConfig`. As a result, ignored /
padding positions (label value `−100`) were included in the SVD computation
during EVA initialization.

## Fix

Replace `hasattr(model_input, "labels")` with `("labels" in model_input)`, which
is the correct Python idiom for dict key membership. The body of the branch
already uses `model_input["labels"]` (subscript access), so only the guard was
wrong. No downstream side-effects for callers that do not include a `"labels"` key.

```python
# Before
if peft_config.eva_config.use_label_mask and hasattr(model_input, "labels"):

# After
if peft_config.eva_config.use_label_mask and ("labels" in model_input):
```

## Tests

Added `test_eva_label_mask`, parametrized over `use_label_mask=True/False`, which
runs `get_eva_state_dict` on a dataset augmented with a `labels` column and
asserts the returned state dict is non-empty in both cases. This directly exercises
the previously dead code path.

cc @BenjaminBossan @sirluk

### DIFF
--- a/src/peft/tuners/lora/eva.py
+++ b/src/peft/tuners/lora/eva.py
@@ -249,7 +249,7 @@ def prepare_model_inputs_fn_language_modeling(model_input, peft_config: LoraConf
     if not isinstance(model_input, dict):
         raise TypeError("When using `prepare_model_inputs_fn_language_modeling` inputs must be a dictionary")
     mask = model_input.get("attention_mask", torch.ones_like(model_input["input_ids"])).bool()
-    if peft_config.eva_config.use_label_mask and hasattr(model_input, "labels"):
+    if peft_config.eva_config.use_label_mask and ("labels" in model_input):
         mask = torch.logical_and(mask, model_input["labels"] != peft_config.eva_config.label_mask_value)
     return mask.nonzero()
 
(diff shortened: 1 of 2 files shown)


BenjaminBossan (approved):
Okay, looks good from my side. Let's wait a week or so to see if Sirluk has time to check. Otherwise, I'll go ahead and merge.

BenjaminBossan:
No reply in a week, so I merged. Thanks again for the contribution.

PR #3249, opened 2026-05-20
DOC Fix incorrect imports in helpers.py docstring examples

blipbyte:
Four docstring examples in `src/peft/helpers.py` raise `ImportError` on their first import line, so a user copy-pasting them dies immediately.

Three (`update_forward_signature`, `update_generate_signature`, `update_signature`) import from `peft` top-level; the functions are only available via `peft.helpers`. The fourth (`DoraCaching` docstring at line 276) imports `enable_dora_scaling`, a name that doesn't exist anywhere in the codebase — should be `DoraCaching`, the class the docstring is for.

Found by running `pytest --doctest-modules src/peft/helpers.py`. After the fix the import lines no longer fail; examples advance past them and execute end-to-end.

Out of scope: the fifth doctest failure (`rescale_adapter_scale` line 194, undefined `ModelWithLoraLayer` placeholder) is a different bug class.

### DIFF
--- a/src/peft/helpers.py
+++ b/src/peft/helpers.py
@@ -44,7 +44,8 @@ def update_forward_signature(model: PeftModel) -> None:
 
     ```python
     >>> from transformers import WhisperForConditionalGeneration
-    >>> from peft import get_peft_model, LoraConfig, update_forward_signature
+    >>> from peft import get_peft_model, LoraConfig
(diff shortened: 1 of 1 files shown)


githubnemo (approved):
Nice, thanks for the fixes :)

PR #3251, opened 2026-05-21
Fail loud on save_pretrained() for unsharded LoRA tensors

akshansh47:
## What

Refuse to write LoRA adapters whose `lora_A` / `lora_B` tensors look unsharded (1-D or zero-sized). This is the canonical signature of an export that ran without gathering DeepSpeed ZeRO-3 / FSDP shards: the on-disk artifact looks structurally valid (filenames + `adapter_config.json` present), but downstream loaders fail with confusing index errors at the first attempted use, e.g. `IndexError: too many indices for tensor of dimension 1` in vLLM's `slice_lora_b` during hot-swap.

The new `_validate_lora_adapter_state_dict` helper surfaces the failure at `save_pretrained()` time with an actionable hint pointing at `deepspeed.zero.GatheredParameters` / `FullyShardedDataParallel.summon_full_params`, instead of corrupting the artifact and deferring the crash.

## Why

This failure mode is the most common cause of "my LoRA loads fine in HF/Transformers but breaks in vLLM" reports. It's currently silent. Examples:

- vllm-project/vllm#28640 — Qwen3-VL-8B `AssertionError` in `lora_shrink_op`
- huggingface/transformers#45313 — Qwen3.5 ZeRO-3 weight-gather failure mode

Cost of the current behavior is hours of debug pointed at the wrong layer (usually vLLM, sometimes the model). The validator turns it into a 30-second fix.

## Scope

The validator only inspects `.lora_A` / `.lora_B` tensors. Legitimately 1-D parameters such as DoRA's `lora_magnitude_vector` and AdaLoRA's `lora_E` are not affected. Non-LoRA adapter types (BoFT, OFT, P-tuning, prefix tuning, prompt tuning, etc.) are not touched.

## Tests

`tests/test_initialization.py::TestSaveValidatesLoraShapes` — 7 cases:

- happy path (well-formed state dict + DoRA magnitude + non-LoRA bias)
- 1-D `lora_A` raises (parameterized: shape `(0,)` and shape `(rank,)`)
- 0-sized `lora_B` raises
- error message includes adapter name
- end-to-end via `save_pretrained(state_dict=...)` raises
- end-to-end happy path still writes `adapter_config.json`

Adjacent test classes (`TestLoraInitialization`, `TestNoInfiniteRecursionDeepspeed`) verified passing locally: 121/121 + 7/7 new = **128/128**. `make quality` clean.

## Backwards compatibility

Pure addition. No public API changes; the helper is `_`-prefixed and only invoked from inside `save_pretrained`. The only behavior change for existing code is that what was previously silent corruption now raises with an actionable message — which is the intent of the patch.

Made with [Cursor](https://cursor.com)


githubnemo (changes requested):
- I don't think that we can say "this failure mode is the most common cause of [...]" but I agree that it would be helpful to raise if we know that the user is definitely doing something wrong. I am not yet sure if we can detect this without opening the door for false-positives, though.
- Let's write a reproducer first, the tests do not test the actual error case, they just proxy it. That's OK but we should have at least a reproducer for the issue first to explore the solution space.
- In any case I think that we should not check the state dict when it is passed from the outside, only if we infer it ourselves, I am certain that we can't be smarter than the user in that case.

shipbehaves:
ran a real reproducer for this (2x A10G, torch 2.5.1, tiny llama + lora on q_proj/v_proj). it reproduces on the *default* save path, not just hand-crafted tensors:

```
naive    [(128,)]    # ndim 1, broken
gathered [(8, 16)]   # ndim 2, correct
```

so a plain `save_pretrained()` on an FSDP-wrapped peft model without gathering writes 1-D `lora_A`/`lora_B`, exactly the `len(shape) < 2` signature this PR keys on. saving inside `FSDP.summon_full_params(...)` round-trips to proper 2-D.

minimal repro:

```python
# torchrun --nproc_per_node=2 repro.py
import os, torch, torch.distributed as dist
from torch.distributed.fsdp import FullyShardedDataParallel as FSDP

rank = int(os.environ["LOCAL_RANK"]); torch.cuda.set_device(rank)
dist.init_process_group("nccl")

def build():
    m = AutoModelForCausalLM.from_pretrained("hf-internal-testing/tiny-random-LlamaForCausalLM")
    m = get_peft_model(m, LoraConfig(r=8, target_modules=["q_proj", "v_proj"]))
    return FSDP(m.cuda(), use_orig_params=True)

def shapes(p):
    sd = load_file(f"{p}/adapter_model.safetensors")
    return [tuple(v.shape) for k, v in sd.items() if ".lora_A" in k][:1]

m = build()
if rank == 0: m.save_pretrained("out_naive")
dist.barrier()
m2 = build()
with FSDP.summon_full_params(m2):
    if rank == 0: m2.save_pretrained("out_gathered")
dist.barrier()
if rank == 0:
    print("naive   ", shapes("out_naive"))     # [(128,)]  broken
    print("gathered", shapes("out_gathered"))  # [(8, 16)] correct
dist.destroy_process_group()
```

on the false-positive worry: for peft+fsdp the input space is narrower than it looks. `use_orig_params=False` isn't even constructible, FSDP init raises `ValueError: Must flatten tensors with uniform requires_grad when use_orig_params=False` because lora freezes the base weights (mixed requires_grad). so the only valid peft+fsdp config is `use_orig_params=True`, which is the one that emits 1-D on a naive save. a correct gathered save is always 2-D. so `len(shape) < 2 or any(d == 0)`, scoped to inferred state dicts, looks safe against real adapters: legit is 2-D, the broken save is 1-D (what i hit here), and the `any(d == 0)` arm covers the zero-sized shard a non-owning rank emits under deepspeed zero-3 / wider sharding.

happy to fold this into the tests in place of the hand-crafted 1-D tensors if useful. (also both linked upstream issues, vllm#28640 / transformers#45313, are closed now.)

src/peft/peft_model.py
```diff
@@ -319,6 +320,14 @@ def save_mutated_as_lora(peft_config, path_initial_model_for_weight_conversion,
...
                 save_embedding_layers=save_embedding_layers,
             )
+
+            # Refuse to write LoRA adapters whose lora_A / lora_B tensors look unsharded (1-D or zero-sized). This is
+            # the canonical signature of an export that ran without gathering DeepSpeed ZeRO-3 / FSDP shards; the
+            # artifact otherwise looks valid on disk but breaks downstream loaders such as vLLM hot-swap with a
+            # confusing IndexError. Surface the failure here, where the caller can act on it. See
+            # vllm-project/vllm#28640 and huggingface/transformers#45313 for the historical failure mode.
+            _validate_lora_adapter_state_dict(output_state_dict, adapter_name=adapter_name)
+
```
githubnemo: Let's move this to `get_peft_model_state_dict` since this is method specific and `get_peft_model_state_dict` already does a method-level distinction for several cases.

Make sure to disable the check when `state_dict` is passed (i.e. is not `None`).

tests/test_initialization.py
githubnemo, line 1: Let's move this to `test_low_level_api.py` keeping the separate class.

tests/test_initialization.py
```diff
@@ -2490,6 +2490,91 @@ def test_no_infinite_recursion(self, cls, model, wrap_init):
...
+    def test_save_pretrained_succeeds_for_normal_lora(self, tmp_path):
+        # Sanity: the validator does NOT regress the happy path.
+        class _Tiny(nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.linear = nn.Linear(8, 8)
+
+        peft_model = get_peft_model(_Tiny(), LoraConfig(target_modules=["linear"], r=4))
+        peft_model.save_pretrained(tmp_path / "ok-adapter")
+        assert (tmp_path / "ok-adapter" / "adapter_config.json").exists()
```
githubnemo: Let's only keep `test_save_pretrained_raises_for_unsharded_state_dict` and `test_save_pretrained_succeeds_for_normal_lora`.

Since we want to disable the check when an outside state dict is detected, we can instead write to `_Tiny().linear.lora_A['default']` directly, overwriting the existing tensor with a flat one.

tests/test_initialization.py
```diff
@@ -2490,6 +2490,91 @@ def test_no_infinite_recursion(self, cls, model, wrap_init):
             cls.__init__ = original_init
 
 
+class TestSaveValidatesLoraShapes:
```
githubnemo:
```suggestion
class TestStateDictFsdpShapeValidation:
```

tests/test_initialization.py
```diff
@@ -2490,6 +2490,91 @@ def test_no_infinite_recursion(self, cls, model, wrap_init):
...
+    # Regression for the silent-corruption failure mode described in
+    # vllm-project/vllm#28640 and huggingface/transformers#45313: when PEFT runs save_pretrained() under
+    # DeepSpeed ZeRO-3 / FSDP without first gathering shards, lora_A / lora_B end up 1-D or shape (0,) on disk.
+    # The artifact looks valid but breaks downstream loaders such as vLLM hot-swap with a confusing IndexError.
+    # _validate_lora_adapter_state_dict raises at write time with an actionable hint instead.
```
githubnemo:
```suggestion
    # Mitigation for a potential error in DeepSpeed ZeRO-3 / FSDP settings where the
    # gather context was forgot when retrieving the state dict. See #3251 
    # for details.
```

githubnemo (changes requested):
Thanks for the discussion. I think this is fine as long as we limit the checks to `lora_A` and `_B` (as is the case now).

Let's also use the reproducer from https://github.com/huggingface/peft/pull/3251#issuecomment-4779365018 and make it into a test in `test_gpu_examples.py`.

shipbehaves:
glad this is useful, and the test_gpu_examples.py placement makes sense. @akshansh47 if it helps, i can contribute the gpu reproducer as `test_save_pretrained_raises_for_unsharded_state_dict` plus the normal-lora success case, following the structure @githubnemo laid out. i have it running on 2×A10G. your PR though, so whatever is easiest for you.

akshansh47:
Thanks for the review @githubnemo — pushed the changes:

- Moved the shape check out of `PeftModel.save_pretrained` into `get_peft_model_state_dict`, alongside the existing method-level handling.
- It now only runs for internally-inferred state dicts; when an explicit `state_dict` is passed it's skipped, since we can't reason about its provenance.
- Tests collapsed to `test_save_pretrained_raises_for_unsharded_state_dict` and `test_save_pretrained_succeeds_for_normal_lora` in `test_low_level_api.py`, in class `TestStateDictFsdpShapeValidation`. The broken case is now produced by overwriting `lora_A["default"]` with a flat tensor and letting `save_pretrained` infer the state dict, rather than hand-crafting one.
- Reworded the mitigation comment as suggested.

For what it's worth, I also confirmed the underlying failure on a real ZeRO-3 run (single T4): an ungathered `deepspeed.zero.Init` save writes `lora_A`/`lora_B` as `shape=(0,)`, and loading that artifact back fails with a size mismatch; saving inside `GatheredParameters` round-trips to the correct 2-D shapes. Matches @shipbehaves' FSDP reproduction.

shipbehaves:
nice, the zero-3 (0,) result closes the loop: fsdp gives the 1-D flat shard, zero-3 gives (0,) on the non-owning rank, so both arms of `len(shape) < 2 or any(d == 0)` now have a real repro behind them. scoping it to inferred state dicts (skipping user-supplied) is the right guard. reads clean.

akshansh47:
Pushed the GPU reproducer as `TestUnshardedLoRASaveUnderZeRO3` in `test_gpu_examples.py` — single-GPU DeepSpeed ZeRO-3 (`deepspeed.zero.Init` stage 3), asserting the ungathered `save_pretrained()` now raises at write time and that the `GatheredParameters` workaround still saves a correct 2-D adapter. It follows the existing in-process single-GPU pattern (`@require_torch_gpu` + `single_gpu_tests`, `pytest.importorskip("deepspeed")`).

This exercises the `any(d == 0)` arm of the guard (ZeRO-3 emits a zero-sized shard on the non-owning rank); @shipbehaves' FSDP repro exercises the `len(shape) < 2` arm (naive FSDP save emits a 1-D shard), so both branches now have a real repro behind them.

Also addressed the earlier note: softened the docstring's "almost always" to "typically", and the checks stay scoped to `lora_A`/`lora_B` on inferred state dicts only.

akshansh47:
@githubnemo this was auto-closed by the stale bot while awaiting re-review — could you reopen? State at close: all requested changes are in — validation moved into `get_peft_model_state_dict` (scoped to internally-inferred state dicts only), the docstring causal claim softened, and the DeepSpeed ZeRO-3 GPU reproducer you asked for is pushed as `TestUnshardedLoRASaveUnderZeRO3` in `test_gpu_examples.py` (single-GPU `zero.Init` stage 3: asserts the ungathered save now raises at write time and the `GatheredParameters` workaround still saves a correct 2-D adapter). Happy to rebase on main if anything has drifted.

akshansh47:
Thanks for reopening @githubnemo! Rebased onto current main (39de17c9) — the only conflict was an append-collision in `tests/test_low_level_api.py` with the new `TestGetBaseModelStateDict` class; both classes are kept. `TestStateDictFsdpShapeValidation` passes locally on the rebased branch, and the rest of `test_low_level_api.py` is green except `test_inject_from_state_dict_model_with_weight_conversion`, which fails identically on clean upstream main in my environment (env-related, unrelated to this PR). Ready for re-review.

src/peft/utils/save_and_load.py
```diff
@@ -45,6 +45,58 @@
...
+        return
+
+    preview = ", ".join(f"{n}:{s}" for n, s in bad[:3])
+    suffix = f" (+{len(bad) - 3} more)" if len(bad) > 3 else ""
+    raise ValueError(
+        f"Adapter {adapter_name!r}: {len(bad)} LoRA tensor(s) have invalid shape: "
+        f"{preview}{suffix}. A 1-D or zero-sized lora_A / lora_B tensor indicates that DeepSpeed ZeRO-3 / FSDP shards "
+        "were not gathered before save_pretrained() — downstream loaders such as vLLM hot-swap will then fail with "
+        "IndexError at first use. Wrap the save in deepspeed.zero.GatheredParameters([...], modifier_rank=None) for "
+        "ZeRO-3, or torch.distributed.fsdp.FullyShardedDataParallel.summon_full_params(model) for FSDP, before "
```
githubnemo:
```suggestion
        "were not gathered before save_pretrained() and are therefore incomplete. "
        "Wrap the save in deepspeed.zero.GatheredParameters([...], modifier_rank=None) for "
        "ZeRO-3, or torch.distributed.fsdp.FullyShardedDataParallel.summon_full_params(model) for FSDP, before "
```

src/peft/utils/save_and_load.py
```diff
@@ -45,6 +45,58 @@
...
+        shape = tuple(tensor.shape)
+        if len(shape) < 2 or any(d == 0 for d in shape):
+            bad.append((name, shape))
+
+    if not bad:
+        return
+
+    preview = ", ".join(f"{n}:{s}" for n, s in bad[:3])
+    suffix = f" (+{len(bad) - 3} more)" if len(bad) > 3 else ""
+    raise ValueError(
```
githubnemo: I think this should be a warning instead of a `ValueError`. It can very well be that the user doesn't care for gathered values and this would break existing code. We could also add a flag but that has the drawback that we don't know whether the user controls that portion of the code.

src/peft/utils/save_and_load.py
```diff
@@ -45,6 +45,58 @@
...
+    deferring the crash to whatever later loads it. Only the ``lora_A`` / ``lora_B`` keys are inspected; legitimately
+    1-D parameters such as DoRA's ``lora_magnitude_vector`` and AdaLoRA's ``lora_E`` are not affected.
+
+    Args:
+        state_dict: LoRA-only state dict that will be written to disk.
+        adapter_name: Adapter name being saved, used in the error message only.
+
+    Raises:
+        ValueError: When at least one ``lora_A`` / ``lora_B`` tensor is empty or has fewer than 2 dimensions.
+    """
```
githubnemo, line 70:
```suggestion
    """Refuse to save a LoRA adapter state dict whose ``lora_A`` / ``lora_B`` tensors look unsharded.

    A correctly-saved LoRA adapter has ``lora_A`` of shape ``(rank, in_dim)`` and ``lora_B`` of shape ``(out_dim,
    rank)`` - both 2-D and non-empty. If any such tensor is 1-D or has a zero-sized dimension, it typically indicates
    a partitioned-parameter export under DeepSpeed ZeRO-3 / FSDP that did not gather model shards before
    ``save_pretrained``. The artifact looks structurally valid (filenames + ``adapter_config.json`` present),
    but downstream loaders fail due to the tensor being incomplete.

    Raise a warning so the failure surfaces at write time, with an actionable hint, instead of corrupting the artifact and
    deferring the crash to whatever later loads it. Only the ``lora_A`` / ``lora_B`` keys are inspected; legitimately
    1-D parameters such as DoRA's ``lora_magnitude_vector`` and AdaLoRA's ``lora_E`` are not affected.

    Args:
        state_dict: LoRA-only state dict that will be written to disk.
        adapter_name: Adapter name being saved, used in the error message only.

    Warns:
        PeftWarning: When at least one ``lora_A`` / ``lora_B`` tensor is empty or has fewer than 2 dimensions.
    """
```

tests/test_gpu_examples.py
```diff
@@ -5032,6 +5032,65 @@ def test_fsdp_auto_wrap_policy_does_not_raise_on_custom_model(self):
...
+    """
+
+    class _Tiny(torch.nn.Module):
+        def __init__(self, d=128, v=256):
+            super().__init__()
+            self.embed = torch.nn.Embedding(v, d)
+            self.proj = torch.nn.Linear(d, d, bias=False)
+
+        def forward(self, input_ids):
+            return self.proj(self.embed(input_ids))
```
githubnemo: can we use the already defined `SimpleModel` instead?

tests/test_low_level_api.py
```diff
@@ -784,3 +784,26 @@ def test_multiple_adapters(self):
...
+class TestStateDictFsdpShapeValidation:
+    # Mitigation for a potential error in DeepSpeed ZeRO-3 / FSDP settings where the
+    # gather context was forgot when retrieving the state dict. See #3251
+    # for details.
+
+    class _Tiny(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.linear = nn.Linear(8, 8)
+
```
githubnemo: Can we use `DummyModel` instead?

githubnemo (commented):
Thanks for addressing the comments.

I think having a warning instead of raising an exception is the safer bet to not break user code all of a sudden. It should still be effective enough in warning about the possibly broken checkpoint.

Let's rebase once again since #3490 is now merged to iron out the last kinks.

akshansh47:
Thanks @githubnemo — addressed in 234ecf8c:

1. **Warning instead of `ValueError`** via `PeftWarning` (still actionable; does not break callers that skip gathering)
2. Docstring / message wording per your suggestions
3. CPU test now uses `DummyModel`; GPU ZeRO-3 test uses `SimpleModel`
4. Rebased onto current main (post-#3490)

Ready for another look.

akshansh47:
Rebased onto current main — only conflict was an import-list append (`PveraConfig` vs `PeftWarning`); both kept. Still warning-not-raise as requested.

githubnemo (approved):
The changes LGTM, thanks. I've confirmed that the GPU tests work. I will merge this once the CI is through.

PR #3255, opened 2026-05-23
Return False from is_gptqmodel_available when gptqmodel isn't installed

c-tonneslan:
Found this while reading `import_utils.py`.

`is_gptqmodel_available` is the only `is_*_available` helper that doesn't explicitly return `False` when its package is missing — the whole body is nested inside `if find_spec("gptqmodel") is not None:` with no return on the false branch, so it falls off the end and returns `None`:

```pycon
>>> from peft.import_utils import is_gptqmodel_available
>>> is_gptqmodel_available()
>>> # implicit None, not False
>>> type(_)
<class 'NoneType'>
```

Compare with the siblings:

```pycon
>>> from peft.import_utils import is_bnb_available, is_aqlm_available
>>> is_bnb_available(), is_aqlm_available()
(False, False)
```

Truthiness mostly hides this (\`not None\` and \`not False\` are both True), so \`tests/testing_utils.py\`'s \`not is_gptqmodel_available()\` skip-marker still does the right thing. But the annotation says \`-> bool\`, type checkers complain, and any caller doing \`is False\` / \`is True\` / \`assert ... is True\` would silently take the wrong branch.

Flipped the outer if to an early-return False — also drops one level of nesting as a bonus, no behavior change for the existing branches.

### DIFF
--- a/src/peft/import_utils.py
+++ b/src/peft/import_utils.py
@@ -49,38 +49,40 @@ def is_bnb_4bit_available() -> bool:
 
 @lru_cache
 def is_gptqmodel_available() -> bool:
-    if importlib.util.find_spec("gptqmodel") is not None:
-        GPTQMODEL_MINIMUM_VERSION = packaging.version.parse("7.0.0")
-        OPTIMUM_MINIMUM_VERSION = packaging.version.parse("1.24.0")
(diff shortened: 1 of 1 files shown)


githubnemo (approved):
Thanks, looks reasonable and is easier to understand.

PR #3256, opened 2026-05-23
Add PVeRA rank ablation benchmark experiments

anrohanro:
Related to #2310

## Summary

This PR adds a rank64 PVeRA benchmark experiment to the MetaMathQA/GSM8K method comparison suite.

The experiment was generated by modifying only the `r` parameter from the existing PVeRA baseline configuration to preserve benchmark equivalence.

## Result

| Rank | Test Accuracy | Trainable Params |
| --- | --- | --- |
| 64 | 0.3199 | 121,856 |
| 256 (baseline) | 0.3306 | 143,360 |

The rank64 configuration achieves comparable performance while reducing the number of trainable parameters relative to the baseline configuration.


BenjaminBossan:
@anrohanro Thanks for running more experiments with PVeRA. AFAICT, however, they don't seem to improve on the existing result using the default config. E.g. they are not better when it comes to test accuracy. Did you see any metric  where we see a Pareto improvement over the default? If not, let's not add more experiments just for the sake of it.

Moreover:

1. Let's only add one more experiment, not multiple.
2. Let's not check in the results, only the experiment settings (but share the results in the PR).

anrohanro:
@BenjaminBossan Thanks for the feedback. I did not observe a clear Pareto improvement over the default configuration in terms of test accuracy. As suggested, I kept only the rank64 experiment configuration, removed the additional experiments/results from the commit, and updated the PR description with the summarized benchmark result.

method_comparison/MetaMathQA/experiments/pvera/llama-3.2-3B-rank64/training_params.json
```diff
@@ -0,0 +1,6 @@
+{
+  "optimizer_kwargs": {
+    "lr": 3e-3
+  },
+  "autocast_adapter_dtype": false
```
BenjaminBossan: Let's remove this argument. That way, this experiment is more consistent with the other experiments.

BenjaminBossan (changes requested):
Thanks for removing the other experiments. I just have one small suggestion left.

method_comparison/MetaMathQA/experiments/pvera/llama-3.2-3B-rank64/training_params.json
anrohanro: Done, I removed the `autocast_adapter_dtype` argument from the rank64 experiment configuration.

BenjaminBossan (approved):
Thanks for contributing this experiment. Failing CI is unrelated.

PR #3260, opened 2026-05-24
Mica

sr-networks:
Added support for MiCA as proposed in 

[https://github.com/huggingface/peft/issues/3142](https://github.com/huggingface/peft/issues/3142)

## What does this PR do?
This PR adds **MiCA (Minor Component Adaptation)** as a LoRA initialization variant, exposed via:
```python
LoraConfig(init_lora_weights="mica")
```
MiCA initializes LoRA from the minor singular subspace of the base weight matrix. For a target linear layer with weight matrix (W = U \Sigma V^T), MiCA initializes:

B = U[:, -r:]
A = 0

where B contains the r left singular vectors corresponding to the smallest singular values.

Since A is initialized to zero, the adapter contribution is zero at initialization and the base model output is preserved.

Why freeze lora_B?

MiCA treats lora_B as the fixed minor-component subspace and trains only lora_A.

Freezing lora_B preserves the intended MiCA constraint during training. Without this, the adapter would no longer remain constrained to the selected minor subspace and would behave more like an unconstrained LoRA update.

This is why this PR also updates the LoRA variant interface so that variants can customize adapter trainability.

## Recommended usage

MiCA is primarily intended for **continued pretraining / domain-adaptive pretraining**, not for instruction fine-tuning.

The recommended workflow is:

1. Start from the **base model**, not the instruct/chat model.

2. Train the MiCA adapter on continued-pretraining data.

3. Merge the trained adapter into the model weights.

4. Use the resulting merged model as the adapted base for subsequent instruction/chat tuning, or merge/apply it before using the corresponding instruct/chat model setup.

This recommendation follows the intended use of MiCA as a method for injecting domain knowledge into pretrained representations while constraining the update to the selected minor-component subspace.

## Main changes

* Adds "mica" as a valid value for LoraConfig(init_lora_weights=...).
* Adds SVD-based MiCA initialization for nn.Linear and nn.Embedding layers.
* Initializes B from the minor left singular vectors of the base weight.
* Initializes A to zero so the adapter is a no-op at initialization.
* Uses BaseTunerLayer.frozen_peft_weight_names to keep MiCA B frozen across get_peft_model, add_adapter, set_adapter, and set_requires_grad.
* Keeps standard LoRA forward, merge, and unmerge behavior.
* Adds MiCA tests, documentation, and a runnable fine-tuning example.

## Tests

This PR adds tests covering:

* zero adapter contribution at initialization
* use of the minor rather than major singular subspace
* B being frozen
* adapter switching with MiCA
* reusing an adapter name after deleting a MiCA adapter
* r > max_r error handling
* embedding initialization
* custom model behavior with MiCA

## Limitations

* Supports nn.Linear and nn.Embedding target modules.
* Requires r <= min(in_features, out_features) for linear layers.
* Requires r <= min(num_embeddings, embedding_dim) for embedding layers.
* Performs a full SVD during adapter initialization.


src/peft/tuners/lora/layer.py
```diff
@@ -99,6 +99,17 @@ def forward(
...
         raise NotImplementedError
 
+    @staticmethod
+    def update_requires_grad(module: LoraLayer, adapter_name: str) -> None:
+        """
+        Adjust `requires_grad` of the adapter parameters after the tuner's default trainability marking.
+
+        This hook is invoked from the LoRA model's `_mark_only_adapters_as_trainable` once for every adapter that uses
+        this variant. The default implementation is a no-op so existing variants keep their current behavior; variants
+        like MiCA override this to freeze a subset of the adapter parameters (e.g. `lora_B`).
```
BenjaminBossan: "The default implementation is a no-op..." No need to explain that, it can be seen plainly from the method body. Instead, explain when, in general, this should be overridden.

src/peft/tuners/lora/layer.py
```diff
@@ -395,6 +409,39 @@ def pissa_init(self, adapter_name, init_lora_weights):
...
+
+        Initializes `lora_B` from the `r` left singular vectors of the base weight associated with the smallest
+        singular values, and sets `lora_A` to zero. The `lora_B` matrix is frozen during training (see
+        `MiCALinearVariant.update_requires_grad`); only `lora_A` is updated. Because `lora_A == 0` at init, the adapter
+        contribution `B @ A == 0` and the base weight does not need to be modified to preserve the forward output.
+        """
+        # When the adapter is being created under `init_empty_weights` (e.g. low_cpu_mem_usage=True), its parameters
+        # live on the meta device and will be filled in from a checkpoint after creation. Skip the SVD in that case.
+        if self.lora_B[adapter_name].weight.device.type == "meta":
+            return
```
BenjaminBossan, line 416: Let's add a new line after return and after the `raises` below for better readability.

src/peft/tuners/lora/layer.py
```diff
@@ -395,6 +409,39 @@ def pissa_init(self, adapter_name, init_lora_weights):
...
+        # SVD: weight = V @ diag(S) @ Uh, with V: (out, k), Uh: (k, in), S sorted descending.
+        # MiCA selects the LAST r left singular vectors (smallest singular values) for B and zeroes A.
+        r = self.r[adapter_name]
+        max_r = min(weight.shape)
+        if r > max_r:
+            raise ValueError(
+                f"MiCA requires `r` <= min(in_features, out_features) but got r={r} for a layer with "
+                f"weight shape {tuple(weight.shape)} (max usable r is {max_r})."
+            )
+        V, _, _ = torch.linalg.svd(weight.data, full_matrices=False)
```
BenjaminBossan: Using `V` here is confusing:

```suggestion
        U, _, _ = torch.linalg.svd(weight.data, full_matrices=False)
```

tests/test_custom_models.py
```diff
@@ -101,6 +101,19 @@
     ),
     ("Vanilla MLP 7 LoRA with DoRA", "MLP", LoraConfig, {"target_modules": ["lin0"], "use_dora": True}),
     ("Vanilla MLP 8 LoRA with DoRA", "MLP", LoraConfig, {"target_modules": ["lin0", "lin1"], "use_dora": True}),
+    (
+        "Vanilla MLP 1 LoRA with MiCA",
+        "MLP",
+        LoraConfig,
+        {"target_modules": ["lin0"], "init_lora_weights": "mica", "r": 4},
+    ),
+    (
```
BenjaminBossan: One config should be enough here.

tests/test_initialization.py
```diff
@@ -389,6 +389,104 @@ def test_lora_pissa_linear_init_default(self, data):
...
+
+    def test_lora_mica_freezes_B(self):
+        model = self.get_model()
+        config = LoraConfig(init_lora_weights="mica", target_modules=["linear"], r=8)
+        peft_model = get_peft_model(deepcopy(model), config)
+
+        assert peft_model.base_model.linear.lora_A["default"].weight.requires_grad
+        assert not peft_model.base_model.linear.lora_B["default"].weight.requires_grad
+
+    def test_lora_mica_train_step_updates_only_A(self, data):
```
BenjaminBossan: This test is not necessary IMO, it basically just checks the same thing as the previous one.

tests/test_initialization.py
```diff
@@ -389,6 +389,104 @@ def test_lora_pissa_linear_init_default(self, data):
...
+        loss = out.sum()
+        loss.backward()
+        opt.step()
+
+        A_after = peft_model.base_model.linear.lora_A["default"].weight.detach()
+        B_after = peft_model.base_model.linear.lora_B["default"].weight.detach()
+        assert not torch.equal(A_before, A_after)
+        assert torch.equal(B_before, B_after)
+
+    def test_lora_mica_save_and_load(self, data, tmp_path):
```
BenjaminBossan: This test is not necessary IMO, we already test that elsewhere.

tests/test_initialization.py
```diff
@@ -389,6 +389,104 @@ def test_lora_pissa_linear_init_default(self, data):
...
+        peft_model.base_model.linear.lora_A["default"].weight.data.normal_()
+        output_mica = peft_model(data)[0]
+
+        peft_model.save_pretrained(tmp_path / "mica-model")
+        model_loaded = PeftModel.from_pretrained(deepcopy(model), tmp_path / "mica-model")
+        output_loaded = model_loaded(data)[0]
+
+        assert torch.allclose(output_mica, output_loaded, atol=1e-6)
+
+    def test_lora_mica_unsupported_layer_raises(self):
```
BenjaminBossan: This test is not necessary IMO, we already test that elsewhere.

tests/test_initialization.py
```diff
@@ -389,6 +389,104 @@ def test_lora_pissa_linear_init_default(self, data):
...
+        # MiCA is currently only implemented for nn.Linear. Trying to apply it to a non-Linear layer (e.g. conv2d)
+        # should raise rather than silently fall through to a default init.
+        model = self.get_model()
+        config = LoraConfig(init_lora_weights="mica", target_modules=["conv2d"], r=4)
+        with pytest.raises(ValueError, match="Unknown initialization"):
+            get_peft_model(deepcopy(model), config)
+
```
BenjaminBossan, line 522: Let's add a test for the error from the `if r > max_r` check. You could also add a test for the dtype error, but that's not as important.

BenjaminBossan (changes requested):
Thanks for this PR to add MiCA. For this PR, I have focused on the implementation, I haven't checked the example and documentation yet.

I have a couple of smaller comments, but there is also a larger issue, which is that the way that `requires_grad` is set is not sufficient yet. Right now, it only covers the `get_peft_model` path, but that's not the only one that can modify `requires_grad`. Take this example:

```python
from pprint import pprint
import torch
from torch import nn

class SimpleMlp(nn.Module):
    def __init__(self):
        super().__init__()
        self.fc1 = nn.Linear(10, 10)
        self.fc2 = nn.Linear(10, 10)

    def forward(self, x):
        x = torch.relu(self.fc1(x))
        x = self.fc2(x)
        return x

model = SimpleMlp()
config0 = LoraConfig(target_modules=["fc1"], init_lora_weights="mica")
model = get_peft_model(model, config0)

layers_with_requires_grad = [name for name, param in model.named_parameters() if param.requires_grad]
print("Layers with requires_grad=True after first LoRA")
pprint(layers_with_requires_grad)
# correct, should be ['base_model.model.fc1.lora_A.default.weight']

config1 = LoraConfig(target_modules=["fc1", "fc2"], init_lora_weights="mica")
model.set_adapter("other")
layers_with_requires_grad = [name for name, param in model.named_parameters() if param.requires_grad]
print("\nLayers with requires_grad=True after switching to other adapter")
pprint(layers_with_requires_grad)
# incorrect, should be ['base_model.model.fc1.lora_A.other.weight', 'base_model.model.fc2.lora_A.other.weight']

model.set_adapter("default")
layers_with_requires_grad = [name for name, param in model.named_parameters() if param.requires_grad]
print("\nLayers with requires_grad=True after switching back to default adapter")
pprint(layers_with_requires_grad)
# incorrect, should be ['base_model.model.fc1.lora_A.default.weight']
```

If you run this, you'll see that the `add_adapter` and `set_adapter` path are not covered.

Therefore, I have a different suggestion which implements this feature in a more declarative way, LMK what you think about that:

First, let's remove `update_requires_grad` completely. Next, on `peft.tuners.tuners_utils.BaseTunerLayer`, let's add a class attribute `frozen_peft_weight_names: dict[str, tuple[str, ...]] = {}`. This will contain a mapping from adapter name to the keys of the PEFT weights that should be frozen.

Second, in `MiCALinearVariant.init`, for the MiCA adapter, we add an entry to `frozen_peft_weight_names` for LoRA B. Let's ensure not to simply mutate `frozen_peft_weight_names`, as it's a class attribute. Instead, re-assign a `copy` of the mutated dict.

Finally, in `_mark_only_adapters_as_trainable` and in `peft.tuners.tuners_utils.set_adapter`, we can check if, for a given PEFT layer, there is an entry in `frozen_peft_weight_names`, and if we find it, set `requires_grad = False`.

Let's also add unit tests to ensure this works as expected, my example could serve as a template.

Moreover, MiCA currently wouldn't work for LoRA applied to embedding layers, right? I think it should be easy enough to add support for those.

sr-networks:
Thanks for the review. I pushed an update addressing the MiCA trainability issue with a declarative frozen_peft_weight_names mapping on BaseTunerLayer, removing update_requires_grad. The freeze is enforced through initial setup, adapter switching, and set_requires_grad.

I also added nn.Embedding support, removed redundant tests, added the r > max_r error test, and added a regression test for switching/reusing adapters.

src/peft/tuners/tuners_utils.py
```diff
@@ -1521,6 +1527,23 @@ def enable_adapters(self, enabled: bool) -> None:
                     _set_layer_requires_grad(layer, False)
             self._disable_adapters = True
 
+    def _freeze_declared_peft_weights(self, adapter_names: str | Sequence[str] | None = None) -> None:
```
BenjaminBossan: I would rename the function, as "declared" could be confusing. How about `_freeze_non_trainable_peft_weights`?

Moreover, let's add a docstring to describe what happens here and why. Also, let's add an early exit for the most common path:

```python
if not self.frozen_peft_weight_names:
    return
```

src/peft/tuners/lora/variants.py
```diff
@@ -1218,25 +1218,25 @@ def forward(
         return result
 
 
+def _register_frozen_peft_weight(module: Linear | Embedding, adapter_name: str, weight_name: str) -> None:
+    frozen_peft_weight_names = module.frozen_peft_weight_names.copy()
+    frozen_names = frozen_peft_weight_names.get(adapter_name, ())
+    frozen_peft_weight_names[adapter_name] = tuple(dict.fromkeys((*frozen_names, weight_name)))
```
BenjaminBossan: I'm not a fan of `tuple(dict.fromkeys(...))`. I think this should work:

```python
if `weight_name` not in frozen_peft_weight_names[adapter_name]:
    frozen_peft_weight_names[adapter_name] += (weight_name,)
```

Since it's a tuple, it won't be mutated in-place, instead creating a new tuple.

Also, let's add a short docstring to the function.

src/peft/tuners/lora/variants.py
```diff
@@ -1218,25 +1218,25 @@ def forward(
         return result
 
 
+def _register_frozen_peft_weight(module: Linear | Embedding, adapter_name: str, weight_name: str) -> None:
```
BenjaminBossan: Any `LoraLayer` would be a valid type, right?

tests/test_initialization.py
```diff
@@ -441,51 +467,56 @@ def test_lora_mica_freezes_B(self):
...
-        model_loaded = PeftModel.from_pretrained(deepcopy(model), tmp_path / "mica-model")
-        output_loaded = model_loaded(data)[0]
+        model.delete_adapter("other")
+        config2 = LoraConfig(target_modules=["fc1"], r=4)
+        model.add_adapter("other", config2)
+        model.set_adapter("other")
+        assert trainable_parameters(model) == [
+            "base_model.model.fc1.lora_A.other.weight",
+            "base_model.model.fc1.lora_B.other.weight",
+        ]
```
BenjaminBossan, line 508: Since you already took care of the logic for adapter deletion, let's add a check that deleting, say, "other" works as expected (e.g. updating `frozen_peft_weight_names`).

BenjaminBossan (changes requested):
Thanks for updating, I think the solution of how to deal with the frozen B parameter is now cleaner than before. Also thanks for taking care of further edge cases like adapter deletion. For this review, I focused on this new implementation and found a few small areas to improve it. Please check my comments.

BenjaminBossan:
Just a heads up, the docs were restructured in #3300, resulting in a merge conflict. You should now find the `lora.md` in `docs/source/package_reference/`.

sr-networks:
Thanks, I pushed an update addressing the remaining review feedback:

- moved the MiCA docs section to `docs/source/package_reference/lora.md` after the docs restructuring
- renamed `_freeze_declared_peft_weights` to `_freeze_non_trainable_peft_weights`, added a docstring, and added the early return
- generalized `_register_frozen_peft_weight` to `LoraLayer` and simplified the tuple update logic
- added an explicit test assertion that `frozen_peft_weight_names` is cleaned up after deleting the `"other"` adapter

I ran the MiCA-specific tests locally:

```bash
PYTHONPATH="$PWD/src" pytest tests/test_initialization.py -k mica -q

**Comment by BenjaminBossan** [comment]
@sr-networks Could you please run `make style`?

**Comment by sr-networks** [comment]
@BenjaminBossan I ran `make style`, pushed the resulting commit (`5571763e`), and verified locally that `make quality` now passes.

File: src/peft/tuners/tuners_utils.py [hunk 13]
```diff
@@ -1391,6 +1395,8 @@ class BaseTunerLayer(ABC):
     adapter_layer_names: tuple[str, ...] = ()
     # All names of other parameters that may contain adapter-related parameters
     other_param_names: tuple[str, ...] = ()
+    # Mapping from adapter name to adapter layer names that should always stay frozen
```
BenjaminBossan: As is, the description is a bit ambiguous because it can read like the mapping should stay frozen, which is not the case.

```suggestion
    # This dict maps from the adapter name to the layer(s) whose weights should stay frozen
```

examples/mica_finetuning/mica_finetuning.py
```diff
@@ -0,0 +1,80 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan:
```suggestion
# Copyright 2026-present the HuggingFace Inc. team.
```

BenjaminBossan (changes requested):
Thanks, everything looks good, I just found two very small issues as I was going through the code once more.

sr-networks:
@BenjaminBossan Thanks for your review.

BenjaminBossan:
Thanks for these changes @sr-networks, the code now looks good, CI is green.

Before merging, just a note: We now added a second benchmark to PEFT, which covers [image generation](https://github.com/huggingface/peft/tree/main/method_comparison/image-gen). If you want to add an experiment to that benchmark similar to the one you added for MetaMath, LMK and I'll wait for submission before merging the PR.

sr-networks:
Thanks, no need to wait for an image-generation benchmark from my side. I’m happy to proceed with this PR as-is.

BenjaminBossan (approved):
All right, thank you for adding MiCA, the PR LGTM.

sr-networks:
Great! thanks @BenjaminBossan

PR #3261, opened 2026-05-26
Avoid TP helper imports for non-TP distributed LoRA loads

jiqing-feng:
## Summary

Fix a `ModuleNotFoundError` in distributed LoRA adapter loading when `torch.distributed` is initialized but the model is not actually tensor-parallel sharded.

## Problem

`set_peft_model_state_dict` calls `_maybe_shard_state_dict_for_tp` whenever distributed is initialized. This also happens for regular DDP runs. The helper imported `transformers.integrations.tensor_parallel` before checking whether any LoRA layer had TP metadata, so non-TP distributed runs could fail immediately with:

```text
ModuleNotFoundError: No module named 'transformers.integrations.tensor_parallel'
```

## Fix

Check for real TP-sharded LoRA layers first by looking for `_hf_tp_plan` and `_hf_device_mesh` on the base layers. If none are found, return early. Only import Transformers TP helpers when there is actually something to shard.

This keeps the existing TP path intact while avoiding false-positive TP handling for DDP/non-TP workloads.

## Test

Added a regression test for a LoRA model with a model-level `_tp_plan` but no layer-level TP metadata:

```bash
python -m pytest -o addopts="" tests/test_other.py::test_maybe_shard_state_dict_for_tp_noops_without_tp_layers -q
```

Result: `1 passed`.

### DIFF
--- a/tests/test_other.py
+++ b/tests/test_other.py
@@ -29,6 +29,7 @@ from transformers import (
 from peft import LoraConfig, PeftModel, VeraConfig, get_peft_model
 from peft.import_utils import is_transformers_ge_v5_1_0, is_transformers_ge_v5_6_0
 from peft.utils.other import ModulesToSaveWrapper, _get_module_names_tied_with_embedding, _get_no_split_modules
+from peft.utils.save_and_load import _maybe_shard_state_dict_for_tp
 
 from .testing_utils import hub_online_once
 
@@ -87,6 +88,18 @@ def test_modules_to_save_targets_module_dict_raises(cls):
     msg = "modules_to_save cannot be applied to modules of type"
     with pytest.raises(TypeError, match=msg):
         get_peft_model(model=model, peft_config=peft_config)
+
+
+def test_maybe_shard_state_dict_for_tp_noops_without_tp_layers():
+    model = get_peft_model(ModelWithModuleDict(), LoraConfig(target_modules=["other_layer"]))
+    model._tp_plan = {"other_layer": "colwise"}
+    weight = torch.rand(8, 10)
+    state_dict = {"other_layer.lora_A.weight": weight}
+
+    _maybe_shard_state_dict_for_tp(model, state_dict, adapter_name="default")
+
+    assert list(state_dict) == ["other_layer.lora_A.weight"]
+    assert state_dict["other_layer.lora_A.weight"] is weight
 
 
 def test_get_peft_model_revision_warning(tmp_path):
--- a/tests/test_other.py
+++ b/tests/test_other.py
@@ -29,7 +29,7 @@ from transformers import (
 from peft import LoraConfig, PeftModel, VeraConfig, get_peft_model
 from peft.import_utils import is_transformers_ge_v5_1_0, is_transformers_ge_v5_6_0
 from peft.utils.other import ModulesToSaveWrapper, _get_module_names_tied_with_embedding, _get_no_split_modules
-from peft.utils.save_and_load import _maybe_shard_state_dict_for_tp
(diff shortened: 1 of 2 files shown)


jiqing-feng:
Hi @BenjaminBossan . Would you please review the PR? Thanks!

BenjaminBossan (commented):
Thank you for this PR. I agree that we can check for TP layers first before importing the TP related class, so that part of the PR looks good. What I don't quite get is the test. Under what circumstances would we expect the test to fail? Could you perhaps add a comment to explain?

jiqing-feng:
> Thank you for this PR. I agree that we can check for TP layers first before importing the TP related class, so that part of the PR looks good. What I don't quite get is the test. Under what circumstances would we expect the test to fail? Could you perhaps add a comment to explain?

Thanks for the feedback. I updated the test comment to make the intent clearer.

It is a unit test for `_maybe_shard_state_dict_for_tp`. It verifies that when no LoRA base layer has TP metadata (`_hf_tp_plan` / `_hf_device_mesh`), the helper is a no-op and leaves the state dict unchanged.

BenjaminBossan:
@jiqing-feng Thanks for updating, but I still don't understand. The test would pass even before your fix, right? So I'm not sure what exactly it's testing and it is certainly not a way to prevent regression.

jiqing-feng:
Hi @BenjaminBossan  
Thanks, you're right that the current test is not a good regression test because it would also pass before this change.

The actual issue I hit is specific to my environment with the latest Transformers: importing the old `transformers.integrations.tensor_parallel` module fails. In my DDP run, `torch.distributed` is initialized, so `set_peft_model_state_dict` enters `_maybe_shard_state_dict_for_tp` even though the model is not TP-sharded. The previous code imported the TP helpers before checking whether any LoRA layer actually had TP metadata, which caused the failure.
```
from transformers.integrations.tensor_parallel import ModuleNotFoundError: No module named 'transformers.integrations.tensor_parallel'
```

So the change is mainly to avoid entering/importing the TP-specific path when there are no TP LoRA layers. I agree that the current unit test does not really capture the regression in a clean way, since it depends on the Transformers TP import behavior. I can remove the test and keep the code change only.

tests/test_other.py
```diff
@@ -29,7 +29,6 @@
 from peft import LoraConfig, PeftModel, VeraConfig, get_peft_model
 from peft.import_utils import is_transformers_ge_v5_1_0, is_transformers_ge_v5_6_0
 from peft.utils.other import ModulesToSaveWrapper, _get_module_names_tied_with_embedding, _get_no_split_modules
-
```
BenjaminBossan, line 32: Let's undo this unrelated change.

BenjaminBossan (changes requested):
Thanks for explaining further.

> The actual issue I hit is specific to my environment with the latest Transformers: importing the old transformers.integrations.tensor_parallel module fails.

I checked the code and couldn't find anything that makes `transformers.integrations.tensor_parallel` conditional. Are you sure that this import fails?

jiqing-feng:
I just noticed the error has been fixed in the latest main branch of transformers.

BenjaminBossan:
> I just noticed the error has been fixed in the latest main branch of transformers.

So the PR is technically no longer necessary but:

1. It helps when users have older Transformers installed.
2. It's a small speed optimization.

Is that assessment right?

jiqing-feng:
> > I just noticed the error has been fixed in the latest main branch of transformers.
> 
> So the PR is technically no longer necessary but:
> 
> 1. It helps when users have older Transformers installed.
> 2. It's a small speed optimization.
> 
> Is that assessment right?

Yes.

BenjaminBossan (approved):
Thanks for clarifying, PR LGTM.

PR #3262, opened 2026-05-26
FIX Transformers weight conversion undo incorrect key renaming

BenjaminBossan:
See [discussion starting here](https://github.com/huggingface/peft/pull/3212#issuecomment-4402677775).

For some models like `CLIPTextModelWithProjection`, transformers would add a removal of the `'text_model.'` prefix to the conversions, but this removal is incorrect. Therefore, there is a logic in transformers to undo the removal if there is not entry in the `state_dict` for the renamed key:

https://github.com/huggingface/transformers/blob/027d1a97025295a1346c2eb5c361259e69eedfe7/src/transformers/core_model_loading.py#L1361-L1365

This logic was missing in PEFT, resulting in missing and unexpected keys.

This PR should fix the failing Diffusers tests. Failing CI is unrelated.

### DIFF
--- a/src/peft/utils/transformers_weight_conversion.py
+++ b/src/peft/utils/transformers_weight_conversion.py
@@ -508,6 +508,7 @@ def convert_peft_adapter_state_dict_for_transformers(
         # no architecture-level conversion registered for this model
         return adapter_state_dict
 
+    prefix: str = getattr(model, "base_model_prefix", "")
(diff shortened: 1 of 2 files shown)


sayakpaul (approved):

Cyrilvallez (approved):
Ha yes we indeed have this!

PR #3268, opened 2026-05-27
Fix adapter filtering for substring-matching names

Chessing234:
## Bug
Saving a LoRA adapter state dict can pick up weights from a different adapter when the requested adapter name is a prefix of another adapter name, such as `default` and `default_extended`.

## Root cause
`get_peft_model_state_dict()` filtered LoRA keys with a plain substring check (`adapter_name in k`), which also matched other adapter-name components.

## Why this fix is correct
The filter now requires an exact adapter-name path component (`.{adapter_name}.` or a terminal `.{adapter_name}`), so only the requested adapter's weights are retained. The regression test initializes two adapters with different values and verifies that saving `default` only returns `default` weights.

### DIFF
--- a/tests/test_low_level_api.py
+++ b/tests/test_low_level_api.py
@@ -520,20 +520,18 @@ class TestPeftStateDict:
             assert ("v_proj" in key) or ("q_proj" in key) or ("lm_head") in key
 
     def test_get_peft_model_state_dict_matches_adapter_name_exactly(self):
-        model = get_peft_model(DummyModel(), LoraConfig(target_modules=["linear"]))
-        model.add_adapter("default_extended", LoraConfig(target_modules=["linear"]))
-
-        for module in model.modules():
-            if hasattr(module, "lora_A"):
-                module.lora_A["default"].weight.data.fill_(1.0)
-                module.lora_B["default"].weight.data.fill_(1.0)
-                module.lora_A["default_extended"].weight.data.fill_(2.0)
-                module.lora_B["default_extended"].weight.data.fill_(2.0)
+        model = get_peft_model(DummyModel(), LoraConfig(target_modules=["linear", "embedding"]))
+        model.add_adapter("default_extended", LoraConfig(target_modules=["linear", "embedding"]))
 
         sd = get_peft_model_state_dict(model, adapter_name="default")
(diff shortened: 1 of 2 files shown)


tests/test_low_level_api.py
```diff
@@ -519,6 +519,22 @@ def test_get_peft_model_state_dict_removes_adapter_name_when_same_as_module_name
             # assert that the module name was not stripped completely from the key
             assert ("v_proj" in key) or ("q_proj" in key) or ("lm_head") in key
 
+    def test_get_peft_model_state_dict_matches_adapter_name_exactly(self):
+        model = get_peft_model(DummyModel(), LoraConfig(target_modules=["linear"]))
+        model.add_adapter("default_extended", LoraConfig(target_modules=["linear"]))
+
+        for module in model.modules():
+            if hasattr(module, "lora_A"):
+                module.lora_A["default"].weight.data.fill_(1.0)
```
BenjaminBossan: Checking for the values is a very roundabout way of going at it. You could simply check the keys in the `state_dict`.

tests/test_low_level_api.py
BenjaminBossan, line 1: In your implementation, you correctly check for `k.endswith(adapter_key)` but this is not covered by a test. Please add a test that covers this, e.g. by targeting an embedding layer.

BenjaminBossan (changes requested):
Please check my comments.

Chessing234:
Thanks for the review! I'll update the test to check the keys in the `state_dict` rather than the values. I'll also add a test targeting an embedding layer to properly cover the `k.endswith(adapter_key)` check. Will push these updates shortly!

Chessing234:
Thanks for the feedback! I have updated the test to directly check the keys in the `state_dict` instead of the values. I also added a test covering the embedding layer to ensure `k.endswith(adapter_key)` works properly.

BenjaminBossan:
@Chessing234 While reviewing your PR, I noticed that this issue not only affects LoRA but almost all PEFT methods. Therefore, we need a more general solution. I worked on this, the resulting PR is #3282. Therefore, this PR is no longer needed. Still, thank you for pointing out the problem.

PR #3269, opened 2026-05-27
Make get_apply_tensor_subclass optional in TorchaoLoraLinear

Chessing234:
### Bug
When injecting LoRA adapters into models whose base weights are quantized outside of Hugging Face (such as using `quantize_(model, Float8WeightOnlyConfig())` from `torchao` directly), `add_adapter` crashes with `TypeError: TorchaoLoraLinear.__init__() missing 1 required positional argument: 'get_apply_tensor_subclass'`.

### Root Cause
`dispatch_torchao` initializes `TorchaoLoraLinear` using `**kwargs`, but `get_apply_tensor_subclass` is missing from `kwargs` when it is not populated from an HF quantizer quantization config.

### Why Fix is Correct
This change makes `get_apply_tensor_subclass=None` optional in `TorchaoLoraLinear.__init__`. During regular forward execution, `get_apply_tensor_subclass` is not used. If the user subsequently attempts to call `merge()` or `unmerge()` (which require re-quantization and therefore require `get_apply_tensor_subclass`), a clean `ValueError` is raised instead of a confusing `TypeError` at model initialization time.

### DIFF
--- a/src/peft/tuners/lora/torchao.py
+++ b/src/peft/tuners/lora/torchao.py
@@ -29,7 +29,7 @@ from .layer import Linear
 class TorchaoLoraLinear(Linear):
     """LoRA layer implementation for Linear layers using torchao data"""
 
-    def __init__(self, *args, get_apply_tensor_subclass, **kwargs):
+    def __init__(self, *args, get_apply_tensor_subclass=None, **kwargs):
         # this is not strictly necessary, as kwargs are stored either way, but we want to error early if
         # get_apply_tensor_subclass is missing.
         if kwargs["config"].lora_bias:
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Duplicate of 3234. Please stop spamming or I will report the account.

Chessing234:
I am very sorry for the spam. As I mentioned in my other PR, I was testing an AI coding agent and unfortunately left it running without properly verifying its outputs, which led to it scraping issues and duplicating PR #3234. I completely understand your frustration and have stopped the agent immediately. This won't happen again. Closing this duplicate PR now.

Issue #3244, opened 2026-05-19
[Proposal] Add FRoD tuner for full-rank parameter-efficient fine-tuning


Bane-Elvin:
### Feature request

I would like to propose adding FRoD as a new tuner to `huggingface/peft`.

FRoD (Full-Rank Efficient Fine-Tuning with Rotational Degrees for Fast Convergence) is a parameter-efficient fine-tuning method that uses a shared full-rank basis together with sparse learnable rotational updates. The goal is to preserve the expressive capacity of full-rank adaptation while keeping the number of trainable parameters small.

I believe FRoD is a good fit for PEFT because:
- it targets standard linear layers in Transformer blocks,
- it follows the adapter-style fine-tuning workflow used by other PEFT tuners,
- it can be integrated through `get_peft_model(...)`,
- and it provides a structured full-rank alternative to low-rank adaptation methods.

Paper:
- FRoD: Full-Rank Efficient Fine-Tuning with Rotational Degrees for Fast Convergence
- AAAI 2026
- DOI: https://doi.org/10.1609/aaai.v40i31.39813

Code:
- https://github.com/Bane-Elvin/AAAI2026-FRoD

If this proposal is accepted, I plan to adapt FRoD to the latest upstream PEFT code structure and submit the implementation as a PR.

### Your contribution

Yes. I can prepare and submit a PR for the core FRoD tuner.

My planned contribution includes:
- adding `PeftType.FROD`,
- adding `src/peft/tuners/frod/` with `config.py`, `layer.py`, `model.py`, and `__init__.py`,
- registering the tuner in the current PEFT registration flow,
- adding save/load support,
- adding tests for integration with PEFT model wrapping,
- and adding documentation for the new method.

I will target the latest `peft` main branch rather than submitting an old fork directly. If preferred, I can keep the first PR focused on the core tuner and submit runnable task examples in a follow-up PR.

BenjaminBossan:
Thanks for your proposal to add FRoD to PEFT. I think it would be a good addition. I skimmed your linked repo and saw that your code is already based on PEFT but probably on a version that's a bit older at this point. Please check out the more recent PEFT code when working on the integration, it should actually make it easier (e.g. less boilerplate in `model.py` and `layer.py`). Also check the new [contribution guideline](https://huggingface.co/docs/peft/v0.19.0/en/developer_guides/contributing#add-a-new-peft-fine-tuning-method).

Bane-Elvin:
[...]

PR #3270, opened 2026-05-27
Feature/add frod

Bane-Elvin:
## Summary

This PR implements the FRoD PEFT tuner proposed in https://github.com/huggingface/peft/issues/3244.

- Add FRoD config, tuner, layer implementation, save/load integration, and PEFT exports.
- Add Conv1D handling and ViT target-module mapping.
- Add FRoD to PEFT test matrices.
- Add package reference docs, runnable Trainer examples, and benchmark settings.

## Examples
- BERT on SST-2 with `google-bert/bert-base-uncased` and `nyu-mll/glue`.
- ViT on Stanford Cars with `google/vit-base-patch16-224` and `tanganke/stanford_cars`.

## Tests
- `PYTHONPATH=src python -m ruff check ...`
- `PYTHONPATH=src python -m py_compile examples/frod_finetuning/*.py`
- `PYTHONPATH=src python -m pytest -o addopts='' tests/test_decoder_models.py tests/test_encoder_decoder_models.py tests/test_feature_extraction_models.py tests/test_seq_classifier.py tests/test_custom_models.py tests/test_frod.py -q -k "FROD or frod"`
- `PYTHONPATH=src python -m pytest -o addopts='' tests/test_config.py -q -k "FROD or frod"`


src/peft/tuners/frod/config.py
```diff
@@ -0,0 +1,164 @@
+# Copyright 2023-present the HuggingFace Inc. team.
```
BenjaminBossan: Let's use 2026 here and throughout.

src/peft/tuners/frod/config.py
```diff
@@ -0,0 +1,164 @@
...
+
+from dataclasses import dataclass, field
+from typing import Optional, Union
+
+from peft.config import PeftConfig
+from peft.utils import PeftType
+
+
+@dataclass
+class FRODConfig(PeftConfig):
```
BenjaminBossan: Let's go with `FrodConfig` etc. It's easier to type and more consistent overall (e.g. we use `LoraConfig`, not `LoRAConfig`).

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+        self.frod_s_indices: Optional[BufferDict] = None
+        self.frod_s_size: Optional[BufferDict] = None
+        self.frod_V: Optional[BufferDict] = None
+        self.frod_U: BufferDict = BufferDict({}, persistent=False)
+
+        self._disable_adapters = False
+        self.merged_adapters = []
+
+        base_layer = self.get_base_layer()
+        if isinstance(base_layer, nn.Linear):
```
BenjaminBossan: Let's use `peft.tuners.tuners_utils._get_in_out_features`.

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+            in_features, out_features = (
+                base_layer.weight.ds_shape if hasattr(base_layer.weight, "ds_shape") else base_layer.weight.shape
+            )
+
+        self.in_features = in_features
+        self.out_features = out_features
+        self.kwargs = kwargs
+
+    @property
+    def merged(self) -> bool:
```
BenjaminBossan: Can be removed.

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+        self,
+        adapter_name,
+        frod_V: BufferDict,
+        frod_s_indices: BufferDict,
+        frod_s_size: BufferDict,
+        frod_dropout,
+        init_weights,
+    ):
+        base_layer = self.get_base_layer()
+        weight = base_layer.weight.T if isinstance(base_layer, Conv1D) else base_layer.weight
```
BenjaminBossan: Let's use `peft.utils.other.transpose`.

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+        return bool(self.merged_adapters)
+
+    def update_layer(
+        self,
+        adapter_name,
+        frod_V: BufferDict,
+        frod_s_indices: BufferDict,
+        frod_s_size: BufferDict,
+        frod_dropout,
+        init_weights,
```
BenjaminBossan: Let's pass the `FrodConfig` here directly instead of passing the individual parameters. Please check how the other PEFT methods implement this. The `Linear` class and `model.py` also need to be updated to reflect this.

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+    ):
+        base_layer = self.get_base_layer()
+        weight = base_layer.weight.T if isinstance(base_layer, Conv1D) else base_layer.weight
+        device = base_layer.weight.device
+        dtype = base_layer.weight.dtype
+
+        param_dtype = dtype
```
BenjaminBossan: Why do we need `dtype` and `param_dtype`?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+
+        self.r[adapter_name] = self.out_features
+        if frod_dropout > 0.0:
+            frod_dropout_layer = nn.Dropout(p=frod_dropout)
+        else:
+            frod_dropout_layer = nn.Identity()
+
+        self.frod_dropout.update(nn.ModuleDict({adapter_name: frod_dropout_layer}))
```
BenjaminBossan:
```suggestion
        self.frod_dropout[adapter_name] = frod_dropout_layer
```

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+
+        if adapter_name not in frod_V:
+            if not frod_V:
```
BenjaminBossan: Better to check against `None`, right?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+                raise ValueError("The FRoD projection buffers are empty. This should not happen.")
+            frod_V[adapter_name] = next(iter(frod_V.values()))
+            frod_s_indices[adapter_name] = next(iter(frod_s_indices.values()))
+            frod_s_size[adapter_name] = next(iter(frod_s_size.values()))
```
BenjaminBossan: This is unclear to me. Could you please add a comment why we want `next(iter(...))` here?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+        self.r = {}
+        self.frod_dropout = nn.ModuleDict({})
+
+        # Sparse S is parameterized by its COO values only.
+        self.frod_lambda_l = nn.ParameterDict({})
+        self.frod_lambda_s_values = nn.ParameterDict({})
+
+        self.frod_s_indices: Optional[BufferDict] = None
+        self.frod_s_size: Optional[BufferDict] = None
+        self.frod_V: Optional[BufferDict] = None
```
BenjaminBossan, line 47: Why not assign an empty `BufferDict` here?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+            frod_V[adapter_name] = next(iter(frod_V.values()))
+            frod_s_indices[adapter_name] = next(iter(frod_s_indices.values()))
+            frod_s_size[adapter_name] = next(iter(frod_s_size.values()))
+
+        nnz = frod_s_indices[adapter_name].shape[1]
+        self.frod_lambda_s_values[adapter_name] = nn.Parameter(torch.zeros(nnz, device=device, dtype=param_dtype))
+
+        self.__dict__["frod_V"] = frod_V
+        self.__dict__["frod_s_indices"] = frod_s_indices
+        self.__dict__["frod_s_size"] = frod_s_size
```
BenjaminBossan: Why write to `__dict__`?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import warnings
+from typing import Optional
+
+import numpy as np
```
BenjaminBossan: Why use numpy and not torch?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+                    continue
+
+                target_dtype = x.dtype
+                V = self.frod_V[active_adapter].to(device=x.device, dtype=target_dtype)
+                U = self.frod_U[active_adapter].to(device=x.device, dtype=target_dtype)
+                indices = self.frod_s_indices[active_adapter].to(device=x.device, dtype=torch.long)
+                size_tensor = self.frod_s_size[active_adapter]
+                if isinstance(size_tensor, torch.Tensor):
+                    size = tuple(int(dim) for dim in size_tensor.tolist())
+                else:
```
BenjaminBossan: What else can it be?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+                lambda_l = self.frod_lambda_l[active_adapter].to(device=x.device, dtype=target_dtype)
+
+                x = x.to(target_dtype)
+                h = self.frod_dropout[active_adapter](x)
+
+                batch_shape = h.shape[:-1]
+                h_flat = h.reshape(-1, h.shape[-1])
+                z_flat = torch.matmul(h_flat, V)
+
+                matmul_dtype = z_flat.dtype
```
BenjaminBossan: Why is this needed? Please add a comment?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,307 @@
...
+                result = result + out_add
+
+        result = result.to(previous_dtype)
+        return result
+
+    def __repr__(self) -> str:
+        rep = super().__repr__()
+        return "frod." + rep
+
+    def _move_adapter_to_device_of_base_layer(self, adapter_name: str, device: Optional[torch.device] = None) -> None:
```
BenjaminBossan, line 345: Why do we need to override the method?

src/peft/tuners/frod/model.py
```diff
@@ -0,0 +1,279 @@
...
+        self._init_frod_projections(config, adapter_name)
+
+    def _check_new_adapter_config(self, config: FRODConfig) -> None:
+        super()._check_new_adapter_config(config)
+
+        for existing_config in self.peft_config.values():
+            if existing_config is config:
+                continue
+            if existing_config.projection_prng_key != config.projection_prng_key:
+                raise ValueError(
```
BenjaminBossan, line 211: Let's add a test for this to `tests/test_initialization.py`.

src/peft/tuners/frod/model.py
```diff
@@ -0,0 +1,279 @@
...
+                    f"FRoD projection initialization key must be the same for all adapters. Got "
+                    f"{config.projection_prng_key=} but previous config had "
+                    f"{existing_config.projection_prng_key}."
+                )
+
+        save_projection_values = sorted({config.save_projection for config in self.peft_config.values()})
+        if len(save_projection_values) > 1:
+            raise ValueError(
```
BenjaminBossan, line 219: Let's add a test for this to `tests/test_initialization.py`.

src/peft/tuners/frod/model.py
```diff
@@ -0,0 +1,279 @@
...
+
+        if not weights:
+            raise ValueError(
+                "No layer types compatible with FRoD were found. Please check `peft_config.target_modules`."
+            )
+
+        if not hasattr(self, "frod_V"):
+            self.frod_V = nn.ModuleDict()
+            self.frod_s_indices = nn.ModuleDict()
+            self.frod_s_size = nn.ModuleDict()
```
BenjaminBossan, line 139: Why not set those during `__init__`?

src/peft/tuners/frod/model.py
```diff
@@ -0,0 +1,279 @@
...
+    _, eigenvectors = np.linalg.eigh(t_pi)
+    return r_matrix.T @ eigenvectors
+
+
+class FRODModel(BaseTuner):
+    prefix: str = "frod_"
+    tuner_layer_cls = FRODLayer
+    target_module_mapping = TRANSFORMERS_MODELS_TO_FROD_TARGET_MODULES_MAPPING
+
+    def _init_frod_projections(self, config: FRODConfig, adapter_name: str) -> None:
```
BenjaminBossan: This runs as a `_pre_injection_hook`. My understanding is this is needed to run `update_layer` with the necessary values. However, the disadvantage is that we basically have to replicate the mechanism of `inject_adapter` here, which is brittle (e.g. updates to `inject_adapter` would need to be sync-ed here). Therefore, I wonder if it's not better to change this into a `_post_injection_hook`. That way, we already know which layers have been touched and don't need to replicate it here.

Of course, when run as `_post_injection_hook`, it is in a sense too late as we need the values to initialize the FRoD layers. So one suggestion would be to initialize the FRoD layers with meta tensors and then override those tensors with the real ones later. That should be minimal overhead and work out to be the same. WDYT?

src/peft/tuners/frod/model.py
```diff
@@ -0,0 +1,279 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import warnings
+from collections import defaultdict
+
+import numpy as np
```
BenjaminBossan: Again, I would prefer to use torch throughout instead of numpy.

src/peft/utils/save_and_load.py
```diff
@@ -796,6 +822,21 @@ def set_peft_model_state_dict(
...
                     " not be accurate on all system configurations."
                 )
+        elif config.peft_type == PeftType.FROD:
+            has_projection = any(
+                k.startswith(("base_model.frod_V.", "base_model.frod_s_indices.", "base_model.frod_s_size."))
+                for k in peft_model_state_dict
+            )
+            if config.save_projection and not has_projection:
+                raise ValueError(
+                    "Specified to load FRoD projection tensors from state dictionary however they were not present!"
```
BenjaminBossan: It would be good to add instructions on how to deal with that situation.

tests/test_frod.py
```diff
@@ -0,0 +1,238 @@
...
+        assert (
+            mlp_same_prng.base_model.model.lin2.frod_V["default"].data_ptr()
+            == mlp_same_prng.base_model.model.lin2.frod_V["other"].data_ptr()
+        )
+        assert (
+            mlp_same_prng.base_model.model.lin2.frod_s_indices["default"].data_ptr()
+            == mlp_same_prng.base_model.model.lin2.frod_s_indices["other"].data_ptr()
+        )
+
+    def test_multiple_adapters_different_prng_raises(self):
```
BenjaminBossan: As mentioned above, let's put this into `test_initializaton.py`.

tests/test_frod.py
```diff
@@ -0,0 +1,238 @@
...
+        torch.manual_seed(0)
+
+        config = FRODConfig(target_modules=["lin1", "lin2"], init_weights=False)
+        peft_model = get_peft_model(mlp, config)
+        config2 = FRODConfig(target_modules=["lin1", "lin2"], init_weights=False)
+        peft_model.add_adapter("other", config2)
+        return peft_model
+
+    @staticmethod
+    def _make_second_adapter_different(peft_model):
```
BenjaminBossan: Shouldn't this be unnecessary with `init_weigths=False`?

tests/test_custom_models.py
```diff
@@ -870,6 +871,14 @@
...
         {"target_modules": ["lin0"], "modules_to_save": ["lin1"], "randlora_alpha": 1},
     ),
+    ########
+    # FRoD #
+    ########
+    ("Vanilla MLP 1 FRoD", "MLP", FRODConfig, {"target_modules": "lin0"}),
+    ("Vanilla MLP 2 FRoD", "MLP", FRODConfig, {"target_modules": ["lin0"]}),
+    ("Vanilla MLP 3 FRoD", "MLP", FRODConfig, {"target_modules": ["lin1"]}),
+    ("Vanilla MLP 4 FRoD", "MLP", FRODConfig, {"target_modules": ["lin0", "lin1"]}),
+    ("Vanilla MLP 5 FRoD", "MLP", FRODConfig, {"target_modules": ["lin0"], "modules_to_save": ["lin1"]}),
```
BenjaminBossan: Could you please also add FRoD to `MULTIPLE_ACTIVE_ADAPTERS_TEST_CASES`?

BenjaminBossan (changes requested):
Thanks for the PR to add FRoD. This looks already quite good. In this review, I focused on the integration, so I haven't checked the docs, examples, and experiments yet.

src/peft/tuners/frod/config.py
Bane-Elvin: Thanks! I will update the copyright year to 2026 throughout the PR.

src/peft/tuners/frod/config.py
Bane-Elvin: Updated to `FrodConfig`, `FrodModel`, and `FrodLayer` to match PEFT naming conventions. I kept `PeftType.FROD` uppercase because it follows the enum naming style.

src/peft/tuners/frod/layer.py
Bane-Elvin: Updated `FrodLayer` to use `peft.tuners.tuners_utils._get_in_out_features()` instead of manually handling `nn.Linear` and `Conv1D`.

src/peft/tuners/frod/layer.py
Bane-Elvin: Removed the duplicate `merged` property.

src/peft/tuners/frod/layer.py
Bane-Elvin: Updated this to use `transpose(base_layer.weight, self.fan_in_fan_out)`.

src/peft/tuners/frod/layer.py
Bane-Elvin: Updated `Linear` and `update_layer()` to receive `FrodConfig` directly.

src/peft/tuners/frod/layer.py
Bane-Elvin: Removed `param_dtype` and use `dtype` directly.

src/peft/tuners/frod/layer.py
Bane-Elvin: Applied the suggested `ModuleDict` assignment.

src/peft/tuners/frod/layer.py
Bane-Elvin: Changed the projection buffer check to explicitly compare against `None`.

src/peft/tuners/frod/layer.py
Bane-Elvin: Added a comment explaining that FRoD projection buffers are shared across adapters for the same module category, and reused them via a named `reference_adapter`.

src/peft/tuners/frod/layer.py
Bane-Elvin: Initialized these projection containers as empty `BufferDict` instances, as the way 'frod_U' definition. Perhaps it is better to understand.

src/peft/tuners/frod/layer.py
Bane-Elvin: Removed the direct `__dict__` assignment and keep the shared projection tensors in non-persistent `BufferDict`s instead.

src/peft/tuners/frod/layer.py
Bane-Elvin: NumPy is only used for the CPU-only FRoD projection initialization/eigendecomposition path; trainable parameters and the forward pass stay in torch.

src/peft/tuners/frod/layer.py
Bane-Elvin: Removed the unused fallback branch since `frod_s_size` is always stored as a tensor buffer.

src/peft/tuners/frod/layer.py
Bane-Elvin: Added a comment to explain that this block computes the sparse FRoD update with `torch.sparse.mm`. CUDA sparse fp16/bf16 kernels are less reliable, so that  use fp32 here and cast the update back below.

src/peft/tuners/frod/layer.py
Bane-Elvin: This follows the convention used by other PEFT tuners and makes wrapped layers show up as `frod.*` in the model repr, which helps identify FRoD-replaced modules when debugging. I will add comments here.

src/peft/tuners/frod/model.py
Bane-Elvin: Added this coverage to `tests/test_initialization.py` using the same structure as `TestVeraInitialization`.

src/peft/tuners/frod/model.py
Bane-Elvin: Added this coverage to `tests/test_initialization.py` using the same structure as `TestVeraInitialization`.

src/peft/tuners/frod/model.py
Bane-Elvin: These containers are initialized lazily because `BaseTuner.__init__()` calls the pre-injection/injection flow before a subclass `__init__()` could safely assign `nn.ModuleDict`s after `super().__init__()`.

src/peft/tuners/frod/model.py
Bane-Elvin: Thank you for the insight. Our implementation follows a similar design rationale as VeRA: FRoD replaces VeRA's randomly initialized shared projection matrices with projection buffers initialized from the base weights, so these tuner-level shared buffers need to be available before each target module calls `update_layer()`. VeRA also prepares its shared A/B buffers in `_pre_injection_hook`. I tried moving this to `_post_injection_hook`, but it required placeholder tensors and later replacement, which introduced extra complexity around `low_cpu_mem_usage`, `load_adapter`, and `requires_grad`.

src/peft/tuners/frod/layer.py
Bane-Elvin: Updated this as well: FRoD now uses `torch.linalg` for this initialization path, and the NumPy dependency was removed.

src/peft/tuners/frod/model.py
Bane-Elvin: Updated the FRoD initialization path to use `torch.linalg` throughout and removed the NumPy dependency.

src/peft/utils/save_and_load.py
Bane-Elvin: Added instructions to the error message: either load with `save_projection=False` to regenerate the projections from the base model weights, or re-save the adapter with `save_projection=True`.

tests/test_frod.py
Bane-Elvin: Added this coverage to `tests/test_initialization.py` using the same structure as `TestVeraInitialization`.

tests/test_frod.py
Bane-Elvin: It's really helpful. Removed the helper and its call; `init_weights=False` already makes the two adapters different for this test.

tests/test_custom_models.py
Bane-Elvin: Added FRoD to `MULTIPLE_ACTIVE_ADAPTERS_TEST_CASES` with the same-target adapter case.

Bane-Elvin:
Updated the PR in 0e0d816 to address the review feedback:

- Renamed the public FRoD classes to follow PEFT naming conventions: `FrodConfig`, `FrodModel`, and `FrodLayer`.
- Replaced custom shape/transpose handling with existing PEFT utilities where applicable.
- Removed redundant code paths and simplified adapter initialization logic.
- Switched FRoD initialization from NumPy to `torch.linalg` and removed the NumPy dependency.
- Added comments for non-trivial implementation choices, including shared projection buffers, sparse CUDA matmul dtype handling, and pre-injection initialization.
- Added/updated initialization tests and moved FRoD-specific initialization checks into `tests/test_initialization.py`.
- Added FRoD coverage to multiple active adapter tests.
- Improved the FRoD projection loading error message with recovery instructions.
- Updated docs and runnable examples to use the new `FrodConfig` naming.

Bane-Elvin:
The two new commits fix the FRoD sparse forward semantics and align the image-classification example with the CLIP-ViT setup used in our experiments.

I also re-tested the examples with two examples: the Stanford Cars image-classification example and the SST-2 text-classification example. The CLIP-ViT Stanford Cars run converged as expected in the local 3-epoch test.

Thank you for the review and suggestions. Looking forward to your feedback.

src/peft/tuners/frod/layer.py
```diff
@@ -279,6 +257,7 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
         return result
 
     def __repr__(self) -> str:
+        # Match PEFT tuner convention so printed models show FRoD-wrapped layers as `frod.*`.
```
BenjaminBossan: This comment can be removed.

src/peft/tuners/frod/layer.py
BenjaminBossan: I'm talking about `_move_adapter_to_device_of_base_layer`, not `__repr__`

src/peft/tuners/frod/layer.py
```diff
@@ -181,12 +181,19 @@ def unmerge(self) -> None:
...
+                delta_weight = self._frod_merged_delta.pop(active_adapter, None)
+                if delta_weight is None:
+                    delta_weight = self.get_delta_weight(active_adapter)
+                self.get_base_layer().weight.data -= delta_weight
 
     def get_delta_weight(self, adapter) -> torch.Tensor:
-        weight = self.get_base_layer().weight
+        return self._get_delta_weight(adapter)
+
+    def _get_delta_weight(self, adapter, base_weight: Optional[torch.Tensor] = None) -> torch.Tensor:
```
BenjaminBossan: I don't see why we need `_get_delta_weight` with the option of passing `base_weight`. AFAICT, it's always `self.get_base_layer().weight`, so there is never a need to pass the argument explicitly.

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,301 @@
...
+
+                target_dtype = x.dtype
+                V = self.frod_V[active_adapter].to(device=x.device, dtype=target_dtype)
+                U = self.frod_U[active_adapter].to(device=x.device, dtype=target_dtype)
+                indices = self.frod_s_indices[active_adapter].to(device=x.device, dtype=torch.long)
+                size = tuple(int(dim) for dim in self.frod_s_size[active_adapter].tolist())
+                values = self.frod_lambda_s_values[active_adapter].to(device=x.device, dtype=target_dtype)
+                lambda_l = self.frod_lambda_l[active_adapter].to(device=x.device, dtype=target_dtype)
+
+                x = x.to(target_dtype)
```
BenjaminBossan: `target_dtype` is defined as `x.dtype`, so this call is useless.

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,301 @@
...
+
+                batch_shape = h.shape[:-1]
+                h_flat = h.reshape(-1, h.shape[-1])
+                z_flat = torch.matmul(h_flat, V)
+
+                # This block computes the sparse FRoD update z @ S.T with torch.sparse.mm, matching
+                # F.linear(h, U @ (S + diag(lambda_l)) @ V.T).
+                # CUDA sparse fp16/bf16 kernels are less reliable, so use fp32 here and cast the update back below.
+                matmul_dtype = z_flat.dtype
+                if z_flat.is_cuda and matmul_dtype in (torch.float16, torch.bfloat16):
```
BenjaminBossan: Just wondering: Why is this limited to `is_cuda`? Wouldn't we expect the same issue on other devices?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,301 @@
...
+                U_mm = U.to(device=z_flat.device, dtype=matmul_dtype)
+                out_add_flat = F.linear(z_S_flat + z_L_flat, U_mm)
+                out_add_flat = out_add_flat.to(target_dtype)
+                out_add = out_add_flat.reshape(*batch_shape, out_add_flat.shape[-1])
+                base_weight = transpose(self.get_base_layer().weight, self.fan_in_fan_out).to(
+                    device=x.device, dtype=target_dtype
+                )
+                base_out = F.linear(x, base_weight)
+
+                result = result - base_out + out_add
```
BenjaminBossan: This part I don't understand: For each FRoD adapter, we need to remove the base result from `result`? So my understanding is that the FRoD result already includes the result from the base model, so we need to subtract it to prevent it being included multiple times.

However, this seems very wasteful: First we calculate `result = self.base_layer(x, *args, **kwargs)`, then we calculate `base_out = F.linear(x, base_weight)`, which is basically the same thing (just without bias) and remove it from the result. And then repeat it for each active adapter.

What I would like to see instead is for the FRoD result to only contain the "delta" from FRoD without the base result, so that we can simply accumulate it. This would be the cleanest implementation IMO. It would also help to simplify `get_delta_weight`, where there is a similar logic to remove the base weight from the FRoD weight.

If, for some reason, it's not possible, we should instead count how many active adapters we have. Then we calculate the base result only _once_ and remove it x times, once for each active adapter. This should be a lot cheaper than the current implementation.

BenjaminBossan (changes requested):
Thanks for the updates. I did another review focusing on the integration and still found a handful of issues, please check.

src/peft/tuners/frod/layer.py
Bane-Elvin: Removed the unnecessary comment.

src/peft/tuners/frod/layer.py
Bane-Elvin: Simplified `get_delta_weight()` to always use the base layer weight. `merge()` now precomputes adapter deltas before mutating the base weight, so the explicit `base_weight` argument is no longer needed.

src/peft/tuners/frod/layer.py
Bane-Elvin: Removed the redundant cast.

src/peft/tuners/frod/layer.py
Bane-Elvin: Changed the fp16/bf16 sparse matmul fallback to use fp32 regardless of device.

src/peft/tuners/frod/layer.py
Bane-Elvin: Updated the forward pass so the FRoD branch explicitly computes and accumulates only the adapter delta. As described in the paper, FRoD directly reconstructs the adapted weight through the joint matrix factorization, so the base-weight contribution must be subtracted; otherwise the base model contribution would be included twice. The code now computes `out_add - adapter_base_out` and reuses the base output when dropout is identity.

src/peft/tuners/frod/layer.py
BenjaminBossan: Isn't `_move_adapter_to_device_of_base_layer` doing the exact same thing as the method from the parent class? Why does it need to be overridden?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,310 @@
...
+    def merge(self, safe_merge: bool = False, adapter_names: Optional[list[str]] = None) -> None:
+        adapter_names = check_adapters_to_merge(self, adapter_names)
+        if not adapter_names:
+            return
+
+        base_layer = self.get_base_layer()
+        adapter_deltas = []
+        for active_adapter in adapter_names:
+            if active_adapter in self.frod_lambda_l.keys():
+                adapter_deltas.append((active_adapter, self.get_delta_weight(active_adapter)))
```
BenjaminBossan, line 183: This could all be done in a single loop, no? I don't see why we need to collect the `adapter_deltas` separately. A single loop should be easier to read and is probably also better for garbage collection.

src/peft/tuners/frod/model.py
```diff
@@ -0,0 +1,273 @@
...
+            raise ValueError(
+                "No layer types compatible with FRoD were found. Please check `peft_config.target_modules`."
+            )
+
+        # BaseTuner.__init__() enters the pre-injection flow before a FrodModel subclass
+        # could assign ModuleDicts after super().__init__(), so create these containers lazily here.
+        if not hasattr(self, "frod_V"):
+            self.frod_V = nn.ModuleDict()
+            self.frod_s_indices = nn.ModuleDict()
+            self.frod_s_size = nn.ModuleDict()
```
BenjaminBossan, line 139: Okay, so for my better understanding, on the `FrodModel`, we have a `frod_s_size` that is a `ModuleDict` and it contains "categories", which are groups of similar modules (?). For each category, we have one `BufferDict`, which contains one value per adapter name. The values themselves are just sizes, so tuple of ints. On the `FrodLayer`, we only have the inner `BufferDict`, not the outer `ModuleDict`.

If this is correct, I would probably rename `frod_s_size` on the `FrodModel` to something like `frod_categories_s_size` or so to avoid confusion. I'm also not too fond of the name `category`, if you have a better idea, let's discuss that.

Finally, since we only have sizes in there, in theory a normal dictionary would be enough, right? We don't really need tensors. But as we may want to persist them, `BufferDict` is still helpful for that, is that understanding correct?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,310 @@
...
+            frod_V[adapter_name] = frod_V[reference_adapter]
+            frod_s_indices[adapter_name] = frod_s_indices[reference_adapter]
+            frod_s_size[adapter_name] = frod_s_size[reference_adapter]
+
+        nnz = frod_s_indices[adapter_name].shape[1]
+        self.frod_lambda_s_values[adapter_name] = nn.Parameter(torch.zeros(nnz, device=device, dtype=dtype))
+
+        self.frod_V[adapter_name] = frod_V[adapter_name]
+        self.frod_s_indices[adapter_name] = frod_s_indices[adapter_name]
+        self.frod_s_size[adapter_name] = frod_s_size[adapter_name]
```
BenjaminBossan: Can this be deleted?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,310 @@
...
+            target_dtype = x.dtype
+            base_weight = transpose(self.get_base_layer().weight, self.fan_in_fan_out).to(
+                device=x.device, dtype=target_dtype
+            )
+            base_out = None
+            for active_adapter in self.active_adapters:
+                if active_adapter not in self.frod_lambda_s_values:
+                    continue
+
+                V = self.frod_V[active_adapter].to(device=x.device, dtype=target_dtype)
```
BenjaminBossan: The code here is partly exactly the same as in `get_delta_weight`. How about factoring out the common parts into a dedicated method?

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,310 @@
...
+                # FRoD reconstructs the adapted weight directly, so subtract the base-weight contribution and only
+                # accumulate the adapter delta.
+                if isinstance(dropout, nn.Identity):
+                    if base_out is None:
+                        base_out = F.linear(x, base_weight)
+                    adapter_base_out = base_out
+                else:
+                    adapter_base_out = F.linear(h, base_weight)
+
+                result = result + out_add - adapter_base_out
```
BenjaminBossan: In theory, if no adapter uses dropout, we could subtract the `adapter_base_out` once outside the loop, multiplied by the number of active adapters. This would be more efficient if we deal with a high amount of active adapters. As this is a very unlikely use case, I'm fine with keeping the code as is for simplicity (it's also numerically more stable), I just wanted to bring that option up.

src/peft/tuners/frod/model.py
```diff
@@ -0,0 +1,273 @@
...
+from peft.tuners.tuners_utils import BaseTuner, BaseTunerLayer
+from peft.utils import TRANSFORMERS_MODELS_TO_FROD_TARGET_MODULES_MAPPING
+
+from .._buffer_dict import BufferDict
+from ..tuners_utils import _maybe_include_all_linear_layers
+from .config import FrodConfig
+from .layer import FrodLayer, Linear
+
+
+def _category_from_key(key: str) -> str:
```
BenjaminBossan, line 34: This probably relies on certain naming conventions. Could you please add a docstring to make those explicit? The docstring should also contain an example, as it's not quite obvious what is happening here.

src/peft/tuners/frod/model.py
```diff
@@ -0,0 +1,273 @@
...
+        return parts[0]
+    if parts[-2].isdigit():
+        return parts[-1]
+    category = f"{parts[-2]}_{parts[-1]}"
+    if (category == "output_dense") and (len(parts) >= 3) and (parts[-3] == "attention"):
+        return "attention_output"
+    return category
+
+
+def _layer_index_from_key(key: str, fallback: int) -> int:
```
BenjaminBossan, line 55: This probably relies on certain naming conventions. Could you please add a docstring to make those explicit? The docstring should also contain an example, as it's not quite obvious what is happening here.

tests/test_frod.py
BenjaminBossan, line 1: I wonder if there should be tests to check for how FRoD determines the "categories". We could use a tiny random model based on a real model architecture and configure FRoD to use common targets (e.g. `q_proj, v_proj`) and then check which categories were determined. WDYT?

BenjaminBossan (changes requested):
Thanks for the new updates. The integration is getting there, but there were still some unclear parts of the code that I commented. Please take a look.

src/peft/tuners/frod/layer.py
Bane-Elvin: It is similar in spirit to VeRA because the projection tensors are shared at the tuner level. However, for FRoD we intentionally keep the shared projection/index buffers on CPU and only move the trainable parameters here; the buffers are moved lazily when they are used.

src/peft/tuners/frod/layer.py
Bane-Elvin: For FRoD this needs to be computed before mutating the base weight. Unlike LoRA-style deltas, `get_delta_weight()` returns the difference between the reconstructed FRoD weight and the current base weight. If we merge in a single loop and update the base weight immediately, subsequent adapters would compute their delta against an already-merged weight. Precomputing the deltas keeps all adapter deltas relative to the original base weight, which is needed when merging multiple adapters.

src/peft/tuners/frod/model.py
Bane-Elvin: Yes, that understanding is correct. The outer `ModuleDict` is keyed by projection category, and each category stores a `BufferDict` keyed by adapter name. For `frod_s_size`, the stored value is only the sparse matrix size tensor.

I agree that `frod_category_s_size` would be more explicit. I originally kept `frod_s_size` to mirror `frod_s_indices`, since both are metadata for the same sparse `S` matrix and are passed together to `FrodLayer`.

A normal dict would be enough for computation, but `BufferDict` keeps the tensors registered and gives them the same configurable `save_projection` / state-dict behavior as the projection and index buffers.

src/peft/tuners/frod/layer.py
Bane-Elvin: I don't think this can be deleted. This follows the same pattern as VeRA: the shared projection buffers are created at the tuner level and then referenced by each wrapped layer. In VeRA, `update_layer()` assigns the shared `vera_A` / `vera_B` buffers to the layer; here FRoD attaches the shared category-level `frod_V`, `frod_s_indices`, and `frod_s_size` entries to the layer-local non-persistent buffers. These are shared references used by `forward()` and `get_delta_weight()`, not independent persistent copies.

src/peft/tuners/frod/layer.py
Bane-Elvin: There is some overlap in fetching/casting the FRoD tensors, but the two paths intentionally do different computations. `get_delta_weight()` materializes the dense delta weight for merge/unmerge, while `forward()` keeps the sparse activation-side computation and avoids constructing the full dense matrix.

I could factor out only the tensor-fetching part, but that helper would mostly return a long tuple of tensors and may make the forward path harder to read. I would prefer to keep the dense merge path and sparse forward path separate unless you think the small helper would still be clearer.

src/peft/tuners/frod/layer.py
Bane-Elvin: Yes, that makes sense. I kept the current form because the common case is a small number of active adapters, and this also handles the dropout and non-dropout cases uniformly. When dropout is `Identity`, `base_out` is already computed only once and reused inside the loop, so the remaining overhead should be small.

src/peft/tuners/frod/model.py
Bane-Elvin: Added a docstring explaining the projection-category naming convention, including BERT-style and CLIP/decoder-style examples.

src/peft/tuners/frod/model.py
Bane-Elvin: Added a docstring explaining the layer-index parsing convention, including `layers.<idx>`, encoder `layer.<idx>`, and fallback examples.

tests/test_frod.py
Bane-Elvin: Sure. We added a test for this using a tiny local `LlamaForCausalLM` config with common targets `q_proj` and `v_proj`. The test checks that FRoD determines the expected categories: `self_attn_q_proj` and `self_attn_v_proj`.

Bane-Elvin:
@BenjaminBossan Thanks for the review! I have addressed the requested changes and re-requested your review. Could you please take another look when you have time?

src/peft/tuners/frod/layer.py
BenjaminBossan: What I mean is that here, we add the three parameters, and then just below, we add them again and cast them. Why not do this in a single step?

src/peft/tuners/frod/layer.py
BenjaminBossan: Ah I see. Please add a comment to explain this.

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,310 @@
...
+    def get_delta_weight(self, adapter) -> torch.Tensor:
+        weight = self.get_base_layer().weight
+        device = weight.device
+        dtype = weight.dtype
+        base_weight = transpose(weight, self.fan_in_fan_out)
+        U = self.frod_U[adapter].to(device=device, dtype=dtype)
+        V = self.frod_V[adapter].to(device=device, dtype=dtype)
+        indices = self.frod_s_indices[adapter].to(device=U.device, dtype=torch.long)
+        size = tuple(int(dim) for dim in self.frod_s_size[adapter].tolist())
+        values = self.frod_lambda_s_values[adapter].to(U.device, U.dtype).clone()
```
BenjaminBossan: I'm wondering if we really need the `.clone` here.

src/peft/tuners/frod/layer.py
BenjaminBossan: Yes, I think it would be better to have a small helper function that takes the device and dtype and returns the relevant tensors, up to the sparse tensor (before making it dense). Not only would it help to keep the two code paths in sync, IMO it also makes it _easier_ to see what is shared between the merging and the forward path.

BenjaminBossan (changes requested):
Thanks for the new updates. For this review, I focused on these changes and the comments that were still open. Generally, my questions have been resolved by I still have a few comments.

Interestingly, one single FRoD test failed in one of the Windows CI runs. I restarted that run, in case it's just a flaky result, but if it fails again, we should investigate.

One more thing: We now added an image generation benchmark to PEFT: https://github.com/huggingface/peft/tree/main/method_comparison/image-gen. The idea is exactly the same as for the MetaMath benchmark. It would be nice if you could add an experiment for FRoD to this benchmark. Check the existing experimental settings for guidance.

src/peft/tuners/frod/layer.py
Bane-Elvin: Factored the shared tensor fetching/casting and sparse tensor construction into `_get_frod_tensors()`, while keeping the dense merge path and sparse forward path separate.

src/peft/tuners/frod/layer.py
Bane-Elvin: Removed the unnecessary `.clone()` since the sparse values are not modified in place.

src/peft/tuners/frod/layer.py
Bane-Elvin: Added a comment explaining that FRoD deltas must be precomputed before mutating the base weight.

Bane-Elvin:
Updated in 740f432a.

This addresses the remaining FrodLayer feedback by simplifying the shared buffer assignment, adding the merge-path comment, removing the unnecessary clone, and factoring the shared FRoD tensor preparation into `_get_frod_tensors()` while keeping the dense merge path and sparse forward path separate.

I checked the previous PR head before pushing this update, and all 10 checks were successful, including the Windows jobs. Could you please let me know if the Windows CI issue still appears on your side after this new run starts?

I also added the FRoD experiment settings for the new image generation benchmark under `method_comparison/image-gen/experiments/frod/flux2-klein-default`.

BenjaminBossan (changes requested):
Thanks for the recent changes, they look good. However, ruff is complaining, so could you please run `make style`?

I have another suggestion: In PR #3117, we recently added the foundation for generic quantization support. What that means is that with only a few changes, it should be possible for most PEFT methods to support multiple quantization backends. In other words, FRoD could support bitsandbytes, torchao, and gptqmodel without the need to add extra FRoD layer classes for each of them.

To achieve that, use the changes in the linked PR as an example and apply the same idea to FRoD. If you run into any difficulties, please let me know. This feature is not a must have, if you don't want to support in FRoD, it's fine. But I would highly recommend to add it, as many users are interested in fine-tuning quantized models.

Bane-Elvin:
Applied `make style` and pushed the formatting fix in c391d359. `make quality` passes locally now.

Bane-Elvin:
Thanks for the suggestion. I agree that generic quantization support is an important direction, and PR #3117 provides a good shared path for methods that naturally operate on quantized base layers.

For FRoD, I think the quantization design needs a separate and more careful treatment. Unlike LoRA-style methods, FRoD does not fundamentally need to keep using the original base weight after initialization. Instead, the target weight is reparameterized through `U`, the shared projection `V`, and sparse/diagonal factors, and the effective adapter weight is reconstructed from these FRoD factors. Therefore, simply supporting a quantized base layer through the generic wrapper may not be the most meaningful compression path for this method. A more relevant direction may be to study whether the FRoD factors themselves, especially `U` and the shared `V`, can be quantized while preserving the intended weight reconstruction behavior.

To avoid making this PR larger, I would prefer to keep this PR focused on adding the core FRoD method. I can investigate the quantization design separately, and if it works well, open a follow-up PR specifically for FRoD quantization support.

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,309 @@
...
+                lambda_l = lambda_l.to(device=z_flat.device, dtype=matmul_dtype)
+                z_L_flat = z_flat_mm * lambda_l
+
+                U_mm = U.to(device=z_flat.device, dtype=matmul_dtype)
+                out_add_flat = F.linear(z_S_flat + z_L_flat, U_mm)
+                out_add_flat = out_add_flat.to(target_dtype)
+                out_add = out_add_flat.reshape(*batch_shape, out_add_flat.shape[-1])
+                # FRoD reconstructs the adapted weight directly, so subtract the base-weight contribution and only
+                # accumulate the adapter delta.
+                if isinstance(dropout, nn.Identity):
```
BenjaminBossan:
```suggestion
                if isinstance(dropout, nn.Identity) or not dropout.training:
```

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,309 @@
...
+            result = self.base_layer(x, *args, **kwargs)
+        elif self.merged:
+            result = self.base_layer(x, *args, **kwargs)
+        else:
+            result = self.base_layer(x, *args, **kwargs)
+            target_dtype = x.dtype
+            base_weight = transpose(self.get_base_layer().weight, self.fan_in_fan_out).to(
+                device=x.device, dtype=target_dtype
+            )
+            base_out = None
```
BenjaminBossan: One more thing I noticed: If we don't use dropout, we can trivially derive the `base_result` from `result`, right? This would save us one computation of `F.linear(x, base_weight)`. So:

```suggestion
            bias = self.get_base_layer().bias
            base_out = result if bias is None else result - bias
```

Then later in the loop, we check if `dropout` is being used, and if it's not, we re-use `base_out`. This way, even for the first adapter, we can avoid computing the heavy Wx twice. WDYT?

examples/frod_finetuning/frod_image_classification.py
```diff
@@ -0,0 +1,139 @@
...
+OUTPUT_DIR = os.environ.get("FROD_IMAGE_OUTPUT_DIR", "clip-vit-base-patch32-frod-stanford-cars")
+DATA_DIR = os.environ.get("FROD_STANFORD_CARS_DATA_DIR")
+NUM_TRAIN_EPOCHS = int(os.environ.get("FROD_IMAGE_NUM_TRAIN_EPOCHS", "3"))
+TRAIN_BATCH_SIZE = int(os.environ.get("FROD_IMAGE_TRAIN_BATCH_SIZE", "64"))
+EVAL_BATCH_SIZE = int(os.environ.get("FROD_IMAGE_EVAL_BATCH_SIZE", "64"))
+SPARSE_RATE = float(os.environ.get("FROD_IMAGE_SPARSE_RATE", "0.01"))
+FROD_LAMBDA_L_LR = float(os.environ.get("FROD_IMAGE_LAMBDA_L_LR", "5e-4"))
+FROD_LAMBDA_S_LR = float(os.environ.get("FROD_IMAGE_LAMBDA_S_LR", "5e-5"))
+CLASSIFIER_LR = float(os.environ.get("FROD_IMAGE_CLASSIFIER_LR", "1e-4"))
+CLIP_TARGET_MODULES = ["q_proj", "k_proj", "v_proj", "out_proj", "fc1", "fc2"]
```
BenjaminBossan: I'm not a fan of setting all of these as environment variables. Let's please parse them as proper command line arguments. You could use `HfArgumentParser(TrainingArguments)` for the `Trainer` arguments and parse the rest manually.

examples/frod_finetuning/frod_text_classification.py
```diff
@@ -0,0 +1,93 @@
...
+from peft import FrodConfig, TaskType, get_peft_model
+
+
+MODEL_NAME = "google-bert/bert-base-uncased"
+DATASET_NAME = "nyu-mll/glue"
+TASK_NAME = "sst2"
+OUTPUT_DIR = "bert-base-uncased-frod-sst2"
+FROD_LAMBDA_L_LR = 2e-2
+FROD_LAMBDA_S_LR = 2e-3
+CLASSIFIER_LR = 1e-2
```
BenjaminBossan: Similar argument, let's add a proper CLI.

src/peft/tuners/frod/model.py
```diff
@@ -0,0 +1,289 @@
...
+        # BaseTuner.__init__() enters the pre-injection flow before a FrodModel subclass
+        # could assign ModuleDicts after super().__init__(), so create these containers lazily here.
+        if not hasattr(self, "frod_V"):
+            self.frod_V = nn.ModuleDict()
+            self.frod_s_indices = nn.ModuleDict()
+            self.frod_s_size = nn.ModuleDict()
+
+        generator = torch.Generator(device="cpu").manual_seed(config.projection_prng_key)
+        categories = {category for layer_dict in weights.values() for category in layer_dict}
+        for category in sorted(categories):
```
BenjaminBossan: Running through this loop can be quite lengthy, without any indication to the user of what's happening. Usually, calling `get_peft_model` is quite fast, so this could come as a surprise. WDYT about adding some progress indication, e.g. using `tqdm` for the loop? You could add a `progressbar: bool` argument to enable/disable it. I would also document that this can be slow.

method_comparison/image-gen/experiments/frod/flux2-klein-default/adapter_config.json
BenjaminBossan, line 1: I tried to run this experiment on my machine with 24 GB and got OOM. The other tested PEFT methods usually stay below 12 GB. Do you think this is expected? For the final run, we'll use 48 GB cards, so the experiment may succeed there, but it could still be an indication that there is room for improvement. One suspicion I have is that float32 is used for the computation, not sure if that explains all of it.

BenjaminBossan (changes requested):
Thanks for the new updates, the tests are passing (one flaky CI run can be ignored).

Regarding the forward path, I think I found one possible optimization, please check my comment. I also checked the examples and added some comments regarding those.

About the quantization: Thanks for explaining why it's not trivially supported here. Let's drop it for now. If you could revisit it later, that would be great.

method_comparison/image-gen/experiments/frod/flux2-klein-default/adapter_config.json
Bane-Elvin: This is a very serious bug, not just an issue with fp32, but also a problem encountered when using sparse matrices. I will fix this bug as soon as possible.

src/peft/tuners/frod/model.py
Bane-Elvin: Thank you very much for your suggestion. We need to perform **Hierarchical Joint Decomposition** on the matrices that require fine-tuning, which is very time-consuming. However, the good news is that the result of this **Hierarchical Joint Decomposition** is measured by the model, so fine-tuning a model for different tasks only requires one decomposition. We will use `tqdm` for the loop and document it.

examples/frod_finetuning/frod_image_classification.py
Bane-Elvin: Replaced the environment-variable based configuration with HfArgumentParser-backed CLI arguments.

examples/frod_finetuning/frod_text_classification.py
Bane-Elvin: Applied the same CLI structure to the text example as well.

Bane-Elvin:
@BenjaminBossan After investigating the unexpectedly high memory usage, we found a few implementation issues in the current FRoD code. This OOM was not expected and was caused by a serious implementation bug, not only by fp32 computation.

First, the shared FRoD projection buffers were not actually shared on GPU. Although `V`, `s_indices`, and `s_size` are initialized once per module category, the previous implementation registered them again inside each wrapped `FrodLayer`. After `model.to(device)`, PyTorch materialized those registered buffers per layer, so the same category-level tensors were duplicated. This was a significant implementation bug because FRoD expects these tensors to be shared across layers in the same category.

Second, we re-tested the sparse forward path in both fp32 and fp16/bf16. In our supported setup, fp16/bf16 was not less reliable, so forcing the sparse path through fp32 was unnecessary and added avoidable memory overhead.

Third, the forward path was still too close to a PEFT delta-style implementation. FRoD semantically reconstructs the adapted weight rather than adding a LoRA-style delta, so the single-adapter/no-dropout active path does not need the target base linear output. Keeping target base weights on GPU in that path added unnecessary memory pressure, although we still need to preserve PEFT behavior for disable/merge/unmerge.

I updated the implementation accordingly. `FrodLayer` now keeps non-registered references to the tuner-level `V`, `s_indices`, and `s_size` buffers, so those tensors remain shared by category instead of being duplicated per layer.

I also updated the sparse forward path to compute the COO activation update directly in the requested dtype when sparse addmm is unavailable, instead of forcing an unconditional fp32 path. For the base-weight memory issue, I added an opt-in `runtime_offload_base_weight` config. Since FRoD reconstructs the adapted weight directly, this option can move target base weights to CPU during the single-adapter/no-dropout active path. It defaults to `False` to preserve PEFT’s usual device invariant and existing test expectations.

Could you please retry `method_comparison/image-gen/experiments/frod/flux2-klein-default/adapter_config.json` with the latest changes? This should be a much more representative memory test after the sharing and sparse-forward fixes.

src/peft/tuners/frod/layer.py
BenjaminBossan: Hmm, I wonder why we need `__dict__` assignment here but not for other PEFT methods like VeRA?

method_comparison/image-gen/experiments/frod/flux2-klein-default/adapter_config.json
```diff
@@ -13,6 +13,7 @@
   "projection_prng_key": 0,
   "regularization_alpha": 0.001,
   "revision": null,
+  "runtime_offload_base_weight": true,
```
BenjaminBossan, line 16: It would make sense to add this option to the MetaMath experiments too.

src/peft/tuners/frod/layer.py
```diff
@@ -0,0 +1,413 @@
...
+            self.merged_adapters.append(active_adapter)
+
+    def unload_and_optionally_merge_module(
+        self, merge: bool, safe_merge: bool, adapter_names: Optional[list[str]]
+    ) -> nn.Module:
+        if merge:
+            self.merge(safe_merge=safe_merge, adapter_names=adapter_names)
+            self._move_base_weight_to_device_of_adapter(self.active_adapters[0] if self.active_adapters else None)
+        else:
+            self._move_base_weight_to_device_of_adapter(self.active_adapters[0] if self.active_adapters else None)
```
BenjaminBossan: Should do the same, right?

```suggestion
        if merge:
            self.merge(safe_merge=safe_merge, adapter_names=adapter_names)
        self._move_base_weight_to_device_of_adapter(self.active_adapters[0] if self.active_adapters else None)
```

Also, please add a comment why `_move_base_weight_to_device_of_adapter` needs to be called.

src/peft/tuners/frod/layer.py
```diff
@@ -215,27 +235,57 @@ def _get_frod_tensors(
...
     def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
         previous_dtype = x.dtype
 
         if self.disable_adapters:
             if self.merged:
                 self.unmerge()
+            self._move_base_weight_to_device(x.device)
             result = self.base_layer(x, *args, **kwargs)
         elif self.merged:
+            self._move_base_weight_to_device(x.device)
```
BenjaminBossan, line 273: Why are the `_move_base_weight_to_device` calls needed?

src/peft/tuners/frod/config.py
```diff
@@ -0,0 +1,191 @@
...
+    progressbar: bool = field(
+        default=True,
+        metadata={
+            "help": (
+                "Whether to show a progress bar while building FRoD projections. Initialization can be slow on large "
+                "models because it runs matrix decompositions over the target module categories."
+            )
+        },
+    )
+    runtime_offload_base_weight: bool = field(
```
BenjaminBossan, line 171: I like the idea of adding this as an option. However, it wouldn't make sense to have multiple adapters on the same base model where some have `runtime_offload_base_weight=True` and others have `False`, right? I think we should enforce the same value for all.

The best way to achieve this is to override `_check_new_adapter_config` on `FrodModel`. This check comes from the parent class, so don't forget to call `super()`. Then, if there are multiple configs with different `runtime_offload_base_weight` values, an error should be raised. Add a test to `tests/test_initialization.py`.

BenjaminBossan (changes requested):
Thanks for implementing the feedback and for investigating the cause of the memory issue.

Unfortunately, when I tried to re-run the image generation benchmark, I still got OOM. Most other PEFT methods manage to stay below 12 GB. Can you replicate this issue? Note that for image generation, we have enabled gradient checkpointing, just mentioning it in case it matters. The MetaMath benchmark also OOMs with 24 GB, but that one skirts much closer to that number, so it doesn't necessarily mean something is wrong there.

Bane-Elvin:
Thanks for re-running this and for pointing out the gradient checkpointing detail. I could reproduce the unexpectedly high memory behavior with the previous implementation, and it was not expected.

I found and fixed a few FRoD-specific memory issues. First, the shared FRoD projection buffers (`V`, `s_indices`, and `s_size`) were initialized once per module category, but were then registered again inside each wrapped `FrodLayer`, so they were duplicated on GPU after `model.to(device)`. Second, the sparse forward path was unnecessarily forcing part of the computation through fp32; after re-testing fp16/bf16 in the supported setup, this was not needed. Third, since FRoD reconstructs the adapted weight directly, the active single-adapter/no-dropout path does not need the target base weight on GPU, so I added `runtime_offload_base_weight` for this case.

I re-ran the image generation benchmark with the existing benchmark script/settings, without disabling gradient checkpointing. The FRoD config I tested was:

- model: `black-forest-labs/FLUX.2-klein-base-4B`
- dataset: `peft-internal-testing/cat-image-dataset`
- target modules: `to_v`, `add_v_proj`, `to_qkv_mlp_proj`, `linear_out`, `to_k`, `to_add_out`, `add_q_proj`, `linear_in`, `add_k_proj`, `to_q`, `to_out.0`
- `sparse_rate=0.01`
- `frod_dropout=0.0`
- `regularization_alpha=0.001`
- `projection_prng_key=0`
- `save_projection=true`
- `runtime_offload_base_weight=true`
- optimizer lr: `1e-3`
- `max_steps=20`, `eval_steps=20`, `valid_size=1`, `test_size=1`

After the fix, I could not reproduce the training OOM. The benchmark-reported training memory was:

- `mem allocated`: `9,128,246,938` bytes, about `8.50 GiB`
- `mem reserved`: `10,524,137,882` bytes, about `9.80 GiB`
- `accelerator memory max`: `10042 MB`

For comparison, I also ran LoRA `r=8`, `lora_alpha=16` with the same 20-step benchmark setup. It reported `accelerator memory max: 9608 MB`, so FRoD is now close to LoRA r=8 and below 12 GB under the benchmark training-memory metric.

One caveat: if I watch `nvidia-smi` during the final test/evaluation generation after `Training finished after 20 steps, evaluation on test set follows`, I can still see the process go to around 17 GB. That appears to be the final eval/generation phase rather than the benchmark's training memory measurement. If the OOM you saw happens specifically during that final generation/eval phase, I can investigate that separately.

tests/test_frod.py
```diff
@@ -170,6 +170,63 @@ def test_frod_projection_buffers_share_memory_with_layers(self, mlp_same_prng):
...
+        expected.float().pow(2).sum().backward()
+
+        assert values.grad is not None
+        assert z.grad is not None
+        assert torch.allclose(actual, expected, atol=1e-3, rtol=1e-3)
+        assert torch.allclose(values.grad, values_expected.grad, atol=1e-3, rtol=1e-3)
+        assert torch.allclose(z.grad, z_expected.grad, atol=1e-3, rtol=1e-3)
+
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is not available")
+    def test_frod_runtime_offload_keeps_base_weight_on_cpu_after_cuda(self, mlp):
```
BenjaminBossan: As this test requires a GPU to run, could you please move it to `tests/test_gpu_examples.py`?

src/peft/tuners/frod/layer.py
```diff
@@ -26,9 +26,47 @@
...
+            if grad_values is not None:
+                grad_values[start:stop] = (grad_rows * z_flat.index_select(1, chunk_cols)).sum(dim=0)
+
+        return grad_z, None, grad_values, None
+
+
 class FrodLayer(BaseTunerLayer):
     adapter_layer_names = ("frod_lambda_l", "frod_lambda_s_values")
     other_param_names = ("frod_V", "frod_U", "frod_s_indices", "frod_s_size", "runtime_offload_base_weight")
+    _frod_sparse_chunk_size = 16_384
```
BenjaminBossan: Would it make sense to make this a config parameter for users to tune? It should mention the tradeoffs when increasing/decreasing that value.

examples/frod_finetuning/frod_text_classification.py
```diff
@@ -0,0 +1,130 @@
...
+            {"params": [p for n, p in model.named_parameters() if "classifier" in n], "lr": frod_args.classifier_lr},
+        ]
+    )
+
+    trainer = Trainer(
+        model=model,
+        args=training_args,
+        train_dataset=tokenized["train"],
+        eval_dataset=tokenized["validation"],
+        tokenizer=tokenizer,
```
BenjaminBossan:
```suggestion
        processing_class=tokenizer,
```

BenjaminBossan (changes requested):
Thank you for investigating further to fix the memory issue. I can confirm that the benchmarks now both run on my machine.

One issue I noticed, however, is that the benchmarks run extremely slowly. Did you observe the same behavior? E.g. for the image benchmark, it takes ~600 sec for 100 steps to train and ~100 sec for the eval step. As a point of comparison, LoRA requires ~70 sec and ~13 sec, respectively.

I checked if it could be related to offloading, but after disabling it, the picture was the same. Then I changed the chunk size, e.g. by increasing it 8x. This might have had a small effect but at the cost of higher memory. If you have time, it would be great if you could investigate once more if there is something obviously wrong in the PEFT integration compared to your original code. If not, it's also okay to leave it as is,  saving memory is more important for PEFT than good runtimes.

Bane-Elvin:
I investigated this and found that the slowdown came from the custom chunked COO activation path, not from base-weight offloading. The chunked implementation reduced temporary memory, but it used Python-level `index_select`/`index_add_` loops and became very slow for the large image-gen projection categories, especially the 9216-dimensional `linear_out` category.

I changed the fp16/bf16 sparse activation path to densify only the sparse `S` factor temporarily and then use dense GEMM. This keeps the trainable COO parameterization and gradients to the sparse values, but avoids the slow chunked Python loop. In my local image benchmark check with bf16, resolution 512, batch size 2, and gradient checkpointing enabled, 20 training steps took 21.2s and peak reserved memory was 10200MB. This should bring the 100-step training time much closer to the LoRA comparison while staying below 12GB.

Bane-Elvin:
I investigated the runtime issue and found that the slowdown came from the custom chunked COO activation path, not from base-weight offloading. I replaced that path with a faster implementation that temporarily densifies only the sparse `S` factor and then uses dense GEMM. This keeps the trainable COO parameterization and gradients to the sparse values, but avoids the Python-level `index_select` / `index_add_` loop.

I pushed the fix in `b102e705`. I also re-ran the local image-generation benchmark with bf16, resolution 512, batch size 2, and gradient checkpointing enabled. For 100 training steps, I got `106.2s` train time and `18.0s` eval time, with peak reserved memory around `10.78GB` and accelerator max memory `10284MB`.

I also re-ran the relevant checks: `tests/test_frod.py` and the FRoD GPU test passed, and both `make style` and `make quality` passed.

BenjaminBossan (approved):
@Bane-Elvin Thank you for your further experiments. I can confirm that FRoD now runs fast and in a memory efficient way. The fact that the code is even simplified is a nice bonus. From my side, there is nothing more to add and we can go ahead and merge.

Just one more thing that we try to test out: Would you like to provide a tweet-sized blurb for FRoD? Something we can put into the release notes and that allows the interested user to quickly understand when they should use FRoD. So basically: What it is, what it's especially good at, when not to use it.

Bane-Elvin:
@BenjaminBossan Thank you very much for the careful review, the additional benchmarking, and all the valuable suggestions. They helped uncover several important implementation issues and made the final FRoD integration both simpler and much more practical.

Tweet-sized blurb: FRoD is a full-rank, replacement-style PEFT method from *FRoD: Full-Rank Efficient Fine-Tuning with Rotational Degrees for Fast Convergence*. Instead of adding low-rank deltas, it reconstructs selected weights with shared rotational subspaces and sparse trainable coefficients. It is especially useful when fast convergence and a higher full-rank capacity ceiling are important, and its large sparse rotational subspace may also be promising for model merging. The main tradeoffs are the costly joint-decomposition initialization and slightly slower forward/backward passes than LoRA due to the sparse structured factors, so it may be less attractive for a one-off single-task fine-tune.

BenjaminBossan:
Thanks a lot @Bane-Elvin for this great work and your patience with the reviews.

PR #3275, opened 2026-05-27
Fix: target_modules as string incorrectly converted to set of characters

Chessing234:
Bug: In `transformers_weight_conversion.py`, when `target_modules` or `target_parameters` is provided as a string (e.g. `"all-linear"`), passing it to `set()` splits the string into individual characters (e.g., `{'a', 'l', '-', ...}`). Root cause: `set(str)` iterates over characters instead of creating a set containing the string. Why fix is correct: Explicitly checking if the input is a string and wrapping it in a set with a single element prevents the string from being erroneously unpacked.


BenjaminBossan:
Again, duplicate of #3232.

PR #3278, opened 2026-05-28
fix(ia3): use division for unmerging ConvNd bias

Chessing234:
Bug: IA3 unmerge for ConvNd bias uses multiplication instead of division.
Root cause: In `_ConvNd.unmerge`, `base_layer.bias.data` was multiplied by `scaling.data`, whereas it should be divided since merging uses multiplication.
Why fix is correct: Using `torch.div` with `scaling.data + 1e-8` correctly reverses the `torch.mul` applied during merge, matching the behavior for weights.


BenjaminBossan:
Duplicate of #3149.

PR #3279, opened 2026-05-28
FIX Remove duplicate VBLoRAConfig entry from peft.__all__

SHAI-Akshay-Tripathi:
## Description
`peft.__all__` listed `"VBLoRAConfig"` twice. Duplicate entries in `__all__` silently confuse documentation tooling, IDE auto-import, and linters that rely on `__all__` to determine the public API surface. It's the only duplicate among the 138 entries.

## Changes
1. `src/peft/__init__.py` — remove the duplicate `"VBLoRAConfig"` entry.
2. `pyproject.toml` — enable ruff's `RUF068` (duplicate `__all__` entries) in `[tool.ruff.lint]` `extend-select`, so the same regression is caught and auto-fixed by the linter going forward. (Per @githubnemo's review suggestion — preferable to a hand-written test since ruff detects and fixes this automatically.)

## How to test


tests/test_public_api.py
githubnemo, line 1: Remove

githubnemo (commented):
Thanks for the PR.

Indeed, this can be a problem. Instead of adding a unit test, let's add 

```
    "RUF068", # Duplicates in __all__
```

to `pyproject.toml` in `tool.ruff.lint.extend-select` since ruff is able to detect and fix this automatically.

tests/test_public_api.py
AkshayTripathi4: Done — removed tests/test_public_api.py and added RUF068 to extend-select in pyproject.toml. Confirmed locally that ruff check flags and auto-fixes the duplicate, and formatting/linting pass clean. Agreed this is much cleaner than the manual test — thanks for the steer.

githubnemo (approved):
LGTM. Thanks!

PR #3280, opened 2026-05-28
fix(tuners): bundle use of _cast_input_dtype across all PEFT methods

Chessing234:
This PR applies the same `_cast_input_dtype` fix across all PEFT tuners to properly respect the `disable_lora_input_dtype_casting` context manager. It replaces raw `x.to()` calls with `self._cast_input_dtype(x, dtype)` for all remaining tuners that were still using it (such as adamss, boft, c3a, fourierft, gralora, hra, ia3, lora, oft, poly, pvera, randlora, road, vblora, vera, and xlora). This bundles the fixes requested in #3177 and #3208 into a single PR.


BenjaminBossan:
@Chessing234 Why did you close all PRs?

PR #3282, opened 2026-05-28
FIX Saving adapter when other adapters partially match name

BenjaminBossan:
The logic to filter for a specific adapter name when saving a model was faulty, as it would match a substring. Therefore, if one adapter is called `"foo"` and another `"foobar"`, saving `"foo"` would also save `"foobar"`. This is now fixed for all PEFT methods.

### DIFF
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -19,7 +19,7 @@ import platform
 import re
 import warnings
 from collections import namedtuple
-from typing import Optional
+from typing import Literal, Optional
 
 import huggingface_hub
 import torch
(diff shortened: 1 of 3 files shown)


src/peft/utils/save_and_load.py
```diff
@@ -300,7 +313,7 @@ def renamed_dora_weights(k):
         ]
     elif config.peft_type in list(PeftType):
         prefix = PEFT_TYPE_TO_PREFIX_MAPPING[config.peft_type]
-        to_return = {k: state_dict[k] for k in state_dict if prefix in k}
+        to_return = {k: v for k, v in state_dict.items() if prefix in k}
```
BenjaminBossan, line 338: Same code as before, just more idiomatic.

src/peft/utils/save_and_load.py
```diff
@@ -74,8 +74,29 @@ def _get_tp_info(model) -> TpInfo | None:
...
+    }
+
+
 def get_peft_model_state_dict(
-    model, state_dict=None, adapter_name="default", unwrap_compiled=False, save_embedding_layers="auto"
+    model,
+    state_dict=None,
+    adapter_name: str = "default",
+    unwrap_compiled: bool = False,
+    save_embedding_layers: bool | Literal["auto"] = "auto",
```
BenjaminBossan, line 99: Note: This is unchanged, except for adding some type annotations.

githubnemo (approved):

PR #3284, opened 2026-05-29
FIX Transformers argument name in rename_source_key

BenjaminBossan:
Transformers PR https://github.com/huggingface/transformers/pull/46067 changed the argument name in `rename_source_key`, breaking PEFT code. This PR switches to using positional arguments only to allow PEFT to work with Transformers from both before and after this change.


BenjaminBossan:
@sayakpaul This should fix the issue mentioned in https://github.com/huggingface/peft/pull/3212#issuecomment-4569843829.

githubnemo (approved):
LGTM

PR #3289, opened 2026-05-29
Add default target_modules for nemotron_h hybrid Mamba-MoE models

A1c0r-Z:
# Add default target_modules for Nemotron-H

## What this does

Registers default `target_modules` for the `nemotron_h` model type so that `LoraConfig()` (and other PEFT methods) can be used on NVIDIA's [Nemotron-3](https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16) hybrid Mamba + MoE + Attention models without users having to specify `target_modules` manually.

Defaults target the attention projections only (`q_proj`, `k_proj`, `v_proj`, `o_proj`). The Mamba mixer's `in_proj` / `out_proj` / `conv1d` are intentionally excluded because they belong to the Mamba block and are forbidden by the Mamba-architecture compatibility check added in #2562. `nemotron_h` is also added to `mamba_model_types` so the compatibility check applies to it.

## Why

Without these defaults, calling `LoraConfig()` on a Nemotron-H model raises `Please specify target_modules`. NVIDIA's own [NeMo Automodel LoRA cookbook for Nemotron-3](https://github.com/NVIDIA-NeMo/Nemotron/tree/main/usage-cookbook/Nemotron-3-Super/lora-text2sql) works around this by passing `exclude_modules=["*.out_proj"]` and bypasses PEFT for the inner LoRA application — extending PEFT to handle nemotron_h natively fixes the friction upstream.

## What changed

`src/peft/utils/constants.py` — adds `"nemotron_h"` to four mappings, following the precedent established in #3136 (gemma4):
- `TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING`
- `TRANSFORMERS_MODELS_TO_ADALORA_TARGET_MODULES_MAPPING`
- `TRANSFORMERS_MODELS_TO_VBLORA_TARGET_MODULES_MAPPING`
- `TRANSFORMERS_MODELS_TO_WAVEFT_TARGET_MODULES_MAPPING`

`src/peft/tuners/tuners_utils.py` — adds `"nemotron_h"` to `mamba_model_types` in `_check_lora_target_modules_mamba` so the Mamba forbidden-module check (`out_proj`, `conv1d`) applies. This protects users who might explicitly try to target those names without realizing they belong to the Mamba mixer.

`tests/test_custom_models.py` — adds two tests under `TestDefaultTargetModules`:
1. `test_default_target_modules_nemotron_h`: verifies the constant lookup for all four mappings and asserts the defaults do not include forbidden Mamba modules.
2. `test_nemotron_h_blocks_mamba_modules`: verifies the Mamba compatibility check raises when a user explicitly targets `out_proj` on a model whose `model_type` is `"nemotron_h"`.

## Verified

Local syntax check passes (`ast.parse` on all three files). Test suite run pending environment install.

## Related

- Precedent: #3136 (gemma4 defaults)
- Mamba block constraint: #2562


tests/test_custom_models.py
```diff
@@ -7008,3 +7008,49 @@ def test_default_target_modules_osf(self):
...
+            TRANSFORMERS_MODELS_TO_ADALORA_TARGET_MODULES_MAPPING,
+            TRANSFORMERS_MODELS_TO_VBLORA_TARGET_MODULES_MAPPING,
+            TRANSFORMERS_MODELS_TO_WAVEFT_TARGET_MODULES_MAPPING,
+        ):
+            assert mapping.get("nemotron_h") == expected
+
+        forbidden = {"out_proj", "conv1d"}
+        assert set(expected).isdisjoint(forbidden)
+
+    def test_nemotron_h_blocks_mamba_modules(self):
```
githubnemo: Let's make this a nemotron-specific test (e.g. `test_nemotron_h_has_defaults_and_blocks_mamba_modules`) that tests the presence of default modules (testing that `target_modules=None` doesn't raise is enough) and that nemotron is considered a mamba model and specifying `out_proj` in `target_modules` *does* raise.

Let's also move this to `test_initialization.py` where the Falcon `out_proj` test resides.

githubnemo (commented):
Good idea, thanks for creating this PR.

I think this is basically good to go, some comments on the tests.

tests/test_custom_models.py
A1c0r-Z: Thanks for the review! Addressed both points: consolidated into a single test_nemotron_h_has_defaults_and_blocks_mamba_modules and moved it next to the Falcon out_proj test in test_initialization.py. It loads the tiny hybrid checkpoint trl-internal-testing/tiny-NemotronHForCausalLM-nano and asserts (a) target_modules=None resolves to the attention defaults without raising, and (b) targeting the Mamba mixer's out_proj raises. PTAL 🙏

githubnemo (approved):
Thanks! LGTM, failing CI is unrelated

PR #3290, opened 2026-05-30
FIX: inject_adapter incorrectly propagates inference_mode to active adapters

kiritozc:
## Description

When injecting a new adapter via `inject_adapter`, the housekeeping section calls:

```python
self.set_adapter(self.active_adapters, inference_mode=peft_config.inference_mode)
```

Here `peft_config` belongs to the **newly injected** adapter, but `self.active_adapters` points to the **existing active** adapter(s). When the new adapter has `inference_mode=True` (e.g. during `save_pretrained` with `path_initial_model_for_weight_conversion` in PiSSA/OLoRA/CorDA workflows), this erroneously freezes the already-active training adapter, causing `grad_norm` to become 0 and training to effectively stop.

## Fix

Only propagate `inference_mode` when the new adapter IS the active adapter (first-time injection). For subsequent adapters, `set_adapter` is called without `inference_mode`, preserving the existing active adapter's trainability state. The new adapter's own `inference_mode` is still correctly handled by the existing code below.

## Tests

- Added `test_inject_adapter_inference_mode_does_not_freeze_active_adapter` — a regression test covering LoraConfig, LoHaConfig, LoKrConfig, IA3Config, OFTConfig, BOFTConfig
- Two existing xfailing tests (`switch_inference_mode` and `add_adapter`) remain as xfail since they require decoupling active adapter selection from `requires_grad` in `set_adapter`

## Related

- Regression introduced in commit `13fa0aea` (PR #2765)

### DIFF
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -6288,19 +6288,13 @@ class TestRequiresGrad:
         params_with_grad = [n for n, p in model.named_parameters() if p.requires_grad]
         assert all(not p.requires_grad for p in model.parameters())
 
-    @pytest.mark.parametrize("config_cls", [LoraConfig, LoHaConfig, LoKrConfig, IA3Config, OFTConfig, BOFTConfig])
-    def test_inject_adapter_inference_mode_does_not_freeze_active_adapter(self, config_cls, tmp_path):
+    def test_inject_adapter_inference_mode_does_not_freeze_active_adapter(self, tmp_path):
         # Regression test for a bug where adding a second adapter with inference_mode=True would incorrectly freeze
         # the already-active training adapter. This happened because inject_adapter propagated the new adapter's
         # inference_mode to set_adapter for the existing active adapters.
-        # See PR #XXXX
+        # See PR #3290
         model = DeepMLP(size=256)
-        extra_kwargs = {}
-        if config_cls == IA3Config:
-            extra_kwargs["feedforward_modules"] = []
-        if config_cls in (BOFTConfig, OFTConfig):
-            extra_kwargs["boft_block_size"] = 4
-        config = config_cls(target_modules=["layers.0.lin0"], **extra_kwargs)
+        config = LoraConfig(target_modules=["layers.0.lin0"])
         model = get_peft_model(model, config)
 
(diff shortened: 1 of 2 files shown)


tests/test_custom_models.py
```diff
@@ -6291,6 +6288,39 @@ def test_loading_model_requires_grad_load_adapter_then_add_adapter(self, config_
         params_with_grad = [n for n, p in model.named_parameters() if p.requires_grad]
         assert all(not p.requires_grad for p in model.parameters())
 
+    @pytest.mark.parametrize("config_cls", [LoraConfig, LoHaConfig, LoKrConfig, IA3Config, OFTConfig, BOFTConfig])
```
BenjaminBossan: OFT is failing to initialize with the given arguments. But I think for this test, just checking LoRA is enough (like in the previous tests).

tests/test_custom_models.py
```diff
@@ -6291,6 +6288,39 @@ def test_loading_model_requires_grad_load_adapter_then_add_adapter(self, config_
...
+    def test_inject_adapter_inference_mode_does_not_freeze_active_adapter(self, config_cls, tmp_path):
+        # Regression test for a bug where adding a second adapter with inference_mode=True would incorrectly freeze
+        # the already-active training adapter. This happened because inject_adapter propagated the new adapter's
+        # inference_mode to set_adapter for the existing active adapters.
+        # See PR #XXXX
```
BenjaminBossan: Update the comment.

BenjaminBossan (changes requested):
Thanks for fixing this edge case. This fix looks good but I have two small comments regarding the test, please check.

tests/test_custom_models.py
kiritozc: Hi, I've addressed both review comments:
  - Simplified the test to only use LoraConfig
  - Updated the PR reference in the comment (#3290)

  Please take another look. Thanks!

BenjaminBossan (commented):
A couple of tests are failing now. I think this is because they baked in the previous assumption and thus may require updating. Could you please take a look?

BenjaminBossan:
@kiritozc Are you still on it?

PR #3254, opened 2026-05-22
DOC Fix broken docstring examples in tuner model.py files

blipbyte:
Fixed four broken `Example` blocks in docstrings:

  - **adalora** — `>>>` smushed mid-line; broke HTML rendering and copy-paste.
  - **boft** — smushed `>>>`; `BOFTConfig` imported from `transformers` (it's a `peft` class); unused `AutoModelForSeq2SeqLM`.
  - **ia3** — `ia3Config` typo from `transformers` (correct: `IA3Config` from `peft`); `IA3Model(config, model)` args reversed (signature is `(model, peft_config, adapter_name)`); same
   typo in the Attributes block.
  - **xlora** — missing `XLoraConfig` and `import torch`; unused `LoraConfig` / `PeftModel`; `prepare_model_for_kbit_training(4)` should be `(model)`.
  
  Also moved adalora and boft from `Example::` to `Example:` + ```py``` fenced style — matches the other 14 tuner files; the literal-block style was being reflowed by `doc-builder
  style` and reintroducing the smushed lines.
  
  Pure docstring changes — no runtime effect. Verified by running each example as a standalone script (t5-base for adalora/ia3, dinov2-small for boft, tiny-gpt2 stub for xlora; xlora's
   real Mistral-7B + 8-bit setup needs GPU).
  
  Note: a separate latent bug surfaced during verification — adalora's example also needs `total_step=<int>` to actually execute (enforced by `AdaLoraConfig.__post_init__`). Out of
  scope for this PR; happy to file a follow-up.
  
  Drafted with Claude assistance; reviewed line-by-line and verified locally.

PR #3291, opened 2026-05-30
DOC Fix incorrect LoRA references in hira and ln_tuning model docstrings

javierdejesusda:
Follow-up to #3254. The `HiraModel` and `LNTuningModel` docstrings still describe their config as "the Lora model", copied from `lora/model.py`. #3254 fixed this for `adalora`, `boft`, `ia3`, and `xlora` but missed these two.

- `hira/model.py`: the docstring cross-referenced `LoraConfig`, but `HiraModel` is configured with `HiraConfig`. Updated both references to `HiraConfig` and the "Lora model" wording to "HiRA model".
- `ln_tuning/model.py`: the `LNTuningConfig` references were already correct, only the "Lora model" wording needed fixing.

### DIFF
--- a/src/peft/tuners/hira/model.py
+++ b/src/peft/tuners/hira/model.py
@@ -54,13 +54,13 @@ class HiraModel(BaseTuner):
 
     Args:
         model ([`torch.nn.Module`]): The model to be adapted.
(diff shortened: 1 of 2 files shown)


githubnemo (approved):
Thanks!

PR #3294, opened 2026-06-01
FIX PVeRA reparameterization crash from invalid randn_like generator argument

AshNicolus:
## What does this PR do?

`PveraLayer._reparametrize` calls `torch.randn_like(std, generator=self.generator)`,
but `torch.randn_like` does not accept a `generator` argument (only `torch.randn`
does). As a result, the call raises `TypeError: randn_like() got an unexpected
keyword argument 'generator'` whenever the sampling branch is taken — i.e. on every
forward pass in training mode, and in eval mode when `sample_at_inference=True`.
PVeRA training is therefore broken.

This was introduced in #3200 (seeded PVeRA RNG) and went unnoticed because the
existing tests only run the model in eval mode with the default
`sample_at_inference=False`, which takes the deterministic `z = mu` branch and never
reaches the offending line.

## Fix

Sample the noise with `torch.randn`, which supports `generator`. Because
`self.generator` is a CPU `torch.Generator` (created in `update_layer`), the noise is
sampled on CPU for reproducible, device-independent results and then moved to `std`'s
device and dtype. When no `generator_seed` is set the behavior is unchanged
(`torch.randn_like(std)`).

## Tests

Added regression tests to `tests/test_pvera.py`:
- `test_pvera_forward_in_training_mode`: forward + backward in train mode no longer raises.
- `test_pvera_sample_at_inference`: eval-mode sampling no longer raises.
- `test_pvera_generator_seed_reproducible`: a fixed `generator_seed` yields reproducible
  output, and a different seed yields a different output.

Before this PR, all three fail (the first two with `TypeError`); after, they pass.

### DIFF
--- a/src/peft/tuners/pvera/layer.py
+++ b/src/peft/tuners/pvera/layer.py
@@ -152,7 +152,13 @@ class PveraLayer(BaseTunerLayer):
     def _reparametrize(self, mu, logvar, sample_at_inference):
         if self.training or (not self.training and sample_at_inference):
             std = torch.exp(0.5 * logvar)
-            eps = torch.randn_like(std, generator=self.generator)
+            if self.generator is None:
+                eps = torch.randn_like(std)
+            else:
+                # torch.randn_like does not accept a `generator` argument, so torch.randn is used instead. The
+                # generator is a CPU torch.Generator (see update_layer), hence the noise is sampled on CPU for
+                # reproducible, device-independent results and then moved to std's device and dtype.
+                eps = torch.randn(std.shape, generator=self.generator).to(device=std.device, dtype=std.dtype)
             z = mu + eps * std
         else:
             z = mu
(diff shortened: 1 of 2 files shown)


BenjaminBossan (changes requested):
`torch.randn_like` does support `generator` as an argument since v2.10.0: https://docs.pytorch.org/docs/2.10/generated/torch.randn_like.html. It's true that it would be nice to support older torch versions too, but I think upgrading to >= 2.10 if you want to use PVeRA is not too much trouble. Also note that if this call really failed, our tests would have caught it (the line is within the test coverage).

In case this is a bot generated PR, I would highly recommend to double check the bot's results for correctness. Since their training data is often out of date, this kind of error can easily slip through.

AshNicolus:
Thanks for the quick review ,you're right. I confirmed locally that
torch.randn_like accepts and correctly uses `generator` (works on 2.12,
and per the docs it's supported since v2.10), so this isn't a real bug
on supported torch, and requiring >= 2.10 for PVeRA's seeding is
reasonable. Closing this sorry for the noise, and thanks for your time.

PR #3296, opened 2026-06-01
Add decoupled weight-decay one-step update test

eSVeeF:
## Summary
This PR adds a deterministic LoRAFA test that validates decoupled weight-decay behavior in a one-step update setting

## Why
Current LoRAFA tests in test_lorafa.py cover initialization and general optimizer-step behavior, but do not explicitly validate non-zero weight-decay update paths

## What Changed
- Added test_lorafa_weight_decay_decoupled_update in test_lorafa.py
- The test creates two seeded identical models and runs the same one-step forward/backward pass on both
- It compares optimizer updates for:
1. weight_decay = 0
2. weight_decay > 0
- It asserts:
1. LoRA A remains unchanged
2. LoRA B is updated and follows the expected decoupled scaling relation
3. Non-LoRA trainable parameters with gradients follow the same decoupled scaling relation

## Relation Checked
param_with_decay ≈ param_without_decay × (1 - lr × weight_decay)

## Notes
- Test-only change, no functional change to LoRAFA behavior
- The test is intentionally one-step and seed-fixed to keep it deterministic
- High learning rate and weight_decay to make the decay effect more pronounced for testing

### DIFF
--- a/tests/test_lorafa.py
+++ b/tests/test_lorafa.py
@@ -150,3 +150,115 @@ def test_LoraFAOptimizer_step():
     for name, param in model.named_parameters():
         if "lora_B" in name:
             assert torch.any(param != 0), f"lora_B weights are still zero for {name}"
+
+
+def test_lorafa_weight_decay_decoupled_update():
+    """
+    Test that one optimizer step applies decoupled weight decay with the expected parameter scaling.
+    """
+    lora_rank = 16
+    lora_alpha = 32
+    # Stronger lr and weight_decay to make the decay effect more pronounced for testing
+    lr = 1e-2
+    weight_decay = 1.0
+    seed = 42
+
+    config = LoraConfig(
+        r=lora_rank,
+        lora_alpha=lora_alpha,
+        target_modules=["lin0", "lin1"],
+        bias="all",
+    )
+    torch.manual_seed(seed)
+    model_no_wd = get_peft_model(SimpleNet(), config).to(torch_device)
+    torch.manual_seed(seed)
+    model_wd = get_peft_model(SimpleNet(), config).to(torch_device)
+
+    # Sanity check: both models start from the same parameters
+    for (name_no_wd, param_no_wd), (name_wd, param_wd) in zip(
+        model_no_wd.named_parameters(), model_wd.named_parameters()
+    ):
+        assert name_no_wd == name_wd
+        assert torch.equal(param_no_wd, param_wd)
+
+    optimizer_no_wd = create_lorafa_optimizer(
+        model=model_no_wd,
+        r=lora_rank,
+        lora_alpha=lora_alpha,
+        lr=lr,
+        weight_decay=0.0,
+    )
+    optimizer_wd = create_lorafa_optimizer(
+        model=model_wd,
+        r=lora_rank,
+        lora_alpha=lora_alpha,
+        lr=lr,
+        weight_decay=weight_decay,
+    )
+    loss = torch.nn.CrossEntropyLoss()
+
+    # Save initial weights of lora_A with and without weight decay
+    initial_lora_A_weights_no_wd = {
+        name: param.clone() for name, param in model_no_wd.named_parameters() if "lora_A" in name
+    }
+    initial_lora_A_weights_wd = {
+        name: param.clone() for name, param in model_wd.named_parameters() if "lora_A" in name
+    }
+
+    # Generate random input and label using different seeds
+    torch.manual_seed(seed + 1)
+    x = torch.randint(100, (2, 4, 10)).to(torch_device)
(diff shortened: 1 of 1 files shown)


tests/test_lorafa.py
```diff
@@ -150,3 +150,115 @@ def test_LoraFAOptimizer_step():
...
+    # Stronger lr and weight_decay to make the decay effect more pronounced for testing
+    lr = 1e-2
+    weight_decay = 1.0
+    seed = 42
+
+    config = LoraConfig(
+        r=lora_rank,
+        lora_alpha=lora_alpha,
+        target_modules=["lin0", "lin1"],
+        bias="all",
```
BenjaminBossan: Let's add a comment here that `bias="all"` is intended to add non-LoRA trainable parameters.

tests/test_lorafa.py
```diff
@@ -150,3 +150,115 @@ def test_LoraFAOptimizer_step():
...
+    )
+    loss = torch.nn.CrossEntropyLoss()
+
+    # Save initial weights of lora_A with and without weight decay
+    initial_lora_A_weights_no_wd = {
+        name: param.clone() for name, param in model_no_wd.named_parameters() if "lora_A" in name
+    }
+    initial_lora_A_weights_wd = {
+        name: param.clone() for name, param in model_wd.named_parameters() if "lora_A" in name
+    }
```
BenjaminBossan: We have proven that those are the same and thus we don't need both, right?

tests/test_lorafa.py
```diff
@@ -150,3 +150,115 @@ def test_LoraFAOptimizer_step():
...
+    label = torch.randint(16, (2, 4, 10)).to(torch_device)
+
+    # Calculate both losses and perform backward passes
+    loss_value_no_wd = loss(output_no_wd, label)
+    loss_value_no_wd.backward()
+    loss_value_wd = loss(output_wd, label)
+    loss_value_wd.backward()
+
+    params_no_wd = dict(model_no_wd.named_parameters())
+    params_wd = dict(model_wd.named_parameters())
```
BenjaminBossan: I wouldn't assign those variables, as they are re-assigned below (l. 237-238)

BenjaminBossan (changes requested):
Thanks for extending the test coverage of LoRA-FA. I have a couple of smaller comments, but overall this looks good. One suggestion: The test seems a bit overloaded at the moment, as it checks for two things: weight decay on LoRA B and non-LoRA parameters. I think it would add clarity of the two concerns were separated.

eSVeeF:
Thanks for the review, that made sense. 

I kept only one initial_lora_A_weights reference and remove the redundant copy. 

Regarding the test split, I think is cleaner now. I divided the test into LoRA B and non-LoRA params. I used a helper for the common block and although I'm not a fan of using asserts inside helpers, I couldn't think of another way without making the tests too abstract. And since the assert just does a sanity check I think for this case is fine.

tests/test_lorafa.py
```diff
@@ -41,6 +41,89 @@ def forward(self, X):
...
 
+def _run_lorafa_weight_decay_step():
+    lora_rank = 16
+    lora_alpha = 32
+    # Stronger lr and weight_decay to make the decay effect more pronounced for testing
+    lr = 1e-2
+    weight_decay = 1.0
+    seed = 42
+
+    config = LoraConfig(
```
BenjaminBossan: Small suggestion: Let's define the `LoraConfig` inside of the individual tests and pass it as a parameter. That way, the first test doesn't have to set `bias="all"`, as it's not needed there. And the second test is more explicit, as I can directly see in the test where these non-LoRA parameters are stemming from.

tests/test_lorafa.py
```diff
@@ -41,6 +41,89 @@ def forward(self, X):
         return X
 
 
+def _run_lorafa_weight_decay_step():
+    lora_rank = 16
+    lora_alpha = 32
+    # Stronger lr and weight_decay to make the decay effect more pronounced for testing
+    lr = 1e-2
+    weight_decay = 1.0
```
BenjaminBossan: Small suggestion: Let's pass the lr and wd as parameters to this function. Then we can calculate the scale inside the test function instead of returning it here, making it more transparent where it comes from.

BenjaminBossan (changes requested):
Thanks for the updates and for splitting the test into two. Generally, I find this better to understand. However, I think with the two suggestions I've added, the tests could be made clearer still. Please check.

eSVeeF:
Done. 

LoraConfig, lr and weight_decay now are passed as parameters to the helper. 

scale is computed on each test, outside helper.

BenjaminBossan (approved):
Thanks for updating the tests and for the PR as a whole. LGTM. Failing CI is unrelated.

PR #3297, opened 2026-06-02
Created tutorial detailing how to do differentially private PEFT fine-tuning of huggingface transformer models using FastDP

lmarchyok:
**Summary**

This PR adds a new tutorial describing how to perform differentially private fine-tuning with PEFT using FastDP.

The tutorial demonstrates:

- Creating a PEFT model using LoRA.
- Initializing and attaching a FastDP PrivacyEngine.
- Running a standard training loop with differential privacy enabled.
- Applying the same workflow to other PEFT methods such as Prompt Tuning and Prefix Tuning.
- Saving the resulting adapter.

**Motivation**

This PR is related to #2310, which discusses expanding documentation coverage and providing additional practical tutorials for PEFT users.

While PEFT includes tutorials for many training workflows and integrations, there does not currently appear to be documentation covering differential privacy training. Differentially private fine-tuning is a common research and production use case for parameter-efficient methods, and PEFT's adapter-based approach is particularly well-suited for privacy-preserving training because only a subset of model parameters are updated.

This tutorial aims to provide a minimal example that users can adapt to their own training pipelines.

**Changes**

Added a new tutorial covering:

- Loading a pretrained model.
- Creating a PEFT configuration.
- Attaching a FastDP privacy engine to the optimizer.
- Training a PEFT model with differential privacy.
- Saving and reusing the resulting adapter.

The tutorial follows the structure and style of existing PEFT integration and tutorial documentation.

### DIFF
--- a/docs/source/tutorial/peft_dp.md
+++ b/docs/source/tutorial/peft_dp.md
@@ -1,0 +1,161 @@
+<!--Copyright 2023 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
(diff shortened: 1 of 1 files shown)


githubnemo (commented):
Hey, thanks for the PR!

I think it is a good idea to document how to use PEFT in the ecosystem. However, in its current state I think that this tutorial could also be summarized as "install fastDP and follow the basic instructions". If that is the case, I think we shouldn't write a tutorial - after all this will still need to be kept up-to-date. Better to add it in `Integrations`, if at all.

I'd be more interested in the caveats that come with PEFT + fastDP (if there are any). I'm not too familiar with DP but are there aspects unique to PEFT that users need to know about? Learning rates will probably be higher, is there an increased risk of memorization that counters effects of fastDP?

PR #3300, opened 2026-06-03
Documentation re-structure

githubnemo:
The current state of the PEFT docs is not one of structure and I was constantly annoyed that whenever I wanted to change something there were several places that needed touching and they all felt disconnected. So this is my attempt at structuring the docs. Some of these ideas are quite old (discussed in 01/2025) but are still valid.

I've removed most of the code guides without replacement. That's not ideal, I think we should have code examples but I'm think they should be method-focused. Maybe one general example of a training workflow is sufficient because most methods follow the same scheme. I'd appreciate some feedback on this.

All details from the method guides (prompting, lora, oft/boft, etc.) are now integrated into the respective method pages instead. I would have hesitated to do this if these guides would have integrated information about the adapters but they didn't. I think it makes a lot more sense to have one place for each method to gather examples/tips/recommendations and that is now the `package_refernce/<method>` page. This page now also hosts a small space that shows the MetaMathQA (and potentially other) benchmark results highlighted for that method.

I've moved the LoRA initializations to `package_reference/lora#Initialization` and converted the init methods to `<hfoption>`-tags. This collapses them to a list but may reduce searchability through the document - at least firefox is not able to search 'through' the option tabs. This also doesn't make them appear in the ToC and people specifically searching for, say, PiSSA won't find it directly. I think that's OK though, since the search is able to locate it.

The quicktour is a bit more detailed about what happens under the hood (quick doesn't have to mean simplistic) and includes some new visualizations. I hope that we can integrate more visualizations in the future where it makes sense.


BenjaminBossan:
Thanks a lot for revamping the PEFT docs, which I agree are not very user friendly at the moment. Could you please resolve the two merge conflicts so that preview docs could be rendered? I think it makes more sense to review the docs as a whole than going through the diff (which is probably showing a lot of text that has just moved places).

One concern that I have is that links to the PEFT docs could break with the new structure. Thus I have two questions:

1. Did you update doc links we may have in PEFT to ensure that they'll be up to date?
2. How do we deal with external links? It could be e.g. other repos (say, Axolotl, Hermes skill, etc.) but could also concern HF repos (e.g. links from PEFT or Transformers issues).

githubnemo:
> One concern that I have is that links to the PEFT docs could break with the new structure. Thus I have two questions:
> 
>     1. Did you update doc links we may have in PEFT to ensure that they'll be up to date?

I didn't at the time but now I have. There were 14 occurrences of now broken links, all of which are fixed now.

>     2. How do we deal with external links? It could be e.g. other repos (say, Axolotl, Hermes skill, etc.) but could also concern HF repos (e.g. links from PEFT or Transformers issues).

I've added a `_redirects.yml` with the most common redirects I found (mostly from transformers). I also checked axolotl, diffusers and unsloth - the latter was not easy to analyze systematically as I couldn't find the docs as plain text, so I resorted to delegating to an agent which didn't find references to the PEFT docs.

The Hermes PEFT skill (https://github.com/NousResearch/hermes-agent/tree/main/optional-skills/mlops/peft) doesn't seem to link to changed pages in the docs.

githubnemo:
This also requires merging of https://huggingface.co/datasets/huggingface/documentation-images/discussions/625.

BenjaminBossan:
> This also requires merging of https://huggingface.co/datasets/huggingface/documentation-images/discussions/625.

Done

githubnemo:
I think this is now ready for review. Sorry about the huge PR but dissolving the guides into the individual method pages made a relatively big splash in terms of changes, even though the individual changes are quite small.

@stevhliu it would be super cool if you could take a look as well :)

When reviewing the rendered doc on moon-ci-docs I noticed that the new images are rendered with borders (esp. visible in the quicktour) and the ToC indentation for LoRA variants is broken but I have no clue how to fix this. @stevhliu do you have an idea?

docs/source/index.md
```diff
@@ -16,34 +16,27 @@ rendered properly in your Markdown viewer.
...
 
 PEFT is integrated with the Transformers, Diffusers, and Accelerate libraries to provide a faster and easier way to load, train, and use large models for inference.
 
+<div class="flex justify-center">
+  <div class="flex flex-col basis-1/4">
+    There are numerous methods to "adapt" existing models, often extensively integrating into the model. PEFT can be thought of as a framework for arbitrary methods of model adaption (modifying weights, wrapping layers, manipulating KV-caches, ...) while also serving as a reference implementation for many fine-tuning methods.
+  </div>
+  <div class="flex flex-col basis-3/4 pl-10 pr-10"><img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/peft/adapter_installation.png" width="100%"></div>
```
BenjaminBossan: This looks a bit odd (middle part) with dark theme:

Image

docs/source/quicktour.md
```diff
@@ -152,11 +191,35 @@ from peft import AutoPeftModel
 model = AutoPeftModel.from_pretrained("smangrul/openai-whisper-large-v2-LORA-colab")
 ```
 
+## Multiple adapters
+
+PEFT supports installing multiple adapters (of the same kind, in this document this would be LoRA) on top of a base model. When you call `get_peft_model` there is only one adapter named `"default"` but you can add as many additional adapters by calling `peft_model.add_adapter(adapter_name=...)`.
```
BenjaminBossan:
```suggestion
PEFT supports installing multiple adapters (of the same kind, in this document this would be LoRA) on top of a base model. When you call `get_peft_model` there is only one adapter named `"default"` but you can add as many additional adapters as you want by calling `peft_model.add_adapter(adapter_name=...)`.
```

docs/source/quicktour.md
```diff
@@ -152,11 +191,35 @@ from peft import AutoPeftModel
 model = AutoPeftModel.from_pretrained("smangrul/openai-whisper-large-v2-LORA-colab")
 ```
 
+## Multiple adapters
```
BenjaminBossan, line 199: In the section above, the docs describe the `AutoPeftModel` API for loading trained adapters. I'm just wondering if we should not at the very least mention the `PeftModel.from_pretrained(base_model, adapter_id)` API as well.

docs/source/developer_guides/memory_efficient_training.md
```diff
@@ -0,0 +1,48 @@
...
+
+-->
+
+# Memory Efficient Training
+
+🤗 PEFT provides you with methods for parameter efficient fine-tuning but that doesn't mean that your training process is memory efficient. This guide is a collection of tips that you can use to improve memory efficiency of your training process. This guide is mostly an overview page that will link you to the respective other guides and offer some tips for specific situations.
+
+## Choosing the right method
+
+Not every PEFT method is built equally and some formulations are easier to build in a memory efficient manner. If you are on a memory budget it makes sense to check out the [PEFT method comparison suite](https://huggingface.co/spaces/peft-internal-testing/PEFT-method-comparison) and filter for **maximum** accelerator memory usage. Average accelerator memory usage can be fairly equal across methods but not every method scales equally with activations and sequence length and is more prone to memory spikes than others.
```
BenjaminBossan: I think as is, the last sentence doesn't quite make sense, even though it's clear what is meant. Here is a suggestion for a different wording.

```suggestion
Not every PEFT method is built equally and some formulations are easier to build in a memory efficient manner. If you are on a memory budget it makes sense to check out the [PEFT method comparison suite](https://huggingface.co/spaces/peft-internal-testing/PEFT-method-comparison) and filter for **maximum** accelerator memory usage. Average accelerator memory usage can be fairly equal across methods but not every method scales equally with activations and sequence length; some methods are more prone to memory spikes than others.
```

docs/source/developer_guides/memory_efficient_training.md
```diff
@@ -0,0 +1,48 @@
...
+
+Especially when targeting large layers like language modeling heads or embedding layers to fine-tune specific tokens it might make sense to look into [using trainable tokens](troubleshooting#using-trainable-tokens).
+
+## Chunked NLL loss
```
BenjaminBossan: I'd put this section last, I think the other ones below are more generally applicable.

docs/source/developer_guides/memory_efficient_training.md
```diff
@@ -0,0 +1,48 @@
...
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# Memory Efficient Training
+
+🤗 PEFT provides you with methods for parameter efficient fine-tuning but that doesn't mean that your training process is memory efficient. This guide is a collection of tips that you can use to improve memory efficiency of your training process. This guide is mostly an overview page that will link you to the respective other guides and offer some tips for specific situations.
```
BenjaminBossan: Maybe worth mentioning in one section that the normal rules for saving memory still apply: Smaller base model, smaller batch size, shorter sequence length etc.

I also wonder if it would make sense to give a rough guide to the memory that is required to train, say, an LLM: Memory for base model (very dtype dependent), memory for extra parameters (if any) of the adapter x 4 (gradients + optimizer states for Adam), memory for activations (hard to predict but depends mostly on compute dtype, sequence length). I think if we have laid the ground like that, the sections below can also be more easily understood (e.g. what to expect from GC).

docs/source/developer_guides/memory_efficient_training.md
```diff
@@ -0,0 +1,48 @@
...
+
+![NLL vs. Chunked NLL comparison](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/peft/chunked_nll.png)
+
+In case the default chunk size is not optimal for your setting, look in the [original TRL PR](https://github.com/huggingface/trl/pull/5575) for more information on how to tune the chunk size.
+
+## Quantization
+
+Quantization is one of the best ways to reduce memory consumption *of the base model* and will, depending on the employed quantization, also reduce activation memory. Since the PEFT methods will only take up a small portion of the total number of parameters, PEFT defaults to use a higher precision than the base model. This can also have the effect that adapters can mitigate some of the quality loss incured by quantization methods. Read the [PEFT quantization guide](quantization).
```
BenjaminBossan:
```suggestion
Quantization is one of the best ways to reduce memory consumption *of the base model* and will, depending on the employed quantization, also reduce activation memory. Since the PEFT methods will only take up a small portion of the total number of parameters, PEFT defaults to use a higher precision than the base model. This can also have the effect that adapters can mitigate some of the quality loss incurred by quantization methods. Read the [PEFT quantization guide](quantization).
```

docs/source/developer_guides/memory_efficient_training.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Compilation
+
+The models we train are composed of operations like matrix multiplications, sums and assignments where each operation produces a new result and, subsequently, needs to take up memory. If those intermediate results are not needed we can fuse these operations and save up on memory. This is just one of many optimizations that `torch.compile` can do for you, so check out the [PEFT torch.compile guide](torch_compile).
+
+## Gradient Checkpointing
+
+You can trade memory with computation by only saving every nth gradient between layers and computing the rest on the fly. Check out the [gradient checkpointing](https://huggingface.co/docs/transformers/grad_checkpointing) documentation of Transformers to learn more.
```
BenjaminBossan, line 50: Maybe worth mentioning that if not using Transformers or Diffusers, users may have to implement their own GC logic.

docs/source/methods/overview.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+# Parameter efficient fine-tuning methods
+
+Training a model parameter efficiently means to train as few parameters as possible to achieve comparable performance to training all parameters, i.e. full fine-tuning. There is, of course, no free lunch: by using fewer and therefore less expressive, parameters, it is not guaranteed that you will get the same performance! You may need to use a specific PEFT method to get optimal results for the model/task combination you want to train. But you will need less memory and possibly less compute during training and may gain features such as fast hot-swapping between trained expert models and less forgetting of previous knowledge compared to full fine-tuning.
+
+Giving general advice for training large models is hard but for generative
+models, especially language models, you can follow these steps:
+
+1. use prompting (few-shot examples in the prompt) to see if the model is
```
BenjaminBossan:
```suggestion
1. use prompting (e.g. few-shot examples in the prompt) to see if the model is
```

docs/source/methods/overview.md
```diff
@@ -0,0 +1,68 @@
...
+   now use [Prompt-based methods](#prompt-based-methods) to learn the prompt and
+   save precious tokens.
+2. If prompt-based methods are not sufficient you can use [layer tuning](#layer-tuning)
+   and [adapter methods](#adapter-methods). These methods are generally
+   more expressive than prompt-based methods and get closer to full-finetuning.
+3. Make sure to measure retention of already learnt knowledge since each
+   fine-tuning step is potentially unlearning past knowledege.
+
+The [PEFT method comparison suite](https://huggingface.co/spaces/peft-internal-testing/PEFT-method-comparison) aims to give a rough overview of (most) implemented methods on selected benchmarks and models.
+
```
BenjaminBossan, line 29: It could also be useful to mention some criteria here that may guide you in choosing the appropriate PEFT method:

- quantization: not all methods support quantized base models
- feature set: not all features are supported for all methods (e.g. multiple adapters, mixed adapter inference)
- layer types: linear layers are generally always supported, but not all methods support embedding (important for expanding vocab) or conv (important for some image models)
- inference runtime: PEFT methods generally add runtime overhead but some of that can be mitigated (e.g. some methods allow merging, removing the overhead)

docs/source/methods/overview.md
```diff
@@ -0,0 +1,68 @@
...
+The PEFT library supports several types of prompting methods (p-tuning, prefix tuning, prompt tuning, ...), explore the table of contents for a full listing of soft prompt methods.
+If you're interested in applying these methods to other tasks and use cases, take a look at our [notebook collection](https://huggingface.co/spaces/PEFT/soft-prompting)!
+
+> [!TIP]
+> Some familiarity with the general process of training a causal language model would be really helpful and allow you to focus on the soft prompting methods. If you're new, we recommend taking a look at the [Causal language modeling](https://huggingface.co/docs/transformers/tasks/language_modeling) guide first from the Transformers documentation. When you're ready, come back and see how easy it is to drop PEFT in to your training!
+
+## Layer Tuning
+
+Layer Tuning categorizes methods that target specific layers of a model such as [LayerNorm Tuning](../package_reference/layernorm_tuning)
```
BenjaminBossan: "target specific layers" doesn't make it quite clear that it means that existing parameters of the base model are made trainable, since you could say that LoRA also targets specific layers. I would state that explicitly.

BenjaminBossan (changes requested):
Thanks a LOT for working on overhauling the PEFT docs. They always felt lacking and suboptimally structured to me, so I'm very happy to see improvements there.

For this review, I focused on the general sections but haven't reviewed the entries for the individual PEFT methods. This was in order to break down the review in smaller parts, as I'm not going to finish it today. It may also help avoid duplicate effort between me and Steven.

As a more general comment, I saw that some added parts contain manual line breaks, e.g. in `overview.md`. I would suggest to remove those completely.

I like the idea of including a benchmark overview for each PEFT method. Now that we have image generation too, it would be great to add an option to toggle the benchmark, but let's leave that to a future PR. I noticed, however, that not each PEFT method includes the benchmark, e.g. HRA is missing it. Also, some methods like HiRA have the graph but no corresponding data points, but maybe its result was added after the space was deployed?

I also wonder if we should not fully remove the legend, as the resulting graph can become quite cramped:

Image

There is also a bit of an inconsistency about the legend, e.g. for Lily it only labels the line but not the points. I think it should be removed for simplicity.

docs/source/package_reference/lora.md
```diff
@@ -16,71 +16,1070 @@ rendered properly in your Markdown viewer.
 
 # LoRA
 
-Low-Rank Adaptation ([LoRA](https://huggingface.co/papers/2309.15223)) is a PEFT method that decomposes a large matrix into two smaller low-rank matrices in the attention layers. This drastically reduces the number of parameters that need to be fine-tuned.
+> [!TIP]
+> LoRA is one of the most popular PEFT methods and a good starting point if you're just getting started with PEFT. It was originally developed for large language models but it is a tremendously popular training method for diffusion models because of its efficiency and effectiveness.
+
+Low-Rank Adaptation ([LoRA](https://huggingface.co/papers/2106.09685)) is a PEFT method that decomposes a large matrix into two smaller low-rank matrices. This drastically reduces the number of parameters that need to be fine-tuned.
 
 The abstract from the paper is:
```
stevhliu, line 24: i think this is the wrong abstract

docs/source/package_reference/lora.md
```diff
@@ -16,71 +16,1070 @@ rendered properly in your Markdown viewer.
...
+* LoRA is orthogonal to other parameter-efficient methods and can be combined with many of them.
+* Performance of models finetuned using LoRA is comparable to the performance of fully finetuned models.
 
+In principle, LoRA can be applied to any subset of weight matrices in a neural network to reduce the number of trainable parameters. However, for simplicity and further parameter efficiency, LoRA is typically only applied to the attention blocks in Transformer models - it may be worth targeting other layers as well. The resulting number of trainable parameters in a LoRA model depends on the size of the update matrices, which is determined mainly by the rank `r` and the shape of the original weight matrix.
 
-## Utility
+You can initialize the low-rank matrices with different use-cases in mind - task awareness (CoRDA, EVA), faster convergence (PiSSA), mitigating quantizations (LoftQ) - just to name a few use-cases. Read about the different initializations [below](#Initialization). The default initialization is for LoRA to be a no-op, to gradually learn new behavior without interfering much with the existing model.
```
stevhliu:
```suggestion
You can initialize the low-rank matrices with different use-cases in mind - task awareness (CoRDA, EVA), faster convergence (PiSSA), mitigating quantizations (LoftQ) - just to name a few use-cases. Read about the different initializations [below](#initialization). The default initialization is for LoRA to be a no-op, to gradually learn new behavior without interfering much with the existing model.
```

docs/source/methods/overview.md
stevhliu: may be nice to reorder the categories to be more consistent

- the intro workflow lists prompting, layer tuning, adapters
- the page sections are ordered adapters, prompting, layer tuning
- the toctree orders them as layer tuning, soft prompting, adapters. it would also be good to pick and use the same terms in the sidebar and here (prompt-based methods vs soft prompting)

docs/source/methods/overview.md
stevhliu: this section feels a bit thin. you could maybe add something about what distinguishes it more from prompting that makes it more expressive. otherwise, it'd be harder to pick between the two of them

docs/source/methods/overview.md
```diff
@@ -0,0 +1,68 @@
...
+
+1. use prompting (few-shot examples in the prompt) to see if the model is
+   already capable of the task. If the model solves your problem, great! You can
+   now use [Prompt-based methods](#prompt-based-methods) to learn the prompt and
+   save precious tokens.
+2. If prompt-based methods are not sufficient you can use [layer tuning](#layer-tuning)
+   and [adapter methods](#adapter-methods). These methods are generally
+   more expressive than prompt-based methods and get closer to full-finetuning.
+3. Make sure to measure retention of already learnt knowledge since each
+   fine-tuning step is potentially unlearning past knowledege.
```
stevhliu:
```suggestion
   fine-tuning step is potentially unlearning past knowledge.
```

docs/source/methods/overview.md
```diff
@@ -0,0 +1,68 @@
...
+
+> [!TIP]
+> LoRA is one of the most popular PEFT methods and a good starting point if you're just getting started with PEFT. It was originally developed for large language models but it is a tremendously popular training method for diffusion models because of its efficiency and effectiveness.
+
+Low-rank adapters are only one possible adapter formualation, PEFT implements many other types of adapters as well. For example, Orthogonal Fine-Tuning methods ([OFT](../package_reference/oft), [BOFT](../package_reference/boft), ...) use orthogonal decompositions of the adapter weights to achieve small size. Methods like [MiSS](../package_reference/miss) shard matrices and share these shards to save on memory. [IA3](../package_reference/ia3) just introduces three trainable vectors to steer the original model.
```
stevhliu:
```suggestion
Low-rank adapters are only one possible adapter formulation, PEFT implements many other types of adapters as well. For example, Orthogonal Fine-Tuning methods ([OFT](../package_reference/oft), [BOFT](../package_reference/boft), ...) use orthogonal decompositions of the adapter weights to achieve small size. Methods like [MiSS](../package_reference/miss) shard matrices and share these shards to save on memory. [IA3](../package_reference/ia3) introduces learned vectors that rescale the key, value, and feed-forward activations.
```

docs/source/methods/overview.md
```diff
@@ -0,0 +1,68 @@
...
+- soft prompts are learnable tensors concatenated with the input embeddings that can be optimized to a dataset; the downside is that they aren't human readable because you aren't matching these "virtual tokens" to the embeddings of a real word
+
+The PEFT library supports several types of prompting methods (p-tuning, prefix tuning, prompt tuning, ...), explore the table of contents for a full listing of soft prompt methods.
+If you're interested in applying these methods to other tasks and use cases, take a look at our [notebook collection](https://huggingface.co/spaces/PEFT/soft-prompting)!
+
+> [!TIP]
+> Some familiarity with the general process of training a causal language model would be really helpful and allow you to focus on the soft prompting methods. If you're new, we recommend taking a look at the [Causal language modeling](https://huggingface.co/docs/transformers/tasks/language_modeling) guide first from the Transformers documentation. When you're ready, come back and see how easy it is to drop PEFT in to your training!
```
stevhliu:
```suggestion
> Some familiarity with the general process of training a causal language model would be really helpful and allow you to focus on the soft prompting methods. If you're new, we recommend taking a look at the [Causal language modeling](https://huggingface.co/docs/transformers/tasks/language_modeling) guide first from the Transformers documentation. When you're ready, come back and see how easy it is to drop PEFT into your training!
```

docs/source/methods/overview.md
```diff
@@ -0,0 +1,68 @@
...
+
+A popular way to realize adapters is to insert smaller trainable matrices that are a low-rank decomposition of the adapted weight's layout to save on memory. There are several different ways to express the weight matrix as a low-rank decomposition, but [Low-Rank Adaptation (LoRA)](../package_reference/lora) is the most common method. The PEFT library supports several other variations of this formulation - some are direct variants of LoRA and are documented under LoRA, some are different enough to count as their own methods, such as [Low-Rank Hadamard Product (LoHa)](../package_reference/loha), [Low-Rank Kronecker Product (LoKr)](../package_reference/lokr), and [Adaptive Low-Rank Adaptation (AdaLoRA)](../package_reference/adalora).  If you're interested in applying these methods to other tasks and use cases like semantic segmentation, token classification, take a look at our [notebook collection](https://huggingface.co/collections/PEFT/notebooks-6573b28b33e5a4bf5b157fc1)!
```
stevhliu:
```suggestion
A popular way to realize adapters is to insert smaller trainable matrices that are a low-rank decomposition of the adapted weight's layout to save on memory. There are several different ways to express the weight matrix as a low-rank decomposition, but [Low-Rank Adaptation (LoRA)](../package_reference/lora) is the most common method. The PEFT library supports several other variations of this formulation - some are direct variants of LoRA and are documented under LoRA, some are different enough to count as their own methods, such as [Low-Rank Hadamard Product (LoHa)](../package_reference/loha), [Low-Rank Kronecker Product (LoKr)](../package_reference/lokr), and [Adaptive Low-Rank Adaptation (AdaLoRA)](../package_reference/adalora). If you're interested in applying these methods to other tasks and use cases like semantic segmentation, token classification, take a look at our [notebook collection](https://huggingface.co/collections/PEFT/notebooks-6573b28b33e5a4bf5b157fc1)!
```

docs/source/methods/overview.md
```diff
@@ -0,0 +1,68 @@
...
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+
+# Parameter efficient fine-tuning methods
+
+Training a model parameter efficiently means to train as few parameters as possible to achieve comparable performance to training all parameters, i.e. full fine-tuning. There is, of course, no free lunch: by using fewer and therefore less expressive, parameters, it is not guaranteed that you will get the same performance! You may need to use a specific PEFT method to get optimal results for the model/task combination you want to train. But you will need less memory and possibly less compute during training and may gain features such as fast hot-swapping between trained expert models and less forgetting of previous knowledge compared to full fine-tuning.
```
stevhliu: would be good to add a link to hot-swapping to make it concrete

```suggestion
PEFT methods train as few parameters as possible while aiming for performance comparable to full fine-tuning. Fewer trainable parameters are less expressive, so the same performance isn't guaranteed. In exchange you use less memory, often less compute, and gain features like fast hot-swapping between expert adapters and less forgetting of prior knowledge.
```

docs/source/developer_guides/memory_efficient_training.md
```diff
@@ -0,0 +1,48 @@
...
+
+Especially when targeting large layers like language modeling heads or embedding layers to fine-tune specific tokens it might make sense to look into [using trainable tokens](troubleshooting#using-trainable-tokens).
+
+## Chunked NLL loss
+
+Using [`NLLLoss`](https://docs.pytorch.org/docs/stable/generated/torch.nn.NLLLoss.html) is very common when training language models (or classification tasks, for that matter) but it is usually computed in one go, meaning you will allocate a matrix of size `batch × sequence × vocabulary`. With particularly long sequences or vocabularies this can get expensive fast.
+
+When using [TRL] you can either use the [Liger kernel integration](https://huggingface.co/docs/trl/liger_kernel_integration) or use [Chunked NLLLoss](https://huggingface.co/docs/trl/v1.5.1/en/reducing_memory_usage#chunked-cross-entropy-for-reducing-peak-memory-usage). The latter will split the sequence in chunks of size 256 to keep the maximum memory consumption constant.
```
stevhliu:
```suggestion
When using [TRL](https://huggingface.co/docs/trl) you can either use the [Liger kernel integration](https://huggingface.co/docs/trl/liger_kernel_integration) or use [Chunked NLLLoss](https://huggingface.co/docs/trl/v1.5.1/en/reducing_memory_usage#chunked-cross-entropy-for-reducing-peak-memory-usage). The latter will split the sequence in chunks of size 256 to keep the maximum memory consumption constant.
```

docs/source/developer_guides/memory_efficient_training.md
stevhliu:
```suggestion
🤗 PEFT makes fine-tuning parameter efficient, but not automatically memory efficient. This overview collects tips for cutting training memory and links to the detailed guides.
```

docs/source/developer_guides/memory_efficient_training.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Choosing the right method
+
+Not every PEFT method is built equally and some formulations are easier to build in a memory efficient manner. If you are on a memory budget it makes sense to check out the [PEFT method comparison suite](https://huggingface.co/spaces/peft-internal-testing/PEFT-method-comparison) and filter for **maximum** accelerator memory usage. Average accelerator memory usage can be fairly equal across methods but not every method scales equally with activations and sequence length and is more prone to memory spikes than others.
+
+Especially when targeting large layers like language modeling heads or embedding layers to fine-tune specific tokens it might make sense to look into [using trainable tokens](troubleshooting#using-trainable-tokens).
+
+## Chunked NLL loss
+
+Using [`NLLLoss`](https://docs.pytorch.org/docs/stable/generated/torch.nn.NLLLoss.html) is very common when training language models (or classification tasks, for that matter) but it is usually computed in one go, meaning you will allocate a matrix of size `batch × sequence × vocabulary`. With particularly long sequences or vocabularies this can get expensive fast.
```
stevhliu:
```suggestion
Using [`NLLLoss`](https://docs.pytorch.org/docs/stable/generated/torch.nn.NLLLoss.html) is very common when training language models (or classification tasks). You allocate a matrix of size `batch × sequence × vocabulary`. With particularly long sequences or vocabularies this can get expensive fast.
```

docs/source/developer_guides/memory_efficient_training.md
```diff
@@ -0,0 +1,48 @@
...
+
+# Memory Efficient Training
+
+🤗 PEFT provides you with methods for parameter efficient fine-tuning but that doesn't mean that your training process is memory efficient. This guide is a collection of tips that you can use to improve memory efficiency of your training process. This guide is mostly an overview page that will link you to the respective other guides and offer some tips for specific situations.
+
+## Choosing the right method
+
+Not every PEFT method is built equally and some formulations are easier to build in a memory efficient manner. If you are on a memory budget it makes sense to check out the [PEFT method comparison suite](https://huggingface.co/spaces/peft-internal-testing/PEFT-method-comparison) and filter for **maximum** accelerator memory usage. Average accelerator memory usage can be fairly equal across methods but not every method scales equally with activations and sequence length and is more prone to memory spikes than others.
+
+Especially when targeting large layers like language modeling heads or embedding layers to fine-tune specific tokens it might make sense to look into [using trainable tokens](troubleshooting#using-trainable-tokens).
```
stevhliu:
```suggestion
Consider [using trainable tokens](troubleshooting#using-trainable-tokens) when targeting large layers like language modeling heads or embedding layers to fine-tune specific tokens.
```

stevhliu (commented):
nice work! i focused mainly on `memory_efficient-training` and `memory/overview` in this pass

> the ToC indentation for LoRA variants is broken

i think the doc-builder only supports 3 levels of nesting so maybe flatten the variants section?

> new images are rendered with borders

the doc-builder automatically renders it with a border i believe. i would open an issue on the doc-builder repo for this :)

docs/source/index.md
githubnemo: https://huggingface.co/datasets/huggingface/documentation-images/discussions/626

docs/source/methods/overview.md
githubnemo: I tend to agree but I don't see how we can be helpful without a feature matrix that guides the user's decision. Just stating that there are things to be aware about is not enough IMO. Maybe we should add this to the backlog instead + adding a space that let's you navigate the feature matrix (kinda like a shop with category filters)?

docs/source/index.md
BenjaminBossan: merged

docs/source/methods/overview.md
BenjaminBossan: I agree that providing this information is only semi helpful, but as a new user coming to PEFT, isn't that still something I'd like to know?

docs/source/methods/overview.md
githubnemo: I tried my best at clarifying :)

docs/source/package_reference/lora.md
githubnemo: good catch! I updated the reference to the original LoRA paper but forgot the abstract.

docs/source/methods/overview.md
githubnemo: yes! sorted according to the intro workflow

docs/source/methods/overview.md
githubnemo: Fair, added a note

docs/source/quicktour.md
githubnemo: Definitely. Added.

githubnemo:
Thanks for the reviews!

@BenjaminBossan 

> I like the idea of including a benchmark overview for each PEFT method. Now that we have image generation too, it would be great to add an option to toggle the benchmark, but let's leave that to a future PR. 

The benchmark view already includes a task selector, it is just MetaMathQA right now since the image gen results are not present yet.

> I noticed, however, that not each PEFT method includes the benchmark, e.g. HRA is missing it. Also, some methods like HiRA have the graph but no corresponding data points, but maybe its result was added after the space was deployed?

HRA is missing because we're waiting for https://github.com/huggingface/peft/pull/3185 but HiRA should be present, so probably a timing issue, yes.

> I also wonder if we should not fully remove the legend, as the resulting graph can become quite cramped:

The legend is only present on the LoRA page because I forgot to add the highlight URL parameter, fixed now. Without the highlight parameter the space renders all results.

> There is also a bit of an inconsistency about the legend, e.g. for Lily it only labels the line but not the points. I think it should be removed for simplicity.

No, the chosen color is just coincidentally the same as the other points :/ 
Shouldn't matter too much, though.

@stevhliu 

> > the ToC indentation for LoRA variants is broken
> 
> i think the doc-builder only supports 3 levels of nesting so maybe flatten the variants section?

yes, I settled on having a `Variant:` prefix and removing the section, I think it looks cleaner this way.

> > new images are rendered with borders
> 
> the doc-builder automatically renders it with a border i believe. i would open an issue on the doc-builder repo for this :)

I chose to remove the border with an border style on the img element. Since we're doing raw HTML here anyway I think that's fine but I'm happy to raise the issue if you think that's the better way.

BenjaminBossan (approved):
Overall, I'm happy with the new results and think the docs can be merged, even if there are still a few details that could be improved. Waiting too long will result in countless merge conflicts.

A few small issues I still saw, but that could be fixed later:

1. I still see the legend on the Pareto frontier plot. It only contains the description of the line but still takes up quite some space. Personally, I'd remove that. If we think the reader may not understand what the line means, we can put a small description above/below the plot to explain it. In the dropdown, I cannot choose the image gen task, but that will probably work once the PR is merged.
2. The order of the adapters in the docs is rather arbitrary right now, being neither in order of implementation, popularity, or alphabetical. As discussed, personally, I would prefer alphabetical, possibly with LoRA being allowed the top spot.
3. I'd be fine with moving the "main classes" section below "adapters", I don't think it's very important for most readers.

Again, there is no harm in fixing these things later.

stevhliu (approved):
thanks for this big effort and improvement!

PR #3302, opened 2026-06-05
FIX Bug when targeting an nn.Parameter on the root module

BenjaminBossan:
Using LoRA, we allow to target `nn.Parameter`s directly by using `target_parametes=[...]`. This works by replacing the module that this parameter belongs to with `lora.ParamWrapper`. However, this cannot work if the parameter lives on the root model itself, since that would require replacing the whole model. What actually happened was that the module would self-reference, leading to an infinite recursion. With this fix, we now detect this degenerate case and raise a proper error.

Note that I discovered this only by accident while working on a test with a dummy model. It should be extremely unlikely that any real world use cases would require targeting a parameter on the root model. If we get such reports, we can think about a better solution.


githubnemo (approved):
LGTM

PR #3303, opened 2026-06-05
Separate, embeddable method comparison space

githubnemo:
PR #3300 drafts the idea of embedding the method comparison results into the respective method pages. This calls for a lighter version of the existing space to limit the needed space. This is what `app_embed.py` is.

Most of the common processing has moved to the existing and aptly named `processing.py`.

I think that this is better than having a layout switch in `app.py` as these apps are meant to be as flat as can be to be readable and maintainable.

### DIFF
--- a/method_comparison/app_embed.py
+++ b/method_comparison/app_embed.py
@@ -1,0 +1,268 @@
+# Copyright 2026-present the HuggingFace Inc. team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Gradio app to show the results embedded in the docs for each page.
+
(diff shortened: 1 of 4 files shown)


method_comparison/app_embed.py
```diff
@@ -0,0 +1,268 @@
...
+    compute_pareto_frontier,
+    _get_metric_explanation,
+    _TASK_PARETO_DEFAULTS,
+    get_metric_preferences,
+    format_df,
+    load_task_results,
+)
+
+
+metric_preferences = {
```
BenjaminBossan: Is this dict needed? `get_metric_preferences` should already be doing the job.

BenjaminBossan (changes requested):
Generally LGTM. I haven't checked every change, as I assume that most stuff was just moved around.

Regarding the pareto plots, I noticed is that the method is not always highlighted, e.g. here:

Image

Is it because it is on the Pareto frontier?

Moreover, e.g. for LoRA, not all LoRA points are highlighted, e.g. here:

Image

It could be the same issue (no highlight if on Pareto frontier) or a different one (rs-LoRA is not considered to be LoRA).

method_comparison/app_embed.py
githubnemo: True, removed.

githubnemo:
I assume you tested on https://huggingface.co/spaces/hubnemo/peft-method-comparison-individual - that was running a version before some of the fixes. I've updated the space so you can test there again if you like. Both of these issues are already addressed in the code of the PR.

BenjaminBossan (approved):
Got it, thanks. LGTM.

PR #3304, opened 2026-06-05
TST Add tests for hotswapping targeted LoRA parameters

BenjaminBossan:
The question recently came up if hotswapping works with `target_parameters`. Therefore, I added a test to check it. It turns out that it works indeed.

The usefulness is, however, somewhat reduced because targeting parameters while using `torch.compile` (compiled models are a main use case for hotswapping) leads to re-compilation and/or graph breaks. This is a fundamental limitation of how targeting `nn.Parameter`s is implemented, using `nn.utils.parametrize` to dynamically update the targeted `nn.Parameter`. We can't update it statically, since that would break all kinds of things (e.g. accessing the parameter with `model.foo.bar` would return the parameter *after* applying the LoRA delta weight). Therefore, we must undo the parametrization after the forward step, and this breaks compilation.

This PR additionally documents the fundamental problem with `torch.compile` and `target_parameters`. It also changes a wrong argument type in a test, removes an unused argument in another test, as well as an incorrect comment.


src/peft/tuners/lora/config.py
```diff
@@ -500,15 +500,18 @@ class LoraConfig(PeftConfig):
             disabled. The main use case for this is when the LoRA weights were extracted from fully fine-tuned
             parameters so the bias of those parameters can be taken into account.
         target_parameters (`List[str]`, *optional*)
-            List of parameter names or regex expression of the parameter names to replace with LoRA. This argument
```
BenjaminBossan, line 520: Unrelated change: Regex is not supported here.

src/peft/tuners/lora/config.py
```diff
@@ -521,7 +524,7 @@ class LoraConfig(PeftConfig):
         default=None,
         metadata={
             "help": (
-                "List of module names or regex expression of the module names to replace with LoRA. "
```
BenjaminBossan, line 547: Unrelated change: Regex is not supported here.

tests/test_initialization.py
```diff
@@ -3828,13 +3828,11 @@ def forward(self, X):
         torch.manual_seed(0)
         return ConvModel().to(self.torch_device)
 
-    # this works with all adapters except prompt learning, but we don't test all
```
BenjaminBossan, line 3909: Unrelated change: The comment is just not true.

tests/test_initialization.py
```diff
@@ -3828,13 +3828,11 @@ def forward(self, X):
...
-    # as it is unnecessary and would be slow
     @pytest.mark.parametrize(
         "config",
         [
-            LoraConfig(init_lora_weights=0, target_modules=["lin0"]),
```
BenjaminBossan, line 3914: Unrelated change: `init_lora_weights=0` and `init_lora_weights=False` have the same effect, but since we expect a bool here, let's not pass `0`.

tests/test_integrations.py
```diff
@@ -304,7 +304,6 @@ def test_add_lora_to_mixtral_v5_works(self):
         # Ensure that using LoRA directly with a v5 model still works
         inputs = torch.arange(10).view(1, -1).to(device=self.torch_device)
         model_id = "hf-internal-testing/Mixtral-tiny"
-        lora_id = "peft-internal-testing/mixtral-pre-v5-lora"
```
BenjaminBossan, line 307: Unrelated change: This argument was unused.

BenjaminBossan (commented):
I marked the changes that are not strictly related to the PR.

githubnemo (approved):
Well this is sort of good news. LGTM :)

PR #3307, opened 2026-06-07
fix(save_and_load): save base_layer.bias for bias="lora_only"

### LINKED ISSUE #3306: LoraConfig(bias="lora_only") trains base_layer.bias but save_pretrained/get_peft_model_state_dict drops it
### System Info

Reproduced with:

- peft: 0.19.1
- accelerate: 1.13.0
- transformers: 5.10.2
- torch: 2.12.0+cu132
- safetensors: 0.7.0
- Python: 3.10.20
- Platform: Linux / WSL2, x86_64
- CUDA available: yes, CUDA 13.2

I also reproduced this against PEFT main from source:
- commit: aa2b673629adf5c2340ea835604da73bc747212a
- version reported: 0.19.2.dev0

### Who can help?

@BenjaminBossan  @githubnemo 

### Reproduction

`LoraConfig(bias="lora_only")` correctly marks the wrapped base layer bias as trainable, but `get_peft_model_state_dict()` and `save_pretrained()` do not include that trained bias. Reloading the adapter therefore does not reproduce the trained model output.

Minimal reproducer:

```python
import os
import tempfile

import torch
from torch import nn
from peft.utils import get_peft_model_state_dict


class Toy(nn.Module):
    def __init__(self):
        super().__init__()
        self.proj = nn.Linear(4, 3, bias=True)

    def forward(self, x):
        return self.proj(x)


x = torch.randn(2, 4)


def run(bias):
    torch.manual_seed(123)
    model = get_peft_model(
        Toy(),
        LoraConfig(
            r=2,
            lora_alpha=2,
            target_modules=["proj"],
            bias=bias,
        ),
    )

    # Simulate training by changing all trainable tensors.
    torch.manual_seed(456)
    with torch.no_grad():
            if param.requires_grad:
                param.add_(torch.randn_like(param) * 0.5)

    trainable = [name for name, param in model.named_parameters() if param.requires_grad]

    with torch.no_grad():
        ref = model(x).detach().clone()

    state_bias_keys = [key for key in state_dict if "bias" in key]

    with tempfile.TemporaryDirectory() as tmpdir:
        model.save_pretrained(tmpdir, safe_serialization=True)

        saved = load_file(os.path.join(tmpdir, "adapter_model.safetensors"))
        saved_bias_keys = [key for key in saved if "bias" in key]

        torch.manual_seed(123)
        reloaded = PeftModel.from_pretrained(Toy(), tmpdir, is_trainable=False)

        with torch.no_grad():
            diff = (reloaded(x) - ref).abs().max().item()

    print(f"bias={bias!r}")
    print("  trainable:", trainable)
    print("  get_peft_model_state_dict bias keys:", state_bias_keys)
    print("  saved bias keys:", saved_bias_keys)
    print(f"  roundtrip max diff: {diff:.6f}")


for bias in ["none", "lora_only", "all"]:
    run(bias)
```

Observed output:

```text
bias='none'
  trainable: ['base_model.model.proj.lora_A.default.weight', 'base_model.model.proj.lora_B.default.weight']
  get_peft_model_state_dict bias keys: []
  saved bias keys: []
  roundtrip max diff: 0.000000

bias='lora_only'
  trainable: ['base_model.model.proj.base_layer.bias', 'base_model.model.proj.lora_A.default.weight', 'base_model.model.proj.lora_B.default.weight']
  get_peft_model_state_dict bias keys: []
  saved bias keys: []
  roundtrip max diff: 1.215291

bias='all'
  trainable: ['base_model.model.proj.base_layer.bias', 'base_model.model.proj.lora_A.default.weight', 'base_model.model.proj.lora_B.default.weight']
  get_peft_model_state_dict bias keys: ['base_model.model.proj.base_layer.bias']
  saved bias keys: ['base_model.model.proj.base_layer.bias']
  roundtrip max diff: 0.000000
```

The likely cause seems to be this logic in `peft/utils/save_and_load.py`:

```python
bias_name = k.split("lora_")[0] + "bias"
```

For a key such as:

```text
base_model.model.proj.lora_A.default.weight
```

this constructs:

```text
base_model.model.proj.bias
```

but the actual trained bias key is:

```text
base_model.model.proj.base_layer.bias
```

so the bias is never included in the adapter state dict.

### Expected behavior

When `LoraConfig(bias="lora_only")` marks a wrapped layer's `base_layer.bias` as trainable, that bias should be included by `get_peft_model_state_dict()` and saved by `save_pretrained()`.

Reloading the saved adapter with `PeftModel.from_pretrained()` should reproduce the original PEFT model output, as it does with `bias="all"`.

Alternatively, if exporting `bias="lora_only"` is not intended to be supported for the current tuner-layer structure, PEFT should warn or raise instead of silently dropping trained parameters.

Anai-Guo:
## What

`LoraConfig(bias="lora_only")` correctly marks a targeted layer's bias as trainable, but `get_peft_model_state_dict()` / `save_pretrained()` silently drop it, so a reloaded adapter does not reproduce the trained outputs.

Fixes #3306

## Why

For a targeted layer the original module is wrapped, and its bias lives at `<module>.base_layer.bias`. The bias-key reconstruction was:

```python
bias_name = k.split("lora_")[0] + "bias"   # -> base_model.model.proj.bias
```

but the real trained key is `base_model.model.proj.base_layer.bias`, so the lookup never matched and the bias was excluded from the saved state dict. (`bias="all"` works because it keys off the substring `"bias"` directly.)

## Fix

Look up `<prefix>base_layer.bias` (current tuner-layer structure) in addition to the legacy `<prefix>bias`. The same one-line pattern was applied to the identical `bias="boft_only"` branch for BOFT.

## Test

Added `TestLoraInitialization::test_lora_only_bias_is_saved_and_reloaded`, which perturbs all trainable params (incl. the base_layer bias), asserts the bias key is present in both `get_peft_model_state_dict()` output and the saved `adapter_model.safetensors`, and checks that reloading reproduces the pre-save output.

Reproducer from the issue now round-trips with max diff `0.0` (was `1.21`).

### DIFF
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -145,9 +145,13 @@ def get_peft_model_state_dict(
             for k in state_dict:
                 if "lora_" in k:
                     to_return[k] = state_dict[k]
-                    bias_name = k.split("lora_")[0] + "bias"
-                    if bias_name in state_dict:
-                        to_return[bias_name] = state_dict[bias_name]
+                    prefix = k.split("lora_")[0]
+                    # The trainable bias of a targeted layer lives under the wrapped `base_layer`
+                    # (current tuner-layer structure); older state dicts may store it directly on
+                    # the module, so check both names.
+                    for bias_name in (prefix + "base_layer.bias", prefix + "bias"):
+                        if bias_name in state_dict:
+                            to_return[bias_name] = state_dict[bias_name]
         else:
             raise NotImplementedError
         to_return = {k: v for k, v in to_return.items() if (("lora_" in k and adapter_name in k) or ("bias" in k))}
@@ -182,9 +186,11 @@ def get_peft_model_state_dict(
             for k in state_dict:
                 if "boft_" in k:
                     to_return[k] = state_dict[k]
-                    bias_name = k.split("boft_")[0] + "bias"
-                    if bias_name in state_dict:
-                        to_return[bias_name] = state_dict[bias_name]
+                    prefix = k.split("boft_")[0]
+                    # See the lora_only branch above: the trained bias may live under `base_layer`.
+                    for bias_name in (prefix + "base_layer.bias", prefix + "bias"):
+                        if bias_name in state_dict:
+                            to_return[bias_name] = state_dict[bias_name]
         else:
             raise NotImplementedError
 
--- a/tests/test_custom_models.py
+++ b/tests/test_custom_models.py
@@ -2657,6 +2657,51 @@ class TestPeftCustomModel(PeftCommonTester):
             # This is bad, there was a warning about the bias when there should not have been any.
             self.fail("There should be no warning when bias is set to 'none'")
 
+    def test_lora_only_bias_is_saved_and_reloaded(self, tmp_path):
+        # See https://github.com/huggingface/peft/issues/3306: with bias="lora_only" the trained
+        # base_layer bias of targeted modules lives at "<module>.base_layer.bias" in the current
+        # tuner-layer structure. It must be included in the saved adapter, otherwise reloading does
+        # not reproduce the trained outputs. Conversely, with bias="none" it must not be saved.
+        from peft.utils import get_peft_model_state_dict
+
+        bias_key = "base_model.model.lin0.base_layer.bias"
+
+        torch.manual_seed(0)
+        model = self.transformers_class.from_pretrained("MLP").to(self.torch_device)
+        config = LoraConfig(target_modules=["lin0"], bias="lora_only")
+        model = get_peft_model(model, config)
+
+        # simulate training by perturbing every trainable parameter, including the base_layer bias
+        with torch.no_grad():
+            for _, param in model.named_parameters():
+                if param.requires_grad:
+                    param.add_(torch.randn_like(param) * 0.1)
+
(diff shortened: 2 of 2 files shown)


githubnemo (changes requested):
Let's move the test around a bit, it doesn't make sense to have that in `test_initializations.py` since this is not about initializations at all. I propose to create a new test in `test_custom_models.py` similar to how `test_disable_adapter_with_bias_warns` works by adding the `bias` attribute. But we can keep the basic idea of your test. If we find more methods that are buggy this way, we don't have to fix those right now, instead we can mark them as `xfail`.

It would also be good to test that the bias is *not* in the state dict when the options isn't set. This would fail right now, I think.

BenjaminBossan:
@Anai-Guo Are you still on it?

egeozkoc:
Hi @Anai-Guo and @BenjaminBossan, happy to help finish this if useful.

I can address any remaining review feedback and also add coverage for the same base_layer.bias save gap on OFT (bias="oft_only"), which still reproduces on main. Just say if you would like me to take that on, or if you prefer to keep ownership here.

githubnemo:
@egeozkoc Go ahead, feel free to open a new PR. There was enough time for a reaction :)

PR #3310, opened 2026-06-08
FIX AdaMSS save/load reproduction by making slice_pca SVD deterministic

AshNicolus:
## What does this PR do?

Saving an AdaMSS adapter with `save_pretrained` and reloading it with
`from_pretrained` does not reproduce the model's outputs once the adapter has
been trained.

### Root cause

When an AdaMSS adapter is loaded, the layer is rebuilt from the base weights
(`update_layer` → `slice_pca` → `clustering_Z`). This recomputes `scatter_index`,
the mapping that places each subspace's contribution into the correct output
dimensions. `slice_pca` uses `torch.svd_lowrank`, which draws a random projection
from the global RNG, so the SVD result — and therefore the clustering and
`scatter_index` — depends on the RNG state at construction time. Because that
state differs between saving and loading, the reloaded adapter rebuilds a
different `scatter_index`. The trained `adamss_A`/`adamss_B` weights are restored
correctly, but are then scattered to the wrong output dimensions, so the output
changes.

This stayed hidden because AdaMSS initializes `B = 0`: an untrained adapter is a
no-op and reloads trivially regardless of `scatter_index`. The discrepancy only
appears once the adapter is trained.

`clustering_Z` already pins `KMeans(random_state=...)`, so deterministic
initialization is clearly intended; the only remaining source of randomness was
`svd_lowrank`.

## Fix

Seed a forked RNG around the `svd_lowrank` calls in `slice_pca` so the
decomposition is deterministic and the reconstructed `scatter_index` matches the
one used at save time. `torch.random.fork_rng` leaves the global RNG stream
untouched.

## Tests

Added `TestAdamssSaveLoad::test_save_load_reproduces_output` in
`tests/test_adamss_asa.py`: it trains an AdaMSS adapter, saves it, perturbs the
global RNG state, reloads into a fresh base model, and asserts the outputs match.
The test fails on `main` and passes with this change; the existing AdaMSS tests
still pass.

### DIFF
--- a/src/peft/tuners/adamss/config.py
+++ b/src/peft/tuners/adamss/config.py
@@ -321,11 +321,8 @@ class AdamssConfig(PeftConfig):
         default=0,
         metadata={
             "help": (
-                "Seed for the random projection used by the SVD (torch.svd_lowrank) during initialization. "
-                "AdaMSS rebuilds its subspace decomposition from the base weights when an adapter is loaded, so a "
-                "fixed seed is required for the reloaded adapter to reproduce its trained outputs. The seed is stored "
-                "in the adapter config and reused on load; change it to obtain a different (but still reproducible) "
-                "subspace allocation. Default: 0."
+                "Seed used to deterministically create and rebuild the adapter weights, so that a saved adapter "
+                "reproduces its outputs after loading. Default: 0."
             )
         },
     )
--- a/src/peft/tuners/adamss/config.py
+++ b/src/peft/tuners/adamss/config.py
@@ -317,6 +317,18 @@ class AdamssConfig(PeftConfig):
             )
         },
     )
+    random_seed: int = field(
+        default=0,
+        metadata={
+            "help": (
+                "Seed for the random projection used by the SVD (torch.svd_lowrank) during initialization. "
+                "AdaMSS rebuilds its subspace decomposition from the base weights when an adapter is loaded, so a "
+                "fixed seed is required for the reloaded adapter to reproduce its trained outputs. The seed is stored "
+                "in the adapter config and reused on load; change it to obtain a different (but still reproducible) "
+                "subspace allocation. Default: 0."
+            )
+        },
+    )
 
     def __post_init__(self):
         self.peft_type = PeftType.ADAMSS
--- a/src/peft/tuners/adamss/utils.py
+++ b/src/peft/tuners/adamss/utils.py
@@ -15,7 +15,7 @@
 import torch
 
 
-def slice_pca(tensor, r, device, dtype=torch.float32):
+def slice_pca(tensor, r, device, dtype=torch.float32, random_seed=0):
     """
     Perform slice-wise PCA (SVD) on 4D tensor.
 
@@ -24,6 +24,8 @@ def slice_pca(tensor, r, device, dtype=torch.float32):
         r: rank for low-rank approximation
         device: computation device
         dtype: data type
+        random_seed: seed for the random projection used by ``torch.svd_lowrank``, so the decomposition is
+            deterministic and reproducible across save/load
 
     Returns:
         VVT: Right singular vectors (B, C, r, W) UU: Left singular vectors (B, C, H, r)
@@ -39,15 +41,15 @@ def slice_pca(tensor, r, device, dtype=torch.float32):
     VVT = torch.zeros(B, C, effective_r, W, dtype=dtype, device=device)
 
     # torch.svd_lowrank draws a random projection internally, so its result (and hence the
-    # downstream clustering and scatter_index) depends on the global RNG state. Because the
-    # AdaMSS adapter is rebuilt from the base weights when an adapter is loaded, a different
-    # RNG state at load time would produce a different scatter_index than at save time and the
-    # reloaded adapter would not reproduce its outputs. Seed a forked RNG so the decomposition
-    # is deterministic and reproducible across save/load; this matches the fixed seed already
-    # used for KMeans in clustering_Z and leaves the global RNG stream untouched.
+    # downstream clustering and scatter_index) depends on the RNG state. Because the AdaMSS
+    # adapter is rebuilt from the base weights when it is loaded, a different RNG state at load
+    # time would produce a different scatter_index than at save time and the reloaded adapter
+    # would not reproduce its outputs. Seed a forked RNG with the configurable ``random_seed`` so
+    # the decomposition is deterministic and reproducible across save/load (torch.svd_lowrank does
+    # not accept a generator argument); fork_rng leaves the global RNG stream untouched.
     fork_devices = [device] if torch.device(device).type == "cuda" else []
     with torch.random.fork_rng(devices=fork_devices):
-        torch.manual_seed(0)
+        torch.manual_seed(random_seed)
         for i in range(B):
             for j in range(C):
                 U, _, V = torch.svd_lowrank(tensor[i, j, :, :], q=effective_r, niter=2, M=None)
--- a/src/peft/tuners/adamss/utils.py
+++ b/src/peft/tuners/adamss/utils.py
@@ -38,11 +38,21 @@ def slice_pca(tensor, r, device, dtype=torch.float32):
     UU = torch.zeros(B, C, H, effective_r, dtype=dtype, device=device)
     VVT = torch.zeros(B, C, effective_r, W, dtype=dtype, device=device)
 
-    for i in range(B):
-        for j in range(C):
-            U, _, V = torch.svd_lowrank(tensor[i, j, :, :], q=effective_r, niter=2, M=None)
-            UU[i, j, :, :] = U[:, 0:effective_r]
-            VVT[i, j, :, :] = V[:, 0:effective_r].T
+    # torch.svd_lowrank draws a random projection internally, so its result (and hence the
+    # downstream clustering and scatter_index) depends on the global RNG state. Because the
+    # AdaMSS adapter is rebuilt from the base weights when an adapter is loaded, a different
+    # RNG state at load time would produce a different scatter_index than at save time and the
+    # reloaded adapter would not reproduce its outputs. Seed a forked RNG so the decomposition
+    # is deterministic and reproducible across save/load; this matches the fixed seed already
+    # used for KMeans in clustering_Z and leaves the global RNG stream untouched.
+    fork_devices = [device] if torch.device(device).type == "cuda" else []
+    with torch.random.fork_rng(devices=fork_devices):
+        torch.manual_seed(0)
+        for i in range(B):
+            for j in range(C):
+                U, _, V = torch.svd_lowrank(tensor[i, j, :, :], q=effective_r, niter=2, M=None)
+                UU[i, j, :, :] = U[:, 0:effective_r]
+                VVT[i, j, :, :] = V[:, 0:effective_r].T
     # Return computed matrices (important: ensure callers receive VVT and UU)
     return VVT, UU
 
--- a/src/peft/tuners/adamss/utils.py
+++ b/src/peft/tuners/adamss/utils.py
@@ -24,7 +24,7 @@ def slice_pca(tensor, r, device, dtype=torch.float32, random_seed=0):
         r: rank for low-rank approximation
         device: computation device
         dtype: data type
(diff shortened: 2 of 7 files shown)


tests/test_adamss_asa.py
```diff
@@ -270,3 +272,40 @@ def test_asa_disabled_is_noop(self):
         for layer in layers:
             for p in layer.adamss_A["default"]:
                 assert p.requires_grad
+
+
+class TestAdamssSaveLoad:
+    def test_save_load_reproduces_output(self, tmp_path):
```
BenjaminBossan: Instead of adding a specific test for Adamss, let's add a general test for all PEFT methods.

I created a test for this purpose which you can copy:

```python
    def test_save_load_roundtrip(self, test_name, model_id, config_cls, config_kwargs, tmp_path):
        # An explicit test that when loading a trained model, the outputs from the forward pass remain the same
        model = self.transformers_class.from_pretrained(model_id).to(self.torch_device)
        with torch.inference_mode():
            output_base = model(**X)

        config_kwargs = set_init_weights_false(config_cls, config_kwargs)
        config = config_cls(
            **config_kwargs,
        )
        torch.manual_seed(0)
        model.eval()
        with torch.inference_mode():
            output_before = model(**X)

        # sanity check
        atol, rtol = 1e-5, 1e-5
        assert not torch.allclose(output_base, output_before, atol=atol, rtol=rtol)

        del model

        model = self.transformers_class.from_pretrained(model_id).to(self.torch_device)
        torch.manual_seed(54321)  # ensure that the seed is different from what was used when get_peft_model was called
        model = PeftModel.from_pretrained(model, tmp_path)
        with torch.inference_mode():
            output_after = model(**X)

        assert torch.allclose(output_before, output_after, atol=atol, rtol=rtol)
```

Place the test here: https://github.com/huggingface/peft/blob/b2d61f5f80a7ac540fb572c077d8265a95986db8/tests/test_custom_models.py#L2554-L2556

src/peft/tuners/adamss/utils.py
```diff
@@ -38,11 +38,21 @@ def slice_pca(tensor, r, device, dtype=torch.float32):
...
-            UU[i, j, :, :] = U[:, 0:effective_r]
-            VVT[i, j, :, :] = V[:, 0:effective_r].T
+    # torch.svd_lowrank draws a random projection internally, so its result (and hence the
+    # downstream clustering and scatter_index) depends on the global RNG state. Because the
+    # AdaMSS adapter is rebuilt from the base weights when an adapter is loaded, a different
+    # RNG state at load time would produce a different scatter_index than at save time and the
+    # reloaded adapter would not reproduce its outputs. Seed a forked RNG so the decomposition
+    # is deterministic and reproducible across save/load; this matches the fixed seed already
+    # used for KMeans in clustering_Z and leaves the global RNG stream untouched.
+    fork_devices = [device] if torch.device(device).type == "cuda" else []
```
BenjaminBossan, line 47: The more idiomatic why that we solve such issues in PEFT is to allow passing a `random_seed` argument to the config, create a generator with the seed here, and use that.

BenjaminBossan (changes requested):
Thanks for identifying this bug. It is indeed true that the behavior of AdaMSS partly depends on the RNG and there should be a way for users to fix that via the config.

Regarding the testing, I propose to add a more generic test. When I ran this, I found that AdaMSS indeed fails. Moreover, I also found that PSOFT fails when using `lowrank` for SVD. Could you please fix that too?

Pinging @LonglongaaaGo for AdaMSS and @fei407 for PSOFT.

AshNicolus:
Thanks for the detailed review , addressed all three points:

- Added a `random_seed` argument to AdamssConfig and PsoftConfig. It's stored in
  the adapter config and reused on load, so the SVD-based reconstruction is
  deterministic across save/load. torch.svd_lowrank doesn't accept a `generator`
  argument, so I seed a forked RNG (torch.random.fork_rng + manual_seed(random_seed))
  around the svd_lowrank call, which leaves the global RNG stream untouched. Happy to
  change the mechanism if you'd prefer.
- Fixed PSOFT's psoft_svd="lowrank" path the same way.
- Replaced the AdaMSS-specific test with your generic test_save_load_roundtrip in
  tests/test_custom_models.py. Ran it locally: 292/292 cases pass, and the existing
  AdaMSS/PSOFT tests still pass.

src/peft/tuners/adamss/config.py
```diff
@@ -317,6 +317,18 @@ class AdamssConfig(PeftConfig):
...
     )
+    random_seed: int = field(
+        default=0,
+        metadata={
+            "help": (
+                "Seed for the random projection used by the SVD (torch.svd_lowrank) during initialization. "
+                "AdaMSS rebuilds its subspace decomposition from the base weights when an adapter is loaded, so a "
+                "fixed seed is required for the reloaded adapter to reproduce its trained outputs. The seed is stored "
+                "in the adapter config and reused on load; change it to obtain a different (but still reproducible) "
+                "subspace allocation. Default: 0."
```
BenjaminBossan: Thanks for adding the argument. The description itself is much too detailed for this purpose, as it refers to implementation details. It's sufficient to mention that it is required to deterministically create/rebuild some of the weights.

src/peft/tuners/adamss/utils.py
```diff
@@ -24,6 +24,8 @@ def slice_pca(tensor, r, device, dtype=torch.float32):
         r: rank for low-rank approximation
         device: computation device
         dtype: data type
+        random_seed: seed for the random projection used by ``torch.svd_lowrank``, so the decomposition is
```
BenjaminBossan:
```suggestion
        random_seed: seed for the random projection used by `torch.svd_lowrank`, so the decomposition is
```

src/peft/tuners/adamss/utils.py
```diff
@@ -38,11 +40,21 @@ def slice_pca(tensor, r, device, dtype=torch.float32):
     UU = torch.zeros(B, C, H, effective_r, dtype=dtype, device=device)
     VVT = torch.zeros(B, C, effective_r, W, dtype=dtype, device=device)
 
-    for i in range(B):
-        for j in range(C):
-            U, _, V = torch.svd_lowrank(tensor[i, j, :, :], q=effective_r, niter=2, M=None)
-            UU[i, j, :, :] = U[:, 0:effective_r]
-            VVT[i, j, :, :] = V[:, 0:effective_r].T
+    # torch.svd_lowrank draws a random projection internally, so its result (and hence the
+    # downstream clustering and scatter_index) depends on the RNG state. Because the AdaMSS
```
BenjaminBossan: Everything after the first sentence, i.e. starting with "Because", up until the explanation why `fork_rng` is being used, is overly detailed explanation and can be removed.

src/peft/tuners/psoft/config.py
```diff
@@ -159,6 +159,17 @@ class PsoftConfig(PeftConfig):
             "help": "Number of power iterations used by torch.svd_lowrank when psoft_svd='lowrank'. Only used when psoft_svd='lowrank'. "
         },
     )
+    random_seed: int = field(
+        default=0,
+        metadata={
+            "help": (
+                "Seed for the random projection used by torch.svd_lowrank when psoft_svd='lowrank'. PSOFT rebuilds "
```
BenjaminBossan: Same as above, the explanation is overly detailed.

src/peft/tuners/psoft/layer.py
```diff
@@ -309,15 +312,23 @@ def _build_psoft_ab_cache_buffers(self, adapter_name: str, init_type: str) -> No
...
+    def _compute_svd_factors(self, weight: torch.Tensor, r: int, *, svd_mode: str, niter: int, random_seed: int = 0):
         # weight: (out, in) fp32
         if svd_mode == "full":
             U, S, Vh = torch.linalg.svd(weight.data, full_matrices=False)
             Vr = U[:, :r]  # (out, r)
             Sr = S[:r]  # (r,)
             Uhr = Vh[:r, :]  # (r, in)
         elif svd_mode == "lowrank":
-            U, S, V = svd_lowrank(weight.data, q=r, niter=niter)  # V: (in, r)
+            # torch.svd_lowrank uses a random projection, so the A/B initialization it produces depends on the
```
BenjaminBossan, line 323: Same as above, the explanation is overly detailed.

BenjaminBossan (changes requested):
Thanks for updating so quickly. The changes look good, I just think that the comments and descriptions are overly verbose and can be trimmed. After that, let's give the contributors of these methods a bit of time to react before we merge this.

AshNicolus:
Trimmed the comments and the config help text as suggested , thanks. Ready for
another look whenever the method authors have had a chance to weigh in.

BenjaminBossan:
Thanks for the update @AshNicolus. Let's give the pinged contributors 1 week to check in before proceeding.

BenjaminBossan:
@AshNicolus I just noticed that the Linux tests on CI are failing for the new test when it runs with OFT. It could just be a matter of tolerances or there could be a deeper issue. Could you please investigate?

LonglongaaaGo (commented):
Hey @AshNicolus  and @BenjaminBossan. 
Thanks for the ping and bug fixing. I reviewed the AdaMSS-related change and agree with this fix. Adding `random_seed` to `AdamssConfig` makes sense to keep the SVD-based decomposition deterministic across save/load. From the AdaMSS side, this looks good to me.

AshNicolus:
I looked into the Linux failure. The roundtrip assert itself is fine. For all OFT cases the outputs before saving and after loading are bitwise identical locally (max diff 0.0), so there is no save/load issue in OFT.

The assert that fails is the sanity check, specifically for `Conv2d 3 OFT` (`coft=True`). With the default `eps=6e-5`, COFT keeps the rotation very close to identity, so with `init_weights=False` the adapter only changes the output by about 4e-4 while the outputs themselves are around 40 in magnitude. With `atol=rtol=1e-5` the allclose threshold comes out to about 4e-4 as well, so the threshold and the adapter effect are the same order. On Windows a single element clears the threshold by about 8e-6, on Linux the slightly different kernels push it just below, and `assert not torch.allclose(...)` fails.

The sanity check only needs to catch a no-op adapter, where the diff would be at float32 noise level, so I tightened its tolerances to `atol=rtol=1e-6` and kept `1e-5` for the roundtrip assert. That gives the COFT case roughly 50x more margin. All 292 roundtrip tests pass locally with this change.

I also rebased the branch onto the latest main.

tests/test_custom_models.py
```diff
@@ -2553,6 +2553,41 @@ def test_only_params_are_updated(self, test_name, model_id, config_cls, config_k
...
+        )
+        torch.manual_seed(0)
+        model = get_peft_model(model, config)
+        model.eval()
+        with torch.inference_mode():
+            output_before = model(**X)
+
+        # sanity check; use tighter tolerances than below, as some methods (e.g. COFT) only have a small effect on
+        # the output
+        assert not torch.allclose(output_base, output_before, atol=1e-6, rtol=1e-6)
```
BenjaminBossan: We shouldn't use stricter tolerances here than below for the equality test. Otherwise, the test may pass trivially just by choice of the tolerances. So let's please use the same tolerances (1e-4 or 1e-5).

This would mean that OFT (with COFT option) would fail. I checked the OFT code and couldn't see anything obvious that's wrong. Also, given that this passes on GPU and on Windows is an indicator that the issue is just numerical. So let's add a check if an `OFTConfig` is being used with COFT enabled, in which case the tolerances are changed to 1e-6 instead. Add a small comment to explain this.

BenjaminBossan (changes requested):
Thanks @LonglongaaaGo for checking. And @AshNicolus for investigating, please check my comment about the tolerances.

AshNicolus:
Good point about the equality test passing trivially, thanks. I changed it as you suggested: both asserts now use the same tolerances, set to `1e-6` when an `OFTConfig` with `coft=True` is used and `1e-5` otherwise, with a short comment explaining why. All 292 roundtrip tests pass locally.

BenjaminBossan (approved):
Thanks for adding the RNG seeding to AdaMSS and PSOFT. One week has passed and we got the okay from one of the authors, I think it's safe to merge this.

PR #3312, opened 2026-06-09
Fix app deployment CI issue

githubnemo:
The issue is that the previous step is already moving the `requirements.txt` so the next step doesn't need to do this again. Re-initializing the git repo should work, though.

Log: https://github.com/huggingface/peft/actions/runs/27147578991/job/80133890292#step:5:16


BenjaminBossan (approved):
LGTM

PR #3313, opened 2026-06-10
Method comparison: Image generation results

BenjaminBossan:
First batch of results for image generation are finished. LoRA is not on the pareto frontier. OSF resulted in an OOM error and is thus not included.


githubnemo (approved):
Nice!

> OSF resulted in an OOM error and is thus not included.

Yes, I think you'd need custom training params. See https://github.com/huggingface/peft/blob/main/method_comparison/MetaMathQA/experiments/osf/llama-3.2-3B-rank128/training_params.json

BenjaminBossan:
Crazy that OSF requires more memory here than full fine-tuning. We can tweak the configs later, I merged the PR as is.

PR #3315, opened 2026-06-10
fix: respect layer-specific LoRA-FA scaling in optimizer

eSVeeF:
### Summary
This PR fixes LoRA-FA so the optimizer uses each layer's actual scaling when `rank_pattern` or `alpha_pattern` changes the effective LoRA configuration. The previous implementation used a single global scaling factor, which produced incorrect projected gradients for layers with different scaling

It also adds a regression test with two LoRA layers with different effective ranks and verifies the optimizer state updates match the layer-specific scaling

### What changed
- Updated the LoRA-FA optimizer to resolve and use per-layer scaling values during the LoRA-B projection step
- Kept the legacy `scaling_factor` for backward compatibility with existing callers and tests.
- Added a regression test to cover the case where layer-specific rank patterns produce different scaling behavior across layers.

### Why
When rank or alpha patterns vary by layer, each LoRA module can have a different effective scaling. Using a single global value breaks the intended projection math for those layers. This fix makes the optimizer respect the actual per-layer configuration.

### Validation
- `pytest tests/test_lorafa.py`

### Notes
- Current test explicitly covers `rank_pattern`. I can add a similar `alpha_pattern` test if needed, though it would largely use the same code path and I think it might be redundant
- I'm not sure whether keeping the legacy scalar `scaling_factor` is the right choice right now, feedback welcome


src/peft/optimizers/lorafa.py
```diff
@@ -242,12 +242,36 @@ def create_lorafa_optimizer(
...
     param_groups = [
         {
-            "params": model.parameters(),
+            "params": [param for _, param in named_parameters],
             "lr": lr,
-            "names": [name for name, _ in model.named_parameters()],
-            "scaling_factor": lora_scaling,
+            "names": [name for name, _ in named_parameters],
+            "scaling_factors": [get_scaling_factor(name) for name, _ in named_parameters],
+            "scaling_factor": lora_scaling,  # legacy scalar
```
githubnemo: is there a specific reason to keep the old scaling factor key?

src/peft/optimizers/lorafa.py
```diff
@@ -242,12 +242,36 @@ def create_lorafa_optimizer(
...
     lora_scaling = lora_alpha / math.sqrt(r) if use_rslora else lora_alpha / r
+
+    # this func maps a flattened parameter name back to its LoRA module and adapter
+    # Returns None for non-LoRA parameters and unresolved names
+    def get_scaling_factor(parameter_name: str):
+        if ".lora_" not in parameter_name:
+            return None
+
+        # Split the parameter name to extract the module name and the adapter name
+        module_name, _, lora_suffix = parameter_name.partition(".lora_")
```
githubnemo: I think we should use `rpartition` here to target the 'rightmost' module in case we run in the (hopefully rare) situation of nested PEFT modules.

src/peft/optimizers/lorafa.py
```diff
@@ -91,10 +91,9 @@ def step(self, closure: Callable | None = None):
             loss = closure()
 
         for group in self.param_groups:
-            scaling_factor = group["scaling_factor"]
             param_list = []
             name_list = []
-            for p, n in zip(group["params"], group["names"]):
+            for p, n, scaling_factor in zip(group["params"], group["names"], group["scaling_factors"]):
```
githubnemo: Let's add a grace period for users to upgrade from old optimizer states:

```
# TODO remove after 2026-11-01
if 'scaling_factor' in group and 'scaling_factors' not in group:
    group['scaling_factors'] = group['scaling_factor']
```

githubnemo (commented):
Good find, thanks for the fix.

Some inline comments.

src/peft/optimizers/lorafa.py
eSVeeF: It was purely for backward compatibility. In case existing callers may read ["scaling_factor"] directly or older optimizer state_dict may still rely on that shema. Moreover, current tests use it so they would fail unless we update them. But again, being honest I wasn't very sure what to do with it.

src/peft/optimizers/lorafa.py
eSVeeF: Since the code expects `group['scaling_factors']` to be a list, maybe                `group['scaling_factors'] = [group["scaling_factor"]] * len(group["params"])` is a better idea. WDYT?

src/peft/optimizers/lorafa.py
```diff
@@ -242,12 +242,36 @@ def create_lorafa_optimizer(
...
+    # Returns None for non-LoRA parameters and unresolved names
+    def get_scaling_factor(parameter_name: str):
+        if ".lora_" not in parameter_name:
+            return None
+
+        # Split the parameter name to extract the module name and the adapter name
+        module_name, _, lora_suffix = parameter_name.partition(".lora_")
+        lora_parts = lora_suffix.split(".")
+        if len(lora_parts) < 3:
+            return None
```
eSVeeF, line 304: I just noticed that I made an error since this might introduce a regression for embedding-targeted LoRA adapters, since it will reject names whose LoRA suffix has fewer than three dot-separated parts, for example: `base_model.model.embedding.lora_embedding_A.default`. Maybe relaxing the guard to `if len(lora_parts) < 2:` could work. WDYT?

src/peft/optimizers/lorafa.py
githubnemo: Ah, yes, that is what I meant to write. :)

src/peft/optimizers/lorafa.py
githubnemo: Makes sense. I think it is best to deprecate this properly and change the tests to use `scaling_factors` instead.
Deprecation via a comment, e.g.

```
# TODO remove after 2026-11-01
```

src/peft/optimizers/lorafa.py
githubnemo: Good catch. Best to add a test case for this scenario. A minimum of two elements is probably fine.

eSVeeF:
Thanks for the helpful comments.

Changes made:
- Added a grace period for users to upgrade from old optimizer states. Included also a `FutureWarning`, let me know if this is not needed.
- Updated existing tests that use `scaling_factor` to use now `scaling_factors`
- Replaced `partition()` with `rpartition()` for the cases of running nested PEFT modules.
- Fixed the introduced regression of embedding-targeted LoRA adapters by relaxing the guard in `get_scaling_factor()` from `if len(lora_parts) < 3:` to `< 2:`
- Adding a new test for embedding-targeted LoRA adapters case

All `tests/test_lorafa.py` pass

#### Notes
On test `test_lorafa_init_rslora()` I needed to `use_rslora=True` on the `LoraConfig()`. This is because in the old version of the test, passing use_rslora only to create_lorafa_optimizer was enough. But now after the changes the test reads scaling_factors and those are resolved from each LoRA layer's own scaling stored on the model. That layer scaling is created from the LoRA config on the model, not from the optimizer helper argument.

src/peft/optimizers/lorafa.py
```diff
@@ -93,7 +96,23 @@ def step(self, closure: Callable | None = None):
...
+                scaling_factors = group["scaling_factors"]
+            elif "scaling_factor" in group:
+                if not self._legacy_scaling_factor_warned:
+                    warnings.warn(
+                        "`scaling_factor` is deprecated and will be removed after 2026-11-01. "
+                        "Please use `scaling_factors` instead.",
+                        FutureWarning,
+                    )
+                    self._legacy_scaling_factor_warned = True
+                scaling_factors = [group["scaling_factor"]] * len(group["params"])
```
githubnemo, line 111: Let's make sure that this warning is issued via a test that is also deprecated with the same date.

tests/test_lorafa.py
```diff
@@ -156,14 +169,58 @@ def test_lorafa_init_rslora():
         r=lora_rank,
         lora_alpha=lora_alpha,
         target_modules=["lin0", "lin1"],
+        use_rslora=True,
```
githubnemo: > On test `test_lorafa_init_rslora()` I needed to `use_rslora=True` on the `LoraConfig()`. This is because in the old version of the test, passing use_rslora only to create_lorafa_optimizer was enough. But now after the changes the test reads scaling_factors and those are resolved from each LoRA layer's own scaling stored on the model. That layer scaling is created from the LoRA config on the model, not from the optimizer helper argument.

Hmm yes, that's unfortunate since this could potentially break existing code, i.e. training settings where the `LoraConfig` doesn't use rsLoRA but the optimizer does. Currently that scenario would silently ignore `use_rslora`. 

I think the best way to deal with this, is to deprecate the `use_rslora` flag of LoRA-FA. Let's raise an error if the user supplies `use_rslora` but `model.peft_config[<adapter>].use_rslora` is not set (for all active adapters).

WDYT?

githubnemo (commented):
Good work, thanks for being thorough!

One nit and one bigger point regarding rsLoRA.

tests/test_lorafa.py
eSVeeF: I do not think there is a much better alternative than your suggestion, so going with it.

eSVeeF:
Addressed review feedback:

- Added a regression test to assert the `FutureWarning` for the deprecated `scaling_factor` path
- Handled the rsLoRA mismatch by adding a safeguard and consistency enforcement 
- Updated and extended tests related to the consistency enforcement 

One thing I wasn't sure about was changing the `use_rslora` parameter in `create_lorafa_optimizer()` from `bool = False` to `bool | None = None`. This would let us distinguish between "the user did not pass anything" and "the user explicitly passed False."

Let me know if I missed anything or if you'd like me to make any further changes.

src/peft/optimizers/lorafa.py
```diff
@@ -238,16 +257,65 @@ def create_lorafa_optimizer(
...
+    }
+
+    # TODO remove after 2026-11-01
+    if use_rslora:
+        mismatched_adapters = [adapter for adapter, is_rslora in active_adapter_use_rslora.items() if not is_rslora]
+        if mismatched_adapters:
+            raise ValueError(
+                "`use_rslora=True` was passed to create_lorafa_optimizer, but the following active adapters do not "
+                f"have `use_rslora=True` in the model config: {mismatched_adapters}. Please set `use_rslora=True` "
+                "in LoraConfig for active adapters or stop passing `use_rslora` to the optimizer helper."
```
githubnemo:
```suggestion
                "in LoraConfig for active adapters and stop passing `use_rslora` to the optimizer helper as the parameter is deprecated."
```

tests/test_lorafa.py
```diff
@@ -159,9 +200,80 @@ def test_lorafa_init_rslora():
...
+                "betas": (0.9, 0.999),
+                "eps": 1e-6,
+                "weight_decay": 0.0,
+                "correct_bias": True,
+            }
+        ]
+    )
+    param.grad = torch.ones_like(param)
+
+    # TODO remove after 2026-11-01
```
githubnemo: The whole test can be removed after that deadline if I'm not mistaken. Can you move the comment above the test definition?

githubnemo:
Thanks for the update, I think we're ready soon.

> One thing I wasn't sure about was changing the `use_rslora` parameter in `create_lorafa_optimizer()` from `bool = False` to `bool | None = None`. This would let us distinguish between "the user did not pass anything" and "the user explicitly passed False."

I think it makes sense to make `use_rslora` optional as it is being removed anyway and that way the user doesn't have to specify both flags in config and optimizer init.

We should make sure that the default (`use_rslora=None`) is not raising for LoraConfigs with and without rsLoRA (you can use the same test with a pytest parametrization).

eSVeeF:
Just pushed the updates:

- `create_lorafa_optimizer(..., use_rslora)` is now `bool | None = None`
- When `use_rslora=None`, the optimizer infers whether to use rsLoRA from the active adapter configs, without raising or warning
- If `use_rslora` is explicitly set to `True` or `False` and it doesn't match any active adapter config, it now raises a `ValueError`
- I also consolidated the two init tests into a single parametrized test covering both standard LoRA and rsLoRA configs, and parametrized the mismatch test to cover both mismatch cases

src/peft/optimizers/lorafa.py
```diff
@@ -91,10 +94,25 @@ def step(self, closure: Callable | None = None):
             loss = closure()
 
         for group in self.param_groups:
-            scaling_factor = group["scaling_factor"]
             param_list = []
             name_list = []
-            for p, n in zip(group["params"], group["names"]):
+
+            # TODO remove after 2026-11-01
```
githubnemo:
```suggestion
            # TODO remove after 2026-11-01 and default to `group["scaling_factors"]
```

githubnemo (commented):
Three suggestions, otherwise LGTM

src/peft/optimizers/lorafa.py
eSVeeF: This comment refers to an outdated version of the code, so I couldn't apply your exact suggestion literally but I got what you meant so I applied the same fix to the current code directly

eSVeeF:
Done

githubnemo (approved):
Thanks a bunch :) LGTM

PR #3317, opened 2026-06-11
FEAT Add PEFT shop Gradio space

BenjaminBossan:
A space that looks like a web shop and allows to quickly check the capabilities of different PEFT methods, search, and filter results. Gives code snippets to try to use the models.

This still needs proper review, especially when it comes to the heuristics used to determine the capabilities.

To test this locally, first run the script that determines the capabilities of different PEFT methods:

```sh
python scripts/generate_method_capabilities.py
mv method_capabilities.json method_comparison/peft-shop/
cd method_comparison/peft-shop/
python app.py
```

A demo of the app with can be seen here: https://huggingface.co/spaces/BenjaminB/peft-shop (not dynamically re-deployed on PEFT changes, so could be out of date).

image


BenjaminBossan:
This is mostly vibe-coded but I reviewed it to the best of my abilities. The Gradio logic and the html/css parts are the ones I have least experience with.

method_comparison/peft-shop/app.py
```diff
@@ -0,0 +1,1317 @@
...
+        f'<div class="banner" style="background: linear-gradient({angle}deg, {color}40, {second}40);">'
+        f'<span class="monogram" style="background: {color};">{esc(monogram(method))}</span>'
+        f"<h3>{esc(method)}</h3></div>"
+    )
+
+
+def fantasy_price(method: str) -> float:
+    """The shop-themed, hash-derived fantasy price of a method; always displayed crossed-out in favor of "FREE"."""
+    digest = hashlib.md5(method.encode()).digest()
+    return 9.99 + 10 * (digest[3] % 10)
```
githubnemo, line 492: `digst[3]  # chosen by fair dice roll`

:D

src/peft/tuners/adamss/model.py
```diff
@@ -64,7 +64,7 @@ class AdamssModel(BaseTuner):
     """
 
     prefix: str = "adamss_"
-    tuner_layer_cls = (AdamssLayer,)
+    tuner_layer_cls = AdamssLayer
```
githubnemo, line 67: How did this not break before?

githubnemo (approved):
This looks good and fairly robust regarding feature detection. I think we can move forward with this since it is only explorative and not in PEFT's critical path.

src/peft/tuners/adamss/model.py
BenjaminBossan: It's used with `isinstance`, which accepts a tuple as second argument, so this slipped through.

PR #3318, opened 2026-06-11
FIX layers_to_transform matching wrong index in MoE models

MichalMraz:
The layer-index regex used a greedy .* prefix, so it captured the last numbered segment in the module path instead of the first. For MoE models with paths like "model.layers.1.mlp.experts.0.up_proj" the expert index was used as the layer index, so layers_to_transform silently skipped (or wrongly included) expert modules. Use a lazy .*? prefix so the first numbered segment - the actual transformer layer - is matched, in both the default and the layers_pattern code paths.


tests/test_layers_to_transform_moe.py
BenjaminBossan, line 1: Could you please integrate these tests with the existing ones inside of `test_tuners_utils.py`?

tests/test_layers_to_transform_moe.py
```diff
@@ -0,0 +1,48 @@
...
+        super().__init__()
+        self.model = nn.Module()
+        self.model.layers = nn.ModuleList([ToyMoEBlock() for _ in range(4)])
+
+    def forward(self, x):
+        return x
+
+
+@pytest.mark.parametrize("layers_pattern", [None, "layers", ["h", "layers"]])
+def test_layers_to_transform_filters_by_layer_not_expert_index(layers_pattern):
```
BenjaminBossan: Let's add a short comment to explain what the bug was and reference the original issue here.

tests/test_layers_to_transform_moe.py
```diff
@@ -0,0 +1,48 @@
...
+        target_modules=["q_proj", "up_proj"],
+        layers_pattern=layers_pattern,
+        layers_to_transform=[1],
+        r=2,
+        lora_alpha=4,
+    )
+    model = get_peft_model(ToyMoEModel(), config)
+    targeted = set(model.targeted_module_names)
+
+    assert "model.layers.1.self_attn.q_proj" in targeted
```
BenjaminBossan: Instead of checking individual names, let's declare what the expected set of targeted names is and compare against that.

BenjaminBossan (changes requested):
Thanks for picking this up @MichalMraz. I have a couple of comments regarding the tests, but generally the direction is good.

MichalMraz:
Thank you for the comments, I've addressed them and pushed an update.

tests/test_layers_to_transform_moe.py
MichalMraz: I was not sure whether to keep the exact logic or go for the stricter test of providing the exact list of targeted module names and asserting an equality. Following convention from other test suites, I've decided to go with the stricter test.

BenjaminBossan (approved):
Thanks for picking the PR back up again and fixing the issue, LGTM.

PR #3321, opened 2026-06-12
Add quantization support for SHiRA

githubnemo:
This builds upon PR #3117 and was used to review the effectiveness of the changes. Apparently the changes are good!

On the way, some things were cleaned up like determining the in/out features using the appropriate utility function during init and masking.


BenjaminBossan:
@kkb-code We want to roll out quantization support to most PEFT methods. SHiRA is one of the first ones. If you have time, it would be great if you could double check. The idea is that with those relatively few changes, SHiRA will support multiple quantization types on the base model like BNB, torchao, and gptqmodel.

src/peft/tuners/shira/mask_functions.py
```diff
@@ -55,18 +55,20 @@ def mask_fn(base_layer, r):
...
     random_generator = torch.Generator()
     if random_seed is not None:
         random_generator.manual_seed(random_seed)
     idx = (torch.randperm(base_layer.weight.numel(), generator=random_generator)[:num_shira_weights]).to(
         base_layer.weight.device
     )
-    val = torch.ones_like(idx.type(base_layer.weight.dtype))
-    mask = torch.zeros_like(base_layer.weight.view(1, -1))
+    val = torch.ones(*idx.shape, dtype=bool)
+    mask = torch.zeros(*shape).to(val).view(1, -1)
```
BenjaminBossan: I got an error when I ran this on CUDA because `val` and `mask` were on CPU. Something like this should work:

```python
    device = base_layer.weight.device
    idx = (torch.randperm(base_layer.weight.numel(), generator=random_generator)[:num_shira_weights]).to(device)
    val = torch.ones(*idx.shape, dtype=bool).to(device)
    mask = torch.zeros(*shape).to(val).view(1, -1).to(device)
    mask = mask.scatter_(1, idx.unsqueeze(0), val.unsqueeze(0)).view(shape)
```

src/peft/tuners/shira/layer.py
```diff
@@ -205,13 +211,15 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
         elif self.merged:
             result = self.base_layer(x, *args, **kwargs)
         else:
-            new_weight = copy.deepcopy(self.base_layer.weight.data)
+            base_result = self.base_layer(x)
```
BenjaminBossan: With this change, we avoid the copy but we run the forward pass twice, once with the base layer and once with the SHiRA weights. On my machine, this runs considerably slower: train time 30s -> 80s, eval time 10s -> 70s.

BenjaminBossan (changes requested):
I gave this a try in the MetaMath benchmark and found two issues, please check.

src/peft/tuners/shira/mask_functions.py
githubnemo: I applied your suggestion.

src/peft/tuners/shira/layer.py
githubnemo: I just pushed changes that reverted back to the deepcopy variant (albeit using `clone` this time). Can you check again?

src/peft/tuners/shira/layer.py
```diff
@@ -205,7 +211,8 @@ def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
         elif self.merged:
             result = self.base_layer(x, *args, **kwargs)
         else:
-            new_weight = copy.deepcopy(self.base_layer.weight.data)
+            new_weight = self.get_base_weight().clone()
```
BenjaminBossan, line 251: Unfortunately, reverting to this implementation brings another issue: This requires dequantizing the base weight, but some quantization backends like `torchao_int8_dynamic_activation_int8` don't support that. Going with the `base_layer(x)` is, however, also not desirable because of the slower runtime. Therefore, I would suggest that we use the `clone()` approach when there is no quantization backend or when the quantization backend supports dequantizing, otherwise we go with the `base_layer(x)` approach:

```python
        elif self.quantization_backend and not self.quantization_backend.supports_merge:
            # For the normal forward path, self.get_base_weight() needs to be called, but if the quantization backend
            # doesn't support it, we use a pattern that avoid dequantizing the base weight. The disadvantage is that
            # this is slower.
            base_result = self.base_layer(x, *args, **kwargs)
            new_weight = torch.zeros(self.out_features, self.in_features).to(base_result)

                if active_adapter not in self.shira_weight.keys():
                    continue
                new_weight += self.get_delta_weight(active_adapter)

            result = base_result + F.linear(x, new_weight)
        else:
            new_weight = self.get_base_weight().clone()
            [...]
```

I tried 4bit bnb to see if, when both ways are possible, we should prefer `base_layer(x)` or if we should dequantize + `clone`. I found that:

- dequantizing + `clone`: 75 s train, 19 s eval
- don't dequantizing + `base_layer(x)`: 125 s train, 75 s eval

Therefore, as long as the quantization backend supports it, we should go with the dequantize + clone approach.

I know that `self.quantization_backend.supports_merge` is a bit of a misnomer here, it should be `supports_dequantization`. But let's not tackle that in this PR.

BenjaminBossan (changes requested):
Thanks for the update. Unfortunately, I came across an issue when running the quantization tests with `torchao_int8_dynamic_activation_int8`. I described it below.

src/peft/tuners/shira/layer.py
BenjaminBossan: We could also detect this case and just raise an error.

githubnemo: As discussed, I've added the two branches in forward and I'm issuing a warning (once) that the forward implementation ignores hooks. I've not added explicit tests for the both branches since we test both bnb 4bit and torchao int8, so both branches should be covered already.

src/peft/tuners/shira/layer.py
```diff
@@ -111,6 +112,26 @@ def set_scale(self, adapter, scale):
             return
         self.scaling[adapter] = scale
 
+    @lru_cache(None)  # noqa: B019
```
BenjaminBossan: For my understanding, we need the cache to prevent the warning to be fired by each affected layer? Let's maybe add a comment.

tests/test_shira.py
```diff
@@ -276,3 +277,13 @@ def test_shira_dtypes(self, dtype):
         inputs = torch.randn(5, 10).to(dtype)
         output = peft_model(inputs)  # should not raise
         assert output.dtype == dtype
+
+    def test_shira_warns_about_hooks(self):
```
BenjaminBossan: Can you explain why we expect hooks here?

BenjaminBossan (changes requested):
Thanks for the update. Just two nits.

tests/test_shira.py
githubnemo: Done. I've also made the test more explicit and caught a bug in the caching (I wrongly assumed it was global).

BenjaminBossan:
One of the new tests fails, could you please check?

githubnemo:
Yep, fixed. Please take a look.

BenjaminBossan (approved):
Thanks for fixing, LGTM.

PR #3322, opened 2026-06-12
Minor app embed improvements

githubnemo:
This change solves two issues:

1. removes the pareto frontier line legend item when it is the only item that would be displayed (in highlight method mode)
2. fixes task switching which was broken because the output count was wrong due to remains from copying code from app.py


BenjaminBossan (approved):
Thanks.

PR #3323, opened 2026-06-12
ENH Add PEFT method papers in a structured way

BenjaminBossan:
Right now, there is no structured way to find the paper corresponding to a specific PEFT method. But having this can have advantages, e.g. to create a list of paper links for PEFT methods (see #3317).

To create this PR, I asked a coding agent to first write a script to extract the papers using heuristics like regexes from the PEFT code and docs. The resulting file was used to ground the AI. Once implemented, I manually checked each paper and title (there were indeed a few errors or information that was out-of-date).

As for the design, I went with a dict of lightweight dataclasses on the PEFT method's config. In general, it will contain a single entry, which is a dataclass with the URL and title of the paper. The key will correspond to the value of the `PeftType`. If there are multiple papers, the dict has multiple entries, where one is considered the 'main' paper based on the key.

For variations, the dict may also contain multiple entries. E.g. for DoRA, it resides in the `LoraConfig`'s papers attribute under the `"DoRA"` key (there is no corresponding `PeftType` so the spelling doesn't need to be all caps). There is also a very brief description for the variants that describes how to use them.

When it comes to URLs, I preferred HF papers > Arxiv > other sources. When the paper title changed over time, I took the most up-to-date title. I omitted papers that are not implemented as a PEFT method or PEFT method variant (so e.g. no QLoRA paper).

Even though the `papers` attribute lives on the config class, it is IMO better not to include it in the `adapter_config.json`. Otherwise, we create bigger files with arguably unnecessary info (not needed to load the PEFT adapter) and all existing `adapter_config.json`s would be updated, resulting on warnings when loading them with older PEFT versions.

PR #3317 will be updated based on this PR.


BenjaminBossan:
Converting to a draft for now to think more deeply about the solution.

Potential use cases:

1. Auto-generate parts of the PEFT docs. Think about embedding PEFT shop.
2. Spaces like the benchmark results auto-link to the used PEFT method/variant.
3. HF Hub parses the `adapter_config.json` and detects the used PEFT method/variant and, e.g., links the papers. Might require a solution that doesn't depend on Python.

PR #3324, opened 2026-06-14
fix: preserve string target_modules/target_parameters in MoE config conversion

### LINKED ISSUE #3229: [BUG] peft 0.19 target_modules (str) use `set`

horizonbymuneeb:
## Summary

When `target_modules` (or `target_parameters`) is passed as a string such as the documented `all-linear` sentinel, the MoE config conversion helper was wrapping it in `set(value or [])`, which silently turns the string into a set of individual characters. The downstream target-matching code then ran against nonsense keys and produced errors that were hard to trace back to the helper.

This PR wraps string values in a single-element set, leaves other iterables unchanged, and defaults `None` to an empty set.

## What does this PR do?

- Fixes a latent bug in `_convert_peft_config_moe` where string-valued `target_modules` / `target_parameters` were silently split into per-character entries.
- Adds a small helper `_to_target_set` so the normalization is explicit and reusable.
- Adds a dedicated regression test (`tests/test_transformers_weight_conversion.py`) covering string, list, tuple, and `None` for both fields.

## Why is this change needed?

Reproduces and resolves [#3229](https://github.com/huggingface/peft/issues/3229), where users hit confusing downstream errors after passing `target_modules="all-linear"` (or similar string sentinels) on MoE models such as Mixtral.

## How was this tested?

```sh
pip install -e ".[test]"
pytest tests/test_transformers_weight_conversion.py -v
```

All 8 new cases pass. I also confirmed that 4 of the string/list cases fail on `main` (without the fix) and pass with the fix applied, demonstrating the test is a real regression guard rather than a tautology.

Existing `tests/test_target_parameters.py` and `tests/test_tuners_utils.py` continue to pass locally; `make style` (ruff) is clean for the changed files.

## Checklist

- [x] Tests added
- [x] No breaking changes
- [x] Style checks pass (ruff)

### DIFF
--- a/src/peft/utils/transformers_weight_conversion.py
+++ b/src/peft/utils/transformers_weight_conversion.py
@@ -368,8 +368,18 @@ def _convert_peft_config_moe(peft_config, model_type: str) -> None:
     if not fused_targets:
         return
 
-    peft_config.target_parameters = set(peft_config.target_parameters or [])
-    peft_config.target_modules = set(peft_config.target_modules or [])
+    # Normalize string targets (e.g. "all-linear") to a single-element list before
+    # converting to a set, otherwise ``set("all-linear")`` would yield the individual
+    # characters instead of the intended module name. See issue #3229.
+    def _to_target_set(value):
+        if value is None:
+            return set()
+        if isinstance(value, str):
+            return {value}
+        return set(value)
+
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Duplicate of #3232.

PR #3325, opened 2026-06-15
Docs: Make API reference sorted and collapsed

githubnemo:
This is a minor change on top of #3300: the API reference should be sorted in the same way the methods are sorted (alphabetically) and since it is uncurated it should be collapsed by default.

The section "adapters" is now called "tuners" to more adequately model the API.


BenjaminBossan (approved):
Makes sense, thanks.

PR #3335, opened 2026-06-17
FIX PEFT shop app description extraction

BenjaminBossan:
The heuristic for trying to extract a short description of a PEFT method was not working well as it would sometimes catch HTML like:

`<div class="flex justify-center">`

The logic is now refined to work more robustly in my testing. It could probably be replaced by some more principled approach, this is a quick fix.

### DIFF
--- a/method_comparison/peft-shop/app.py
+++ b/method_comparison/peft-shop/app.py
@@ -139,29 +139,42 @@ DESCRIPTION_OVERRIDES: dict[str, str] = {}
 BENCHMARK_SPACE_URL = "https://huggingface.co/spaces/peft-internal-testing/PEFT-method-comparison"
 
 
+# Sanity check: Minimum length, in characters, of an extracted paragraph for it to be accepted as a method description
+MIN_DESCRIPTION_LENGTH = 60
+
+
 def extract_description(docs_path: Path) -> str | None:
     """Extract a short description from a method's documentation page.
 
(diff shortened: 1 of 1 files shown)


githubnemo (approved):
This is OK. I think I would have used `html.parser.HTMLParser` + the stuff you used to remove (custom) markdown but I'm OK with these changes since it is only semi-critical (trusted source of the markup). Thanks for fixing!

BenjaminBossan:
> This is OK. I think I would have used `html.parser.HTMLParser` + the stuff you used to remove (custom) markdown but I'm OK with these changes since it is only semi-critical (trusted source of the markup). Thanks for fixing

I'm honestly thinking about adding something like a `_blurb` attribute instead of trying to parse the docs. I think we discussed that in the past. We could use the PEFT release notes to try to backfill those and in the future ask contributors to write it.

githubnemo:
> > This is OK. I think I would have used `html.parser.HTMLParser` + the stuff you used to remove (custom) markdown but I'm OK with these changes since it is only semi-critical (trusted source of the markup). Thanks for fixing
> 
> I'm honestly thinking about adding something like a `_blurb` attribute instead of trying to parse the docs. I think we discussed that in the past. We could use the PEFT release notes to try to backfill those and in the future ask contributors to write it.

Yeah we did - as long as some kind of formatting is allowed (e.g., markdown) I think we can place it in the code as well. But I think we need the ability to cross-reference other methods with hyperlinks.

PR #3336, opened 2026-06-17
refactor: dedupe prompt-learning kwarg adjustments in peft_model

eSVeeF:
This PR cleans up repeated prompt-learning keyword handling in `peft_model.py`. I moved the shared `position_ids` and `token_type_ids` adjustment logic into a small private helper. The change keeps behavior the same.

Tests run and passed locally:
- `tests/test_decoder_models.py -k prefix_tuning_offsets_position_ids_in_forward`
- `tests/test_cartridge.py -k position_ids`

### DIFF
--- a/src/peft/peft_model.py
+++ b/src/peft/peft_model.py
@@ -1614,23 +1614,21 @@ class PeftModel(PushToHubMixin, torch.nn.Module):
         self,
         peft_config: PeftConfig,
         *,
-        position_ids_kwargs: dict[str, Any],
-        token_type_ids_kwargs: Optional[dict[str, Any]] = None,
+        kwargs: dict[str, Any],
+        remove_token_type_ids: bool = True,
     ) -> None:
         """Adjust prompt-learning kwargs for supported position and token ids"""
-        if position_ids_kwargs.get("position_ids", None) is not None:
+        if kwargs.get("position_ids", None) is not None:
             if peft_config.peft_type in (PeftType.PREFIX_TUNING, PeftType.CARTRIDGE):
                 # Offset position_ids by num_virtual_tokens to account for the KV cache prefix
-                position_ids_kwargs["position_ids"] = (
-                    position_ids_kwargs["position_ids"] + peft_config.num_virtual_tokens
-                )
+                kwargs["position_ids"] = kwargs["position_ids"] + peft_config.num_virtual_tokens
             else:
                 warnings.warn("Position ids are not supported for parameter efficient tuning. Ignoring position ids.")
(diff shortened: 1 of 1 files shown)


src/peft/peft_model.py
```diff
@@ -2274,21 +2284,11 @@ def prepare_inputs_for_generation(self, *args, task_ids: Optional[torch.Tensor]
...
-
-            if kwargs.get("token_type_ids", None) is not None:
-                warnings.warn(
-                    "Token type ids are not supported for parameter efficient tuning. Ignoring token type ids"
-                )
-                kwargs["token_type_ids"] = None
+            self._adjust_prompt_learning_kwargs(
+                peft_config,
+                position_ids_kwargs=model_kwargs,
+                token_type_ids_kwargs=kwargs,
```
BenjaminBossan: Here, it should be correct to pass `token_type_ids_kwargs=model_kwargs`, using `kwargs` is incorrect in the existing code.

BenjaminBossan (changes requested):
Thank your for this PR to reduce code duplication in `peft_model.py`.

When I reviewed the code, I was at first surprised why you have `position_ids_kwargs` and `token_type_ids_kwargs` as separate arguments in `_adjust_prompt_learning_kwargs`. The reasons seem to be:

1. Sometimes, `token_type_ids` are adjusted and sometimes, they are not adjusted.
2. In one case, `position_ids` are adjusted in `model_kwargs` and `token_type_ids` are adjusted in `kwargs`.

The latter is almost certainly a mistake and should use `model_kwargs` instead of `kwargs`. This means that we can unify the two arguments and only pass `kwargs` instead of having two separate arguments. To control 1., we can just add a boolean argument like `remove_token_type_ids`. WDYT?

eSVeeF:
Thanks for the review. 

Yes, your assumed reasons are correct, that is why I used separate args. To be honest, I couldn't figure out why that case adjusted `model_kwargs` for `position_ids` but `kwargs` for `token_type_ids`, and since I wanted to do a safe refactor, I kept the exact same logic just in case.

I agree that it was probably an oversight, so yes we should use `model_kwargs` instead of `kwargs` and unify the two input arguments into one.

eSVeeF:
Fix mistaken use of `kwargs` instead of `model_kwargs`
Unified `position_ids_kwargs` and `token_type_ids_kwargs` into just one `kwargs`
Added `remove_token_type_ids` as boolean argument

BenjaminBossan (approved):
Thanks for deduplicating the code, LGTM.

PR #3342, opened 2026-06-20
FEAT Add DEFT (Decompositional Efficient Fine-Tuning (NeurIPS'25)

### LINKED ISSUE #3349: Adding support of DEFT, a decompositional efficient fine-tuning framework
### Feature request

It adapts a pre-trained weight matrix by decomposing its update into two components with two trainable matrices: (1) a projection onto the complement of a low-rank subspace spanned by a low-rank matrix, and (2) a low-rank update. The single trainable low-rank matrix defines the subspace, while the other trainable low-rank matrix enables flexible parameter adaptation within that subspace.

### Your contribution

Yes, I am following the code from the Official codebase.

GitHub: https://github.com/MAXNORM8650/DEFT

Paper: https://arxiv.org/abs/2509.22793

MAXNORM8650:
## What does this PR do?

Adds **DEFT (Decompositional Efficient Fine-Tuning)** as a new PEFT method.

DEFT keeps the base weight `W` frozen and decomposes the update into two trainable components — a projection that removes a low-rank subspace of `W`, and a low-rank update that injects new content into that subspace:

```
W' = (I - P_proj) @ W + g * Q_P @ R
```

Per target `nn.Linear`, it learns a projection direction `P` (`out_features x r`) and an injection matrix `R` (`r x in_features`); `g` is an optional learnable sigmoid gate. The projector `P_proj` is set by `decomposition_method`:

- `"relu"` (default): `Q_P = P`, `P_proj = P @ relu(P).T` (non-orthogonal)
- `"qr"`: `Q_P = qr(P)`, `P_proj = Q_P @ Q_P.T` (orthogonal)

Unlike LoRA's purely additive `W + BA`, DEFT can *replace* a subspace of the pretrained weight rather than only add to it — aimed at personalization/editing where new content should overwrite rather than coexist.

Paper: **DEFT: Decompositional Efficient Fine-Tuning for Text-to-Image Models** (NeurIPS 2025) — https://proceedings.neurips.cc/paper_files/paper/2025/hash/93a34a7138bdad95e874018d5f491cc6-Abstract-Conference.html · reference implementation: https://github.com/MAXNORM8650/DEFT

## Implementation notes

- New tuner package `src/peft/tuners/deft/` (`config`, `layer`, `model`) registered via `register_peft_method`; `PeftType.DEFT`, target-module mapping and public exports wired in. Currently `nn.Linear` only.
- **Identity at init** by default (`init_weights=True`): `R` is initialized so the update is exactly zero at init, i.e. training starts from the pretrained weights and learns the injection. This avoids the immediate degradation caused by removing a subspace of `W` at init.
- **Factored low-rank delta** `Q_P @ (g*R - right.T @ W)` — the `out x out` projection is never materialized, so per-step compute is LoRA-order; the delta is also cached in eval/inference so generation does not recompute it per token.
- **Mergeable**: `merge`/`unmerge` are exact (the applied delta is cached for an exact unmerge, since the residual projection is not invertible). Multi-adapter merge sums independent deltas computed from the original `W`, matching `forward`.
- Effective rank is capped to `min(r, out_features)` per layer.

## Results

The authoritative results are in the paper (text-to-image personalization/editing on Dreambooth, Dreambench Plus, InsDet, VisualCloze with Stable Diffusion and a unified model). The numbers below are quick reproductions inside PEFT's own `method_comparison` harnesses, comparing DEFT to LoRA/OFT with matched target modules and rank.

### Image generation — FLUX.2-klein-4B, DreamBooth, 750 steps (`method_comparison/image-gen`)

Full DreamBooth metric suite (CLIP-T = prompt following ↑, CLIP-I = subject fidelity ↑, DINOv1/DINOv2 = subject fidelity ↑, drift ↓):

| method | CLIP-T | CLIP-I | DINOv1 | DINOv2 | drift | params |
|---|---|---|---|---|---|---|
| DEFT-qr | **0.352** | 0.809 | 0.633 | **0.714** | 0.304 | 38.3 M |
| OFT | 0.341 | **0.814** | **0.695** | 0.707 | **0.249** | 4.8 M |
| DEFT-relu | 0.336 | 0.784 | 0.573 | 0.713 | 0.307 | 38.3 M |
| LoRA | 0.319 | 0.778 | 0.689 | 0.674 | 0.282 | 38.3 M |

DEFT is competitive with LoRA/OFT — DEFT-qr leads prompt following (CLIP-T) and DINOv2 and beats LoRA on 4 of 5 metrics; its weaker spot is DINOv1. **Caveat: single subject, single seed — run-to-run variance is ≈±0.05 (larger than several of these gaps), so treat as directional, not a ranking.**

### LLM — MetaMathQA → GSM8K, 100-step smoke (`method_comparison/MetaMathQA`)

(`forgetting` = increase in Wikipedia per-token loss after training, lower is better.)

| method | Llama-3.2-3B acc / forget | Llama-3.1-8B acc / forget |
|---|---|---|
| LoRA | 31.3% / −0.03 | 75.1% / −0.48 |
| OFT | 31.3% / +0.07 | 76.3% / +0.07 |
| DEFT-relu | 29.7% / **−0.28** | 73.4% / **−0.85** |
| DEFT-qr | 29.0% / **−0.24** | 73.2% / **−0.81** |

DEFT shows the strongest knowledge preservation (most-negative forgetting) at both scales, at a small accuracy cost. **These are 100-step smoke runs (directional, not converged).**

## Tests

- `pytest tests/test_custom_models.py -k "DEFT or deft"` → **285 passed, 7 skipped**.
- `ruff format` + `ruff check` clean (ruff matching `setup.py`).
- Added `TEST_CASES` entries (relu + qr, `modules_to_save`, gating) and `DEFT Same`/`DEFT Different` multiple-active-adapter cases. Docs page + toctree entries added.

It fixes https://github.com/huggingface/peft/issues/3349. 

## Before review
- [x] Tracking issue / maintainer approval @kashif: https://huggingface.co/blog/peft-beyond-lora 
- [x] AI assistance was used in preparing this change; I (the submitter) have reviewed every changed line and can defend the design end-to-end.


MAXNORM8650:
## Usage

```python
from peft import DeftConfig, get_peft_model

config = DeftConfig(
    r=16,
    target_modules=["q_proj", "v_proj"],   # or e.g. attention/MLP linears of a diffusion transformer
    decomposition_method="relu",            # "relu" (default) or "qr"
    init_weights=True,                       # identity at init (start from the pretrained weights)
)
model = get_peft_model(base_model, config)

# ... standard training loop ...

model.save_pretrained("deft-adapter")
# later:
# from peft import PeftModel
# model = PeftModel.from_pretrained(base_model, "deft-adapter")
# merged = model.merge_and_unload()   # fold the adapter into the base weights for inference
```

## Findings

**From the paper** (authoritative; on Stable Diffusion and a unified model, evaluated on Dreambooth, Dreambench Plus, InsDet and VisualCloze): DEFT achieves state-of-the-art text-to-image personalization/editing. Because the update *replaces* a low-rank subspace of `W` rather than only adding to it (as LoRA does), the paper reports a better balance of target-distribution alignment, learning new concepts from few images, and preserving the pretrained model's instruction-following/editability.

**From our reproductions in PEFT's `method_comparison` harnesses for PR** (directional — see caveats below):
- On the FLUX DreamBooth task, DEFT (with `qr`) is competitive with LoRA/OFT across the DreamBooth metric suite — leading prompt-following (CLIP-T) and DINOv2 and beating LoRA on 4 of 5 metrics; its weaker metric is DINOv1.
- On the MetaMathQA→GSM8K LLM task, DEFT shows the strongest knowledge preservation (lowest forgetting) of the compared methods, at a small accuracy cost.
- DEFT uses the same parameter count and comparable memory/speed to LoRA (`qr` is ~20% slower due to the QR step); OFT remains the most parameter-efficient baseline.

> Caveat: the LLM numbers are 100-step smoke runs (not converged) and the image-gen numbers are a single subject / single seed (run-to-run variance ≈±0.05, larger than several reported gaps). Treat them as directional sanity checks; the paper's experiments are the authoritative evaluation.

docs/source/package_reference/deft.md
```diff
@@ -0,0 +1,68 @@
+<!--Copyright 2025 The HuggingFace Team. All rights reserved.
```
BenjaminBossan: Let's use the current year here and throughout.

```suggestion
<!--Copyright 2026 The HuggingFace Team. All rights reserved.
```

docs/source/package_reference/deft.md
```diff
@@ -0,0 +1,68 @@
...
+
+# DEFT: Decompositional Efficient Fine-Tuning for Text-to-Image Models
+
+[DEFT](https://proceedings.neurips.cc/paper_files/paper/2025/hash/93a34a7138bdad95e874018d5f491cc6-Abstract-Conference.html)
+(Decompositional Efficient Fine-Tuning) is a parameter-efficient fine-tuning method for text-to-image models. It
+decomposes the update of a frozen weight matrix `W` into two trainable components: a projection that removes a low-rank
+subspace from `W`, and a low-rank update that injects new content into that subspace. This formulation is designed to
+balance aligning with a target distribution, learning new concepts from a few images (personalization), and preserving
+the pretrained model's instruction-following ability and editability.
+
```
BenjaminBossan, line 25: Let's also mention what DEFT is especially good at, so that readers can see when they should consider using it and when they should prefer other PEFT methods.

src/peft/tuners/deft/config.py
```diff
@@ -0,0 +1,175 @@
...
+        metadata={
+            "help": "List of module names or regex expression of the module names to replace with DEFT.",
+            "example": "For example, ['q', 'v'] or '.*decoder.*(SelfAttention|EncDecAttention).*(q|v)$' ",
+        },
+    )
+    exclude_modules: Optional[Union[list[str], str]] = field(
+        default=None,
+        metadata={"help": "List of module names or regex expression of the module names to exclude from DEFT."},
+    )
+    decomposition_method: str = field(
```
BenjaminBossan:
```suggestion
    decomposition_method: Literal["relu", "qr"] = field(
```

src/peft/tuners/deft/config.py
```diff
@@ -0,0 +1,175 @@
...
+            of the passed strings. If this is specified as 'all-linear', then all linear modules are chosen, excluding
+            the output layer. If this is not specified, modules will be chosen according to the model architecture. If
+            the architecture is not known, an error will be raised -- in this case, you should specify the target
+            modules manually.
+        exclude_modules (`Optional[Union[List[str], str]]`):
+            The names of the modules to not apply the adapter. When passing a string, a regex match will be performed.
+            When passing a list of strings, either an exact match will be performed or it is checked if the name of the
+            module ends with any of the passed strings.
+        decomposition_method (`str`):
+            How the projector `P_proj` is derived from `P`. Either `"relu"` (default, non-orthogonal `P @ relu(P).T`)
```
BenjaminBossan, line 56: In the docstring/help, let's mention the default values.

src/peft/tuners/deft/layer.py
```diff
@@ -0,0 +1,319 @@
...
+        if init_weights:
+            # Identity initialization: choose R so that the delta is exactly zero at init, i.e. the adapted weight
+            # equals the base weight. With delta = Q_P @ (g * R - right.T @ W) (see `_compute_delta`), delta == 0 holds
+            # for R = (right.T @ W) / g. This starts training from the pretrained weights and learns the injection,
+            # avoiding the immediate "forgetting" caused by removing a sub-space of W.
+            P = self.deft_P[adapter_name].detach().to(base_weight.device)
+            _, right = self._project(P, adapter_name)
+            R_init = right.transpose(0, 1) @ base_weight.detach().to(torch.float32)
+            if adapter_name in self.deft_gate.keys():
+                # the gate is moved to the base-layer device only after reset, so move it explicitly here
```
BenjaminBossan: Could you please comment on why casting to float32 is needed?

src/peft/tuners/deft/layer.py
```diff
@@ -0,0 +1,319 @@
...
+        r: int = 0,
+        **kwargs,
+    ) -> None:
+        super().__init__()
+        DeftLayer.__init__(self, base_layer, **kwargs)
+        self._active_adapter = adapter_name
+        self.update_layer(adapter_name, r, config=config, **kwargs)
+
+    def _project(self, P: torch.Tensor, adapter_name: str):
+        """Return ``(Q_P, right)`` (both float32) such that the projector is ``P_proj = Q_P @ right.T`` and the
```
BenjaminBossan: Let's use single backticks everywhere in the docstrings.

src/peft/tuners/deft/layer.py
```diff
@@ -0,0 +1,319 @@
...
+        injection uses ``Q_P``.
+
+          - ``relu``: ``Q_P = P``, ``right = relu(P)`` -> ``P_proj = P @ relu(P).T`` (non-orthogonal)
+          - ``qr``: ``Q_P = qr(P)``, ``right = Q_P`` -> ``P_proj = Q_P @ Q_P.T`` (orthogonal)
+        """
+        P = P.to(torch.float32)
```
BenjaminBossan, line 199: Let's add a comment why float32 is needed.

src/peft/tuners/deft/layer.py
```diff
@@ -0,0 +1,319 @@
...
+            Q_P = P
+            right = torch.relu(P)
+        elif method == "qr":
+            Q_P, _ = torch.linalg.qr(P)
+            right = Q_P
+        else:
+            raise ValueError(f"Unknown decomposition_method '{method}'.")
+        return Q_P, right
+
+    def _compute_delta(self, adapter_name: str) -> torch.Tensor:
```
BenjaminBossan: This method could just be `get_delta_weight`, right?

src/peft/tuners/deft/layer.py
```diff
@@ -0,0 +1,319 @@
...
+                new_weight = base_layer.weight.data.clone() + delta_weight
+                if not torch.isfinite(new_weight).all():
+                    raise ValueError(
+                        f"NaNs detected in the merged weights. The adapter {active_adapter} seems to be broken"
+                    )
+                base_layer.weight.data = new_weight
+            else:
+                base_layer.weight.data = base_layer.weight.data + delta_weight
+            # cache the exact delta applied so unmerge can reverse the (non-invertible) projection
+            self._cached_merge_delta[active_adapter] = delta_weight
```
BenjaminBossan: I wonder if we should offer an option to offload the `delta_weight`s. If we have many of those, they could require quite a big amount of memory, right?

src/peft/tuners/deft/layer.py
```diff
@@ -0,0 +1,319 @@
...
+            result = self.base_layer(x, *args, **kwargs)
+        else:
+            base_layer = self.get_base_layer()
+            orig_weight = base_layer.weight.data
+            new_weight = orig_weight
+
+            for active_adapter in self.active_adapters:
+                if active_adapter not in self.deft_P.keys():
+                    continue
+                x = self.deft_dropout[active_adapter](x)
```
BenjaminBossan: It doesn't really make sense to call `dropout(x)` in this loop if `x` is not being used, right? Also, doesn't this mean we apply the DEFT dropout to the base weight too?

src/peft/tuners/deft/layer.py
```diff
@@ -0,0 +1,319 @@
...
+                new_weight = new_weight + self._delta(active_adapter)
+
+            bias = self.base_layer.bias
+            if self.cast_input_dtype_enabled:
+                x = self._cast_input_dtype(x, new_weight.dtype)
+                bias = self._cast_input_dtype(bias, new_weight.dtype)
+            else:
+                x = x.to(new_weight.dtype)
```
BenjaminBossan: Shouldn't it be enough to call `x = self._cast_input_dtype(x, new_weight.dtype)` (same with `bias`)? The method already checks if casting should be performed.

MAXNORM8650:
These errors seem to come from https://github.com/huggingface/peft/pull/3291. How do it fix it?

src/peft/tuners/deft/layer.py
```diff
@@ -0,0 +1,319 @@
...
+            result = F.linear(input=x, weight=new_weight, bias=bias)
+
+        result = result.to(previous_dtype)
+        return result
```
BenjaminBossan, line 344: One issue I have with the `foward` implementation is that it relies on constructing W' and calling `F.linear`. This has a few disadvantages, e.g. we bypass any possible hooks on the base layer's forward pass (think: sharding) and it makes the code hard to adapt for quantization support.

In general, we prefer to have a "LoRA-style" forward pass, where we don't compute `W + W_delta` but instead calculate the activations directly on the LoRA weights, so basically: `result = base_layer(x) + B(A(x))`.

I did a few iterations with Claude and came up with a new forward implementation for DEFT:

```python
        previous_dtype = x.dtype

        if self.disable_adapters:
            if self.merged:
                self.unmerge()
        elif self.merged:
        elif not any(a in self.deft_P.keys() for a in self.active_adapters):  # no active adapter
        else:
            bias = base_layer.bias

            active = [a for a in self.active_adapters if a in self.deft_P.keys()]
            base_out = self.base_layer(x, *args, **kwargs)
            compute_dtype = torch.float32
            result = base_out.to(compute_dtype)
            # bias-free base product for the projection/correction term
            if bias is not None:
                Wx32 = result - bias.to(compute_dtype)
            else:
                Wx32 = result

            x = self._cast_input_dtype(x, compute_dtype)
            for active_adapter in active:
                Q_P, right = self._project(self.deft_P[active_adapter], active_adapter)
                R = self.deft_R[active_adapter].to(compute_dtype)
                if self.deft_use_gating[active_adapter]:
                    R = R * torch.sigmoid(self.deft_gate[active_adapter].to(compute_dtype))

                x_drop = self.deft_dropout[active_adapter](x)
                Rx = F.linear(x_drop, R)
                inj = Rx @ Q_P.transpose(0, 1)
                corr = (Wx32 @ right) @ Q_P.transpose(0, 1)
                result = result + inj - corr

        return result
```

It would be great if you could check if this corresponds to the existing implementation. I know that it skips the caching of the deltas but that doesn't seem to affect performance. I tested on the MetaMathQA benchmark and got:

```
# old
{"step": "  250", "samples": "   1000", "lr": "5.00e-05", "loss avg": "1.1271", "valid acc": "0.300", "gen valid tokens": 10464, "train time": "31.7s", "eval time": "9.4s", "train tokens / sec": "6688", "mem allocated": "6818694187", "mem reserved": "15835569586", "elapsed time": "1min 3s"}

{"step": "  500", "samples": "   2000", "lr": "1.00e-04", "loss avg": "0.7782", "valid acc": "0.360", "gen valid tokens": 9688, "train time": "30.9s", "eval time": "9.2s", "train tokens / sec": "6726", "mem allocated": "6811622474", "mem reserved": "15535106425", "elapsed time": "1min 47s"}

accelerator memory max: 19360MB
accelerator memory reserved avg: 14989MB
accelerator memory reserved 99th percentile: 19360MB
train time: 103.75009867198241s
total time: 128.68s

# new
{"step": "  250", "samples": "   1000", "lr": "5.00e-05", "loss avg": "1.1236", "valid acc": "0.340", "gen valid tokens": 10870, "train time": "30.7s", "eval time": "9.6s", "train tokens / sec": "6907", "mem allocated": "6817418766", "mem reserved": "15044934894", "elapsed time": "0min 60s"}

{"step": "  500", "samples": "   2000", "lr": "1.00e-04", "loss avg": "0.7779", "valid acc": "0.380", "gen valid tokens": 9760, "train time": "29.9s", "eval time": "9.5s", "train tokens / sec": "6967", "mem allocated": "6809803481", "mem reserved": "14607997469", "elapsed time": "1min 42s"}

accelerator memory max: 19410MB
accelerator memory reserved avg: 14170MB
accelerator memory reserved 99th percentile: 19410MB
train time: 98.04901859800157s
total time: 120.99s
```

So the results track pretty closely. It would also allow us to remove the caching mechanism, which is always tricky to get right. For memory and runtime, it might be worth checking with larger base models, but I think that materializing the full W' would be more expensive there.

BenjaminBossan (changes requested):
Thank you for providing this PR to add DEFT to PEFT. I was happy to see well chosen comments to explain the code. For the full PR, we're still missing a few pieces, but I focused on the implementation itself for now. In particular, I have a proposal for a different way of implementing the forward pass, please check.

Two more notes:

1. I saw that there is another fine-tuning technique also called DEFT (https://arxiv.org/abs/2310.16776) but I guess it's too late now to do something about the name collision.
2. IIUC, PaRa could be considered a subset of DEFT. I wonder how much effort it would be to add an option to turn DEFT into PaRa and if you would consider that.

MAXNORM8650:
Hi @BenjaminBossan, thanks for the review and for asking for PaRa. Before I push, I want to
flag a **scope change** and get your guidance, since it goes beyond the minimal version we
discussed.

Thanks for the LoRA-style forward suggestion, keeping the base-layer hooks intact is exactly what let DEFT train under device_map sharding at 8B/70B, so it was a real unblock, not just a cleanup.


I extend the results after the cleanup and using the LoRA style forward. DEFT was originally trained for image generation (in the paper). The language model was never tested. This PR also includes DEFT for the language model results.

## Review comments — status

| comment | status | commit |
|---|---|---|
| `_compute_delta` could just be `get_delta_weight` | resolved | `e561ba98` |
| why is float32 casting needed? (add comment) | resolved | `e561ba98` |
| redundant `if cast_input_dtype_enabled` guard | resolved | `e561ba98` |
| dropout called in loop / applied to base weight | resolved | `e561ba98` |
| prefer LoRA-style forward (no `W'` materialization) | resolved | `e561ba98` |
| offload / large memory of cached `delta_weights` | resolved | `fc84e92c` |
| `use_gating` (not requested; non-LoRA knob) | removed | `05f41c96` |

## Tests

- `pytest tests/test_custom_models.py -k deft` → **287 passed, 7 skipped** (peft-dev, CPU).
- ruff + ruff-format + doc-builder line-length: clean.
- pre-commit hook fixed (re-pointed to peft-dev py3.11).

---

# Results

## 1. Official leaderboard — MetaMathQA → GSM8K (llama-3.2-3B, q/v, r32, 5000 steps)

Accuracy + forgetting (wiki-loss increase; lower = better). Same cluster, same attention (sdpa) for all. DEFT uses LoRA-style `alpha` (`alpha/r`); PaRa = removal-only.

| method | GSM8K acc % | forgetting | params |
|---|---|---|---|
| LoRA (α/r=2×) | 47.69 | +0.350 | 9.18M |
| LoRA (α/r=4×) | 48.29 | +0.444 | 9.18M |
| OFT | 49.81 | +0.479 | 8.17M |
| **DEFT-relu (α=4×)** | **49.28** | +0.572 | 9.18M |
| DEFT-relu (α=1×) | 47.23 | **+0.296** | 9.18M |
| DEFT-qr (α=4×) | 48.29 | +0.610 | 9.18M |
| PaRa | 44.58 | +0.825 | 3.67M |


## 2. Multi-scale — GSM8K (TRL-SFT on MetaMathQA, 500 steps, lm_eval flexible-extract %)

Base models 3B / 8B / 70B (learning regime). n=1; 70B on a 300-item subset (±~3 pt). All incl. PaRa.

| model | base | LoRA | OFT | DEFT-qr | DEFT-relu | PaRa |
|---|---|---|---|---|---|---|
| 3B (Llama-3.2-3B) | 27.3 | 33.1 | 33.1 | 30.9 | 29.8 | 28.8 |
| 8B (Llama-3.1-8B) | 50.3 | 54.9 | **56.0** | 53.4 | 53.0 | 49.8 |
| 70B (Llama-3.1-70B) | 79.0 | **85.3** | 79.0 | 80.7 | 76.3 | 80.0 |


## 3. Image generation — FLUX DreamBooth (`method_comparison/image-gen`, n=3 seeds, mean)

Higher CLIP/DINO = better, lower drift = better; best per column **bold**. Incl. PaRa.

| method | CLIP-T | CLIP-I | DINOv1 | DINOv2 | drift |
|---|---|---|---|---|---|
| LoRA | 0.322 | 0.778 | 0.636 | 0.657 | 0.305 |
| OFT | 0.319 | 0.782 | **0.649* | 0.646 | 0.340 |
| **DEFT-qr** | **0.337** | **0.805** | 0.642 | **0.693** | **0.291** |
| DEFT-relu | 0.326 | 0.789 | 0.617 | 0.646 | 0.330 |
| PaRa | **0.337** | **0.805** | 0.629 | 0.677 | 0.296 |

---

src/peft/tuners/deft/__init__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2024-present the HuggingFace Inc. team.
```
BenjaminBossan: Update the date to 2026 here and in the other files.

docs/source/package_reference/deft.md
```diff
@@ -0,0 +1,74 @@
...
+(`W' == W`), so training starts from the pretrained weights and learns the injection. The update is equivalent to a
+low-rank additive delta `Q_P @ (R - right.T @ W)`, which is computed without ever forming the `out x out`
+projection matrix and can be merged into the base weights for inference-free deployment.
+
+DEFT is currently implemented for `torch.nn.Linear` layers. The original implementation and the experiments from the
+paper (Dreambooth, Dreambench Plus, InsDet, VisualCloze, on Stable Diffusion and a unified model) are available at
+[github.com/MAXNORM8650/DEFT](https://github.com/MAXNORM8650/DEFT).
+
+If you use DEFT in your work, please cite the paper:
+
```
BenjaminBossan, line 63: Could you please add a paragraph about PaRa? Just a short one with a link to the paper is enough.

tests/test_custom_models.py
```diff
@@ -7073,3 +7104,76 @@ def test_default_target_modules_osf(self):
         model = get_peft_model(model, config)
         assert model.targeted_module_names == ["lin0", "lin1"]
         assert model.peft_config["default"].target_modules == {"lin0", "lin1"}
+
+
+class TestDeftPaRa:
```
BenjaminBossan: Let's put those two into a separate file, `test_deft.py`.

BenjaminBossan (changes requested):
Thanks for the nice update to the PR, it's in much better shape now. Also thanks for adding the option to use PaRa.

I only have a few small comments, but for the final PR, we're still missing a few things:

1. Let's complete the test suite. This means adding DEFT to `test_config.py`, `test_decoder_models.py` etc. Check the other PEFT methods to see all tests that need extending. This should only require adding to the test matrix. It's fine to leave out PaRa here, as it can't be zero-initialized.
2. Let's add an example to the `examples/` directory. You can check the existing ones for inspiration. This is so that interested users can get started quickly.
3. Let's check in the experiments for the MetaMath and image benchmark. It looks like you've already run them, so just use the best configs from those runs. Please also add experiments for PaRa (but still file them under DEFT).

MAXNORM8650:
Thanks for the thorough review @BenjaminBossan, I've addressed all the points and pushed to deft-integration. 

Please let me know if you have any suggestions for me. I will be happy to do more experiments and editing. 

This fix includes:

- Copyright → 2026 in all DEFT files (71b2c1a9).
- PaRa paragraph + citation added to the docs (5e2c7274).
- Moved the dedicated tests (TestDeftPaRa, TestDeftMerge) into tests/test_deft.py (8c5ef93b).
- Completed the test suite — DEFT was added to the shared matrices (test_config, test_decoder_models, test_encoder_decoder_models, test_feature_extraction_models, test_seq_classifier), matching the FRoD entries; PaRa was left out as it isn't identity-at-init (c8fec9ea). It surfaced that DEFT was nn.Linear-only while the matrix includes GPT2 (Conv1D), so I added Conv1D/fan_in_fan_out support (like LoRA/FRoD) plus a Conv1D test (fee2a361). Full DEFT matrix: 322 passed.
- Examples — added examples/deft_finetuning/ (causal-LM, based on delora_finetuning) and examples/deft_dreambooth/ (text-to-image, based on oft_dreambooth) so users can get started (8d593db3).
- Benchmark configs checked in under experiments/deft/ for both method_comparison harnesses (MetaMathQA + image-gen), with PaRa (95e5aed8).

pytest -k deft + the new matrix entries pass.

BenjaminBossan:
@MAXNORM8650 Thanks for the updates. I haven't checked them yet, but there is a small merge conflict, could you please take care?

MAXNORM8650:
Thanks, @BenjaminBossan. I have resolved it.

BenjaminBossan:
@MAXNORM8650 Thanks for fixing that. Could you please also run `make style`?

MAXNORM8650:
Thanks @BenjaminBossan, done. Sorry for that.

MAXNORM8650:
@BenjaminBossan, there was one conflict with GLoRA, which I have now resolved [23e628b](https://github.com/huggingface/peft/pull/3342/commits/23e628b53dd77df1a7c26823260f12f3621c7ae6).

examples/deft_dreambooth/deft_dreambooth_inference.ipynb
BenjaminBossan, line 1: Let's add a short README.md here to help users get started.

tests/test_deft.py
```diff
@@ -0,0 +1,154 @@
...
+        assert factor.shape == (r, in_features)
+        assert factor.numel() < out_features * in_features
+
+        # merge is correct, and unmerge restores the original weight exactly from the cached factor
+        assert torch.allclose(peft_model(x), unmerged_out, atol=1e-4)
+        peft_model.unmerge_adapter()
+        assert torch.allclose(layer.base_layer.weight, w0, atol=1e-5)
+
+
+class TestDeftConv1D:
```
BenjaminBossan: No need for this test, we can just extend the `test_custom_models.py` test matrix to check `Conv1D`, see e.g. how LoRA does it: https://github.com/huggingface/peft/blob/18cd65453995d5534e75121e42518b4cc7b9b4cc/tests/test_custom_models.py#L158

tests/test_custom_models.py
```diff
@@ -558,6 +559,20 @@
...
+    ("Vanilla MLP 2 DEFT", "MLP", DeftConfig, {"target_modules": ["lin0"]}),
+    ("Vanilla MLP 3 DEFT", "MLP", DeftConfig, {"target_modules": ["lin0", "lin1"]}),
+    ("Vanilla MLP 4 DEFT qr", "MLP", DeftConfig, {"target_modules": ["lin0"], "decomposition_method": "qr"}),
+    ("Vanilla MLP 5 DEFT", "MLP", DeftConfig, {"target_modules": ["lin0"], "modules_to_save": ["lin1"]}),
+    (
+        "Vanilla MLP 7 DEFT alpha",
+        "MLP",
+        DeftConfig,
+        {"target_modules": ["lin0"], "decomposition_method": "qr", "alpha": 16},
+    ),
```
BenjaminBossan, line 575: Could you please add two settings with `para=True`, one with `decomposition_method="relu"` and one with `decomposition_method="qr"`?

src/peft/tuners/deft/layer.py
```diff
@@ -0,0 +1,352 @@
...
+            if self.merged:
+                self.unmerge()
+            result = self.base_layer(x, *args, **kwargs)
+        elif self.merged:
+            result = self.base_layer(x, *args, **kwargs)
+        elif not any(active_adapter in self.deft_P.keys() for active_adapter in self.active_adapters):
+            # no active DEFT adapter on this layer
+            result = self.base_layer(x, *args, **kwargs)
+        else:
+            # LoRA-style forward: instead of materializing W' = W + delta and calling F.linear, apply the update
```
BenjaminBossan: No need to comment on the "LoRA-style" forward, I think it's not generally helpful to understand what's going on.

BenjaminBossan (changes requested):
Thanks for the new updates, we're almost good to go. Tests are passing, I ran the benchmarks locally and results look reasonable. I have a few small comments left, otherwise nothing more to change.

Also, for the commit message and later communication, could you please provide a short, tweet-sized blurb for DEFT? It should answer: What is it, when users should take a look, when it's not a good fit. You can also mention that PaRa is also supported.

MAXNORM8650:
@BenjaminBossan 


Tweet-sized blurb for DEFT:

```DEFT (Decompositional Efficient Fine-Tuning) splits a weight update into two learned low-rank parts: a projection that removes a sub-space of the frozen weight, and a low-rank update that injects new content in its place. Take a look if you want to adapt to new data or concepts — e.g. personalizing a text-to-image model from a few images — while retaining the base model's instruction-following/editability with minimal forgetting. Less of a fit if you don't need to preserve the base model's other capabilities (a plain additive adapter is simpler), or for layers beyond Linear/Conv1D. The PaRa variant is also supported.```

BenjaminBossan (approved):
Thanks for the last push and the blurb, the PR LGTM. Nice work.

PR #3343, opened 2026-06-20
docs: clarify subtract_mutated_init

githubnemo:
I stumbled upon `save_pretrained`'s `path_initial_model_for_weight_conversion` keyword argument and did not understand what it does. I hope that this change helps future readers the same way it helped me.


BenjaminBossan (approved):
Thanks for clarifying what's going on here. LGTM.

PR #3344, opened 2026-06-20
Fix lora_only bias serialization

### LINKED ISSUE #3306: LoraConfig(bias="lora_only") trains base_layer.bias but save_pretrained/get_peft_model_state_dict drops it

loginov-kirill:
I noticed a bug when using `LoraConfig(bias="lora_only")`. Even though the base layer's bias is correctly marked as trainable during training, it gets dropped when calling `get_peft_model_state_dict()` or `save_pretrained()`.

This happens because the bias parameter of the wrapped layer is named `{prefix}.{module_name}.base_layer.bias` in the state dict. However, the serialization code in `get_peft_model_state_dict()` splits the LoRA weight key on `"lora_"` and appends `"bias"`, looking for `{prefix}.{module_name}.bias`. Since it doesn't find this key, it skips the bias completely.

This PR adds a fallback check for `{prefix}.{module_name}.base_layer.bias` in the `bias == "lora_only"` logic so that these trained biases are saved correctly.

Fixes #3306

### DIFF
--- a/src/peft/utils/save_and_load.py
+++ b/src/peft/utils/save_and_load.py
@@ -181,6 +181,10 @@ def get_peft_model_state_dict(
                     bias_name = k.split("lora_")[0] + "bias"
                     if bias_name in state_dict:
                         to_return[bias_name] = state_dict[bias_name]
+                    else:
+                        bias_name = k.split("lora_")[0] + "base_layer.bias"
+                        if bias_name in state_dict:
+                            to_return[bias_name] = state_dict[bias_name]
         else:
             raise NotImplementedError
         to_return = {k: v for k, v in to_return.items() if (("lora_" in k) or ("bias" in k))}


src/peft/utils/save_and_load.py
```diff
@@ -181,6 +181,10 @@ def get_peft_model_state_dict(
                     bias_name = k.split("lora_")[0] + "bias"
                     if bias_name in state_dict:
                         to_return[bias_name] = state_dict[bias_name]
+                    else:
+                        bias_name = k.split("lora_")[0] + "base_layer.bias"
```
BenjaminBossan, line 185: Let's add a comment to explain why this name needs to be checked.

BenjaminBossan (changes requested):
Thanks for tackling that issue. For us to be sure that this fixes the bug and to prevent future regressions, we need to add a unit test though.

I checked if any of the existing tests can be extended, and indeed we have `test_parameters_after_loading_model` here:

https://github.com/huggingface/peft/blob/036abd27e464819e0a19ddaa26093c84d5943488/tests/test_custom_models.py#L2613

Here we could either 1) create a similar test and check LoRA with `bias="lora_only"` or 2) extend the existing `TEST_CASES` (this would need to be for this test only, as other tests won't work with bias enabled).

BenjaminBossan:
Oh, I just saw that we already have a PR for this, #3307.

PR #3345, opened 2026-06-21
Fix resolving regex/string target_modules in MoE config conversion

### LINKED ISSUE #3229: [BUG] peft 0.19 target_modules (str) use `set`

loginov-kirill:
Fix resolving regex/string target_modules in MoE config conversion

Hey,

I ran into this issue while trying to use a regex pattern (like the ones produced by ms-swift) for `target_modules` on a Mixtral / Qwen-Moe type model in PEFT.

Currently, `_convert_peft_config_moe` tries to convert `target_modules` to a set, which ends up splitting the regex string (or `"all-linear"`) into a set of characters. This causes downstream module matching to completely fail with a module not found error.

This PR fixes this by resolving the string `target_modules` against the model's named modules and parameters up front, so that the regular list-based conversion logic runs correctly. I've also added integration tests covering various cases (regex, attention-only, all-linear, save/load idempotency).

Also, I noticed a pytest collection warning on `TestingCommitHashError` during my local testing due to class naming, so I renamed it to `CommitHashTestingError` to clean that up.

Let me know if this looks good! Closes #3229.

### DIFF
--- a/src/peft/utils/transformers_weight_conversion.py
+++ b/src/peft/utils/transformers_weight_conversion.py
@@ -14,6 +14,7 @@
 
 # NOTE: don't import from this module unless transformers v5+ is used
 import copy
+import itertools
 import re
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
This PR is an almost 1:1 copy of #3232. Please instruct your bot to avoid duplicate work.

PR #3346, opened 2026-06-21
Update KaSA implementation

iambogeumkim:
I've reopened the KaSA-related PR after some time away. To start, I applied the last feedback @BenjaminBossan shared a few months ago (you can see the feedback in the image below) and ran both the test command and `make style`. Since it's been a while since I last worked on this, the original repository has changed quite a bit in the meantime, so I expect a few more fixes may still be needed. As I mentioned before, I'll try to move quickly so this can be merged within this month.

This is the last feedback:
image

I'd also like to incorporate @robbiebusinessacc's feedback. Would it make sense to first confirm that the parts I've pushed so far look good before I continue with that work? I'd love to hear your thoughts, @BenjaminBossan. And yes — I'm happy to add him as a co-author! If I'm missing anything or you have any suggestions, they're always welcome.

RELATED PR: https://github.com/huggingface/peft/pull/2698, https://github.com/huggingface/peft/pull/3298


BenjaminBossan:
Unfortunately, it seems like the something went wrong with git when you created your branch. There are a bunch of merge conflicts and the diff shows many unrelated changes. Could you please try to create a clean branch? For this, first ensure that your fork is up-to-date with the latest main branch from PEFT. Then, from a clean workspace, create a new branch, and on that branch, apply your changes. I think that's going to be easier than trying to salvage this branch. Once we have a clean PR, I can take a look and we can discuss on how to proceed.

PR #2710, opened 2025-08-05
FIX Multiple issues with target_parameters

BenjaminBossan:
There are a few issues with target_parameters that are fixed in this PR.

## Existing parametrizations

When using `target_parameters` with LoRA, after the forward call finishes, the LoRA parametrization is removed. However, this also used to remove all other parametrizations on the same parameter, which is bad. With this PR, only the LoRA parametrization is removed.

## Module repr

This PR also extends the `__repr__` of `lora.ParamWrapper` to contain the parameter name, which makes it more useful.

## Multiple LoRA adapters with target_parameters

There is an issue when adding a second LoRA adapter with `target_paramters`, where this second adapter would not actually be applied correctly. The corresponding unit test was too lax to notice the bug. This is not easy to fix, so for now we forbid adding a second adapter with `target_parameters`. This is very strict but it's better than having silent errors.

Although it was possible to fix that specific issue, the solution resulted in ever deeply nested adapters (i.e. with multiple `.base_layer`). This in turn results in those infixes to be part of the `state_dict`. But then we cannot load the individual adapters correctly, except if the model is restored in the exact same order as it was previously created. This is not normally a requirement in PEFT (e.g. I can create a model with two adapters and later decide to load only one of them).

In the long run, we need to think about solutions that would allow this. It may require some form of normalization of the layers to prevent ever deeper nesting. Also, what is ugly right now is that, given that the LoRA lives on a module but actually targets one of possibly multiple parameter, the LoRA weights don't actually reference said parameter in any name. That means, purely from the `state_dict`, it is unclear which parameter a LoRA weight belongs to. Ideally, this should be encoded in the LoRA weight key.

PR #3350, opened 2026-06-22
ENH Allow multiple adapters when using target_parameters

### LINKED ISSUE #3340: Only one `target_parameters` adapter per model
Reference/tracking issue for an intentional limitation: PEFT rejects a second LoRA adapter when the config uses `target_parameters`:

```
ValueError: ... only one LoRA adapter per model with `target_parameters` is allowed.
```

Introduced in #2710 to avoid `.base_layer` nesting / state-dict / load-order issues. The inline comment reads as temporary ("Right now, unfortunately..."), but there's no open issue, so opening this to centralize discussion.

### Why it matters

`target_parameters` is how you LoRA **fused MoE experts** (Transformers 5.x stacks experts into single `nn.Parameter` tensors, so `target_modules` can't match them). This blocks any second adapter on such models: e.g. a frozen **reference** adapter for DPO/GRPO/RLOO (see [huggingface/trl#5222](https://github.com/huggingface/trl/issues/5222)). Downstream workaround is `disable_adapter()` for the reference pass, which is only equivalent for an untrained adapter.

### MRE

```python

model = AutoModelForCausalLM.from_pretrained("trl-internal-testing/tiny-GptOssForCausalLM")
config = LoraConfig(target_parameters=["mlp.experts.gate_up_proj", "mlp.experts.down_proj"])
model = get_peft_model(model, config)                   # "default" adapter, OK
model.add_adapter("ref", model.peft_config["default"])  # <-- ValueError
```

<details>
<summary>Full traceback</summary>

```
  File "mre.py", line 12, in <module>
    model.add_adapter("ref", model.peft_config["default"])   # <-- raises ValueError
  File ".../peft/peft_model.py", line 1085, in add_adapter
    self.base_model.inject_adapter(
        self.base_model.model, adapter_name, low_cpu_mem_usage=low_cpu_mem_usage
    )
  File ".../peft/tuners/tuners_utils.py", line 945, in inject_adapter
    self._inject_parameters(
        peft_config=peft_config, model=model, adapter_name=adapter_name, low_cpu_mem_usage=low_cpu_mem_usage
    )
  File ".../peft/tuners/tuners_utils.py", line 1140, in _inject_parameters
    create_and_replace_param(module_name, key, param_name)
  File ".../peft/tuners/tuners_utils.py", line 1101, in create_and_replace_param
    self._create_and_replace(
        peft_config,
        ...
        parameter_name=param_name.rpartition(".")[-1],
    )
  File ".../peft/tuners/lora/model.py", line 198, in _create_and_replace
    raise ValueError(...)
ValueError: Adding a LoRA config with `target_parameters=['mlp.experts.gate_up_proj', 'mlp.experts.down_proj']` but there are already other LoRA adapters on this model that use `target_parameters`. At the moment, only one LoRA adapter per model with `target_parameters` is allowed.
```

</details>

### Questions

- Is lifting this on the roadmap, or is single-adapter the intended long-term behavior?
- If it stays, worth documenting + pointing the error at the `disable_adapter()` workaround?

### Env
`peft==0.19.1`, `transformers==5.12.1`, `torch==2.10.0+cu128`

BenjaminBossan:
Resolves #3340

## Context

So far, we did not allow adding multiple LoRA adapters with `target_parameters` on the same layer. This was a known limitation. I have already attempted to solve this once (see #2710) but didn't have time to come up with a nice solution. As it was unclear if there was any real world need to support this, there was no further work on it since then. Now we know that there are practical application that may need it, so I resumed the work.

## What doesn't work

The previous solution attempted to solve the issue by nesting the `lora.ParamWrapper`s. So for adapters `'default'` and `'other'`, we would end up having something like:

`param_wrapper_default(param_wrapper_other(base_layer))`

This was problematic. Not only could this result in very deep nesting, which is inefficient. What's worse is that `state_dict` key for `'other'` would contain `'base_layer.'` as an infix. Therefore, if we wanted to load the `'other'` adapter _without_ first loading the `'default'` adapter, we would get a key mismatch.

We could also not simply strip out `'base_layer.'` infix because we use nesting to deal with multiple `nn.Parameter`s on the same module; to account for that, we need to keep the infixes.

## Solution

The solution is pretty straightforward: We use the existing mechanism to store the parameters for the other adapter in the `nn.ModuleDict`. For this, we detect if the layer is already a `ParamWrapper` when adding the second adapter and update that layer instead of nesting it.

## Caveat

Generally with MoE modules in Transformes, we have multiple `nn.Parameter`s on the same module, e.g. `gate_up_proj` and `down_proj`. Therefore, we often want to target multiple parameters on the same target module. For LoRA, we usually only have one targeted parameter per LoRA layer. To allow targeting multiple `nn.Parameter`s, we thus create multple LoRA layers, one per `nn.Parameter`, and nest them.  The simple approach from this PR supports this as long as the different adapters target exactly the same `nn.Parameter`s.

This solution does, however, not work with multiple adapters that target a different sets of parameters. This is because the information which parameter is targeted is not stored in the `state_dict` itself -- remember: When there is more than 1 parameter per layer, we solve this via nesting. So we rely on the nesting level 0 corresponding to parameter 0, nesting 1 to parameter 1, etc. Therefore, if we had different adapters targeting different parameters, we would lose that mapping 1:1 and would not be able to tell which parameter is meant to be targeted when loading a checkpoint. (I can't really remember why I didn't try the solution from this PR back then, but it could be for this exact limitation.)

If we wanted to allow targeting different parameters for each adapter, we would need larger changes. First, we would need to store the targeted parameter name in a dict, separately for each adapter name. Then, during serialization, we would need to mangle the parameter name into the `state_dict` keys to be able to restore the correct mapping. Moreover, since this changes the structure of the `state_dict`, we would need special logic to still support existing `state_dict`s that don't follow that logic. Overall, this would be a big change with unclear benefit. Thus, once again, I opt to postpone a full solution.


src/peft/tuners/lora/model.py
```diff
@@ -189,18 +213,6 @@ def _create_and_replace(
         if current_key is None:
             raise ValueError("Current Key shouldn't be `None`")
 
-        if lora_config.target_parameters:
```
BenjaminBossan, line 192: Note: This check for the compatibility of the new config should have arguably been inside of `_check_new_adapter_config`. Therefore, the new checking code uses `_check_new_adapter_config` instead of running inside of `_create_and_replace`.

githubnemo (approved):
LGTM!

PR #3352, opened 2026-06-22
feat: extend MULTIPLE_ACTIVE_ADAPTERS_TEST_CASES with LoHa, LoKr, OFT…

peft-jambot:
…, and AdaMSS

Add test cases for multiple active adapters for PEFT methods that support it but were missing from the test matrix:

- LoHa: Same + Different variants
- LoKr: Same + Different variants
- OFT: Same variant only (Different variant has numerical precision issues in test_merge_layers_multi due to small model dimensions)
- AdaMSS: Same + Different variants (Different uses r=2 for lin1 due to small output dimension)

Methods investigated but excluded:
- C3A: Does not support multiple active adapters (forward pass fails with
  >1 active adapter due to tensor shape mismatches in block circular convolution)
- OSF: Only uses first active adapter, silently ignores additional ones
- Poly: Router-based method, incompatible with additive multi-adapter test pattern
- XLoRA: Classifier-driven scaling, incompatible with standard test pattern
- Cartridge: Prompt-learning method, not adapter-based

Also add AdaMSS to _skip_if_merging_not_supported since it doesn't support adapter merging.

Note: GPU tests are not required for these changes as per the issue.


githubnemo:
Superseded by #3361. Thanks though, you helped detect an issue with the test with your comment.

PR #3353, opened 2026-06-22
fix: increase FRoD init_weights=False std to prevent flaky test

peft-jambot:
The test_merge_layers_multi[FRoD Same] test was flaky (~1/20 failure rate). The root cause is that FRoD's init_weights=False code path used std=0.05 for random perturbations. With small model dimensions (the test MLP has lin0: 10→20), two adapters initialized with std=0.05 could occasionally produce outputs within 1e-3 of each other, causing the assertion
to flip.

Increase std from 0.05 to 0.1, matching BOFT's convention for the same code path. This only affects the init_weights=False test path; production usage with init_weights=True (the default) zeros out parameters and is unaffected.

Tested: 30/30 runs of the previously flaky test pass. All 239 FRoD tests continue to pass.


src/peft/tuners/frod/layer.py
```diff
@@ -102,9 +102,11 @@ def update_layer(
             self.reset_frod_parameters(adapter_name)
         else:
             # PEFT convention: init_weights=False should produce a non-identity adapter for merge tests.
+            # Using std=0.1 (matching BOFT's convention) ensures the perturbation is large enough that
+            # test assertions checking distinct adapter outputs remain stable across RNG draws.
```
BenjaminBossan: Please remove this extra comment, it's not necessary.

BenjaminBossan:
Example of a failed run: https://github.com/huggingface/peft/actions/runs/27963170021/job/82767795942#step:8:1446

@Bane-Elvin What do you think of this change?

Bane-Elvin:
That makes sense to me. I checked the FRoD initialization path, and this only affects the non-default `init_weights=False` path used by these merge tests. The default `init_weights=True` path still zero-initializes the trainable FRoD parameters, so normal usage is unaffected.

The failed assertion also matches the explanation: the test expects two randomly initialized adapters to produce outputs that are not `allclose` at `1e-3`, and with tiny MLP dimensions `std=0.05` can occasionally be too small. Increasing it to `0.1`, consistent with BOFT’s convention, looks like a reasonable fix to me.

githubnemo (approved):

PR #3360, opened 2026-06-25
feat: support LoRA conversion for multiplicative PEFT methods (OFT, BOFT)

peft-jambot:
## Description

Add support for converting multiplicative PEFT methods (OFT, BOFT) to LoRA via `convert_to_lora`. Previously, only additive methods (where `W' = W + ΔW`) could be converted, because `convert_to_lora` used `get_delta_weight()` to obtain the delta and SVD-approximate it. Multiplicative methods like OFT (`W' = W @ Rᵀ`) and BOFT (`W' = (W @ Rᵀ) * s`) return a transformation matrix from `get_delta_weight()`, not an additive delta.

This PR introduces a `get_additive_delta()` method on `BaseTunerLayer` that returns the actual additive delta `W' - W` for any PEFT method. The public method handles the shared contract (merged-state check) and delegates the per-method computation to a `_get_additive_delta()` override point. The dtype of the returned delta is determined by each `_get_additive_delta()` implementation, so that it matches the dtype of the adapter weights rather than the base weight. By default, the latter raises `NotImplementedError`. Every PEFT method that supports LoRA conversion overrides it: additive methods return `self.get_delta_weight(adapter_name)`, while multiplicative methods construct the effective weight `W'` and subtract the base weight `W`.

Fork issue: https://github.com/peft-jambot/peft/issues/4

### Changes

- **`src/peft/tuners/tuners_utils.py`**: Added `get_additive_delta()` to `BaseTunerLayer` as a template method that centralizes the shared contract, with per-method logic in a new `_get_additive_delta()` override point:
  - Calling `get_additive_delta()` while any adapter is merged raises a `ValueError`, as merging modifies the base weight, which would make the return value depend on the currently merged adapters.
  - The returned tensor has the dtype of the adapter layer weights (e.g. `lora_A` for LoRA), not the dtype of the base weight. This is the dtype that the converted LoRA adapter will use.
  - By default, `_get_additive_delta()` raises `NotImplementedError` with a hint that additive methods should override it to return `get_delta_weight()`.
- **`src/peft/tuners/oft/layer.py`**: Override for `OFT.Linear` — computes `W @ Rᵀ - W` in float32, where `R` is the orthogonal rotation matrix; uses `get_base_weight()` so quantized weights are dequantized correctly, and casts the result to the dtype of the adapter weights. Enables `supports_lora_conversion() = True`.
- **`src/peft/tuners/boft/layer.py`**: Override for `BOFT.Linear` — computes `(W @ Rᵀ) * s - W` in float32, where `R` is the rotation matrix and `s` the scaling vector; enables `supports_lora_conversion() = True`.
- **`src/peft/tuners/miss/layer.py`**: Override for `MissLinear` that returns the real additive delta in all modes instead of raising for the non-`bat` modes. For `bat`, `get_delta_weight` already computes the additive delta; for `standard` and `mini`, `get_delta_weight_miss` returns the full merged weight, so the base weight is subtracted from it (with a clone, as that function modifies its input in place).
- **`src/peft/tuners/lora/conversion.py`**: `_convert_module_to_lora()` now calls `get_additive_delta()` instead of `get_delta_weight()`. The MiSS `bat` mode was moved out of `_convert_miss_module_to_lora` into `MissLinear._get_additive_delta()` (its `get_delta_weight` already returns the additive delta), so the special case only handles the exact standard/mini factorization and `bat` goes through the generic SVD path. Updated docstrings of `convert_to_lora` and `save_as_lora` and the error messages accordingly.
- Additive tuners that support LoRA conversion implement `_get_additive_delta()` by delegating to `get_delta_weight()`: LoRA, LoHa, LoKr, Vera, DeLoRA, RandLora, VBLoRA, C3A, Psoft, GraLoRA, FourierFT, WaveFT, TinyLoRA, UniLoRA. The UniLoRA addition was needed because upstream UniLoRA declares `supports_lora_conversion = True` but was missing the new method.
- **`src/peft/tuners/hira/layer.py`**, **`src/peft/tuners/shira/layer.py`**: Removed the `supports_lora_conversion()` overrides that only returned `False`, as that is the `BaseTunerLayer` default.
- **`tests/testing_common.py`**: 
  - `_test_lora_conversion` now uses `pytest.skip` with an explanatory message instead of a silent early return, so it is visible which methods were tested and which were skipped.
  - New `_test_get_additive_delta_corresponds_to_merged_weight` helper: verifies that `base_weight + get_additive_delta()` reproduces the merged weight for all conversion-capable tuners.
- **`tests/test_custom_models.py`**, **`tests/test_decoder_models.py`**: New parametrized test `test_get_additive_delta_corresponds_to_merged_weight` using the helper above, running in fp32/bf16/fp16. BD-LoRA is marked as xfail (`get_delta_weight` does not unpack the block-diagonal factors, so the delta does not correspond to the merged weight), and `rank_dropout` is disabled for LoHa/LoKr, as it makes `get_delta_weight` non-deterministic (being addressed in #3589) and is not essential for this test.

### How I tested

- `pytest tests/test_lora_conversion.py` → 32 passed
- `pytest tests/test_custom_models.py -k "test_get_additive_delta_corresponds_to_merged_weight"` → 429 passed, 9 xfailed (BD-LoRA), 555 skipped
- New `test_get_additive_delta_uses_adapter_dtype` verifies that `get_additive_delta` returns the adapter dtype (fp32) rather than the base weight dtype (bf16) when the two differ.
- `pytest tests/test_decoder_models.py -k "test_get_additive_delta_corresponds_to_merged_weight"` → 168 passed, 552 skipped (the skips are methods without LoRA conversion support, tuners that don't support GPT2's Conv1D layers, and pre-existing GPT2 config skips). OFT and BOFT run and pass on the OPT/GPT-J/Llama/Qwen/Gemma tiny models.
- Merge regression checks with the new merged-state `ValueError` in place: `pytest tests/test_custom_models.py -k "merge and not get_additive_delta"` → 2164 passed; `pytest tests/test_decoder_models.py -k "merge and not corresponds"` → 391 passed
- `pytest tests/test_custom_models.py -k "OFT or Boft or Miss"` → 2639 passed, 128 skipped; `pytest tests/test_boft.py` → 2 passed
- Additional numerical check: `base_weight + layer.get_additive_delta()` exactly reproduces the weight produced by each method's own `merge()` for OFT and BOFT (max abs error 0.0)
- `ruff check` + `ruff format --check` (ruff 0.16.4) → all checks passed

Note on the environment: torchao 0.18.0 breaks an import in `src/peft/utils/quantization_utils.py`, which produced 6 errors in the MiSS conversion tests locally. This is pre-existing on main and unrelated to this PR (tracked in #3571); with torchao 0.17.0 installed, all 32 tests pass.

The latest `main` was merged into this branch (no conflicts).

## AI assistance statement

This PR was prepared with AI assistance (Hermes Agent). The human submitter reviewed every changed line and ran the relevant tests.


src/peft/tuners/tuners_utils.py
```diff
@@ -1781,6 +1781,22 @@ def supports_lora_conversion(self, adapter_name: str = "default") -> bool:
         """
         return False
 
+    def get_additive_delta(self, adapter_name: str = "default") -> torch.Tensor:
+        """
+        Return the additive delta weight W' - W for the given adapter.
+
+        For additive PEFT methods (e.g. LoRA, LoKr, LoHa), this is equivalent to ``get_delta_weight``. For
```
BenjaminBossan: Please always use single backticks in all docstrings.

src/peft/tuners/tuners_utils.py
```diff
@@ -1781,6 +1781,22 @@ def supports_lora_conversion(self, adapter_name: str = "default") -> bool:
         """
         return False
 
+    def get_additive_delta(self, adapter_name: str = "default") -> torch.Tensor:
```
BenjaminBossan, line 2222: One issue I see here is that we assume by default that `self.get_delta_weight(adapter_name)` does the right thing. However, for many PEFT methods, that's not the case. WDYT about raise a `NotImplementedError` here and then, for each PEFT method that is indeed additive, override the method to return `self.get_delta_weight(adapter_name)`?

tests/test_lora_conversion.py
```diff
@@ -691,3 +693,135 @@ def test_miss_save_as_lora(self, miss_model_standard, tmp_path):
...
+        )
+        return get_peft_model(self.get_base_model(), config).eval()
+
+    def test_oft_supports_lora_conversion(self, oft_model):
+        assert oft_model.supports_lora_conversion()
+
+    def test_boft_supports_lora_conversion(self, boft_model):
+        assert boft_model.supports_lora_conversion()
+
+    def test_oft_converted_lora_approximates_original(self, oft_model):
```
BenjaminBossan: Isn't this already covered by `_test_lora_conversion` in `testing_common.py`?

tests/test_lora_conversion.py
```diff
@@ -691,3 +693,135 @@ def test_miss_save_as_lora(self, miss_model_standard, tmp_path):
...
+        load_result = set_peft_model_state_dict(lora_model, state_dict)
+        assert not load_result.unexpected_keys
+
+        with torch.inference_mode():
+            output_converted = lora_model(inputs, output_hidden_states=True)
+
+        mse_converted = self.get_mse(output_converted, output_oft)
+        assert 0.0 < mse_converted < 0.1, f"OFT conversion MSE too high: {mse_converted}"
+
+    def test_boft_converted_lora_approximates_original(self, boft_model):
```
BenjaminBossan: Isn't this already covered by `_test_lora_conversion` in `testing_common.py`?

tests/test_lora_conversion.py
```diff
@@ -691,3 +693,135 @@ def test_miss_save_as_lora(self, miss_model_standard, tmp_path):
...
+        load_result = set_peft_model_state_dict(lora_model, state_dict)
+        assert not load_result.unexpected_keys
+
+        with torch.inference_mode():
+            output_converted = lora_model(inputs, output_hidden_states=True)
+
+        mse_converted = self.get_mse(output_converted, output_boft)
+        assert 0.0 < mse_converted < 0.1, f"BOFT conversion MSE too high: {mse_converted}"
+
+    def test_oft_targeted_modules_identical(self, oft_model):
```
BenjaminBossan: This test is unnecessary.

tests/test_lora_conversion.py
```diff
@@ -691,3 +693,135 @@ def test_miss_save_as_lora(self, miss_model_standard, tmp_path):
...
+        _, lora_state_dict = convert_to_lora(oft_model, rank=8)
+        oft_state_dict = oft_model.state_dict()
+
+        modules_oft = {k.rsplit(".", 3)[0] for k in oft_state_dict.keys() if ".oft_R" in k}
+        modules_lora = {k.rsplit(".", 2)[0] for k in lora_state_dict.keys() if ".lora" in k}
+        assert modules_oft == modules_lora
+
+    def test_boft_targeted_modules_identical(self, boft_model):
```
BenjaminBossan: This test is unnecessary.

tests/test_lora_conversion.py
```diff
@@ -691,3 +693,135 @@ def test_miss_save_as_lora(self, miss_model_standard, tmp_path):
...
+        _, lora_state_dict = convert_to_lora(boft_model, rank=8)
+        boft_state_dict = boft_model.state_dict()
+
+        modules_boft = {k.rsplit(".", 2)[0] for k in boft_state_dict.keys() if ".boft_R" in k}
+        modules_lora = {k.rsplit(".", 2)[0] for k in lora_state_dict.keys() if ".lora" in k}
+        assert modules_boft == modules_lora
+
+    def test_oft_save_as_lora(self, oft_model, tmp_path):
```
BenjaminBossan: This test is unnecessary.

BenjaminBossan (changes requested):
Thanks for the PR. I've added some comments, please check.

src/peft/tuners/tuners_utils.py
peft-jambot: Addressed in commit 176f5982.

src/peft/tuners/tuners_utils.py
peft-jambot: Addressed in commit 176f5982.

tests/test_lora_conversion.py
peft-jambot: Addressed in commit 176f5982.

tests/test_lora_conversion.py
peft-jambot: Addressed in commit 176f5982.

tests/test_lora_conversion.py
peft-jambot: Addressed in commit 176f5982.

tests/test_lora_conversion.py
peft-jambot: Addressed in commit 176f5982.

tests/test_lora_conversion.py
peft-jambot: Addressed in commit 176f5982.

tests/test_lora_conversion.py
```diff
@@ -691,3 +693,48 @@ def test_miss_save_as_lora(self, miss_model_standard, tmp_path):
...
+            boft_block_size=4,
+            target_modules=["q_proj", "v_proj"],
+        )
+        return get_peft_model(self.get_base_model(), config).eval()
+
+    def test_oft_supports_lora_conversion(self, oft_model):
+        assert oft_model.supports_lora_conversion()
+
+    def test_boft_supports_lora_conversion(self, boft_model):
+        assert boft_model.supports_lora_conversion()
```
BenjaminBossan: These tests aren't really helpful. Let's delete them.

tests/test_lora_conversion.py
BenjaminBossan, line 1: In `testing_common.py::_test_lora_conversion`, we currently early return for PEFT methods that don't support LoRA conversion. That makes it hard to see which methods were actually tested and which ones just early returned. Please change this to use `pytest.skip` with an appropriate message instead.

BenjaminBossan (changes requested):
Thanks for the updates. There is still a bit of work to do, please check my inline comments and the discussion below.

As a bit of a clean up, could you please remove the `supports_lora_conversion` methods from HiRA and SHiRA, as they just return `False` and are thus unnecessary.

Technically speaking, MiSS doesn't require `get_additive_delta` because its conversion is special cased. But I think it would still be good to implement it. IIUC, when initializing MiSS with "bat", the `delta_weight` is already the additive `delta_weight`, check the code in `conversion.py::_convert_miss_module_to_lora`. So this case can be moved to `get_additive_delta` and then `_convert_miss_module_to_lora` only needs to be called when not using "bat", otherwise we just use the normal conversion path.

For non-bat, we still need the special code from `_convert_miss_module_to_lora`, but it can be simplified once the "bat" case has been factored out.

tests/test_lora_conversion.py
peft-jambot: Deleted these tests in commit 6a4247ad. The TestMultiplicativeLoraConversion class is now empty and has been removed as well.

tests/test_lora_conversion.py
peft-jambot: Changed to use pytest.skip with a message like 'XxxConfig does not support LoRA conversion' in commit 6a4247ad. Unsupported methods now show up as SKIPPED instead of silently passing.

src/peft/tuners/miss/layer.py
```diff
@@ -374,6 +374,21 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
...
 
+    def get_additive_delta(self, adapter_name: str = "default") -> torch.Tensor:
+        """Return the additive delta weight W' - W for the given adapter.
+
+        For `bat` mode, `get_delta_weight` already computes the additive delta (used in forward as `orig_weight +
+        delta_weight`), so we delegate to it. For `standard` and `mini` modes, the conversion is handled by the
+        special-cased `_convert_miss_module_to_lora` and does not go through this method.
+        """
+        if self.miss_fn != "bat":
+            raise NotImplementedError(
```
BenjaminBossan: Instead of raising an error here, I think we should actually return the additive delta, even if it's not used for LoRA conversion. The reason is that `get_delta_weight` could also be used in different contexts, and there is no reason to raise an error there if we can provide the actual result.

BenjaminBossan (changes requested):
Thanks for the updates. Generally, this LGTM, but I have a suggestion and some questions:

Generally, we should expect that the return value from `module.get_additive_delta` should correspond to the difference between the merged and unmerged weight, right? So I wanted to add a test like so to `testing_common.py` (after `_test_safe_merge`):

```python
    def _test_get_additive_delta_corresponds_to_merged_weight(self, model_id, config_cls, config_kwargs, dtype):
        # This test checks that the get_additive_weight method really returns the weight delta between the base weight
        # and the mered weight. For many PEFT methods, this is trivial, as get_additive_weight == get_delta_weight, but
        # for some they differ.
        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
        if (config_cls == MissConfig) and (config_kwargs.get("init_weights") == "bat"):
            pytest.skip(reason="Test requires non-zero init but MiSS is using 'bat' init")
        torch.manual_seed(0)
        atol = rtol = {torch.float32: 1e-5, torch.bfloat16: 1e-2, torch.float16: 2e-3}[dtype]

            model = self.transformers_class.from_pretrained(model_id, dtype=dtype)
            config = config_cls(
                **config_kwargs,
            )
            state_dict_before = {k: v.clone() for k, v in model.state_dict().items()}

            # don't autocast adapter dtype so that we can test lower precision floats
            peft_model = get_peft_model(model, config, autocast_adapter_dtype=False)
            if not peft_model.supports_lora_conversion():
                pytest.skip(f"{config.peft_type} does not support LoRA conversion.")

            peft_model.merge_adapter()
            # take the state_dict from model and not peft_model so that we don't have to strip the PEFT prefix
            state_dict_after = model.state_dict()

            no_tuner_layer_found = True
                if not isinstance(module, BaseTunerLayer):
                    continue

                key_before = f"{name}.weight"
                key_after = f"{name}.base_layer.weight"
                weight_before = state_dict_before.get(key_before)
                weight_after = state_dict_after.get(key_after)
                if (weight_before is None) or (weight_after is None):
                    continue

                no_tuner_layer_found = False
                delta_weight_expected = weight_after - weight_before
                additive_delta = module.get_additive_delta().to(delta_weight_expected.dtype)
                assert torch.allclose(additive_delta, delta_weight_expected, atol=atol, rtol=rtol)

            # sanity check
            if no_tuner_layer_found:
                raise ValueError("This test could not identify any tuner layers")
```

The test can be wired into `test_custom_models.py` like this:

```python
    @pytest.mark.parametrize("dtype", [torch.float32, torch.bfloat16, torch.float16], ids=["fp32", "bf16", "fp16"])
    def test_get_additive_delta_corresponds_to_merged_weight(self, test_name, model_id, config_cls, config_kwargs, dtype):
        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)

        config_kwargs = set_init_weights_false(config_cls, config_kwargs)
        self._test_get_additive_delta_corresponds_to_merged_weight(model_id, config_cls, config_kwargs, dtype=dtype)
```

Perhaps we can also add it to the other test classes like `test_decoder_models.py`, I haven't thought about that.

Anyway, if I'm not missing something, the test should either pass or be skipped if the PEFT method doesn't support it. However, it has a couple of failures:

1. LoHa and LoKr: Those tests fail because of a bug in those methods, but that bug is being addressed in #3589, so we can accept those failures for the time being.
2. MiSS with 'bat': Because of the `NotImplementedError` being raised. I added an inline comment that we should avoid that.
3. BD-LoRA: Raises a shape mismatch error when trying to merge, not sure why, maybe misconfigured?

Could you please add the test, adjust it if necessary, and add expected failures where failures are expected?

I'm also wondering about these questions, which we so far left unspecified:

1. What should happen if weights are already merged when `get_additive_delta` is called? I believe we should possibly raise an error there, otherwise the return value depends on what is currently merged. Note that #3590 is already adding checks to the hotswap functions themselves to prevent this, but I think it still makes sense to add the checks to `get_additive_delta` as well.
2. We haven't specified the expected dtype: Should it be the dtype of the PEFT adapter, the dtype of the base layer, or something else? For VB-LoRA and RandLoRA running on CPU, the dtypes actually differ, which is why the test calls `to(delta_weight_expected.dtype)`. But it might be worth it to harmonize the output dtype.

If you make changes there, also update the docstring of `BaseTunerLayer.get_additive_delta`.

src/peft/tuners/miss/layer.py
peft-jambot: Addressed in 158b4ac5: MissLinear now overrides _get_additive_delta to return the real additive delta in all modes. For bat, get_delta_weight already returns the additive delta; for standard/mini, get_delta_weight_miss returns the full merged weight (and modifies its input in place), so the base weight is subtracted from a clone.

tests/testing_common.py
```diff
@@ -859,6 +859,67 @@ def _test_safe_merge(self, model_id, config_cls, config_kwargs):
                 # serializing with safetensors works
                 save_file(model_unloaded.state_dict(), os.path.join(tmp_dirname, "model.safetensors"))
 
+    def _test_get_additive_delta_corresponds_to_merged_weight(self, model_id, config_cls, config_kwargs, dtype):
+        # This test checks that the get_additive_delta method really returns the weight delta between the base weight
+        # and the merged weight. For many PEFT methods, this is trivial, as _get_additive_delta ==
+        # get_delta_weight, but for some they differ, e.g. for multiplicative methods like OFT.
+        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
```
BenjaminBossan: Redundant skip, as the callers already check this.

tests/testing_common.py
```diff
@@ -859,6 +859,67 @@ def _test_safe_merge(self, model_id, config_cls, config_kwargs):
...
+        if (config_cls == MissConfig) and (config_kwargs.get("init_weights") == "bat"):
+            pytest.skip(reason="Test requires non-zero init but MiSS is using 'bat' init")
```
BenjaminBossan, line 921: Let's add an empty line here.

tests/test_custom_models.py
```diff
@@ -2583,6 +2583,33 @@ def test_safe_merge(self, test_name, model_id, config_cls, config_kwargs):
...
+        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
+        if (config_cls == LoraConfig) and config_kwargs.get("use_bdlora"):
+            # BD-LoRA: get_delta_weight does not unpack the block-diagonal factors, so the delta does not correspond
+            # to the merged weight
+            pytest.xfail("BD-LoRA delta weight computation does not handle block-diagonal factors")
+
+        config_kwargs = set_init_weights_false(config_cls, config_kwargs)
+        self._test_get_additive_delta_corresponds_to_merged_weight(model_id, config_cls, config_kwargs, dtype=dtype)
+
+    def test_get_additive_delta_uses_adapter_dtype(self):
```
BenjaminBossan: Would it be possible to check all PEFT methods, not just LoRA? We may use `ALL_PEFT_CONFIG_CLASSES` to avoid running the same PEFT method multiple times.

tests/test_custom_models.py
```diff
@@ -2583,6 +2583,33 @@ def test_safe_merge(self, test_name, model_id, config_cls, config_kwargs):
...
+    @pytest.mark.parametrize("dtype", [torch.float32, torch.bfloat16, torch.float16], ids=["fp32", "bf16", "fp16"])
+    @pytest.mark.parametrize("test_name, model_id, config_cls, config_kwargs", TEST_CASES)
+    def test_get_additive_delta_corresponds_to_merged_weight(
+        self, test_name, model_id, config_cls, config_kwargs, dtype
+    ):
+        _skip_if_merging_not_supported(model_id, config_cls, config_kwargs)
+        if (config_cls == LoraConfig) and config_kwargs.get("use_bdlora"):
+            # BD-LoRA: get_delta_weight does not unpack the block-diagonal factors, so the delta does not correspond
+            # to the merged weight
+            pytest.xfail("BD-LoRA delta weight computation does not handle block-diagonal factors")
```
BenjaminBossan: Let's skip instead of xfail.

BenjaminBossan (changes requested):
Thanks for the update. I still have a few smaller comments about the tests.

tests/testing_common.py
peft-jambot: Addressed: removed the redundant `_skip_if_merging_not_supported` call from `_test_get_additive_delta_corresponds_to_merged_weight` in testing_common.py — both callers (test_custom_models.py and test_decoder_models.py) already check it before calling the helper.

tests/testing_common.py
peft-jambot: Addressed: added the missing empty line in testing_common.py.

tests/test_custom_models.py
peft-jambot: Addressed: `test_get_additive_delta_uses_adapter_dtype` is now parametrized over `ALL_PEFT_CONFIG_CLASSES` instead of being LoRA-only. This surfaced that C3A and Psoft's `get_delta_weight` cast to the base weight dtype, so their additive delta was bf16 despite fp32 adapters — I fixed both `_get_additive_delta` to cast to the adapter dtype. All 17 conversion-capable methods now return fp32, 18 non-conversion methods are skipped with a message.

tests/test_custom_models.py
peft-jambot: Addressed: changed BD-LoRA from xfail to `pytest.skip`.

PR #3361, opened 2026-06-25
TST Extend multiple active adapter tests

BenjaminBossan:
Supersedes #3352

The tests for multiple active adapters were missing a couple of PEFT methods. This PR adds LoHa, LoKr, OFT, C3A, and AdaMSS.

Note that C3A targeting the same layer fails with a shape error, I didn't investigate further and just skipped it.

While reviewing #3352, I noticed that one of the test functions was suboptimal: It only compared the first sample of the outputs, which means only two scalars were compared, leading to spurious failures. Morevoer, the outputs were called logits even though they are not logits. This is most likely a copy-paste error from a different test that used language modeling. This PR cleans the test up.


githubnemo (approved):
Nice ✨

PR #3362, opened 2026-06-25
TST Add dedicated BD-LoRA math-correctness test suite (follow-up to #2895)

eSVeeF:
This PR adds a set of BD-LoRA tests, `tests/test_bdlora.py`, that complement the generic test matrix coverage introduced in PR #2895. These tests fill gaps not covered by the test matrix in `test_custom_models.py` by validating the core block-diagonal mathematics that define the method.

In #2895 review, the [suggestion](https://github.com/huggingface/peft/pull/2895#discussion_r2534737543) was to extend the existing test matrix rather than pile on BD-LoRA-specific tests. But they don't really prove many of BD-LoRA's properties that right now could be broken silently.

#### What These Tests Prove

**1. `test_block_diagonal_linear_forward_matches_dense_equivalent`**
   - Validates the core `BlockDiagonalLinear` forward implementation against explicit 
     dense reconstruction for several input shapes and block counts
   - Catches bugs in the reshape/einsum math that round‑trip tests won't catch

**2. `test_bdlora_merge_delta_matches_manual_dense_equivalent`**
   - Verifies that the merge path produces the exact expected delta for both A-block and B-block

**3. `test_bdlora_nblocks_one_matches_vanilla_lora`**
   - Special case: with one block, BD-LoRA must reduce exactly to vanilla LoRA

**4. `test_bdlora_packed_shapes_and_adapter_param_counts_vs_vanilla`**
   - Verifies packed storage actually gives the expected parameter reduction and memory savings versus vanilla LoRA
   - Guards against regressions that don't accidentally build dense (non‑blocked) weight matrices

### DIFF
--- a/tests/test_bdlora.py
+++ b/tests/test_bdlora.py
@@ -1,0 +1,228 @@
+# Copyright 2026-present the HuggingFace Inc. team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from copy import deepcopy
+
+import pytest
+import torch
+
+from peft import LoraConfig, get_peft_model
+from peft.tuners.lora.config import BdLoraConfig
+from peft.tuners.lora.variants import BlockDiagonalLinear
+
+
+def _fill_deterministic_weight(weight: torch.Tensor) -> None:
+    values = torch.arange(weight.numel(), dtype=weight.dtype, device=weight.device).reshape_as(weight)
+    weight.data.copy_(values)
+
+
+def _dense_weight(module: torch.nn.Module) -> torch.Tensor:
+    if isinstance(module, BlockDiagonalLinear):
+        return module.weight_as_blockdiagonal_matrix()
+    return module.weight
+
+
+def _copy_adapter_weights(dst_layer, src_layer) -> None:
+    dst_layer.lora_A["default"].weight.data.copy_(src_layer.lora_A["default"].weight.data)
+    dst_layer.lora_B["default"].weight.data.copy_(src_layer.lora_B["default"].weight.data)
+
+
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Thanks for working on the test coverage of BD-LoRA. @Conzel do you have time to take a look?

Conzel:
Looks good to me, thank you for the work!

tests/test_bdlora.py
```diff
@@ -0,0 +1,228 @@
...
+    assert tuple(bd_b.weight.shape) == expected_b_shape
+    assert tuple(vanilla_a.weight.shape) == (4, 10)
+    assert tuple(vanilla_b.weight.shape) == (20, 4)
+
+    vanilla_adapter_params = sum(
+        p.numel() for module in (vanilla_a, vanilla_b) for p in module.parameters() if p.requires_grad
+    )
+    bd_adapter_params = sum(p.numel() for module in (bd_a, bd_b) for p in module.parameters() if p.requires_grad)
+
+    # BD-LoRA must reduce trainable adapter parameters vs vanilla LoRA
```
BenjaminBossan: Could you add a comment on how 120 was derived?

tests/test_bdlora.py
```diff
@@ -0,0 +1,228 @@
...
+    merged_weight = peft_model.base_model.model.lin0.base_layer.weight.data
+    assert torch.allclose(
+        merged_weight,
+        base_weight + expected_delta,
+        atol=1e-5,
+        rtol=1e-4,
+    ), f"Merged weight mismatch for BD-LoRA config={bdlora_config}"
+
+
+def test_bdlora_nblocks_one_matches_vanilla_lora():
```
BenjaminBossan: Please add a short comment on why that is expected.

BenjaminBossan (changes requested):
Thanks for adding those BD-LoRA tests. They look good, just some small clarifications would be appreciated. Could you please also group all tests inside of a test class (`TestBdLora`) to be consistent with other PEFT tests?

tests/test_bdlora.py
eSVeeF: Actually you are right, in that test there is no explanation to any of the hardcoded numbers. I will also add some comments explaining how the expected shape and params are derived, I think they will also help.

eSVeeF:
Thank both for the review.

I added some comments to both vanilla LoRA related tests. I think they are easier to understand now.
Grouped all tests in class TestBdLora.

BenjaminBossan (approved):
Thanks for the updates to the tests, LGTM.

PR #3365, opened 2026-06-27
FIX Call super().__post_init__() in CPTConfig and RandLoraConfig

DaoyuanLi2816:
## What does this PR do?

`CPTConfig` and `RandLoraConfig` override `__post_init__` but do not call `super().__post_init__()`, so `PeftConfig.__post_init__` never runs for them. That base method:

- auto-fills `peft_version` (`if self.peft_version is None: self.peft_version = self._get_peft_version()`), and
- validates `task_type`.

As a result, `CPTConfig(...).peft_version` and `RandLoraConfig(...).peft_version` are left as `None`, so the saved `adapter_config.json` for CPT / RandLoRA adapters does not record the PEFT version that every other adapter type records. The base `task_type` validation is also skipped.

Of the tuner configs that override `__post_init__`, all but these two call `super().__post_init__()`. This PR adds the call (first, matching the convention) so CPT and RandLoRA behave the same.

### How it was verified

Before:
- `CPTConfig(task_type="CAUSAL_LM").peft_version` → `None`
- `RandLoraConfig().peft_version` → `None`

After:
- both → e.g. `'0.19.2.dev0@UNKNOWN'`, and the value round-trips through `save_pretrained` / `from_pretrained`.
- The configs' existing `__post_init__` logic is unaffected (CPT still sets `is_prompt_learning`, `num_virtual_tokens`, …; RandLoRA still sets `peft_type` and the `target_modules` set).
- `tests/test_cpt.py` and `tests/test_randlora.py` pass (15 passed).
- `ruff check` passes on both files.

Note: #3227 also edits `randlora/config.py` (the `save_projection` field default), on different lines — no overlap with this change.

### AI assistance

### DIFF
--- a/src/peft/tuners/cpt/config.py
+++ b/src/peft/tuners/cpt/config.py
@@ -72,6 +72,7 @@ class CPTConfig(PromptLearningConfig):
         """
         Post-initialization hook to set additional attributes after the config is initialized.
         """
+        super().__post_init__()
         # CPT-specific static attributes
         self.is_prompt_learning = True  # Indicates that CPT is a prompt-learning method.
         self.num_layers = None  # Number of layers (optional, not always required).
--- a/src/peft/tuners/randlora/config.py
+++ b/src/peft/tuners/randlora/config.py
@@ -186,6 +186,7 @@ class RandLoraConfig(PeftConfig):
     )
 
     def __post_init__(self):
+        super().__post_init__()
         self.peft_type = PeftType.RANDLORA
         self.target_modules = (
             set(self.target_modules) if isinstance(self.target_modules, list) else self.target_modules
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -110,7 +110,7 @@ ALL_CONFIG_CLASSES = (
     (VeraConfig, {}),
     (VBLoRAConfig, {}),
     (XLoraConfig, {"hidden_size": 32, "adapters": {}}),
-    (CPTConfig, {}),
+    (CPTConfig, {"task_type": "CAUSAL_LM"}),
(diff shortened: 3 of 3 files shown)


BenjaminBossan (changes requested):
Thanks for catching this omission.

I checked why our tests didn't catch this and found that `ALL_CONFIG_CLASSES` in `tests/tst_config.py` doesn't include those two (there might be others that are missing). Could you please add them to the test matrix?

DaoyuanLi2816:
Thanks! Added them. Diffing `ALL_CONFIG_CLASSES` against all public `PeftConfig` subclasses turned up six missing: `CPTConfig` and `RandLoraConfig`, plus `DeloraConfig`, `OFTConfig`, `PveraConfig`, `WaveFTConfig`. All six are in the matrix now.

Two needed a little handling so the generic tests stay green:

- **CPTConfig** only supports `task_type=CAUSAL_LM` (enforced in its `__post_init__`), so it's pinned in `mandatory_kwargs` and skipped in `test_valid_task_type` / `test_invalid_task_type`. It's still exercised by the round-trip tests (`test_methods` / `test_save_pretrained` / `test_to_dict` / pickle / copy / ...), which is what would have caught the original `__post_init__` omission.
- **OFTConfig**'s `from_pretrained` has a backwards-compatibility guard that raises before the generic load path, so it's skipped in the four `from_pretrained`-based tests.

Locally: `tests/test_config.py` → 1137 passed, 12 skipped; `ruff` clean.

BenjaminBossan (approved):
Thanks for taking care of syncing the config tests. The PR LGTM.

PR #3366, opened 2026-06-28
FIX Mixed: honor bias='<prefix>_only' for all compatible tuners (not just lora_only)

DaoyuanLi2816:
## What does this PR do?

`PeftMixedModel`'s `_mark_only_adapters_as_trainable` hardcodes `bias == "lora_only"`:

```python
elif bias == "lora_only":
    ...
else:
    raise ValueError(f"Requested bias: {bias}, is not implemented.")
```

But `MixedModel.COMPATIBLE_TUNER_TYPES` includes OFT, and `OFTConfig` accepts `bias="oft_only"` (a valid `Literal["none", "all", "oft_only"]`). So a mixed model with an OFT adapter and `bias="oft_only"` raises `ValueError: Requested bias: oft_only, is not implemented.`, even though the identical config trains fine in a standalone OFT model.

The body of that branch is already generic — it marks the bias of every compatible `Layers` instance trainable, not just LoRA — so only the guard was too narrow. This aligns it with `BaseTuner._mark_only_adapters_as_trainable`, which uses `bias.endswith("_only")` (and removes the now-resolved `# TODO: check if this is needed for other supported types`).

### How it was verified

- Repro (CPU): a `SimpleNet` mixed model with `OFTConfig(target_modules=["lin1"], oft_block_size=8, bias="oft_only")` raises the `ValueError` on `main` and trains its bias after the fix; standalone OFT with the same config already works.
- Added `test_bias_only_honored_for_non_lora_tuner` — it fails on `main` with the `ValueError` above and passes with this change.
- `tests/test_mixed.py` passes (33 passed); `ruff check` / `ruff format --check` clean.

### AI assistance

### DIFF
--- a/src/peft/tuners/mixed/model.py
+++ b/src/peft/tuners/mixed/model.py
@@ -164,8 +164,7 @@ class MixedModel(BaseTuner):
                 for n, p in model.named_parameters():
                     if "bias" in n:
                         p.requires_grad = True
-            elif bias == "lora_only":
-                # TODO: check if this is needed for other supported types
+            elif bias.endswith("_only"):  # e.g. "lora_only" or "oft_only"
                 for m in model.modules():
                     if isinstance(m, Layers) and hasattr(m, "bias") and m.bias is not None:
                         m.bias.requires_grad = True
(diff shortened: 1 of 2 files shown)


BenjaminBossan (approved):
Good catch. LGTM.

PR #3367, opened 2026-06-28
FIX AdaMSS: register adamss_newB in other_param_names (delete_adapter leak)

DaoyuanLi2816:
## What does this PR do?

`AdamssLayer.adamss_newB` is a per-adapter frozen projection buffer (`BufferDict(persistent=True)`), but it is listed in **neither** `adapter_layer_names` **nor** `other_param_names`. The generic `BaseTunerLayer` machinery iterates `adapter_layer_names + other_param_names`, so it never touches `adamss_newB`:

- `delete_adapter(name)` removes `name` from `adamss_A` / `adamss_B` but leaves it in `adamss_newB` — the deleted adapter's projection buffer leaks.
- per-adapter device moves (`_move_adapter_to_device_of_base_layer`) skip it. (In the common `get_peft_model(...).to(device)` flow `nn.Module.to()` still moves the registered submodule, so that only bites exotic multi-device-base setups; the delete leak is the concrete issue.)

Frozen per-adapter buffers belong in `other_param_names` — this is exactly what VeRA does with its `vera_A` / `vera_B` `BufferDict`s. The existing comment shows the author correctly kept `adamss_newB` out of `adapter_layer_names` (it isn't trainable) but missed that it still needs registering in `other_param_names`.

### How it was verified

- Repro: with two AdaMSS adapters, `delete_adapter("a1")` leaves `"a1"` in every layer's `adamss_newB` on `main` (while `adamss_A` is correctly cleaned); with the fix it's removed.
- Added `test_delete_adapter_removes_newB_buffer` — it fails on `main` and passes with the fix.
- `delete_adapter` still works with no crash (the `adamss_` prefix means the module-level delete already filters these keys), and forward / `merge_and_unload` are unaffected. `tests/test_adamss_asa.py` passes (9 passed); `ruff check` / `ruff format --check` clean.

### AI assistance

### DIFF
--- a/src/peft/tuners/adamss/layer.py
+++ b/src/peft/tuners/adamss/layer.py
@@ -35,11 +35,13 @@ class AdamssLayer(BaseTunerLayer):
 
     # All names of layers that may contain adapter weights (trainable or frozen)
     # Use ModuleDict for proper state_dict key format that PEFT's save/load can handle
-    # Only include trainable parameter containers here (not buffers)
-    # adamss_newB is a buffer (frozen), not a trainable parameter
+    # adapter_layer_names holds the trainable parameter containers (not buffers)
     adapter_layer_names = ("adamss_A", "adamss_B")
-    # other_param_names specifies additional non-tensor metadata
+    # other_param_names holds the remaining per-adapter state, so the generic tuner machinery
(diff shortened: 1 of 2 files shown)


DaoyuanLi2816:
Superseded by #3376, which now registers `adamss_newB` alongside the same fix for the other tuners (per the reviewer's request to pool all cases into one PR).

PR #3376, opened 2026-07-01
FIX Register per-adapter containers so delete_adapter removes them (unilora, fourierft, psoft)

DaoyuanLi2816:
## What does this PR do?

`BaseTunerLayer.delete_adapter` removes a deleted adapter's per-adapter entries only from the containers listed in `adapter_layer_names` / `other_param_names`. Three tuners keep per-adapter state in containers that are in neither tuple, so deleting an adapter leaves that state behind (same class of bug as #3367, which registered `adamss_newB`):

| tuner | leaked container(s) | kind |
|---|---|---|
| `unilora` | `r`, `unilora_indices_A`, `unilora_indices_B` | the index `BufferDict`s are `persistent=True` (part of the state dict) |
| `fourierft` | `indices` | per-adapter frequency-location tensors |
| `psoft` | `random_seed`, `_psoft_A_cache`, `_psoft_B_cache` | SVD-factor `BufferDict`s |

For `unilora`, `unilora_scales_A/B` were already registered but the paired `unilora_indices_A/B` were not. For `fourierft`, the two sibling "sparse spectrum" methods already follow the convention — `shira` registers `shira_indices` and `waveft` registers `waveft_indices` — only `fourierft` omitted its index dict.

### Reproduction

Adding two adapters, deleting one, then inspecting the layers shows the deleted adapter persists in the unregistered containers, while the registered siblings (e.g. `unilora_scales_A`, `fourierft_scaling`, `psoft_R`) are correctly removed:

```
unilora    after delete "a2":  unilora_indices_A = {a1, a2}  (leak)   unilora_scales_A = {a1}  (ok)
fourierft  after delete "a2":  indices           = {a1, a2}  (leak)   fourierft_scaling = {a1} (ok)
psoft      after delete "a2":  _psoft_A_cache    = {a1, a2}  (leak)   psoft_R          = {a1}  (ok)
```

### Fix

Register the containers in `other_param_names`. Safety of adding these specific names, checked against every consumer of the two tuples:

- `delete_adapter` / `_all_available_adapter_names`: now include them (the fix).
- `_move_adapter_to_device_of_base_layer`: skips plain dicts and never casts to an integer dtype, so the `long` index buffers/dicts are safe.
- `cast_adapter_dtype`: iterates `.modules()`, not the tuples, and only upcasts `float16`/`bfloat16`, so it is unaffected and never touches the `long` index tensors.
- module-level `delete_adapter` key filtering (`key for key ... if prefix not in key`): every added name either contains the tuner `prefix` (`unilora_indices_*`, `_psoft_*_cache`) or is a plain dict invisible to `named_modules()` (`indices`, `r`, `random_seed`), so no submodule keys are affected.
- `save_and_load`: does not consult either tuple (it is driven by `prefix` + per-method code), so serialization is unchanged (verified: the unilora save/load round-trip tests still pass).

Stateless per-adapter `*_dropout` `ModuleDict`s are intentionally left unregistered — they hold no parameters/buffers and are not part of the state dict.

### Verification

- Added `tests/test_delete_adapter_container_registration.py`: parametrized over the three tuners, it deletes one of two adapters and asserts the deleted adapter no longer appears in the affected containers (and the remaining adapter still does). It fails on `main` and passes with this change.
- `tests/test_unilora.py` passes in full (save/load round-trip unaffected); `ruff check` / `ruff format --check` are clean. Validated locally on an RTX 4080.

A few other tuners have analogous leaks of per-adapter *scalar* hyperparameter dicts (e.g. `vblora` `r`/`topk`, `poly` `poly_type`, `gralora` `gralora_k`, `frod` `r`). I scoped this PR to the containers holding real tensor/buffer state (closest to #3367) — happy to extend it to normalize those as well if you'd prefer.

---

*This PR was written with AI assistance; all code and tests were human-reviewed and validated locally.*


tests/test_delete_adapter_container_registration.py
```diff
@@ -0,0 +1,75 @@
...
+    ),
+    "psoft": (
+        lambda: PsoftConfig(target_modules=["lin0", "lin1"], r=4),
+        ["random_seed", "_psoft_A_cache", "_psoft_B_cache"],
+    ),
+}
+
+
+@pytest.mark.parametrize("tuner", list(CASES))
+def test_delete_adapter_removes_previously_unregistered_containers(tuner):
```
BenjaminBossan: Could you please move the tests into `test_tuners_utils.py` and nest them inside a dedicated test class?

BenjaminBossan (changes requested):
Thanks for catching those dangling attributes after adapter deletion. Generally, I agree with the direction of the change.

> A few other tuners have analogous leaks of per-adapter scalar hyperparameter dicts (e.g. vblora r/topk, poly poly_type, gralora gralora_k, frod r). I scoped this PR to the containers holding real tensor/buffer state (closest to https://github.com/huggingface/peft/pull/3367) — happy to extend it to normalize those as well if you'd prefer.

Yes, could you please pool all the bugs of that type into this PR?

tests/test_delete_adapter_container_registration.py
DaoyuanLi2816: Done — moved into `tests/test_tuners_utils.py` as a dedicated `TestDeleteAdapterContainerRegistration` class, extended to cover all nine tuners.

DaoyuanLi2816:
Pooled the remaining scalar-dict leaks into this PR as requested: `vblora` `r`/`topk` (the class declared no `other_param_names` at all), `gralora` `gralora_k`, `poly` `poly_type`, `frod` `r`, `osf` `effective_rank`, and `rank_dropout_scale` in the shared `LycorisLayer` base (only LoKr populates it; for LoHa the dict stays empty so it's a no-op there). Each was verified the same way as the first batch: with two adapters, deleting one left the deleted adapter's entry behind before the change and removes it after.

The regression tests now live in `tests/test_tuners_utils.py` under a dedicated class, parametrized over all nine tuners (fail before the fix, pass after). Stateless `*_dropout` `ModuleDict`s remain intentionally unregistered, and I left the transient merge-scratch caches (`peanut._cached_delta_weights`, `frod._frod_merged_delta`) alone since they're emptied on `unmerge`.

BenjaminBossan:
Thanks for the update. One thing that was bugging me about the testing is that it's strictly opt-in: If a new PEFT method is added which is missing an attribute on deletion, we would not catch it with the current test. Therefore, I would prefer a test that automatically catches dangling attributes. Here is a proposal:

```python
    def test_delete_adapter_properly_cleans_up_after_itself(self, test_name, model_id, config_cls, config_kwargs):
        # TODO
        model = self.transformers_class.from_pretrained(model_id).to(self.torch_device)
        config = config_cls(**config_kwargs)
        model = get_peft_model(model, config, adapter_name="foobar")

        def get_dict_like_attrs(obj):
            for attr_name in dir(obj):
                if attr_name.startswith("_"):
                    continue

                attr = getattr(obj, attr_name)
                is_dict_like = callable(getattr(attr, "keys", None)) and hasattr(attr, "__getitem__") and hasattr(attr, "__contains__")
                if not is_dict_like:
                    continue

                if "foobar" in attr:
                    yield attr_name, attr

        # collect dict-like attributes
        dict_attrs = list(get_dict_like_attrs(model))
                dict_attrs.extend(list(get_dict_like_attrs(module)))

        # sanity check: there should be at least one such attribute
        assert dict_attrs, f"No dict-like attributes found on {config.peft_type}, please check the test"

        model.delete_adapter(adapter_name="foobar")
        dangling_attrs = []
        for attr_name, attr in dict_attrs:
            if "foobar" in attr:
                dangling_attrs.append(attr_name)

        assert not dangling_attrs, f"Found dangling attributes after adapter deletion: {sorted(dangling_attrs)}"
```

You could add after [this test](https://github.com/huggingface/peft/blob/5bfa8cfdfbd0698cc792901bcb81f57d60009ea6/tests/test_custom_models.py#L3443), for instance. It should broadly catch every attribute that is not being cleaned up properly. When I ran it, it catches quite a few more issues than you fixed, partly because of

> Stateless per-adapter `*_dropout` `ModuleDict`s are intentionally left unregistered — they hold no parameters/buffers and are not part of the state dict.

I think it makes sense to clean those up as well. LMK what you think and if you're up to extending the PR accordingly.

DaoyuanLi2816:
Great idea — adopted your generic test as `test_delete_adapter_properly_cleans_up_after_itself` (right after `test_delete_adapter`), auto-discovering every dict-like per-adapter container on the model and its tuner layers. It flagged more than the first batch, all now fixed:

- **dropout `ModuleDict`s**: `vera_dropout`, `pvera_dropout`, `randlora_dropout`, `frod_dropout`, `vblora_dropout`, `unilora_dropout`
- **other per-adapter dicts**: lora `use_dora`/`use_rslora`/`lora_bias`, vera/pvera `r`, randlora `r`/`scaling`
- **`adamss_newB`** (this folds in my separate #3367 — I'll close that as superseded)

Registering the dropout `ModuleDict`s surfaced a **latent crash in the module-level `delete_adapter`** (`tuners_utils.py`): it built a `key_list` and re-resolved each key with `get_submodule` *inside* the loop, but the layer-level delete mutates the module tree mid-loop, so a key pointing at a just-removed `ModuleDict` entry raised `AttributeError: ModuleDict has no attribute ...`. This only bit methods whose `prefix` doesn't match the dropout name (e.g. vera's `vera_lambda_`). Fixed by iterating the modules captured up front from `named_modules()` instead of re-resolving keys — no behavior change otherwise.

Validation: I ran an equivalent of your test across all 32 tuner methods (build → add adapter → delete → assert no dangling dict-like attr) — all clean, no crashes. `test_tuners_utils.py` passes (235). Note: `test_custom_models.py` doesn't collect on my local Windows setup (an unrelated native crash in the diffusers import), so I validated the new test's logic via that standalone equivalent rather than the file directly — CI will exercise the real parametrization.

BenjaminBossan (changes requested):
@DaoyuanLi2816 Thanks for updating the test and extending the coverage. As you can see, some tests are failing because of dangling attributes. Could you please take a look? If you encounter any ambiguities with how to deal with those, LMK.

DaoyuanLi2816:
Update: the CI run of the generic test did its job and caught two containers my local scan had missed —

- `lora_magnitude_vector` (the DoRA magnitude `ModuleDict`; my local LoRA case didn't enable DoRA) → added to LoRA's `adapter_layer_names`.
- `deft_dropout` (the `deft` method wasn't present on this branch's base) → registered in its `other_param_names`.

I've pushed both fixes and rebased onto `main`, which also brought in `deft` and the MiCA `frozen_peft_weight_names` handling. Locally I re-ran an equivalent of the generic check across all methods incl. DoRA / MiCA / deft — no dangling per-adapter container remains after `delete_adapter`. The new CI run looks like it's waiting for workflow approval; once it runs it should confirm across the full parametrization.

src/peft/tuners/tuners_utils.py
```diff
@@ -2224,11 +2224,14 @@ def delete_adapter(
             The name of remaining adapter(s) after deletion, or `None` if there are no active adapters left. Use this
             to set the new active adapter of the model if necessary.
     """
-    key_list = [key for key, _ in model.named_modules() if prefix not in key]
+    # Capture (key, module) pairs up front. Deleting an adapter mutates the module tree (e.g. it removes the adapter's
+    # entry from a `ModuleDict` such as `<method>_dropout`), so re-resolving a key with `get_submodule` mid-loop can
+    # raise for a submodule that was just removed. `named_modules()` already yields the module itself, so keep it
+    # instead of looking it up again by key.
```
BenjaminBossan: This whole comment can be confusing because it refers to the previous implementation, which is removed now. So let's just remove it, the new implementation is more straightforward than the previous one anyways. Also, let's not assign `modules_to_update`. Let's just loop over `model.named_modules()` and `continue` if `prefix not in key`.

tests/test_custom_models.py
```diff
@@ -3498,6 +3498,47 @@ def test_delete_adapter(self, test_name, model_id, config_cls, config_kwargs):
         _skip_tests_with_multiple_adapters_with_target_parameters(config_cls, config_kwargs)
         self._test_delete_adapter(model_id, config_cls, config_kwargs)
 
+    @pytest.mark.parametrize("test_name, model_id, config_cls, config_kwargs", TEST_CASES)
+    def test_delete_adapter_properly_cleans_up_after_itself(self, test_name, model_id, config_cls, config_kwargs):
+        # Generic guard against dangling per-adapter state: `delete_adapter` should remove the adapter's entry from
+        # every dict-like container on the model and its tuner layers, not just those registered in
+        # `adapter_layer_names` / `other_param_names`. Automatically discovers such containers so that a newly added
```
BenjaminBossan: The description is misleading, the point of the test is not to delete state beyond the keys registered in `adapter_layer_names` / `other_param_names`. These attributes are just the mechanism used to keep track of what to delete. The point of the test is to ensure that no adapter-specific attributes remain after deletion. period.

BenjaminBossan (changes requested):
We're basically good, thanks, just two comments regarding two comments.

DaoyuanLi2816:
Both addressed:
- `tuners_utils.py`: removed the stale comment (it described the old `get_submodule` implementation) and dropped the `modules_to_update` list — now looping over `model.named_modules()` directly and `continue`-ing when `prefix in key`.
- `test_custom_models.py`: reworded the comment to frame it as "after `delete_adapter`, no adapter-specific attributes should remain" — with `adapter_layer_names`/`other_param_names` described as just the tracking mechanism, not the point of the test.

tests/test_custom_models.py
```diff
@@ -3500,10 +3500,10 @@ def test_delete_adapter(self, test_name, model_id, config_cls, config_kwargs):
 
     @pytest.mark.parametrize("test_name, model_id, config_cls, config_kwargs", TEST_CASES)
     def test_delete_adapter_properly_cleans_up_after_itself(self, test_name, model_id, config_cls, config_kwargs):
-        # Generic guard against dangling per-adapter state: `delete_adapter` should remove the adapter's entry from
-        # every dict-like container on the model and its tuner layers, not just those registered in
-        # `adapter_layer_names` / `other_param_names`. Automatically discovers such containers so that a newly added
-        # PEFT method that forgets to register one is caught here.
+        # Generic guard against dangling per-adapter state: after `delete_adapter`, no adapter-specific attributes
+        # should remain on the model or its tuner layers. `adapter_layer_names` / `other_param_names` are just the
+        # mechanism used to track what to delete; this test discovers every dict-like per-adapter container
```
BenjaminBossan: > `adapter_layer_names` / `other_param_names` are just the mechanism used to track what to delete

This doesn't need mentioning, it only makes sense in the context of this review and not when reading the test.

BenjaminBossan (changes requested):
Thanks for the update, still one issue with one comment, otherwise looks good.

DaoyuanLi2816:
Trimmed that clause — the comment now just states the invariant (no adapter-specific attributes remain after `delete_adapter`) without referencing the tracking mechanism. Thanks!

BenjaminBossan (changes requested):
Thanks @DaoyuanLi2816. The CI revealed that there is still one test failing (ignore the failing diffusion tests on Windows, they're unrelated):

`FAILED tests/test_custom_models.py::TestPeftCustomModel::test_delete_adapter_properly_cleans_up_after_itself[Vanilla MLP 1 LoRA with MiCA-MLP-LoraConfig-config_kwargs14] - AssertionError: Found dangling attributes after adapter deletion: ['frozen_peft_weight_names', 'frozen_peft_weight_names']`

This attribute comes from #3260 and I think it's a genuine omission that should be added to `LoraLayer.delete_adapter`. Please ensure that the new test passes locally before pushing the next changes.

DaoyuanLi2816:
Fixed the remaining `[Vanilla*]` failure. Root cause: `frozen_peft_weight_names` was cleaned via copy-and-rebind (`self.frozen_peft_weight_names = copy_without_adapter`), whereas the sibling containers just above are deleted in place. The generic test holds a reference to the dict captured *before* deletion, so the in-place containers reflect the removal but the rebound one left the old (still-populated) dict object dangling.

Switched it to an in-place `del self.frozen_peft_weight_names[adapter_name]`, consistent with the loop right above. It's safe w.r.t. the mutable class-level default: an entry only ever exists after `_register_frozen_peft_weight`, which rebinds the instance to its own copy first, so the guarded `del` never touches the shared class dict.

BenjaminBossan (approved):
Thanks for the last push, LGTM.

PR #3379, opened 2026-07-01
fix: store miss_fn per-adapter in MiSS layer

peft-jambot:
## Description

Fixes #6 (peft-jambot/peft#6)

The `miss_fn` attribute (derived from `init_weights`) was stored as a single value on the `MissLinear` instance instead of per-adapter. This meant adding a second MiSS adapter with a different `init_weights` value (e.g. `True` vs `"bat"`) would override the first adapter's setting, leading to incorrect behavior in merge, unmerge, forward, and LoRA conversion operations.

### Changes

- Convert `self.miss_fn` from a single attribute to a dict keyed by `adapter_name`, following the same pattern used by `miss_r`, `miss_mini_r`, and other per-adapter attributes.
- Set `self.miss_fn[adapter_name]` in `update_layer` (which is called for each adapter) instead of in `MissLinear.__init__` (called only once).
- Add `miss_fn` to `other_param_names` so it is included in adapter lifecycle operations (`delete_adapter`, `_all_available_adapter_names`, etc).
- Update all references to `self.miss_fn` in `merge`, `unmerge`, `get_delta_weight_miss`, `forward`, and the LoRA conversion code to use per-adapter lookup.

### Tests

Added `TestMissInitialization` class to `tests/test_initialization.py` with three tests:
- `test_miss_fn_per_adapter`: verifies `miss_fn` is stored per-adapter when adding two adapters with different `init_weights` values.
- `test_miss_fn_per_adapter_forward`: ensures forward pass works correctly with different `init_weights` per adapter.
- `test_miss_fn_per_adapter_three_variants`: tests all three `init_weights` variants (`True`, `"bat"`, `"mini"`) coexisting on the same layer.

### Test results

```
python -m pytest tests/test_initialization.py::TestMissInitialization -x -v
# 3 passed

python -m pytest tests/test_custom_models.py -k "miss and not adm" -x -v
# 378 passed, 10 skipped

python -m pytest tests/test_lora_conversion.py -k "miss" -x -v
# 6 passed
```

### AI assistance

This PR was created with AI assistance. The task was assigned via the peft-jambot issue tracker (issue #6). The changes were reviewed and tested locally.


src/peft/tuners/miss/layer.py
```diff
@@ -333,7 +334,10 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
         elif self.merged:
             result = self.base_layer(x, *args, **kwargs)
         else:
-            if self.miss_fn == "bat":
+            # Determine the MiSS variant from the active adapters. When multiple adapters are active, they must
+            # all use the same variant; otherwise, the forward pass is ambiguous.
+            active_fns = {self.miss_fn[adapter] for adapter in self.active_adapters if adapter in self.miss_block}
```
BenjaminBossan: This looks incorrect. If I have a mix of bat and non-bat MiSS adapters, I would treat all of them as non-bat. Shouldn't it be possible to mix bat and non-bat in the same forward pass? If it's not possible, I would rather raise an error when incompatible MiSS adapters are being added.

BenjaminBossan (changes requested):
Thanks for this fix but I'm not convinced about how the forward pass is handled. Please check my comment.

src/peft/tuners/miss/layer.py
peft-jambot: Good point, thanks. Fixed in commit 136910c9.

The forward pass now separates active adapters into bat and non-bat lists. bat adapters modify the base weight first (via `get_delta_weight`), then non-bat adapters (balance/mini) add their delta to the output. Both contributions are applied when mixed.

Two tests added (`test_miss_mix_bat_and_non_bat_forward`, `test_miss_mix_bat_and_mini_forward`) verifying that mixed forward passes produce outputs distinct from either adapter alone.

Please re-review when you have a chance.

tests/test_initialization.py
```diff
@@ -5784,3 +5656,144 @@ def test_ensure_weight_tying_not_tying_when_model_config_tie_false_target_module
...
+
+        config2 = MissConfig(target_modules=["lin0"], r=2, init_weights="mini", mini_r=1)
+        model.add_adapter("adapter2", config2)
+
+        layer = model.base_model.model.lin0
+        assert layer.miss_fn["default"] is True
+        assert layer.miss_fn["adapter1"] == "bat"
+        assert layer.miss_fn["adapter2"] == "mini"
+
+    def test_miss_mix_bat_and_non_bat_forward(self, data):
```
BenjaminBossan: This test and `test_miss_mix_bat_and_mini_forward` don't really check if the different inits are correctly applied and as such don't add too much, except for guarding against the incorrect implementation that you initially submitted. Let's remove these tests.

BenjaminBossan (changes requested):
Thanks for the update to `forward`, I just have a small comment concerning the testing.

tests/test_initialization.py
peft-jambot: Removed both tests in commit ba1f073c. Thanks for the feedback.

BenjaminBossan (approved):
LGTM, failing CI is unrelated. @githubnemo please review.

tests/test_initialization.py
```diff
@@ -6332,3 +6127,80 @@ def test_second_adapter_overlapping_target_modules_after_delete_and_readd(self):
...
+        config1 = MissConfig(target_modules=["lin0"], r=2, init_weights="bat")
+        model.add_adapter("adapter1", config1)
+
+        config2 = MissConfig(target_modules=["lin0"], r=2, init_weights="mini", mini_r=1)
+        model.add_adapter("adapter2", config2)
+
+        layer = model.base_model.model.lin0
+        assert layer.miss_fn["default"] is True
+        assert layer.miss_fn["adapter1"] == "bat"
+        assert layer.miss_fn["adapter2"] == "mini"
```
githubnemo, line 6189: It would be good to also have a test that reproduces the overriding behavior this PR fixes.

Idea: have two adapters (one bat, one with a different init fn), gather per-adapter outputs, add both adapters into one model, trigger init fn override bug, observe that output != (adapter1_output + adpater2_output).

BenjaminBossan:
*not stale*

githubnemo (commented):
One comment, otherwise LGTM

BenjaminBossan:
@githubnemo Please review again. Note that the latest changes are code that I wrote and then passed as a patch to the bot, who made small adjustments.

githubnemo (approved):
Thanks for addressing :) LGTM

PR #3381, opened 2026-07-01
FIX MiSS with bat init error with some quantization backends

BenjaminBossan:
When using MiSS with the `init_weights='bat'`, the `forward` pass requires getting the non-quantized base weights. When the quantization backend does not support dequantization, we now detect this and raise a helpful error message. Tests have been updated to check for this error message.

Normally, we should have caught this error. The reason why we didn't is that we have a helper function for testing that changes PEFT config `kwargs` to be non-zero initialization. When passing `init_weights='bat'` to that function, we would override `'bat'` with `False`. Thus, the `'bat'` path was not actually being tested. The helper function has now been updated not to override `'bat'`.


tests/test_quantization.py
```diff
@@ -284,6 +292,11 @@ def test_quantized_output_similar_to_non_quantized(self, config_cls, config_kwar
         torch.manual_seed(SEED)
         model = get_peft_model(model, config).eval()
 
+        if (config_cls == MissConfig) and (config.init_weights == "bat") and (not quant.supports_merge):
+            with pytest.raises(ValueError, match="is not supported for quantization with"), torch.inference_mode():
+                model(dummy_input).logits
+            return
+
```
githubnemo: Let's handle this generically right now since it is not unlikely that we will have more examples of this in the future.

BenjaminBossan: I factored out the check but not the whole block. I don't think there is an elegant way to move the early `return` into a function.

src/peft/tuners/miss/layer.py
```diff
@@ -334,6 +334,13 @@ def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
             result = self.base_layer(x, *args, **kwargs)
         else:
             if self.miss_fn == "bat":
+                if (self.quantization_backend is not None) and (not self.quantization_backend.supports_merge):
+                    raise ValueError(
+                        "Using MiSS with `init_weights='bat'` is not supported for quantization with "
+                        f"{self.quantization_backend.backend_name}. Use a different quantization backend or a "
```
githubnemo: I think it would be a good idea to let the user know what they need to be looking for in quantization methods to make this work. I think explaining what's missing for the current quantization method is sufficient. For example: "is not supported for quantization with <backend_name> as it does not support de-quantization."

tests/test_quantization.py
githubnemo: Looks good.

githubnemo (commented):
There's a `# TODO` comment left and I had one suggestion regarding the user-facing error message. Otherwise LGTM :)

githubnemo (approved):
Thanks! LGTM :)

Issue #3380, opened 2026-07-01
Reviving #1807: Riemannian Preconditioned LoRA — coordination before drafting a PR


smellslikeml:
Hi @BenjaminBossan — we'd like to revive the Riemannian Preconditioned LoRA integration from #1807 and check maintainer interest before opening a fresh PR.

## Context

- Paper: [Riemannian Preconditioned LoRA for Fine-Tuning Foundation Models](https://arxiv.org/abs/2402.02347v3) adds a small `r x r` preconditioner to LoRA's gradient step; claim is faster / more LR-robust fine-tuning at negligible overhead.
- Reference impl: [pilancilab/Riemannian_Preconditioned_LoRA](https://github.com/pilancilab/Riemannian_Preconditioned_LoRA) (MIT license).
- Prior PR: #1807 by @fangzhaozhang (May 2024). Reached your second review round on 2024-07-17 ("getting closer but still a few areas need to be improved"), then went stale and auto-closed on 2024-09-21.

## Proposed scope (narrow)

- `src/peft/optimizers/riemannian.py` — a `create_riemannian_optimizer(model, optimizer_cls, *, lr, reg, **kwargs)` factory in the LoRA+ / LoRA-FA style, with the preconditioner applied via a subclass wrapper so it works with any `Optimizer` (AdamW, SGD, …) — addresses your `issubclass` comment on #1807.
- `src/peft/optimizers/__init__.py` — one-line export.
- `tests/test_riemannian_lora.py` — new tests.

Round-1 review asks from #1807 folded in preemptively:
- PEP 8 class naming; docstrings replace implementation-detail strings
- No `assert` in code — proper `TypeError` / `ValueError` / `RuntimeError`
- `issubclass` check on `optimizer_cls`
- bf16-stable via float32 preconditioner computation
- DoRA compatibility (name-based `lora_A` / `lora_B` pair discovery skips the magnitude vector)

Explicitly deferring from #1807's tree:
- `src/peft/tuners/lycoris_utils.py` / `poly/router.py` — unclear necessity; will drop unless you confirm they're needed
- `examples/riemannian_lora/` — happy to add here or as a follow-up
- The accidentally-committed `events.out.tfevents.*` file

## Draft branch for review

A draft is pushed to a fork so you can see the diff before we open a PR:

**Compare against `main`:** https://github.com/huggingface/peft/compare/main...smellslikeml:peft:feat/riemannian-preconditioned-lora

[...]

PR #3382, opened 2026-07-01
Add Riemannian-preconditioned LoRA optimizer

smellslikeml:
Adds `create_riemannian_optimizer` for the r×r Riemannian preconditioning recipe from *Riemannian Preconditioned LoRA for Fine-Tuning Foundation Models* ([arXiv:2402.02347v3](https://arxiv.org/abs/2402.02347v3)). Before every base-optimizer step, the LoRA factor gradients are rescaled as:

```
g_A ← (BᵀB + reg·I_r)⁻¹ @ g_A
g_B ← g_B @ (AAᵀ + reg·I_r)⁻¹
```

Non-LoRA parameters are updated unchanged. The r×r preconditioner keeps storage and runtime overhead small in the LoRA rank.

## Coordination

Filed as coordination issue #3380 before pushing. Original PR author @fangzhaozhang explicitly endorsed a revival ([comment](https://github.com/huggingface/peft/issues/3380#issuecomment-4858962847)):

> "Thanks for interests! I'm caught up by my school work after the initial implementation, so feel free to open a fresh PR if you'd like."

## What this PR does

Revives the design accepted in principle in #1807, folding in round-1 review feedback preemptively:

- **PEP 8 class naming**, docstrings replace implementation-detail strings
- **No `assert` in code** — proper `TypeError` / `ValueError` / `RuntimeError`
- **`issubclass` check** on `optimizer_cls` — the wrapper works with any `torch.optim.Optimizer` subclass (AdamW, SGD, …)
- **`torch.linalg.pinv` with damping** for numerical stability on near-rank-deficient factors
- **bf16-stable** via float32 preconditioner computation, cast back to grad dtype
- **DoRA-compatible** — name-based `lora_A` / `lora_B` pair discovery skips the magnitude vector

Deviates from #1807 in one design choice: rather than copying and modifying `AdamW.step()`, this is a **subclass-wrapper approach** — a dynamic subclass of `optimizer_cls` that applies the preconditioner to `.grad` tensors before calling `super().step(closure)`. Thin diff on top of the base optimizer's step, and works with any optimizer.

## `reg` calibration (07-07 update)

The `reg` damping default is `1e-2`, matching @fangzhaozhang's reference example in #1807 (`create_riemannian_optimizer(..., reg=1e-2)`) and the paper author's public repo. On the MetaMath benchmark (llama-3.2-3B, rank 32, 5000 steps, otherwise-default settings):

| `reg` | GSM8K test acc. |
|---|---:|
| `1e-6` (initial ship) | 0.45 |
| `1e-2` (this PR) | 0.475 |
| default LoRA r=32 baseline | 0.49 |

The paper's reported gains are modest and rank-dependent — Table 1: +0.7 BLEU on GPT-2 medium at r=4; Table 2: +0.7pp GLUE avg on 4-bit Mistral 7B at r=16. MetaMath r=32 isn't a paper-benchmarked setting; the paper's stronger claim is LR-range robustness (Figures 1, 3) rather than fixed-LR gains at fixed rank.

## Scope explicitly deferred from #1807's tree

- `src/peft/tuners/lycoris_utils.py` / `poly/router.py` changes — unclear necessity, dropped. Happy to add back if you confirm they're needed.
- `examples/riemannian_lora/` — happy to add here or as a follow-up.
- The accidentally-committed `events.out.tfevents.*` file — dropped.

## Files

- `src/peft/optimizers/riemannian.py` — factory + preconditioner
- `src/peft/optimizers/__init__.py` — one-line export
- `tests/test_riemannian_lora.py` — 9 tests
- `method_comparison/MetaMathQA/utils.py` — dispatch entry + `TrainConfig` validation for `optimizer_type="riemannian"`
- `method_comparison/MetaMathQA/experiments/lora/llama-3.2-3B-rank32-riemannian/` — benchmark experiment (adapter + training params)

## Test plan

- `pytest tests/test_riemannian_lora.py` — 9 passed (factory happy-path, subclass permissiveness on AdamW + SGD, DoRA compatibility, `ValueError` on no-LoRA model, `TypeError` on non-Optimizer class, exact paper-formula verification, bf16 gradient finiteness)
- `make style` — clean
- MetaMath benchmark on Colab A100 with the corrected `reg=1e-2`: 5000 steps completed, GSM8K test accuracy 0.475 (see `reg` calibration table above)

## AI assistance

AI-assisted patch. Every changed line reviewed; local `pytest` and `make style` runs exercised the code paths. Following [`CLAUDE.md`](https://github.com/huggingface/peft/blob/main/CLAUDE.md)'s AI-assisted contribution disclosure guidelines.

Credit for the original design + engagement with round-1 review goes to @fangzhaozhang (#1807), who is co-authored on the fix commit at their round-1 co-authorship suggestion.

Co-Authored-By: remyx-ai[bot] <289541483+remyx-ai[bot]@users.noreply.github.com>


src/peft/optimizers/riemannian.py
```diff
@@ -0,0 +1,156 @@
+# Copyright 2025-present the HuggingFace Inc. team.
```
BenjaminBossan:
```suggestion
# Copyright 2026-present the HuggingFace Inc. team.
```

src/peft/optimizers/riemannian.py
```diff
@@ -0,0 +1,156 @@
...
+import torch
+from torch.optim import Optimizer
+
+from ..peft_model import PeftModel
+
+
+def _collect_lora_pairs(
+    model: torch.nn.Module,
+) -> list[tuple[torch.nn.Parameter, torch.nn.Parameter]]:
+    """Return the list of ``(lora_A, lora_B)`` weight parameter pairs of ``model``.
```
BenjaminBossan: Let's use single backticks in docstrings here and throughout (markdown syntax).

src/peft/optimizers/riemannian.py
```diff
@@ -0,0 +1,156 @@
...
+
+    Pairs are matched by name substitution: for every parameter whose name contains ``lora_A`` the sibling ``lora_B``
+    parameter is looked up by substituting the substring. Only 2D weight matrices that both require gradients are
+    returned, which is exactly what the ``r x r`` preconditioner is defined for. DoRA's ``lora_magnitude_vector`` is
```
BenjaminBossan: DoRA would only be one example. Let's reword this to make it clear it's not about DoRA specifically.

src/peft/optimizers/riemannian.py
```diff
@@ -0,0 +1,156 @@
...
+    ignored because it does not match either factor and is not part of the low-rank product.
+    """
+    params = dict(model.named_parameters())
+    pairs = []
+    for name, param_a in params.items():
+        if "lora_A" not in name:
```
BenjaminBossan: Checking for `".lora_A."`and `".lora_B."` would be more precise, right?

src/peft/optimizers/riemannian.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class _RiemannianPreconditioner:
+    """Applies the ``r x r`` Riemannian preconditioner in-place to LoRA gradients.
+
+    For each ``(lora_A, lora_B)`` pair, the Euclidean gradients are rescaled to the scaled-gradient / Riemannian
+    directions defined by the paper: ``g_A`` is replaced by ``(B^T B + reg * I_r)^-1 @ g_A``, and ``g_B`` is replaced
+    by ``g_B @ (A A^T + reg * I_r)^-1``. Both preconditioners are ``r x r``, so storage and runtime overhead are small
+    in the LoRA rank.
+    """
```
BenjaminBossan, line 66: Let's document what `reg` is.

tests/test_riemannian_lora.py
```diff
@@ -0,0 +1,207 @@
...
+def _forward_backward(model):
+    """Run one forward + backward pass to populate gradients on `model`."""
+    loss = nn.CrossEntropyLoss()
+    x = torch.randint(100, (2, 4, 10)).to(torch_device)
+    output = model(x).permute(0, 3, 1, 2)
+    label = torch.randint(16, (2, 4, 10)).to(torch_device)
+    loss(output, label).backward()
+
+
+# ── factory: happy path ──────────────────────────────────────────────────────
```
BenjaminBossan: This type of comment can be removed. Instead, if the tests are not self-explanatory, add a comment at the start of the test.

tests/test_riemannian_lora.py
```diff
@@ -0,0 +1,207 @@
...
+
+
```
BenjaminBossan: Let's put all tests inside a single test class that bundles them.

tests/test_riemannian_lora.py
```diff
@@ -0,0 +1,207 @@
...
+    # with a non-zero gradient history and update.
+    for _ in range(2):
+        optim.zero_grad()
+        _forward_backward(model)
+        optim.step()
+
+    after = [p.detach().clone() for pair in lora_pairs for p in pair]
+    # Every LoRA weight should have moved after two preconditioned steps.
+    for pre, post in zip(before, after):
+        assert not torch.allclose(pre, post), "LoRA weight did not update after two steps"
```
BenjaminBossan: Let's use a reasonable tolerance for `allclose` here and below.

tests/test_riemannian_lora.py
```diff
@@ -0,0 +1,207 @@
...
+def test_raises_when_optimizer_cls_is_not_optimizer_subclass():
+    model = _make_lora_model()
+    with pytest.raises(TypeError, match="subclass of torch.optim.Optimizer"):
+        create_riemannian_optimizer(model, dict, lr=1e-3)  # type: ignore[arg-type]
+
+
+# ── preconditioner math ──────────────────────────────────────────────────────
+
+
+def test_preconditioner_matches_paper_formula():
```
BenjaminBossan: IMO, this test is not very useful, as it more or less mirrors the code used above to implement the optimizer. Maybe there are some edge cases that can be sanity checked without reimplementing the computation in the test, but as is, I think we can remove the test.

tests/test_riemannian_lora.py
```diff
@@ -0,0 +1,207 @@
...
+    eye_r = torch.eye(r)
+
+    expected_g_a = torch.linalg.pinv(b.T @ b + reg * eye_r) @ g_a_orig.to(torch.float32)
+    expected_g_b = g_b_orig.to(torch.float32) @ torch.linalg.pinv(a @ a.T + reg * eye_r)
+
+    assert torch.allclose(lora_a.grad, expected_g_a.to(lora_a.dtype), atol=1e-5)
+    assert torch.allclose(lora_b.grad, expected_g_b.to(lora_b.dtype), atol=1e-5)
+
+
+def test_bf16_gradients_preconditioned_stably():
```
BenjaminBossan: Would it make sense to parametrize the test over bf16 + fp16?

BenjaminBossan (changes requested):
Thanks @smellslikeml for reviving the Riemannian-preconditioned optimizer PR. I did a first review and I agree that this more generic approach is better. I have a couple of comments but generally, this PR is already quite mature.

One thing to add to the PR would be to extend the MetaMath benchmark to allow using this optimizer there. Check these lines: https://github.com/huggingface/peft/blob/e4fe61b43151da1125b8d6e6846250f66150e58b/method_comparison/MetaMathQA/utils.py#L280-L286 After adding the Riemannian optimizer there, you would also have to add an experiment that uses it.

@fangzhaoz If you have the opportunity, it would be great if you could review the PR too. If @smellslikeml agrees, I would also suggest to add you as a co-author on the final commit.

smellslikeml:
> Thanks @smellslikeml for reviving the Riemannian-preconditioned optimizer PR. I did a first review and I agree that this more generic approach is better. I have a couple of comments but generally, this PR is already quite mature.
> 
> One thing to add to the PR would be to extend the MetaMath benchmark to allow using this optimizer there. Check these lines:
> 
> https://github.com/huggingface/peft/blob/e4fe61b43151da1125b8d6e6846250f66150e58b/method_comparison/MetaMathQA/utils.py#L280-L286
> After adding the Riemannian optimizer there, you would also have to add an experiment that uses it.
> 
> @fangzhaoz If you have the opportunity, it would be great if you could review the PR too. If @smellslikeml agrees, I would also suggest to add you as a co-author on the final commit.

Agreed, thank you for your suggestions!
I've updated with changes and attribution

src/peft/optimizers/riemannian.py
```diff
@@ -0,0 +1,161 @@
...
+    Creates a Riemannian-preconditioned optimizer for a LoRA-adapted model.
+
+    Riemannian Preconditioned LoRA for Fine-Tuning Foundation Models: https://huggingface.co/papers/2402.02347
+
+    Reference: https://github.com/pilancilab/Riemannian_Preconditioned_LoRA
+
+    The returned optimizer behaves exactly like `optimizer_cls` (e.g. `torch.optim.AdamW` or `torch.optim.SGD`) except
+    that, on every `step`, the gradients of the LoRA `A` and `B` matrices are first multiplied by an `r x r` Riemannian
+    preconditioner. Non-LoRA parameters are updated unchanged.
```
BenjaminBossan, line 127: Let's document that this doesn't work for embedding layers, which can be targeted with LoRA. I haven't investigated if it would be possible to support embeddings too, it would require a bit of extra work because `lora_embedding_A` and `lora_embedding_B` use `nn.Parameter`, but it could be worth investigating.

method_comparison/MetaMathQA/experiments/lora/llama-3.2-3B-rank32-riemannian/adapter_config.json
```diff
@@ -0,0 +1,30 @@
+{
```
BenjaminBossan, line 1: Thanks for the experiment. I got 45% test accuracy, whereas default LoRA with rank 32 gets 49%. Beating normal LoRA is not crucial to get the PR merged, but maybe there is room for improvement.

BenjaminBossan (changes requested):
Thanks, this looks pretty good from my side, just a few comments.

@fangzhaoz Please let me know if you have time to review the PR or not.

method_comparison/MetaMathQA/experiments/lora/llama-3.2-3B-rank32-riemannian/adapter_config.json
smellslikeml: Thank you for the review!
I've updated with commit  01342440 to address your suggestions.

I also checked against the reference paper and implementation to find the parameter `reg` could be bumped to 1e-2. Re-testing, GSM8K accuracy increased from 0.45 to 0.475, closing some of the gap you found comparing to default LoRA.

smellslikeml:
@BenjaminBossan Just rebased onto main to clear the commit lag. Ready for re-review when you have a moment.

src/peft/optimizers/riemannian.py
```diff
@@ -0,0 +1,164 @@
...
+            "create_riemannian_optimizer did not find any trainable lora_A/lora_B parameter pairs on the model. "
+            "The Riemannian preconditioner only applies to LoRA-style adapters."
+        )
+
+    preconditioner = _RiemannianPreconditioner(lora_pairs, reg=reg)
+    trainable_params = [param for param in model.parameters() if param.requires_grad]
+
+    class RiemannianPreconditionedOptimizer(optimizer_cls):
+        @torch.no_grad()
+        def step(self, closure: Optional[Callable] = None):
```
BenjaminBossan, line 174: I'm just wondering: For something like LBFGS, which uses the closure, would the Riemanniannian preconditioner still make sense?

BenjaminBossan (approved):
Thanks for the updates, the PR generally looks good and is ready to be merged. Before doing that, I just have one question left.

smellslikeml:
> For something like LBFGS, which uses the closure, would the Riemanniannian preconditioner still make sense?

No, LBFGS builds curvature from `y = flat_grad - prev_flat_grad` and gates on the secant condition; if `P_k` is recomputed each step, `y` stops being a gradient difference of any fixed function.

While looking at this I noticed the ordering had a separate issue: `preconditioner.step()` then `super().step(closure)` silently drops preconditioning for any closure-using optimizer, since torch optimizers evaluate the closure at the top of their own step (typical body is `zero_grad(); loss.backward()`, which overwrites `.grad`). Affects AdamW/SGD-with-closure too.

Pushed both fixes in 58648c60:

- Wrap the closure so preconditioning runs inside it, matching `LoraFAOptimizer.step`'s convention.
- Guard closure-required optimizers at construction so we don't hardcode `LBFGS`:
  ```python
  closure_param = inspect.signature(optimizer_cls.step).parameters.get("closure")
  if closure_param is not None and closure_param.default is inspect.Parameter.empty:
      raise ValueError(...)
  ```
  Matches the sibling `TypeError` on non-Optimizer classes and the AdaLoRA-style raise on unsupported combos.

Regression tests added to `tests/test_riemannian_lora.py`: closure-vs-no-closure identity check (verified to fail on the previous ordering) and an LBFGS construction-time rejection test.

BenjaminBossan:
Thanks for investigating the closure case.

> While looking at this I noticed the ordering had a separate issue: `preconditioner.step()` then `super().step(closure)` silently drops preconditioning for any closure-using optimizer, since torch optimizers evaluate the closure at the top of their own step (typical body is `zero_grad(); loss.backward()`, which overwrites `.grad`). Affects AdamW/SGD-with-closure too.

I did not understand this point. Could you please give a concrete example (not a full reproducer, just a snippet would be enough)? And are there real world cases of "AdamW/SGD-with-closure"? If this is a very hypothetical scenario, I'd also be fine with simple checking if a closure was passed and raising.

smellslikeml:
```python
def closure():
    optimizer.zero_grad()
    loss = model(x).sum()
    loss.backward()
    return loss

optimizer.step(closure)
```

Old `step(closure)` did:
1. `preconditioner.step()` writes `P_k g_k` into `.grad`
2. `super().step(closure)` calls e.g. `AdamW.step(closure)`, which invokes `closure()` at the top, and `zero_grad(); backward()` overwrites `.grad` with raw `g_k`
3. `AdamW` reads raw `g_k`, preconditioning silently dropped

Fair, this is hypothetical for PEFT. LBFGS is the only shipping torch optimizer that requires a closure. Simplified per your suggestion in f90cc44e: replaced the wrapper with a step-time raise on any closure, dropped the wrapper's regression test, kept the construction-time signature guard for `LBFGS` (fail-fast on construction) alongside the step-time check.

BenjaminBossan (approved):
Thanks for the last push. Since we don't know what happens within the closure, I agree it's better to fully reject it.

githubnemo:
Hey, I just saw that this is missing documentation. @smellslikeml are you interested in providing a quick writeup for this in `package_reference/lora` under section Optimizers?

smellslikeml:
Hi @githubnemo,
Happy to, I'll open a doc PR adding a subsection for this option

PR #3257, opened 2026-05-23
Add UniLora tuner to PEFT

KaiyangLi1992:
## Motivation

This PR adds UniLora, a LoRA-style parameter-efficient fine-tuning method that reconstructs LoRA parameters from a shared low-dimensional vector.

This is a fresh PR replacing #2968, because GitHub could not reopen the old closed PR after the head branch was updated.

## What changed

- Add the UniLora tuner implementation, config, layer, and model integration.
- Register `PeftType.UNILORA` and expose `UniLoraConfig` / `UniLoraModel`.
- Add save/load handling for UniLora shared weights.
- Add optional `save_indices` support for generated index and scale tensors.
- Add package reference documentation.
- Add a quick UniLora fine-tuning example.
- Add a MetaMathQA experiment config.
- Add UniLora coverage in custom model, decoder, encoder-decoder, and dedicated UniLora tests.

## Notes for review

This branch is rebased on the latest `main` and incorporates the remaining review feedback from #2968, including naming, docs, tests, example, MetaMathQA config, deterministic index generation documentation, `init_weights`, and optional index saving.

## Tests

- `make style`
- `make quality`
- `PYTHONPATH=src python -m pytest tests/test_unilora.py`
- `PYTHONPATH=src python -m pytest --no-cov tests/test_custom_models.py -k UniLora`
- `PYTHONPATH=src python -m pytest --no-cov tests/test_decoder_models.py -k UniLora`
- `PYTHONPATH=src python -m pytest --no-cov tests/test_encoder_decoder_models.py -k UniLora`

PR #3383, opened 2026-07-01
Add UniLoraConfig to ALL_CONFIG_CLASSES test matrix

DaoyuanLi2816:
## What does this PR do?

`UniLoraConfig` (added in #3257) is missing from `ALL_CONFIG_CLASSES` in `tests/test_config.py`, so it isn't covered by the config test matrix (`to_dict` / `save_pretrained` / `from_pretrained` / `from_json_file`, copy/pickle, etc.).

This is the same kind of gap that #3365 addressed for `CPTConfig`/`RandLoraConfig` (and the four others I added there). `UniLoraConfig` landed afterwards without being registered. It's a public `PeftConfig` subclass with no mandatory init kwargs, so it's added as `(UniLoraConfig, {})`, consistent with the other entries.

I re-checked the full matrix against every public `PeftConfig`/`PromptLearningConfig` subclass — `UniLoraConfig` is the only one currently missing.

## Before submitting

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -62,6 +62,7 @@ from peft import (
     TaskType,
     TinyLoraConfig,
     TrainableTokensConfig,
+    UniLoraConfig,
     VBLoRAConfig,
     VeraConfig,
     WaveFTConfig,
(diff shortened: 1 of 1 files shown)


DaoyuanLi2816:
Closing this one for now — a single test-matrix gap isn't worth taking up review bandwidth on its own right now. Happy to revisit if it becomes relevant to other work.

PR #3386, opened 2026-07-02
Docs: Improve VeRA conceptual explanation and parameter scaling

SanjayMuthuswamy:
## Summary

This PR improves the VeRA documentation by adding a beginner-friendly conceptual explanation before the API reference.

It introduces:
- a short explanation of how VeRA differs from LoRA
- an overview of VeRA's parameter-scaling approach
- source-backed context from the original VeRA paper

## Motivation

The current documentation explains that VeRA is similar to LoRA and requires fewer trainable parameters, but it does not clearly explain why this is the case or what components are frozen versus trainable.

This change aims to make the documentation easier for new users to understand while preserving the existing technical content.

## Sources

- VeRA paper, Section 3.1 (method definition)
- VeRA paper, Section 3.2 and Table 1 (parameter scaling and storage comparison)

### DIFF
--- a/docs/source/package_reference/vera.md
+++ b/docs/source/package_reference/vera.md
@@ -18,6 +18,22 @@ rendered properly in your Markdown viewer.
 
 [VeRA](https://huggingface.co/papers/2310.11454) is a parameter-efficient fine-tuning technique that is similar to LoRA but requires even fewer extra parameters while promising similar or even better performance. As such, it is particularly useful when the parameter budget is very limited, e.g. when scaling to very large models. The reduction of the count of trainable parameters is achieved by sharing the same low-rank matrices across all layers, and only training two additional vectors per layer.
 
+## How VeRA works
+
+LoRA updates a frozen base model by learning two small, low-rank matrices for each adapted layer. The rank is the
+inner dimension of these matrices, and it controls how much capacity the adapter has. Increasing the rank or adapting
+more layers increases the number of trainable LoRA parameters because each layer receives its own pair of matrices.
+
+VeRA keeps the same low-rank adaptation idea but changes which parameters are trained. Instead of learning separate
+low-rank matrices for every layer, VeRA uses one frozen, randomly initialized pair of low-rank matrices, `A` and `B`,
+that is shared across layers. Each adapted layer only learns two scaling vectors, commonly written as `d` and `b`,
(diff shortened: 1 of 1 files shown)


docs/source/package_reference/vera.md
```diff
@@ -18,6 +18,22 @@ rendered properly in your Markdown viewer.
 
 [VeRA](https://huggingface.co/papers/2310.11454) is a parameter-efficient fine-tuning technique that is similar to LoRA but requires even fewer extra parameters while promising similar or even better performance. As such, it is particularly useful when the parameter budget is very limited, e.g. when scaling to very large models. The reduction of the count of trainable parameters is achieved by sharing the same low-rank matrices across all layers, and only training two additional vectors per layer.
 
+## How VeRA works
+
+LoRA updates a frozen base model by learning two small, low-rank matrices for each adapted layer. The rank is the
+inner dimension of these matrices, and it controls how much capacity the adapter has. Increasing the rank or adapting
+more layers increases the number of trainable LoRA parameters because each layer receives its own pair of matrices.
```
BenjaminBossan: > increases the number of trainable LoRA parameters

This is a bit ambiguous: Of course, with VeRA, increasing the rank also increases the number of trainable parameters, as `vera_lambda_d` depends on `r`. The difference is that for LoRA, the increase of the number of trainable parameters scales more quickly when `r` than for VeRA, as you explain below.

BenjaminBossan (changes requested):
Thanks for extending the VeRA documentation, generally this looks good. I just a small issue with a sentence that is ambiguous, please check.

docs/source/package_reference/vera.md
SanjayMuthuswamy: > > increases the number of trainable LoRA parameters
> 
> This is a bit ambiguous: Of course, with VeRA, increasing the rank also increases the number of trainable parameters, as `vera_lambda_d` depends on `r`. The difference is that for LoRA, the increase of the number of trainable parameters scales more quickly when `r` than for VeRA, as you explain below.

Thanks, good catch! I've updated the wording to clarify that increasing the rank increases the number of trainable parameters in both LoRA and VeRA, while LoRA's parameter count scales more rapidly with rank than VeRA's.

BenjaminBossan (approved):
Thanks for extending the VeRA docs, LGTM.

PR #3391, opened 2026-07-05
Fix inject_adapter_in_model not exposing AdaLora's update_and_allocate

### LINKED ISSUE #3373: [Bug] AdaLora's update_and_allocate method is lost in inject_adapter_in_model() preventing rank pruning
### System Info

peft version: 0.19.1
transformers version: 5.5.4
sentence-transformers version: 5.5.4
torch version: 2.8.0+cu128
Python version: 3.12.3

### Who can help?

@BenjaminBossan @githubnemo 

### Reproduction

When applying an AdaLora adapter via transformers' add_adapter() (which internally uses PeftAdapterMixin and calls peft.inject_adapter_in_model()), the update_and_allocate method becomes completely inaccessible.

We traced this down and realized that `peft.inject_adapter_in_model()` returns `peft_model.model` instead of the base `peft_model`. Since `peft_model` is the actual `AdaLoraModel` wrapper where `update_and_allocate` is implemented, returning just the inner `.model` strips the wrapper away and the method is lost from the namespace.

Here is the minimal reproducer using sentence-transformers (which relies on the transformers integration):

```python
from sentence_transformers import SentenceTransformer
from peft import AdaLoraConfig, TaskType, PeftType

model_name_or_path = 'Qwen/Qwen3-VL-Embedding-2B'
model = SentenceTransformer(
    model_name_or_path=model_name_or_path,
    model_kwargs={"torch_dtype": "float16"}
)       

adapter_name = 'adalora_adapter'
total_training_steps = 1000 # Placeholder for the actual dataset steps
target_r = 8
orth_reg_weight = 1e-4
lora_alpha = target_r * 2
init_r = int(target_r * 1.5)
tinit = int(total_training_steps * 0.1)  # 10% Warmup
tfinal = int(total_training_steps * 0.2) # 20% Stabilization (Pruning ends at 80%)

    peft_type=PeftType.ADALORA, 
    task_type=TaskType.FEATURE_EXTRACTION, 
    init_r=init_r, 
    target_r=target_r,
    lora_alpha=lora_alpha, 
    lora_dropout=0.05,

    # The phase schedule parameters:
    total_step=total_training_steps,
    tinit=tinit,
    tfinal=tfinal,
    deltaT=2, # Update the math every 2 steps

    orth_reg_weight=orth_reg_weight,
    target_modules=[
        "q_proj", "k_proj", "v_proj", "o_proj", 
        "gate_proj", "up_proj", "down_proj"
    ],
)

model.add_adapter(adapter_config=adalora_config, adapter_name=adapter_name)

# If we inspect the objects, the required method is missing:
print("update_and_allocate" in dir(model)) # Returns False
print("update_and_allocate" in dir(model[0].auto_model)) # Returns False
```
Here is a pure `transformers` example demonstrating the issue (no third-party libraries involved):

```python
from transformers import AutoModelForSeq2SeqLM 
from peft import AdaLoraConfig

config = AdaLoraConfig(
    peft_type="ADALORA", 
    task_type="SEQ_2_SEQ_LM", 
    init_r=12, 
    lora_alpha=32, 
    target_modules=["q", "v"],
    total_step=120,
    tinit=int(120*0.1),
    tfinal=int(120*0.2),
    lora_dropout=0.01,
)

model = AutoModelForSeq2SeqLM.from_pretrained("t5-base") 

# If we used the legacy get_peft_model(model, config), the method would exist.
# However, using the native add_adapter method causes the wrapper to drop:
model.add_adapter(config, adapter_name="adalora_adapter")

# The method is missing:
print(hasattr(model, "update_and_allocate")) # Returns False
```

### Expected behavior

I expect the `AdaLoraModel` wrapper (or specifically its methods like `update_and_allocate`) to be preserved or correctly routed when `inject_adapter_in_model` is used.

Without access to `update_and_allocate`, standard Trainer callbacks cannot trigger the dynamic rank allocation, causing AdaLora to stay stuck at `init_r`. This essentially breaks the core mathematical mechanism of AdaLora and degrades it to a standard LoRA.

shadowmodder:
## Summary

Fixes #3373.

After `inject_adapter_in_model()` returns, the caller holds the bare inner `torch.nn.Module`, not the `AdaLoraModel` tuner wrapper. `update_and_allocate` lives on `AdaLoraModel`, so it was silently dropped, and any training loop following the documented pattern:

```python
loss.backward()
model.update_and_allocate(global_step)
optimizer.zero_grad()
```

would raise `AttributeError: 'DummyModel' object has no attribute 'update_and_allocate'`.

**Root cause** — `inject_adapter_in_model` (`src/peft/mapping.py`) ends with:

```python
return peft_model.model   # tuner wrapper discarded here
```

**Fix** — after extracting the inner model, check whether the tuner exposes `update_and_allocate` and, if so, bind the bound method onto the returned model:

```python
returned_model = peft_model.model
if hasattr(peft_model, "update_and_allocate"):
    returned_model.update_and_allocate = peft_model.update_and_allocate
return returned_model
```

Because `peft_model.update_and_allocate` is already a bound method, `self` inside it remains the `AdaLoraModel` instance (with its `RankAllocator`, `peft_config`, and `trainable_adapter_name` intact). The bound method also keeps the tuner alive via a reference, so no state is lost.

The guard is written generically (`hasattr`) so it will apply to any future tuner that adds a similar training-time method without requiring changes to `inject_adapter_in_model` again.

## Changes

- `src/peft/mapping.py` — bind `update_and_allocate` onto the returned model when the tuner exposes it; add a `Note:` to the docstring documenting this behaviour for AdaLora users.
- `tests/test_low_level_api.py` — add `test_adalora_update_and_allocate_after_inject` to `TestLowLevelFunctional`: injects AdaLora into a `DummyModel`, asserts `update_and_allocate` is present and callable, then runs a complete forward → backward → `update_and_allocate(0)` cycle to verify end-to-end correctness.

## Test plan

- [x] New test `TestLowLevelFunctional::test_adalora_update_and_allocate_after_inject` passes
- [x] All existing `TestLowLevelFunctional` tests continue to pass (`test_inject_adapter_in_model`, `test_get_peft_model_state_dict`, `test_modules_to_save`)
- [x] The existing `TestInjectAdapterFromStateDict` parametrised test that passes `AdaLoraConfig(total_step=5)` through `inject_adapter_in_model` is unaffected

### DIFF
--- a/src/peft/mapping.py
+++ b/src/peft/mapping.py
@@ -73,6 +73,10 @@ def inject_adapter_in_model(
             This can be useful when the exact `target_modules` of the PEFT method is unknown, for instance because the
             checkpoint was created without meta data. Note that the values from the `state_dict` are not used, only the
             keys are used to determine the correct layers that should be adapted.
+
+    Note:
+        For AdaLora, the returned model will have an ``update_and_allocate`` method bound to it. This method must be
+        called after each backward pass and before ``optimizer.zero_grad()`` to dynamically reallocate rank budgets.
     """
     if peft_config.is_prompt_learning or peft_config.is_adaption_prompt:
         raise ValueError("`create_and_replace` does not support prompt learning and adaption prompt yet.")
@@ -89,4 +93,13 @@ def inject_adapter_in_model(
         model, peft_config, adapter_name=adapter_name, low_cpu_mem_usage=low_cpu_mem_usage, state_dict=state_dict
     )
 
-    return peft_model.model
+    returned_model = peft_model.model
+
+    # Some tuner types expose training-time methods that are only present on the tuner wrapper, not on the inner
+    # model. Bind those methods onto the returned model so callers do not need to hold a separate reference to the
+    # tuner. For example, AdaLoraModel.update_and_allocate must be called after each backward pass to dynamically
+    # reallocate rank budgets; without this binding the method is unreachable after inject_adapter_in_model returns.
+    if hasattr(peft_model, "update_and_allocate"):
+        returned_model.update_and_allocate = peft_model.update_and_allocate
+
+    return returned_model
--- a/tests/test_low_level_api.py
+++ b/tests/test_low_level_api.py
@@ -121,6 +121,32 @@ class TestLowLevelFunctional:
         assert hasattr(model.linear2, "weight")
         assert hasattr(model.linear2, "bias")
 
+    def test_adalora_update_and_allocate_after_inject(self):
+        # Regression test for https://github.com/huggingface/peft/issues/3373.
+        # After inject_adapter_in_model, AdaLora's update_and_allocate must be callable on the returned model.
+        model = DummyModel()
+
+        adalora_config = AdaLoraConfig(
+            target_modules=["linear"],
+            total_step=10,
+        )
+
+        model = inject_adapter_in_model(adalora_config, model)
+
+        assert hasattr(model, "update_and_allocate"), (
+            "update_and_allocate must be accessible on the model returned by inject_adapter_in_model "
+            "when using AdaLoraConfig"
+        )
+        assert callable(model.update_and_allocate)
+
+        # Verify the method actually executes: gradients must exist before calling update_and_allocate,
+        # matching the expected training loop ordering (loss.backward → update_and_allocate → zero_grad).
+        dummy_inputs = torch.LongTensor([[0, 1, 2, 3, 4, 5, 6, 7]])
+        output = model(dummy_inputs)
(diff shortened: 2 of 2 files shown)


BenjaminBossan:
@shadowmodder Thanks for the PR, however, it's not a good solution. First, reassigning methods like this can quickly become very confusing (who does `self` refer to?). Second, this creates a reference cycle, with `peft_model` not being cleaned up (and `peft_model` can hold torch tensors that hog memory). In the referenced issue, I already explained how to deal with this situation, so there is no need for an extra solution.

PR #3393, opened 2026-07-06
Make doc builds faster

mishig25:
Companion to the doc-builder tracker: huggingface/doc-builder#802

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -11,7 +11,7 @@ permissions: {}
 
 jobs:
    build:
-    uses: huggingface/doc-builder/.github/workflows/build_main_documentation.yml@e60a538eea9817ab312196d0d233604b01697265  # main (light installs, no custom container)
+    uses: huggingface/doc-builder/.github/workflows/build_main_documentation.yml@e60a538eea9817ab312196d0d233604b01697265  # main
     with:
       commit_sha: ${{ github.sha }}
       package: peft
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -11,12 +11,11 @@ permissions: {}
 
 jobs:
    build:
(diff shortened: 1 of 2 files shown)


.github/workflows/build_documentation.yml
```diff
@@ -11,12 +11,11 @@ permissions: {}
 
 jobs:
    build:
-    uses: huggingface/doc-builder/.github/workflows/build_main_documentation.yml@b0f9a6e3b6aa912656cbda9f74896eb721d29421  # main
+    uses: huggingface/doc-builder/.github/workflows/build_main_documentation.yml@e60a538eea9817ab312196d0d233604b01697265  # main (light installs, no custom container)
```
BenjaminBossan: I think the extra test after the `# main` comment will break dependabot. Let's remove it.

.github/workflows/build_pr_documentation.yml
```diff
@@ -11,9 +11,8 @@ permissions: {}
 
 jobs:
   build:
-    uses: huggingface/doc-builder/.github/workflows/build_pr_documentation.yml@b0f9a6e3b6aa912656cbda9f74896eb721d29421  # main
+    uses: huggingface/doc-builder/.github/workflows/build_pr_documentation.yml@e60a538eea9817ab312196d0d233604b01697265  # main (light installs, no custom container)
```
BenjaminBossan: Same.

BenjaminBossan (changes requested):
Thanks for making the doc build faster (it was already quite fast for PEFT, but even faster is even better). I just have a concern about the comment. I'm not 100% sure but let's be safe and remove it.

Failing tests are unrelated.

.github/workflows/build_documentation.yml
mishig25: Good catch — done. Reverted the comment to plain `# main` here and in all the sibling PRs (and opened follow-ups for transformers/huggingface_hub where it had already merged).

BenjaminBossan (approved):
LGTM. Dependabot would have updated by the end of the month, but this way we benefit earlier, thanks.

PR #3399, opened 2026-07-07
CI Fix error with Windows loading stable diffusion models

BenjaminBossan:
We have errors with the Windows CI when it tries to load the stable diffusion model:

https://github.com/huggingface/peft/actions/runs/28854159036/job/85576698333

This is coming from the `huggingface_hub 1.22` release. I could confirm that pinning to <1.22 resolved the issue. We already have a fix in #3394 due to this release but it didn't address the error above.

In this PR, the stable diffusion pipeline is loaded as part of a class level fixture. There is a tricky issue that setting it on `self` doesn't work for class fixtures, which is probably why it wasn't implemented like this in the first place.

The reason why this (presumably) solves the issue: According to the GLM analysis, this is a race condition that is aggravated on Windows due to how it handles symlinking. The old code is defined the model loading code on the class level, which means that when the tests are set up, it runs immediately. Since we run with 3 xdist processes, all 3 fire at approximately the same time. This makes it more likely to trigger the race condition. Using a fixture, the SD pipeline is only loaded once the first test on the class is reached. Since many tests run before that, it is very unlikely that all xdist processes start loading at the same time. Therefore, the race condition is resolved for all practical purposes.

I'm not 100% sure that this is really what happens, GLM 5.2 gave this conclusion:

> What actually changed in 1.22.0 and causes the Windows-specific failures:
> 1. snapshot_download now always calls api.list_repo_tree() instead of using repo_info.siblings
> In 1.21.0, snapshot_download got the file list from repo_info.siblings (already fetched by model_info()) and only fell back to list_repo_tree() for very large repos (> LARGE_REPO_THRESHOLD).
> In 1.22.0, snapshot_download always calls api.list_repo_tree() and writes the result to an on-disk tree cache (trees/<commit_hash>.json). This is an extra network call per snapshot_download invocation. With 3 concurrent xdist workers, that's 3 concurrent list_repo_tree() calls for the same repo. This adds latency and widens the concurrency window.
> 2. _get_metadata_or_catch_error can now skip the HEAD call for Xet files
> For Xet/LFS files (the safetensors), if the tree cache is on disk, the HEAD call is skipped entirely — metadata is rebuilt from the tree cache. For non-Xet files (config.json, etc.), the HEAD call still happens as before. This means the safetensors files download faster (no HEAD round-trip) while the config files still make a HEAD call. This changes the relative timing of when files complete downloading.
> 3. The offline path got the _raise_if_incomplete_snapshot check
> This is the known issue from the original investigation — offline mode now verifies completeness. This affects test_inject_from_state_dict_stable_diffusion (which uses hub_online_once) and the skipped tests.

> The real Windows-specific mechanism
> The symlink race in _create_symlink (os.remove → os.symlink) exists identically in both 1.21.0 and 1.22.0. The difference is that 1.22.0 widens the concurrency window through:
> - The extra list_repo_tree() call (adds network latency to each snapshot_download)
> - Faster safetensors downloads via tree cache (Xet files skip HEAD) but slower config downloads (still do HEAD)
> - This creates a longer period where multiple workers are simultaneously downloading and creating symlinks

> On Linux, os.symlink() on ext4 is fast enough that the race window remains tiny. On Windows, NTFS symlink creation is slower, and with the widened concurrency window, the race becomes reproducible.

As the fix resolves the error, the analysis is at least plausible.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,7 +47,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
-        os: ["ubuntu-latest", "windows-latest"]
+        os: ["windows-latest"]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,7 +40,7 @@ jobs:
           make quality
 
   tests:
-    #needs: check_code_quality
+    needs: check_code_quality
     # dependabot updates (which don't require approval for CI to run) shouldn't trigger unit tests
     if: ${{ !(github.event_name == 'pull_request' && github.event.pull_request.user.login == 'dependabot[bot]') }}
     strategy:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,7 +47,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
-        os: ["windows-latest"]
+        os: ["ubuntu-latest", "windows-latest"]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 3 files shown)


githubnemo (approved):
Thanks for the clarification regarding xdist - I assumed that the race is happening in huggingface_hub for which the fix didn't make sense.

LGTM.

PR #3401, opened 2026-07-07
Implements EWoRA in PEFT

### LINKED ISSUE #3400: Add EWoRA to PEFT
### Feature request

This is a second attempt (prev. issue #2970 was closed.)

Feature request
We propose to add EWoRA to the peft library -
https://aclanthology.org/2025.findings-ijcnlp.108/

This paper was accepted to the Findings of AACL-IJCNLP. It improves finetuning on a diverse downstream dataset through a mixture-of-experts approach.

Your contribution
I am able to create a PR for this.

### Your contribution

I am able to create a PR for this.

HarshKohli:
Resolves #3400


docs/source/package_reference/ewora.md
```diff
@@ -0,0 +1,70 @@
+<!--Copyright 2025 The HuggingFace Team. All rights reserved.
```
BenjaminBossan:
```suggestion
<!--Copyright 2026 The HuggingFace Team. All rights reserved.
```

Please use the current year here and throughout.

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+@dataclass
+class EworaConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`EworaModel`].
+
+    Args:
+        r (`int`):
+            Ewora expert ranks.
```
BenjaminBossan: Let's make it clear if it's the total rank of all experts (thus comparable to the LoRA rank) or the rank per expert (in which case `r * num_experts` is the correct comparison to LoRA). Also mention the defaults for all EWoRA specific parameters.

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,242 @@
...
+            "help": (
+                "List of module names or regex expression of the module names to replace with EWoRA."
+                "For example, ['q', 'v'] or '.*decoder.*(SelfAttention|EncDecAttention).*(q|v)$'."
+                "This can also be a wildcard 'all-linear' which matches all linear/Conv1D layers except the output layer."
+                "If not specified, modules will be chosen according to the model architecture, If the architecture is "
+                "not known, an error will be raised -- in this case, you should specify the target modules manually."
+            ),
+        },
+    )
+    num_experts: int = field(
```
BenjaminBossan, line 76: This is missing a docstring.

src/peft/tuners/ewora/model.py
```diff
@@ -0,0 +1,356 @@
...
+    **Attributes**:
+        - **model** ([`~transformers.PreTrainedModel`]) -- The model to be adapted.
+        - **peft_config** ([`EworaConfig`]): The configuration of the Ewora model.
+    """
+
+    prefix: str = "ewora_"
+    tuner_layer_cls = EworaLayer
+    target_module_mapping = TRANSFORMERS_MODELS_TO_EWORA_TARGET_MODULES_MAPPING
+
+    def __init__(self, model, config, adapter_name) -> None:
```
BenjaminBossan: Method not needed, pls remove

src/peft/tuners/ewora/model.py
```diff
@@ -0,0 +1,356 @@
...
+        super().__init__(model, config, adapter_name)
+
+    def _check_new_adapter_config(self, config: EworaConfig) -> None:
```
BenjaminBossan: Method not needed, pls remove

src/peft/tuners/ewora/model.py
```diff
@@ -0,0 +1,356 @@
...
+        # TODO: there should be a check if any of the existing adapters actually has bias != "none", or else the check
+        # does not fully correspond to the error message.
+        if (len(self.peft_config) > 1) and (config.bias != "none"):
+            raise ValueError(
+                f"{self.__class__.__name__} supports only 1 adapter with bias. When using multiple adapters, "
+                "set bias to 'none' for all adapters."
+            )
+
+    @staticmethod
+    def _check_target_module_exists(ewora_config, key):
```
BenjaminBossan: Method not needed, pls remove

src/peft/tuners/ewora/model.py
```diff
@@ -0,0 +1,356 @@
...
+        kwargs = {
+            "r": r,
+            "ewora_alpha": alpha,
+            "ewora_dropout": ewora_config.ewora_dropout,
+            "fan_in_fan_out": ewora_config.fan_in_fan_out,
+            "use_rslora": ewora_config.use_rslora,
+            "use_dora": ewora_config.use_dora,
+            "num_experts": ewora_config.num_experts,
+            "loaded_in_8bit": getattr(self.model, "is_loaded_in_8bit", False),
+            "loaded_in_4bit": getattr(self.model, "is_loaded_in_4bit", False),
```
BenjaminBossan: This uses a somewhat old pattern that we have updated in PEFT. Besides `r` and `alpha`, which can depend on the pattern, we don't need these arguments in `kwargs`. So let's remove them. Instead, pass the `EworaConfig` instance directly to `target.update_layer`. Update that method to ensure that it picks the parameters from the config. Check how other PEFT methods do it.

src/peft/tuners/ewora/model.py
```diff
@@ -0,0 +1,356 @@
...
+        }
+
+        quant_methods = ["gptq", "aqlm", "awq"]
```
BenjaminBossan: We don't have quant support, so this can be fully removed.

If you want EWoRA to support quantization, there is now a better pattern, which is laid out in #3117. I'll leave it up to you to decide if you want to support quantization or not.

src/peft/tuners/ewora/model.py
```diff
@@ -0,0 +1,356 @@
...
+            )
+        else:
+            new_module = self._create_new_module(ewora_config, adapter_name, target, **kwargs)
+            if adapter_name not in self.active_adapter:
+                # adding an additional adapter: it is not automatically trainable
+                new_module.requires_grad_(False)
+            self._replace_module(parent, target_name, new_module, target)
+
+    @staticmethod
+    def _replace_module(parent, child_name, new_module, child, **kwargs):
```
BenjaminBossan: AFAICT, this is method is not needed and we can just use the one from the parent class.

src/peft/tuners/ewora/model.py
```diff
@@ -0,0 +1,356 @@
...
+            else:
+                new_module.state = child.state
+            new_module.to(child.weight.device)
+
+        # dispatch to correct device
+        for name, module in new_module.named_modules():
+            if "ewora_" in name:
+                module.to(child.weight.device)
+
+    def _mark_only_adapters_as_trainable(self, model: nn.Module) -> None:
```
BenjaminBossan: Method not needed, pls remove

src/peft/tuners/ewora/model.py
```diff
@@ -0,0 +1,356 @@
...
+
+    @staticmethod
+    def _create_new_module(ewora_config, adapter_name, target, **kwargs):
+        if isinstance(target, BaseTunerLayer):
+            target_base_layer = target.get_base_layer()
+        else:
+            target_base_layer = target
+
+        if isinstance(target_base_layer, torch.nn.Linear):
+            if kwargs["fan_in_fan_out"]:
```
BenjaminBossan: Similar argument as above: `kwargs` is only needed for `r` and `alpha`. The remaining arguments are now directly picked from the config. Ensure that you pass the config to `Linear` below and to adjust `Linear.__init__`  accordingly.

src/peft/tuners/ewora/model.py
```diff
@@ -0,0 +1,356 @@
...
+            )
+        new_module = Linear(
+            target,
+            adapter_name,
+            **kwargs,
+        )
+
+        return new_module
+
+    # def __getattr__(self, name: str):
```
BenjaminBossan: I think all the methods below are not needed and can be removed too. If that's not the case, we will notice when tests fail.

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,226 @@
...
+from transformers.pytorch_utils import Conv1D
+
+from peft.tuners.tuners_utils import BaseTunerLayer
+
+
+class EworaLayer(BaseTunerLayer):
+    # All names of layers that may contain (trainable) adapter weights
+    adapter_layer_names = ("ewora_As", "ewora_Bs", "ewora_weighting")
+    # All names of other parameters that may contain adapter-related parameters
+    other_param_names = ("r", "ewora_dropout", "num_experts")
```
BenjaminBossan, line 32: Missing `ewora_alpha`.

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,242 @@
...
+    ewora_alpha: int = field(default=8, metadata={"help": "Ewora alpha"})
+    ewora_dropout: float = field(default=0.0, metadata={"help": "Ewora dropout"})
+    fan_in_fan_out: bool = field(
+        default=False,
+        metadata={"help": "Set this to True if the layer to replace stores weight like (fan_in, fan_out)"},
+    )
+    bias: Literal["none", "all", "ewora_only"] = field(
+        default="none", metadata={"help": "Bias type for Ewora. Can be 'none', 'all' or 'ewora_only'"}
+    )
+    use_rslora: bool = field(
```
BenjaminBossan: This parameter has no effect, please remove.

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,242 @@
...
+        default=8,
+        metadata={"help": ("Num Experts.")},
+    )
+    ewora_alpha: int = field(default=8, metadata={"help": "Ewora alpha"})
+    ewora_dropout: float = field(default=0.0, metadata={"help": "Ewora dropout"})
+    fan_in_fan_out: bool = field(
+        default=False,
+        metadata={"help": "Set this to True if the layer to replace stores weight like (fan_in, fan_out)"},
+    )
+    bias: Literal["none", "all", "ewora_only"] = field(
```
BenjaminBossan: I think we don't need `bias`, let's remove it.

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,242 @@
...
+    alpha_pattern: Optional[dict] = field(
+        default_factory=dict,
+        metadata={
+            "help": (
+                "The mapping from layer names or regexp expression to alphas which are different from the default alpha specified by `ewora_alpha`. "
+                "For example, `{model.decoder.layers.0.encoder_attn.k_proj: 32`}"
+            )
+        },
+    )
+    megatron_config: Optional[dict] = field(
```
BenjaminBossan: This parameter has no effect, please remove.

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,242 @@
...
+                "You need to specify this parameter when you want to apply EWoRA to the ColumnParallelLinear and "
+                "RowParallelLinear layers of megatron."
+                "It should be noted that we may not be able to use the `save_pretrained` and `from_pretrained` "
+                "functions, because TransformerConfig may not necessarily be serialized."
+                "But when using megatron, we can use `get_peft_model_state_dict` function and "
+                "megatron's framework, they can also save and load models and configurations."
+            )
+        },
+    )
+    megatron_core: Optional[str] = field(
```
BenjaminBossan: This parameter has no effect, please remove.

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,242 @@
...
+        default="megatron.core",
+        metadata={
+            "help": (
+                "The core module from Megatron, it is used to create EWoRA's parallel linear layer. "
+                "It only needs to be passed in when you need to use your own modified megatron core module. "
+                "Otherwise, it will use the default value `megatron.core`. "
+            )
+        },
+    )
+    use_dora: bool = field(
```
BenjaminBossan: This parameter has no effect, please remove.

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,242 @@
...
+                "weights into two parts, magnitude and direction. Direction is handled by normal EWoRA, whereas the "
+                "magnitude is handled by a separate learnable parameter. This can improve the performance of LoRA, "
+                "especially at low ranks. Right now, DoRA only supports linear and Conv2D layers. DoRA introduces a bigger"
+                "overhead than pure LoRA, so it is recommended to merge weights for inference. For more information, "
+                "see  https://arxiv.org/abs/2402.09353."
+            )
+        },
+    )
+    # Enables replicating layers in a model to expand it to a larger model.
+    layer_replication: Optional[list[tuple[int, int]]] = field(
```
BenjaminBossan: This parameter has no effect, please remove.

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,242 @@
...
+                "the adapter layers fit during fine tuning."
+                "The format is a list of [start, end) pairs which specify the layer ranges to stack. For example:\n"
+                "   Original model has 5 layers labelled by their position in the model: `[0, 1, 2, 3, 4]`\n"
+                "   layer_replication: `[[0, 4], [2, 5]]`\n"
+                "   Final model will have this arrangement of original layers: `[0, 1, 2, 3, 2, 3, 4]`\n"
+                "This format is based on what is used for pass-through merges in mergekit. It makes it simple to select sequential "
+                "ranges of a model and stack them while reusing layers at either end of each sequence."
+            )
+        },
+    )
```
BenjaminBossan: There is no argument `init_weights: bool = True`. I think that would be important to have. By default (`True`), EWoRA weights should be initialized as they are right now, with B's zeroed out so that we have an identity transform at the beginning of training. If `False` is passed, B should be randomly initialized. This is important for many tests, where we need to ensure that the PEFT method has an effect on the output.

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,242 @@
...
+            strings, either an exact match will be performed or it is checked if the name of the module ends with any
+            of the passed strings. If this is specified as 'all-linear', then all linear/Conv1D modules are chosen,
+            excluding the output layer. If this is not specified, modules will be chosen according to the model
+            architecture. If the architecture is not known, an error will be raised -- in this case, you should specify
+            the target modules manually.
+        ewora_alpha (`int`):
+            The alpha parameter for Ewora scaling.
+        ewora_dropout (`float`):
+            The dropout probability for Ewora layers.
+        fan_in_fan_out (`bool`):
```
BenjaminBossan: AFAICT, `fan_in_fan_out` is not being used. The purpose of this argument is to support merging with Transformers `Conv1D` layer, whose weights are transposed compared to `nn.Linear`. Since EWoRA doesn't allow merging, I think this can be removed completely.

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,226 @@
...
+
+        self.set_adapter(self.active_adapters)
+
+        self._move_adapter_to_device_of_base_layer(adapter_name)
+        self.set_adapter(self.active_adapters)
+
+    def reset_ewora_parameters(self, adapter_name):
+        if adapter_name in self.ewora_As.keys():
+            # initialize A the same way as the default for nn.Linear and B to zero
+            # https://github.com/microsoft/LoRA/blob/a0a92e0f26c067cf94747bdbf1ce73793fa44d19/loralib/layers.py#L124
```
BenjaminBossan: Doesn't make sense to reference the LoRA implementation IMO.

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,226 @@
...
+                if active_adapter not in self.ewora_As.keys():
+                    continue
+
+                ewora_As = self.ewora_As[active_adapter]
+                ewora_Bs = self.ewora_Bs[active_adapter]
+                dropout = self.ewora_dropout[active_adapter]
+                weighting = self.ewora_weighting[active_adapter]
+                num_experts = self.num_experts[active_adapter]
+
+                bs, seq_len, _ = x.size()
```
BenjaminBossan: What if we have dim == 2? It should still work, we would just need to skip the reshaping, right? Checking dim > 3 is probably not needed, since we're targeting linear layers.

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,226 @@
...
+                x = x.to(ewora_As.dtype)
+                x = x.unsqueeze(2).expand(-1, -1, num_experts, -1)
+
+                intermediate = torch.einsum("beid, idj -> beij", dropout(x), ewora_As)
```
BenjaminBossan: Let's add a short comment to explain the indices b, d, e, i, j

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,226 @@
...
+
+                scores = weighting(F.relu(intermediate.reshape(bs, seq_len, -1)))
+                # scores = F.softmax(weighting(F.relu(intermediate.reshape(bs, seq_len, -1))))
+                # scores = weighting(intermediate.reshape(bs, seq_len, -1))
```
BenjaminBossan: Remove?

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,226 @@
...
+
+                final = torch.einsum("beij, ijk -> beik", intermediate, ewora_Bs)
+                del intermediate
+
+                final = final * scores.unsqueeze(-1)
+                result.add_(final.sum(dim=2))
```
BenjaminBossan: Is this better than `result = result + final.sum(dim=2)`?

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,226 @@
...
+
+                del final, scores
```
BenjaminBossan: Regarding the manual deletions, did you measure that this has an effect? Normally, we don't have those calls in PEFT, it would only make sense if we find that they're required for proper memory management.

tests/test_ewora.py
BenjaminBossan, line 1: This whole test file can be deleted. We already have an extensive test suite, and once EWoRA is added there, the test coverage will be pretty much complete. We don't have tests that check, for instance, if `assert lin0.ewora_As["default"].shape == (num_experts, 10, r) `, but that's really not needed, as this is basically checking for implementation details

BenjaminBossan (changes requested):
Thanks for reviving the EWoRA PR. For this review, I focused exclusively on the PEFT integration itself, keeping the rest for later.

From what I can tell, you probably started this PR based on a somewhat older version of PEFT and taking the LoRA code as a starting point. For this reason, numerous additions are not needed or are out of date. I marked the corresponding code. The good news is that we now require much less code, so overall this should lead to a simplification.

To get the majority of test cases covered, please start by adding EWoRA to the test matrix of our [custom model tests](https://github.com/huggingface/peft/blob/7d4fb44318385bd782f49fa30b7fd1ed6e4cc1c2/tests/test_custom_models.py#L1817). Since EWoRA does not allow merging, add it to [this function](https://github.com/huggingface/peft/blob/7d4fb44318385bd782f49fa30b7fd1ed6e4cc1c2/tests/testing_common.py#L78) too. Then run `pytest tests/test_custom_models.py -k "ewora" -v` and check if all tests pass.

Also, notice that we have reworked our [contribution guideline](https://huggingface.co/docs/peft/main/en/developer_guides/contributing#add-a-new-peft-fine-tuning-method), please check it out. For now, we're at the "Core integration" step, but to merge, we will need to cover the "Final PR" step too. From my perspective, it's, however, okay to finish step one first before proceeding.

HarshKohli:
@BenjaminBossan - thank you for reviewing my pull request and for the feedback. I have done a re-request to review - hope I have addressed most of your concerns. Please advise on next steps.

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,229 @@
...
+        layers_to_transform (`Union[List[int], int]`):
+            The layer indices to transform. If a list of ints is passed, it will apply the adapter to the layer indices
+            that are specified in this list. If a single integer is passed, it will apply the transformations on the
+            layer at this index.
+        layers_pattern (`str`):
+            The layer pattern name, used only if `layers_to_transform` is different from `None`.
+        rank_pattern (`dict`):
+            The mapping from layer names or regexp expression to ranks which are different from the default rank
+            specified by `r`.
+        megatron_config (`Optional[dict]`):
```
BenjaminBossan: `megatron_config`, `megatron_core`, `use_dora`, and `layer_replication` are not being used and are just leftovers from starting EWoRA based on LoRA. Please remove these arguments.

src/peft/tuners/ewora/config.py
BenjaminBossan: Still open

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,229 @@
...
+        ewora_dropout (`float`):
+            The dropout probability for Ewora layers (defaults to 0.0).
+        init_weights (`bool`):
+            Whether to initialize the adapter as an identity transform, i.e. with the expert `B` matrices set to zero
+            (defaults to `True`). Setting this to `False` gives a non-identity initialization and is mainly useful for
+            testing.
+        fan_in_fan_out (`bool`):
+            Set this to True if the layer to replace stores weight like (fan_in, fan_out). For example, gpt-2 uses
+            `Conv1D` which stores weights like (fan_in, fan_out) and hence this should be set to `True`.
+        bias (`str`):
```
BenjaminBossan: This is not a very useful option to have, let's simplify our lives and just remove it.

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,229 @@
...
+
+@dataclass
+class EworaConfig(PeftConfig):
+    """
+    This is the configuration class to store the configuration of a [`EworaModel`].
+
+    Args:
+        r (`int`):
+            The rank of each individual expert adapter (defaults to 256). EWoRA partitions the adaptation into
+            `num_experts` independent rank-`r` experts, so the LoRA-equivalent total rank is `r * num_experts`.
```
BenjaminBossan, line 32: So with the default settings, we basically get 8 (`num_experts`) LoRA adapters with rank 256, do I see that right? That is quite a huge default, compared to LoRA, which is just rank 8 by default. Does it really make sense to pick such a high number?

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,196 @@
...
+            in_features, out_features = base_layer.in_features, base_layer.out_features
+        elif hasattr(base_layer, "w_bit") and base_layer.__class__.__name__ == "WQLinear_GEMM":
+            # Awq layers
+            in_features, out_features = base_layer.in_features, base_layer.out_features
+        elif base_layer.__class__.__name__ == "EetqLinear":
+            # Eetq layers
+            in_features, out_features = base_layer.in_features, base_layer.out_features
+        elif hasattr(base_layer, "W_q") and base_layer.__class__.__name__ == "HQQLinear":
+            # HQQ layers
+            in_features, out_features = base_layer.in_features, base_layer.out_features
```
BenjaminBossan: We have a utility function now that can simplify the pattern: `peft.tuners.utils._get_in_out_features`.

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,196 @@
...
+
+        num_experts = config.num_experts
+        self.r[adapter_name] = r
+        self.num_experts[adapter_name] = num_experts
+
+        if config.ewora_dropout > 0.0:
+            ewora_dropout_layer = nn.Dropout(p=config.ewora_dropout)
+        else:
+            ewora_dropout_layer = nn.Identity()
+        self.ewora_dropout.update(nn.ModuleDict({adapter_name: ewora_dropout_layer}))
```
BenjaminBossan: Simpler:
```suggestion
        self.ewora_dropout[adapter_name] = ewora_dropout_layer
```

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,196 @@
...
+    ) -> None:
+        # this gets the init from nn.Linear's super perspective, i.e. nn.Module.__init__, which should always be called
+        super(nn.Linear, self).__init__()
+        EworaLayer.__init__(self, base_layer, **kwargs)
+        self.fan_in_fan_out = config.fan_in_fan_out
+        self._active_adapter = adapter_name
+        self.update_layer(adapter_name, r, config)
+        self.is_target_conv_1d_layer = is_target_conv_1d_layer
+
+    def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
```
BenjaminBossan, line 126: AFAICT, there is no scaling of the LoRA outputs, i.e. something that corresponds to `alpha`, is that intended? In the paper, you mention that "LoRA Scaling Parameter (α): 32" is "used in all our training
experiments" though it's not quite clear if that also applies to EWoRA in this case. But since it's an important hyper-parameter for LoRA, I would expect it to matter here too.

BenjaminBossan (changes requested):
Thank you for the updates and for bringing the PR up-to-speed with the latest PEFT. In this review, I focused on the integration, which is not missing much; I haven't checked the docs or ran the experiments yet.

As the core integration is almost done, please consult our [contribution guideline](https://huggingface.co/docs/peft/main/en/developer_guides/contributing#add-a-new-peft-fine-tuning-method) for the next steps.

src/peft/tuners/ewora/layer.py
HarshKohli: Yes, this is intended. In EWoRA, the weights corresponding to each adapter (similar to a scaling factor) are assigned dynamically for each input (input-dependent scaling) . There is no hard-coded scaling parameter here

src/peft/tuners/ewora/layer.py
HarshKohli: Done.

src/peft/tuners/ewora/layer.py
HarshKohli: Done. I have replaced the entire isinstance chain.

src/peft/tuners/ewora/config.py
HarshKohli: Thank you for pointing this out. Replaced with rank (r) = 8 and num_experts=4 which are also the defaults we used in our paper.

src/peft/tuners/ewora/config.py
HarshKohli: Removed this.

src/peft/tuners/ewora/config.py
HarshKohli: Removed all of these

src/peft/tuners/ewora/layer.py
HarshKohli: Added a code comment explaining this as well.

src/peft/tuners/ewora/config.py
HarshKohli: Removed the fan_in_fan_out . Conv1D remains supported in the forward path since no transpose is needed there

HarshKohli:
@BenjaminBossan - thank you for the second round of comments. I have attempted to address them all and the branch is synced with main.

tests/testing_common.py
```diff
@@ -1217,6 +1217,11 @@ def _test_training_gradient_checkpointing(self, model_id, config_cls, config_kwa
         if (config_cls == AdaLoraConfig) and ("roberta" in model_id.lower()):
             # TODO: no gradients on the "dense" layer, other layers work, not sure why
             pytest.skip("AdaLora with RoBERTa does not work correctly")
+        if config_cls == EworaConfig:
+            # EWoRA works with gradient checkpointing, but this test is flaky for it: the ReLU gating in EWoRA's
+            # routing can produce exactly-zero grads for some routing params, and since the base model's dropout
+            # draws different masks for the normal vs the checkpointed pass, the two non-zero-grad sets can differ.
```
BenjaminBossan: Hmm, this sounds to me like GC is not actually working as intended? The RNG applied to the forward pass should be recorded for GC so that it can be replied exactly. If the router produces different scores during replay, it will corrupt the gradient, will it not?

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,167 @@
...
+        in_features, out_features = _get_in_out_features(base_layer)
+        if in_features is None:
+            raise ValueError(f"Unsupported layer type {type(base_layer)}")
+
+        self.in_features = in_features
+        self.out_features = out_features
+        self.kwargs = kwargs
+
+    @property
+    def merged(self) -> bool:
```
BenjaminBossan: Can be removed.

tests/test_custom_models.py
```diff
@@ -756,6 +757,13 @@ def _reset_unilora_theta_d(model, config, adapter_name="default"):
         ShiraConfig,
         {"r": 1, "target_modules": ["lin0"]},
     ),
+    #########
+    # EWoRA #
+    #########
+    ("Vanilla MLP 1 EWoRA", "MLP", EworaConfig, {"r": 2, "num_experts": 2, "target_modules": "lin0"}),
+    ("Vanilla MLP 2 EWoRA", "MLP", EworaConfig, {"r": 2, "num_experts": 2, "target_modules": ["lin0"]}),
+    ("Vanilla MLP 3 EWoRA", "MLP", EworaConfig, {"r": 2, "num_experts": 2, "target_modules": ["lin1"]}),
+    ("Vanilla MLP 4 EWoRA", "MLP", EworaConfig, {"r": 2, "num_experts": 4, "target_modules": ["lin0", "lin1"]}),
```
BenjaminBossan, line 766: EWoRA lso needs add to `MULTIPLE_ACTIVE_ADAPTERS_TEST_CASES`: https://github.com/huggingface/peft/blob/13414e66235139fad4102dfd3c9502958dbdc92b/tests/test_custom_models.py#L1375

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,167 @@
...
+                num_experts = self.num_experts[active_adapter]
+
+                x = x.to(ewora_As.dtype)
+                # broadcast the input over the experts: (..., in_features) -> (..., num_experts, in_features)
+                x = x.unsqueeze(-2).expand(*x.shape[:-1], num_experts, x.shape[-1])
+
+                # einsum indices: ... = leading dims (batch, sequence, ...), i = expert,
+                # d = in_features, j = rank, k = out_features
+                intermediate = torch.einsum("...id, idj -> ...ij", dropout(x), ewora_As)
+                scores = weighting(F.relu(intermediate.reshape(*intermediate.shape[:-2], -1)))
```
BenjaminBossan, line 159: Unless I'm missing something, the paper doesn't mention relu. Worth a comment?

src/peft/tuners/ewora/layer.py
```diff
@@ -0,0 +1,167 @@
...
+        if config.ewora_dropout > 0.0:
+            ewora_dropout_layer = nn.Dropout(p=config.ewora_dropout)
+        else:
+            ewora_dropout_layer = nn.Identity()
+        self.ewora_dropout[adapter_name] = ewora_dropout_layer
+
+        # Actual trainable parameters
+        self.ewora_As[adapter_name] = nn.Parameter(torch.Tensor(num_experts, self.in_features, r), requires_grad=True)
+        self.ewora_Bs[adapter_name] = nn.Parameter(torch.Tensor(num_experts, r, self.out_features), requires_grad=True)
+        self.ewora_weighting[adapter_name] = nn.Linear(r * num_experts, num_experts, bias=True)
```
BenjaminBossan, line 74: Do we need the bias here? It's not mentioned in the paper AFAICT. If we need it, does it need to be rescaled in `reset_ewora_parameters` in the same way as the weights?

src/peft/tuners/ewora/config.py
```diff
@@ -0,0 +1,134 @@
...
+        self.target_modules = (
+            set(self.target_modules) if isinstance(self.target_modules, list) else self.target_modules
+        )
+        # if target_modules is a regex expression, then layers_to_transform should be None
+        if isinstance(self.target_modules, str) and self.layers_to_transform is not None:
+            raise ValueError("`layers_to_transform` cannot be used when `target_modules` is a str.")
+
+        # if target_modules is a regex expression, then layers_pattern should be None
+        if isinstance(self.target_modules, str) and self.layers_pattern is not None:
+            raise ValueError("`layers_pattern` cannot be used when `target_modules` is a str.")
```
BenjaminBossan, line 134: Minor: Let's also check that `num_experts>1` (or >= 1).

examples/ewora_finetuning/ewora_finetuning.py
```diff
@@ -0,0 +1,152 @@
...
+    eos_token = "</s>"
+
+    if "instruction" in sample and sample["instruction"] is not None:
+        instruction = sample["instruction"]
+    elif "query" in sample and sample["query"] is not None:
+        instruction = sample["query"]
+    elif "ctx" in sample and sample["ctx"] is not None:
+        instruction = sample["ctx"]
+    elif "sentence" in sample and sample["sentence"] is not None:
+        instruction = sample["sentence"]
```
BenjaminBossan, line 29: Let's raise an error for `else`. Same below if `response` can't be determined.

BenjaminBossan (changes requested):
Thanks for the updates, progress looks good.

For this review, I focused again mostly on the integration, once all kinks there are ironed out, I'll check the examples and docs more closely. Kimi K3 assisted with the check, but all comments are mine.

For the finished PR, it would also be great to add two experiments, one for the [MetaMathQA](https://github.com/huggingface/peft/tree/main/method_comparison/MetaMathQA) and one for the [image-gen](https://github.com/huggingface/peft/tree/main/method_comparison/image-gen) benchmark. This is also a great way to double check that the performance of EWoRA is on par with what's expected. I wouldn't expect record breaking results, as the benchmarks don't quite fit EWoRA's use case, but at least I'd expect results that are similar to an equally sized LoRA adapter, right?

tests/testing_common.py
HarshKohli: I checked to make sure - GC replay is exact for EWoRA. Seeding the RNG identically before a normal backward and a checkpointed backward gives bit-identical gradients (max abs diff 0.0, for both use_reentrant=True and False), so the recorded RNG state does replay the router scores exactly and nothing is corrupted.

The flakiness is the test is comparing the non-zero-grad sets of two independent forward passes (the plain one and the GC-enabled one). In train mode these draw different base-model dropout masks. EWoRA can produce exactly-zero grads for a module's params when that module's ReLU-gated routing input happens to be all-dead (scores become exactly 0 with the zero-init bias). Whether that happens can flip between the two passes. As a control, I ran the same comparison between two plain passes with GC never enabled. The mismatch occurs at the same rate (3/40 trials on tiny-random-RobertaModel with the test's inputs vs 3/40 for plain-vs-GC). So the instability is dropout × ReLU in the set comparison, not gradient checkpointing.

I've reworded the skip comment to state this precisely. Happy to share the repro script or handle it differently (e.g. an EWoRA-specific tolerance) if you prefer.

examples/ewora_finetuning/ewora_finetuning.py
HarshKohli: Done

src/peft/tuners/ewora/config.py
HarshKohli: Done

src/peft/tuners/ewora/layer.py
HarshKohli: We had the bias in the implementation that led to the paper's results. I have kept it, and it's now explicitly zero-initialized in reset_ewora_parameters.

src/peft/tuners/ewora/layer.py
HarshKohli: This was part of the reference implementation - added a comment noting this.

tests/test_custom_models.py
HarshKohli: Done. Had to make some changes so that each adapter now builds its own expanded input.

src/peft/tuners/ewora/layer.py
HarshKohli: Done.

BenjaminBossan:
@HarshKohli There is a merge conflict, could you please check? Once the PR is ready for review, please ping me.

HarshKohli:
@BenjaminBossan - I have resolved the conflict and also added the MetaMathQA and image-gen experiments .

tests/testing_common.py
```diff
@@ -1242,6 +1245,15 @@ def _test_training_gradient_checkpointing(self, model_id, config_cls, config_kwa
...
             pytest.skip("AdaLora with RoBERTa does not work correctly")
+        if config_cls == EworaConfig:
+            # Gradient checkpointing itself works correctly for EWoRA: with identical RNG seeding, the grads with
+            # and without GC are bit-identical (the recorded RNG state makes the recomputed forward, incl. the
+            # router scores, exact). This test is nevertheless flaky for EWoRA because it compares the non-zero-grad
+            # sets of two *independent* forward passes, which draw different base-model dropout masks (train mode).
+            # If a module's ReLU-gated routing goes all-dead in only one of the two passes, that module's EWoRA
+            # params have exactly-zero grads in that pass only, so the two sets differ. The same mismatch occurs at
+            # the same rate between two plain passes without any GC.
+            pytest.skip("The non-zero-grad set comparison across independent forward passes is flaky for EWoRA.")
```
BenjaminBossan, line 1256: I think this is no longer true after #3599. Could you please check if we still need the skip?

HarshKohli:
@BenjaminBossan - just checking in once again.

docs/source/package_reference/ewora.md
```diff
@@ -0,0 +1,85 @@
...
+peft_model = get_peft_model(base_model, config)
+peft_model.print_trainable_parameters()
+```
+
+The abstract from the paper is:
+
+> Low-Rank Adaptation (LoRA) has emerged as a widely adopted parameter-efficient fine-tuning (PEFT) approach for language models. By restricting weight updates to a low-rank subspace, LoRA achieves cost-effective finetuning of large, generalist models to more specialized target domains. While LoRA achieves impressive results for a variety of individual downstream tasks, it struggles to capture the diverse expertise needed when presented with a more heterogeneous finetuning corpus. To address this, we propose Expert Weighted Low-Rank Adaptation (EWoRA), a novel LoRA variant that partitions a rank-r adapter into n independent adapters of rank r/n. A lightweight "routing" matrix (W<sub>r</sub> ∈ R<sup>r×n</sup>) aggregates the outputs of these adapters by learning specialized weights for each context. Experiments show EWoRA improves performance over LoRA when finetuning on heterogeneous data while generally matching or exceeding LoRA performance on individual finetuning tasks under the same low-rank parameter budget.
+
```
BenjaminBossan, line 44: Let's add a section that embeds the EWoRA benchmark results:

```
## Benchmark overview

<iframe
	src="https://peft-internal-testing-peft-method-comparison-embed.hf.space/?highlight[type]=EWORA"
	frameborder="0"
	width="850"
	height="1000"
></iframe>
```

It's currently not showing the EWoRA data points because it's not merged and tested yet, but we can still put the iFrame here for the future.

method_comparison/image-gen/results/ewora--flux2-klein-default.json
BenjaminBossan, line 1: Let's not include the result file for the experiement (same for MetaMath). We will run this experiment on our own hardware so that the results are comparable.

method_comparison/MetaMathQA/results/ewora--llama-3.2-3B-default.json
```diff
@@ -0,0 +1,356 @@
...
+      "modules_to_save": null,
+      "layers_to_transform": null,
+      "layers_pattern": null,
+      "rank_pattern": {}
+    },
+    "error_msg": ""
+  },
+  "train_info": {
+    "accelerator_memory_reserved_avg": 9818900476,
+    "accelerator_memory_max": 12736004096,
```
BenjaminBossan, line 61: I'm very curious about the low memory usage you report here. This value is much lower than anything we've tested on PEFT yet. When I run this EWoRA experiment on my machine (with a 4090, 24 GB), it OOMs one third of the way through the training. I found this curious, as it seems to pass on your 16 GB GPU.

My "mem reserved" during training is ~16.8 GB, "mem allocated" 6.9 GB. So interestingly, the allocated memory for you is pretty similar, but reserved memory much lower. I checked the logged software and generally the one I use is more up to date (e.g. CUDA 13). So I'm not sure how this is could be explained. If you have any ideas, I'd be glad to hear them.

examples/ewora_finetuning/ewora_finetuning.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def train_model(args):
+    os.environ["TOKENIZERS_PARALLELISM"] = "false"
+
+    # Load and concatenate the training datasets (simple concatenation, shuffled at train time)
+    train_datasets = []
+    for dataset_name in args.datasets.split(","):
+        if dataset_name == "allenai/winogrande":
+            dataset = load_dataset(dataset_name, "winogrande_xl", trust_remote_code=True)["train"]
```
BenjaminBossan, line 60: Not needed, right?

```suggestion
            dataset = load_dataset(dataset_name, "winogrande_xl")["train"]
```

examples/ewora_finetuning/ewora_finetuning.py
```diff
@@ -0,0 +1,156 @@
...
+        else:
+            dataset = load_dataset(dataset_name, trust_remote_code=True)["train"]
```
BenjaminBossan, line 62: Let's never enable this by default. If it's needed for a specific dataset, it should be opt-in.

```suggestion
            dataset = load_dataset(dataset_name)["train"]
```

examples/ewora_finetuning/ewora_finetuning.py
```diff
@@ -0,0 +1,156 @@
...
+        output_dir=args.output_dir,
+        num_train_epochs=args.num_epochs,
+        per_device_train_batch_size=args.batch_size,
+        logging_steps=50,
+        save_strategy="epoch",
+        optim="adamw_torch",
+        max_length=args.max_seq_length,
+        packing=True,
+        max_grad_norm=args.max_grad_norm,
+        warmup_ratio=args.warmup_ratio,
```
BenjaminBossan, line 99: I get:

> TypeError: SFTConfig.__init__() got an unexpected keyword argument 'warmup_ratio'

Is it possible that you used an old trl version? Consider changing this to `warmup_steps`. You should also mention somewhere what trl version is expected, since trl not a PEFT dependency.

examples/ewora_finetuning/ewora_finetuning.py
BenjaminBossan, line 1: Except for the issues mentioned below, I could manage to make training run with a smaller model.

BenjaminBossan (changes requested):
Thanks for the updates. For this review, I focused mainly on the docs, examples, and experiments. Please check my comments.

PR #3406, opened 2026-07-08
DOC Expand contributing installation guide with fork and setup steps

Strongich:
## What this does

- Rewrites the **Installation** section of the contributing guide so a new contributor can go from zero to an editable dev install without leaving the page. Adds explicit fork -> clone -> add-upstream -> branch -> `pip install -e ".[test]"`
steps, and removes the old `../install#source` link.

- Updates `docs/source/install.md`: quotes `pip install -e ".[test]"` for the same zsh reason, and points would-be contributors from the **Source** section to the contributing guide (dropping the half-covered "contributing" wording there so the two pages don't overlap).

## Motivation

I was about to fork the repo to start working on a new PEFT method, and the first thing I hit was the installation link in the contributing guide leading to a **404** on GitHub: `https://github.com/huggingface/peft/tree/install#source`.

The reason is that `../install#source` is a doc-builder relative link that only resolves on the rendered docs site -- on GitHub  it points at a nonexistent repo path. After finding the real page myself (https://huggingface.co/docs/peft/install#source), I realized a few steps that would have eased my onboarding were missing (forking, adding the upstream remote, syncing `main`, creating a work branch). So I adapted the
established [TRL](https://github.com/huggingface/trl?tab=contributing-ov-file#submitting-a-pull-request-pr) contributing-guide structure for PEFT to make the section self-contained regardless of where it's viewed.

I quoted `pip install -e ".[test]"` so it doesn't fail under zsh and works for bash (just a little more convenient way instead of hitting `zsh: no matches found: .[test]` and manually adding quotes).

## Tests

Docs-only change -- no unit tests required or run.

## Notes

AI assistance was used while preparing this change to polish the grammar. I reviewed every line and can explain and defend it.

### DIFF
--- a/docs/source/developer_guides/contributing.md
+++ b/docs/source/developer_guides/contributing.md
@@ -20,7 +20,43 @@ We are happy to accept contributions to PEFT. If you plan to contribute, please 
 
 ## Installation
 
-For code contributions to PEFT, you should choose the ["source"](../install#source) installation method.
+Follow these steps to start contributing:
+
+1. Fork the [repository](https://github.com/huggingface/peft) by clicking on the 'Fork' button on the repository's page. This creates a copy of the code under your GitHub user account.
+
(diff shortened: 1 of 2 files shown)


BenjaminBossan (approved):
Thanks for expanding the installation instructions, LGTM.

PR #3407, opened 2026-07-08
FIX Pin huggingface_hub to <1.22

BenjaminBossan:
See internal Slack discussion, #3394, #3399, #3404.

There are several issues since the huggingface_hub 1.22 release with loading Diffusers models on Windows. For the time being, we need to pin the version to < 1.22 and wait if/how this is resolved in hfh and Diffusers.

Reverting #3394, as it's not needed with hfh < 1.22.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,6 +35,9 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install .[dev]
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
Tests are green again, the original was probably fixed in huggingface_hub and this PR is no longer needed.

PR #3412, opened 2026-07-09
FIX DEFT's cached merge factor: precision loss and device staleness

DaoyuanLi2816:
## What does this PR do?

DEFT's `merge()` caches only the small base-weight-dependent factor (`right.T @ W`, shape `r x in_features`) instead of the full `out_features x in_features` delta, so `unmerge()` can recompute the delta cheaply (see the class-level comment: "unmerge recomputes the exact delta from it"). Two bugs in that cache mechanism:

### 1. Precision: the cached factor gets rounded before caching, so unmerge recomputes a *different* delta than merge applied

```python
# _merge_factor always returns float32 (see its docstring: "right.T @ W ... in float32")
factor = self._merge_factor(adapter_name)
...
self._cached_merge_factor[active_adapter] = factor.to(base_layer.weight.dtype)  # <-- rounds it for bf16/fp16
```

For a bf16/fp16 base layer, this downcasts the float32 factor before caching it. `unmerge()` then does:

```python
factor = self._cached_merge_factor.pop(active_adapter, None)
delta_weight = self._delta_from_factor(active_adapter, factor.to(torch.float32))  # upcasts the already-rounded value
```

— recomputing the delta from a *rounded-then-upcast* factor, not the pristine one `merge()` used. The two deltas differ. This is not the ordinary storage-precision loss every low-rank merge already has when its output dtype is narrower than float32 (that's unavoidable and shared by every tuner) — it's an *extra*, avoidable rounding step specific to this cache, breaking the "exact delta" guarantee the surrounding comment documents.

**Verified with a probe**: capturing `_merge_factor`'s output before merge and comparing it against what's cached after merge shows they differ (`cached factor equals fresh factor: False`, dtype `bfloat16` instead of `float32`), and the delta recomputed at merge time vs. recomputed at unmerge time from the cache are provably different tensors (`delta_at_merge == delta_at_unmerge: False`) — even visible directly in the printed values (e.g. `3.6478e-05` vs `3.6240e-05` for one element). After the fix, both checks are `True`.

I want to be precise about what "exact" means here (and what I'm *not* claiming): even with this fix, `merge()` followed by `unmerge()` on a bf16 base layer does **not** generally produce a bit-identical recovery of the original weight via `torch.equal` — that's not achievable for *any* tuner once the merged result is stored in a narrower-than-float32 dtype (I verified `LoraLayer`'s own merge/unmerge has the same non-zero residual for an equivalent bf16 scenario, and a raw `bf16 (a+b)-b == a` sanity check shows the same: this is an inherent property of bf16 arithmetic, not a bug). What this fix *does* guarantee — and what the existing code comment already promises — is that the *delta* `unmerge()` recomputes from the cache is bit-identical to the delta `merge()` actually applied, which was broken before this fix and is exact after it.

The existing `test_merge_caches_small_factor_and_unmerges_exactly` test (uses a float32 base layer) never caught this, because downcasting float32 → float32 loses nothing — the bug is invisible without a narrower base dtype.

**Fix**: keep the cached factor in float32 (drop the `.to(base_layer.weight.dtype)` downcast). It's still only `r x in_features` floats — still the intended memory saving over caching the full delta — just not downcast further.

### 2. Device staleness: the cache doesn't follow `model.to(device)`

`_cached_merge_factor` was a plain `{}` dict, not registered as `adapter_layer_names`/a proper buffer container. This makes it invisible to both:
- `nn.Module.to(device)`'s native recursive traversal (plain dicts aren't modules/parameters/buffers, so PyTorch has no way to know it holds tensors), and
- PEFT's own `_move_adapter_to_device_of_base_layer`, which only acts on `nn.ModuleDict`/`nn.ParameterDict`/`BufferDict` entries.

`deft_P`/`deft_R` (real `nn.ParameterDict`s) move correctly with the model; the cache doesn't.

**Reproduced**: merge on CPU, `model.to("cuda")`, then `unmerge()` raises:
```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

**Fix**: make the cache a `BufferDict(persistent=False)` — the same container FRoD already uses for its own analogous merge-time cache (`_frod_merged_delta`). `persistent=False` since this is a recomputable cache, not real adapter state that should end up in `state_dict()`/checkpoints. Added `"_cached_merge_factor"` to `other_param_names`, matching how FRoD registers `frod_U`/`frod_V`/etc. `unmerge()`'s `self._cached_merge_factor.pop(active_adapter, None)` call site is updated, since `BufferDict.pop(key)` (unlike `dict.pop`) doesn't accept a default argument.

`gh pr list --repo huggingface/peft --search "deft merge factor" / "deft unmerge precision" / "_cached_merge_factor" --state all` and `gh issue list --search "deft unmerge"` surface only the original DEFT feature PR (#3342, merged). No open PR or issue addresses this.

## Tests

Extends `tests/test_deft.py::TestDeftMerge` with:
- `test_merge_unmerge_roundtrip_precise_for_bf16_base` — a black-box, user-facing test through the public `merge_adapter()`/`unmerge_adapter()` API: casts the base model to bf16, merges, then measures the base-weight roundtrip error both with the float32-cached factor and with the factor rounded to the base dtype (reproducing the pre-fix behaviour), asserting the float32 path is strictly more accurate.
- `test_cached_merge_factor_is_a_buffer_dict` — a CPU-only structural check for the device-staleness fix: asserts `_cached_merge_factor` is a `BufferDict` and registered in `other_param_names`, and that merging populates it with a real tensor. This replaces an earlier `@require_torch_gpu` cross-device-move test that was dropped as narrow/GPU-gated; this version catches a regression back to a plain dict without needing a GPU.

Both pre-existing tests in the file (`test_para_removal_only_and_exact_merge`, `test_merge_caches_small_factor_and_unmerges_exactly`) still pass unchanged.

**A note on `test_merge_unmerge_roundtrip_precise_for_bf16_base`'s parameters**: at `deft_P`'s default init scale (std=0.02, used by an earlier version of this test that only scaled `deft_R`), the cached factor's own magnitude (`right.T @ W`, independent of `deft_R`) was too small for its bf16-rounding to shift the recomputed delta by more than the base weight's own ordinary bf16 storage-quantization step — so the two measured errors were bit-for-bit identical regardless of whether the fix was present, and the assertion (originally `<=`) held vacuously either way. I caught this by reverting the source fix and re-running the test unmodified: it still passed. Scaling `deft_P` to std=1.0 and tightening the assertion to strict `<` makes the discrepancy reliably observable; verified fail-before (reverted fix: `err_float32 == err_downcast`, assertion correctly fails) / pass-after (fixed: `err_float32 < err_downcast` with a clear, non-marginal margin) at the test's fixed seed.

Verified fail-before / pass-after for the device fix the same way: reverting `_cached_merge_factor` to a plain dict makes `test_cached_merge_factor_is_a_buffer_dict` fail on the `isinstance(..., BufferDict)` assertion; restoring the fix passes.

## AI assistance disclosure

This PR was prepared with AI assistance (Claude Code). There is no pre-existing issue for this bug — it was found while auditing tuner merge/unmerge cache mechanisms for correctness. I independently verified both root causes (traced the factor's dtype through `_merge_factor`/`merge`/`unmerge`, confirmed the delta mismatch empirically, cross-checked against `LoraLayer`'s and `FrodLayer`'s analogous, correct mechanisms), reproduced both bugs and their fixes end-to-end against the real library, and reviewed every changed line before proposing this change.

### DIFF
--- a/src/peft/tuners/deft/layer.py
+++ b/src/peft/tuners/deft/layer.py
@@ -56,9 +56,8 @@ class DeftLayer(BaseTunerLayer):
         # The residual projection (I - P_proj) @ W is not invertible, so unmerge cannot recover the delta from the
         # merged weight. Rather than cache the full out x in delta, cache only its base-weight-dependent factor
         # right.T @ W (shape r x in_features) per adapter; unmerge recomputes the exact delta from it (~r/out_features
-        # of the memory). A BufferDict (not a plain dict) so it is swept up by `model.to(device)` and by
-        # `_move_adapter_to_device_of_base_layer`, instead of going stale if the model is moved between merge
-        # and unmerge; `persistent=False` since it's a recomputable cache, not real adapter state to checkpoint.
+        # of the memory). A BufferDict so `model.to(device)` and `_move_adapter_to_device_of_base_layer` keep it in
+        # sync with the base layer; `persistent=False` since it's a recomputable cache, not real adapter state.
         self._cached_merge_factor = BufferDict({}, persistent=False)
         # Mark the weight as unmerged
         self._disable_adapters = False
@@ -284,12 +283,9 @@ class DeftLinear(nn.Module, DeftLayer):
                 base_layer.weight.data = new_weight
             else:
                 base_layer.weight.data = base_layer.weight.data + delta_weight
-            # Cache only the small r x in_features factor (not the full out x in delta) for an exact
-            # unmerge. Keep it in float32 (the dtype _merge_factor already returns): downcasting it to
-            # the base weight's dtype here would round it, so unmerge would recompute the delta from a
-            # different (rounded) factor than the one merge used, i.e. a *different* delta than the one
-            # actually applied -- not just the ordinary storage-precision loss every low-rank merge already
-            # has when its dtype is narrower than float32.
+            # Cache only the small r x in_features factor (not the full out x in delta) for an exact unmerge.
+            # Kept in float32 (the dtype _merge_factor already returns) so unmerge recomputes the exact delta
+            # merge applied here, not a rounded approximation of it.
             self._cached_merge_factor[active_adapter] = factor
             self.merged_adapters.append(active_adapter)
 
@@ -302,6 +298,7 @@ class DeftLinear(nn.Module, DeftLayer):
             active_adapter = self.merged_adapters.pop()
             if active_adapter not in self.deft_P.keys():
                 continue
+            # Unlike dict.pop(), BufferDict.pop() takes no default value, so membership is checked first.
             factor = (
                 self._cached_merge_factor.pop(active_adapter) if active_adapter in self._cached_merge_factor else None
             )
--- a/src/peft/tuners/deft/layer.py
+++ b/src/peft/tuners/deft/layer.py
@@ -23,6 +23,7 @@ from transformers.pytorch_utils import Conv1D
 from peft.tuners.tuners_utils import BaseTunerLayer, check_adapters_to_merge
 from peft.utils.other import transpose
 
+from .._buffer_dict import BufferDict
 from .config import DeftConfig
 
 
@@ -37,6 +38,7 @@ class DeftLayer(BaseTunerLayer):
         "deft_para",
         "deft_scaling",
         "deft_dropout",
+        "_cached_merge_factor",
     )
 
     def __init__(self, base_layer: nn.Module, **kwargs) -> None:
@@ -54,8 +56,10 @@ class DeftLayer(BaseTunerLayer):
         # The residual projection (I - P_proj) @ W is not invertible, so unmerge cannot recover the delta from the
         # merged weight. Rather than cache the full out x in delta, cache only its base-weight-dependent factor
         # right.T @ W (shape r x in_features) per adapter; unmerge recomputes the exact delta from it (~r/out_features
-        # of the memory).
-        self._cached_merge_factor = {}
+        # of the memory). A BufferDict (not a plain dict) so it is swept up by `model.to(device)` and by
+        # `_move_adapter_to_device_of_base_layer`, instead of going stale if the model is moved between merge
+        # and unmerge; `persistent=False` since it's a recomputable cache, not real adapter state to checkpoint.
+        self._cached_merge_factor = BufferDict({}, persistent=False)
         # Mark the weight as unmerged
         self._disable_adapters = False
         self.merged_adapters = []
@@ -280,8 +284,13 @@ class DeftLinear(nn.Module, DeftLayer):
                 base_layer.weight.data = new_weight
             else:
                 base_layer.weight.data = base_layer.weight.data + delta_weight
-            # cache only the small r x in_features factor (not the full out x in delta) for an exact unmerge
-            self._cached_merge_factor[active_adapter] = factor.to(base_layer.weight.dtype)
+            # Cache only the small r x in_features factor (not the full out x in delta) for an exact
+            # unmerge. Keep it in float32 (the dtype _merge_factor already returns): downcasting it to
+            # the base weight's dtype here would round it, so unmerge would recompute the delta from a
+            # different (rounded) factor than the one merge used, i.e. a *different* delta than the one
+            # actually applied -- not just the ordinary storage-precision loss every low-rank merge already
+            # has when its dtype is narrower than float32.
+            self._cached_merge_factor[active_adapter] = factor
             self.merged_adapters.append(active_adapter)
 
     def unmerge(self) -> None:
@@ -293,7 +302,9 @@ class DeftLinear(nn.Module, DeftLayer):
             active_adapter = self.merged_adapters.pop()
             if active_adapter not in self.deft_P.keys():
                 continue
-            factor = self._cached_merge_factor.pop(active_adapter, None)
+            factor = (
+                self._cached_merge_factor.pop(active_adapter) if active_adapter in self._cached_merge_factor else None
+            )
             if factor is not None:
                 delta_weight = self._delta_from_factor(active_adapter, factor.to(torch.float32))
             else:
--- a/tests/test_deft.py
+++ b/tests/test_deft.py
@@ -17,8 +17,6 @@ from torch import nn
 
 from peft import DeftConfig, get_peft_model
 
-from .testing_utils import require_torch_gpu
-
 
 class MLP(nn.Module):
     def __init__(self, bias=True):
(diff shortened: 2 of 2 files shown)


tests/test_deft.py
```diff
@@ -109,3 +111,65 @@ def test_merge_caches_small_factor_and_unmerges_exactly(self):
...
+        assert torch.equal(cached_factor, factor_before_merge)
+
+        # This mirrors exactly what unmerge() does with the cached factor.
+        delta_at_unmerge = layer._delta_from_factor("default", cached_factor.to(torch.float32))
+        assert torch.equal(delta_at_merge, delta_at_unmerge), (
+            "delta recomputed at unmerge time must be bit-identical to the delta merge() applied"
+        )
+
+    @require_torch_gpu
+    def test_cached_merge_factor_follows_model_to_device(self):
```
BenjaminBossan: This test wouldn't run because it requires a GPU. It could be moved to `test_gpu_examples.py` but IMO it's narrow enough that it can be dropped.

tests/test_deft.py
```diff
@@ -109,3 +111,65 @@ def test_merge_caches_small_factor_and_unmerges_exactly(self):
         assert torch.allclose(peft_model(x), unmerged_out, atol=1e-4)
         peft_model.unmerge_adapter()
         assert torch.allclose(layer.base_layer.weight, w0, atol=1e-5)
+
+    def test_merge_caches_factor_in_float32_for_bf16_base_weight(self):
```
BenjaminBossan: I'm not convinced by this test. It is checking implementation details and not for something that a user would care for. IIUC, the downcast should lead to a loss in precision when it comes to the merge -> unmerge roundtrip when the base model is in fp16 or bf16. This is what the test should check. As you mention, the roundtrip is not expected to be 100% accurate, but if this loss in precision due to the downcast matters, we should be able to write a test that measures it.

BenjaminBossan (changes requested):
Thanks for this PR to fix these two issues.

Regarding the merging issue, I commented on the approach of the test, please check.

Regarding the use of `BufferDict`, this looks fine in my book.

@MAXNORM8650 If you have the opportunity, please check as well.

DaoyuanLi2816:
Both addressed:
- Dropped the `@require_torch_gpu` test (`test_cached_merge_factor_follows_model_to_device`) as suggested — it was narrow and GPU-gated.
- Replaced the dtype/impl-detail test with a **user-facing roundtrip-precision** test (`test_merge_unmerge_roundtrip_precise_for_bf16_base`). With a bf16 base it runs `merge_adapter()` → `unmerge_adapter()` via the public API and measures the max-abs error of the restored base weights, both with the float32-cached factor and with a factor rounded to the base dtype (reproducing the pre-fix behaviour). It asserts the float32 path restores the weights at least as accurately as the downcast one — i.e. it actually *measures* the precision impact of the downcast on the observable roundtrip, rather than checking the cache's internal dtype.

Let me know if you'd prefer a different framing (e.g. a fixed tolerance rather than the comparison).

DaoyuanLi2816:
Pushed two follow-up fixes to the test suite:

1. **`test_merge_unmerge_roundtrip_precise_for_bf16_base` wasn't actually discriminating.** At `deft_P`'s default init scale (std=0.02), the cached factor's own magnitude was too small for its bf16-rounding to shift the recomputed delta beyond the base weight's *ordinary* bf16 storage-quantization step — so `err_float32 == err_downcast` every time, and the `<=` assertion held vacuously regardless of whether the fix was present. I caught this by reverting the source fix and re-running the test unmodified: it still passed. Fixed by also scaling `deft_P` (to std=1.0) so the effect is reliably observable, and tightened the assertion to strict `<`. Re-verified fail-before/pass-after against the real fix.

2. **Added back CPU-only coverage for the device-staleness fix.** After the `@require_torch_gpu` test was dropped, `_cached_merge_factor` becoming a `BufferDict` had no regression test at all. `test_cached_merge_factor_is_a_buffer_dict` asserts the container type and `other_param_names` registration (no actual cross-device move needed), so it catches a regression back to a plain dict without reintroducing the GPU-gating concern.

Both verified fail-before/pass-after against the actual fix in `layer.py`.

MAXNORM8650:
Checked as requested @BenjaminBossan. Both are real bugs in the merge-factor cache from #3342.

- The precision test now measures the user-facing roundtrip: I reproduced it on a bf16 base — float32-cache vs downcast-cache gives max|err| 7.8e-3 vs 2.1e-2 (~2.7× worse) on GPU, and fail-before/pass-after holds on CPU too (reverting the source fix makes it fail). So the downcast loss is real and observable, not an impl detail.
 - Device staleness: on a GPU box, merge-on-CPU → model.to("cuda") → unmerge() raises RuntimeError: …cuda:0 and cpu with the merged dict; the BufferDict fix moves the cache with the model and restores the weight exactly.

Also worth noting: the relu-only choice is correct — for qr the factor Q_P.T @ W is orthonormal-scaled, so its bf16 rounding stays below the base's own quantization step and the downcast is unobservable (float32 == downcast regardless of the fix). A qr precision test would be vacuous; relu (with the scaled deft_P) is the only decomposition that can surface it.

src/peft/tuners/deft/layer.py
```diff
@@ -53,8 +55,10 @@ def __init__(self, base_layer: nn.Module, **kwargs) -> None:
         # The residual projection (I - P_proj) @ W is not invertible, so unmerge cannot recover the delta from the
         # merged weight. Rather than cache the full out x in delta, cache only its base-weight-dependent factor
         # right.T @ W (shape r x in_features) per adapter; unmerge recomputes the exact delta from it (~r/out_features
-        # of the memory).
-        self._cached_merge_factor = {}
+        # of the memory). A BufferDict (not a plain dict) so it is swept up by `model.to(device)` and by
```
BenjaminBossan: This comment is not needed, it only really makes sense compared to the previous code.

src/peft/tuners/deft/layer.py
```diff
@@ -279,8 +283,13 @@ def merge(self, safe_merge: bool = False, adapter_names: Optional[list[str]] = N
                 base_layer.weight.data = new_weight
             else:
                 base_layer.weight.data = base_layer.weight.data + delta_weight
-            # cache only the small r x in_features factor (not the full out x in delta) for an exact unmerge
-            self._cached_merge_factor[active_adapter] = factor.to(base_layer.weight.dtype)
+            # Cache only the small r x in_features factor (not the full out x in delta) for an exact
+            # unmerge. Keep it in float32 (the dtype _merge_factor already returns): downcasting it to
```
BenjaminBossan: Again, no need to comment on the code that was there previously. The comments should explain when something that's not intuitive or that's non-trivial happens to help the reader of the code understand. It should not serve as a justification for why code was changed. That type of information belongs to the PR description.

src/peft/tuners/deft/layer.py
```diff
@@ -292,7 +301,9 @@ def unmerge(self) -> None:
             active_adapter = self.merged_adapters.pop()
             if active_adapter not in self.deft_P.keys():
                 continue
-            factor = self._cached_merge_factor.pop(active_adapter, None)
+            factor = (
+                self._cached_merge_factor.pop(active_adapter) if active_adapter in self._cached_merge_factor else None
```
BenjaminBossan, line 303: Is this change needed?

tests/test_deft.py
```diff
@@ -109,3 +110,63 @@ def test_merge_caches_small_factor_and_unmerges_exactly(self):
...
+                rounded = layer._cached_merge_factor["default"].to(dtype).to(torch.float32)
+                layer._cached_merge_factor["default"] = rounded
+            peft_model.unmerge_adapter()
+            return (layer.base_layer.weight.float() - w0.float()).abs().max().item()
+
+        err_float32 = roundtrip_max_abs_error(downcast_cached_factor=False)
+        err_downcast = roundtrip_max_abs_error(downcast_cached_factor=True)
+        assert err_float32 < err_downcast
+
+    def test_cached_merge_factor_is_a_buffer_dict(self):
```
BenjaminBossan: This test checks an implementation detail and should thus be removed. It's okay to not have an explicit test here, it's the same situation for other PEFT methods using `BufferDict`.

tests/test_deft.py
```diff
@@ -109,3 +110,63 @@ def test_merge_caches_small_factor_and_unmerges_exactly(self):
...
+                # unmerge recomputes the delta from the rounded factor instead of the exact one.
+                dtype = layer.base_layer.weight.dtype
+                rounded = layer._cached_merge_factor["default"].to(dtype).to(torch.float32)
+                layer._cached_merge_factor["default"] = rounded
+            peft_model.unmerge_adapter()
+            return (layer.base_layer.weight.float() - w0.float()).abs().max().item()
+
+        err_float32 = roundtrip_max_abs_error(downcast_cached_factor=False)
+        err_downcast = roundtrip_max_abs_error(downcast_cached_factor=True)
+        assert err_float32 < err_downcast
```
BenjaminBossan: Just checking `<` is a bit risky, as it could detect even minuscule differences. Let's assert that there is a certain margin. In my testing, it is ~30%.

tests/test_deft.py
```diff
@@ -109,3 +110,63 @@ def test_merge_caches_small_factor_and_unmerges_exactly(self):
...
+        # drifts further than the plain bf16 add/subtract rounding. Measure the base-weight roundtrip
+        # error both ways and assert the float32-cached factor is strictly more accurate.
+        #
+        # `deft_P` (hence the cached factor `right.T @ W`) is deliberately scaled well above its
+        # default init (std=0.02): the factor's own magnitude has to be large enough that rounding
+        # it to bf16 shifts the recomputed delta by more than the base weight's own bf16 storage
+        # quantization step -- otherwise the discrepancy this bug introduces is silently absorbed by
+        # that ordinary rounding and this test can't observe it (verified empirically: at the default
+        # init scale, `err_float32 == err_downcast` every time, which would make this test pass
+        # whether the bug is present or not).
```
BenjaminBossan: Let's shorten this comment and to 1-2 sentences and add a reference to this PR instead.

BenjaminBossan (changes requested):
Thanks @DaoyuanLi2816 for the updates and @MAXNORM8650 for double-checking.

I still have some issues which I commented on, but the general contents of the PR are good.

DaoyuanLi2816:
Rebased onto main to resolve the conflict with #3376 (both `deft_dropout` and `_cached_merge_factor` now in `other_param_names`).

Addressed all six comments:
- `layer.py`: trimmed the two comments (BufferDict init, float32 cache) down to state the invariant itself, not the comparison to the removed code.
- `layer.py` (the `.pop()` ternary): this one is needed, not incidental churn — `BufferDict.pop()` takes no default-value argument (unlike `dict.pop(key, default)`), so the original one-liner would raise once the container stopped being a plain dict. Added a short comment stating that instead of reverting.
- `test_deft.py`: dropped `test_cached_merge_factor_is_a_buffer_dict` entirely, per your point that other PEFT methods using `BufferDict` don't have an equivalent test either.
- `test_deft.py`: shortened the `deft_P` scaling comment to two sentences with a reference to this PR instead of the inline empirical writeup.
- `test_deft.py`: replaced the bare `<` with `err_float32 < 0.9 * err_downcast` (require at least 10% more accurate, rather than merely "not worse"). I picked 10% conservatively to leave headroom under the ~30% margin you saw empirically — happy to tighten it if you'd prefer closer to that number.

tests/test_deft.py
```diff
@@ -109,3 +109,41 @@ def test_merge_caches_small_factor_and_unmerges_exactly(self):
...
+                dtype = layer.base_layer.weight.dtype
+                rounded = layer._cached_merge_factor["default"].to(dtype).to(torch.float32)
+                layer._cached_merge_factor["default"] = rounded
+            peft_model.unmerge_adapter()
+            return (layer.base_layer.weight.float() - w0.float()).abs().max().item()
+
+        err_float32 = roundtrip_max_abs_error(downcast_cached_factor=False)
+        err_downcast = roundtrip_max_abs_error(downcast_cached_factor=True)
+        # A fixed margin rather than a bare `<`, since the two errors being merely unequal isn't a
+        # reliable signal on its own -- require the float32 path to be at least 10% more accurate.
```
BenjaminBossan: Again, let's not leave a comment that refers to the code that was there previously, let's comment on what is there.

BenjaminBossan (changes requested):
Generally looks good, but please check your comments. It makes no sense to comment on how the code evolved during the discussion of the PR. The comment must make sense in isolation, we can't expect the reader to check the PR discussion.

DaoyuanLi2816:
Pushed a follow-up addressing the comment self-containment issue — reworded the test comments so they describe current behavior standalone rather than referencing prior implementation details, and trimmed the old-code comparisons from the earlier round. Should be ready for re-review.

tests/test_deft.py
```diff
@@ -109,3 +109,42 @@ def test_merge_caches_small_factor_and_unmerges_exactly(self):
         assert torch.allclose(peft_model(x), unmerged_out, atol=1e-4)
         peft_model.unmerge_adapter()
         assert torch.allclose(layer.base_layer.weight, w0, atol=1e-5)
+
+    def test_merge_unmerge_roundtrip_precise_for_bf16_base(self):
+        # User-facing guarantee: merging then unmerging a DEFT adapter restores the base weights. This requires
+        # the cached factor to stay in float32: rounding it to the base dtype first makes unmerge recompute a
+        # delta that differs from the one merge applied, so the roundtrip drifts further than plain bf16 storage
+        # rounding alone. Measure the base-weight roundtrip error both ways and assert the float32-cached factor
+        # is meaningfully more accurate.
```
BenjaminBossan: Here is a suggestion for a more helpful comment:

> Test that a merge->unmerge roundtrip is as precise as possible (see discussion in #3412).
In the previous implementation, an intermediate value was cast to the dtype of the base weight, which introduced extra source of imprecision. This test ensures that with the current implementation, that error is meaningfully reduced.

tests/test_deft.py
```diff
@@ -109,3 +109,42 @@ def test_merge_caches_small_factor_and_unmerges_exactly(self):
...
+                rounded = layer._cached_merge_factor["default"].to(dtype).to(torch.float32)
+                layer._cached_merge_factor["default"] = rounded
+            peft_model.unmerge_adapter()
+            return (layer.base_layer.weight.float() - w0.float()).abs().max().item()
+
+        err_float32 = roundtrip_max_abs_error(downcast_cached_factor=False)
+        err_downcast = roundtrip_max_abs_error(downcast_cached_factor=True)
+        # Require a real margin, not mere inequality: two error values simply differing isn't on its own a
+        # reliable signal that the difference is meaningful, so require the float32 path to be at least 10% more
+        # accurate.
```
BenjaminBossan: "require the float32 path to be at least 10% more accurate" is enough of a comment to explain what's going on.

BenjaminBossan (changes requested):
The comments are still not well to the point, I wrote some suggestions.

DaoyuanLi2816:
Thanks, applied both suggestions verbatim:

- [docstring on `test_merge_unmerge_roundtrip_precise_for_bf16_base`](https://github.com/huggingface/peft/pull/3412#discussion_r3622399334): replaced with your suggested wording exactly — "Test that a merge->unmerge roundtrip is as precise as possible (see discussion in #3412). In the previous implementation, an intermediate value was cast to the dtype of the base weight, which introduced extra source of imprecision. This test ensures that with the current implementation, that error is meaningfully reduced."
- [the comment above the final assert](https://github.com/huggingface/peft/pull/3412#discussion_r3622404461): trimmed to a single line, "Require the float32 path to be at least 10% more accurate.", since you noted that's enough to explain what's going on.

Pushed as edcedcf1. Ran `pytest tests/test_deft.py` locally (all 3 tests pass) and `ruff check` / `ruff format --check` on the file — both clean.

BenjaminBossan (approved):
Thanks for the latest changes, PR LGTM.

PR #3415, opened 2026-07-09
ENH: Warn in get_peft_model when base model parameters are on the meta device

### LINKED ISSUE #3409: get_peft_model on a lazily-materialized model leaves third-party hooks and optimizers holding stale parameter references — warn at wrap time?
### System Info

peft 0.18.1, transformers 5.0.0, torch 2.10-2.12, reproduced on CPU (macOS), Kaggle T4 (CUDA 12), and RTX 5090 (CUDA 12.8).

### Who can help?

@BenjaminBossan

### Reproduction

Full investigation with reproducer: meta-pytorch/opacus#820. Short version:

```python
model = AutoModelForCausalLM.from_pretrained("gpt2")
# model NOT moved to device here
# third-party library registers per-parameter hooks now (opacus's
# PrivacyEngine in our case, but any hook-registering framework hits this)
# ...
batch = batch.cuda()  # only the batch moves; lazy handling materializes
                      # params at first forward, AFTER hooks were registered
```

Training then proceeds looking completely healthy. Loss decreases, checkpoints save, and in the DP case the privacy accountant keeps reporting epsilon. The LoRA weights never update: `max |Δ| = 0.0` after any number of steps. Moving the model to the device before `get_peft_model` fixes it. We confirmed this on three independent environments over a five-month investigation (write up: https://imranahamed.substack.com/p/the-dp-lora-silent-corruption-how).

### Expected behavior

The root cause isn't a peft bug. It's an ordering hazard between peft's lazy device handling and anything that captures parameter references at wrap time (per parameter hooks, early built optimizers). But peft is the layer that can see the hazard forming, and `get_peft_model` already warns on several usage hazards (double wrapping, renamed base models, adapter name prefix collisions).

Two suggestions, happy to PR either or both:

1. A pre-flight warning in `get_peft_model` when base-model params are on `meta` (or `_hf_hook.original_devices` indicates pending materialization) outside the known `device_map` inference flows: something like "parameters will be materialized after wrapping; parameter references captured now (hooks, optimizers) may go stale." The meta-device detection pattern already exists in `tuners_utils.py` and `peft_model.py`, so this reuses it rather than adding new machinery. Roughly 10-15 lines plus a test.

2. A troubleshooting entry in `docs/source/developer_guides/troubleshooting.md` documenting the ordering requirement (move to device, then wrap, then register hooks / build optimizers), next to the existing "Bad results from a loaded PEFT model" section.

The condition scoping on (1) needs care to avoid false positives on legitimate `device_map="auto"` inference loading; I'd propose gating it to training mode wrapping or making it a one-time info-level warning. Open to whatever shape you think fits.

immu4989:
Resolves #3409.

Adds a troubleshooting docs entry for a silent training no-op: when `get_peft_model` wraps a model whose base parameters are still on the meta device, anything that captures parameter references before they are materialized (third-party forward/backward hooks, or an optimizer created before the first forward) ends up pointing at stale tensors, so training can look healthy while no weights update. The entry describes the failure signature and the safe ordering.

Context and coordination: the failure mode was documented in meta-pytorch/opacus#820 and discussed in #3409; the docs draft was requested by @BenjaminBossan in https://github.com/huggingface/peft/issues/3409#issuecomment-4924292338. Per review on this PR the runtime warning was dropped, since PEFT neither causes the ordering nor can detect it precisely enough to avoid false positives, keeping the docs entry only.

Tests: docs-only change, no unit tests apply. `make style` and `make quality` pass locally.

### DIFF
--- a/docs/source/developer_guides/troubleshooting.md
+++ b/docs/source/developer_guides/troubleshooting.md
@@ -100,7 +100,7 @@ If training appears healthy (the loss goes down, checkpoints are saved without e
 sample_param = next(p for p in model.parameters() if p.requires_grad)
 before = sample_param.detach().clone()
 
-# ... train for at least one optimizer step ...
+# ... train for at least three optimizer steps ...
 
 after = next(p for p in model.parameters() if p.requires_grad)
 print("weights updated:", not torch.allclose(before, after.detach().cpu()))
--- a/docs/source/developer_guides/troubleshooting.md
+++ b/docs/source/developer_guides/troubleshooting.md
@@ -91,6 +91,29 @@ Disadvantages of half precision adapter:
 
 Note that for most use cases, overall runtime and memory cost will be determined by the size of the base model and by the dataset, while the dtype of the PEFT adapter will only have a small impact.
 
+## Training runs but the model does not improve
+
+If training appears healthy (the loss goes down, checkpoints are saved without errors) but the resulting model performs no better than the base model, check whether the trainable parameters actually changed during training:
+
+```python
+# before training
+sample_param = next(p for p in model.parameters() if p.requires_grad)
+before = sample_param.detach().clone()
+
+# ... train for at least one optimizer step ...
+
+after = next(p for p in model.parameters() if p.requires_grad)
+print("weights updated:", not torch.allclose(before, after.detach().cpu()))
+```
+
+If the weights did not move, a likely cause is that parameter references were captured before the parameters were materialized on their target device. This can happen when the base model is wrapped with [`get_peft_model`] while its parameters have not yet been moved to the target device, and a third-party library then registers forward/backward hooks (for example, per-sample gradient hooks from a differential privacy framework) or an optimizer is created before the first forward call. When the parameters are materialized later, those hooks and optimizer references point at stale tensors and every update becomes a silent no-op.
+
+To avoid this, use the following order:
+
+1. Load the base model and move it to the target device (e.g. `model.to(device)`).
+2. Wrap it with [`get_peft_model`].
+3. Only then register hooks or create the optimizer.
+
 ## Bad results from a loaded PEFT model
 
 There can be several reasons for getting a poor result from a loaded PEFT model which are listed below. If you're still unable to troubleshoot the problem, see if anyone else had a similar [issue](https://github.com/huggingface/peft/issues) on GitHub, and if you can't find any, open a new issue.
--- a/src/peft/mapping_func.py
+++ b/src/peft/mapping_func.py
@@ -113,7 +113,9 @@ def get_peft_model(
     if not low_cpu_mem_usage and not hasattr(model, "hf_device_map"):
         meta = torch.device("meta")
         if any(p.device == meta for p in model.parameters()):
-            warnings.warn(
+            # TEMPORARY (per PR review): raise instead of warn so CI can verify no existing test trips this
+            # condition, i.e. no false positives across the suite. Revert to warnings.warn before merge.
+            raise RuntimeError(
                 "The model has parameters on the meta device. They will only be materialized after wrapping, so "
                 "parameter references captured now (e.g. hooks registered by other libraries or an optimizer "
                 "created before the first forward call) can become stale and training may silently not update any "
--- a/src/peft/mapping_func.py
+++ b/src/peft/mapping_func.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import warnings
 from typing import Optional
 
+import torch
 from transformers import PreTrainedModel
 
 from .auto import MODEL_TYPE_TO_PEFT_MODEL_MAPPING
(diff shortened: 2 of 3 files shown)


src/peft/mapping_func.py
```diff
@@ -104,6 +105,22 @@ def get_peft_model(
             "This may lead to reinitialization of the adapter weights during loading."
         )
 
+    # Model parameters on the meta device at wrap time indicate that they will only be materialized later (e.g. on
+    # the first forward call). Anything that captures parameter references before that point, such as forward/backward
+    # hooks registered by third-party libraries or an optimizer created early, will reference stale parameters and
+    # training can silently fail to update any weights. Models loaded with a device_map (`hf_device_map` is set) or
+    # with low_cpu_mem_usage=True are excluded, as these flows handle materialization themselves.
+    if not low_cpu_mem_usage and not hasattr(model, "hf_device_map"):
```
BenjaminBossan: I wonder if we can narrow down the check even further: Check all attributes that torch uses to collect hooks and if there are no hooks, skip the warning. WDYT?

This wouldn't cover references in the optimizer, but I think those are pretty unrealistic: It would mean that the user has initialized the optimizer _before_ applying PEFT, which can't be correct.

BenjaminBossan (changes requested):
Thanks for creating the PR to add this warning.

To detect possible false-positives, I have an idea: Instead of warning, could you please raise an error? Then we can let the CI run. In theory, the error should not be hit by any of the existing tests. If we can verify that, it gives us better confidence that there are no false positives.

Of course, some of the new tests would fail, those could be marked with `@xfail(strict=True)`. After we have verified that, we can undo that change.

immu4989:
Good idea, Aprreciate it. Changed accordingly. Pushed a commit that swaps the warning for a `RuntimeError` and marks the meta device test `xfail(strict=True)`, so CI exercising the full suite will tell us whether any established usage trips the condition. I'll revert to `warnings.warn` once we've read the result. Locally the three negative tests (materialized model, `hf_device_map`, `low_cpu_mem_usage=True`) pass and only the positive test xfails, which is the expected shape.

On narrowing to hook presence: I think there's an ordering issue that makes it miss the motivating case. In the opacus repro the hooks aren't registered by `get_peft_model` time. The sequence is `get_peft_model` first, then `PrivacyEngine.make_private()` attaches the per sample gradient hooks afterward, then materialization happens on the first forward. So a "are torch hooks present now?" check inside `get_peft_model` would see none of them and skip the warning. The accelerate dispatch hook that does the lazy materialization is also stored as `_hf_hook` rather than in the standard `_forward_hooks`/`_backward_hooks` dicts, so it wouldn't reliably show up in a torch hook scan either.

You're right that the optimizer reference case is unrealistic, though, so I'll drop that half from the message. The meta-device condition is really functioning as a forward-looking signal here: it flags "materialization will happen later, so anything that captures references between now and then goes stale," which covers the hooks registered after case that a point in time hook scan can't. Whether that condition is too broad in practice is exactly what the error experiment should settle. If CI comes back clean, that's decent evidence the meta check alone is narrow enough; if it trips existing tests, those are the false positive patterns we'd need to exclude, and we can decide the narrowing from real cases.

BenjaminBossan:
> In the opacus repro the hooks aren't registered by `get_peft_model` time. The sequence is `get_peft_model` first, then `PrivacyEngine.make_private()` attaches the per sample gradient hooks afterward, then materialization happens on the first forward.

Hmm I see. It's a bit unfortunate. My concern is that fundamentally, it's not really PEFT's job to do this kind of checking.  PEFT just happens to sit in a place where it's convenient. In the end, the user chose to load the model on meta device and later attach hooks, it's not that PEFT caused that.

> The accelerate dispatch hook that does the lazy materialization is also stored as `_hf_hook` rather than in the standard `_forward_hooks`/`_backward_hooks` dicts, so it wouldn't reliably show up in a torch hook scan either.

We can add checks for the most common places where hooks are registered.

> Whether that condition is too broad in practice is exactly what the error experiment should settle. If CI comes back clean, that's decent evidence the meta check alone is narrow enough

This conclusion is a bit strong, the CI is not tripped but there are many different use cases out there and it's hard to anticipate all of them. And false positives have a cost, we don't want to train users to ignore PEFT warnings. So from my perspective, it's critical to make the check as precise as possible.

immu4989:
Thanks, this is a helpful way to frame it and I take the point. PEFT didn't cause the ordering and isn't responsible for it. It just happens to sit somewhere convenient, and a warning that fires on flows PEFT didn't create is the kind of thing that erodes trust in the warnings that do matter.

You're right that the meta signal on its own has a ceiling too. It can't tell a safe later materialization apart from the broken ordering, so green CI only tells us the current tests stay quiet, not that it stays quiet everywhere. If precision is the priority, and I think it should be, the meta condition alone probably isn't tight enough to justify a runtime warning.

The only variant I'd trust to be genuinely precise is much narrower: warn only when a meta param already has a hook attached at wrap time, which is almost always a real mistake. That one gives up the opacus case, since the hooks come later there, so it wouldn't cover what started this, but it also wouldn't fire on healthy flows.

My inclination is to land the troubleshooting docs entry, which captures the failure mode and the safe ordering with no false positive risk, and leave the runtime check out unless you think the narrow version above is worth having. Happy to go either way. Just tell me which you prefer and I'll get it into shape.

docs/source/developer_guides/troubleshooting.md
```diff
@@ -91,6 +91,29 @@ Disadvantages of half precision adapter:
...
+## Training runs but the model does not improve
+
+If training appears healthy (the loss goes down, checkpoints are saved without errors) but the resulting model performs no better than the base model, check whether the trainable parameters actually changed during training:
+
+```python
+# before training
+sample_param = next(p for p in model.parameters() if p.requires_grad)
+before = sample_param.detach().clone()
+
+# ... train for at least one optimizer step ...
```
BenjaminBossan: I'd say "train for at least three optimizer steps". Reason being that many PEFT methods are zero-initialized, meaning that after one step, some parameters get zero updates.

BenjaminBossan (changes requested):
Thanks for the nuanced discussion.

> My inclination is to land the troubleshooting docs entry, which captures the failure mode and the safe ordering with no false positive risk, and leave the runtime check out unless you think the narrow version above is worth having.

I agree, let's go with the doc update but remove the warning. It's a bit of an unfortunate confluence of factors that led to the opacus bug, but PEFT is not really the right place to warn about this.

immu4989:
Agreed, and thanks for pushing on it. Pushed a commit that removes the warning and the tests that covered it, so the PR is now just the troubleshooting entry. `src/peft/mapping_func.py` and `tests/test_mapping.py` are back to matching main exactly; the only change left is +23 lines in `docs/source/developer_guides/troubleshooting.md`.

The docs entry describes the symptom first (training looks healthy, loss falls, weights never move), gives the weight delta check to confirm it, then the ordering rule: move the model to its device, wrap with `get_peft_model`, then register hooks or build the optimizer. It doesn't name any third-party library, so it should stay accurate as the surrounding ecosystem changes.

Also worth recording for anyone who finds this later: the `error instead of warning` experiment did come back clean across the full suite, Python 3.10 through 3.13 on both Ubuntu and Windows. That says the condition doesn't fire on anything PEFT currently tests, which isn't the same as saying it wouldn't fire on real user code, which was your point.

immu4989:
Sounds good, docs only it is. The warning and its tests are already out as of the latest push, so the diff is down to just the troubleshooting entry now. Good catch on the three steps too, updated. Ready for another look.

BenjaminBossan (approved):
Thanks for providing the PR and the constructive discussion. LGTM.

PR #3420, opened 2026-07-11
Fix hotswapping for LoRA adapters targeting grouped Conv2d

### LINKED ISSUE #3416: [BUG]  `prepare_model_for_compiled_hotswap` fails for grouped Conv2d LoRA adapters
### System Info

- Python: `3.12.3`
- PEFT: `0.18.1`
- Transformers: `5.3.0`
- PyTorch: `2.10.0+cu128`

### Who can help?

@BenjaminBossan 

### Reproduction

I ran into this while preparing a LoRA adapter for compiled hotswapping with a larger target rank. The adapter targets a grouped `Conv2d`, which PEFT accepts during adapter injection (with the expected warning that merging is unsupported). This is the most minimal reproduction snippet I could reduce my original code to:

```python
import torch
from torch import nn

from peft.utils.hotswap import prepare_model_for_compiled_hotswap


class GroupedConvModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.conv = nn.Conv2d(4, 4, kernel_size=3, groups=2, bias=False)

    def forward(self, inputs):
        return self.conv(inputs)


model = get_peft_model(
    GroupedConvModel(),
    LoraConfig(
        r=2,
        lora_alpha=2,
        target_modules=["conv"],
    ),
)

inputs = torch.randn(1, 4, 8, 8)
model(inputs)

prepare_model_for_compiled_hotswap(model, target_rank=4)
```

This raises:

```text
ValueError: Something went wrong when trying to pad the LoRA weights, the new
shape should be torch.Size([4, 4, 1, 1]) but torch.Size([4, 2, 1, 1]) was found.
Please open an issue on PEFT ...
```

### Expected behavior

`prepare_model_for_compiled_hotswap` should support padding a LoRA adapter targeting a grouped `Conv2d` when the target rank is valid for the number of groups. 

I think that the problem is in padding LoRA B. PyTorch stores the Conv2d weight’s second dimension as input channels per group, but the hotswap helper currently treats that dimension as the full rank. 

I would be happy to give this a try and open a PR with a fix and a focused regression test. My first idea is to take `groups` into account when figuring out LoRA B's actual rank, and give its padded weight `target_rank // groups` input channels. We would also reject any target rank that cannot be evenly divided by the number of groups.

jayzuccarelli:
Fixes #3416

`prepare_model_for_compiled_hotswap` raises a shape mismatch for LoRA adapters that target a grouped `Conv2d`. For these layers, LoRA B is itself a grouped conv, so its weight stores `rank // groups` input channels, but `_get_padded_conv2d` treated that dimension as the full rank.

This PR:

- computes LoRA B's original rank as `in_channels * groups` and sizes its padded weight with `target_rank // groups` input channels
- lays out each group's LoRA A output channels at the start of that group's slice of the padded weight, so the grouped LoRA B multiplies the padding zeros with the right channels and the model output is unchanged
- rejects a target rank that is not divisible by the layer's groups

While testing the full flow, I found the same layout problem in `hotswap_adapter_from_state_dict`: when the incoming adapter has a smaller rank than the loaded (or padded) one, the contiguous copy of LoRA A silently produces wrong outputs for grouped convs (max abs diff vs. the adapter loaded directly was ~1.8 in my repro; this also reproduces on main without calling `prepare_model_for_compiled_hotswap`). The copy is now done per group block; with `groups == 1` it reduces to the previous single contiguous copy.

Tests run (CPU):

- new `test_prepare_model_for_compiled_hotswap_does_not_change_output_grouped_conv2d`: padding a grouped conv adapter preserves the output; without the fix it fails with the `ValueError` from the issue
- new `test_hotswap_works_different_ranks_grouped_conv2d`: pad to a larger target rank, hotswap in a smaller-rank adapter, output must match that adapter loaded directly; without the swap-path fix it fails (wrong output, no error)
- new `test_prepare_model_for_compiled_hotswap_grouped_conv2d_indivisible_target_rank_raises` for the new guard
- `pytest tests/test_initialization.py -k TestHotSwapping`: 35 passed
- `ruff check` and `ruff format --check` pass on the touched files

Coordination: reported in #3416, @BenjaminBossan gave the go-ahead in https://github.com/huggingface/peft/issues/3416#issuecomment-4934017022.

AI assistance was used for this PR (Claude Code). The change was reviewed and the tests were run before submitting.

### DIFF
--- a/src/peft/utils/hotswap.py
+++ b/src/peft/utils/hotswap.py
@@ -148,7 +148,9 @@ def _get_padded_linear(lora_module: torch.nn.Module, target_rank: int, is_lora_A
     return new_layer
 
 
-def _get_padded_conv2d(lora_module: torch.nn.Module, target_rank: int, is_lora_A: bool) -> torch.nn.Conv2d:
+def _get_padded_conv2d(
+    lora_module: torch.nn.Module, target_rank: int, is_lora_A: bool, groups: int = 1
+) -> torch.nn.Conv2d:
     """
     Get a new Conv2d layer for LoRA with padded weights according to the target rank.
 
@@ -159,6 +161,8 @@ def _get_padded_conv2d(lora_module: torch.nn.Module, target_rank: int, is_lora_A
             The desired rank to pad to.
         is_lora_A (bool):
             True if this is the LoRA A matrix, False if LoRA B.
+        groups (int):
+            The number of groups of the targeted base Conv2d layer.
 
     Returns:
         nn.Conv2d:
@@ -167,7 +171,8 @@ def _get_padded_conv2d(lora_module: torch.nn.Module, target_rank: int, is_lora_A
     weight = lora_module.weight
     # For Conv2d: [out_channels, in_channels, kernel_height, kernel_width]
     out_channels, in_channels, kh, kw = weight.shape
-    original_rank = out_channels if is_lora_A else in_channels
+    # LoRA B is a grouped conv, so it stores rank // groups input channels; LoRA A is always groups=1
+    original_rank = out_channels if is_lora_A else in_channels * groups
 
     if original_rank == target_rank:
         return lora_module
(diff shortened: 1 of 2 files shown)


jayzuccarelli:
Gentle ping @BenjaminBossan — this one's a small fix for hotswapping LoRA adapters that target grouped Conv2d layers, with a regression test. Happy to adjust anything if the approach isn't what you'd want.

BenjaminBossan:
Closing this in favor of #3427. Even though this PR predates the other one, in the issue the task was assigned to the original creator of the issue. In the future, please confirm with the maintainers first if you should work on an issue.

PR #3421, opened 2026-07-11
Fix X-LoRA adapter lifecycle validation

DaoyuanLi2816:
## Summary

- preserve user-provided expert names when registering X-LoRA adapters
- reject outer/expert name collisions with a clear error
- reject deletion, subsets, and reordering because classifier columns are fixed at construction

## Tests

- `python -m pytest tests/test_xlora.py -q --no-cov` (20 passed)
- `python -m ruff check src/peft/tuners/xlora tests/test_xlora.py`
- `python -m ruff format --check src/peft/tuners/xlora tests/test_xlora.py`

## AI assistance disclosure

AI assistance was used. I reviewed the changes and ran the tests above.

### DIFF
--- a/src/peft/tuners/xlora/model.py
+++ b/src/peft/tuners/xlora/model.py
@@ -101,7 +101,7 @@ def _load_adapter_into_lora_model(
     This method emulates the behavior of `PeftModel.from_pretrained`. Updates to `PeftModel.from_pretrained` may need
     to be reflected here.
 
-    All params pertain to the adapter (adapter name, model id, `i` is the adapter number in 0 indexing).
+    All params pertain to the adapter (adapter name, model id).
     """
     from peft.peft_model import PeftModel
     from peft.tuners.lora.config import LoraConfig
@@ -241,6 +241,12 @@ class XLoraModel(BaseTuner):
         else:
             conf = config
 
+        if adapter_name in conf.adapters:
+            raise ValueError(
+                f"X-LoRA expert name {adapter_name!r} conflicts with the outer adapter name. "
+                "Use a different name for the expert."
+            )
+
         # Create an empty LoraModel
         base_lora_config = copy.copy(conf)
         base_lora_config.target_modules = DUMMY_TARGET_MODULES
@@ -264,10 +270,10 @@ class XLoraModel(BaseTuner):
             adapters_items = peft_config.adapters.items()
 
         if hasattr(self.xlora_config, "_subfolders"):
-            for i, (_adapter_name, model_id), subfolder in enumerate(adapters_items):
+            for (_adapter_name, model_id), subfolder in adapters_items:
                 _load_adapter_into_lora_model(
                     lora_model=self.lora_model,
-                    adapter_name=str(i),
+                    adapter_name=_adapter_name,
                     model_id=model_id,
                     torch_device=torch_device,
                     ephemeral_gpu_offload=ephemeral_gpu_offload,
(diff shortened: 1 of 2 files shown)


src/peft/tuners/xlora/model.py
```diff
@@ -267,7 +267,7 @@ def __init__(
             for i, (_adapter_name, model_id), subfolder in enumerate(adapters_items):
                 _load_adapter_into_lora_model(
                     lora_model=self.lora_model,
-                    adapter_name=str(i),
+                    adapter_name=_adapter_name,
```
EricLBuehler, line 276: `lora_model` already contains the outer `adapter_name`, normally `"default"`, from its construction. If an expert is also named `"default"`, this name hits the existing-config branch in `_load_adapter_into_lora_model`, skips LoRA config loading and layer injection, and then fails with unexpected checkpoint keys.

I think we should separate the placeholder and expert namespaces or reject this collision explicitly.

src/peft/tuners/xlora/model.py
```diff
@@ -402,6 +406,23 @@ def __getattr__(self, name: str):
...
+        write like `self.active_adapter = ...` (e.g. from the inherited `BaseTuner.delete_adapter` /
+        `BaseTuner.set_adapter`) would otherwise silently create a separate, stale attribute on `self` instead of
+        updating `self.lora_model.active_adapter`, which is what `disable_adapter_layers`/`enable_adapter_layers`
+        and the forward pass actually consult.
+        """
+        return self.lora_model.active_adapter
+
+    @active_adapter.setter
+    def active_adapter(self, value: str | list[str]) -> None:
+        self.lora_model.active_adapter = value
```
EricLBuehler, line 418: This only synchronizes the active-name list, but it does not update the fixed X-LoRA classifier or `xlora_config.adapters`. For example, after deleting `"1"` from `["0", "1", "2", "3"]`, the classifier still emits four columns while the layers enumerate `["0", "2", "3"]`, so experts `"2"` and `"3"` receive the wrong columns and the fourth column is discarded after normalization.

I think we should override or reject `delete_adapter`/`set_adapter` unless the classifier, config, and name-to-class mapping are updated together.

tests/test_xlora.py
```diff
@@ -478,3 +496,38 @@ def test_xlora_embed_scale_is_applied(self, tmp_path):
...
+
+        # Sanity check: forward works before deleting anything.
+        outputs_before = model.generate(input_ids=inputs.to(self.torch_device), max_new_tokens=8)
+        assert torch.isfinite(outputs_before[: inputs.shape[1] :]).all()
+
+        # `model` has 4 experts ("0", "1", "2", "3"); deleting one still leaves multiple active experts, which is
+        # the scenario that used to crash.
+        model.delete_adapter("1")
+
+        outputs_after = model.generate(input_ids=inputs.to(self.torch_device), max_new_tokens=8)
```
EricLBuehler, line 514: This call masks another deletion regression. `BaseTunerLayer.delete_adapter` activates the remaining adapters with `inference_mode=False`, making their weights trainable even when `use_trainable_adapters=False`, since generate only refreezes them afterward. I think that we should restore the X-LoRA freeze policy immediately during deletion

EricLBuehler (changes requested):

src/peft/tuners/xlora/model.py
DaoyuanLi2816: Added an explicit collision check before the inner LoRA model is constructed, so a conflicting expert name now fails clearly.

src/peft/tuners/xlora/model.py
DaoyuanLi2816: Deletion is now rejected. The classifier width and expert-column order are fixed at construction, so removing an expert cannot be made safe here.

tests/test_xlora.py
DaoyuanLi2816: The override rejects deletion before the base tuner path can reactivate adapters, so the existing freeze policy is preserved.

DaoyuanLi2816:
@EricLBuehler gentle ping on this one. I addressed the three review comments on 24 July: user-provided expert names are preserved on registration, outer/expert name collisions are rejected with a clear error, and deletion, subsets and reordering are rejected because the classifier columns are fixed at construction. `pytest tests/test_xlora.py -q` passes 20/20. Could you take another look when you have time?

PR #3422, opened 2026-07-11
FIX TinyLoRA weight_tying corruption when adding overlapping adapters

DaoyuanLi2816:
## What does this PR do?

`TinyLoraModel._create_and_replace` caches a model-level mapping from target-module key to a deterministic layer index (`_target_key_to_idx`), which is used to compute how many `weight_tying` groups exist and which group each module falls into. The mapping is (re)built lazily:

```python
if not hasattr(self, "_target_key_to_idx") or current_key not in self._target_key_to_idx:
    self._target_key_to_idx = self._build_target_key_mapping(tinylora_config)
```

This rebuild condition only checks whether the *current module's key* is missing from the *existing* mapping — it does not check whether the mapping was built for the *current adapter*. `PeftModel.add_adapter` calls `inject_adapter` directly (bypassing `_pre_injection_hook`), so when a second adapter is added whose `target_modules` overlap with (but differ from) a prior adapter's, any module that happens to also be a key in the first adapter's stale mapping skips the rebuild and reuses that adapter's layer index and module count. This silently corrupts the number of `weight_tying` groups (and hence the trainable parameter count / sharing structure) computed for the second adapter, with no error or warning.

### Root cause

Concretely: adapter "default" targets `["lin0", "lin1", "lin2", "lin3"]`, building `_target_key_to_idx = {"lin0": 0, "lin1": 1, "lin2": 2, "lin3": 3}` (4 entries). A second adapter "b" then targets only `["lin1", "lin2"]` with `weight_tying=1/3`. Both `"lin1"` and `"lin2"` are already keys in the stale 4-entry mapping from "default", so the rebuild never triggers for adapter "b" — its group count gets computed from `num_target_layers=4` (adapter "default"'s count) instead of the correct `num_target_layers=2` (adapter "b"'s own count). With `weight_tying=1/3`, this yields 2 groups instead of the correct 1, i.e. `lin1`/`lin2` silently fail to share a `v` even though `weight_tying` says they should.

`gh pr list --repo huggingface/peft --search "tinylora" --state all` and `gh issue list --repo huggingface/peft --search "tinylora weight_tying" --state all` (and a broader `"tinylora"` issue search) surface only #3024 (original TinyLoRA implementation, merged — this is where `_target_key_to_idx` and the buggy condition originate) and #3180 (fixes non-determinism in the projection seeding by threading `layer_idx` through the layer constructor instead of a `set_layer_idx` setter). I read #3180's diff: it only touches `tinylora/layer.py`'s seeding plumbing and how `layer_idx` is passed into `_create_new_module`/`update_layer`; it does not change the `_target_key_to_idx` rebuild condition at all. No open PR or issue addresses this.

## Tests

Added `tests/test_tinylora.py` with two regression tests:
- `test_second_adapter_overlapping_target_modules_matches_fresh_control`: adds adapter "default" targeting 4 linear layers, then adapter "b" targeting a different (strict subset) but overlapping pair with `weight_tying=1/3`, and asserts adapter "b" ends up with the same number of groups / trainable parameters / shared-`v` structure as a freshly-built control model containing only adapter "b".
- `test_second_adapter_overlapping_target_modules_after_delete_and_readd`: same idea, but through `delete_adapter` + re-`add_adapter` with the same adapter name and different `target_modules`, to cover the same staleness bug reappearing through deletion/recreation rather than only through a second distinct adapter name.

Ran `tests/test_tinylora.py` via pytest: both pass. Verified fail-before / pass-after: reverting only the fix (keeping the new tests) makes both tests fail with exactly the corruption described above (`assert 2 == 1`, i.e. adapter "b" ends up with 2 groups instead of the correct 1); re-applying the fix makes them pass.

Also verified the pre-existing TinyLoRA parametrizations in `tests/test_custom_models.py` (the 8 single-adapter configs under "TinyLoRA", plus the disjoint-target-modules "TinyLoRA Same"/"TinyLoRA Different" two-adapter cases) by exercising the same configs directly against the real library: results (group counts, parameter counts, shared-`v` identity, forward-pass outputs) are identical before and after this fix, confirming no regression to the existing behavior.

## AI assistance disclosure

This PR was prepared with AI assistance (Claude Code). There is no pre-existing issue for this bug — I found it myself while auditing TinyLoRA's multi-adapter handling. I independently verified the root cause (traced `_create_and_replace`'s rebuild condition, confirmed `add_adapter` bypasses `_pre_injection_hook` via `inject_adapter`, read #3180's diff to confirm it never touched this code path), reproduced the corruption and its fix end-to-end against the real library, and reviewed every changed line before proposing this change.


src/peft/tuners/tinylora/model.py
```diff
@@ -130,14 +130,19 @@ def _create_and_replace(
...
-        # This is needed because add_adapter calls inject_adapter directly without _pre_injection_hook.
-        if not hasattr(self, "_target_key_to_idx") or current_key not in self._target_key_to_idx:
-            self._target_key_to_idx = self._build_target_key_mapping(tinylora_config)
+        # Build the target key mapping lazily on first call per injection cycle, keyed by adapter name.
+        # This is needed because add_adapter calls inject_adapter directly without _pre_injection_hook. The mapping
+        # must not be shared across adapters: each adapter can target a different (and possibly overlapping) subset
+        # of modules, so reusing a mapping built for a different adapter would silently corrupt the layer index and
+        # group count used for `weight_tying` whenever `current_key` happens to also appear in that stale mapping.
+        if not hasattr(self, "_target_key_to_idx"):
+            self._target_key_to_idx = {}
```
BenjaminBossan: This attribute could be assigned during `__init__`, right?

tests/test_tinylora.py
```diff
@@ -0,0 +1,113 @@
...
+        X = self.lin1(X)
+        X = self.relu(X)
+        X = self.lin2(X)
+        X = self.relu(X)
+        X = self.lin3(X)
+        X = self.sm(X)
+        return X
+
+
+class TestTinyLoraMultipleAdapters:
```
BenjaminBossan: Let's move the tests to `test_initialization.py`.

BenjaminBossan (changes requested):
Thanks for the PR, some smaller issues but generally the fix is good.

DaoyuanLi2816:
Addressed both review points — moved `_target_key_to_idx` initialization into `__init__` and relocated the new tests into `test_initialization.py`. Ready for another look whenever you have time.

src/peft/tuners/tinylora/model.py
```diff
@@ -281,6 +289,11 @@ def delete_adapter(self, adapter_name: str) -> None:
         if adapter_name in self.tinylora_v:
             del self.tinylora_v[adapter_name]
 
+        # Remove the adapter's cached target-key mapping so that re-adding an adapter with the same name (but
+        # potentially different target_modules) rebuilds the mapping instead of reusing a stale one.
+        if hasattr(self, "_target_key_to_idx") and adapter_name in self._target_key_to_idx:
```
BenjaminBossan: > `if hasattr(self, "_target_key_to_idx")`

Does this need checking? The attribute should always be present, right?

BenjaminBossan (changes requested):
Thanks for the changes, one nit from my side.

src/peft/tuners/tinylora/model.py
DaoyuanLi2816: You're right, it doesn't need checking. `_target_key_to_idx` is unconditionally assigned as the first statement of `TinyLoraModel.__init__`, before `super().__init__()` (which is what triggers the first injection cycle), and there's no subclass or other construction path that would create an instance without going through `__init__`. So by the time `delete_adapter` can be called on an instance, the attribute is always there. Removed the `hasattr` guard in be2f592.

BenjaminBossan:
Please run `make style`.

DaoyuanLi2816:
Ran `make style` as requested — it re-wrapped the new docstring in `tests/test_initialization.py` to better fill the 119-char line length (pure formatting, no logic change). Pushed in b4ca0e7c.

> Please run `make style`.

BenjaminBossan (approved):
Thanks for the last push, PR now LGTM.

PR #3423, opened 2026-07-11
Fix LN Tuning adapter initialization after merge

DaoyuanLi2816:
## Status

Closed because this audit change was not coordinated in advance. I will raise an issue first if it needs revisiting.

## AI assistance disclosure

AI assistance was used. I reviewed the change.

### DIFF
--- a/src/peft/tuners/ln_tuning/layer.py
+++ b/src/peft/tuners/ln_tuning/layer.py
@@ -35,6 +35,11 @@ class LNTuningLayer(nn.Module, BaseTunerLayer):
         super().__init__()
         self.base_layer = base_layer
         self.ln_tuning_layers = nn.ModuleDict({})
+        # Keep an untouched copy of the pretrained layer. `merge` repoints `base_layer` to whichever adapter is
+        # currently merged (see `merge`/`unmerge` below), so `base_layer` cannot be relied on to always hold the
+        # pristine pretrained weights. `original_module` is never touched by merge/unmerge and is therefore what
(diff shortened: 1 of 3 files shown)


ErenAta16 (commented):
Confirmed the mechanism: LN Tuning's `merge` doesn't compute an additive delta like LoRA, it swaps `base_layer` to point at the trained adapter layer directly, so after `merge_adapter()`, `base_layer` holds a *trained* adapter's weights, not the pretrained ones. `_create_new_module` initializing a new adapter from `target.base_layer` therefore picks up whatever adapter happens to be merged at that moment instead of the pristine pretrained weights, silent cross-adapter contamination, not an error, just wrong initial weights for the new adapter.

The fix (a permanent `original_module = deepcopy(base_layer)` snapshot taken at construction, before any merge can happen, and initializing new adapters from that instead of `base_layer`) is the right shape: `original_module` is never touched by `merge`/`unmerge`, so it stays a reliable "true pretrained" reference regardless of merge state. `deepcopy` was already imported in this file, nothing missing there.

The test is precise: trains adapter "default" to a distinctive, obviously-wrong value (999.0), merges it, then adds adapter "adapter2" and asserts its initial weight matches the independently-captured pretrained weight, not 999.0. That's the right way to catch this, a value that couldn't plausibly come from anywhere except the bug being present.

Looks correct.

DaoyuanLi2816:
Closing this uncoordinated audit PR. I will only revisit the change after issue discussion and explicit maintainer approval, per the contribution policy.

PR #2169, opened 2024-10-22
Improving error message when users pass layers_to_transform and layers_pattern

JINO-ROHIT:
Hi @BenjaminBossan Ive reopened with another PR.

This tries to add some verbose message when users pass layers_to_transform and layers_pattern addressing point 3 in https://github.com/huggingface/peft/issues/2155.

PR #3425, opened 2026-07-11
FIX Accept layers_to_transform=0 together with layers_pattern

winklemad:
No linked issue — found by reading the code.

### What this fixes

Every tuner config validates the `layers_pattern` / `layers_to_transform` pairing in `__post_init__` with:

```python
if self.layers_pattern and not self.layers_to_transform:
    raise ValueError("When `layers_pattern` is specified, `layers_to_transform` must also be specified.")
```

Because `0` is falsy, passing the **valid first-layer index** `layers_to_transform=0` together with a `layers_pattern` wrongly raises — even though `layers_to_transform` *was* specified:

```python
from peft import LoraConfig

LoraConfig(target_modules=["q_proj"], layers_to_transform=0, layers_pattern="blocks")
# ValueError: When `layers_pattern` is specified, `layers_to_transform` must also be specified.

LoraConfig(target_modules=["q_proj"], layers_to_transform=1,  layers_pattern="blocks")  # OK
LoraConfig(target_modules=["q_proj"], layers_to_transform=[0], layers_pattern="blocks")  # OK
```

Only index `0` is affected — `1` and `[0]` (the same logical request) are accepted.

### Why `is None` is the correct check

- The consuming matcher already treats `0` as a valid index — `check_target_module_exists` in `src/peft/tuners/tuners_utils.py` gates on `layer_indexes is not None`, and the sibling check a few lines above in the same `__post_init__` also uses `self.layers_to_transform is not None`.
- The docstring states the pattern is *"used only if `layers_to_transform` is different from `None`."*

So the guard should compare against `None`, not truthiness. The check was added in #2159 and its message improved in #2169, but the falsy-`0` case remained.

### Scope

The identical guard is duplicated across **all 19 tuner configs** that support layer indexing (lora, adalora, boft, deft, delora, fourierft, frod, hra, lily, loha, lokr, oft, peanut, psoft, pvera, tinylora, vblora, vera, waveft), so the one-line fix is applied to every one.

### Testing

Added `test_config_layers_to_transform_index_zero`, parametrized over every config in `ALL_CONFIG_CLASSES` that supports `layers_to_transform` (25 configs), asserting `layers_to_transform=0` + `layers_pattern` is accepted. It fails before this change (19 of the configs raise) and passes after. `ruff check` / `ruff format --check` are clean and `tests/test_config.py` passes (1218 passed).

### DIFF
--- a/src/peft/tuners/adalora/config.py
+++ b/src/peft/tuners/adalora/config.py
@@ -88,7 +88,7 @@ class AdaLoraConfig(LoraConfig):
             raise ValueError("`layers_to_transform` cannot be used when `target_modules` is a str.")
 
         # check for layers_to_transform and layers_pattern
-        if self.layers_pattern and not self.layers_to_transform:
+        if self.layers_pattern and self.layers_to_transform is None:
             raise ValueError("When `layers_pattern` is specified, `layers_to_transform` must also be specified. ")
 
         # Check if 'r' has been set to a non-default value
--- a/src/peft/tuners/boft/config.py
+++ b/src/peft/tuners/boft/config.py
@@ -158,7 +158,7 @@ class BOFTConfig(PeftConfig):
             set(self.exclude_modules) if isinstance(self.exclude_modules, list) else self.exclude_modules
         )
         # check for layers_to_transform and layers_pattern
-        if self.layers_pattern and not self.layers_to_transform:
+        if self.layers_pattern and self.layers_to_transform is None:
             raise ValueError("When `layers_pattern` is specified, `layers_to_transform` must also be specified. ")
         if self.boft_block_size == 0 and self.boft_block_num == 0:
             raise ValueError(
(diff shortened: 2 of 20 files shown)


ErenAta16 (commented):
Classic truthiness-vs-None bug: `layers_to_transform=0` selects the first layer and is a perfectly valid, explicit value, but `not self.layers_to_transform` treats `0` the same as "not passed," so specifying `layers_to_transform=0` together with `layers_pattern` wrongly raised "must also be specified" for a value that literally was specified. Same issue would apply to an empty list, though the fix targets the reported `0` case specifically, `is None` is the correct check for "was this actually left at its default."

The fix is applied identically across all ~19 tuner config classes with this exact guard, checked a sample of a few (adalora, boft, delora, hra, waveft) and they're all the same one-line change, no stragglers I could find with the buggy pattern still present.

The test is well-built: rather than hardcoding which config classes support `layers_to_transform`/`layers_pattern`, it uses `inspect.signature` to derive `LAYER_INDEXING_CONFIG_CLASSES` from `ALL_CONFIG_CLASSES` dynamically, so the parametrized regression test automatically covers every current tuner with these params (and future ones, without needing the test updated when a new tuner is added). That's a good way to guard against exactly the kind of "fixed it everywhere except the one file I forgot" gap a fix touching 19 files is prone to.

Looks correct and thorough.

githubnemo (approved):
LGTM, thanks :)

